PR #306, opened 2018-04-05
Increase Watch Interval for E2E Tests

alecmerdler:
### Description

Hopefully this makes them faster by reducing the amount of time it takes to confirm `Deployments` and such are created for Open Cloud Services.

Addresses https://jira.coreos.com/browse/ALM-484


alecmerdler:
closing because I don't think it works

PR #335, opened 2018-05-01
Clarify Status Condition when Catalog Source Not Found

alecmerdler:
### Description

Use `InstallCheckFailed` instead of `DependenciesConflict`.

Addresses https://jira.coreos.com/browse/ALM-572

### DIFF
--- a/pkg/api/apis/installplan/v1alpha1/types.go
+++ b/pkg/api/apis/installplan/v1alpha1/types.go
@@ -63,9 +63,9 @@ type InstallPlanConditionReason string
 
 const (
 	InstallPlanReasonPlanUnknown        InstallPlanConditionReason = "PlanUnknown"
+	InstallPlanReasonInstallCheckFailed InstallPlanConditionReason = "InstallCheckFailed"
 	InstallPlanReasonDependencyConflict InstallPlanConditionReason = "DependenciesConflict"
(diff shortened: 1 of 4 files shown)


pkg/api/apis/installplan/v1alpha1/types.go
```diff
@@ -63,9 +63,9 @@ type InstallPlanConditionReason string
 
 const (
 	InstallPlanReasonPlanUnknown        InstallPlanConditionReason = "PlanUnknown"
+	InstallPlanReasonInstallCheckFailed InstallPlanConditionReason = "InstallCheckFailed"
 	InstallPlanReasonDependencyConflict InstallPlanConditionReason = "DependenciesConflict"
```
alecmerdler: I can't reason about `DependenciesConflict` reason, it doesn't seem to apply to any problem we have yet.

ecordell: There's nothing that stops someone from creating catalogs that have different CSVs that claim to own the same CRD. We check for that when compiling the official catalog for publishing, but this should be the error state for users writing their own. (Longer term we probably want to support resolving these decisions, but for now we'll just error out)

alecmerdler:
@ecordell Fixed the tests

PR #338, opened 2018-05-01
[WIP] Resource Ownership Chain Represented and Validated in CSV

alecmerdler:
### Description

In order to correctly trace the `ownerReferences` chain, the order of `resources` matters (`ownerReferences` to the CR are only written to the top level object) and all levels need to be present.

Example: `VaultService` -> `Deployment` (has refs to `VaultService`) -> `ReplicaSet` -> `Pod`

Addresses https://jira.coreos.com/browse/ALM-563


ecordell:
We'll revisit this later

PR #342, opened 2018-05-04
Rebased OSB backbone

ecordell:
See: https://github.com/operator-framework/operator-lifecycle-manager/pull/299


Makefile
```diff
@@ -1,67 +1,102 @@
...
 
-coverage-html: test
+coverage-html: cover.out
 	go tool cover -html=cover.out
 
-run-local: update-catalog
-	. ./scripts/package-release.sh ver=1.0.0-local Documentation/install/resources Documentation/install/local-values.yaml
+build: $(CMDS)
+
+$(CMDS): .FORCE
```
ericavonb: You can delete the builds to trigger a rebuild, you like you can add `clean` as a pre-req to any target that you want to have a full rebuild for.

ecordell: I tried `make clean build` in the dockerfiles and it was still saying `nothing to build`? not sure if I was doing something wrong

ericavonb: Hu. I don't know. That works for me locally. There's always `make clean && make build` in the Dockerfile. I was thinking like having `build: clean $(CMDS)`, then `make build` should rebuild.

PR #345, opened 2018-05-08
test validate service broker api version function

ericavonb:
Add test of the implementation to validate supported versions of the Open Service Broker as per the service broker interface. All current versions (2.11, 2.12, and 2.13) are supported.

### DIFF
--- a/pkg/server/servicebroker/apiserver_test.go
+++ b/pkg/server/servicebroker/apiserver_test.go
@@ -1,0 +1,35 @@
+package servicebroker
+
+import (
+	"testing"
+
+	"github.com/golang/mock/gomock"
+	"github.com/stretchr/testify/require"
+
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/clientset/versioned/fake"
+)
+
+const (
+	testNamespace = "testns"
+)
+
+func mockALMBroker(ctrl *gomock.Controller) *ALMBroker {
+	return &ALMBroker{
+		client:    fake.NewSimpleClientset(),
+		namespace: testNamespace,
+	}
+}
+
+func TestValidateBrokerAPIVersion(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+	// test bad version
+	brokerMock := mockALMBroker(ctrl)
+	err := brokerMock.ValidateBrokerAPIVersion("oops")
+	require.EqualError(t, err, "unknown OpenServiceBroker API Version: oops")
+
+	// supported version
+	require.NoError(t, brokerMock.ValidateBrokerAPIVersion("2.11"))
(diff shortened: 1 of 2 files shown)


pkg/server/servicebroker/apiserver_test.go
```diff
@@ -23,8 +23,15 @@ func mockALMBroker(ctrl *gomock.Controller) *ALMBroker {
 func TestValidateBrokerAPIVersion(t *testing.T) {
 	ctrl := gomock.NewController(t)
 	defer ctrl.Finish()
-	err := mockALMBroker(ctrl).ValidateBrokerAPIVersion("n/a")
-	require.EqualError(t, err, "not implemented")
+	// test bad version
+	brokerMock := mockALMBroker(ctrl)
```
ecordell: Can these be done without the operatorclient mocks? It's just more work to remove operator client if we use them in new code

ericavonb: These use whatever is in the generated clientsets:
https://github.com/operator-framework/operator-lifecycle-manager/pull/345/files/d9dd9c928c52c1d3c32ff2a9cff4717f63da804e#diff-3a028458a1ff8c8a41758933f6670271L9

Do those need to be updated?

ecordell: nope, that's good - I just saw mock and assumed it was using the MockOperatorClient!

pkg/server/servicebroker/apiserver_test.go
```diff
@@ -0,0 +1,35 @@
...
+	"github.com/stretchr/testify/require"
+
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/clientset/versioned/fake"
+)
+
+const (
+	testNamespace = "testns"
+)
+
+func mockALMBroker(ctrl *gomock.Controller) *ALMBroker {
```
ecordell: you can remove ctrl, it's unused

pkg/server/servicebroker/apiserver_test.go
```diff
@@ -0,0 +1,35 @@
...
+	return &ALMBroker{
+		client:    fake.NewSimpleClientset(),
+		namespace: testNamespace,
+	}
+}
+
+func TestValidateBrokerAPIVersion(t *testing.T) {
+	ctrl := gomock.NewController(t)
```
ecordell: same here, you don't need ctrl or the defer

pkg/server/servicebroker/apiserver_test.go
ericavonb: f2f98478

pkg/server/servicebroker/apiserver_test.go
ericavonb: f2f98478

PR #346, opened 2018-05-08
Strip Markdown Descriptions from OSB

alecmerdler:
### Description

Uses https://github.com/writeas/go-strip-markdown to strip Markdown text when building an OSB Service from a `ClusterServiceVersion`.

Addresses https://jira.coreos.com/browse/ALM-567

### DIFF
--- a/pkg/server/servicebroker/types_test.go
+++ b/pkg/server/servicebroker/types_test.go
@@ -1,0 +1,33 @@
+package servicebroker
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	csvv1alpha1 "github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/clusterserviceversion/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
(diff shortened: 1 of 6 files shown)


alecmerdler:
@ecordell Added unit test for description parsing.

pkg/server/servicebroker/types_test.go
```diff
@@ -0,0 +1,33 @@
...
+}
+
+func TestCSVToServiceParsesDescription(t *testing.T) {
+	var parseDesc = func(desc string) string {
+		require.Equal(t, mockCSV().Spec.Description, desc)
+
+		return ""
+	}
+
+	_, err := csvToService(mockCSV(), parseDesc)
```
ecordell: if you add some markdown to the CSV description and write an expectation that it doesn't have markdown here, this looks good to me

alecmerdler: I made `ParseDescription` a function interface that is passed to `csvToService` so we can pass a mock and assert it was called. Testing that Markdown is stripped is the responsibility of the `go-strip-markdown` dependency.

PR #347, opened 2018-05-08
Manual Approval of Install Plans

alecmerdler:
### Description

Implements install plan manual approval.

Addresses https://jira.coreos.com/browse/ALM-339

### DIFF
--- a/deploy/chart/templates/06-installplan.crd.yaml
+++ b/deploy/chart/templates/06-installplan.crd.yaml
@@ -41,3 +41,18 @@ spec:
               - Automatic
               - Manual
               - Update-Only # Will only apply an update if it updates existing packages only and doesn't add any new ones
+            approved:
+              type: boolean
+      anyOf:
+        - properties:
+          approval: 
+            enum: 
+              - Manual
+          required:
+            - approved
+        - properties:
+          approval:
+            enum:
+              - Automatic
+              - Update-Only
+          required: []
--- a/pkg/api/apis/installplan/v1alpha1/types.go
+++ b/pkg/api/apis/installplan/v1alpha1/types.go
@@ -35,6 +35,7 @@ const (
 type InstallPlanSpec struct {
 	ClusterServiceVersionNames []string `json:"clusterServiceVersionNames"`
 	Approval                   Approval `json:"approval"`
+	Approved                   bool     `json:"approved"`
 }
 
 // InstallPlanPhase is the current status of a InstallPlan as a whole.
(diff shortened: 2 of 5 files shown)


pkg/controller/operators/catalog/operator_test.go
```diff
@@ -54,25 +54,35 @@ func TestTransitionInstallPlan(t *testing.T) {
...
 		transError error
+		approval   v1alpha1.Approval
+		approved   bool
 		expected   v1alpha1.InstallPlanPhase
 		condition  *v1alpha1.InstallPlanCondition
 	}{
-		{v1alpha1.InstallPlanPhaseNone, nil, v1alpha1.InstallPlanPhasePlanning, nil},
-		{v1alpha1.InstallPlanPhaseNone, err, v1alpha1.InstallPlanPhasePlanning, nil},
+		{v1alpha1.InstallPlanPhaseNone, nil, v1alpha1.ApprovalAutomatic, false, v1alpha1.InstallPlanPhasePlanning, nil},
+		{v1alpha1.InstallPlanPhaseNone, err, v1alpha1.ApprovalAutomatic, false, v1alpha1.InstallPlanPhasePlanning, nil},
```
ecordell: just for completeness: `ApprovalAutomatic` + `approved = true` should go to planning as well

pkg/api/apis/installplan/v1alpha1/types.go
```diff
@@ -35,6 +35,7 @@ const (
 type InstallPlanSpec struct {
 	ClusterServiceVersionNames []string `json:"clusterServiceVersionNames"`
 	Approval                   Approval `json:"approval"`
+	Approved                   bool     `json:"approved"`
```
ecordell: This needs to be propagated to the CRD schema as well

alecmerdler: Which is the correct file with the `InstallPlan-v1` CRD? And how do I express that `approved` is only required if `approval: Manual`?

ecordell: The json-schema way is:

```json
  "dependencies": {
    "foo": ["bar"]
  }
```
(if foo is present, bar is required)

but IIRC the subset used for CRD validation doesn't allow that. It's worth checking, at least.

The canonical `InstallPlan` lives in `deploy/chart/templates`

deploy/chart/templates/06-installplan.crd.yaml
```diff
@@ -41,3 +41,18 @@ spec:
               - Automatic
               - Manual
               - Update-Only # Will only apply an update if it updates existing packages only and doesn't add any new ones
+            approved:
+              type: boolean
+      anyOf:
```
alecmerdler: Found this here: https://stackoverflow.com/questions/38717933/jsonschema-attribute-conditionally-required/38781027

ecordell: We use that for the install strategy on CSV as well 👍

PR #348, opened 2018-05-09
Add installPlanApproval to Subscription-v1

ecordell:
The `installPlanApproval` field on Subscription will dictate the `approval` field of the InstallPlans that get created by the Subscription.

Also removed the `Upgrade-Only` enum option, until we decide that it's valuable and implement it.


pkg/api/apis/subscription/v1alpha1/types.go
```diff
@@ -64,3 +65,11 @@ type SubscriptionList struct {
 
 	Items []Subscription `json:"items"`
 }
+
+// GetInstallPlanApproval gets the configured install plan approval or the default
+func (s *Subscription) GetInstallPlanApproval() v1alpha1.Approval {
+	if s.Spec.InstallPlanApproval == v1alpha1.ApprovalManual {
```
alecmerdler: I know this works now because there are only 2 possible values, but it should really be:
```go
if s.Spec.InstallPlanApproval == nil {
        return v1alpha1.ApprovalAutomatic
}
return s.Spec.InstallPlanApproval
```

ecordell: I did it this way so that any invalid approvals (`approval: foo` etc) will get defaulted to `Automatic`. If we ever get to more than two approval types I would change this to set logic (if approval not in set_of_allowed_values, return default)

PR #351, opened 2018-05-15
Add Operator metering component to tectonic-feature catalog

chancez:
Adds the operator-metering component to the tectonic-feature catalog. This isn't tectonic specific, and in-fact, this wouldn't work by default on Tectonic, so this location is temporarily until we have multiple catalogs.


ecordell:
What's your goal with this? `tectonic-features` is one of two catalogs shipping with tectonic 1.9, so we do have multiple catalogs already.

If you just want validation of your manifests and such we're working on some tooling for another repo that will do that.

PR #353, opened 2018-05-17
test(servicebroker): unit test GetCatalog endpoint

ericavonb:


pkg/server/servicebroker/apiserver_test.go
```diff
@@ -82,102 +486,168 @@ func TestGetCatalog(t *testing.T) {
 		inputs      args
 		expect      output
 	}{
+		// Loads and Converts Packages to Services
```
ecordell: I was expecting to see the positive test cases where the catalog is actually loaded (like in the commented outline of the tests)?

ericavonb: I realized once I had pulled the loading into an interface that those tests wouldn't really add value. The `ListCatalogs` function does not depend on how the loading is done, so it shouldn't be tested in its unit tests.

I then looked into unit tests for the catalogLoader implementation. That would require fixing the fake catalogSource client Lister somehow or adding another layer of interfaces, which is annoying. Also, loading the fake clients and checking the results would involve basically repeating the logic it's trying to test.

I think the e2e tests would be a more appropriate place to test the catalogsource/configmaps to service catalog flow.

ecordell: But this is a test for `GetCatalog`, which converts an OLM catalog to an OSB Catalog. I don't see why we can't add an additional test case here where the input configmap has packages/csvs (and no errors) and the output response has the expected services. What am I missing?

PR #361, opened 2018-05-30
Tests for Runaway Control Loops

alecmerdler:
### Description

Adds checks in the e2e tests to ensure that the Operators are not rapidly processing OLM custom resources. 

Addresses https://jira.coreos.com/browse/ALM-581

### DIFF
--- a/test/e2e/csv_e2e_test.go
+++ b/test/e2e/csv_e2e_test.go
@@ -295,13 +295,18 @@ func TestCreateCSVRequirementsMet(t *testing.T) {
 	require.NoError(t, err)
 	defer cleanupCSV()
 
-	_, err = fetchCSV(t, c, csv.Name, csvSucceededChecker)
+	fetchedCSV, err := fetchCSV(t, c, csv.Name, csvSucceededChecker)
 	require.NoError(t, err)
 
 	// Should create deployment
 	dep, err := c.GetDeployment(testNamespace, strategy.DeploymentSpecs[0].Name)
 	require.NoError(t, err)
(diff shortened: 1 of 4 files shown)


test/e2e/csv_e2e_test.go
```diff
@@ -116,6 +116,12 @@ var csvReplacingChecker = func(csv *v1alpha1.ClusterServiceVersion) bool {
 	return csv.Status.Phase == v1alpha1.CSVPhaseReplacing || csv.Status.Phase == v1alpha1.CSVPhaseDeleting
 }
 
+var csvResourceVersionCheckerFor = func(csv *v1alpha1.ClusterServiceVersion) csvConditionChecker {
+	return func(fetchedCSV *v1alpha1.ClusterServiceVersion) bool {
+		return csv.GetResourceVersion() == fetchedCSV.GetResourceVersion()
```
alecmerdler: @ecordell what do you think of this approach to catching regressions like this?

ecordell: This will work if we're careful to use it correctly. Calling it twice in a row probably won't find the issues, but calling it once, waiting for a period of time that we don't expect changes to have occurred, and then calling it again will.

ecordell: I'm thinking we could use defer to our advantage here, like whenever we have an object hit a known state that we don't expect to change, go ahead and defer a check for the resourceVersion.

If that's not enough, we could keep a registry of resourceVersionChecks for the entire testsuite, and then check them as a final test.

alecmerdler:
@ecordell The last test run passed, I just removed the `TODO`s, so this is ready for review.

PR #362, opened 2018-05-31
Fix memory usage in catalog operator

ecordell:
Also scopes the catalog operator to only watch the "catalogNamespace" for catalog sources (it was watching all before).

Mem usage before:
screen shot 2018-05-31 at 9 26 34 am

Mem usage after:
screen shot 2018-05-31 at 6 12 54 pm

### DIFF
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -58,15 +58,20 @@ func NewOperator(kubeconfigPath string, wakeupInterval time.Duration, operatorNa
 		return nil, err
 	}
 
-	sharedInformerFactory := externalversions.NewSharedInformerFactory(crClient, wakeupInterval)
-
-	// Create an informer for each namespace.
+	// Create an informer for each watched namespace.
 	ipSharedIndexInformers := []cache.SharedIndexInformer{}
 	subSharedIndexInformers := []cache.SharedIndexInformer{}
 	for _, namespace := range watchedNamespaces {
 		nsInformerFactory := externalversions.NewFilteredSharedInformerFactory(crClient, wakeupInterval, namespace, nil)
 		ipSharedIndexInformers = append(ipSharedIndexInformers, nsInformerFactory.Installplan().V1alpha1().InstallPlans().Informer())
 		subSharedIndexInformers = append(subSharedIndexInformers, nsInformerFactory.Subscription().V1alpha1().Subscriptions().Informer())
+	}
+
+	// Create an informer for each catalog namespace
+	catsrcSharedIndexInformers := []cache.SharedIndexInformer{}
+	for _, namespace := range []string{operatorNamespace} {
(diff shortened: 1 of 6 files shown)


ecordell:
(Pipeline is passing now, status is just slow to update)

pkg/controller/operators/catalog/operator.go
```diff
@@ -69,6 +67,13 @@ func NewOperator(kubeconfigPath string, wakeupInterval time.Duration, operatorNa
 		subSharedIndexInformers = append(subSharedIndexInformers, nsInformerFactory.Subscription().V1alpha1().Subscriptions().Informer())
 	}
 
+	// Create an informer for each catalog namespace
+	catsrcSharedIndexInformers := []cache.SharedIndexInformer{}
+	for _, namespace := range []string{operatorNamespace} {
```
alecmerdler: Why `range` over a slice with a single item?

pkg/controller/operators/catalog/operator.go
```diff
@@ -139,7 +142,7 @@ func (o *Operator) syncCatalogSources(obj interface{}) (syncError error) {
 	defer o.sourcesLock.Unlock()
 	o.sources[catsrc.GetName()] = src
 	o.sourcesLastUpdate = timeNow()
-	return err
+	return nil
```
alecmerdler: We mix this convention in the codebase (we know `err == nil`, but return `err` anyway). Which is preferred?

pkg/controller/registry/configmap_loader.go
```diff
@@ -21,23 +21,22 @@ const (
 
 // ConfigMapCatalogResourceLoader loads a ConfigMap of resources into the in-memory catalog
 type ConfigMapCatalogResourceLoader struct {
-	Catalog   *InMem
```
alecmerdler: Why the change in this interface?

pkg/controller/operators/catalog/operator.go
ecordell: Just because I know we'll implement "local" catalogs in the near future, where we'll want to iterate over `operatorNamespace ∪ watchedNamespaces` instead of just `operatorNamespace`

pkg/controller/operators/catalog/operator.go
ecordell: I feel this is clearer, but I don't know if there's a golang convention for this. Which do you think is more readable?

pkg/controller/registry/configmap_loader.go
ecordell: This is fixing the memory leak, since the catalog returned was a field of the loader, and the loader holds a reference to part of the operator (operator client), every time we sync the catalog we were loading into memory and retaining it indefinitely. 

We might still be leaking the loader itself now that you mention it...but at least that's a tiny amount instead of an entire in memory catalog

PR #364, opened 2018-06-07
Resolve Install Plan before Requiring Approval

alecmerdler:
### Description

Fixes regression introduced which prevented `status.plan` from being filled before a manual install plan is approved. Also updates docs to include the `RequiresApproval` phase.

Addresses https://jira.coreos.com/browse/ALM-623

### DIFF
--- a/pkg/api/apis/installplan/v1alpha1/types.go
+++ b/pkg/api/apis/installplan/v1alpha1/types.go
@@ -54,7 +54,6 @@ type InstallPlanConditionType string
 
 const (
 	InstallPlanResolved  InstallPlanConditionType = "Resolved"
-	InstallPlanApproved  InstallPlanConditionType = "Approved"
 	InstallPlanInstalled InstallPlanConditionType = "Installed"
 )
 
--- a/test/e2e/alm_e2e_test.go
+++ b/test/e2e/alm_e2e_test.go
@@ -114,7 +114,7 @@ func TestCreateInstallPlanManualApproval(t *testing.T) {
 		},
 	}
 
-	// Create a new installplan for vault with manual approval
+	// Create a new InstallPlan for Vault with manual approval
 	cleanup, err := decorateCommonAndCreateInstallPlan(c, vaultInstallPlan)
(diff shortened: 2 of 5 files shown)


pkg/api/apis/installplan/v1alpha1/types.go
```diff
@@ -54,7 +54,6 @@ type InstallPlanConditionType string
 
 const (
 	InstallPlanResolved  InstallPlanConditionType = "Resolved"
-	InstallPlanApproved  InstallPlanConditionType = "Approved"
```
alecmerdler: Unused, should be determined from `spec.approved`.

alecmerdler:
The e2e tests are failing because of `the server could not find the requested resource` when trying to fetch the resources in `status.plan`, which is what we want because the `InstallPlan` is not approved. This test was probably passing before because `status.plan` was empty due to the regression, so we never tried to fetch anything.

test/e2e/alm_e2e_test.go
```diff
@@ -132,21 +132,22 @@ func TestCreateInstallPlanManualApproval(t *testing.T) {
 		if step.Resource.Kind == "CustomResourceDefinition" {
 			_, err := c.GetCustomResourceDefinition(step.Resource.Name)
 
-			require.NoError(t, err)
+			require.Error(t, err)
 			vaultResourcesPresent = vaultResourcesPresent + 1
```
alecmerdler: @ecordell This fails because our CI target cluster always has all the CRDs. Should we just disable this check?

ecordell: It looks like it should be finding the already installed CRDs and counting them, right? I wouldn't expect this to fail

ecordell:
@alecmerdler For the e2e tests, we want them to be aware of the approval state:

1. if plan is resolved but not approved, verify that the resources in the plan haven't been created (as you pointed out we won't be able to do that for the CRDs)
2. if the plan is resolved and approved, verify that all of the resources in the plan exist

PR #365, opened 2018-06-08
Separate Manifest Validator

alecmerdler:
### Description

Moves the catalog validation logic out of `test/schema` and into a separate binary, `cmd/validator`. Makes the necessary changes to be run using `go run validator` instead of `go test ./test/schema`.

Added to `operator-manifests` CI in https://github.com/operator-framework/operator-manifests/pull/2.

Addresses https://jira.coreos.com/browse/ALM-593


cmd/validator/main.go
```diff
@@ -0,0 +1,17 @@
...
+
+	"github.com/operator-framework/operator-lifecycle-manager/cmd/validator/schema"
+)
+
+func main() {
+	// TODO(alecmerdler): Get manifest directory from args
+	manifestDir := os.Args[1]
+
+	// FIXME(alecmerdler): `TestCatalogVersions` is meant to run against built catalog configmaps
+	// schema.TestCatalogVersions(manifestDir)
```
ecordell: We can simplify this check to simply: `for all manifests with replaces field: check that csv being replaced actually exists`. That should make it easier to implement (and generalize for operator-manifests)

Makefile
```diff
@@ -143,7 +143,8 @@ codegen: $(CODEGEN)
 verify-codegen: codegen
 	git diff --exit-code
 
-verify-catalog: schema-check
+verify-catalog:
+	go test -v ./test/schema/catalog_versions_test.go
```
alecmerdler: Verifying catalog version consistency is a unique problem for OLM, and does not need to be part of the separate binary.

Makefile
```diff
@@ -18,7 +18,7 @@ all: test build
 test: schema-check cover.out
 
 schema-check:
-	go test -v ./test/schema
+	go run ./cmd/validator/main.go ./deploy/chart/catalog_resources
```
alecmerdler: `go run validator` accepts a path to a directory of unique catalogs or packages, but not versions of the same catalog.

deploy/chart/catalog_resources/components/chargeback.package.yaml
```diff
@@ -1,4 +1,4 @@
-#! package-manifest: ../../deploy/chart/catalog_resources/components/chargeback.v0.5.1.clusterserviceversion.yaml
+#! package-manifest: ./deploy/chart/catalog_resources/components/chargeback.v0.5.1.clusterserviceversion.yaml
```
alecmerdler: Changed the paths to be relative to project root, where `go run validator` is executed.

alecmerdler:
@ecordell I think this is ready for review (I removed all the `panic(err)` calls).

cmd/validator/schema/upgrade_path.go
```diff
@@ -0,0 +1,65 @@
...
+			return nil, err
+		}
+		matching[name] = bytes
+	}
+
+	return matching, nil
+}
+
+// TestUpgradePath checks that every ClusterServiceVersion in a package directory has a valid `spec.replaces` field.
+func TestUpgradePath(packageDir string) error {
```
ecordell: nit: kind of weird to have a non-test function start with `Test`

Makefile
```diff
@@ -143,7 +143,8 @@ codegen: $(CODEGEN)
 verify-codegen: codegen
 	git diff --exit-code
 
-verify-catalog: schema-check
+verify-catalog:
```
ecordell: we still want to call `schema-check` here too, right?

cmd/validator/schema/schema.go
```diff
@@ -201,93 +221,100 @@ func ValidateCRD(t *testing.T, schemaFileName string, fileBytes []byte) error {
...
-		require.NoError(t, ValidateResource(t, path, f, err))
 	})
-	return nil
+	return err
 }
 
-func TestCatalogResources(t *testing.T) {
-	directoryTester := DirectoryResourceValidator{t}
-	directoryTester.ValidateResources(os.Getenv("GOPATH") + "/src/github.com/operator-framework/operator-lifecycle-manager/deploy/chart/catalog_resources")
+func TestCatalogResources(manifestDir string) error {
```
ecordell: Same here, maybe `CheckCatalogResources` or `ValidateCatalogResources` instead?

Issue #360, opened 2018-05-30
make build fails


tkashem:
`make build` command fails with an error
Steps to reproduce
1. get latest from master branch
2. make build

I am seeing the following error -
```
operator-framework/operator-lifecycle-manager (master) $ make build
if [ 1 = 1true ]; then \
	GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go test -o bin/alm -c -covermode=count -coverpkg ./pkg/... github.com/operator-framework/operator-lifecycle-manager/cmd/alm; \
else \
	GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build -o bin/alm github.com/operator-framework/operator-lifecycle-manager/cmd/alm; \
fi
vendor/github.com/emicklei/go-restful/container.go:17:2: cannot find package "github.com/emicklei/go-restful/log" in any of:
	/{$GOPATH}/src/github.com/operator-framework/operator-lifecycle-manager/vendor/github.com/emicklei/go-restful/log (vendor tree)
	/usr/lib/golang/src/github.com/emicklei/go-restful/log (from $GOROOT)
	/{$GOPATH}/src/github.com/emicklei/go-restful/log (from $GOPATH)
make: *** [Makefile:40: bin/alm] Error 1
```
The sub package `'github.com/emicklei/go-restful/log'` is missing from the vendor tree. 
`make vendor` should fix it, but I am running into another issue - `dep ensure -v -vendor-only` seems to hang indefinitely on my local workstation.

ecordell:
Is `operator-framework/operator-lifecycle-manager` in your gopath? The dependency is checked into the repo so you shouldn't see different dependencies.

Check to see if this exists:
`/{$GOPATH}/src/github.com/operator-framework/operator-lifecycle-manager/vendor/github.com/emicklei/go-restful/log`

You also just double check that you checked out olm into `/{$GOPATH}/src/github.com/operator-framework/operator-lifecycle-manager/`

tkashem:
Yes, `operator-framework/operator-lifecycle-manager` is in my GOPATH.
I have double checked - olm is checked out into the right path.

Looks like `log` sub-package is missing here - `https://github.com/operator-framework/operator-lifecycle-manager/tree/master/vendor/github.com/emicklei/go-restful`

Thanks for getting back to me.

ecordell:
[...]

PR #370, opened 2018-06-13
Remove tectonic-operators

ecordell:
Took the path of least-resistance for this PR:

- Copied over the two parts of `operator-client` that don't have good tooling in client-go:
 - Deployments (`CreateOrRollingDeployment`)
 - CustomResource interactions
- Kept an `operatorclient` abstraction to hang our test mocks off of

Doing things this way required very few changes to our tests, which was the hard part from last time we attempted this.

I expect that in the future, operator-sdk and/or client-go will make all of this custom client work obsolete, so we should plan to update to the best solution when it's available.


alecmerdler:
How should I review this?

ecordell:
I'm mainly interested to make sure none of the changes affected the e2e tests. Another thing to check is that this builds on a fresh clone, to verify https://github.com/operator-framework/operator-lifecycle-manager/issues/360 is fixed.

PR #380, opened 2018-07-11
Add CatalogSource Namespace to Subscription Objects

alecmerdler:
### Description

Makes it possible to find the contents of a `CatalogSource-v1` without needing permission to fetch in _all namespaces_. 

Fixes https://jira.coreos.com/browse/ALM-639

### DIFF
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -34,13 +34,18 @@ const (
 //for test stubbing and for ensuring standardization of timezones to UTC
 var timeNow = func() metav1.Time { return metav1.NewTime(time.Now().UTC()) }
 
+type sourceKey struct {
+	name      string
+	namespace string
+}
+
 // Operator represents a Kubernetes operator that executes InstallPlans by
 // resolving dependencies in a catalog.
 type Operator struct {
 	*queueinformer.Operator
 	client            versioned.Interface
 	namespace         string
-	sources           map[string]registry.Source
+	sources           map[sourceKey]registry.Source
 	sourcesLock       sync.Mutex
 	sourcesLastUpdate metav1.Time
 }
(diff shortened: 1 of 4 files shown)


pkg/controller/operators/catalog/operator.go
```diff
@@ -34,13 +34,18 @@ const (
...
+}
+
 // Operator represents a Kubernetes operator that executes InstallPlans by
 // resolving dependencies in a catalog.
 type Operator struct {
 	*queueinformer.Operator
 	client            versioned.Interface
 	namespace         string
-	sources           map[string]registry.Source
+	sources           map[sourceKey]registry.Source
```
alecmerdler: Use complex `struct` which contains both name and namespace as map key.

ecordell:
Why don't we make this backwards compatible by saying `if CatalogSourceNamespace == ""; CatalogSourceNamespace = <global catalog namespace from command line args>`?

pkg/controller/operators/catalog/operator.go
ecordell: :+1: for compound keys

alecmerdler:
@ecordell Tests are passing, made it backwards compatible.

pkg/controller/operators/catalog/operator.go
```diff
@@ -270,14 +275,14 @@ func (o *Operator) ResolvePlan(plan *v1alpha1.InstallPlan) error {
 	var notFoundErr error
 	for sourceName, source := range o.sources {
```
ecordell: nit: `for key, source`

pkg/controller/operators/catalog/operator.go
```diff
@@ -270,14 +275,14 @@ func (o *Operator) ResolvePlan(plan *v1alpha1.InstallPlan) error {
...
-		notFoundErr = resolveInstallPlan(sourceName, source, plan)
+		plan.EnsureCatalogSource(sourceName.name)
+		notFoundErr = resolveInstallPlan(sourceName.name, source, plan)
 		if notFoundErr != nil {
 			continue
 		}
 
 		// Look up the CatalogSource.
-		catsrc, err := o.client.CatalogsourceV1alpha1().CatalogSources(o.namespace).Get(sourceName, metav1.GetOptions{})
+		catsrc, err := o.client.CatalogsourceV1alpha1().CatalogSources(o.namespace).Get(sourceName.name, metav1.GetOptions{})
```
ecordell: should be `CatalogSources(sourceName.namespace)` just in case?

PR #382, opened 2018-07-12
fix(docs): correct spelling of philosopy.md

aravindhp:
This has the added benefit of fixing the link in README.md


ecordell:
Nice catch! Thanks!

PR #386, opened 2018-07-18
feat(catalog): multiple CatalogSource resolution

njhale:
### Description

Adds multi-catalog resolution of CSVs and CRDs.

Addresses [ALM-641](https://jira.coreos.com/browse/ALM-642)

### DIFF
--- a/pkg/api/apis/installplan/v1alpha1/types.go
+++ b/pkg/api/apis/installplan/v1alpha1/types.go
@@ -33,6 +33,8 @@ const (
 
 // InstallPlanSpec defines a set of Application resources to be installed
 type InstallPlanSpec struct {
+	CatalogSource              string   `json:"sourceName"`
+	CatalogSourceNamespace     string   `json:"sourceNamespace"`
 	ClusterServiceVersionNames []string `json:"clusterServiceVersionNames"`
 	Approval                   Approval `json:"approval"`
 	Approved                   bool     `json:"approved"`
--- a/pkg/api/apis/installplan/v1alpha1/types.go
+++ b/pkg/api/apis/installplan/v1alpha1/types.go
@@ -12,6 +12,7 @@ import (
 	k8sjson "k8s.io/apimachinery/pkg/runtime/serializer/json"
 
 	csvv1alpha1 "github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/clusterserviceversion/v1alpha1"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/registry"
 )
 
 const (
@@ -89,7 +90,7 @@ var ErrInvalidInstallPlan = errors.New("the InstallPlan contains invalid data")
 type InstallPlanStatus struct {
 	Phase          InstallPlanPhase       `json:"phase"`
 	Conditions     []InstallPlanCondition `json:"conditions,omitempty"`
-	CatalogSources []string               `json:"catalogSources"`
+	CatalogSources []registry.SourceKey   `json:"catalogSources"`
 	Plan           []Step                 `json:"plan,omitempty"`
 }
 
@@ -218,14 +219,14 @@ type InstallPlan struct {
 
 // EnsureCatalogSource ensures that a CatalogSource is present in the Status
 // block of an InstallPlan.
-func (p *InstallPlan) EnsureCatalogSource(catalogSourceName string) {
-	for _, source := range p.Status.CatalogSources {
-		if source == catalogSourceName {
+func (p *InstallPlan) EnsureCatalogSource(sourceKey registry.SourceKey) {
+	for _, srcKey := range p.Status.CatalogSources {
+		if srcKey == sourceKey {
 			return
 		}
 	}
 
-	p.Status.CatalogSources = append(p.Status.CatalogSources, catalogSourceName)
+	p.Status.CatalogSources = append(p.Status.CatalogSources, sourceKey)
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
--- a/pkg/api/apis/installplan/v1alpha1/types.go
+++ b/pkg/api/apis/installplan/v1alpha1/types.go
@@ -33,7 +33,7 @@ const (
 
 // InstallPlanSpec defines a set of Application resources to be installed
 type InstallPlanSpec struct {
-	CatalogSource              string   `json:"sourceName"`
+	CatalogSource              string   `json:"source"`
 	CatalogSourceNamespace     string   `json:"sourceNamespace"`
 	ClusterServiceVersionNames []string `json:"clusterServiceVersionNames"`
 	Approval                   Approval `json:"approval"`
--- a/pkg/api/apis/installplan/v1alpha1/types.go
+++ b/pkg/api/apis/installplan/v1alpha1/types.go
@@ -153,12 +153,13 @@ type Step struct {
 // StepResource represents the status of a resource to be tracked by an
 // InstallPlan.
 type StepResource struct {
-	CatalogSource string `json:"sourceName"`
-	Group         string `json:"group"`
-	Version       string `json:"version"`
-	Kind          string `json:"kind"`
-	Name          string `json:"name"`
-	Manifest      string `json:"manifest,omitempty"`
+	CatalogSource          string `json:"sourceName"`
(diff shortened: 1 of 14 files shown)


pkg/controller/registry/resolver/resolver_test.go
```diff
@@ -37,56 +37,202 @@ func TestSingleSourceResolveInstallPlan(t *testing.T) {
...
+		name   string
+		srcKey registry.SourceKey
+	}
+	var table = []struct {
+		description string
+		planCSVName string
+		csvs        map[string]csvName
+		crds        map[string]crdName
+		srcKeys     []registry.SourceKey
+		expectedErr error
```
ecordell: Instead of building the expected steps from the input csvs/crds, let's just have a list of expected steps (can include secrets then, too)?

pkg/controller/operators/catalog/operator.go
```diff
@@ -269,35 +269,46 @@ func (o *Operator) ResolvePlan(plan *v1alpha1.InstallPlan) error {
...
-	var steps []v1alpha1.Step
+	// Set a first source key
+	var firstSrcKey registry.SourceKey
 	for srcKey := range sourcesSnapshot {
-		log.Debugf("resolving against source %v", srcKey)
-		plan.EnsureCatalogSource(srcKey.Name)
-		steps, notFoundErr = o.dependencyResolver.ResolveInstallPlan(sourcesSnapshot, srcKey, CatalogLabel, plan)
-		if notFoundErr != nil {
-			continue
+		// Get the first key (arbitrary)
```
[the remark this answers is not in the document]
njhale: Not that I can see. Could we have a discussion on what the criteria of a desired first catalog is?

pkg/controller/registry/resolver/resolver_test.go
njhale: In order to test ResolveInstallPlan(...), those fields are being used to populate one or more sources with CSVs and CRDs to be resolved.

pkg/controller/registry/resolver/resolver_test.go
```diff
@@ -37,56 +37,202 @@ func TestSingleSourceResolveInstallPlan(t *testing.T) {
...
+		required []string
+		srcKey   registry.SourceKey
+	}
+	type crdName struct {
+		name   string
+		srcKey registry.SourceKey
+	}
+	var table = []struct {
+		description string
+		planCSVName string
```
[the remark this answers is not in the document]
njhale: planCSVName specifies the main CSV the InstallPlan must resolve for each test in the test table. If we remove and hard-code we lose some flexibility with the naming of that CSV. (I'm wondering if it's worth taking out)

pkg/controller/registry/resolver/resolver_test.go
njhale: So then I would build those csvs/crds from the expected steps, store them in their respective sources, and ResolveInstallPlan from there?

pkg/api/apis/installplan/v1alpha1/types.go
```diff
@@ -153,12 +153,13 @@ type Step struct {
...
 // InstallPlan.
 type StepResource struct {
-	CatalogSource string `json:"sourceName"`
-	Group         string `json:"group"`
-	Version       string `json:"version"`
-	Kind          string `json:"kind"`
-	Name          string `json:"name"`
-	Manifest      string `json:"manifest,omitempty"`
+	CatalogSource          string `json:"sourceName"`
+	CatalogSourceNamespace string `json:"sourceNamespace"`
```
alecmerdler: Can we change `status.catalogSources` to be a list with both the `name` and `namespace` of the `CatalogSources` as well?

njhale: `InstallPlanStatus.CatalogSources` as a `[]registry.SourceKey`?

pkg/controller/registry/resolver/resolver.go
```diff
@@ -167,42 +167,47 @@ func (resolver *SingleSourceResolver) resolveCRDDescription(sources map[registry
...
+		csvSRM, used, err := resolver.resolveCSV(sources, preferredSourceKey, catalogLabelKey, plan.Namespace, csvName)
 
 		if err == nil {
 			srm.Combine(csvSRM)
+			usedSourceKeys = append(used, usedSourceKeys...)
 			continue
 		}
 
 		// Attempt to resolve from any other CatalogSource
 		for srcKey := range sources {
```
alecmerdler: Only resolve from `CatalogSources` in the same namespace of the `InstallPlan` *or* the "global" namespace (`o.namespace`).

njhale: I reduced the catalog sources snapshot that gets passed in to only contain sources from `plan.Namespace` and `o.namespace`.  That should address this issue.

pkg/api/apis/installplan/v1alpha1/types.go
njhale: Done.

pkg/controller/operators/catalog/operator.go
```diff
@@ -437,55 +441,40 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
...
-	// Form the subscription key (owning subcription must be in the same namespace as the install plan)
-	subKey := registry.SubscriptionKey{Name: ownerSub.Name, Namespace: plan.Namespace}
-
-	o.subscriptionsLock.RLock()
-	sub, ok := o.subscriptions[subKey]
-	o.subscriptionsLock.RUnlock()
-
-	var preferredSourceKey registry.SourceKey
-
+func (o *Operator) getPreferredSourceKey(plan *v1alpha1.InstallPlan, includedNamespaces map[string]bool) registry.SourceKey {
```
ecordell: I think this method makes more sense if it takes `sourcesSnapshot` as an arg instead of included namespaces, since you've already filtered the sources by includedNamespaces above

pkg/controller/operators/catalog/operator.go
```diff
@@ -437,55 +441,40 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
...
 
-	if ok {
-		// Attempt to get the subscription's source
-		preferredSourceKey = registry.SourceKey{Name: sub.Spec.CatalogSource, Namespace: sub.Spec.CatalogSourceNamespace}
-		_, ok := o.sources[preferredSourceKey]
-		if ok {
+	var preferredSourceKey registry.SourceKey
+
+	// Attempt to get the plan's source
+	if _, ok := includedNamespaces[plan.Spec.CatalogSource]; ok {
```
ecordell: this part wouldn't be needed if the source list were passed in, but as it is now I think it's supposed to be `plan.Spec.CatalogSourceNamespace`

pkg/controller/registry/resolver/resolver.go
```diff
@@ -215,24 +220,24 @@ func (resolver *MultiSourceResolver) resolveCSV(sources map[registry.SourceKey]r
 		}
 
 		// registry.SourceKey for the source containing the CSV
-		csvSrcKey := firstSrcKey
+		csvSourceKey := preferredSourceKey
```
ecordell: `var csvSourceKey SourceKey` 

(easier to catch bugs where this isn't being set properly if it's not initialized to a valid value)

pkg/controller/registry/resolver/resolver.go
```diff
@@ -302,25 +309,27 @@ func (resolver *MultiSourceResolver) resolveCRDDescription(sources map[registry.
 	}
 
 	// registry.SourceKey for source found to contain the CRD
-	crdSrcKey := firstSrcKey
+	crdSourceKey := preferredSourceKey
 
-	// Attempt to get the first source
-	source, ok := sources[firstSrcKey]
+	// Attempt to get the preferredSource
+	source, ok := sources[preferredSourceKey]
```
ecordell: There's a ton of code scattered throughout for managing the "preferred source". Ideally, these methods should not have to care about which source is preferred. This isn't something we have to fix now, it can be a followup or it can wait until we have better plans for a total-ordering of sources.

njhale: We could always change the ResolveInstallPlan sources parameter to be a pre-ordered, pre-filtered slice of Sources. Another possible alternative is that we establish some criteria for general Source priority and pass in a priority queue.

PR #391, opened 2018-07-20
Make CatalogSource Namespace Required Field on Subscription

alecmerdler:
### Description

It was added to the Go types, but not the CRD manifest itself.

Addresses https://jira.coreos.com/browse/ALM-644

### DIFF
--- a/deploy/chart/templates/07-subscription.crd.yaml
+++ b/deploy/chart/templates/07-subscription.crd.yaml
@@ -29,10 +29,15 @@ spec:
           required:
           - source
           - name
+          - sourceNamespace
           properties:
             source:
               type: string
               description: Name of a CatalogSource that defines where and how to find the channel
+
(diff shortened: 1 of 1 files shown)


njhale:
LGTM

deploy/chart/templates/07-subscription.crd.yaml
```diff
@@ -29,11 +29,16 @@ spec:
           required:
           - source
           - name
+          - sourceNamespace
```
ecordell: This shouldn't be required, right?

This will cause existing Subscriptions to be invalid, and I think we can assume that the sourceNamespace is the catalogNamespace if not specified.

ecordell:
@alecmerdler we _do_ want the `sourceNamespace` field to exist in the validation block, it just shouldn't be required

PR #397, opened 2018-07-26
docs: add access control workflow

robszumski:
I keep fielding questions about how the access control works for admins and end users, so I figured a doc can answer these questions for all.


ecordell:
Thanks @robszumski, this is great!

PR #402, opened 2018-07-31
[WIP] feat(resolver): add namespace and channel awareness

njhale:
### Description

Makes dependency resolution aware of default package channels and preexisting CSVs in an install plan's namespace.

Roughly addresses [ALM-641](https://jira.coreos.com/browse/ALM-641) and [ALM-469](https://jira.coreos.com/browse/ALM-469)

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -253,12 +253,12 @@
   version = "v0.3.6"
 
 [[projects]]
-  digest = "1:0243cffa4a3410f161ee613dfdd903a636d07e838a42d341da95d81f42cd1d41"
+  digest = "1:8e36686e8b139f8fe240c1d5cf3a145bc675c22ff8e707857cdd3ae17b00d728"
   name = "github.com/json-iterator/go"
   packages = ["."]
   pruneopts = "NUT"
-  revision = "ab8a2e0c74be9d3be70b3184d9acc634935ded82"
-  version = "1.1.4"
+  revision = "1624edc4454b8682399def8740d46db5e4362ba4"
+  version = "1.1.5"
 
 [[projects]]
   branch = "master"
@@ -442,7 +442,7 @@
   name = "golang.org/x/crypto"
   packages = ["ssh/terminal"]
   pruneopts = "NUT"
-  revision = "c126467f60eb25f8f27e5a981f32a87e3965053f"
+  revision = "f027049dab0ad238e394a753dba2d14753473a04"
 
 [[projects]]
   branch = "master"
@@ -457,7 +457,7 @@
     "idna",
   ]
   pruneopts = "NUT"
-  revision = "f4c29de78a2a91c00474a2e689954305c350adf9"
+  revision = "f9ce57c11b242f0f1599cf25c89d8cb02c45295a"
 
 [[projects]]
   branch = "master"
@@ -472,14 +472,14 @@
 
 [[projects]]
   branch = "master"
-  digest = "1:0710435fba1b90375bbb61cf59572961c72167775854a2162e0859210a381d06"
+  digest = "1:50f05222003d4bacd6cee17200c7ecbd86415a1e745818abe2f06393d6345c63"
   name = "golang.org/x/sys"
   packages = [
     "unix",
     "windows",
   ]
   pruneopts = "NUT"
-  revision = "3dc4335d56c789b04b0ba99b7a37249d9b614314"
+  revision = "acbc56fc7007d2a01796d5bde54f39e3b3e95945"
 
 [[projects]]
   digest = "1:e33513a825fcd765e97b5de639a2f7547542d1a8245df0cef18e1fd390b778a9"
@@ -592,7 +592,7 @@
     "storage/v1beta1",
   ]
   pruneopts = "NUT"
-  revision = "2d6f90ab1293a1fb871cf149423ebb72aa7423aa"
+  revision = "61b11ee6533278ae17d77fd36825d0b47ec3a293"
 
 [[projects]]
   branch = "release-1.11"
@@ -611,7 +611,7 @@
     "pkg/features",
   ]
   pruneopts = "NUT"
-  revision = "bbc52469f98bf18f81d6010a32c3377726cbfd84"
+  revision = "aaea8ffc60577d42bca6f1ea93a10ec85ec4f54e"
 
 [[projects]]
   branch = "release-1.11"
@@ -665,7 +665,7 @@
     "third_party/forked/golang/reflect",
   ]
   pruneopts = "NT"
-  revision = "103fd098999dc9c0c88536f5c9ad2e5da39373ae"
+  revision = "1c9f4d20e8d7ca63f4ce98dc22110138cd6a365f"
 
 [[projects]]
   branch = "release-1.11"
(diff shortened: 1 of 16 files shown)


pkg/controller/operators/catalog/operator.go
```diff
@@ -41,10 +41,10 @@ type Operator struct {
 	*queueinformer.Operator
 	client             versioned.Interface
 	namespace          string
-	sources            map[registry.SourceKey]registry.Source
```
ecordell: This is a bigger wrench than I intended to throw when I started looking at this PR...

But it strikes me that this type of tracking (for both "sources" and "subscriptions") is exactly what the cache and indexers are supposed to do for you in client-go. 

We already have SharedIndexInformers for all of this available to the operator, so we can probably just use the index provided there.

Take a look at the cache tooling and see if you agree: https://github.com/kubernetes/client-go/tree/master/tools/cache

I would guess that it's as simple as keeping track of the indexInformers by namespace and then querying those.

njhale: I think this makes a lot of sense when dealing with subscriptions since we don't need to do any additional work to pull it from the indexer's cache/store. Sources however, need an InMem instance constructed from the catalog source object and its config map before they are useful.

ecordell: Good point!

A future optimization to consider would be to disallow "internal" catalogsources (replacing them with pods that expose an API over the network), which would let us do the same caching for sources and remove the need to parse and keep all the contents in memory.

njhale: Thanks!

Also, it's questionable at this point whether knowing about subscriptions at resolution time matters as much as knowing what's in the target namespace.

pkg/controller/operators/catalog/operator.go
```diff
@@ -499,3 +505,19 @@ func (o *Operator) getSourcesSnapshot(plan *v1alpha1.InstallPlan, includedNamesp
 
 	return sourcesSnapshot
 }
+
+// getExistingCSVNames returns a set of existing CSV names for the given namespace
+func (o *Operator) getExistingCSVNames(namespace string) (map[string]struct{}, error) {
+	// Get a list of CSV CRs in the namespace
+	csvList, err := o.client.OperatorsV1alpha1().ClusterServiceVersions(namespace).List(metav1.ListOptions{Watch: false})
```
ecordell: pretty sure `Watch` is default `false`

pkg/controller/operators/catalog/operator.go
```diff
@@ -499,3 +505,19 @@ func (o *Operator) getSourcesSnapshot(plan *v1alpha1.InstallPlan, includedNamesp
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Create a set of CSV names from the list
+	existing := make(map[string]struct{})
```
ecordell: There's an edge case with this (and the consumer side in `resolveCRDDescription`).  There can be more than one CSV in a namespace that owns a CRD for a couple of reasons:

- Could be in the middle of upgrading
- Could have attempted to create one and failed (so status is failed but owner is still there)
- Anyone with `csv` `create` permission can write arbitrary owned CRDs (they won't be accepted, but they can write them).

The simplest way to fix this would be to just fail the installplan if we see two csvs that own the CRD already in the namespace (we can't tell which to resolve). Ideally we would just retry the resolution after a bit, and only mark the installplan as truly "Failed" if we've tried a few times and never got a single owner for a CRD (which indicates a bad namespace state - something on the CSV side got stuck).

We could try to detect the upgrading states of the CSVs and watch them directly, but I'd rather keep things more decoupled between catalog/olm.

Anyway this is return a list of existing CSVs in the namespace, but I think what you actually want is a map from a CRDName to the CSVName in the namespace that owns it (then you don't have to loop in `resolveCRDDescription`

pkg/controller/operators/catalog/operator.go
ecordell: Ignore this for this PR, this is something that we should do in general and is orthogonal to this change.

pkg/controller/registry/resolver/resolver_test.go
```diff
@@ -256,6 +262,139 @@ func multiSourceResolveInstallPlan(t *testing.T, resolver DependencyResolver) {
...
+			map[registry.ResourceKey]struct{}{
+				registry.ResourceKey{Name: "macaroni-stable", Kind: csvKind}: struct{}{},
+				registry.ResourceKey{Name: "macaroni", Kind: crdKind}:        struct{}{},
+				registry.ResourceKey{Name: "cheese-alpha", Kind: csvKind}:    struct{}{},
+				registry.ResourceKey{Name: "cheese", Kind: crdKind}:          struct{}{},
+			},
+		},
+	}
+
+	for _, tt := range table {
```
ecordell: want some test cases for when there are multiple CSVs that own the CRD in the namespace

pkg/controller/operators/catalog/operator.go
ecordell: Tracked in ALM-679, ignore this for this PR

Gopkg.lock
```diff
@@ -2,380 +2,487 @@
 
 
 [[projects]]
+  digest = "1:d8ebbd207f3d3266d4423ce4860c9f3794956306ded6c7ba312ecc69cdfbf04c"
```
ecordell: can you re-vendor with `dep` stable?

pkg/controller/registry/resolver/resolver.go
```diff
@@ -126,15 +126,16 @@ func (resolver *MultiSourceResolver) resolveCSV(sourceRefs []registry.SourceRef,
...
-	logger := log.WithFields(log.Fields{"kind": crdDesc.Kind, "name": crdDesc.Name, "version": crdDesc.Version})
+func (resolver *MultiSourceResolver) resolveCRDDescription(sourceRefs []registry.SourceRef, existingCRDOwners map[string][]string, catalogLabelKey, planNamespace string, crdDesc v1alpha1.CRDDescription, owned bool) (v1alpha1.StepResource, string, error) {
+	log.Debugf("resolving %#v", crdDesc)
+
 	crdKey := registry.CRDKey{
 		Kind:    crdDesc.Kind,
 		Name:    crdDesc.Name,
 		Version: crdDesc.Version,
 	}
-	logger.Debug("resolving")
```
ecordell: thanks for cleaning up logs!

pkg/controller/registry/resolver/resolver.go
```diff
@@ -184,9 +183,32 @@ func (resolver *MultiSourceResolver) resolveCRDDescription(sourceRefs []registry
 		return v1alpha1.StepResource{}, "", fmt.Errorf("Unknown CRD %s", crdKey)
 	}
 
-	// TODO: Change to lookup the CSV from the preferred or default channel.
-	logger.WithField("owner", csvs[0].CSV.Name).Info("found owner")
-	return v1alpha1.StepResource{}, csvs[0].CSV.Name, nil
+	var ownerName string
+	if owners, ok := existingCRDOwners[crdKey.Name]; ok && len(owners) > 0 {
```
ecordell: could we make this control flow clearer?

doesn't have to be this, but something that removes the nesting:
```go
owners, ok := existingCRDOwners[crdKey.Name]
if !ok {
  ownerCount = 0 
} else {
  ownerCount = len(owners)
}
switch len(owners):
  case 0:
  // get default channel
  case 1:
  // found an owner
  case >1:
  // error
}
```

PR #405, opened 2018-08-03
Use GroupVersionKind in Spec Descriptor for Selector

alecmerdler:
### Description

Better than using `kind`.


alecmerdler:
Dang it, I forgot to re-generate the catalog sources.

PR #406, opened 2018-08-03
Fix e2e Tests

njhale:
### Description

Fixes e2e tests failing due to various race conditions.


njhale:
Found some more unnecessary control loop checks that compare results

ecordell:
merged in #411

PR #407, opened 2018-08-05
fix(csv): CSV update process optimization regression

ecordell:
I recommend reviewing by commit.

This was an adventure:

- The first commit fixes the actual issue; somewhere in the past we lost the GC optimization that can mark multiple CSVs at once
- When fixing that, I switched to using the typed clients instead of operatorclient's CustomResourceClient
- Because of that, I needed to write a test using the fakes instead of the mocks.
- The test was so much nicer and easier to write than the mocks, that I switched the rest of the tests for the OLM operator to it.
- Because we no longer needed the mocks I was able to delete large chunks of test-only code.
- As part of my refactoring, I broke the re-queuing of CSVs to check for changes. Instead of fixing it short-term, I went ahead and implemented a watch on deployments, which will trigger a sync on any CSVs that own them. This addresses ALM-342
- When verifying my work I fixed some issues in the e2e tests. I'm not 100% sure I removed all flakes, but there may be some overlap with #406 

Before merging I want to play with this on a real cluster, but everything looks good locally.

Also:

- e2e tests now run in 2 minutes 32 seconds, down from 7 minutes 20 seconds
- coverage is up ~2% on olm operator


pkg/controller/operators/olm/operator_test.go
```diff
@@ -2429,3 +2367,69 @@ func TestCSVUpgrades(t *testing.T) {
...
+					csv("csv3", namespace, "", installStrategy("dep"), nil, nil, v1alpha1.CSVPhaseSucceeded),
+				},
+			},
+			expected: nil,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			// configure cluster state
+			clientFake := fake.NewSimpleClientset(tt.initial.csvs...)
```
njhale: Really cool.

PR #411, opened 2018-08-06
fix(e2e) + fix(csv upgrade) + fix(leak)

ecordell:
#406 + #407 + #410

Please see the descriptions on the respective PRs.

In addition to the changes in those PRs:

- Updated the installplan loop to requeue the owning subscription when the installplan changes. this is what fixed the tests for subscriptions (which were timing out and failing).
- Fixes the bug where deploying two copies in CI would cause the older one to lose authorization
- Snuck a little ALM-650 in at the end of this one (adding access to OLM types via aggregated roles)
- Left these pods running overnight to make sure they don't leak. The leaks we saw before happen when the pod loses authorization, but I'm not sure it's worth caring about that (the pod will eventually get killed, which will provide a good signal that it's stopped working because it's not authorized to do its job).


pkg/server/servicebroker/apiserver.go
```diff
@@ -59,7 +59,7 @@ func NewALMBroker(kubeconfigPath string, options Options) (*ALMBroker, error) {
 	if err != nil {
 		return nil, err
 	}
-	almOpClient := operatorclient.NewClient(kubeconfigPath)
+	almOpClient := operatorclient.NewClientFromConfig(kubeconfigPath)
```
njhale: olmOpClient :)

PR #412, opened 2018-08-07
chore(vendor): remove vendor directory

ecordell:
this was only needed because of a private dependency which we have
since removed


ecordell:
(pipeline succeeded, but status update didn't, so I'm going to go ahead and merge)

PR #419, opened 2018-08-14
feat(catalog): add catalog status block updates

njhale:
### Description

Enables the status subresource for catalog sources.

### DIFF
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -137,37 +137,42 @@ func (o *Operator) syncCatalogSources(obj interface{}) (syncError error) {
 		return fmt.Errorf("casting CatalogSource failed")
 	}
 
-	src, err := registry.NewInMemoryFromConfigMap(o.OpClient, catsrc.GetNamespace(), catsrc.Spec.ConfigMap)
+	// Get the catalog source's config map
+	configMap, err := o.OpClient.KubernetesInterface().CoreV1().ConfigMaps(catsrc.GetNamespace()).Get(catsrc.Spec.ConfigMap, metav1.GetOptions{})
 	if err != nil {
-		return fmt.Errorf("failed to create catalog source from ConfigMap %s: %s", catsrc.Spec.ConfigMap, err)
+		return fmt.Errorf("failed to get catalog config map %s when updating status: %s", catsrc.Spec.ConfigMap, err)
 	}
 
-	// Update status subresource
-	out := catsrc.DeepCopy()
+	// Check for changes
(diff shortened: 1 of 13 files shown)


test/e2e/installplan_e2e_test.go
```diff
@@ -129,6 +129,10 @@ func TestCreateInstallPlanManualApproval(t *testing.T) {
 		},
 	}
 
+	// Attempt to get the catalog source before creating install plan
+	_, err = fetchCatalogSource(t, crc, ocsConfigMap, testNamespace, catalogSourceSynced)
```
ecordell: after these changes, the test installplan didn't have a `source` and `sourceNamespace` in the spec anymore

pkg/controller/operators/catalog/operator.go
```diff
@@ -136,16 +136,43 @@ func (o *Operator) syncCatalogSources(obj interface{}) (syncError error) {
...
+		out.Status.LastSync = timeNow()
+
+		_, err = o.client.OperatorsV1alpha1().CatalogSources(out.GetNamespace()).UpdateStatus(out)
+		if err != nil {
+			return fmt.Errorf("failed to update catalog source %s status: %s", out.GetName(), err)
+		}
+	}
+
+	// Create a new in-mem registry
+	src, err := registry.NewInMemoryFromConfigMap(o.OpClient, out.GetNamespace(), out.Spec.ConfigMap)
```
ecordell: it would be nice if we only parsed this into memory when it's changed

test/e2e/util_test.go
```diff
@@ -179,6 +205,62 @@ func waitForDelete(checkResource checkResourceFunc) error {
...
+		builder.WriteString(name)
+		if i < len(names)-1 {
+			builder.WriteString(",")
+		}
+	}
+
+	return builder.String()
+}
+
+func cleanupOLM(t *testing.T, namespace string) {
```
ecordell: the frustrating thing about using this everywhere is that it clears out the state of the test when you `defer` a call to it. it sort of defeats the purpose of keeping the test namespace around when the tests fail (there's nothing to inspect).

test/e2e/installplan_e2e_test.go
njhale: It looks like the `source` and `sourceNamespace` are not in [master](https://github.com/operator-framework/operator-lifecycle-manager/blob/207752168e6c798dd252318d8079d6b38ea5ff44/test/e2e/installplan_e2e_test.go#L125) either.

test/e2e/util_test.go
njhale: Makes total sense.

PR #426, opened 2018-08-17
Action Descriptors

alecmerdler:
### Description

Adds another type of CRD descriptor (`actionDescriptors`) which describe certain actions that can be performed on the custom resource as a whole. Examples could be:
- `urn:alm:descriptor:com.tectonic.ui:action:update` - shorthand for an app-specific update 
- `urn:alm:descriptor:com.tectonic.ui:action:link` - external/internal link to a running service


pkg/api/apis/operators/v1alpha1/clusterserviceversion_types.go
```diff
@@ -41,6 +41,15 @@ type SpecDescriptor struct {
 	Value        *json.RawMessage `json:"value,omitempty"`
 }
 
+// ActionDescriptor describes a declarative action that can be performed on a custom resource instance
+type ActionDescriptor struct {
```
alecmerdler: Uses the same schema as both `SpecDescriptor` and `StatusDescriptor`.

PR #430, opened 2018-08-22
fix(build): Fix install_local.sh calls

aravindhp:
Update the template directory of the install_local.sh calls in the run-local and run-local-shift targets.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,13 +52,13 @@ run-local:
 	. ./scripts/build_local.sh
 	mkdir -p build/resources
 	. ./scripts/package-release.sh 1.0.0-local build/resources Documentation/install/local-values.yaml
-	. ./scripts/install_local.sh local build/resources
+	. ./scripts/install_local.sh local build/resources/templates
 
 run-local-shift:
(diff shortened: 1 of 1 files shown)


rodrigovalin:
Waiting for this to be merged, in the meantime, applying changes locally

ecordell:
I'm pretty sure that this is actually an issue with this line: https://github.com/operator-framework/operator-lifecycle-manager/blob/master/scripts/package-release.sh#L28 not being cross-platform.

These scripts work with macOS's version of `cp`, I think we should just find a cross-platform way to do that instead.

njhale:
Maybe [this](https://twilblog.github.io/macos/linux/cp/copy/2017/02/03/macos-unix-cp-madness.html) will help.
```
cp -R ${charttmpdir}/olm/templates/. ${chartdir}
```

aravindhp:
Ah yes! That works for me. Should I close this PR and let you apply that fix?

ecordell:
@aravindhp I just did tacked it onto another branch I will push up shortly, don't worry about it.

Thanks!

PR #433, opened 2018-08-28
[WIP] Package Extension API Server

njhale:
### Description

Implements a package manifest extension-apiserver that uses the [Kubernetes Aggregation Layer](https://kubernetes.io/docs/tasks/access-kubernetes-api/setup-extension-api-server/) to provide parsed package manifests as top-level Kubernetes resources. Manifests are parsed from config maps pointed at by `sourceType: internal` catalog sources and are stored in memory instead of etcd. 

Result of [ALM-648](https://jira.coreos.com/browse/ALM-648)


Gopkg.toml
```diff
@@ -55,3 +55,14 @@
 [[constraint]]
   name = "github.com/writeas/go-strip-markdown"
   version = "2.0.0"
+
+# resolved to 1.1.3 which doesn't support flags used by apimachinary release-1.11
```
ecordell: these were set a while ago, did you try removing them and resolving to see if it built?

deploy/chart/templates/20-packageserver.yaml
```diff
@@ -0,0 +1,146 @@
+apiVersion: apiregistration.k8s.io/v1beta1
+kind: APIService
+metadata:
+  name: v1alpha1.apps.coreos.com
+spec:
+  insecureSkipTLSVerify: true
+  group: apps.coreos.com
```
ecordell: for actually deploying we probably want to think about this group name

deploy/chart/templates/20-packageserver.yaml
```diff
@@ -0,0 +1,146 @@
+apiVersion: apiregistration.k8s.io/v1beta1
```
ecordell: In order for our packaging scripts to work we need each of these yamls to have its own file (threw them all in one for testing)

Alternately you could fix our packaging scripts to not require that each be in a separate file, whichever

pkg/package-server/apis/packagemanifest/install/install.go
```diff
@@ -0,0 +1,13 @@
...
+
+import (
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/package-server/apis/packagemanifest/v1alpha1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+// Install registers the API group and adds types to a scheme
+func Install(scheme *runtime.Scheme) {
+	v1alpha1.AddToScheme(scheme)
+	//utilruntime.Must(v1alpha1.AddToScheme(scheme))
```
ecordell: do we need these?

pkg/package-server/apis/packagemanifest/v1alpha1/types.go
```diff
@@ -0,0 +1,67 @@
...
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// PackageManifest holds information about a package, which is a reference to one (or more)
+// channels under a single package.
+type PackageManifest struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata"`
+
+	Spec PackageManifestSpec `json:"spec"`
+	//TODO: status
```
ecordell: As a super simple status, we could write what catalog a package came from?

pkg/package-server/apiserver/config.go
```diff
@@ -0,0 +1,98 @@
...
+	genericapiserver.CompletedConfig
+	ProviderConfig *generic.ProviderConfig
+}
+
+// Complete fills in any fields not set that are required to have valid data. It's mutating the receiver.
+func (c *Config) Complete(informers informers.SharedInformerFactory) completedConfig {
+	c.GenericConfig.Version = version.VersionInfo()
+
+	// enable OpenAPI schemas
+	// c.GenericConfig.OpenAPIConfig = genericapiserver.DefaultOpenAPIConfig(generatedopenapi.GetOpenAPIDefinitions, openapinamer.NewDefinitionNamer(generic.Scheme))
```
ecordell: we should enable these I think

pkg/package-server/apiserver/scheme_test.go
```diff
@@ -0,0 +1,14 @@
+package apiserver
+
+//
+//
+//import (
+//"testing"
+//
+//"k8s.io/apimachinery/pkg/api/testing/roundtrip"
+//wardlefuzzer "k8s.io/sample-apiserver/pkg/apis/wardle/fuzzer"
```
ecordell: doubt we need the fuzzing since this is read only, but it might be worth looking at to see if it's useful. otherwise let's remove this.

pkg/package-server/provider/inmem.go
```diff
@@ -0,0 +1,181 @@
...
+
+type packageKey struct {
+	catalogSourceName      string
+	catalogSourceNamespace string
+	packageName            string
+}
+
+type InMemoryProvider struct {
+	*queueinformer.Operator
+	// catalogSourceLister catalogv1alpha1.CatalogSourceLister
```
ecordell: needed?

pkg/package-server/provider/inmem.go
```diff
@@ -0,0 +1,181 @@
...
+		}
+
+		for _, spec := range parsedSpecs {
+			found = true
+			manifest := packagev1alpha1.PackageManifest{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      spec.PackageName,
+					Namespace: cm.GetNamespace(),
+				},
+				Spec: spec,
```
ecordell: we should think about what should actually be here with @alecmerdler 

the current "package spec" from the catalog configmap isn't enough to display packages in the ui. the ui is using the package spec to find the CSV in the configmap and then using that for display.

it could probably be as simple as:

```
spec:
  package:
    package spec
  csv: 
    - csv spec for csv in default channel
```

which would match what we currently do. or it could have it's own set of metadata instead.

pkg/package-server/provider/inmem.go
```diff
@@ -0,0 +1,181 @@
...
+			}
+			// TODO: add check for invalid package definitions
+			manifests = append(manifests, manifest)
+		}
+		log.Debugf("Load ConfigMap      -- Found packages: %v", manifests)
```
ecordell: since we're copying this over anyway, could we make this structured with `log.WithField` etc like we do for other parts of olm?

pkg/package-server/provider/inmem.go
```diff
@@ -0,0 +1,181 @@
...
+	if !ok {
+		log.Debugf("wrong type: %#v", obj)
+		return fmt.Errorf("casting catalog source failed")
+	}
+
+	var manifests []packagev1alpha1.PackageManifest
+
+	// check if the sourcetype is internal
+	switch catsrc.Spec.SourceType {
+	case "internal":
```
ecordell: 👍 for future proofing!

pkg/package-server/storage/packagemanifest/reststorage.go
```diff
@@ -0,0 +1,101 @@
...
+	if err != nil {
+		return nil, err
+	}
+	if pm != nil {
+		manifest = *pm
+	}
+
+	return &manifest, nil
+}
+
```
ecordell: So what happens if I try and create a Package in the API? do I get a reasonable error back?

scripts/build_local.sh
```diff
@@ -5,11 +5,12 @@
 
 set -e
 
-ps x | grep -q [m]inikube || minikube start --kubernetes-version="v1.11.1" --extra-config=apiserver.v=4 || { echo 'Cannot start minikube.'; exit 1; }
+ps x | grep -q [m]inikube || minikube start --extra-config=apiserver.v=4 || { echo 'Cannot start minikube.'; exit 1; }
```
ecordell: is minikube 1.11 by default now?

deploy/chart/templates/20-packageserver.yaml
njhale: I was thinking something like `packages.apps.redhat.com`

pkg/package-server/apiserver/config.go
njhale: Almost done with this.

pkg/package-server/provider/inmem.go
njhale: Yup, sounds good.

deploy/chart/templates/22-packageserver.yaml
```diff
@@ -1,9 +1,10 @@
+{{- $ca := genCA "package-server-ca" 365 -}}
 apiVersion: apiregistration.k8s.io/v1beta1
 kind: APIService
 metadata:
   name: v1alpha1.packages.apps.redhat.com
 spec:
-  insecureSkipTLSVerify: true
+  caBundle: {{ $ca.Cert | b64enc | quote }}
```
jpeeler: You should be able to write this as `caBundle: {{ b64enc $ca.Cert }}`.

pkg/package-server/apis/packagemanifest/v1alpha1/types.go
```diff
@@ -0,0 +1,75 @@
...
+
+	Spec   PackageManifestSpec   `json:"spec,omitempty"`
+	Status PackageManifestStatus `json:"status,omitempty"`
+}
+
+// PackageManifestSpec defines the desired state of PackageManifest
+type PackageManifestSpec struct{}
+
+// PackageManifestStatus represents the current status of the PackageManifest
+type PackageManifestStatus struct {
```
ecordell: We should pull info from the CSV into this. And actually, I think some of it should go into labels if it's not too weird (so that when you do a label query, the object you get back has labels to match)

pkg/package-server/client/clientset/versioned/typed/packagemanifest/v1alpha1/packagemanifest.go
```diff
@@ -0,0 +1,174 @@
...
+// newPackageManifests returns a PackageManifests
+func newPackageManifests(c *PackagemanifestV1alpha1Client, namespace string) *packageManifests {
+	return &packageManifests{
+		client: c.RESTClient(),
+		ns:     namespace,
+	}
+}
+
+// Get takes name of the packageManifest, and returns the corresponding packageManifest object, and an error if there is any.
+func (c *packageManifests) Get(name string, options v1.GetOptions) (result *v1alpha1.PackageManifest, err error) {
```
ecordell: Is there a way to tell the Client codegen to only generate get/list/watch?

pkg/package-server/server/server.go
```diff
@@ -0,0 +1,213 @@
...
+	log "github.com/sirupsen/logrus"
+	genericapiserver "k8s.io/apiserver/pkg/server"
+	genericoptions "k8s.io/apiserver/pkg/server/options"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/clientcmd"
+
+	//"k8s.io/sample-apiserver/pkg/admission/plugin/banflunder"
```
ecordell: these can go

pkg/package-server/server/server.go
```diff
@@ -0,0 +1,213 @@
...
+	server, err := config.Complete(informerFactory).New()
+	if err != nil {
+		return err
+	}
+
+	// run the source provider's informers
+	go sourceProvider.Run(stopCh)
+
+	// add health checks
+	// server.AddHealthzChecks(healthz.NamedCheck("healthz", mgr.CheckHealth))
```
ecordell: health checks would be nice if this helper does what it says

deploy/chart/templates/22-packageserver.yaml
```diff
@@ -0,0 +1,157 @@
+{{- $ca := genCA "package-server-ca" 365 -}}
+apiVersion: apiregistration.k8s.io/v1beta1
+kind: APIService
+metadata:
+  name: v1alpha1.packages.apps.redhat.com
+spec:
+  caBundle: {{ $ca.Cert | b64enc | quote }}
+  group: packages.apps.redhat.com
```
ecordell: This seems weird since the other group is `operators.coreos.com`

pkg/package-server/client/clientset/versioned/typed/packagemanifest/v1alpha1/packagemanifest.go
njhale: Yes - I'm testing that out now.

deploy/chart/templates/22-packageserver.yaml
njhale: I was wondering whether we want to continue using `*.coreos.com`. Also, I added the `package` prefix because the package-server is only going to be responsible for package related resources. If I had just used `apps.redhat.com` or `apps.coreos.com`, the group would be unusable by any other CRDs or extension-apiservers.

PR #434, opened 2018-08-29
[WIP] Conditions on Subscription

alecmerdler:
### Description

Adds `status.conditions` block to the `Subscription` resource.

Implements https://jira.coreos.com/browse/ALM-621

### DIFF
--- a/pkg/api/apis/operators/v1alpha1/subscription_types.go
+++ b/pkg/api/apis/operators/v1alpha1/subscription_types.go
@@ -2,6 +2,7 @@ package v1alpha1
 
 import (
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators"
+	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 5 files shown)


pkg/api/apis/operators/v1alpha1/subscription.go
```diff
@@ -0,0 +1,33 @@
+package v1alpha1
+
+import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+// SetPhase sets the current phase and adds a condition if necessary
+func (c *Subscription) SetSubscriptionPhase(phase SubscriptionPhase, reason ConditionReason, message string) {
```
ecordell: 👍 this matches how we're doing conditions on the other resources.

In the future (for the next apiVersion of these objects) I think we should change these to more align with upstream condition use (i.e. not just a record of phase transitions).

ecordell:
Lets keep this branch around for reference, but we need to do our status and conditions properly for all of our resources; it might be easier to do them all at once.

PR #438, opened 2018-08-29
Remove broker files for repo migration

jpeeler:
ALM-630

This should probably be merged after https://github.com/operator-framework/olm-broker/pull/1 is merged.


ecordell:
We'll want to remove it from the packaging scripts and the `deploy` dir as well

Issue #437, opened 2018-08-29
`./scripts/run_console_local.sh` doesn't provide a usable console with `make run-local` or `make run-local-shift`


djwhatle:
I've been trying to get started with interacting with OLM via the Web UI and have run into issues using `minikube` / `minishift` via `make run-local` and `make run-local-shift` followed by `./scripts/run_console_local.sh` to start the OKD console 

Navigating to `https://my_ip:8443` only shows the service catalog, nothing else, so I figured that the `run_console_local.sh` might help. 

Once I am successful, I'm expecting that there will be a new section of the UI that I can interact with as shown in the screenshot from README.md ( below).

![sub-view](https://user-images.githubusercontent.com/11700385/42060125-c3cde42c-7af3-11e8-87ec-e5910a554902.png)

My actual result is a far less functional UI with an "eternal ellipses" that looks something like this:

![image](https://user-images.githubusercontent.com/7576968/44814814-c359ce00-abab-11e8-9414-789ead77e29d.png)

Menu buttons are clickable but nothing else much happens on the page.

I am also seeing a bunch of errors resulting from the `./scripts/run_console_local.sh` script:
```
[dwhatley@precision-t operator-lifecycle-manager]$ ./scripts/run_console_local.sh 
Using https://192.168.42.152:8443
2018/08/29 20:40:27 cmd/main: cookies are not secure because base-address is not https!
2018/08/29 20:40:27 cmd/main: running with AUTHENTICATION DISABLED!
2018/08/29 20:40:27 cmd/main: Binding to 0.0.0.0:9000...
2018/08/29 20:40:27 cmd/main: not using TLS
2018/08/29 20:40:29 http: proxy error: dial tcp 192.168.42.152:8443: connect: connection refused
2018/08/29 20:40:29 http: proxy error: dial tcp 192.168.42.152:8443: connect: connection refused
[... Lots more of these ...]
2018/08/29 20:41:18 http: proxy error: dial tcp 192.168.42.152:8443: connect: connection refused
2018/08/29 20:41:19 http: proxy error: dial tcp 192.168.42.152:8443: connect: connection refused
2018/08/29 20:41:22 http: proxy error: dial tcp 192.168.42.152:8443: connect: connection refused
2018/08/29 20:41:24 http: proxy error: dial tcp 192.168.42.152:8443: connect: connection refused
2018/08/29 20:41:24 http: proxy error: dial tcp 192.168.42.152:8443: connect: connection refused
[...]

PR #441, opened 2018-08-31
Fix Local Console Script with Minikube

alecmerdler:
### Description

Need to use `--net=host` when running against a locally hosted cluster (like `minikube` or `oc cluster up`).

Fixes https://github.com/operator-framework/operator-lifecycle-manager/issues/437


njhale:
No longer working on OSX :(

njhale:
Hmm, it seems like `--net=host` doesn't behave the same way on OSX. We could probably make use of the `OSTYPE` environment variable to do something like:
```
... 
args="--net=host"
if [[ $OSTYPE == darwin* ]]; then 
  args="-p 9000:9000"
fi

docker run -it $args \
 ...
```

PR #442, opened 2018-08-31
chore(deploy): remove ansible scripts for install

ecordell:
These will just live in openshift-ansible from now on. There's enough
divergence that it's not worth trying to generate them from the same
chart we use for upstream.


njhale:
LGTM

PR #447, opened 2018-09-07
feat(catalog): add aggregated roles for each resolved CRD

ecordell:
For each CRD resolved in an installplan, also generate:

- `edit-<crd>-<crd-version>` which gives edit access to the CRs
- `view-<crd>-<crd-version>` which gives view access to the CRs

Both of them get aggregated to the global `view`, `edit`, and `admin` roles

### DIFF
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -4,12 +4,14 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+
 	"sync"
 	"time"
 
 	olmerrors "github.com/operator-framework/operator-lifecycle-manager/pkg/controller/errors"
 	log "github.com/sirupsen/logrus"
 	"k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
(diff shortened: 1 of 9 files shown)


pkg/controller/operators/catalog/operator.go
```diff
@@ -528,6 +531,26 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
...
+			case clusterRoleKind:
+				// Marshal the manifest into a ClusterRole instance.
+				var cr rbacv1.ClusterRole
+				err := json.Unmarshal([]byte(step.Resource.Manifest), &cr)
+				if err != nil {
+					return err
+				}
+
+				// Attempt to create the ClusterRole.
+				_, err = o.OpClient.KubernetesInterface().RbacV1().ClusterRoles().Create(&cr)
```
jpeeler: Is it possible that the cluster role is ever modified? If so, perhaps reconciliation should be used here instead of create?

ecordell: It could be modified, but the InstallPlans are "run-once". So you just run a new installplan and it will re-create if needed.

PR #452, opened 2018-09-12
add metrics

jpeeler:
Expose counts for various OLM specific resources, which currently are:
CSVs
InstallPlans
Subscriptions
CatalogSources

And also a count for CSV upgrades.

### DIFF
--- a/cmd/olm/main.go
+++ b/cmd/olm/main.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/prometheus/client_golang/prometheus"
 	log "github.com/sirupsen/logrus"
 
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/client"
@@ -15,6 +16,7 @@ import (
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/operators/olm"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/operatorclient"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/signals"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/metrics"
 	olmversion "github.com/operator-framework/operator-lifecycle-manager/pkg/version"
 )
 
@@ -53,6 +55,10 @@ var (
 
 	version = flag.Bool("version", false, "displays olm version")
 )
+
+func init() {
+	metrics.Register()
+}
 
 // main function - entrypoint to ALM operator
 func main() {
@@ -109,6 +115,10 @@ func main() {
 	http.HandleFunc("/healthz", func(w http.ResponseWriter, r *http.Request) {
 		w.WriteHeader(http.StatusOK)
 	})
+	// TODO: both of the following require vendor updates (add k8s.io/apiserver and update prometheus)
+	//healthz.InstallHandler(mux) //(less code)
+	//mux.Handle("/metrics", promhttp.Handler()) //other form is deprecated
+	http.Handle("/metrics", prometheus.Handler())
 	go http.ListenAndServe(":8080", nil)
 
 	operator.Run(stopCh)
--- a/pkg/controller/operators/olm/operator.go
+++ b/pkg/controller/operators/olm/operator.go
@@ -21,6 +21,7 @@ import (
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/operatorclient"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/ownerutil"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/queueinformer"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/metrics"
 )
 
 var ErrRequirementsNotMet = errors.New("requirements were not met")
@@ -73,6 +74,7 @@ func NewOperator(crClient versioned.Interface, opClient operatorclient.ClientInt
 			op.annotateNamespace,
 			nil,
 			"namespace",
+			metrics.NewMetricsNil(),
 		)
 		op.RegisterQueueInformer(queueInformer)
 	}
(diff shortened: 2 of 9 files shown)


pkg/lib/queueinformer/queueinformer_operator.go
```diff
@@ -134,6 +134,7 @@ func (o *Operator) sync(loop *QueueInformer, key string) error {
 		return err
 	}
 
+	// TODO: handle resource deletion metrics here?
```
njhale: I'm wondering if we could [pass in a custom event handler](https://github.com/operator-framework/operator-lifecycle-manager/blob/b870d912cd4f7e3e853a347432fa44cb4874ea78/pkg/lib/queueinformer/queueinformer.go#L88) from the operator that has the resource deletion metrics embedded in the custom `cache.ResourceHandlerFuncs.Delete` function

jpeeler: I thought about that. It's worth considering, but I didn't immediately gravitate towards it since I think I'd have to have that delete function be aware of all the types. Then again, the TODO location would have the same problem.

njhale: Good point. I also just remembered that everything in the `lib` directory is stuff that we wish was in an external package.

jpeeler:
@ecordell Is OLM and the catalog ever deployed separately? I'm wondering if metrics should be exposed from both containers. Also, see if this is closer to what you had in mind (while I debug CI meanwhile).

jpeeler:
If you click the Details link, you can see the pipeline passed (didn't the first time). But I don't know how to update the status on the PR. I'll push a new commit soon and everything should refresh anyway.

pkg/lib/queueinformer/queueinformer_operator.go
```diff
@@ -133,7 +136,9 @@ func (o *Operator) sync(loop *QueueInformer, key string) error {
 	if err != nil {
 		return err
 	}
-
+	if err = o.handleMetrics(loop); err != nil {
```
ecordell: if sync gets retried because syncHandler returns false (see `processNextWorkItem`) this will get called more than we really want. 

Would it make sense to trigger only after successful `syncHandler`? e.g. 

```go
success := loop.syncHandler()
if success {
  go o.handleMetrics(loop) // are prom metrics safe to call concurrently?
}
return success
```

pkg/lib/queueinformer/queueinformer_operator.go
```diff
@@ -142,3 +147,34 @@ func (o *Operator) sync(loop *QueueInformer, key string) error {
 	}
 	return loop.syncHandler(obj)
 }
+
+func (o *Operator) handleMetrics(informer *QueueInformer) error {
+	switch informer.name {
+	case "csv":
```
ecordell: Instead of making the QueueInformer know about all of the informers we might construct it with, what if specialize `handleMetrics` per operator/control loop (e.g. CSV control loop triggers CSV metrics, InstallPlan control loop triggers InstallPlan metrics, etc) similar to how we have the operators specify a `syncHandler`.

Then when you make a new CSV QueueInformer (as a strawman) set the `metricHandler` to 

```go 
o.metricHandler := func() {
	cList, err := o.OpClient.ListCustomResource(v1alpha1.GroupName, v1alpha1.GroupVersion, "", v1alpha1.ClusterServiceVersionKind)
	if err != nil {
		return err
	}
	metrics.CSVCount.Set(float64(len(cList.Items)))
}
```

and `handleMetrics` just calls `o.metricHandler()`

(I think a more common go pattern would be to define a `MetricProvider` interface with a `MetricHandler()` method, but if you want to go that route you should update the syncHandler to follow the same pattern 😄)

cmd/olm/main.go
```diff
@@ -105,11 +115,16 @@ func main() {
...
-	http.HandleFunc("/healthz", func(w http.ResponseWriter, r *http.Request) {
+	mux.HandleFunc("/healthz", func(w http.ResponseWriter, r *http.Request) {
 		w.WriteHeader(http.StatusOK)
 	})
-	go http.ListenAndServe(":8080", nil)
+	// TODO: both of the following require vendor updates (add k8s.io/apiserver and update prometheus)
+	//healthz.InstallHandler(mux) //(less code)
+	//mux.Handle("/metrics", promhttp.Handler()) //other form is deprecated
+	mux.Handle("/metrics", prometheus.Handler())
+	go http.ListenAndServe(":8080", mux)
```
ecordell: do you know if there's a convention for this? is it common to serve metrics on the healthz port or do people define a separate port for each?

pkg/controller/operators/olm/operator.go
```diff
@@ -345,6 +346,7 @@ func (a *Operator) checkReplacementsAndUpdateStatus(csv *v1alpha1.ClusterService
 		log.Infof("newer ClusterServiceVersion replacing %s, no-op", csv.SelfLink)
 		msg := fmt.Sprintf("being replaced by csv: %s", replacement.SelfLink)
 		csv.SetPhase(v1alpha1.CSVPhaseReplacing, v1alpha1.CSVReasonBeingReplaced, msg)
+		metrics.CSVUpgradeCount.Inc()
```
ecordell: I wonder if we even need this if we have the other metrics. `count(changes(csv_count{phase=Replacing} / 2))` might get us there?  

Though with a counter we'll eventually get an accurate count even if we can't scrape some of the events...

cmd/olm/main.go
jpeeler: Yeah, I think it's common to use the same port. Here's what the prometheus operator does (though they aren't using healthz here): https://github.com/coreos/prometheus-operator/blob/ec94db4149766312fb60a2d5fdc5aca3153386dd/cmd/operator/main.go#L201

pkg/controller/operators/olm/operator.go
jpeeler: I left it for now, but I'd be happy to remove since it doesn't really fit in with the theme of the rest of them.

cmd/olm/main.go
ecordell: looks like you can do both without a new muxer? https://github.com/coreos/etcd-operator/blob/master/cmd/operator/main.go#L105-L106

PR #453, opened 2018-09-12
feat(olm): attempt to cleanup namespace annotations on shutdown

ecordell:
This is just a best-effort attempt at cleaning up for now. Longer term strategy is to not rely on namespace annotations.


ecordell:
Passed: https://gitlab.com/operator-framework/operator-framework_operator-lifecycle-manager/pipelines/30021912

:)

PR #457, opened 2018-09-14
[WIP] feat(catalog): add RBAC step resolution

njhale:
### Description

Adds resolution of CSV required ServiceAccounts, Roles, and RoleBindings to InstallPlan resolution. 

Addresses [ALM-588](https://jira.coreos.com/browse/ALM-558)

### DIFF
--- a/pkg/api/apis/operators/v1alpha1/installplan_types.go
+++ b/pkg/api/apis/operators/v1alpha1/installplan_types.go
@@ -254,6 +254,37 @@ func NewStepResourcesFromCRD(crd *v1beta1.CustomResourceDefinition) ([]StepResou
 	return []StepResource{crdStep, aggregatedEditClusterRoleStep, aggregatedViewClusterRoleStep}, nil
 }
 
+// NewStepResourceForObject returns a new StepResource for the provided object
+func NewStepResourceFromObject(obj runtime.Object, name string) (StepResource, error) {
+	var resource StepResource
+
+	// set up object serializer
+	serScheme := runtime.NewScheme()
+	k8sscheme.AddToScheme(serScheme)
+	scheme.AddToScheme(serScheme)
+	serializer := k8sjson.NewSerializer(k8sjson.DefaultMetaFactory, serScheme, serScheme, true)
+
+	// create an object manifest
+	var manifest bytes.Buffer
+	err := serializer.Encode(obj, &manifest)
+	if err != nil {
+		return resource, err
+	}
+
+	gvk := obj.GetObjectKind().GroupVersionKind()
+
+	// create the resource
+	resource = StepResource{
+		Name:     name,
+		Kind:     gvk.Kind,
+		Group:    gvk.Group,
+		Version:  gvk.Version,
+		Manifest: manifest.String(),
+	}
+
+	return resource, nil
+}
+
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 // +genclient
 type InstallPlan struct {
--- a/pkg/api/apis/operators/v1alpha1/installplan_types.go
+++ b/pkg/api/apis/operators/v1alpha1/installplan_types.go
@@ -8,11 +8,9 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
-	"k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/scheme"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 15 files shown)


pkg/controller/operators/catalog/operator.go
```diff
@@ -551,6 +554,84 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
 					plan.Status.Plan[i].Status = v1alpha1.StepStatusCreated
 				}
 
+			case roleKind:
```
alecmerdler: Does this cause problems with [the logic in the `olm-operator`](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/controller/install/deployment.go#L68) that creates RBAC resources?

njhale: Good catch. I was planning on changing the [`olm-operator` logic](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/controller/install/deployment.go#L68) to handle only RBAC verification in a follow up, but it's probably better to go ahead and handle that now.

pkg/api/apis/operators/v1alpha1/installplan_types.go
```diff
@@ -254,6 +254,37 @@ func NewStepResourcesFromCRD(crd *v1beta1.CustomResourceDefinition) ([]StepResou
 	return []StepResource{crdStep, aggregatedEditClusterRoleStep, aggregatedViewClusterRoleStep}, nil
 }
 
+// NewStepResourceForObject returns a new StepResource for the provided object
+func NewStepResourceFromObject(obj runtime.Object, name string) (StepResource, error) {
+	var resource StepResource
+
+	// set up object serializer
+	serScheme := runtime.NewScheme()
```
ecordell: maybe we should just build the scheme on `init` and use it here

pkg/controller/operators/catalog/operator.go
```diff
@@ -551,6 +554,66 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
...
+				// Marshal the manifest into a Role instance.
+				var sa corev1.ServiceAccount
+				err := json.Unmarshal([]byte(step.Resource.Manifest), &sa)
+				if err != nil {
+					return err
+				}
+
+				// Attempt to create the Role.
+				_, err = o.OpClient.KubernetesInterface().CoreV1().ServiceAccounts(plan.Namespace).Create(&sa)
+				if k8serrors.IsAlreadyExists(err) {
```
ecordell: I think we need more here.

Suppose I have CSV1 already in my cluster. I create an installplan for CSV2, and CSV2 replaces CSV1, and that they both use the same serviceAccountName in their permission block (a fairly common scenario with our existing set of CSVs)

Applying the installplan for CSV2 will check if the serviceaccount exists, see that it does, and mark it present. But the ownerrefs on the serviceaccount will still point to CSV1. When CSV2 gets applied to the cluster it will kick off the upgrade and delete CSV1, which will clean up the serviceaccount that CSV2 relied upon.

This might make more sense to solve in the CSV upgrade code (adopt the serviceaccount), but one way to do it would be to have it add or update the ownerrefs in this step.

pkg/controller/registry/resolver/resolver.go
```diff
@@ -218,6 +234,88 @@ func (resolver *MultiSourceResolver) resolveCRDDescription(sourceRefs []registry
...
+
+		// Create Role
+		role := &rbac.Role{
+			TypeMeta: metav1.TypeMeta{
+				Kind: "Role",
+			},
+			Rules: permission.Rules,
+		}
+		ownerutil.AddNonBlockingOwner(role, csv)
+		role.SetName(fmt.Sprintf("%s-role-%d", csv.GetName(), i))
```
ecordell: don't we only need `-%d` if there are two permissions with the same service account? not for every role?

ecordell: Also maybe leave off `role` since that's in the kind?

pkg/controller/registry/resolver/resolver.go
```diff
@@ -218,6 +234,88 @@ func (resolver *MultiSourceResolver) resolveCRDDescription(sourceRefs []registry
...
+				Kind:     "Role",
+				Name:     role.GetName(),
+				APIGroup: rbac.GroupName},
+			Subjects: []rbac.Subject{{
+				Kind: "ServiceAccount",
+				Name: permission.ServiceAccountName,
+			}},
+		}
+		ownerutil.AddNonBlockingOwner(roleBinding, csv)
+		roleBinding.SetName(fmt.Sprintf("%s-%s-rolebinding", role.GetName(), permission.ServiceAccountName))
```
ecordell: maybe leave off `-rolebinding` since it's in the Kind already.

pkg/api/apis/operators/v1alpha1/installplan_types.go
njhale: sounds good

PR #459, opened 2018-09-14
feat(olm): support depending on APIservices

ecordell:
- Adds the ability to depend upon apiservices in a cluster
- Doesn't yet support installing Apiservices


deploy/chart/templates/03-clusterserviceversion.crd.yaml
```diff
@@ -180,6 +180,190 @@ spec:
...
+                            x-descriptors:
+                              type: array
+                              description: A list of descriptors for the status entry that indicate the meaning of the field.
+                              items:
+                                type: string
+                            value:
+                              type: object
+                              description: If present, the value of this status is the same for all instances of the CRD and can be found here instead of on the CR.
+
             customresourcedefinitions:
```
alecmerdler: You just duplicated the `customresourcedefinitions` block, right? The plan is to get rid of it in favor of generic `apiservicedefinitions`?

deploy/chart/templates/03-clusterserviceversion.crd.yaml
```diff
@@ -180,6 +180,190 @@ spec:
...
+                    required:
+                    - name
+                    - version
+                    - kind
+                    - displayName
+                    - description
+                    properties:
+                      name:
+                        type: string
+                        description: Fully qualified name of the CustomResourceDefinition (e.g. my-resource-v1.app.coreos.com)
```
alecmerdler: "Fully qualified name of the APIService"

deploy/chart/templates/03-clusterserviceversion.crd.yaml
ecordell: I duplicated it for v1alpha1, and for v2alpha1 the plan is to do something like that ("providedAPIs", "requiredAPIs"). This keeps things simple for backwards-compatibility.

pkg/api/apis/operators/v1alpha1/clusterserviceversion_types.go
```diff
@@ -282,3 +303,32 @@ func (csv ClusterServiceVersion) GetAllCRDDescriptions() []CRDDescription {
 
 	return descs
 }
+
+// GetAllAPIServiceDescriptions returns a deduplicated set of CRDDescriptions that is
```
njhale: nit: CRDDescriptions -> APIServiceDescriptions

ecordell:
I rewrote a chunk of this because I realized I wasn't checking for the availability of the Kind, so I added checking for that via discovery.

test/e2e/csv_e2e_test.go
```diff
@@ -316,6 +372,69 @@ func TestCreateCSVRequirementsMet(t *testing.T) {
...
+	csv := v1alpha1.ClusterServiceVersion{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       v1alpha1.ClusterServiceVersionKind,
+			APIVersion: v1alpha1.ClusterServiceVersionAPIVersion,
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name: genName("csv"),
+		},
+		Spec: v1alpha1.ClusterServiceVersionSpec{
+			InstallStrategy: v1alpha1.NamedInstallStrategy{
```
njhale: You could also use [this helper](https://github.com/operator-framework/operator-lifecycle-manager/blob/678fea44cfea340f798b8aa629aa8f388a32b66f/test/e2e/installplan_e2e_test.go#L86).

PR #465, opened 2018-09-17
All Namespaces Support for PackageManifest API 

alecmerdler:
### Description

Ensure that all `PackageManifests` are returned when fetching using `--all-namespaces`.

Fixes https://jira.coreos.com/browse/ALM-721

### DIFF
--- a/pkg/package-server/provider/inmem.go
+++ b/pkg/package-server/provider/inmem.go
@@ -227,7 +227,6 @@ func (m *InMemoryProvider) syncCatalogSource(obj interface{}) error {
 	return nil
 }
 
-// ListPackageManifests implements PackageManifestProvider.ListPackageManifests()
 func (m *InMemoryProvider) ListPackageManifests(namespace string) (*packagev1alpha1.PackageManifestList, error) {
 	manifestList := &packagev1alpha1.PackageManifestList{}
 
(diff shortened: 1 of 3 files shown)


pkg/package-server/provider/inmem.go
```diff
@@ -247,13 +246,13 @@ func (m *InMemoryProvider) ListPackageManifests(namespace string) (*packagev1alp
...
 }
 
-// GetPackageManifest implements PackageManifestProvider.GetPackageManifest(...)
 func (m *InMemoryProvider) GetPackageManifest(namespace, name string) (*packagev1alpha1.PackageManifest, error) {
 	m.mu.RLock()
 	defer m.mu.RUnlock()
 
 	var manifest packagev1alpha1.PackageManifest
 	for key, pm := range m.manifests {
+		// FIXME(alecmerdler): Needs to work with "all-namespaces"
```
njhale: In the "all-namespaces" case would you return the first `PackageManifest` matching `name`?

alecmerdler: Ah nevermind, the Kubernetes API does not support getting a single item from `--all-namespaces`.
```shell
$ kubectl get pods --all-namespaces mongodb-2-6k7rg
error: a resource cannot be retrieved by name across all namespaces
```

PR #470, opened 2018-09-19
Issue #469 update the service catalog instruction for minishift.

### LINKED ISSUE #468: The `--service-catalog` clusterup option is invalid for OS 3.10+
This causes the `scripts/build-local-shift.sh` to fail. I'm not sure of the best way to determine which flag value to pass, but for 3.10 the syntax should be: 

  `--extra-clusterup-flags "--enable=*,service-catalog"`

### LINKED ISSUE #469: minishift steps need to be updated for service catalog
https://github.com/operator-framework/operator-lifecycle-manager/blob/master/scripts/build_local_shift.sh#L7 is not going to work if you are using the minishift version >= 1.22.0 because from openshift 3.10 cluster up doesn't have `--service-catalog` as flag so this should be updated to `MINISHIFT_ENABLE_EXPERIMENTAL=y minishift start --extra-clusterup-flags "--enable=*,service-catalog'"`

praveenkumar:
Fix #469 
Fix #468

### DIFF
--- a/scripts/build_local_shift.sh
+++ b/scripts/build_local_shift.sh
@@ -4,7 +4,7 @@
 # This is used to start and build services for running e2e tests
 
 set -e
-MINISHIFT_ENABLE_EXPERIMENTAL=y minishift start --extra-clusterup-flags "--service-catalog" \
+minishift start \
     || { echo 'Cannot start shift.'; exit 1; }
 
 eval $(minishift docker-env) \


scripts/build_local_shift.sh
```diff
@@ -4,7 +4,7 @@
 # This is used to start and build services for running e2e tests
 
 set -e
-MINISHIFT_ENABLE_EXPERIMENTAL=y minishift start --extra-clusterup-flags "--service-catalog" \
+MINISHIFT_ENABLE_EXPERIMENTAL=y minishift start --extra-clusterup-flags "--enable='*,service-catalog'" \
```
njhale: We should probably remove the `service-catalog` flag since we no longer need to test its integration in this repo (see [issue 468](https://github.com/operator-framework/operator-lifecycle-manager/issues/468#issuecomment-422563076)).

praveenkumar:
@ecordell ack, will update it.

PR #476, opened 2018-09-19
Implement Watch for `PackageManifest` API

alecmerdler:
### Description

Allows watching the `PackageManifest` resource.

### Future steps

- Return `resourceVersion` in `PackageManifestList` and implement watching based on it
- Implement `Modified` and `Deleted` events when packages are changed or removed

Resolves https://jira.coreos.com/browse/ALM-722

### DIFF
--- a/pkg/package-server/provider/inmem.go
+++ b/pkg/package-server/provider/inmem.go
@@ -39,6 +39,8 @@ type InMemoryProvider struct {
 
 	mu        sync.RWMutex
 	manifests map[packageKey]packagev1alpha1.PackageManifest
+
+	out chan packagev1alpha1.PackageManifest
 }
 
 // NewInMemoryProvider returns a pointer to a new InMemoryProvider instance
@@ -47,6 +49,7 @@ func NewInMemoryProvider(informers []cache.SharedIndexInformer, queueOperator *q
 	prov := &InMemoryProvider{
 		Operator:  queueOperator,
 		manifests: make(map[packageKey]packagev1alpha1.PackageManifest),
+		out:       make(chan packagev1alpha1.PackageManifest),
 	}
 
 	// register CatalogSource informers.
@@ -222,11 +225,28 @@ func (m *InMemoryProvider) syncCatalogSource(obj interface{}) error {
 
 		log.Debugf("storing packagemanifest at %+v", key)
 		m.manifests[key] = manifest
+		m.out <- manifest
 	}
 
 	return nil
 }
 
+// GetPackageManifest implements PackageManifestProvider.GetPackageManifest(...)
+func (m *InMemoryProvider) GetPackageManifest(namespace, name string) (*packagev1alpha1.PackageManifest, error) {
+	m.mu.RLock()
+	defer m.mu.RUnlock()
+
+	var manifest packagev1alpha1.PackageManifest
+	for key, pm := range m.manifests {
+		if key.packageName == name && key.catalogSourceNamespace == namespace {
+			manifest = pm
+		}
+	}
+
+	return &manifest, nil
+}
+
+// ListPackageManifests implements PackageManifestProvider.ListPackageManifests()
 func (m *InMemoryProvider) ListPackageManifests(namespace string) (*packagev1alpha1.PackageManifestList, error) {
 	manifestList := &packagev1alpha1.PackageManifestList{}
 
@@ -248,16 +268,27 @@ func (m *InMemoryProvider) ListPackageManifests(namespace string) (*packagev1alp
 	return manifestList, nil
 }
 
-func (m *InMemoryProvider) GetPackageManifest(namespace, name string) (*packagev1alpha1.PackageManifest, error) {
-	m.mu.RLock()
-	defer m.mu.RUnlock()
-
-	var manifest packagev1alpha1.PackageManifest
-	for key, pm := range m.manifests {
-		if key.packageName == name && key.catalogSourceNamespace == namespace {
-			manifest = pm
+// WatchPackageManifests forwards all PackageManifests matching the given namespace to the given out chan
+func (m *InMemoryProvider) WatchPackageManifests(namespace string, out chan packagev1alpha1.PackageManifest, stop <-chan struct{}) {
+	go func() {
+		// push existing PackageManifests
+		manifestList, err := m.ListPackageManifests(namespace)
+		if err == nil {
+			for _, manifest := range manifestList.Items {
+				out <- manifest
+			}
 		}
-	}
 
-	return &manifest, nil
+		// watch for changes
+		for {
+			select {
+			case manifest := <-m.out:
+				if manifest.GetNamespace() == namespace {
+					out <- manifest
+				}
+			case <-stop:
+				return
+			}
+		}
+	}()
 }
--- a/pkg/package-server/provider/inmem.go
+++ b/pkg/package-server/provider/inmem.go
@@ -22,7 +22,6 @@ import (
 const (
 	// ConfigMapPackageName is the key for package ConfigMap data
 	ConfigMapPackageName = "packages"
-
 	// ConfigMapCSVName is the key for CSV ConfigMap data
 	ConfigMapCSVName = "clusterServiceVersions"
 )
@@ -33,26 +32,26 @@ type packageKey struct {
 	packageName            string
 }
 
-// InMemoryProvider syncs and provides PackageManifests from the cluster in an in-memory cache
+// InMemoryProvider syncs and provides PackageManifests from the cluster using an in-memory cache.
+// Should be a global singleton.
 type InMemoryProvider struct {
 	*queueinformer.Operator
+	mu sync.RWMutex
 
-	mu        sync.RWMutex
 	manifests map[packageKey]packagev1alpha1.PackageManifest
 
-	out chan packagev1alpha1.PackageManifest
+	add    []chan packagev1alpha1.PackageManifest
+	modify []chan packagev1alpha1.PackageManifest
+	delete []chan packagev1alpha1.PackageManifest
 }
 
 // NewInMemoryProvider returns a pointer to a new InMemoryProvider instance
 func NewInMemoryProvider(informers []cache.SharedIndexInformer, queueOperator *queueinformer.Operator) *InMemoryProvider {
-	// instantiate the in-mem provider
 	prov := &InMemoryProvider{
 		Operator:  queueOperator,
 		manifests: make(map[packageKey]packagev1alpha1.PackageManifest),
-		out:       make(chan packagev1alpha1.PackageManifest),
 	}
 
-	// register CatalogSource informers.
 	queue := workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "catalogsources")
 	queueInformers := queueinformer.New(
 		queue,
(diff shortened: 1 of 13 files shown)


pkg/package-server/provider/inmem.go
```diff
@@ -222,11 +225,26 @@ func (m *InMemoryProvider) syncCatalogSource(obj interface{}) error {
 
 		log.Debugf("storing packagemanifest at %+v", key)
 		m.manifests[key] = manifest
+		m.out <- manifest
```
ecordell: Won't this cause every single package to stream to the client during every resync of CatalogSources?

pkg/package-server/provider/inmem.go
```diff
@@ -248,16 +266,28 @@ func (m *InMemoryProvider) ListPackageManifests(namespace string) (*packagev1alp
...
 
-func (m *InMemoryProvider) GetPackageManifest(namespace, name string) (*packagev1alpha1.PackageManifest, error) {
-	m.mu.RLock()
-	defer m.mu.RUnlock()
-
-	var manifest packagev1alpha1.PackageManifest
-	for key, pm := range m.manifests {
-		if key.packageName == name && key.catalogSourceNamespace == namespace {
-			manifest = pm
+func (m *InMemoryProvider) WatchPackageManifests(namespace string, out chan packagev1alpha1.PackageManifest, stop <-chan struct{}) {
```
ecordell: I don't think this is being called anywhere, it looks like the `PackageManifestWatcher` is what's actually being used by the server

pkg/package-server/storage/packagemanifest/watcher.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+
+	return watcher
+}
+
+func (w *PackageManifestWatcher) Run() {
+	go func() {
+		for {
+			select {
+			case manifest := <-w.in:
```
ecordell: it would be nice if we could emit created/deleted here too, but I see that may be difficult. Other controllers can get that info from etcd events, but we don't have those here.

pkg/package-server/storage/packagemanifest/reststorage.go
```diff
@@ -96,6 +98,22 @@ func (m *PackageManifestStorage) Get(ctx context.Context, name string, opts *met
...
+	namespace := genericapirequest.NamespaceValue(ctx)
+
+	// get selector
+	labelSelector := labels.Everything()
+	if options != nil && options.LabelSelector != nil {
+		labelSelector = options.LabelSelector
+	}
+
+	watcher := NewPackageManifestWatcher(namespace, labelSelector)
+	watcher.Run()
```
njhale: It looks like before we call `watcher.Run()` we need to wire its channels to the provider using `provider.WatchPackageManifests(namespace string, out chan v1alpha1.PackageManifest, stop <-chan struct{})`.

alecmerdler:
@ecordell @njhale @jpeeler Ready for another review.

pkg/package-server/provider/interfaces.go
```diff
@@ -5,6 +5,7 @@ import (
 )
 
 type PackageManifestProvider interface {
-	ListPackageManifests(namespace string) (*v1alpha1.PackageManifestList, error)
-	GetPackageManifest(namespace, name string) (*v1alpha1.PackageManifest, error)
+	Get(namespace, name string) (*v1alpha1.PackageManifest, error)
+	List(namespace string) (*v1alpha1.PackageManifestList, error)
+	Subscribe() (add, modify, delete <-chan v1alpha1.PackageManifest)
```
njhale: 👍

pkg/package-server/server/server.go
```diff
@@ -89,7 +89,8 @@ func NewPackageServerOptions(out, errOut io.Writer) *PackageServerOptions {
 		WatchedNamespaces: []string{v1.NamespaceAll},
 		WakeupInterval:    5 * time.Minute,
 
-		Debug: false,
+		DisableAuthForTesting: true,
```
njhale: nit: I think we should default this to `false` if it's being fed to the `genericapiserver`

pkg/package-server/server/server.go
```diff
@@ -154,52 +154,59 @@ func (o *PackageServerOptions) Run(stopCh <-chan struct{}) error {
...
 	queueOperator, err := queueinformer.NewOperator(o.Kubeconfig)
 	if err != nil {
 		return err
 	}
 
 	sourceProvider := provider.NewInMemoryProvider(catsrcSharedIndexInformers, queueOperator)
 
-	// inject the providers into the config
-	config.ProviderConfig.Provider = sourceProvider
+	go func() {
```
njhale: Nice! My only reservation is that without this goroutine, the provider will be blocked until someone calls watch. I almost feel like this belongs in a `Provider.Run(stopCh <-chan struct{})`, just to keep it coupled with the provider.

njhale: Another thought; we could also have the provider store three separate channel slices, `add, modify, delete []chan *PackageManifest`. Each new call to subscribe could make and append the appropriate channels to these slices. When the provider detects an event, it should be emitted to all channels in the appropriate slice. 

Now that I'm thinking about this more deeply. It looks like having a single channel for each event type will cause issues of its own. Channels deliver _once_, so if mutiple goroutines are reading from the channels returned from subscribe, only one of them will get a sent event.

pkg/package-server/storage/packagemanifest/watcher.go
```diff
@@ -0,0 +1,82 @@
...
+
+	source provider.PackageManifestProvider
+
+	stop   chan struct{}
+	result chan watch.Event
+}
+
+var _ watch.Interface = &Watcher{}
+
+func NewWatcher(namespace, name, resourceVersion string, labelSelector labels.Selector, source provider.PackageManifestProvider) *Watcher {
```
ecordell: I think you want to take a `context.Context` here

pkg/package-server/storage/packagemanifest/watcher.go
```diff
@@ -0,0 +1,82 @@
...
+		source:          source,
+		stop:            make(chan struct{}),
+		result:          make(chan watch.Event),
+	}
+}
+
+// Run is a blocking method which starts the watch by subscribing to the source's channels.
+// Should run in a goroutine.
+func (w *Watcher) Run() {
+	add, modify, delete := w.source.Subscribe()
```
ecordell: and I think you want to listen to `ctx.Done` here

pkg/package-server/storage/packagemanifest/watcher.go
```diff
@@ -0,0 +1,82 @@
...
+			w.Modify(manifest)
+		case manifest := <-delete:
+			w.Delete(manifest)
+		case <-w.stop:
+			return
+		}
+	}
+}
+
+func (w *Watcher) Stop() {
```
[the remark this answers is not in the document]
alecmerdler: I'm pretty sure it's internal k8s code, because it's [part of the interface of `watch.Interface`](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/vendor/k8s.io/apimachinery/pkg/watch/watch.go#L290).

pkg/package-server/provider/inmem.go
```diff
@@ -248,16 +265,13 @@ func (m *InMemoryProvider) ListPackageManifests(namespace string) (*packagev1alp
...
 }
 
-func (m *InMemoryProvider) GetPackageManifest(namespace, name string) (*packagev1alpha1.PackageManifest, error) {
-	m.mu.RLock()
-	defer m.mu.RUnlock()
+func (m *InMemoryProvider) Subscribe() (<-chan packagev1alpha1.PackageManifest, <-chan packagev1alpha1.PackageManifest, <-chan packagev1alpha1.PackageManifest, error) {
+	add := make(chan packagev1alpha1.PackageManifest)
+	modify := make(chan packagev1alpha1.PackageManifest)
+	delete := make(chan packagev1alpha1.PackageManifest)
+	m.add = append(m.add, add)
```
njhale: We should probably use a lock to synchronize appending to these slices.

njhale: I think we need to synchronize appending to these slices.

pkg/package-server/storage/packagemanifest/watcher.go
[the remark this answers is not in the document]
njhale: > Do you know what calls `Stop()` on the watcher?
> 
> Also, probably want to guard against calling multiple times (double-close will panic):
> 
> ```go
> w.Lock()
> defer w.Unlock()
> if !w.Stopped {
>   close(w.result)
>   w.Stopped = true
> }
> ```

alternatively:

`w.once.Do(func() { close(w.result) })` where `w.once` is of type `sync.Once`

pkg/package-server/storage/packagemanifest/watcher.go
```diff
@@ -0,0 +1,86 @@
...
+		labelSelector:   labelSelector,
+		source:          source,
+		stop:            make(chan struct{}),
+		result:          make(chan watch.Event),
+	}
+}
+
+// Run is a blocking method which starts the watch by subscribing to the source's channels.
+// Should run in a goroutine.
+func (w *Watcher) Run() {
```
ecordell: I think you want to take a context.Context here

pkg/package-server/storage/packagemanifest/reststorage.go
```diff
@@ -96,7 +103,49 @@ func (m *PackageManifestStorage) Get(ctx context.Context, name string, opts *met
...
+		return nil, err
+	}
+
+	labelSelector := labels.Everything()
+	if options != nil && options.LabelSelector != nil {
+		labelSelector = options.LabelSelector
+	}
+
+	watcher := NewWatcher(namespace, name, options.ResourceVersion, labelSelector, m.prov)
+	go watcher.Run()
```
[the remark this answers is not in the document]
ecordell: notice that the etcdwatcher gets tied back to the request lifecycle: https://github.com/kubernetes/sample-apiserver/blob/df44042/vendor/k8s.io/apiserver/pkg/storage/etcd3/watcher.go#L246

pkg/package-server/storage/packagemanifest/watcher.go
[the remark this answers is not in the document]
ecordell: @alecmerdler You're right, and the code that calls it is here: https://github.com/kubernetes/sample-apiserver/blob/df44042/vendor/k8s.io/apiserver/pkg/endpoints/handlers/watch.go

As long as we're going through this codepath for the aggregated apiserver we're fine. Also note that it servers http or websockets :)

pkg/package-server/storage/packagemanifest/watcher.go
alecmerdler: @ecordell Passing and handling `context.Context` now.

pkg/package-server/provider/inmem.go
njhale: [append is not a thread-safe operation](https://stackoverflow.com/questions/44152988/append-not-thread-safe)

alecmerdler: Yeah, I just discovered this when [running `go test -race` using my `FakeProvider`](https://github.com/operator-framework/operator-lifecycle-manager/pull/476/files#diff-8bcb7f56c2a87a0aecaed913c951245fR44).

pkg/package-server/storage/packagemanifest/watcher.go
```diff
@@ -0,0 +1,102 @@
...
+	"sync"
+
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/watch"
+
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/package-server/apis/packagemanifest/v1alpha1"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/package-server/provider"
+)
+
+type Watcher struct {
```
alecmerdler: Based off of the [example in k8s apimachinery](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/vendor/k8s.io/apimachinery/pkg/watch/watch.go#L159).

PR #477, opened 2018-09-20
Allow resolving ClusterRoles

ecordell:
- Adds `clusterPermissions` to CSV install strategy (just for v1alpha1, we'll revisit for next api version)
- Resolves ClusterRoles and ClusterRoleBindings

This greatly expands the power of operators that can be installed with OLM. Restricting what operators can do can be achieved be reducing the scope of permissions given to OLM.

Fixes #477


pkg/controller/registry/resolver/resolver.go
```diff
@@ -235,7 +234,7 @@ func resolveCRDDescription(sourceRefs []registry.SourceRef, existingCRDOwners ma
 func resolveRBACStepResources(csv *v1alpha1.ClusterServiceVersion) ([]v1alpha1.StepResource, error) {
 	var rbacSteps []v1alpha1.StepResource
 
-	// User a StrategyResolver to
+	// User a StrategyResolver to get the strategy details
```
njhale: 👍

PR #478, opened 2018-09-21
certified-operators: rectify form of business for Dynatrace

baichinger:


jpeeler:
Author works for the company that is having the name tweaked, so I'm sure they know best.
/lgtm

PR #479, opened 2018-09-24
feat(olm): add RBAC requirements check for CSVs

njhale:
### Description
Adds RBAC requirements check in the olm-operator's CSV control loop.

Addresses [ALM-706](https://jira.coreos.com/browse/ALM-706)


pkg/controller/install/rule_checker.go
```diff
@@ -0,0 +1,140 @@
...
+}
+
+// RuleSatisfied returns true if a ServiceAccount is authorized to perform all actions described by a PolicyRule in a namespace
+func (c *CSVRuleChecker) RuleSatisfied(sa *corev1.ServiceAccount, namespace string, rule rbacv1.PolicyRule) (bool, error) {
+	// get attributes set for the given Role and ServiceAccount
+	user := toDefaultInfo(sa)
+	attributesSet := toAttributesSet(user, namespace, rule)
+
+	// create a new RBACAuthorizer
+	rbacAuthorizer := rbacauthorizer.New(c, c, c, c)
```
ecordell: I think this is pretty important: https://github.com/kubernetes/kubernetes/blob/473ebb21d168b74c1d6d729faf513ad311d30e2b/pkg/kubeapiserver/authorizer/config.go#L108-L113

I think we want to wire up to informers like this, so that we pull from caches instead of querying the api many times

njhale: Agreed. I'm thinking that I'll need to write wrappers around each of the listers to filter on ownerref too.

Gopkg.toml
```diff
@@ -70,3 +70,7 @@ required = ["k8s.io/kube-openapi/cmd/openapi-gen"]
 [[constraint]]
   branch = "release-1.11"
   name = "k8s.io/kube-aggregator"
+
+[[constraint]]
```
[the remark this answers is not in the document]
njhale: I switched `k8s.io/kubernetes` to `branch=release-1.11`

njhale:
@ecordell @alecmerdler @jpeeler This should be ready for another review. The `CSVRuleChecker` now uses shared index listers for RBAC resource listing. As a result, I ended up implementing RBAC resource syncing and CSV owner requeueing too.

pkg/controller/install/rule_checker.go
```diff
@@ -0,0 +1,156 @@
...
+	for _, crb := range crbList {
+		if !c.hasOwnerConflicts(crb.GetOwnerReferences()) {
+			filtered = append(filtered, crb)
+		}
+	}
+
+	return filtered, nil
+}
+
+// hasOwnerConflicts checks if the given list of OwnerReferences points to CSVs other than the
```
njhale: I should probably reword this. Something got messed up over a few revisions.

PR #480, opened 2018-09-24
add OperatorGroup

jpeeler:
So far the reconcile loop only looks at the namespace selector and writes the results to the status and an annotation on each deployment.


deploy/chart/templates/14-operatorgroup.crd.yaml
```diff
@@ -0,0 +1,51 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  creationTimestamp: null
+  labels:
+    controller-tools.k8s.io: "1.0"
+  name: operatorgroups.operators.coreos.com.operators.coreos.com
```
njhale: Is this supposed to be "operatorgroups.operators.coreos.com"?

jpeeler: Oh, yep. I think that's a kubebuilder bug as I don't think I messed that part up...

deploy/chart/templates/30_15-operatorgroup.crd.yaml
```diff
@@ -0,0 +1,51 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  creationTimestamp: null
+  labels:
+    controller-tools.k8s.io: "1.0"
+  name: operatorgroups.operators.coreos.com
+spec:
+  group: operators.coreos.com
+  names:
```
njhale: I believe you can get rid of `15-operatorgrouplist.crd.yaml` if you add `listKind: OperatorGroupList` here. Check out [how CSVs are handled in master](https://github.com/operator-framework/operator-lifecycle-manager/blob/05a838747633bb298e27ecf6209e25774ae0e034/deploy/chart/templates/30_02-clusterserviceversion.crd.yaml#L10).

pkg/controller/operators/olm/operator.go
```diff
@@ -158,6 +182,54 @@ func (a *Operator) syncDeployment(obj interface{}) (syncError error) {
...
+	if err != nil {
+		return err
+	}
+	op.Status.Namespaces = namespaceList.Items
+
+	// make string list made up of comma separated namespaces
+	var b strings.Builder
+	nsCount := len(namespaceList.Items)
+	for i := 0; i < nsCount-1; i++ {
+		fmt.Fprintf(&b, "%v,", namespaceList.Items[i])
```
jpeeler: This might require .Name here, but maybe there's a better way to do all this in general?

pkg/controller/operators/olm/operator.go
```diff
@@ -158,6 +182,54 @@ func (a *Operator) syncDeployment(obj interface{}) (syncError error) {
...
+		return fmt.Errorf("casting OperatorGroup failed")
+	}
+
+	// write namespace matches to status field
+	selector, err := metav1.LabelSelectorAsSelector(&op.Spec.Selector)
+	if err != nil {
+		return err
+	}
+	operatorGroupOpts := metav1.ListOptions{LabelSelector: selector.String()}
+	namespaceList, err := a.OpClient.KubernetesInterface().CoreV1().Namespaces().List(operatorGroupOpts)
```
njhale: I think we want to impersonate `op.Spec.ServiceAccount` when listing namespaces.

pkg/controller/operators/olm/operator.go
njhale: If you don't need to use .Name (which I think we may need to), you can use `strings.Join(namespaceList.Items, ",")` here as an alternative.

pkg/controller/operators/olm/operator.go
```diff
@@ -158,6 +182,54 @@ func (a *Operator) syncDeployment(obj interface{}) (syncError error) {
...
+			return fmt.Errorf("error unmarshaling strategy from ClusterServiceVersion '%s' with error: %s", csvName, err)
+		}
+
+		strategyDetailsDeployment, ok := strategy.(*install.StrategyDetailsDeployment)
+		if !ok {
+			return fmt.Errorf("could not assert strategy implementation as deployment for CSV %s", csvName)
+		}
+
+		for _, deploy := range strategyDetailsDeployment.DeploymentSpecs {
+			deploy.Spec.Template.Annotations["olm.targetNamespaces"] = b.String()
```
njhale: Awesome, is this to have the CSV control loop update those CSVs'  deployments? If that's the case then we'll need to update those CSV's in Kube after we edit them. Also, I'm thinking that the deployment installer will need to check for updates to the annotations field; [possibly here](https://github.com/operator-framework/operator-lifecycle-manager/blob/05a838747633bb298e27ecf6209e25774ae0e034/pkg/controller/operators/olm/operator.go#L361:20) or [here](https://github.com/operator-framework/operator-lifecycle-manager/blob/05a838747633bb298e27ecf6209e25774ae0e034/pkg/controller/install/deployment.go#L201:39).

pkg/api/apis/operators/v1alpha2/operatorgroup_types.go
```diff
@@ -0,0 +1,33 @@
+package v1alpha2
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type OperatorGroupSpec struct {
+	Selector       metav1.LabelSelector  `json:"selector"`
+	ServiceAccount corev1.ServiceAccount `json:"serviceAccount"`
```
ecordell: I think we want both of these to be `omittempty`

pkg/api/apis/operators/v1alpha2/operatorgroup_types.go
```diff
@@ -0,0 +1,33 @@
...
+}
+
+type OperatorGroupStatus struct {
+	Namespaces []corev1.Namespace `json:"namespaces"`
```
ecordell: We probably want a "last updated time" here

pkg/api/client/informers/externalversions/generic.go
```diff
@@ -62,6 +63,10 @@ func (f *sharedInformerFactory) ForResource(resource schema.GroupVersionResource
 	case v1alpha1.SchemeGroupVersion.WithResource("subscriptions"):
 		return &genericInformer{resource: resource.GroupResource(), informer: f.Operators().V1alpha1().Subscriptions().Informer()}, nil
 
+		// Group=operators.coreos.com, Version=v1alpha2
```
ecordell: did you add this manually? it should have been generated

pkg/controller/operators/olm/operator.go
ecordell: We will in the near future, but for this PR I think we should stay away from the impersonation. We can get a lot out of the operatorgroup without implementing that just yet.

pkg/controller/operators/olm/operator.go
ecordell: I think it makes sense for `namespaces` in the status to just be a list of strings. Normally when we do object references we try and include GVK but namespaces are pretty ingrained (e.g. `metadata.namespace` is just a string on every kube object).

If we make that change, then this part just becomes what @njhale suggested

pkg/controller/operators/olm/operator.go
```diff
@@ -158,6 +182,54 @@ func (a *Operator) syncDeployment(obj interface{}) (syncError error) {
...
+	if err != nil {
+		return err
+	}
+	operatorGroupOpts := metav1.ListOptions{LabelSelector: selector.String()}
+	namespaceList, err := a.OpClient.KubernetesInterface().CoreV1().Namespaces().List(operatorGroupOpts)
+	if err != nil {
+		return err
+	}
+	op.Status.Namespaces = namespaceList.Items
+
```
ecordell: Once we get here, we want to check and see if the list of namespaces has changed since the last time we ran the selector. If they're the same, exit early here and don't update all of the deployments (no need to, as their annotations wouldn't have changed).

pkg/api/client/informers/externalversions/generic.go
jpeeler: No, the v1alpha1 generation produced a comment like this too.

pkg/api/apis/operators/v1alpha2/operatorgroup_types.go
ecordell: Yep, and in absence of a serviceaccount we'd use the one that OLM is running as (system:controller:operator-lifecycle-manager)

jpeeler:
The failing images test will pass once https://github.com/operator-framework/operator-lifecycle-manager/pull/489 is merged.

pkg/controller/operators/olm/operator.go
```diff
@@ -158,6 +187,80 @@ func (a *Operator) syncDeployment(obj interface{}) (syncError error) {
...
+	// TODO(jpeeler): this needs to use user impersonation with op.Spec.ServiceAccount
+	namespaceList, err := a.OpClient.KubernetesInterface().CoreV1().Namespaces().List(operatorGroupOpts)
+	if err != nil {
+		return err
+	}
+
+	nsCount := len(namespaceList.Items)
+
+	if len(op.Status.Namespaces) == nsCount {
+		for i, v := range namespaceList.Items {
```
ecordell: I think it would be better to implement this as a set (`map[string]struct{}{}`) instead of a list; I'm not sure if we get order guarantees from `List`

pkg/controller/operators/olm/operator.go
```diff
@@ -158,6 +187,80 @@ func (a *Operator) syncDeployment(obj interface{}) (syncError error) {
...
+		}
+
+		strategyDetailsDeployment, ok := strategy.(*install.StrategyDetailsDeployment)
+		if !ok {
+			return fmt.Errorf("could not assert strategy implementation as deployment for CSV %s", csvName)
+		}
+
+		for _, deploy := range strategyDetailsDeployment.DeploymentSpecs {
+			deploy.Spec.Template.Annotations["olm.targetNamespaces"] = nsList.String()
+			_, err := a.client.Operators().ClusterServiceVersions(currentNamespace).Update(csv)
```
ecordell: We'll have a better chance of this succeeding if we send a `Patch` update.

pkg/controller/operators/olm/operator.go
```diff
@@ -158,6 +187,80 @@ func (a *Operator) syncDeployment(obj interface{}) (syncError error) {
...
+			if err != nil {
+				return fmt.Errorf("CSV update for '%v' failed: %v\n", csvName, err)
+			}
+			a.requeueCSV(csvName, currentNamespace)
```
ecordell: this will requeue the same csv multiple times if there're multiple deployments in the same csv

I'm also not sure we need to requeue - if this update causes the csv deployment to go unhealthy, we should detect that as an issue in the CSV anyway.

pkg/controller/operators/olm/operator.go
```diff
@@ -521,19 +624,55 @@ func (a *Operator) crdOwnerConflicts(in *v1alpha1.ClusterServiceVersion, csvsInN
...
 
-	log.Infof("syncing Namespace: %s", namespace.GetName())
+	log.Infof("syncing Namespace: %s", namespaceName)
 	if err := a.annotator.AnnotateNamespace(namespace); err != nil {
-		log.Infof("error annotating namespace '%s'", namespace.GetName())
+		log.Infof("error annotating namespace '%s'", namespaceName)
+		return err
+	}
+
+	opGroupList, err := a.client.OperatorsV1alpha2().OperatorGroups(namespaceName).List(metav1.ListOptions{})
```
ecordell: You can just query the operatorgroup informer here instead. @njhale can point you to an example in the package server :)

pkg/controller/operators/olm/operator.go
```diff
@@ -521,19 +624,55 @@ func (a *Operator) crdOwnerConflicts(in *v1alpha1.ClusterServiceVersion, csvsInN
...
+			for _, seenNamespace := range op.Status.Namespaces {
+				if seenNamespace.Name == namespaceName {
+					goto DoUpdate
+				}
+			}
+			needsUpdate = false
+		}
+	DoUpdate:
+		if needsUpdate {
+			if err := a.updateDeploymentAnnotation(&op); err != nil {
```
ecordell: this is going to update for every single namespace that's new, I don't think we want it to be in the loop. that way you can just break the loop and don't need a label/goto

pkg/controller/operators/olm/operator.go
```diff
@@ -521,19 +624,55 @@ func (a *Operator) crdOwnerConflicts(in *v1alpha1.ClusterServiceVersion, csvsInN
...
+	for _, op := range opGroupList.Items {
+		selector, err := metav1.LabelSelectorAsSelector(&op.Spec.Selector)
+		if err != nil {
+			return err
+		}
+
+		needsUpdate := true
+		if selector.Matches(labels.Set(namespace.GetLabels())) {
+			for _, seenNamespace := range op.Status.Namespaces {
+				if seenNamespace.Name == namespaceName {
```
ecordell: I think we want the same namespace set logic here

pkg/controller/operators/olm/operator.go
njhale: @jpeeler You can get a `Lister` from the `OperatorGroup` informer that lets you query the indexer's cache instead of making an API call. As an example, we [do this for checking RBAC resources](https://github.com/operator-framework/operator-lifecycle-manager/blob/177560083fafc69c3ec94444b2a6370f953a2feb/pkg/controller/operators/olm/operator.go#L122).

pkg/controller/operators/olm/operator.go
jpeeler: Do you mean you want the namespace check to query a map? (If so, this is only doing one namespace check, not an entire namespace list equality check.)

pkg/controller/operators/olm/operator.go
```diff
@@ -241,6 +279,212 @@ func (a *Operator) syncRBAC(obj interface{}) (syncError error) {
...
+		if err != nil {
+			return fmt.Errorf("error unmarshaling strategy from ClusterServiceVersion '%s' with error: %s", csvName, err)
+		}
+
+		strategyDetailsDeployment, ok := strategy.(*install.StrategyDetailsDeployment)
+		if !ok {
+			return fmt.Errorf("could not assert strategy implementation as deployment for CSV %s", csvName)
+		}
+
+		managerPolicyRules := []rbacv1.PolicyRule{}
```
ecordell: the rbac generation is broadly doing the right thing, but the installplan takes care of some of this already:

1. for any CRD it installs, it creates a 
   1. `<that-crd>-edit` ClusterRole
   1. `<that-crd>-view` ClusterRole

All that's really left for this control loop to do is make sure there's an `<operatorgroup>-edit` and an `<operatorgroup>-view` ClusterRole available (which I see that you have!)

So I think most of this will remain the same, you just don't need to worry about the specific edit/view clusterroles

pkg/controller/operators/olm/operator.go
```diff
@@ -241,6 +279,212 @@ func (a *Operator) syncRBAC(obj interface{}) (syncError error) {
...
+		clusterRole := &rbacv1.ClusterRole{
+			Rules: managerPolicyRules,
+		}
+		ownerutil.AddNonBlockingOwner(clusterRole, csv)
+		clusterRole.SetGenerateName(fmt.Sprintf("owned-crd-manager-%s-", csv.Spec.DisplayName))
+		createdClusterRole, err := a.OpClient.KubernetesInterface().RbacV1().ClusterRoles().Create(clusterRole)
+		if err != nil {
+			return err
+		}
+		for _, sa := range strategyDetailsDeployment.Permissions {
```
ecordell: These generated roles aren't for the serviceaccount for the operator to use, they're for end users and administrators.

It's so that a cluster admin can bind `jpeeler` to `etcd-edit` in namespace `foo` to give access to operator-provided services.

(we will need a separate thing to ensure that the serviceaccount for the operator has the correct permissions in target namespaces)

pkg/controller/operators/olm/operator.go
```diff
@@ -421,10 +421,33 @@ func (a *Operator) syncOperatorGroups(obj interface{}) error {
...
+		csvsInNamespace := a.csvsInNamespace(ns.Name)
+		for _, csv := range csvsInNamespace {
+			if csv.Status.Phase == v1alpha1.CSVPhaseSucceeded {
+				newCSV := csv.DeepCopy()
+				newCSV.Status = v1alpha1.ClusterServiceVersionStatus{
+					Message:        "CSV copied to target namespace",
+					Reason:         v1alpha1.CSVReasonCopied,
+					LastUpdateTime: timeNow(),
+				}
+				newCSV.Annotations["OriginalCSV"] = fmt.Sprintf("Namespace:%v, ResourceVersion:%v", csv.GetNamespace(), csv.GetResourceVersion())
```
ecordell: :+1: can we make this more machine readable `<namespace>/<csvname>`?

also we should follow the annotation key convention we used before, I think it would be `olm.originalCSV`?

pkg/controller/operators/olm/operator.go
```diff
@@ -241,6 +279,180 @@ func (a *Operator) syncRBAC(obj interface{}) (syncError error) {
...
+			managerPolicyRules = append(managerPolicyRules, rbacv1.PolicyRule{Verbs: []string{"*"}, APIGroups: []string{owned.Name}, Resources: resourceNames})
+			apiEditPolicyRules = append(apiEditPolicyRules, rbacv1.PolicyRule{Verbs: []string{"create", "update", "patch", "delete"}, APIGroups: []string{owned.Name}, Resources: []string{owned.Kind}})
+			apiViewPolicyRules = append(apiViewPolicyRules, rbacv1.PolicyRule{Verbs: []string{"get", "list", "watch"}, APIGroups: []string{owned.Name}, Resources: []string{owned.Kind}})
+		}
+		for _, owned := range csv.Spec.APIServiceDefinitions.Owned {
+			resourceNames := []string{}
+			for _, resource := range owned.Resources {
+				resourceNames = append(resourceNames, resource.Name)
+			}
+			managerPolicyRules = append(managerPolicyRules, rbacv1.PolicyRule{Verbs: []string{"*"}, APIGroups: []string{owned.Name}, Resources: resourceNames})
```
njhale: I don't think Resources belong to the APIService's API group.

jpeeler: Can you clarify this with respect to the comment made here: https://github.com/operator-framework/operator-lifecycle-manager/pull/480#discussion_r221736953

jpeeler:
@ecordell I'm working on rebasing next, but I just pushed the simple e2e and associated fixes.

jpeeler:
@ecordell Please review when you get a chance. Although it's probably best to review holistically, I tried to make each commit reviewable by itself.

jpeeler:
A few concerns that might need special attention:
1.  rbac - luckily the rules can just be fixed as needed
2. out of sync issues - there's bound to be some type of problem with non-first edits
3. CSV copying - currently it copies CSVs no matter the phase

jpeeler:
/retest
currently timing out on yum update failures

pkg/lib/queueinformer/queueinformer.go
```diff
@@ -105,10 +111,26 @@ func NewInformer(queue workqueue.RateLimitingInterface, informer cache.SharedInd
...
+	if funcs != nil {
+		if funcs.AddFunc != nil {
+			queueInformer.resourceEventHandlerFuncs.AddFunc = func(obj interface{}) {
+				funcs.AddFunc(obj)
+				queueInformer.defaultAddFunc(obj)
+			}
+		}
+		if funcs.DeleteFunc != nil {
+			queueInformer.resourceEventHandlerFuncs.DeleteFunc = func(obj interface{}) {
+				funcs.DeleteFunc(obj)
```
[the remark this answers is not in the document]
jpeeler: It actually doesn't circumvent it though, it just adds a hook. That seem any better?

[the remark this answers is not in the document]
njhale: I was just thinking that the potential is there to perform sync operations as each event comes in before its key get enqueued, which means you'd be able to do work without being rate-limited by the queue. I guess you could do that in previous versions by just passing in a custom event handler struct anyway. My previous comment was just following my thought process through, it's nbd.

pkg/controller/operators/olm/operator.go
```diff
@@ -625,19 +748,21 @@ func (a *Operator) crdOwnerConflicts(in *v1alpha1.ClusterServiceVersion, csvsInN
 	return nil
 }
 
-// annotateNamespace is the method that gets called when we see a namespace event in the cluster
-func (a *Operator) annotateNamespace(obj interface{}) (syncError error) {
+// syncNamespace is the method that gets called when we see a namespace event in the cluster
+func (a *Operator) syncNamespace(obj interface{}) (syncError error) {
 	namespace, ok := obj.(*corev1.Namespace)
+	namespaceName := namespace.GetName()
```
ecordell: this should move below the `ok` check

jpeeler:
/retest

Looks like CI operator is having a bad day:
```
panic: runtime error: index out of range

goroutine 135 [running]:
github.com/openshift/ci-operator/pkg/steps.waitForPodCompletion(0x1285700, 0xc4204b55c0, 0xc4204ac330, 0xe, 0x1275cc0, 0xc4202aa2d0, 0x7ffe49a0613e, 0xf, 0x0, 0xc420859520)
	/go/src/github.com/openshift/ci-operator/pkg/steps/template.go:572 +0x3f7
github.com/openshift/ci-operator/pkg/steps.(*podStep).Run(0xc4205aa1c0, 0x1275a40, 0xc4200517c0, 0x0, 0x5, 0xc42036a400)
	/go/src/github.com/openshift/ci-operator/pkg/steps/test.go:120 +0xc4f
github.com/openshift/ci-operator/pkg/steps/release.(*assembleReleaseStep).Run(0xc4204315e0, 0x1275a40, 0xc4200517c0, 0xbeef49f166006400, 0x65f0f3ff12, 0x1974400)
	/go/src/github.com/openshift/ci-operator/pkg/steps/release/create_release.go:83 +0x773
github.com/openshift/ci-operator/pkg/steps.runStep(0x1275a40, 0xc4200517c0, 0xc420213140, 0xc420322f00, 0x0)
	/go/src/github.com/openshift/ci-operator/pkg/steps/run.go:105 +0x92
created by github.com/openshift/ci-operator/pkg/steps.Run
	/go/src/github.com/openshift/ci-operator/pkg/steps/run.go:71 +0x3a1```

File: pkg/controller/operators/olm/operator.go [hunk 22]
```diff
@@ -32,19 +36,24 @@ import (
...
-	resolver    install.StrategyResolverInterface
-	lister      operatorlister.OperatorLister
-	annotator   *annotator.Annotator
-	recorder    record.EventRecorder
-	cleanupFunc func()
+	csvQueue            workqueue.RateLimitingInterface
+	client              versioned.Interface
+	resolver            install.StrategyResolverInterface
+	lister              operatorlister.OperatorLister
+	csvLister           map[string]csvlister.ClusterServiceVersionLister
```
njhale: After some fixes in [PR-536](https://github.com/operator-framework/operator-lifecycle-manager/pull/536) get in, we can update this to extend `operatorlister.OperatorLister`.

pkg/controller/operators/olm/operator.go
```diff
@@ -21,6 +24,7 @@ import (
 	"k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
```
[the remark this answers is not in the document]
jpeeler: Can you elaborate? I just always let goimports do its thing.

pkg/controller/operators/olm/operator.go
```diff
@@ -330,10 +428,35 @@ func (a *Operator) syncClusterServiceVersion(obj interface{}) (syncError error)
...
+		// ensure "parent" CSV has not been deleted
+		csvs := a.csvsInNamespace(opNamespace)
+		csv, found := csvs[clusterServiceVersion.Name]
+		if !found {
+			targetNamespaces, ok := clusterServiceVersion.Annotations["olm.targetNamespaces"]
+			if !ok {
+				return fmt.Errorf("Did not find targetNamespaces annotation on %v", csv)
+			}
+			for _, ns := range strings.Split(targetNamespaces, ",") {
+				if err := a.client.OperatorsV1alpha1().ClusterServiceVersions(ns).Delete(csv.Name, &metav1.DeleteOptions{}); err != nil {
```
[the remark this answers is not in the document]
jpeeler: You're right, must have missed this during refactoring. It's fixed now.

pkg/controller/operators/olm/operator_test.go
```diff
@@ -1073,6 +1083,238 @@ func TestTransitionCSV(t *testing.T) {
...
+		t.Run(tc.name, func(t *testing.T) {
+			op, err := NewFakeOperator(tc.initialCsvs, tc.initialObjs, tc.initialCrds, tc.initialApis, &install.StrategyResolver{}, tc.namespaces)
+			require.NoError(t, err)
+
+			stopCh := make(chan struct{})
+			defer func() { stopCh <- struct{}{} }()
+			ready, _ := op.Run(stopCh)
+			<-ready
+
+			// Could not put this in initialObjs - got "no kind is registered for the type v1alpha2.OperatorGroup"
```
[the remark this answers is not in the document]
jpeeler: I had forgotten about this. After refactoring a bit, this is now working fine.

pkg/controller/operators/olm/operator_test.go
```diff
@@ -1073,6 +1083,238 @@ func TestTransitionCSV(t *testing.T) {
...
+			_, err = op.client.OperatorsV1alpha2().OperatorGroups(tc.inputGroup.Namespace).Create(&tc.inputGroup)
```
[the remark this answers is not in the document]
jpeeler: I assume after creating everything using NewFakeOperator that this problem should be fixed now.

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -0,0 +1,284 @@
...
+	}
+	for _, v := range statusNamespaces {
+		if _, ok := nsMap[v.Name]; !ok {
+			return true
+		}
+	}
+	return false
+}
+
+func (a *Operator) updateNamespaceList(op *v1alpha2.OperatorGroup) (error, []*corev1.Namespace) {
```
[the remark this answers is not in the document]
jpeeler: Fair enough, fixed.

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -0,0 +1,284 @@
...
+
+func (a *Operator) syncOperatorGroups(obj interface{}) error {
+	op, ok := obj.(*v1alpha2.OperatorGroup)
+	if !ok {
+		log.Debugf("wrong type: %#v\n", obj)
+		return fmt.Errorf("casting OperatorGroup failed")
+	}
+	log.Infof("syncing operator group %v", op)
+
+	err, targetedNamespaces := a.updateNamespaceList(op)
```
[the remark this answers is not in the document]
jpeeler: Sync is definitely called for status updates too? If an exit occurs on initial successful update, the other work will be unnecessarily delayed in my opinion. The namespace list status update is the only change to the operator group that is made. (And in theory if the operator group isn't modified it'd be just a one time extra sync.)

pkg/controller/operators/olm/operator_test.go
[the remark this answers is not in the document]
njhale: Well, if we just use NewFakeOperator without @ecordell 's changes, the sync handler will run anyway when we start the operator (`ready, _ := op.Run(...)`. If this is working for you now without issues that we can always address this in a follow up when those changes are in.

deploy/chart/templates/0000_30_14-operatorgroup.crd.yaml
```diff
@@ -0,0 +1,49 @@
...
+  subresources:
+    # status enables the status subresource.
+    status: {}
+  validation:
+    openAPIV3Schema:
+      properties:
+        spec:
+          properties:
+            selector:
+              type: object
```
ecordell: We should be explicit about accepting a label selector here, see: https://github.com/operator-framework/operator-lifecycle-manager/blob/master/deploy/chart/templates/0000_30_02-clusterserviceversion.crd.yaml#L153-L182

pkg/api/apis/operators/v1alpha2/operatorgroup_types.go
```diff
@@ -0,0 +1,34 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type OperatorGroupSpec struct {
+	Selector       metav1.LabelSelector  `json:"selector,omitempty"`
+	ServiceAccount corev1.ServiceAccount `json:"serviceAccount,omitempty"`
+}
+
+type OperatorGroupStatus struct {
+	Namespaces  []*corev1.Namespace `json:"namespaces"`
```
ecordell: I'm still a little concerned that this could get too long. `[]string` seems slightly safer. Perhaps the real answer is to wait for CRDs to support arbitrary subresources and implement a `/namespaces` subresource (so that we can paginate)

pkg/controller/operators/olm/operator.go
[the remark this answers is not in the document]
njhale: This doesn't look split up correctly. It should be:
`
import(
standard packages

external stuff

internal stuff
)

From what I can see, it looks like:
import (
standard packages

internal stuff
external stuff
)

pkg/controller/operators/olm/operator_test.go
```diff
@@ -99,17 +101,24 @@ func apiResourcesForObjects(objs []runtime.Object) []*metav1.APIResourceList {
...
+func NewFakeOperator(clientObjs []runtime.Object, k8sObjs []runtime.Object, extObjs []runtime.Object, regObjs []runtime.Object, resolver install.StrategyResolverInterface, namespaces []v1.Namespace) (*Operator, error) {
 	clientFake := fake.NewSimpleClientset(clientObjs...)
 	k8sClientFake := k8sfake.NewSimpleClientset(k8sObjs...)
 	k8sClientFake.Resources = apiResourcesForObjects(append(extObjs, regObjs...))
 	opClientFake := operatorclient.NewClient(k8sClientFake, apiextensionsfake.NewSimpleClientset(extObjs...), apiregistrationfake.NewSimpleClientset(regObjs...))
 	annotations := map[string]string{"test": "annotation"}
-	_, err := opClientFake.KubernetesInterface().CoreV1().Namespaces().Create(&v1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: namespace}})
-	if err != nil {
-		return nil, err
+	for _, ns := range namespaces {
```
ecordell: not a big deal, but if you're passing the objects in like this you can just append the namespaces to `k8sObjs` and they'll exist in the fake for testing. or you can remove the explicit `namespaces` param and build the object list in the tests

PR #484, opened 2018-09-26
re-cut 0.7.1

ecordell:
makes `manifests` a real directory and renames the openshift namespace
to include the openshift prefix


jpeeler:
not trying to nit, but is it image-references or image_references?

ecordell:
@jpeeler `image-references`, I just wrote it wrong on the commit message

PR #491, opened 2018-09-28
rh-operators: add FederationV2 v0.0.2

font:
This adds the FederationV2 v0.0.2 controller (operator) to the catalog of OLM Red Hat operators.

cc @ecordell @pmorie


deploy/chart/catalog_resources/rh-operators/federationv2.v0.0.2.clusterserviceversion.yaml
```diff
@@ -0,0 +1,259 @@
+apiVersion: operators.coreos.com/v1alpha1
+kind: ClusterServiceVersion
+metadata:
+  name: federationv2.v0.0.2
+spec:
```
pmorie: We should add a description to this - suggest something like "Kubernetes Federation V2 namespace-scoped installation"

pmorie: I also notice there's a `maturity` field - that should indicate that this is an alpha maturity component.

pmorie: One final note, this also has me thinking about what an icon for federation would look like...

deploy/chart/catalog_resources/rh-operators/federationv2.crd.yaml
```diff
@@ -0,0 +1,1396 @@
...
+              type: array
+          type: object
+      type: object
+  version: v1alpha1
+status:
+  acceptedNames:
+    kind: ""
+    plural: ""
+  conditions: null
+---
```
ecordell: Unfortunately these need to be split out into separate files at the moment

font: Yikes! There are a lot of CRDs.

font:
@ecordell @pmorie I have updated based on feedback.

@ecordell The `/test images` failure appears to be an environment issue.

ecordell:
Thanks @font - I want to check out the installation before we merge. Have you tested that these manifests install federation with OLM?

font:
> Thanks @font - I want to check out the installation before we merge. Have you tested that these manifests install federation with OLM?

@ecordell I have tested on minikube that OLM installs federation after manually creating the CRDs + CSV for federation. I have also tested with minishift using the same approach. Both work.

I have also attempted to try to test with the `make run-local-shift` command but ran into setup errors so haven't been able to test that yet.

Is there a recommended way to test this?

PR #492, opened 2018-10-01
feat(csv): install owned APIServices

njhale:
### Description

Extends the CSV control-loop to install owned `APIServices`. This includes generating and injecting associated serving certificates as volume mounted `Secrets` in matching `Deployments`, creating/updating a `Service` to point to the pods of said `Deployments`, and creating/updating `APIServices` with the  CA bundle of the signing CA. 

### Note:
This is a first attempt at something that _creates_ all the required resources to establish trust with the OLM managed extension api-server. Managing cert expiration will be the next step. At this time tests still need to be written for this.

Related to [ALM-657](https://jira.coreos.com/browse/ALM-657)


pkg/api/apis/operators/v1alpha1/clusterserviceversion_types.go
```diff
@@ -65,9 +66,10 @@ type CRDDescription struct {
 
 // APIServiceDescription provides details to OLM about apis provided via aggregation
 type APIServiceDescription struct {
-	Name              string                 `json:"name"`
+	Group             string                 `json:"group"`
 	Version           string                 `json:"version"`
 	Kind              string                 `json:"kind"`
+	DeploymentName    string                 `json:"deploymentName,omitempty"`
```
ecordell: I wonder if this should accept a container as well? `deploymentName: packages-apiserver/pkg-container`

deploy/chart/templates/30_02-clusterserviceversion.crd.yaml
```diff
@@ -189,15 +189,15 @@ spec:
                   items:
```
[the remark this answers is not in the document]
njhale: yes - thanks for catching that.

ecordell:
Just hit an interesting issue: if the port used to expose the apiserver isn't 443, we get stuck in installing (but no info about why).

We should probably either have a way to specify `deployment/container:port` for an apiserver, or detect when the port we expect (443) isn't exposed on the pod that's designated as an apiserver.

ecordell:
Also, thinking about the volume issue - what if we just had a standard volume name (like `apiserver-certs`) and then if there's no volumeMount for that volume, we add the "standard" one, otherwise we use the one that's there. That lets you take an existing apiserver (that's looking in a particular location for certs) and package it, without adding extra config.

njhale:
> Just hit an interesting issue: if the port used to expose the apiserver isn't 443, we get stuck in installing (but no info about why).
> 
> We should probably either have a way to specify `deployment/container:port` for an apiserver, or detect when the port we expect (443) isn't exposed on the pod that's designated as an apiserver.

So [`PodSpec.Containers[].Ports[]`](https://github.com/kubernetes/api/blob/a191abe0b71e00ce4cde58af8002aa4c1a8bb068/core/v1/types.go#L2063) is optional, and the omission of a container port doesn't preclude it from being exposed to the network. So I'm thinking that we just have an optional `containerPort` field that overrides a default `targetPort` of 443 on the generated Service.

deploy/chart/templates/30_02-clusterserviceversion.crd.yaml
```diff
@@ -189,21 +189,28 @@ spec:
...
                     type: object
                     required:
-                    - name
+                    - group
```
alecmerdler: Most of the e2e failures are regarding API validation for `ClusterServiceVersions` on tests you didn't modify. You may need to make some fields optional to fix them.

pkg/controller/operators/olm/operator.go
```diff
@@ -442,22 +460,37 @@ func (a *Operator) checkReplacementsAndUpdateStatus(csv *v1alpha1.ClusterService
...
 	// installcheck determined we can't progress (e.g. deployment failed to come up in time)
 	if install.IsErrorUnrecoverable(strategyErr) {
 		csv.SetPhase(v1alpha1.CSVPhaseFailed, v1alpha1.CSVReasonInstallCheckFailed, fmt.Sprintf("install failed: %s", strategyErr))
 		return strategyErr
 	}
 
 	// if there's an error checking install that shouldn't fail the strategy, requeue with message
+	if apiServiceErr != nil {
+		csv.SetPhase(v1alpha1.CSVPhaseInstalling, requeueConditionReason, fmt.Sprintf("APIServices installing: %s", apiServiceErr))
+		a.requeueCSV(csv.GetName(), csv.GetNamespace())
```
njhale: I'm not sure that this is the best way to ensure APIService GVK discovery is immediately detected. Watching APIServices only tells us when its status changes to available, but not when the expected GVKs are discoverable (which can happen much later). I expect this to swamp the CSV workqueue. 

@ecordell Could we potentially watch Discovery for changes or start a go-routine when we detect a change to an APIService that requeues the CSV when the appropriate GVKs become available?

pkg/controller/operators/olm/apiservices.go
```diff
@@ -0,0 +1,428 @@
...
+		return nil, err
+	}
+
+	// change depSpec.Template's name to force rollout
+	// ensures that fresh secret is used by apiserver if not hot reloading
+	podTemplateName := names.SimpleNameGenerator.GenerateName(apiServiceName)
+	if podTemplateName == depSpec.Template.GetName() {
+		return nil, fmt.Errorf("a name collision occured when generating name for PodTemplate")
+	}
+	depSpec.Template.SetName(podTemplateName)
```
njhale: This ensures that the extension api-server is actually rolled out in the event of a secret change. Triggering rollouts this way allows us to keep a static secret name.

pkg/controller/operators/olm/apiservices.go
```diff
@@ -0,0 +1,428 @@
...
+
+func (a *Operator) syncAPIServices(obj interface{}) (syncError error) {
+	apiService, ok := obj.(*apiregistrationv1.APIService)
+	if !ok {
+		log.Debugf("wrong type: %#v", obj)
+		return fmt.Errorf("casting APIService failed")
+	}
+	if ownerutil.IsOwnedByKind(apiService, v1alpha1.ClusterServiceVersionKind) {
+		oref := ownerutil.GetOwnerByKind(apiService, v1alpha1.ClusterServiceVersionKind)
+		log.Infof("APIService %s change requeuing CSV %s", apiService.GetName(), oref.Name)
```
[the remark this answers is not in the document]
njhale: switching to `log.Infof(...)`

pkg/controller/operators/olm/operator.go
njhale: [ALM-749](https://jira.coreos.com/browse/ALM-749)

deploy/chart/catalog_resources/rh-operators/prometheusoperator.0.22.2.clusterserviceversion.yaml
```diff
@@ -158,18 +158,18 @@ spec:
               serviceAccount: prometheus-operator-0-22-2
               containers:
               - name: prometheus-operator
-                image: quay.io/coreos/prometheus-operator@sha256:3daa69a8c6c2f1d35dcf1fe48a7cd8b230e55f5229a1ded438f687debade5bcf
+                image: quay.io/coreos/prometheus-operator:master
```
ecordell: could you remove the changes from this file? this was just for testing out the operatorgroup in that combined branch for the demo, it's not something we want to change just yet

PR #496, opened 2018-10-02
Dropping label for origin releasepayload

abhinavdahiya:
The olm manifests are blocking ClusterVersionOperator intergration in
`openshift/installer`. 

xref: https://github.com/openshift/installer/pull/330#issuecomment-426119496
cc @smarterclayton


smarterclayton:
@ecordell can we temporarily disable to unblock?

smarterclayton:
Tagging, we should be able to reenable soon

PR #498, opened 2018-10-04
Fix for Federation CSV

alecmerdler:
### Description

Adds `spec.provider` field.


alecmerdler:
@font Does Federation have a logo you would like to have displayed?

font:
@alecmerdler Unfortunately, we do not have a logo yet. We will need to create one...

PR #499, opened 2018-10-04
Make use a valid clusterPermissions rule verb

wongma7:
"use" is a valid verb for RBAC in openshift 3.11 for SCCs, i.e. when the `resource` is security.openshift.io/securitycontextconstraints.

see: https://github.com/openshift/openshift-docs/pull/10376


njhale:
Thanks for adding this @wongma7 ! 
/approve

PR #504, opened 2018-10-09
fix(ci): use deployment namespace for catalog_namespace

njhale:
### Description

Prepends a catalog_namespace creation step to the CI deploy script.


ecordell:
Maybe we should just change the `catalogNamespace` for staging to be the same as the deployment namespace?

Won't this error out because the namespace is in the chart now?

PR #510, opened 2018-10-10
chore(deploy): update release scripts to not use 3 different images

ecordell:
This is to fix: https://gitlab.com/operator-framework/operator-framework_operator-lifecycle-manager/-/jobs/106854881


.gitlab-ci/base_jobs.libsonnet
```diff
@@ -22,16 +22,16 @@ local appr = utils.appr;
 
     AlmTest: {
```
njhale: OlmTest?

ecordell: Deleted - I removed the unit tests from gitlab since they run in the other ci system now

PR #513, opened 2018-10-11
Rewrite Namespace when Fetching Global PackageManifests

alecmerdler:
### Description

Ensures that all the the `PackageManifests` returned have the same `metadata.namespace` value.

Fixes https://jira.coreos.com/browse/ALM-752

### DIFF
--- a/pkg/package-server/provider/inmem.go
+++ b/pkg/package-server/provider/inmem.go
@@ -248,6 +248,7 @@ func (m *InMemoryProvider) Get(namespace, name string) (*packagev1alpha1.Package
 
 	for key, pm := range m.manifests {
 		if key.packageName == name && (key.catalogSourceNamespace == namespace || key.catalogSourceNamespace == m.globalNamespace) {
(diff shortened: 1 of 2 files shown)


pkg/package-server/provider/inmem.go
```diff
@@ -248,6 +248,7 @@ func (m *InMemoryProvider) Get(namespace, name string) (*packagev1alpha1.Package
 
 	for key, pm := range m.manifests {
 		if key.packageName == name && (key.catalogSourceNamespace == namespace || key.catalogSourceNamespace == m.globalNamespace) {
+			pm.SetNamespace(namespace)
```
njhale: The one thing that I think could become a problem in the future is when we have multiple packages of the same name in both the "global" and request namespace. In that situation you would have a 50% chance of getting the "global" package instead of the one you actually requested. Maybe for a get of an individual resource, you should only get the "global" package if no package of the same name exists in the requested namespace.

PR #514, opened 2018-10-12
fix(requirements): add support for non resource url rules

njhale:
### Description
- Adds support for NonResourceURL rules as CSV requirements
- Adds CRD served version CSV requirement check

Fixes: [ALM-755](https://jira.coreos.com/browse/ALM-755) and [ALM-756](https://jira.coreos.com/browse/ALM-756)

### DIFF
--- a/pkg/controller/install/rule_checker.go
+++ b/pkg/controller/install/rule_checker.go
@@ -1,6 +1,8 @@
 package install
 
 import (
+	"fmt"
+
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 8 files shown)


eriknelson:
Tested with this patch against olm master and it's working great :+1:

pkg/controller/install/rule_checker.go
```diff
@@ -155,3 +163,16 @@ func (c *CSVRuleChecker) hasOwnerConflicts(ownerRefs []metav1.OwnerReference) bo
 
 	return conflicts
 }
+
+func ruleValid(rule rbacv1.PolicyRule) error {
+	if len(rule.Verbs) == 0 {
```
ecordell: does the particular openapi flavor kube supports let you specify `minItems` length on the verb list?

PR #516, opened 2018-10-15
feat(build): allow skipping minikube during local builds

jpeeler:
This only changes behavior if the NO_MINIKUBE variable is set, otherwise minikube calls will be done exactly as before. (This probably only helps me out, but I'd really appreciate not having to comment out lines or apply a patch for each new branch I work on!)

### DIFF
--- a/scripts/build_local.sh
+++ b/scripts/build_local.sh
@@ -5,9 +5,11 @@
 
 set -e
 
-ps x | grep -q [m]inikube || minikube start --kubernetes-version="v1.12.0" --extra-config=apiserver.v=4 || { echo 'Cannot start minikube.'; exit 1; }
-eval $(minikube docker-env) || { echo 'Cannot switch to minikube docker'; exit 1; }
-kubectl config use-context minikube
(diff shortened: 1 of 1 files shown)


njhale:
Is the aim of this to avoid calling `minikube start...` again?

njhale:
What does your test flow look like then? Do you end up using the `olm-e2e` image?

jpeeler:
Yes, I use the olm-e2e image. Everything happens locally on my machine and nothing happens in a VM. I just set up my cluster beforehand and set the appropriate environment variables. (I use https://direnv.net  to manage the collection of environment variables based on the directory I'm in too.)

jpeeler:
Actually, I've been using the e2e-local target which ultimately ends up executing go test directly. But it looks like there's a docker target that can use the olm-e2e image.

I'm asking that this get merged so that I can stop patching every local e2e I run. If others could benefit from a similar workflow, I don't mind working on adding some helper scripts to get a cluster easily running using the upstream hack script later on.

PR #517, opened 2018-10-15
fix(build): add cleanup for package release files

jpeeler:
scripts/run_e2e_local.sh is slightly different from scripts/run_e2e_docker.sh. Make sure to delete directories for both scenarios.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -97,6 +97,7 @@ container: build
 
 clean:
 	rm -rf bin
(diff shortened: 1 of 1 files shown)


jpeeler:
Somehow the gitlab CI ran e2e tests from https://github.com/operator-framework/operator-lifecycle-manager/pull/480?

PR #518, opened 2018-10-15
feat(deploy): add debug variable to all deployments

jpeeler:
This sets the default value of debug to true for local deploys too.


jpeeler:
/test e2e-aws
Test failed due to some networking error.

PR #522, opened 2018-10-16
More Descriptors for Dynatrace Operator

alecmerdler:
### Description

Adds more `spec` and `status` descriptors for the `OneAgent` object based on [the source code](https://github.com/Dynatrace/dynatrace-oneagent-operator/blob/master/pkg/apis/dynatrace/v1alpha1/types.go). Also fixes the `ClusterServiceVersion` CRD to allow default descriptor values of types other than `object`.

Addresses https://jira.coreos.com/browse/ALM-754

### DIFF
--- a/deploy/chart/catalog_resources/certified-operators/mongodb-enterprise.v0.3.2.clusterserviceversion.yaml
+++ b/deploy/chart/catalog_resources/certified-operators/mongodb-enterprise.v0.3.2.clusterserviceversion.yaml
@@ -3,11 +3,9 @@ kind: ClusterServiceVersion
 metadata:
   name: mongodboperator.v0.3.2
   namespace: placeholder
-
   annotations:
     alm-examples: >-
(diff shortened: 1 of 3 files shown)


deploy/chart/catalog_resources/certified-operators/mongodb-enterprise.v0.3.2.clusterserviceversion.yaml
```diff
@@ -85,8 +79,8 @@ spec:
       group: mongodb.com
       displayName: MongoDB Standalone
       description: MongoDB Deployment consisting of only one host. No replication of data.
-
       resources:
+      # FIXME(alecmerdler): These are probably supposed to be different resources for the different CRD configurations
```
alecmerdler: @robszumski I feel like these resources should be different depending on if you are creating a `MongoDbStandalone`, `    kind: MongoDbShardedCluster`, or `MongoDbReplicaSet`.

PR #523, opened 2018-10-16
fix(olm): don't annotate namespaces on operator creation

ecordell:
instead let the standard apimachinary take care of calling `syncNamespace`

There's no need for it, and it fixes OLM startup issues we were seeing on OpenShift 4


njhale:
@wking This looks like a fix for `olm-operator`. This other [PR](https://github.com/operator-framework/operator-lifecycle-manager/pull/524) looks like the fix for `package-server`.

wking:
Looks like it needs a rebase.

PR #525, opened 2018-10-16
feat(csv): add cert rotation for owned APIServices

njhale:
### Description
This PR ensures that generated certs for owned APIServices are rotated before they expire by some minimum amount of time. Two fields are added to `ClusterServiceVersionStatus` to track cert rotation, `CertsRotateAt` and `CertsLastUpdated`. 

### Notes:
APIService e2e tests use a repurposed `package-server` deployment. This can interfere with later package-server tests and is generally messy. In or following this PR I hope to incorporate a new configurable extension api-server that allows arbitrary group/versions to be set via a startup flag. This should help to alleviate some of this messiness.

### DIFF
--- a/deploy/chart/templates/0000_30_02-clusterserviceversion.crd.yaml
+++ b/deploy/chart/templates/0000_30_02-clusterserviceversion.crd.yaml
@@ -180,17 +180,7 @@ spec:
                           values:
                             type: array
                             description: set of values for the expression
-            certValidForDays:
-              type: integer
-              description: Number of days generated APIService certs are valid for
-              format: int32
-              minimum: 1
-              maximum: 730
-            certMinFreshSeconds:
-              type: integer
-              description: Minimum freshness of a cert before it is renewed; in seconds
-              format: int32
-              minimum: 10
+                            
             apiservicedefinitions:
               type: object
               properties:
--- a/deploy/chart/templates/0000_30_02-clusterserviceversion.crd.yaml
+++ b/deploy/chart/templates/0000_30_02-clusterserviceversion.crd.yaml
@@ -180,6 +180,17 @@ spec:
                           values:
                             type: array
                             description: set of values for the expression
+            certValidForDays:
+              type: integer
+              description: Number of days generated APIService certs are valid for
+              format: int32
+              minimum: 1
+              maximum: 730
+            certMinFreshSeconds:
+              type: integer
+              description: Minimum freshness of a cert before it is renewed; in seconds
+              format: int32
+              minimum: 10
             apiservicedefinitions:
               type: object
               properties:
--- a/pkg/api/apis/operators/v1alpha1/clusterserviceversion_types.go
+++ b/pkg/api/apis/operators/v1alpha1/clusterserviceversion_types.go
@@ -135,6 +135,14 @@ type ClusterServiceVersionSpec struct {
 	// Label selector for related resources.
 	// +optional
 	Selector *metav1.LabelSelector `json:"selector,omitempty" protobuf:"bytes,2,opt,name=selector"`
+
+	// Number of days generated owned APIService certs are valid for
+	// +optional
+	CertValidForDays int `json:"certValidFor,omitempty"`
+
+	// Minimum freshness of a cert before it is renewed; in seconds
+	// +optional
+	CertMinFreshSeconds int `json:"certMinFresh,omitempty"`
 }
 
 type Maintainer struct {
@@ -225,7 +233,7 @@ func (csv ClusterServiceVersion) OwnsCRD(name string) bool {
 	return false
 }
 
-// ConditionReason is a camelcased reason for the status of a RequirementStatus or DependentStatus
+// StatusReason is a camelcased reason for the status of a RequirementStatus or DependentStatus
 type StatusReason string
 
 const (
@@ -256,13 +264,6 @@ type RequirementStatus struct {
 	Dependents []DependentStatus `json:"dependents,omitempty"`
 }
 
-type CertExpiration struct {
-	APIServiceGroup   string `json:"apiServiceName"`
-	APIServiceVersion string ``
-	// Time when the APIService cert expires
-	ValidUntil metav1.Time `json:"validUntil"`
-}
-
 // ClusterServiceVersionStatus represents information about the status of a pod. Status may trail the actual
 // state of a system.
 type ClusterServiceVersionStatus struct {
@@ -285,14 +286,14 @@ type ClusterServiceVersionStatus struct {
 	Conditions []ClusterServiceVersionCondition `json:"conditions,omitempty"`
 	// The status of each requirement for this CSV
 	RequirementStatus []RequirementStatus `json:"requirementStatus,omitempty"`
-	// Cert expiration dates for owned APIServices
+	// Time to refresh generated owned APIService certs
 	// +optional
-	CertExpirations []CertExpiration `json:"certExpirations,omitempty"`
+	CertRefresh metav1.Time `json:"requirementStatus,omitempty"`
 }
 
+// ClusterServiceVersion is a Custom Resource of type `ClusterServiceVersionSpec`.
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 // +genclient
-// ClusterServiceVersion is a Custom Resource of type `ClusterServiceVersionSpec`.
 type ClusterServiceVersion struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata"`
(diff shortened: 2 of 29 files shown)


deploy/chart/templates/0000_30_02-clusterserviceversion.crd.yaml
```diff
@@ -180,6 +180,17 @@ spec:
...
+              type: integer
+              description: Number of days generated APIService certs are valid for
+              format: int32
+              minimum: 1
+              maximum: 730
+            certMinFreshSeconds:
+              type: integer
+              description: Minimum freshness of a cert before it is renewed; in seconds
+              format: int32
+              minimum: 10
```
ecordell: (commenting here to so that my comments read correctly in order)

Do we even want to expose these flags? Can't we just pick good defaults and ensure users don't have to worry about them?

pkg/api/apis/operators/v1alpha1/clusterserviceversion_types.go
```diff
@@ -278,11 +287,14 @@ type ClusterServiceVersionStatus struct {
 	Conditions []ClusterServiceVersionCondition `json:"conditions,omitempty"`
 	// The status of each requirement for this CSV
 	RequirementStatus []RequirementStatus `json:"requirementStatus,omitempty"`
+	// Time to refresh generated owned APIService certs
+	// +optional
+	CertRefresh metav1.Time `json:"certRefresh,omitempty"`
```
ecordell: Does it make more sense for the status to store `CertLastUpdated` and `CertExpiresAt`?

pkg/controller/operators/olm/apiservices.go
```diff
@@ -70,34 +91,55 @@ func (a *Operator) installOwnedAPIServiceRequirements(csv *v1alpha1.ClusterServi
...
-
 	// Assume the strategy is for a deployment
 	strategyDetailsDeployment, ok := strategy.(*install.StrategyDetailsDeployment)
 	if !ok {
 		return nil, fmt.Errorf("unsupported InstallStrategy type")
 	}
 
+	// Return early if there are no owned APIServices
+	if len(csv.Spec.APIServiceDefinitions.Owned) == 0 {
+		return strategyDetailsDeployment, nil
```
ecordell: 👍

pkg/controller/operators/olm/apiservices.go
```diff
@@ -378,50 +424,67 @@ func (a *Operator) installAPIServiceRequirements(desc v1alpha1.APIServiceDescrip
...
 	}
 
-	// create APIService with fresh CA bundle
-	caCertPEM, _, err := ca.ToPEM()
-	if err != nil {
-		logger.Debugf("unable to convert CA certificate to PEM format for APIService %s", apiServiceName)
-		return nil, err
+	// changing depSpec.Template's name to force a rollout
+	// ensures that the new secret is used by the apiserver if not hot reloading
+	podTemplateName := names.SimpleNameGenerator.GenerateName(apiServiceName)
```
ecordell: Is there a safer way to rollout? It's unlikely but there may be dependencies on the pod names. 

Does updating metadata trigger the rollout? Could we add a short hash of the cert as an annotation or something?

pkg/api/apis/operators/v1alpha1/clusterserviceversion_types.go
njhale: Yes, it does. I just didn't want to hide when the cert was actually expected to be refreshed: `CertRefresh = ExpiresAt - MinFresh`.

pkg/controller/operators/olm/apiservices.go
njhale: [Yes, it does ](https://github.com/kubernetes/kubernetes/issues/27081#issuecomment-229252547) - will do

pkg/controller/certs/certs.go
```diff
@@ -40,16 +41,23 @@ func (kp *KeyPair) ToPEM() (certPEM []byte, privPEM []byte, err error) {
...
+	if expiresIn > 730 || expiresIn <= 0 {
+		return nil, time.Time{}, fmt.Errorf("invalid cert expiration")
+	}
+
+	notBefore := time.Now()
+	notAfter := notBefore.AddDate(0, 0, expiresIn)
+
 	caDetails := &x509.Certificate{
 		//TODO(Nick): figure out what to use for a SerialNumber
 		SerialNumber: big.NewInt(1653),
```
jpeeler: I'm wondering if you could utilize https://github.com/kubernetes/client-go/blob/982e3f55001f505adeeafb7f559affe3c7594420/util/cert/cert.go#L68 rather than doing all this. I don't think it matters what the serial number is for self-signed certificates, but otherwise I think a random number needs to be used for security reasons.

njhale: Thanks for finding that! My only concern is that it seems like the expiration is hardcoded - though that may not even be a big deal 😄

pkg/controller/operators/olm/apiservices.go
```diff
@@ -35,6 +44,96 @@ func (a *Operator) syncAPIServices(obj interface{}) (syncError error) {
...
+		apiServiceName := fmt.Sprintf("%s.%s", desc.Version, desc.Group)
+		logger := log.WithFields(log.Fields{
+			"csv":        csv.GetName(),
+			"namespace":  csv.GetNamespace(),
+			"apiservice": apiServiceName,
+		})
+
+		// Replace all '.'s with "-"s to convert to a DNS-1035 label
+		serviceName := (strings.Replace(apiServiceName, ".", "-", -1))
+		service, err := a.OpClient.GetService(csv.GetNamespace(), serviceName)
```
ecordell: should we check ownerreferences here?

ecordell: also shouldn't this be querying the lister cache instead of the API?

pkg/controller/operators/olm/apiservices.go
```diff
@@ -35,6 +44,96 @@ func (a *Operator) syncAPIServices(obj interface{}) (syncError error) {
...
+			return err
+		}
+		if !certs.Active(ca) {
+			logger.Warnf("CA cert not active")
+			return fmt.Errorf("CA cert not active")
+		}
+
+		// Check if serving cert is active
+		secretName := apiServiceName + "-cert"
+		secret, err := a.OpClient.GetSecret(csv.GetNamespace(), secretName)
```
ecordell: cache?

pkg/controller/operators/olm/apiservices.go
```diff
@@ -138,8 +247,8 @@ func (a *Operator) installAPIServiceRequirements(desc v1alpha1.APIServiceDescrip
 			Selector: depSpec.Selector.MatchLabels,
 		},
 	}
-	// TODO(Nick): ensure service name is a valid DNS name
-	// replace all '.'s with "-"s to convert to a DNS-1035 label
+
+	// Replace all '.'s with "-"s to convert to a DNS-1035 label
 	service.SetName(strings.Replace(apiServiceName, ".", "-", -1))
```
ecordell: since we use this in multiple places now, we should probably have a helper `apiServiceNameToServiceName`

pkg/controller/operators/olm/apiservices.go
```diff
@@ -184,17 +293,26 @@ func (a *Operator) installAPIServiceRequirements(desc v1alpha1.APIServiceDescrip
 	}
 	secret.SetName(apiServiceName + "-cert")
 	secret.SetNamespace(csv.GetNamespace())
+
+	// Add olmcasha hash as a label to the
+	caPEM, _, err := ca.ToPEM()
```
ecordell: just to verify: can we ever rotate the cert (ourselves) without updating the hash label? Just want to make sure we can't get into a failed state because of actions we made ourselves.

pkg/controller/operators/olm/operator.go
```diff
@@ -373,9 +433,22 @@ func (a *Operator) transitionCSVState(in v1alpha1.ClusterServiceVersion) (out *v
...
+			return
+		}
+
 		if installErr := a.updateInstallStatus(out, installer, strategy, v1alpha1.CSVReasonComponentUnhealthy); installErr != nil {
 			logger.WithField("strategy", out.Spec.InstallStrategy.StrategyName).Infof("unhealthy component: %s", installErr)
 		}
+
+		// Check if it's time to refresh owned APIService certs
+		if a.shouldRotateCerts(out) {
+			out.SetPhase(v1alpha1.CSVPhasePending, v1alpha1.CSVReasonInstallSuccessful, "owned APIServices need cert refresh")
```
ecordell: should the reason be `Successful`?

test/e2e/csv_e2e_test.go
```diff
@@ -6,6 +6,8 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/ownerutil"
```
ecordell: nit: goimports

pkg/controller/operators/olm/apiservices.go
njhale: Yes - but I was going to put this off until [ALM-679](https://jira.coreos.com/browse/ALM-679)

Issue #488, opened 2018-09-27
make run-local-shift fails on minishift due to the lack of helm


tiraboschi:
minishift by default doesn't ship helm (it could be installed via minishift-addons) so  make run-local-shift fails with ./scripts/package-release.sh: line 26: helm: command not found

```
[stirabos@t470s operator-lifecycle-manager]$ make run-local-shift 
. ./scripts/build_local_shift.sh
-- Starting profile 'minishift'
The 'minishift' VM is already running.
Logged into "https://192.168.42.201:8443" as "system:admin" using existing credentials.

You have access to the following projects and can switch between them with 'oc project <projectname>':

    default
    kube-dns
    kube-proxy
    kube-public
    kube-system
  * myproject
    openshift
    openshift-apiserver
    openshift-controller-manager
    openshift-core-operators
    openshift-infra
    openshift-node
    openshift-web-console

Using project "myproject".
Sending build context to Docker daemon 53.41 MB
Step 1/6 : FROM golang:1.10
Trying to pull repository docker.io/library/golang ... 
1.10: Pulling from docker.io/library/golang
05d1a5232b46: Pull complete 
5cee356eda6b: Pull complete 
89d3385f0fd3: Pull complete 
80ae6b477848: Pull complete 
94ebfeaaddf3: Pull complete 
e132030a369d: Pull complete 
c67c5750c788: Pull complete 
Digest: sha256:a325153f1f1e2edb76a76ad789aff172b89dd6178e8f74d39ef87c04d87d1961
Status: Downloaded newer image for docker.io/golang:1.10
 ---> a4afc24299ee
Step 2/6 : WORKDIR /go/src/github.com/operator-framework/operator-lifecycle-manager
 ---> 80c03d68299f
Removing intermediate container 7a9f43d79279
Step 3/6 : COPY . .
 ---> e9063da745a2
Removing intermediate container 2a58de286bf7
Step 4/6 : RUN make build && cp bin/olm /bin/olm && cp bin/catalog /bin/catalog
 ---> Running in 421ecb380f26

building bin/catalog
building bin/olm
building bin/package-server
building bin/validator
 ---> 139d17cad118
Removing intermediate container 421ecb380f26
Step 5/6 : COPY deploy/chart/catalog_resources /var/catalog_resources
 ---> c288d3e16ea7
Removing intermediate container dcb0a8633750
Step 6/6 : CMD /bin/olm
 ---> Running in 6e79ac9a47a2
[...]

PR #528, opened 2018-10-18
[Doc] Add note about helm binary requirement

aditya-konarde:
`make run-local-shift` fails if the helm binary isn't available

This would fix: https://github.com/operator-framework/operator-lifecycle-manager/issues/488


ecordell:
/approve
/lgtm

Thanks for making our docs better!

aditya-konarde:
Thanks for the review, @ecordell :)

PR #530, opened 2018-10-18
Emit Kubernetes Events for ClusterServiceVersion

alecmerdler:
### Description

Emit events whenever we update the status of a `ClusterServiceVersion`. Uses examples from `client-go` and [this Kubernetes blog post](https://kubernetes.io/blog/2018/01/reporting-errors-using-kubernetes-events/).

Addresses https://jira.coreos.com/browse/ALM-378


pkg/api/apis/operators/v1alpha1/clusterserviceversion.go
```diff
@@ -10,6 +12,11 @@ var obsoleteReasons = map[ConditionReason]struct{}{
 	CSVReasonBeingReplaced: {},
 }
 
+func (c *ClusterServiceVersion) SetPhaseWithEvent(phase ClusterServiceVersionPhase, reason ConditionReason, message string, recorder record.EventRecorder) {
```
alecmerdler: I'm okay with the bleeding of concerns here because we always want to emit an event when the `status` changes.

Gopkg.lock
```diff
@@ -3,54 +3,41 @@
 
 [[projects]]
   branch = "default"
-  digest = "1:c0164b6d52877372590085b45f53a160f47c24e8a2312bb815489223044f5914"
```
alecmerdler: Had to update `vendor/` to add `client-go/tools/record`.

pkg/api/apis/operators/v1alpha1/clusterserviceversion.go
ecordell: I agree in general, but I'm not sure this will have the desired behavior.

I assume we only want to emit an event when we have actually committed the status to the object by writing into the kube API. Otherwise we will see events even when there are errors and the phase didn't actually change.

Alternately we could change this event to be a "attempting to enter phase: X" event instead of an "entered phase X" event.

pkg/api/apis/operators/v1alpha1/clusterserviceversion.go
```diff
@@ -10,6 +12,11 @@ var obsoleteReasons = map[ConditionReason]struct{}{
...
+	recorder.Event(c, v1.EventTypeNormal, string(reason), message)
```
ecordell: Also - is this blocking? Does it make sense to just throw it into a goroutine?

pkg/api/apis/operators/v1alpha1/clusterserviceversion.go
alecmerdler: Looks like Kubernetes usually emits events before calling `UpdateStatus()` on a resource:
- [`DeploymentController`](https://sourcegraph.com/github.com/kubernetes/kubernetes@37cb88efd2dbdb5e7d7b63c491dc751e416d7b34/-/blob/pkg/controller/deployment/deployment_controller.go#L586:6)
- [`CronJob`](https://sourcegraph.com/github.com/kubernetes/kubernetes@37cb88efd2dbdb5e7d7b63c491dc751e416d7b34/-/blob/pkg/controller/cronjob/cronjob_controller.go#L207)
- [`Service`](https://sourcegraph.com/github.com/kubernetes/kubernetes/-/blob/pkg/controller/service/service_controller.go#L309:1)

PR #536, opened 2018-10-22
feat(csv): detect req and dep change in succeeded/failed phases

njhale:
### Description
Adds detection of changes to requirements and deployments to succeeded and failed CSV phase sync handler logic. This PR will allow CSVs to transition out of the previously terminal succeeded and failed phases under the right conditions (i.e. requirement or deployment status change). 

Addresses: [ALM-713](https://jira.coreos.com/browse/ALM-713)

### DIFF
--- a/Documentation/design/architecture.md
+++ b/Documentation/design/architecture.md
@@ -54,9 +54,12 @@ While the OLM Operator is often configured to watch all namespaces, it can also 
 ```
            +------------------------------------------------------+
            |                                                      |
-           v                                      +--> Succeeded -+
-None --> Pending --> InstallReady --> Installing -|
-                                                  +--> Failed
+           |                                      +--> Succeeded -+
+           v                                      |               |
+None --> Pending --> InstallReady --> Installing -|               |
+           ^                                       +--> Failed <--+   
+           |                                              |
+           +----------------------------------------------+
 \                                                                 /
  +---------------------------------------------------------------+
     |
@@ -64,16 +67,16 @@ None --> Pending --> InstallReady --> Installing -|
 Replacing --> Deleting
 ```
 
-| Phase      | Description                                                                                                            |
-|------------|------------------------------------------------------------------------------------------------------------------------|
-| None       | initial phase, once seen by the Operator, it is immediately transitioned to `Pending`                                  |
-| Pending    | requirements in the CSV are not met, once they are this phase transitions to `Installing`                              |
-| InstallReady | all requirements in the CSV are present, the Operator will begin executing the install strategy                      |
-| Installing | the install strategy is being executed and resources are being created, but not all components are reporting as ready  |
-| Succeeded  | the execution of the Install Strategy was successful; if requirements disappear, this may transition back to `Pending` |
-| Failed     | upon failed execution of the Install Strategy, the CSV transitions to this terminal phase                              |
-| Replacing | a newer CSV that replaces this one has been discovered in the cluster. This status means the CSV is marked for GC       | 
-| Deleting | the GC loop has determined this CSV is safe to delete from the cluster. It will disappear soon.                          |
+| Phase      | Description                                                                                                                                                                                                                           |
+|------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| None       | initial phase, once seen by the Operator, it is immediately transitioned to `Pending`                                                                                                                                                 |
+| Pending    | requirements in the CSV are not met, once they are this phase transitions to `Installing`                                                                                                                                             |
+| InstallReady | all requirements in the CSV are present, the Operator will begin executing the install strategy                                                                                                                                     |
+| Installing | the install strategy is being executed and resources are being created, but not all components are reporting as ready                                                                                                                 |
+| Succeeded  | the execution of the Install Strategy was successful; if requirements disappear, or an APIService cert needs to be rotated this may transition back to `Pending`; if an installed component dissapears this may transition to `Failed`|
+| Failed     | upon failed execution of the Install Strategy, or an installed component dissapears the CSV transitions to this phase; if the component can be recreated by OLM, this may transition to `Pending`                                     |
+| Replacing  | a newer CSV that replaces this one has been discovered in the cluster. This status means the CSV is marked for GC                                                                                                                     | 
+| Deleting   | the GC loop has determined this CSV is safe to delete from the cluster. It will disappear soon.                                                                                                                                       |
 
 ### Namespace Control Loop
 
--- a/pkg/api/apis/operators/v1alpha1/clusterserviceversion_types.go
+++ b/pkg/api/apis/operators/v1alpha1/clusterserviceversion_types.go
@@ -182,21 +182,24 @@ const (
 type ConditionReason string
 
 const (
-	CSVReasonRequirementsUnknown     ConditionReason = "RequirementsUnknown"
-	CSVReasonRequirementsNotMet      ConditionReason = "RequirementsNotMet"
-	CSVReasonRequirementsMet         ConditionReason = "AllRequirementsMet"
-	CSVReasonOwnerConflict           ConditionReason = "OwnerConflict"
-	CSVReasonComponentFailed         ConditionReason = "InstallComponentFailed"
-	CSVReasonInvalidStrategy         ConditionReason = "InvalidInstallStrategy"
-	CSVReasonWaiting                 ConditionReason = "InstallWaiting"
-	CSVReasonInstallSuccessful       ConditionReason = "InstallSucceeded"
-	CSVReasonInstallCheckFailed      ConditionReason = "InstallCheckFailed"
-	CSVReasonComponentUnhealthy      ConditionReason = "ComponentUnhealthy"
-	CSVReasonBeingReplaced           ConditionReason = "BeingReplaced"
-	CSVReasonReplaced                ConditionReason = "Replaced"
-	CSVReasonNeedCertRotation        ConditionReason = "NeedCertRotation"
-	CSVReasonAPIServiceResourceIssue ConditionReason = "APIServiceResourceIssue"
-	CSVReasonCopied                  ConditionReason = "Copied"
+	CSVReasonRequirementsUnknown              ConditionReason = "RequirementsUnknown"
+	CSVReasonRequirementsNotMet               ConditionReason = "RequirementsNotMet"
+	CSVReasonRequirementsMet                  ConditionReason = "AllRequirementsMet"
+	CSVReasonOwnerConflict                    ConditionReason = "OwnerConflict"
+	CSVReasonComponentFailed                  ConditionReason = "InstallComponentFailed"
+	CSVReasonInvalidStrategy                  ConditionReason = "InvalidInstallStrategy"
+	CSVReasonWaiting                          ConditionReason = "InstallWaiting"
+	CSVReasonInstallSuccessful                ConditionReason = "InstallSucceeded"
+	CSVReasonInstallCheckFailed               ConditionReason = "InstallCheckFailed"
+	CSVReasonComponentUnhealthy               ConditionReason = "ComponentUnhealthy"
+	CSVReasonBeingReplaced                    ConditionReason = "BeingReplaced"
+	CSVReasonReplaced                         ConditionReason = "Replaced"
+	CSVReasonNeedsReinstall                   ConditionReason = "NeedsReinstall"
+	CSVReasonNeedsCertRotation                ConditionReason = "NeedsCertRotation"
+	CSVReasonAPIServiceResourceIssue          ConditionReason = "APIServiceResourceIssue"
+	CSVReasonAPIServiceResourcesNeedReinstall ConditionReason = "APIServiceResourcesNeedReinstall"
+	CSVReasonAPIServiceInstallFailed          ConditionReason = "APIServiceInstallFailed"
+	CSVReasonCopied                           ConditionReason = "Copied"
 )
 
 // Conditions appear in the status as a record of state transitions on the ClusterServiceVersion
@@ -220,8 +223,20 @@ type ClusterServiceVersionCondition struct {
 
 // OwnsCRD determines whether the current CSV owns a paritcular CRD.
 func (csv ClusterServiceVersion) OwnsCRD(name string) bool {
-	for _, crdDescription := range csv.Spec.CustomResourceDefinitions.Owned {
-		if crdDescription.Name == name {
+	for _, desc := range csv.Spec.CustomResourceDefinitions.Owned {
+		if desc.Name == name {
+			return true
+		}
+	}
+
+	return false
+}
+
+// OwnsAPIService determines whether the current CSV owns a paritcular APIService.
+func (csv ClusterServiceVersion) OwnsAPIService(name string) bool {
+	for _, desc := range csv.Spec.APIServiceDefinitions.Owned {
+		apiServiceName := fmt.Sprintf("%s.%s", desc.Version, desc.Group)
+		if apiServiceName == name {
 			return true
 		}
 	}
--- a/pkg/controller/certs/certs.go
+++ b/pkg/controller/certs/certs.go
@@ -15,10 +15,6 @@ import (
 	"time"
 )
 
-const (
-	Organization = "Red Hat, Inc."
-)
-
 // KeyPair stores an x509 certificate and its ECDSA private key
 type KeyPair struct {
 	Cert *x509.Certificate
@@ -49,7 +45,7 @@ func (kp *KeyPair) ToPEM() (certPEM []byte, privPEM []byte, err error) {
 }
 
 // GenerateCA generates a self-signed CA cert/key pair that expires in expiresIn days
-func GenerateCA(notAfter time.Time) (*KeyPair, error) {
+func GenerateCA(notAfter time.Time, organization string) (*KeyPair, error) {
 	notBefore := time.Now()
 	if notAfter.Before(notBefore) {
 		return nil, fmt.Errorf("invalid notAfter: %s before %s", notAfter.String(), notBefore.String())
@@ -63,7 +59,7 @@ func GenerateCA(notAfter time.Time) (*KeyPair, error) {
 	caDetails := &x509.Certificate{
 		SerialNumber: serial,
 		Subject: pkix.Name{
-			Organization: []string{Organization},
+			Organization: []string{organization},
 		},
 		NotBefore:             notBefore,
 		NotAfter:              notAfter,
@@ -98,7 +94,7 @@ func GenerateCA(notAfter time.Time) (*KeyPair, error) {
 }
 
 // CreateSignedServingPair creates a serving cert/key pair signed by the given ca
-func CreateSignedServingPair(notAfter time.Time, ca *KeyPair, hosts []string) (*KeyPair, error) {
+func CreateSignedServingPair(notAfter time.Time, organization string, ca *KeyPair, hosts []string) (*KeyPair, error) {
 	notBefore := time.Now()
 	if notAfter.Before(notBefore) {
 		return nil, fmt.Errorf("invalid notAfter: %s before %s", notAfter.String(), notBefore.String())
@@ -112,7 +108,7 @@ func CreateSignedServingPair(notAfter time.Time, ca *KeyPair, hosts []string) (*
 	certDetails := &x509.Certificate{
 		SerialNumber: serial,
 		Subject: pkix.Name{
-			Organization: []string{Organization},
+			Organization: []string{organization},
 		},
 		NotBefore:             notBefore,
 		NotAfter:              notAfter,
@@ -185,8 +181,10 @@ func Active(cert *x509.Certificate) bool {
 	return active
 }
 
+// PEMHash returns a hash of the given PEM encoded cert
 type PEMHash func(certPEM []byte) (hash string)
 
+// PEMSHA256 returns the hex encoded SHA 256 hash of the given PEM encoded cert
 func PEMSHA256(certPEM []byte) (hash string) {
 	hasher := sha256.New()
 	hasher.Write(certPEM)
--- a/pkg/controller/operators/olm/apiservices.go
+++ b/pkg/controller/operators/olm/apiservices.go
@@ -26,22 +26,9 @@ const (
 	DefaultCertValidFor = time.Hour * 24 * 730
 	// OLMCAHashAnnotationKey is the label key used to store the hash of the CA cert
 	OLMCAHashAnnotationKey = "olmcahash"
+	// Organization is the organization name used in the generation of x509 certs
+	Organization = "Red Hat, Inc."
 )
-
-func (a *Operator) syncAPIServices(obj interface{}) (syncError error) {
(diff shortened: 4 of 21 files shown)


ecordell:
This looks great, but I think we'll have a better time if we merge your `lister` work first; I'd rather not have O(csvsInCluster) api calls issued every resyncPeriod.

njhale:
> This looks great, but I think we'll have a better time if we merge your `lister` work first; I'd rather not have O(csvsInCluster) api calls issued every resyncPeriod.

I concur

jpeeler:
At a high level are the finalization additions to allow you to ensure deletion is happening? I'm wondering if being able to have a specific delete function handler might help: https://github.com/operator-framework/operator-lifecycle-manager/pull/480/files#diff-c09af0ae4ca3b7464670dbd80be66c70R187. There are probably some nuances I don't understand yet, more of a learning comment than proper review.

njhale:
> At a high level are the finalization additions to allow you to ensure deletion is happening? I'm wondering if being able to have a specific delete function handler might help: https://github.com/operator-framework/operator-lifecycle-manager/pull/480/files#diff-c09af0ae4ca3b7464670dbd80be66c70R187. There are probably some nuances I don't understand yet, more of a learning comment than proper review.

It definitely does help to have that - The only reason for using finalizers is because I wasn't able to circumvent the workqueue and indexer lookup without custom delete handlers. I'm in the process of removing this and getting ready to use the changes you linked! 🙂

cmd/olm/main.go
```diff
@@ -8,16 +8,15 @@ import (
...
-	log "github.com/sirupsen/logrus"
-
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/client"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/install"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/operators/olm"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/operatorclient"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/signals"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/metrics"
 	olmversion "github.com/operator-framework/operator-lifecycle-manager/pkg/version"
+	"github.com/prometheus/client_golang/prometheus"
```
alecmerdler: Fix import order.

pkg/api/apis/operators/v1alpha1/clusterserviceversion_types.go
```diff
@@ -182,20 +182,23 @@ const (
...
+	CSVReasonWaiting                          ConditionReason = "InstallWaiting"
+	CSVReasonInstallSuccessful                ConditionReason = "InstallSucceeded"
+	CSVReasonInstallCheckFailed               ConditionReason = "InstallCheckFailed"
+	CSVReasonComponentUnhealthy               ConditionReason = "ComponentUnhealthy"
+	CSVReasonBeingReplaced                    ConditionReason = "BeingReplaced"
+	CSVReasonReplaced                         ConditionReason = "Replaced"
+	CSVReasonNeedsReinstall                   ConditionReason = "NeedsReinstall"
+	CSVReasonNeedsCertRotation                ConditionReason = "NeedsCertRotation"
+	CSVReasonAPIServiceResourceIssue          ConditionReason = "APIServiceResourceIssue"
+	CSVReasonAPIServiceResourcesNeedReinstall ConditionReason = "APIServiceResourcesNeedReinstall"
```
alecmerdler: Are the [lifecycle docs](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/Documentation/design/architecture.md#clusterserviceversion-control-loop) still correct once we add this change?

pkg/controller/install/deployment.go
```diff
@@ -3,14 +3,13 @@ package install
 import (
 	"fmt"
 
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/wrappers"
```
alecmerdler: Fix import order.

pkg/controller/install/deployment_test.go
```diff
@@ -5,16 +5,15 @@ import (
 	"fmt"
 	"testing"
 
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
```
alecmerdler: Fix import order.

pkg/controller/operators/olm/apiservices.go
```diff
@@ -263,9 +321,9 @@ func (a *Operator) installAPIServiceRequirements(desc v1alpha1.APIServiceDescrip
 	existingService, err := a.lister.CoreV1().ServiceLister().Services(csv.GetNamespace()).Get(service.GetName())
 	if err == nil {
 		// Check if the only owners are this CSV or in this CSV's replacement chain
-		if !ownerutil.Adoptable(csv, existingService.GetOwnerReferences()) {
-			return nil, fmt.Errorf("service %s not safe to replace: extraneous ownerreferences found", service.GetName())
-		}
+		// if ownerutil.Adoptable(csv, existingService.GetOwnerReferences()) {
```
alecmerdler: Should this be commented out?

pkg/controller/operators/olm/apiservices.go
```diff
@@ -156,16 +210,20 @@ func (a *Operator) areAPIServicesAvailable(descs []v1alpha1.APIServiceDescriptio
...
 		}
 
 		if !a.isAPIServiceAvailable(apiService) {
-			return false, fmt.Errorf("APIService %s not available", apiService.GetName())
+			return false, nil //fmt.Errorf("APIService %s not available", apiService.GetName())
 		}
 
 		if err := a.isGVKRegistered(desc.Group, desc.Version, desc.Kind); err != nil {
-			return false, fmt.Errorf("group: %s, version: %s, kind: %s not registered", desc.Group, desc.Version, desc.Kind)
+			return false, nil //fmt.Errorf("group: %s, version: %s, kind: %s not registered", desc.Group, desc.Version, desc.Kind)
```
alecmerdler: Should this be commented out?

pkg/controller/operators/olm/apiservices.go
```diff
@@ -156,16 +210,20 @@ func (a *Operator) areAPIServicesAvailable(descs []v1alpha1.APIServiceDescriptio
...
+			return false, nil
+		}
+
 		if err != nil {
 			return false, err
 		}
 
 		if !a.isAPIServiceAvailable(apiService) {
-			return false, fmt.Errorf("APIService %s not available", apiService.GetName())
+			return false, nil //fmt.Errorf("APIService %s not available", apiService.GetName())
```
alecmerdler: Should this be commented out?

pkg/controller/operators/olm/operator.go
```diff
@@ -212,16 +243,16 @@ func NewOperator(crClient versioned.Interface, opClient operatorclient.ClientInt
...
 
+// Cleanup cleans up operator resources on the cluster
 func (a *Operator) Cleanup() {
 	a.cleanupFunc()
 }
 
 func (a *Operator) requeueCSV(name, namespace string) {
 	// we can build the key directly, will need to change if queue uses different key scheme
 	key := fmt.Sprintf("%s/%s", namespace, name)
+	log.Debugf("requeueing CSV %s", key)
```
alecmerdler: Nice!

pkg/controller/operators/olm/operator.go
```diff
@@ -230,91 +261,76 @@ func (a *Operator) syncDeployment(obj interface{}) (syncError error) {
...
-		a.requeueCSV(oref.Name, secret.GetNamespace())
+	switch v := obj.(type) {
+	case *corev1.Secret:
+		logger.Debugf("syncing Secret %s in namespace %s", v.GetName(), v.GetNamespace())
+	case *corev1.Service:
+		logger.Debugf("syncing Service %s in namespace %s", v.GetName(), v.GetNamespace())
+	case *corev1.ServiceAccount:
+		logger.Debugf("syncing ServiceAccount %s in namespace %s", v.GetName(), v.GetNamespace())
+	default:
+		syncError = errors.New("attempted to sync non Core resource with Core sync handler")
```
alecmerdler: Should we include the `GroupVersionKind` when logging this? Also, saying "non Core" after we only checked if it's one of three different resources is disingenuous.

pkg/controller/operators/olm/apiservices.go
```diff
@@ -128,16 +129,69 @@ func (a *Operator) checkAPIServiceResources(csv *v1alpha1.ClusterServiceVersion,
...
+
+		// system:auth-delegator
+		authDelegatorClusterRole, err := a.lister.RbacV1().ClusterRoleLister().Get("system:auth-delegator")
+		if err != nil {
+			logger.Warnf("could not retrieve ClusterRole system:auth-delegator")
+			return err
+		}
+		rulesMap[metav1.NamespaceAll] = append(rulesMap[metav1.NamespaceAll], authDelegatorClusterRole.Rules...)
+
+		for namespace, rules := range rulesMap {
```
njhale: Is checking if the required rules are satisfied better than just checking if the generated Roles/Bindings exist by name? I'm wondering if we can also get into an infinite CSV transition (Installing -> Failed -> Pending -> Installing -> Failed) because OLM doesn't have permissions to create these resources. I think the answer is no, but would like a second pair of eyes to validate that.

pkg/controller/operators/olm/requirements_test.go
```diff
@@ -5,16 +5,15 @@ import (
 	"fmt"
 	"testing"
 
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
```
alecmerdler: Fix import order.

pkg/lib/operatorlister/clusterserviceversion.go
```diff
@@ -0,0 +1,95 @@
+package operatorlister
+
+import (
+	"fmt"
+	"sync"
+
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
```
alecmerdler: Fix import order.

pkg/lib/operatorlister/clusterserviceversion.go
```diff
@@ -0,0 +1,95 @@
...
+	"k8s.io/apimachinery/pkg/types"
+)
+
+type UnionClusterServiceVersionLister struct {
+	csvListers map[string]listers.ClusterServiceVersionLister
+	csvLock    sync.RWMutex
+}
+
+// List lists all ClusterServiceVersions in the indexer.
+func (ucl *UnionClusterServiceVersionLister) List(selector labels.Selector) (ret []*v1alpha1.ClusterServiceVersion, err error) {
```
alecmerdler: Is this implementing some required interface? It seems like these could just be pure functions...

pkg/controller/operators/olm/apiservices.go
```diff
@@ -263,9 +321,9 @@ func (a *Operator) installAPIServiceRequirements(desc v1alpha1.APIServiceDescrip
...
 		// Check if the only owners are this CSV or in this CSV's replacement chain
-		if !ownerutil.Adoptable(csv, existingService.GetOwnerReferences()) {
-			return nil, fmt.Errorf("service %s not safe to replace: extraneous ownerreferences found", service.GetName())
-		}
+		// if ownerutil.Adoptable(csv, existingService.GetOwnerReferences()) {
+		// 	return nil, fmt.Errorf("service %s not safe to replace: extraneous ownerreferences found", service.GetName())
+		// }
 
 		// Append previous ownerrefs to the new Service
 		service.SetOwnerReferences(append(service.GetOwnerReferences(), existingService.GetOwnerReferences()...))
```
njhale: Is stealing a service from an existing extension api-server deployment bad?. If it's managed by another CSV it will be recreated anyway.

pkg/controller/operators/olm/operator.go
njhale: That makes sense. Honestly, the separate sync functions that don't perform actions are just for the sake of logging.  As long as we can assert it's of type `metav1.Object`, we can now requeue any owner CSVs.

pkg/controller/operators/olm/operator.go
```diff
@@ -667,3 +730,56 @@ func (a *Operator) isReplacing(in *v1alpha1.ClusterServiceVersion) *v1alpha1.Clu
...
+}
+
+func (a *Operator) requeueOwnerCSVs(ownee metav1.Object) {
+	logger := log.WithFields(log.Fields{
+		"ownee":    ownee.GetName(),
+		"selflink": ownee.GetSelfLink(),
+	})
+
+	// Get all existing CSVs from the indexer
+	csvs, err := a.lister.OperatorsV1alpha1Lister().ClusterServiceVersionLister().List(labels.Everything())
```
njhale: I think going by UID and looking through all CSVs on the cluster is the best way to handle requeueing in the general case (Since cluster level resources can have ownerreferences to CSVs in multiple namespaces). Using the indexer means we don't need to query the Kube API, and if by chance a CSV doesn't exist in the indexer cache yet, it will be by at most the time of the next resync.

pkg/controller/operators/olm/operator.go
jpeeler: Latest code looks much better!

pkg/lib/operatorlister/clusterserviceversion.go
njhale: This is implementing `v1alpha1.ClusterServiceVersionLister`.

pkg/lib/queueinformer/queueinformer.go
```diff
@@ -105,10 +111,26 @@ func NewInformer(queue workqueue.RateLimitingInterface, informer cache.SharedInd
 		name:            name,
 		MetricsProvider: metrics,
 	}
-	if funcs == nil {
-		queueInformer.resourceEventHandlerFuncs = queueInformer.defaultResourceEventHandlerFuncs()
-	} else {
-		queueInformer.resourceEventHandlerFuncs = funcs
+	queueInformer.resourceEventHandlerFuncs = queueInformer.defaultResourceEventHandlerFuncs()
+	if funcs != nil {
+		if funcs.AddFunc != nil {
```
ecordell: Do we always want to run the default logic? I would think we'd want to overwrite it if you pass one in.

test/e2e/csv_e2e_test.go
```diff
@@ -545,9 +546,10 @@ func TestCreateCSVRequirementsMetCRD(t *testing.T) {
 			},
 			Scope: "Namespaced",
 		},
-	})
+	}
+	cleanupCRD, err := createCRD(c, crd)
 	require.NoError(t, err)
-	defer cleanupCRD()
+	// defer cleanupCRD()
```
ecordell: nit: delete line

pkg/controller/operators/olm/operator.go
ecordell: I think I'm missing why we would bother listing and checking for existence here? requeueing is fast (adds a key to the queue), and we'll detect that it's not there anymore when we pop it. The first thing queue processing does is check if the object exists in the cache.

pkg/controller/operators/olm/operator.go
```diff
@@ -667,3 +704,56 @@ func (a *Operator) isReplacing(in *v1alpha1.ClusterServiceVersion) *v1alpha1.Clu
 	}
 	return previous
 }
+
+func (a *Operator) handleDeletion(obj interface{}) {
```
ecordell: 👍

pkg/controller/operators/olm/apiservices.go
```diff
@@ -262,11 +305,6 @@ func (a *Operator) installAPIServiceRequirements(desc v1alpha1.APIServiceDescrip
 
 	existingService, err := a.lister.CoreV1().ServiceLister().Services(csv.GetNamespace()).Get(service.GetName())
 	if err == nil {
-		// Check if the only owners are this CSV or in this CSV's replacement chain
-		if !ownerutil.Adoptable(csv, existingService.GetOwnerReferences()) {
```
ecordell: why don't we need this anymore?

pkg/controller/operators/olm/apiservices.go
ecordell: Definitely better to check the rules when we can. 

You bring up a good point, we ideally only want to transition back to pending _if_ we have reason to believe that the final result could change (e.g. previously rules were not met, but now they are, or vice-versa).

I think we probably want to address this by recording on the status each succeess/failure check, as a "real" Condition (as opposed to our Conditions which are just a history of transitions). Then there could be a "RulesMet" condition, a "ComponentHealth" condition (or many of them), etc. When we're in success/failure, we want to detect if those things have changed in a way that would move the needle, and then flip back to pending.

I'm not sure we need to do this here (we do have a story for Conditions) unless you think that's the only way to reasonably avoid the loops you're talking about. I think it should be enough to carry this discussion into the design docs for Conditions and work on it later.

pkg/controller/certs/certs.go
```diff
@@ -100,9 +124,6 @@ func GenerateCA(notAfter time.Time) (*KeyPair, error) {
 // CreateSignedServingPair creates a serving cert/key pair signed by the given ca
 func CreateSignedServingPair(notAfter time.Time, ca *KeyPair, hosts []string) (*KeyPair, error) {
 	notBefore := time.Now()
-	if notAfter.Before(notBefore) {
```
ecordell: why did we stop checking this?

pkg/controller/certs/certs.go
```diff
@@ -48,12 +50,34 @@ func (kp *KeyPair) ToPEM() (certPEM []byte, privPEM []byte, err error) {
 	return
 }
 
+// ToTLSSecret converts the KeyPair to a Kubernetes TLS Secret
+func (kp *KeyPair) ToTLSSecret(name, namespace string) (*corev1.Secret, error) {
```
ecordell: I can't see where these are used?

Documentation/design/architecture.md
```diff
@@ -54,26 +54,29 @@ While the OLM Operator is often configured to watch all namespaces, it can also
 ```
            +------------------------------------------------------+
            |                                                      |
-           v                                      +--> Succeeded -+
-None --> Pending --> InstallReady --> Installing -|
-                                                  +--> Failed
+           |                                      +--> Succeeded -+
```
ecordell: 👍

pkg/lib/queueinformer/queueinformer.go
njhale: I agree.

pkg/controller/certs/certs.go
njhale: I wanted to be able to create bad serving cert pairs for testing purposes. I thought this would be a reasonable approach since we were thinking about converting to the serving cert signing service in the future anyway.

pkg/controller/operators/olm/apiservices.go
njhale: This must have been removed while I was testing. Will put back soon.

pkg/controller/operators/olm/operator.go
njhale: We are listing so we can requeue owner CSVs by the UID listed in the object's OwnerReferences. We need to do it this way because objects like ClusterRoles and ClusterRoleBindings have no context to figure out which namespace their owner CSV hails from. Requeueing is quick, but why not avoid requeueing all CSVs on the cluster every time? The "existence" check is really just a byproduct of creating a set to efficiently find which CSVs to requeue anyway.

pkg/api/apis/operators/v1alpha1/clusterserviceversion_types.go
njhale: I've updated them.

pkg/controller/certs/certs.go
njhale: Testing. I can move these to the appropriate test files.

pkg/controller/operators/olm/apiservices.go
```diff
@@ -156,16 +231,20 @@ func (a *Operator) areAPIServicesAvailable(descs []v1alpha1.APIServiceDescriptio
...
+			return false, nil
+		}
+
 		if err != nil {
 			return false, err
 		}
 
 		if !a.isAPIServiceAvailable(apiService) {
-			return false, fmt.Errorf("APIService %s not available", apiService.GetName())
+			return false, nil //fmt.Errorf("APIService %s not available", apiService.GetName())
```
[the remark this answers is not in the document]
njhale: This was to distinguish between a terminal error state and "not available yet".

pkg/controller/certs/certs.go
ecordell: Make a function in a test file that generates bad certs? let's err on the side of more sanity checks for prod code

pkg/controller/operators/olm/operator.go
ecordell: In that case, I think this should only fall back to using the lister, e.g.:

```
 if owner ref has name/namespace:
    requeue(name, namespace)
 if owner ref missing namespace:
    list csvs
    find namespace
    requeue(name, namespace)
```

pkg/controller/operators/olm/apiservices.go
ecordell: this is still commented out?

pkg/controller/certs/certs.go
njhale: got it

pkg/controller/operators/olm/apiservices.go
njhale: shouldn't be. I amended the commit.

pkg/controller/operators/olm/operator.go
```diff
@@ -20,40 +33,34 @@ import (
...
-	csvLister           map[string]csvlister.ClusterServiceVersionLister
-	operatorGroupLister map[string]operatorgrouplister.OperatorGroupLister
-	annotator           *annotator.Annotator
-	recorder            record.EventRecorder
-	cleanupFunc         func()
+	csvQueue workqueue.RateLimitingInterface
+	client   versioned.Interface
+	resolver install.StrategyResolverInterface
+	lister   operatorlister.OperatorLister
+	// csvLister           map[string]csvlister.ClusterServiceVersionLister
```
ecordell: if these aren't needed anymore, let's remove them?

pkg/controller/operators/olm/operator.go
```diff
@@ -616,14 +659,17 @@ func (a *Operator) findIntermediatesForDeletion(csv *v1alpha1.ClusterServiceVers
 	return nil
 }
 
-// csvsInNamespace finds all CSVs in a namespace
-func (a *Operator) csvsInNamespace(namespace string) map[string]*v1alpha1.ClusterServiceVersion {
-	csvsInNamespace, err := a.client.OperatorsV1alpha1().ClusterServiceVersions(namespace).List(metav1.ListOptions{})
+// csvSet gathers all CSVs in the given namespace into a map keyed by CSV name; if metav1.NamespaceAll gets the set across all namespaces
+func (a *Operator) csvSet(namespace string) map[string]*v1alpha1.ClusterServiceVersion {
```
ecordell: Awesome, thanks

PR #537, opened 2018-10-25
Fix template values for 'olm' parameters

rhuss:
as it was still referring to 'alm' instead of 'olm', which breaks `make run-local-shift`


ecordell:
/approve
/lgtm

Thanks for fixing this! The minishift scripts became out of date (due to the lag in 3.11 availability in minishift)

jpeeler:
/retest
hit network quote limits...

PR #541, opened 2018-10-26
Verify Native APIs Present for ClusterServiceVersion

alecmerdler:
### Description

Adds the ability to enforce that certain "native" APIs exist in a Kubernetes cluster in order to install an Operator using a `ClusterServiceVersion`. This is a more complete solution than depending on some "version" of Kubernetes, because there are so many flavors and distributions that may or may not contain the APIs consumed by the Operator.

### Changes

- Adds `spec.nativeAPIs` field to `ClusterServiceVersion` (array of `GroupVersionKinds`)
- Verifies required native APIs are present using basic discovery as part of CSV control loop
- Writes out to `requirementStatus` just like CRDs and apiservices

Addresses https://jira.coreos.com/browse/ALM-666

### DIFF
--- a/deploy/chart/templates/0000_30_02-clusterserviceversion.crd.yaml
+++ b/deploy/chart/templates/0000_30_02-clusterserviceversion.crd.yaml
@@ -180,7 +180,25 @@ spec:
                           values:
                             type: array
                             description: set of values for the expression
-                            
+            nativeAPIs:
+              type: array
+              description: What resources are required by the Operator, but must be provided by the underlying cluster and not as an extension.
+              items:
+                type: object
+                required:
+                - group
+                - version
+                - kind
+                properties:
+                  group:
+                    type: string
+                    description: Group of the API resource
+                  version:
+                    type: string
+                    description: Version of the API resource
+                  kind:
+                    type: string
+                    description: Kind of the API resource        
             apiservicedefinitions:
               type: object
               properties:
--- a/pkg/api/apis/operators/v1alpha1/clusterserviceversion_types.go
+++ b/pkg/api/apis/operators/v1alpha1/clusterserviceversion_types.go
@@ -102,14 +102,15 @@ type APIServiceDefinitions struct {
 	Required []APIServiceDescription `json:"required,omitempty"`
 }
 
-// ClusterServiceVersionSpec declarations tell the OLM how to install an operator
-// that can manage apps for given version and AppType.
+// ClusterServiceVersionSpec declarations tell OLM how to install an operator
+// that can manage apps for a given version.
 type ClusterServiceVersionSpec struct {
 	InstallStrategy           NamedInstallStrategy      `json:"install"`
 	Version                   semver.Version            `json:"version,omitempty"`
 	Maturity                  string                    `json:"maturity,omitempty"`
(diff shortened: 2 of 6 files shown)


pkg/api/apis/operators/v1alpha1/clusterserviceversion_types.go
```diff
@@ -102,14 +102,23 @@ type APIServiceDefinitions struct {
 	Required []APIServiceDescription `json:"required,omitempty"`
 }
 
-// ClusterServiceVersionSpec declarations tell the OLM how to install an operator
-// that can manage apps for given version and AppType.
+// NativeAPI describes the GVK of a Kubernetes resource that is required by an
+// Operator, but must be provided by the underlying cluster and not an extension.
+type NativeAPI struct {
```
alecmerdler: There might be an existing struct that can be used here that is JSON-serializable.

ecordell:
Looks great :) just need an e2e test and we can merge

deploy/chart/templates/0000_30_02-clusterserviceversion.crd.yaml
```diff
@@ -180,7 +180,25 @@ spec:
                           values:
                             type: array
                             description: set of values for the expression
-                            
+            nativeAPIs:
```
alecmerdler: We can't enforce this as a required field, but it should be.

pkg/controller/operators/olm/requirements.go
```diff
@@ -238,7 +260,7 @@ func (a *Operator) requirementAndPermissionStatus(csv *v1alpha1.ClusterServiceVe
 		return false, nil, fmt.Errorf("could not cast install strategy as type %T", strategyDetailsDeployment)
 	}
 
-	reqMet, reqStatuses := a.requirementStatus(strategyDetailsDeployment, csv.GetAllCRDDescriptions(), csv.GetOwnedAPIServiceDescriptions(), csv.GetRequiredAPIServiceDescriptions())
+	reqMet, reqStatuses := a.requirementStatus(strategyDetailsDeployment, csv.GetAllCRDDescriptions(), csv.GetOwnedAPIServiceDescriptions(), csv.GetRequiredAPIServiceDescriptions(), csv.Spec.NativeAPIs)
```
jpeeler: Does the GUI ever make use of the previous getter methods? I'm just wondering if it's worth making one for the Native APIs to return an alphabetical listing (or deduplication) like the others.

alecmerdler: I don't understand how the frontend app would call into the Operator Go source...?

alecmerdler:
@ecordell Looks like this passed CI!

pkg/api/apis/operators/v1alpha1/clusterserviceversion_types.go
jpeeler: Like https://godoc.org/k8s.io/apimachinery/pkg/apis/meta/v1#GroupVersionKind ?

pkg/controller/operators/olm/requirements.go
jpeeler: I didn't mean call directly. But based on your response there's zero coupling.

PR #548, opened 2018-11-02
Add test case for supporting multi-version CRD

dinhxuanvu:
Test updating CSV which owns different versions of the same CRD to
ensure CSV is created correctly and deployed successfully.

Jira: <https://jira.coreos.com/browse/ALM-629>


jpeeler:
oops :)

PR #551, opened 2018-11-05
Add CRD to "View" RBAC Role

alecmerdler:
### Description

When creating the "view" `Role` for instances of a custom resource, add the CRD itself so users can fetch the validation and additional printer columns.

Addresses https://jira.coreos.com/browse/ALM-785


ecordell:
This looks good, we should support apiservices as well! I'm about to push up a PR that adds the pluralname to apiservicedefinitions in the CSV so that we can construct the RBAC for it

alecmerdler:
@ecordell Is this blocked by #571?

ecordell:
This is included in #618

PR #552, opened 2018-11-07
feat(make): add e2e command for running from ci-operator

ecordell:


pkg/package-server/provider/inmem.go
```diff
@@ -188,7 +189,7 @@ func (m *InMemoryProvider) syncCatalogSource(obj interface{}) error {
 	// assert as CatalogSource
 	catsrc, ok := obj.(*operatorsv1alpha1.CatalogSource)
 	if !ok {
-		log.Debugf("wrong type: %#v", obj)
+		logrus.Debugf("wrong type: %#v", obj)
```
njhale: At some point I think this needs to be injected as a dependency.

scripts/install_bare.sh
```diff
@@ -0,0 +1,33 @@
...
+    then
+    	# skip olm and catalog operator deployment
+    	continue
+    else
+        kubectl apply -f ${f};
+    fi
+done
+
+# wait for package server to be ready
+#kubectl rollout status -w deployment.apps/package-server --namespace=${namespace}
```
njhale: nit: commented rollout

PR #553, opened 2018-11-07
Verify CRD's condition to ensure it's registered with k8s API

dinhxuanvu:
OLM needs to verify CRD's condition to make sure it is registered
with k8s API during CSV control loop before installing operator.
This verification will help to avoid errors related to the
unavailability of an API when installing an operator via OLM.

Jira: https://jira.coreos.com/browse/ALM-775


ecordell:
/approve

looks good! just need some tests for this.

test/e2e/csv_e2e_test.go
```diff
@@ -312,6 +313,85 @@ func TestCreateCSVWithUnmetPermissionsCRD(t *testing.T) {
...
+
+	c := newKubeClient(t)
+	crc := newCRClient(t)
+
+	depName := genName("dep-")
+
+	// Create dependency first (CRD)
+	crdPlural := genName("ins2")
+	crdName := crdPlural + ".cluster.com"
+	crd := extv1beta1.CustomResourceDefinition{
```
Miciah: I suggest adding an ampersand so you're passing a pointer to `UpdateStatus` below: `&extv1beta1.CustomResourceDefinition{`.  That should resolve the compiler error.

pkg/controller/operators/olm/requirements.go
```diff
@@ -59,6 +60,23 @@ func (a *Operator) requirementStatus(strategyDetailsDeployment *install.Strategy
 			met = false
 			statuses = append(statuses, status)
 		}
+
+		// Check if CRD has succes registered with k8s API
```
njhale: nit: successfully

test/e2e/csv_e2e_test.go
```diff
@@ -312,6 +312,88 @@ func TestCreateCSVWithUnmetPermissionsCRD(t *testing.T) {
 
 }
 
+func TestCreateCSVWithNotReadyCRD(t *testing.T) {
```
njhale: I feel like this would be easier to test with a unit test where you can manually set the status without worrying about fighting the CRD controller.

pkg/controller/operators/olm/requirements.go
```diff
@@ -39,24 +40,41 @@ func (a *Operator) requirementStatus(strategyDetailsDeployment *install.Strategy
...
 
-		served := false
-		for _, version := range crd.Spec.Versions {
-			if version.Name == r.Version {
-				if version.Served {
-					status.Status = v1alpha1.RequirementStatusReasonPresent
-					status.UUID = string(crd.GetUID())
-					statuses = append(statuses, status)
-					served = true
+		// Check if CRD has successfully registered with k8s API
```
jpeeler: I suggest moving this code block (and the !registered) to above so that only bad statuses are written. Otherwise, you end up getting a "Present" status and a "PresetNotAvailable", which seems slightly confusing to me.

pkg/controller/operators/olm/requirements_test.go
```diff
@@ -433,13 +433,13 @@ func TestRequirementAndPermissionStatus(t *testing.T) {
 				namespace,
 				"",
 				installStrategy("csv1-dep", nil, nil),
-				[]*v1beta1.CustomResourceDefinition{crd("c1", "v2")},
+				[]*v1beta1.CustomResourceDefinition{crd("c1", "v1")},
 				nil,
```
jpeeler: Is the inversion here and below intentional?

pkg/api/apis/operators/v1alpha1/clusterserviceversion_types.go
```diff
@@ -252,6 +252,7 @@ const (
 	RequirementStatusReasonPresent             StatusReason = "Present"
 	RequirementStatusReasonNotPresent          StatusReason = "NotPresent"
 	RequirementStatusReasonPresentNotSatisfied StatusReason = "PresentNotSatisfied"
+	RequirementStatusReasonNotAvailable        StatusReason = "PresentNotAvailable"
```
jpeeler: I think a comment here for what exactly not available means would be good - "set when when CRD status condition is missing established" or something.

ecordell:
@dinhxuanvu I think you need to rebase, but then this PR should be good to go

jpeeler:
Since the PR from my branch is all green, I'm closing this one.

PR #554, opened 2018-11-07
Switch to go modules

ecordell:
See: https://github.com/golang/go/wiki/Modules

Go modules will be default soon anyway, but this is required now for managing the circular dependency between OLM and operator-registry properly. Switching to modules makes it much easier to share tools as well, see the diffs for details.

The only changes required for developing OLM is that you shouldn't keep it in your GOPATH (if you were before). The rest is wrapped up in the makefile.

Additional changes:

- Codegen is now run in a container, which should make it easier to use and more consistent. This is because kube codegen doesn't know how to run outside of gopath yet.
- We manually pull down branches for kube projects, because they don't yet use go module appropriate tags, and go.mod doesn't support tracking a branch by name. `make vendor` takes care of downloading, pruning, and vendoring all of the dependencies together.
- Added back a `.dockerignore` so that we aren't sending a giant context to builds.
- Lots of small `Makefile` improvements, should be easier to read through now


Makefile
```diff
@@ -108,46 +106,37 @@ $(CI):
...
+# use container-codegen
+codegen:
+	cp scripts/generate_groups.sh vendor/k8s.io/code-generator/generate_groups.sh
+	mkdir -p vendor/k8s.io/code-generator/hack
+	cp boilerplate.go.txt vendor/k8s.io/code-generator/hack/boilerplate.go.txt
+	go run vendor/k8s.io/kube-openapi/cmd/openapi-gen/openapi-gen.go --logtostderr -i ./vendor/k8s.io/apimachinery/pkg/runtime,./vendor/k8s.io/apimachinery/pkg/apis/meta/v1,./vendor/k8s.io/apimachinery/pkg/version,./pkg/package-server/apis/packagemanifest/v1alpha1 -p $(PKG)/pkg/package-server/apis/openapi -O zz_generated.openapi -h boilerplate.go.txt -r /dev/null
 	$(CODEGEN) all $(PKG)/pkg/api/client $(PKG)/pkg/api/apis "operators:v1alpha1,v1alpha2"
 	$(CODEGEN) all $(PKG)/pkg/package-server/client $(PKG)/pkg/package-server/apis "packagemanifest:v1alpha1"
 
+container-codegen:
```
njhale: 👍

PR #555, opened 2018-11-07
Creating csv, crd, and packages for clusterlogging and elasticsearch operators

ewolinetz:
Some of the `statusDescriptors` depend on:

- openshift/elasticsearch-operator#26
- openshift/cluster-logging-operator#36

I believe this should replace https://github.com/operator-framework/operator-manifests/pull/7 but am not sure.

PTAL


ewolinetz:
want to update clo csv to pull in additional status from https://github.com/openshift/cluster-logging-operator/pull/25

Is there a way to conditionally pull information for a statusDescriptor? If we choose to deploy `rsyslog` instead of `fluentd` we would not see the path in the `status` field for clo

deploy/chart/catalog_resources/rh-operators/clusterloggings.v0.0.1.clusterserviceversion.yaml
```diff
@@ -0,0 +1,195 @@
+apiVersion: operators.coreos.com/v1alpha1
+kind: ClusterServiceVersion
+metadata:
+  name: clusterlogging.v0.0.1
```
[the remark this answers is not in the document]
ewolinetz: Do I need to update the package.yaml file then?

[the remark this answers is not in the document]
alecmerdler: Yes, it matches the `metadata.name` of the CSV, not the filename.

ewolinetz:
Made updates based on gist

deploy/chart/catalog_resources/rh-operators/clusterlogging.v0.0.1.clusterserviceversion.yaml
```diff
@@ -0,0 +1,263 @@
+apiVersion: operators.coreos.com/v1alpha1
+kind: ClusterServiceVersion
+metadata:
+  name: clusterlogging.v0.0.1
+  namespace: placeholder
+spec:
+  displayName: Cluster Logging Operator
+
+  description: |
+    The Cluster Logging Operator for OKD provides easy means for configuring and managing your aggregated logging stack.
```
alecmerdler: This Operator won't run on upstream Kubernetes?

jcantrill: @alecmerdler It has not in the past but does it matter in context of openshift and OLM?  We don't have a requirement to support an upstream kubernetes cluster

alecmerdler: OLM has a requirement to be good upstream citizens in order to increase community adoption and contribution. I'm not sure if we can include Operators that are only for OKD; most authors have different images/manifests that work for upstream.

jcantrill: Given how we support logging in Openshift making use of multitenancy, this is not directly translatable to use with upstream Kubernetes because of the authn/authz bits:
Quoting @enj here:
>Mo Khan [9:19 AM]
you have like zero chance of converting that stuff to kube
you can certainly have a kube-only flow
that does some limited subset
but no way to get the same behavior

>Mo Khan [9:19 AM]
unlike openshift, you have 0 guarantees about how authn and authz are configured

> Mo Khan [9:21 AM]
if you have a token, you can do the upstream token review.  how you actually get a token (nevermind scopes), is guess work
basically the part that does work upstream is, "if you somehow have a token for a user, and you are trusted entity allowed to see user's token, and you call token review, you can learn the user/token/extra/audience for that token"
so that only works for things you are happy seeing tokens, and if you just ignore the whole "how does a user get a token??"
basically, it only works for headless automated flows
and there is no way to make it "safe", because there is no concept of scopes
at least, no way to always make it safe, kube could be configured is whatever magical way to possibly make it safe
maybe
our apis are very aware of scopes

If this will block the merge of this PR then we need to re-evaluate how OKD cluster-logging is included as part of the platform

PR #556, opened 2018-11-08
Create registry pods for ConfigMap CatalogSources

ecordell:
This is step 2 in a long line of changes to switch to using operator-registry in OLM. 

- [x] Add pod reconciler that creates registry pods from configmaps
- [x] Add ConfigMap sourcetype as an alias for internal
- [x] Write tests

Note that this _doesn't_ yet use the pods for resolving dependencies, so the in memory registry is staying for now.


pkg/controller/operators/catalog/operator.go
```diff
@@ -7,12 +7,17 @@ import (
 	"sync"
 	"time"
 
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/operatorlister"
```
alecmerdler: Fix import order.

pkg/controller/registry/reconciler.go
```diff
@@ -0,0 +1,360 @@
...
+	rbacv1 "k8s.io/api/rbac/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+var timeNow = func() metav1.Time { return metav1.NewTime(time.Now().UTC()) }
+
+// catalogSourceDecorator wraps CatalogSource to add additional methods
+type catalogSourceDecorator struct {
```
alecmerdler: Why not just use pure functions instead of embedding a struct?

pkg/controller/registry/reconciler.go
```diff
@@ -0,0 +1,360 @@
...
+			APIGroup: "rbac.authorization.k8s.io",
+			Kind:     "Role",
+			Name:     s.roleName(),
+		},
+	}
+	ownerutil.AddOwner(rb, s.CatalogSource, false, false)
+	return rb
+}
+
+type RegistryReconciler interface {
```
alecmerdler: Can we potentially not introduce [another class that ends in -er](https://www.yegor256.com/2015/03/09/objects-end-with-er.html)?

pkg/controller/operators/catalog/operator.go
```diff
@@ -141,57 +151,211 @@ func NewOperator(kubeconfigPath string, wakeupInterval time.Duration, operatorNa
...
+		roleBindingInformer := informerFactory.Rbac().V1().RoleBindings()
+		serviceAccountInformer := informerFactory.Core().V1().ServiceAccounts()
+		serviceInformer := informerFactory.Core().V1().Services()
+		podInformer := informerFactory.Core().V1().Pods()
+		configMapInformer := informerFactory.Core().V1().ConfigMaps()
+
+		queueInformers := []*queueinformer.QueueInformer{
+			queueinformer.NewInformer(roleQueue, roleInformer.Informer(), op.syncObject, handleDelete, "role", metrics.NewMetricsNil()),
+			queueinformer.NewInformer(roleBindingQueue, roleBindingInformer.Informer(), op.syncObject, handleDelete, "rolebinding", metrics.NewMetricsNil()),
+			queueinformer.NewInformer(serviceAccountQueue, serviceAccountInformer.Informer(), op.syncObject, handleDelete, "serviceaccount", metrics.NewMetricsNil()),
```
njhale: Do we also need to watch clusterroles/rolebindings?

pkg/controller/operators/catalog/operator.go
```diff
@@ -141,57 +151,211 @@ func NewOperator(kubeconfigPath string, wakeupInterval time.Duration, operatorNa
...
+		"name":      metaObj.GetName(),
+		"namespace": metaObj.GetNamespace(),
+	})
+
+	logger.Debug("syncing")
+
+	if ownerutil.IsOwnedByKind(metaObj, v1alpha1.CatalogSourceKind) {
+		logger.Debug("requeueing owner CatalogSource")
+		owner := ownerutil.GetOwnerByKind(metaObj, v1alpha1.CatalogSourceKind)
+		o.catSrcQueue.AddRateLimited(fmt.Sprintf("%s/%s", metaObj.GetNamespace(), owner.Name))
```
njhale: I wonder how this will work with cross namespace catalog sources.

pkg/controller/operators/catalog/operator.go
```diff
@@ -141,57 +151,211 @@ func NewOperator(kubeconfigPath string, wakeupInterval time.Duration, operatorNa
...
 
+	logger := log.WithFields(log.Fields{
+		"source": catsrc.GetName(),
+	})
+
+	if catsrc.Spec.SourceType == v1alpha1.SourceTypeInternal || catsrc.Spec.SourceType == v1alpha1.SourceTypeConfigmap {
+		return o.syncConfigMapSource(logger, catsrc)
+	}
+
+	logger.WithField("sourceType", catsrc.Spec.SourceType).Warn("unknown source type")
```
njhale: should this be written back to the catalog source status?

pkg/lib/operatorlister/service.go
```diff
@@ -111,6 +111,6 @@ func (n *NullServiceNamespaceLister) Get(name string) (*v1.Service, error) {
 }
 
 // GetPodServices returns nil and an error explaining that this is a NullServiceNamespaceLister.
-func (n *NullServiceAccountNamespaceLister) GetPodServices(pod *v1.Pod) ([]*v1.Service, error) {
+func (n *NullServiceNamespaceLister) GetPodServices(pod *v1.Pod) ([]*v1.Service, error) {
```
njhale: nice catch

pkg/controller/registry/reconciler.go
ecordell: It's mostly a style issue; I personally think writing compositions of objects in go ends up more readable than writing pure functions. 

That said, all of the methods I defined on the decorator require state from the CatalogSource. Wrapping it in a (private) struct is equivalent to defining a set of functions that all take the catalog source as a first argument.

The way I'm using the struct here is way more like you'd use a State Monad to thread common state throughout pure function calls.

pkg/controller/registry/reconciler.go
ecordell: In general I think naming doesn't matter nearly as much as interface shapes, and names should just be whatever will help a reader the most. So if there's a name that would help you read it better, I would be happy to change it here 😁

pkg/controller/registry/reconciler.go
```diff
@@ -0,0 +1,360 @@
...
+	if err := c.ensureServiceAccount(source, overwrite); err != nil {
+		return err
+	}
+	if err := c.ensureRole(source, overwrite); err != nil {
+		return err
+	}
+	if err := c.ensureRoleBinding(source, overwrite); err != nil {
+		return err
+	}
+	if err := c.ensurePod(source, overwritePod); err != nil {
```
njhale: We could make use of a Deployment instead of a Pod. Then we would get rolling updates and a safeguard against losing a working registry pod (if the CatalogSource's ConfigMap is no longer valid).

cmd/catalog/main.go
```diff
@@ -19,6 +19,7 @@ import (
 const (
 	defaultWakeupInterval   = 15 * time.Minute
 	defaultCatalogNamespace = "tectonic-system"
```
njhale: tectonic-system?

pkg/api/apis/operators/v1alpha1/catalogsource_types.go
```diff
@@ -1,20 +1,31 @@
...
+
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators"
 )
 
 const (
 	CatalogSourceCRDAPIVersion = operators.GroupName + "/" + GroupVersion
 	CatalogSourceKind          = "CatalogSource"
 )
 
+// Approval is the user approval policy for an InstallPlan.
```
alecmerdler: Wrong doc comment.

PR #560, opened 2018-11-08
fix(queueInformers): use separate queue for each namespace

njhale:
### Description
- Uses a separate queue for each namespace
- Extends `make e2e-local` test timeout to 20 minutes
- Checks for olm.operatorNamespace difference to determine copied CSVs


test/e2e/operator_groups_e2e_test.go
```diff
@@ -105,39 +103,33 @@ func TestOperatorGroup(t *testing.T) {
...
+			},
+		},
 	}
 	_, err = crc.OperatorsV1alpha2().OperatorGroups(testNamespace).Create(&operatorGroup)
 	require.NoError(t, err)
 	expectedOperatorGroupStatus := v1alpha2.OperatorGroupStatus{
 		Namespaces: []*corev1.Namespace{createdOtherNamespace},
 	}
 
-	// instead of setting the label selector initially, do it here to force immediate reconcile
```
jpeeler: Did you take this out because a user would never do this?

test/e2e/operator_groups_e2e_test.go
```diff
@@ -105,39 +103,33 @@ func TestOperatorGroup(t *testing.T) {
 
 	oldCommand := patchOlmDeployment(t, c, otherNamespaceName)
 
-	log.Debug("Creating CSV")
```
jpeeler: I can understand removing the default log level of debug, but for a five minute long test I'd rather have some real time output for the logging.

ecordell:
```
# --- PASS: TestOperatorGroup (11.61s)
```
looks like it's running in 12s?

test/e2e/operator_groups_e2e_test.go
njhale: I took this out because it we shouldn't need to do it. The creation of an OperatorGroup should trigger an immediate reconcile. After taking it out I noticed it wasn't, which highlighted a bug with how we were registering multiple queue informers. This PR also addresses that bug! 🙂

jpeeler: Wow, ok. Good job!

test/e2e/operator_groups_e2e_test.go
jpeeler: After the explanation of above, this is fine (that is logging with t.Log).

ecordell:
/approved
/lgtm

(don't see any issue with re-vendoring)

PR #562, opened 2018-11-08
Service Catalog CSV: update resource limitations

jboyd01:
Some users are getting errors with etcd container unable to allocate memory.  I think these were copied from a default template - we've never had resource limitations on our containers in either OpenShift Service Catalog or upstream.


ecordell:
Is there any way that we could just adjust those to values that are sensible for service catalog?

jboyd01:
Let's go with bumping up the limits for the embedded etcd container.  Note that this container will be removed (hopefully soon) and we'll be using openshift's etcd instance.  Thanks for the discussion on this @ecordell

PR #564, opened 2018-11-12
feat(mocks): generate fakes and mocks in a container

njhale:
### Description

Adds the `container-mockgen` make rule for generating fakes and mocks in a container.


mockgen.Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM golang:1.10
```
jpeeler: Does this not work correctly with 1.11?

njhale: doesn't seem to for me.

Makefile
```diff
@@ -117,26 +118,33 @@ codegen:
 	$(CODEGEN) all $(PKG)/pkg/package-server/client $(PKG)/pkg/package-server/apis "packagemanifest:v1alpha1"
 
 container-codegen:
-	docker build -t olm:codegen -f codegen.Dockerfile .
+	docker build -t olm:codegen -f mockgen.Dockerfile .
```
ecordell: shouldn't this still be `codegen.Dockerfile`?

njhale: yup

PR #567, opened 2018-11-13
Cut 0.8.0

ecordell:


manifests/0000_30_07-certified-operators.configmap.yaml
```diff
@@ -1152,46 +1166,40 @@ data:
...
-      
           - name: mongodbshardedclusters.mongodb.com
             description: MongoDB Sharded Cluster Deployment
             displayName: MongoDB Sharded Cluster
             group: mongodb.com
             kind: MongoDbShardedCluster
             version: v1
-      
             resources:
+            # FIXME(alecmerdler): These are probably supposed to be different resources for the different CRD configurations
```
njhale: Do we want this in our release artifacts?

ecordell: I think so?

PR #571, opened 2018-11-13
Create role bindings for operator service accounts 

ecordell:
- Support operatorgroups watching all namespaces. In this case, `targetNamespaces` will be `""`
 - Create role bindings for operator service accounts in target namespaces
 - If target namespaces == all namespaces, generate cluster role bindings instead
 - Always include the operator namespace in the targetnamespace list.
 - OperatorGroup status uses a list of strings instead of a list of `Namespace` objects
 - Generated rolebinings have ownerreferences pointing to the _target_ CSV, so that they're cleaned up automatically if the target namespaces change.
 - Rewrote some of the operatorgroup testing to check both operator namespace and target namespace expectations.
 - Moved deployment template annotation to CSV loop
 - Removed a bunch of dead code / tests for installing permissions from the CSV loop (installplan generates them instead now)


jpeeler:
A few minor fixes you can steal are here: https://github.com/jpeeler/operator-lifecycle-manager/tree/evan-pr-571. (Can I push to your branch directly?)

pkg/controller/install/deployment.go
```diff
@@ -2,15 +2,11 @@ package install
 
 import (
 	"fmt"
-
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/wrappers"
```
njhale: nit: goimports

pkg/controller/install/deployment.go
```diff
@@ -224,6 +160,16 @@ func (i *StrategyDeploymentInstaller) checkForDeployments(deploymentSpecs []Stra
 		if !ready {
 			return StrategyError{Reason: StrategyErrReasonWaiting, Message: fmt.Sprintf("waiting for deployment %s to become ready: %s", dep.Name, reason)}
 		}
+
+		// check annotations
+		if len(i.templateAnnotations) > 0 && dep.Spec.Template.Annotations == nil {
+			return StrategyError{Reason: StrategyErrReasonAnnotationsMissing, Message: fmt.Sprintf("no annotations found on deployment")}
+		}
+		for key, value := range i.templateAnnotations {
+			if dep.Spec.Template.Annotations[key] != value {
```
njhale: At a glance this looks like it has the potential to panic if an annotation we expect to exist on the deployment template does not.

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -2,6 +2,7 @@ package olm
 
 import (
 	"fmt"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/registry/resolver"
```
njhale: nit: goimports

pkg/controller/registry/resolver/rbac.go
```diff
@@ -0,0 +1,152 @@
+package resolver
+
+import (
+	"fmt"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
```
njhale: nit: goimports

pkg/controller/registry/resolver/rbac.go
```diff
@@ -0,0 +1,152 @@
...
+			ownerutil.AddNonBlockingOwner(serviceAccount, csv)
+
+			permissions[permission.ServiceAccountName] = NewOperatorPermissions(serviceAccount)
+		}
+
+		// Create ClusterRole
+		role := &rbacv1.ClusterRole{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:            fmt.Sprintf("%s-%d", csv.GetName(), i),
+				Namespace:       csv.GetNamespace(),
```
njhale: nit: `ClusterRole/Bindings`s  don't need `ObjectMeta.Namespace`

pkg/controller/operators/olm/operator_test.go
```diff
@@ -2292,24 +2528,43 @@ func TestSyncOperatorGroups(t *testing.T) {
 			err = op.syncOperatorGroups(tt.initial.operatorGroup)
 			require.NoError(t, err)
 
+			// Sync csvs enough to get them back to succeeded state
+			for i := 0; i < 6; i++ {
```
jpeeler: Clever

pkg/controller/install/deployment.go
ecordell: if the key doesn't exist, you get the zero value for the value, not a panic: https://play.golang.org/p/Se7UYpYnleR

ecordell:
@jpeeler @njhale addressed comments!

pkg/controller/install/deployment.go
njhale: cool - my bad.

PR #572, opened 2018-11-15
fix(csv-requeue): requeue on namespace all if all namespaces are watched

njhale:
### Description

Fixes CSV requeuing issue when watching all namespaces.

### DIFF
--- a/pkg/controller/operators/olm/operator.go
+++ b/pkg/controller/operators/olm/operator.go
@@ -7,7 +7,6 @@ import (
 	"time"
 
 	log "github.com/sirupsen/logrus"
-	"k8s.io/api/apps/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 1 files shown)


pkg/controller/operators/olm/operator.go
```diff
@@ -271,29 +270,19 @@ func (a *Operator) requeueCSV(name, namespace string) {
 	// We can build the key directly, will need to change if queue uses different key scheme
 	key := fmt.Sprintf("%s/%s", namespace, name)
 	logger := log.WithField("key", key)
-	logger.Debugf("requeueing CSV")
+	logger.Debug("requeueing CSV")
 
-	if queue, ok := a.csvQueues[namespace]; ok {
+	if queue, ok := a.csvQueues[metav1.NamespaceAll]; len(a.csvQueues) == 1 && ok {
```
alecmerdler: Technically we can't pass any other namespaces when using `NamespaceAll`, but that's only enforced from runtime args and not the Go application code, so it's fine.

PR #573, opened 2018-11-16
docs(arch): add operator groups

jpeeler:
This adds a section on operator groups. It also removes the obsolete
section "namespace control loop", since namespaces aren't annotated any
more.

I tried to take into account https://github.com/operator-framework/operator-lifecycle-manager/pull/571, since it'll soon be done. It will still have to be further modified if work is done to make CSV visibility more granular, but not too much.

### DIFF
--- a/Documentation/design/architecture.md
+++ b/Documentation/design/architecture.md
@@ -10,6 +10,7 @@ Each of these Operators are responsible for managing the CRDs that are the basis
 | InstallPlan           | IP         | Catalog | calculated list of resources to be created in order to automatically install/upgrade a CSV |
 | CatalogSource         | CS         | Catalog | a repository of CSVs, CRDs, and packages that define an application                        |
 | Subscription          | Sub        | Catalog | used to keep CSVs up to date by tracking a channel in a package                            |
+| OperatorGroup         | <None>     | OLM     | method to group multiple namespaces and prepare for use by an operator                     |
 
 Each of these Operators are also responsible for creating resources:
 
@@ -17,8 +18,8 @@ Each of these Operators are also responsible for creating resources:
 |----------|----------------------------|
 | OLM      | Deployment                 |
 | OLM      | Service Account            |
-| OLM      | Roles                      |
-| OLM      | RoleBindings               |
+| OLM      | (Cluster)Roles             |
+| OLM      | (Cluster)RoleBindings      |
 | Catalog  | Custom Resource Definition |
 | Catalog  | ClusterServiceVersion   |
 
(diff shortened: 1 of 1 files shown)


Documentation/design/architecture.md
```diff
@@ -140,3 +135,16 @@ Package {name}
   |
   +-- Channel {name} --> CSV {version}
 ```
+
+## Operator Group design
+An operator group consists of a group of namespaces as specified by the label selector in the spec. There are two subsets within this selection that needs distinguishing: the operator namespace where the operator actually runs in and the target namespace(s) which are namespaces the operator have permissions to operate in. Note that the operator namespace is considered part of the operator group's target namespaces even if that particular namespace does not match the label selector.
```
njhale: > An operator group consists of a group of namespaces as specified by the label selector in the spec. There are two subsets within this selection that needs distinguishing: the operator namespace where the operator actually runs in and the target namespace(s)

To me this reads like the operator namespace is selected by the label selector. I thought the operator namespace was implicitly the `OperatorGroup`'s `ObjectMeta.Namespace`.

Documentation/design/architecture.md
```diff
@@ -140,3 +135,16 @@ Package {name}
...
+
+Once an operator group has been created, the focus is around the target namespaces and the resources contained in those namespaces. The status for an operator group is constantly updated to always list the namespaces matching the label selector as specified in the spec. In addition to maintaining an updated status, the operator group control loop also maintains creating RBAC rules, providing operator group knowledge to operators, and copying CSVs, each of which is described further below.
+
+RBAC rules are created for two reasons. The first reason is for restricting access to the API (CRDs) of the installed operators. It is possible that the administrator does not want the full functionality of the operator to be granted in all cases. The second reason is to give the operator the necessary permissions to operate, which is tied to the specified operator group service account.
+
+Operator group target namespace information is made available to operators via annotations on the deployment. A second step of using either an environment variable or downward API is technically necessary to pass this information, which is used to know where the operator has permissions to operate.
```
njhale: >  of using either an environment variable or downward API ...

The [downward API]( https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/#the-downward-api) is used for both environment variable and volumes.

Documentation/design/architecture.md
jpeeler: Ok see what you think now.

njhale: awesome!

jpeeler:
This has now been updated to account for changes in:
https://github.com/operator-framework/operator-lifecycle-manager/pull/589
https://github.com/operator-framework/operator-lifecycle-manager/pull/601

PR #575, opened 2018-11-20
Make use a valid clusterPermissions rule verb

wongma7:
Repeat of https://github.com/operator-framework/operator-lifecycle-manager/pull/499 because it got overwritten somehow


wongma7:
@ecordell pls /retest, seems this merge is blocked on it, thanks!

PR #576, opened 2018-11-20
Propagate Labels from `CatalogSource` to `PackageManifests`

alecmerdler:
### Description

All `metadata.labels` from a given `CatalogSource` will be copied to each of its `PackageManifest` objects. This enables us to label packages for server-side filtering/sorting (for both `kubectl` and console).

Addresses https://jira.coreos.com/browse/ALM-809

### DIFF
--- a/pkg/package-server/provider/inmem.go
+++ b/pkg/package-server/provider/inmem.go
@@ -76,7 +76,7 @@ func NewInMemoryProvider(informers []cache.SharedIndexInformer, queueOperator *q
 }
 
 // parsePackageManifestsFromConfigMap returns a list of PackageManifests from a given ConfigMap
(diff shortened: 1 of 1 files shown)


pkg/package-server/provider/inmem.go
```diff
@@ -76,7 +76,7 @@ func NewInMemoryProvider(informers []cache.SharedIndexInformer, queueOperator *q
 }
 
 // parsePackageManifestsFromConfigMap returns a list of PackageManifests from a given ConfigMap
-func parsePackageManifestsFromConfigMap(cm *corev1.ConfigMap, catalogSourceName, catalogSourceNamespace, catalogSourcePublisher, catalogSourceDisplayName string) ([]packagev1alpha1.PackageManifest, error) {
+func parsePackageManifestsFromConfigMap(cm *corev1.ConfigMap, catsrc *operatorsv1alpha1.CatalogSource) ([]packagev1alpha1.PackageManifest, error) {
```
alecmerdler: Change the function signature to just accept the whole `CatalogSource` rather than individual fields.

PR #578, opened 2018-11-21
Add packagemanifest:aggregated-apiserver-clusterrole

sttts:
Fixes https://github.com/operator-framework/operator-lifecycle-manager/issues/577

PR #583, opened 2018-11-26
Add `packagemanifest:aggregated-apiserver-clusterrole`

### LINKED ISSUE #577: ClusterRoleBinding against aggregated-apiserver-clusterrole without role manifest
There is a dependency against the aggregated-apiserver-clusterrole cluster role, but the role is not installed into the cluster: https://github.com/operator-framework/operator-lifecycle-manager/blob/master/manifests/0000_30_13-packageserver.yaml#L64

This causes https://github.com/openshift/cluster-openshift-apiserver-operator/issues/38.

The cluster role originally comes from the sample-apiserver https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/sample-apiserver/artifacts/example/rbac.yaml. Please add this to the manifests here to stop the noise in the logs.

alecmerdler:
### Description

Updates the `ClusterRole` used by the `package-manifest` apiserver.

Fixes #577 
Closes #578

### DIFF
--- a/deploy/chart/templates/0000_30_13-packageserver.yaml
+++ b/deploy/chart/templates/0000_30_13-packageserver.yaml
@@ -53,6 +53,15 @@ subjects:
   name: olm-operator-serviceaccount
   namespace: {{ .Values.namespace }}
 ---
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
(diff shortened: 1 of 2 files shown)


ecordell:
Force merging; I think this will fix a number of the e2e tests in e2e-aws-olm

PR #585, opened 2018-11-27
fix(annotations): merge CSV and pod template annotations when installing deployments

njhale:
### Description 
Merges CSV and pod template annotations when installing deployments


pkg/controller/install/deployment.go
```diff
@@ -63,7 +63,17 @@ func (i *StrategyDeploymentInstaller) installDeployments(deps []StrategyDeployme
 		dep := &appsv1.Deployment{Spec: d.Spec}
 		dep.SetName(d.Name)
 		dep.SetNamespace(i.owner.GetNamespace())
-		dep.Spec.Template.SetAnnotations(i.templateAnnotations)
+
+		// Merge annotations (to avoid losing info from pod template)
+		annotations := make(map[string]string)
```
njhale: This where we needed to merge annotations, instead of just replacing them.

PR #587, opened 2018-11-28
Adding description to package manifest object

galletti94:
Solves [ALM-819](https://jira.coreos.com/browse/ALM-819)

/cc @alecmerdler

### DIFF
--- a/pkg/package-server/apis/openapi/zz_generated.openapi.go
+++ b/pkg/package-server/apis/openapi/zz_generated.openapi.go
@@ -152,6 +152,13 @@ func schema_package_server_apis_packagemanifest_v1alpha1_CSVDescription(ref comm
 							},
 						},
 					},
+					"description": {
+						SchemaProps: spec.SchemaProps{
+							Description: "LongDescription is the CSV's description",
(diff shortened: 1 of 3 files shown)


alecmerdler:
@galletti94 I think you need to run `make codegen` and commit the generated code.

njhale:
> @galletti94 I think you need to run `make codegen` and commit the generated 

Since codegen has issues with go modules, it's recommended to use `make gen-all` which runs codegen in a go 1.10 container.

galletti94:
@njhale turns out I had an easier time with `make codegen` for some reason. 

I'm still trying to get `make gen-all` to work. The generated code should be the same right?

njhale:
The only difference should be the code generated for openapi; which it looks like did change `pkg/package-server/apis/openapi/zz_generated.openapi.go` 🙂

galletti94:
@njhale @alecmerdler what is the command for re-launching the gitlab-ci checks?

alecmerdler:
@galletti94 Normally I just `git push --force` with the same commit. However, please note that we are having CI issues at the moment.

PR #588, opened 2018-11-28
fix(olm): update CSV status properly

jpeeler:
This sets status after creation.


jpeeler:
Although this will fix the e2e, it may not be the best fix after all.
/hold

jpeeler:
This fix was included in https://github.com/operator-framework/operator-lifecycle-manager/pull/603.

PR #589, opened 2018-11-28
Don't process CSVs without operatorgroup

ecordell:
This PR also moves the copying logic into the CSV control loop so that we don't have to requeue operatorgroups every time a CSV changes.

We probably shouldn't merge until there's a default operatorgroup shipping with OLM (otherwise you won't be able to install any operators)

/hold


pkg/controller/operators/olm/operator.go
```diff
@@ -419,9 +427,61 @@ func (a *Operator) syncClusterServiceVersion(obj interface{}) (syncError error)
...
+
+	operatorGroup, err := a.lister.OperatorsV1alpha2().OperatorGroupLister().OperatorGroups(csv.GetNamespace()).Get(operatorGroupName)
+	// operatorgroup not found
+	if err != nil {
+		logger.Info("operatorgroup not found")
+		return nil
+	}
+
+	// target namespaces don't match
+	if annotations[operatorGroupTargetsAnnotationKey] != strings.Join(operatorGroup.Status.Namespaces, ",") {
```
njhale: We may need to rethink this part if we aren't going to have target namespaces annotated on CSVs.

pkg/controller/operators/olm/operator_test.go
```diff
@@ -391,6 +391,17 @@ func withAnnotations(obj runtime.Object, annotations map[string]string) runtime.
 	return meta.(runtime.Object)
 }
 
+func addAnnotations(annotations map[string]string, add map[string]string) map[string]string {
```
njhale: 👍

deploy/chart/templates/0000_30_16-operatorgroup-default.yaml
```diff
@@ -2,4 +2,14 @@ apiVersion: operators.coreos.com/v1alpha2
 kind: OperatorGroup
 metadata:
   name: global-operators
-  namespace: {{ .Values.operator_namespace }}
\ No newline at end of file
+  namespace: {{ .Values.operator_namespace }}
+---
+apiVersion: operators.coreos.com/v1alpha2
```
ecordell: Added an operatorgroup for the packages server, since they're required now

test/e2e/operator_groups_e2e_test.go
```diff
@@ -274,9 +316,78 @@ func TestOperatorGroup(t *testing.T) {
...
+
+	ruleChecker := install.NewCSVRuleChecker(roleInformer.Lister(), roleBindingInformer.Lister(), clusterRoleInformer.Lister(), clusterRoleBindingInformer.Lister(), &aCSV)
+
+	t.Log("Waiting for operator to have rbac in target namespace")
+	err = wait.Poll(pollInterval, pollDuration, func() (bool, error) {
+		for _, perm := range permissions {
+			sa, err := c.GetServiceAccount(opGroupNamespace, perm.ServiceAccountName)
+			require.NoError(t, err)
+			for _, rule := range perm.Rules {
+				satisfied, err := ruleChecker.RuleSatisfied(sa, otherNamespaceName, rule)
```
ecordell: added a test to verify RBAC in target namespaces (kinda funky with informers in the test, but it works)

test/e2e/operator_groups_e2e_test.go
```diff
@@ -274,9 +316,78 @@ func TestOperatorGroup(t *testing.T) {
...
+				if !satisfied {
+					return false, nil
+				}
+			}
+		}
+		return true, nil
+	})
+
+	// validate provided API clusterroles for the operatorgroup
+	adminRole, err := c.KubernetesInterface().RbacV1().ClusterRoles().Get(operatorGroup.Name+"-admin", metav1.GetOptions{})
```
ecordell: This validates both that the operatorgroup clusterrole was created, and that it has correctly aggregated the csv-specific roles

test/e2e/operator_groups_e2e_test.go
```diff
@@ -286,8 +397,4 @@ func TestOperatorGroup(t *testing.T) {
 	})
 	require.NoError(t, err)
 
-	err = c.KubernetesInterface().CoreV1().Namespaces().Delete(otherNamespaceName, &metav1.DeleteOptions{})
```
ecordell: these moved into `defer` blocks further up

pkg/controller/operators/olm/operator.go
```diff
@@ -403,7 +404,31 @@ func (a *Operator) syncClusterServiceVersion(obj interface{}) (syncError error)
 		"phase":     clusterServiceVersion.Status.Phase,
 	})
 
-	operatorNamespace, ok := clusterServiceVersion.GetAnnotations()["olm.operatorNamespace"]
+	operatorGroup := a.operatorGroupForActiveCSV(logger, clusterServiceVersion)
```
ecordell: this is the main part of the PR: gating csv processing on whether there are operatorgroups or not

pkg/controller/operators/olm/operator.go
```diff
@@ -429,9 +454,67 @@ func (a *Operator) syncClusterServiceVersion(obj interface{}) (syncError error)
 		}
 		syncError = fmt.Errorf("error transitioning ClusterServiceVersion: %s and error updating CSV status: %s", syncError, updateErr)
 	}
+
+	// Check if we need to do any copying / annotation for the operatorgroup
+	if err := a.copyCsvToTargetNamespace(updatedCSV, operatorGroup); err != nil {
```
ecordell: I moved these three pieces from the operatorgroup control loop to here. The symptom was the e2e test taking a very long time (because csv changes didn't requeue their operatorgroups). 

One option would've been to retrigger the owning operatorgroup on CSV change, but then in the operatorgroup side, we reach out and collect up each CSV to perform some of these actions. Moving them here removes all of that processing and lets us treat each CSV in isolation.

pkg/controller/operators/olm/operator.go
```diff
@@ -429,9 +454,67 @@ func (a *Operator) syncClusterServiceVersion(obj interface{}) (syncError error)
...
+		logger.WithError(err).Info("couldn't copy CSV to target namespaces")
+	}
+
+	// Ensure operator has access to targetnamespaces
+	if err := a.ensureRBACInTargetNamespace(updatedCSV, operatorGroup); err != nil {
+		logger.WithError(err).Info("couldn't ensure RBAC in target namespaces")
+	}
+
+	// Ensure cluster roles exist for using provided apis
+	if err := a.ensureClusterRolesForCSV(updatedCSV, operatorGroup); err != nil {
```
ecordell: Previously we were generating operatorgroup clusterroles directly (e.g. an operatorgroup with permissions on all provided apis in the group). 

To speed things up, the operatorgroup control loop just writes out a ClusterRole with aggregation rules, and each individual CSV writes out clusterroles for its provided apis with the correct label to get aggregated to the operatorgroup clusterrole. There's an e2e test to verify the aggregated roles get the correct permissions.

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -18,63 +17,137 @@ import (
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/ownerutil"
 )
 
+const (
+	operatorGroupAnnotationKey          = "olm.operatorGroup"
```
jpeeler: One of us is going to have to rebase a bit. I ended up pulling this into types so the e2e can also use them in https://github.com/operator-framework/operator-lifecycle-manager/pull/616/. Will finish reviewing this tomorrow.

ecordell: I think #616 will need to change (slightly) anyway - see the `operatorGroupForActiveCSV ` method.

pkg/controller/operators/olm/operator.go
```diff
@@ -403,7 +404,31 @@ func (a *Operator) syncClusterServiceVersion(obj interface{}) (syncError error)
...
+		if len(opgroups) == 1 {
+			a.addOperatorGroupAnnotations(&clusterServiceVersion.ObjectMeta, opgroups[0])
+			_, err := a.client.OperatorsV1alpha1().ClusterServiceVersions(clusterServiceVersion.GetNamespace()).Update(clusterServiceVersion)
+			if err != nil {
+				logger.WithField("opgroup", opgroups[0].GetName()).Error("error adding operatorgroup annotation")
+			}
+			return
+		}
+		if len(opgroups) > 1 {
+			logger.Warn("csv created in namespace with multiple operatorgroups, can't pick one automatically")
```
njhale: Do we ever want to write the status back to the CSV when there are operator-group issues?

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -18,63 +17,137 @@ import (
...
+		if err := a.ensureProvidedAPIClusterRole(operatorGroup, csv, namePrefix, EditSuffix, group, plural); err != nil {
+			return err
+		}
+		if err := a.ensureProvidedAPIClusterRole(operatorGroup, csv, namePrefix, ViewSuffix, group, plural); err != nil {
 			return err
 		}
 	}
-	log.Debug("CSV annotation completed")
+	for _, owned := range csv.Spec.APIServiceDefinitions.Owned {
+		namePrefix := fmt.Sprintf("%s-%s-", owned.Name, owned.Version)
```
njhale: Is `ownee.Name` always set for `APIServiceDefinitions`?

ecordell: It is not, but I want to review all of the current ones, add them, and then go make it `required` in the json schema

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -217,66 +290,90 @@ func (a *Operator) ensureTenantRBAC(operatorNamespace, targetNamespace string, c
...
+					logger.WithError(err).Error("status update for target CSV failed")
+					return err
+				}
+			}
+
 			continue
 		} else if k8serrors.IsNotFound(err) {
 			newCSV := csv.DeepCopy()
 			newCSV.SetNamespace(ns)
 			newCSV.SetResourceVersion("")
```
njhale: `ObjectMeta.ResourceVersion` is read only so I don't think this needs to be cleared here.

pkg/controller/operators/olm/operator.go
```diff
@@ -429,9 +454,67 @@ func (a *Operator) syncClusterServiceVersion(obj interface{}) (syncError error)
...
+
+	operatorGroup, err := a.lister.OperatorsV1alpha2().OperatorGroupLister().OperatorGroups(csv.GetNamespace()).Get(operatorGroupName)
+	// operatorgroup not found
+	if err != nil {
+		logger.Info("operatorgroup not found")
+		return nil
+	}
+
+	// target namespaces don't match
+	if annotations[operatorGroupTargetsAnnotationKey] != strings.Join(operatorGroup.Status.Namespaces, ",") {
```
jpeeler: Would this scenario only occur if an operator group was modified after annotations had been previously correct?

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -18,63 +17,137 @@ import (
...
+				kubeRBACAggregationKeyPrefix + suffix:       "true",
+				operatorGroupAggregrationKeyPrefix + suffix: operatorGroup.GetName(),
+			},
+		},
+		Rules: []rbacv1.PolicyRule{{Verbs: VerbsForSuffix[suffix], APIGroups: []string{group}, Resources: []string{resource}}},
+	}
+	ownerutil.AddNonBlockingOwner(clusterRole, csv)
+	existingCR, err := a.OpClient.KubernetesInterface().RbacV1().ClusterRoles().Create(clusterRole)
+	if k8serrors.IsAlreadyExists(err) {
+		if existingCR != nil && reflect.DeepEqual(existingCR.Labels, clusterRole.Labels) && reflect.DeepEqual(existingCR.Rules, clusterRole.Rules) {
```
jpeeler: Does createPatch not handle a no-op or is this just an optimization?

PR #591, opened 2018-11-29
Add operator namespace

jpeeler:
This creates a new namespace for operators, sets OLM to also watch that namespace, and creates an operator group to watch all namespaces. This is part one of a final goal to have all operators aware of the available target namespaces. 

ALM-727


scripts/run_e2e_local.sh
```diff
@@ -8,8 +8,7 @@ set -e
 timestamp=$(date +%s)
 namespace="e2e-tests-${timestamp}-$RANDOM"
 
-tmpdir=`mktemp -d 2>/dev/null || mktemp -d -t 'valuetmpdir'`
-cp test/e2e/e2e-values.yaml ${tmpdir}/e2e-values.yaml
```
jpeeler: If you're looking at this line in the only one commit at a time view, I made a mistake here with removing this line. But it's rectified in the last commit.

deploy/chart/templates/0000_30_10-olm-operator.deployment.yaml
```diff
@@ -22,10 +22,6 @@ spec:
         - name: olm-operator
           command:
           - /bin/olm
-          {{- if .Values.watchedNamespaces }}
```
ecordell: We don't want to remove this option (it gets used by some of the testing still), we just don't want to set it it some cases

deploy/chart/templates/0000_30_15-operatorgroup-default.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: operators.coreos.com/v1alpha2
+kind: OperatorGroup
+metadata:
+  name: olm-operators
```
ecordell: how about `global-operators`?

deploy/okd/values.yaml
```diff
@@ -1,6 +1,7 @@
 rbacApiVersion: rbac.authorization.k8s.io
 namespace: openshift-operator-lifecycle-manager
 catalog_namespace: openshift-operator-lifecycle-manager
+operator_namespace: operators
```
ecordell: for ocp/okd we should make this `openshift-operators`

manifests/0000_30_00-namespace.yaml
```diff
@@ -6,3 +6,8 @@ metadata:
   name: openshift-operator-lifecycle-manager
   labels:
     openshift.io/run-level: "1"
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: operators
```
ecordell: name should come from the template, right?

manifests/0000_30_15-operatorgroup-default.yaml
```diff
@@ -0,0 +1,7 @@
+---
+# Source: olm/templates/0000_30_15-operatorgroup-default.yaml
+apiVersion: operators.coreos.com/v1alpha2
+kind: OperatorGroup
+metadata:
+  name: olm-operators
+  namespace: operators
```
ecordell: namespace should come from the template

Documentation/install/local-values-shift.yaml
```diff
@@ -2,6 +2,7 @@ rbacApiVersion: authorization.openshift.io
 namespace: local
 watchedNamespaces: local
 catalog_namespace: local
```
njhale: I would like us to be consistent with camelCase if we can.

PR #579, opened 2018-11-21
feat(apiservices): implement apiservice mock server

njhale:
### Description
- Implements a mock extension api-server (`mock-ext-server`) which can be configured to register mock GVKs with discovery.
- Uses `mock-ext-server` in `TestCreateCSVWithOwnedAPIService` e2e test.
- Uses `default` ServiceAccount if none are listed in a CSVs owned APIService deployment spec.
- Runs `goimports` on project.

PR #593, opened 2018-12-03
test(csv): use mock-ext-server for owned apiservice e2e test

njhale:
Description:

Uses the [mock-extension-apiserver](https://github.com/operator-framework/mock-extension-apiserver) for all e2e tests involving an owned APIService.

Supersedes [#579](https://github.com/operator-framework/operator-lifecycle-manager/pull/579)


pkg/controller/operators/olm/apiservices.go
```diff
@@ -129,6 +130,9 @@ func (a *Operator) checkAPIServiceResources(csv *v1alpha1.ClusterServiceVersion,
 
 		// Ensure the Deployment's ServiceAccount exists
 		serviceAccountName := deployment.Spec.Template.Spec.ServiceAccountName
+		if serviceAccountName == "" {
```
njhale: use `default` if one does not exist

PR #595, opened 2018-12-03
Fix copied CSV leaks targeted namespaces

dinhxuanvu:
The CSVs should be copied to targeted namespaces first with the
original contents. Then, the main CSVs are annotated with targeted
namespaces afterwards so that the copied CSVs won't contain object
meta with targeted namespaces information.

Jira: https://jira.coreos.com/browse/ALM-824


dinhxuanvu:
Replaced by #601

PR #582, opened 2018-11-26
[WIP] Reduce `ClusterServiceVersion` E2E Test Time

alecmerdler:
### Description

Reduces the time for running certain E2E tests for `ClusterServiceVersions` by creating the CSV first, then its dependencies (CRD, APIService) and adding `ownerReferences`.

PR #599, opened 2018-12-04
fix(e2e): fix CSV tests

njhale:
### Description

- Looks in `testNamespace` instead of `operatorNamespace` for created CSVs and Deployments in `TestCreateCSVWithOwnedAPIService`.
- Creates CSV first before use in CRD OwnerReferences (UID was empty which meant that changes to the CRD would not requeue the CSV).

Supersedes [#582](https://github.com/operator-framework/operator-lifecycle-manager/pull/582)

### DIFF
--- a/test/e2e/csv_e2e_test.go
+++ b/test/e2e/csv_e2e_test.go
@@ -596,11 +596,16 @@ func TestCreateCSVRequirementsMetCRD(t *testing.T) {
 		},
 	}
 
-	// Create dependency first (CRD)
+	// Create CSV first, knowing it will fail
+	cleanupCSV, err := createCSV(t, c, crc, csv, testNamespace, true, false)
+	require.NoError(t, err)
+	defer cleanupCSV()
+
+	fetchedCSV, err := fetchCSV(t, crc, csv.Name, csvPendingChecker)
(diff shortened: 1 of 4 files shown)


test/e2e/csv_e2e_test.go
```diff
@@ -714,39 +725,12 @@ func TestCreateCSVRequirementsMetCRD(t *testing.T) {
 	_, err = c.CreateClusterRoleBinding(&nonResourceClusterRoleBinding)
 	require.NoError(t, err, "could not create ClusterRoleBinding")
 
-	cleanupCSV, err := createCSV(t, c, crc, csv, testNamespace, true, false)
-	require.NoError(t, err)
-	defer cleanupCSV()
-
-	fmt.Println("checking for deployment")
-	// Poll for deployment to be ready
-	err = wait.Poll(pollInterval, pollDuration, func() (bool, error) {
```
ecordell: is there a reason we removed the deployment check?

(It's fine as long as there's another test somewhere that can verify that if the CSV succeeded, the deployment is actually running)

ecordell: The `ocs` e2e tests poll for deployment status, so nevermind

PR #600, opened 2018-12-05
feat(ci): remove gitlab / failfast integration

ecordell:
We've migrated to openshift ci and prow

🎉 🎈


njhale:
We can also get rid of `./scripts/wait_in_queue.sh`.

.gitlab-ci.jsonnet
```diff
@@ -1,202 +0,0 @@
...
-        script+: [
-            "curl -X POST --data-urlencode \"payload={\\\"text\\\": \\\"New OLM Operator quay.io/coreos/olm:${CI_COMMIT_REF_SLUG}-${CI_COMMIT_SHA} deployed to ${TEAMUI_HOST}/k8s/ns/operator-lifecycle-manager/deployments/alm-operator\\\"}\" ${TEAMUI_SLACK_URL}",
-        ],
-        environment+: {
-            name: "teamui",
-        },
-        only: [''],
-    },
-
-    "deploy-openshift": baseJob.Deploy {
```
alecmerdler: How can we do this using Prow?

ecordell: When the e2e tests run, you can login to the cluster it spins up if that's good enough.

If you want a cluster that won't disappear after the tests, you can pull the `origin-release` image built by the PR and start a cluster with installer (locally with libvirt or remotely in AWS)

It's not _as_ easy, but it's not particularly difficult either. If we miss this too much we can invest some time making additional tide commands to do this for us maybe?

ecordell:
@njhale removed `wait_in_queue`

.gitlab-ci.jsonnet
alecmerdler: We need `master` of OLM to be deployed to the same persistent cluster where the end-to-end tests for `openshift/console` runs.

ecordell: We can keep it, but console just needs to run e2e tests in openshift ci and it will automatically get tested against master. That seems easier to maintain long term

njhale: @alecmerdler I'm wondering why the `openshift/console` ci doesn't deploy the `master` version of OLM to its persistent cluster. It seems like the responsibility should be inverted from what we had before.

ecordell:
See #602 for now

.gitlab-ci.jsonnet
alecmerdler: > console just needs to run e2e tests in openshift ci and it will automatically get tested against master

True, but we can't really talk since this literal PR is what takes us off our own ad-hoc CI.

> I'm wondering why the openshift/console ci doesn't deploy the master version of OLM to its persistent cluster

Keyword here is `persistent`. That 3.11 cluster is spun up every once in a while (when it eventually dies) so doesn't receive updates.

PR #609, opened 2018-12-07
feat(package-server): create subscription manifest

njhale:
### Description

- Adds helm template for package-service Subscription manifest
- Adds helm template for olm-operators CatalogSource/ConfigMap
- Generates manifests for package-service Subscription
- Generates manifests for olm-operators CatalogSource/ConfigMap
- Adds Subscription requeuing on CatalogSource update


ecordell:
@njhale It looks like running `gofmt` on the things we codegen is causing `verify` to fail?

njhale:
> @njhale It looks like running `gofmt` on the things we codegen is causing `verify` to fail?

Yup, looks like it. I ran `goimports` on the `pkg/...` earlier. I'll amend this commit with the regenerated code soon.

PR #613, opened 2018-12-10
Use CSVPhaseSucceeded for Copied ClusterServiceVersions

alecmerdler:
### Description

Was previously using `phase: Installing`.


ecordell:
In #589 the control loop is updated to copy the CSV status every time it changes, so that target CSVs always have status that matches the primary. (e.g. if they go unhealthy)

alecmerdler:
@ecordell Ah okay, I was definitely seeing different `status.phase` for copied CSVs when using an `OperatorGroup`.

ecordell:
@alecmerdler I think this is fixed now that #589 is merged. I'll close for now and we can re-open if you find any issues in the UI

PR #614, opened 2018-12-10
Verify CRD's condition to ensure it's registered with k8s API (rebased)

jpeeler:
This is Vu's https://github.com/operator-framework/operator-lifecycle-manager/pull/553, but rebased and with tests passing (should be at least!). It has been modified from his original code slightly - ~mainly instead of creating a crdWithStatus, instead uses crdWithoutStatus which required far less code modification.~

### DIFF
--- a/pkg/controller/operators/olm/requirements.go
+++ b/pkg/controller/operators/olm/requirements.go
@@ -30,8 +30,8 @@ func (a *Operator) requirementStatus(strategyDetailsDeployment *install.Strategy
 		// check if CRD exists - this verifies group, version, and kind, so no need for GVK check via discovery
 		crd, err := a.lister.APIExtensionsV1beta1().CustomResourceDefinitionLister().Get(r.Name)
 		if err != nil {
-			a.Log.Debugf("Setting 'met' to false, %v with err: %v", r.Name, err)
 			status.Status = v1alpha1.RequirementStatusReasonNotPresent
+			a.Log.Debugf("Setting 'met' to false, %v with status %v, with err: %v", r.Name, status, err)
 			met = false
 			statuses = append(statuses, status)
 			continue
@@ -53,6 +53,7 @@ func (a *Operator) requirementStatus(strategyDetailsDeployment *install.Strategy
 
 			if !served {
 				status.Status = v1alpha1.RequirementStatusReasonNotPresent
+				a.Log.Debugf("Setting 'met' to false, %v with status %v, CRD version %v not found", r.Name, status, r.Version)
 				met = false
 				statuses = append(statuses, status)
 				continue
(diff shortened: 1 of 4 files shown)


jpeeler:
Copied from old PR:

OLM needs to verify CRD's condition to make sure it is registered
with k8s API during CSV control loop before installing operator.
This verification will help to avoid errors related to the
unavailability of an API when installing an operator via OLM.

Jira: https://jira.coreos.com/browse/ALM-775

pkg/controller/operators/olm/requirements_test.go
```diff
@@ -453,6 +474,58 @@ func TestRequirementAndPermissionStatus(t *testing.T) {
...
+			},
+			expectedError: nil,
+		},
+		{
+			description: "RequirementNotMet/NamesConflictedCRDVersion",
+			csv: csv("csv1",
+				namespace,
+				"",
+				installStrategy("csv1-dep", nil, nil),
+				[]*v1beta1.CustomResourceDefinition{crdWithoutStatus("c1", "v2")},
```
ecordell: I think I'm missing what makes this CRD be unavailable? What's the difference between this test and the previous?

pkg/controller/operators/olm/requirements.go
```diff
@@ -3,11 +3,13 @@ package olm
 import (
 	"encoding/json"
 	"fmt"
+
 	"github.com/sirupsen/logrus"
 
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
 	olmErrors "github.com/operator-framework/operator-lifecycle-manager/pkg/controller/errors"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/install"
+	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
```
njhale: nit: import ordering

jpeeler: I thought my tooling was out of date, but I just did a `go get golang.org/x/tools/cmd/goimports` and ran goimports on requirements.go and it still put the import line in the same spot. Where does yours get put?

jpeeler:
/retest
This passed for me locally.

pkg/controller/operators/olm/requirements.go
```diff
@@ -28,36 +30,59 @@ func (a *Operator) requirementStatus(strategyDetailsDeployment *install.Strategy
...
+					if version.Served {
+						status.Status = v1alpha1.RequirementStatusReasonPresent
+						served = true
+					}
+					break
+				}
+			}
+
+			if !served {
+				status.Status = v1alpha1.RequirementStatusReasonNotPresent
```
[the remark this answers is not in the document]
chancez: Oh god please yes. This is the worst part of debugging these things.

PR #616, opened 2018-12-11
fix(olm): don't annotate target namespace on copied CSVs

jpeeler:
This removes target namespace information on all copied CSVs and supercedes https://github.com/operator-framework/operator-lifecycle-manager/pull/601.

I believe the approach of deleting the annotation is clearer than adding separate specific annotations for copied CSVs.

ALM-824

### DIFF
--- a/pkg/controller/operators/olm/operator_test.go
+++ b/pkg/controller/operators/olm/operator_test.go
@@ -2385,7 +2385,7 @@ func TestSyncOperatorGroups(t *testing.T) {
 					annotatedDeployment,
 				},
 				targetNamespace: {
-					withAnnotations(targetCSV.DeepCopy(), map[string]string{"olm.targetNamespaces": targetNamespace + "," + operatorNamespace, "olm.operatorGroup": "operator-group-1", "olm.operatorNamespace": operatorNamespace}),
+					withAnnotations(targetCSV.DeepCopy(), map[string]string{"olm.operatorGroup": "operator-group-1", "olm.operatorNamespace": operatorNamespace}),
 					&rbacv1.Role{
 						TypeMeta: metav1.TypeMeta{
 							Kind:       "Role",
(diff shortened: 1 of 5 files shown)


pkg/controller/operators/olm/operator_test.go
```diff
@@ -2358,7 +2358,7 @@ func TestSyncOperatorGroups(t *testing.T) {
 					annotatedDeployment,
 				},
 				targetNamespace: {
-					withAnnotations(targetCSV.DeepCopy(), map[string]string{"olm.targetNamespaces": targetNamespace + "," + operatorNamespace, "olm.operatorGroup": "operator-group-1", "olm.operatorNamespace": operatorNamespace}),
+					withAnnotations(targetCSV.DeepCopy(), map[string]string{"olm.operatorGroup": "operator-group-1", "olm.operatorNamespace": operatorNamespace}),
```
alecmerdler: Why aren't we using the constants here?

jpeeler: I fixed up all of the string references now.

test/e2e/operator_groups_e2e_test.go
```diff
@@ -80,13 +80,25 @@ func patchOlmDeployment(t *testing.T, c operatorclient.ClientInterface, newNames
 }
 
 func checkOperatorGroupAnnotations(obj metav1.Object, op *v1alpha2.OperatorGroup, targetNamespaces string) error {
```
ecordell: What about just passing in the boolean `mustHaveTargetNamespaces` as an arg?

Just thinking about when I'm reading back over this in the future and I come across the `"-"` vs looking up the function signature and seeing a boolean with a descriptive name.

ecordell: (or even having two separate functions, on for checking targets and one for not)

jpeeler: I saw some precedent of using "-" in k8s, but what you suggested is better.

ecordell:
/lgtm 

the test failures don't look related, but we should figure out why they're happening (probably need to re-lgtm, but doing it now just in case they pass)

jpeeler:
/retest
Will investigate further if things don't look better.

jpeeler:
/retest
other PR passed, so I'm going to try this one more time

PR #622, opened 2018-12-14
[WIP] Install operatorsource crd and default cr

kevinrizza:
-Add the operatorsource crd in templates
-Add the operatorsource cr global-operators in templates
-Add yaml files to latest release via make release
-Install default appregistry in cluster by default


deploy/chart/templates/0000_30_19-operatorsource-default.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: "marketplace.redhat.com/v1alpha1"
+kind: "OperatorSource"
+metadata:
+  name: "marketplace-operators"
+  namespace: "openshift-operators"
```
ecordell: this should be `namespace: {{ .Values.operator_namespace }}`

kevinrizza: @ecordell That sets the namespace in the upstream/manifests/ folder to "operators" instead of "openshift-operators". Is that intentional?

kevinrizza: To follow up on this, the upstream/manifests/ folder is for non-openshift upstream kubernetes distributions which is the reason for the namespace change.

I have addressed the requested change and updated the pr to reflect it.

PR #626, opened 2018-12-17
Write cluster operator status after successful startup

jpeeler:
There's an included CLI arg to disable this behavior. I'm not sure which way the default should be, but currently it's turned on by default.
Update: the CLI arg now specifies the name to write the status to as suggested.

ALM-788


manifests/0000_30_18-operatorstatus.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: operatorstatus.openshift.io
+kind: ClusterOperator
+metadata:
+  namespace: openshift-operator-lifecycle-manager
+  name: operator-lifecycle-manager
+status:
+  version: 1.0.0
```
jpeeler: Should the version match the latest release manifest (looks like 0.8.0 currently)?

cmd/olm/main.go
```diff
@@ -106,6 +123,27 @@ func main() {
 	http.Handle("/metrics", prometheus.Handler())
 	go http.ListenAndServe(":8080", nil)
 
-	_, done := operator.Run(stopCh)
+	ready, done := operator.Run(stopCh)
+	<-ready
+
+	if *writeStatus == true {
+		existing, err := configClient.ClusterOperators().Get("operator-lifecycle-manager", metav1.GetOptions{})
```
jpeeler: I'm assuming I don't need to worry about creation, since that should already be done.

cmd/olm/main.go
```diff
@@ -106,6 +123,27 @@ func main() {
 	http.Handle("/metrics", prometheus.Handler())
 	go http.ListenAndServe(":8080", nil)
 
-	_, done := operator.Run(stopCh)
+	ready, done := operator.Run(stopCh)
+	<-ready
```
jpeeler: Is it necessary to set the status to failed if ready doesn't happen in a certain amount of time?

ecordell: I think that CVO will essentially do that for you (if ClusterStatus isn't ready after the global timeout, install failed)

manifests/0000_30_18-operatorstatus.yaml
ecordell: Hm, not sure here either. OpenShift tracks master so I think the best thing would be do the git sha if possible.

jpeeler: Would modifying "make release" be appropriate for replacing the SHA before building an image or is there somewhere else I should be looking?

cmd/olm/main.go
```diff
@@ -38,6 +43,8 @@ var (
 			"If not set, or set to the empty string (e.g. `-watchedNamespaces=\"\"`), "+
 			"olm operator will watch all namespaces in the cluster.")
 
+	writeStatus = flag.Bool("writeStatus", true, "Writes status to ClusterOperator status on OpenShift")
```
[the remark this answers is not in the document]
jpeeler: Good idea, updated.

jpeeler:
/retest
CI is really messed up.

jpeeler:
e2e tests pass locally for me and has been tested on a 4.0 cluster. The code is slightly different from before, so it's worth another look. Not much changed though.

From what I can tell, "progressing" and "failing" doesn't make too much sense for OLM. https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/clusteroperator.md

jpeeler:
Reopening in a new PR.

PR #631, opened 2018-12-18
remove crds from "all" category

juanvallejo:
cc @soltysh

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1658065

CRDs should not register into the "all" category, as that category is reserved for frequently used core resources

### DIFF
--- a/deploy/ocp/manifests/0.7.1/0000_30_02-clusterserviceversion.crd.yaml
+++ b/deploy/ocp/manifests/0.7.1/0000_30_02-clusterserviceversion.crd.yaml
@@ -16,7 +16,6 @@ spec:
     shortNames:
     - csv
     categories:
-    - all
     - olm
   additionalPrinterColumns:
   - name: Display
--- a/deploy/ocp/manifests/0.7.1/0000_30_03-installplan.crd.yaml
+++ b/deploy/ocp/manifests/0.7.1/0000_30_03-installplan.crd.yaml
@@ -21,7 +21,6 @@ spec:
     kind: InstallPlan
     listKind: InstallPlanList
     categories:
-    - all
     - olm
   additionalPrinterColumns:
   - name: CSV
--- a/deploy/ocp/manifests/0.7.1/0000_30_04-subscription.crd.yaml
+++ b/deploy/ocp/manifests/0.7.1/0000_30_04-subscription.crd.yaml
@@ -21,7 +21,6 @@ spec:
     kind: Subscription
     listKind: SubscriptionList
     categories:
-    - all
     - olm
   additionalPrinterColumns:
   - name: Package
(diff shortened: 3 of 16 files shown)


ecordell:
@juanvallejo @soltysh I left a comment [on the bug report](https://bugzilla.redhat.com/show_bug.cgi?id=1658065#c5) about this change, but I'll copy it here:

I'd like to clarify the expected behavior here.

What permissions does the user have in the namespace to delete "all"? OLM aggregates `delete` on its CRs to the standard kube `edit` role (see the [ClusterRoles here](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/manifests/0000_30_12-aggregated.clusterrole.yaml)). So if you make a RoleBinding from user user to `edit` in the namespace, you should be able to delete all of the resources there, including OLM's

What I want to figure out is:

1. Is the aggregation to `edit` not working properly?
2. Is there a different role (other than `edit`) that the user in this bug report has that we should be aggregating to so that `delete` includes OLM types?
3. Should we remove the `all` category from our types (the changes in the PR)? Is there any documentation around what goes into the `all` category? It seems useful to see the objects provided by OLM with a `get all`, right?

soltysh:
>     3\. Should we remove the `all` category from our types (the changes in the PR)? Is there any documentation around what goes into the `all` category? It seems useful to see the objects provided by OLM with a `get all`, right?

Yes, that's the only viable option, see my email. I've tried looking for `all` docs but haven't found anything useful. It's more of a tribal knowledge, although at the same time a lot of users is overusing `all` and I want to ensure we don't accidentally pick up any of the operators in that category. Also see my email about having a common category for all operators. Finally, there's nothing stopping you from creating your own category just for your operators, I see you have `olm` and I think that's a reasonable think to leave.

juanvallejo:
@ecordell 
>> Should we remove the all category from our types (the changes in the PR)? Is there any documentation around what goes into the all category? It seems useful to see the objects provided by OLM with a get all, right?

@soltysh
> Yes, that's the only viable option, see my email. I've tried looking for all docs but haven't found anything useful. It's more of a tribal knowledge, although at the same time a lot of users is overusing all and I want to ensure we don't accidentally pick up any of the operators in that category. Also see my email about having a common category for all operators. Finally, there's nothing stopping you from creating your own category just for your operators, I see you have olm and I think that's a reasonable think to leave.

Yes, to reiterate what Maciej has stated, we would encourage operator CRDs to fall under a common category, but `all` would not be it as it has special meaning in `kubectl` and `oc`.

I have [left a comment in that bugzilla](https://bugzilla.redhat.com/show_bug.cgi?id=1658065#c6) which links to the current definition of "all" that we have. Linking that doc here as well: https://github.com/kubernetes/community/blob/master/contributors/devel/kubectl-conventions.md#rules-for-extending-special-resource-alias---all

ecordell:
I think the only guideline that comes close for us is this one:

> * No namespace admin level resources (limits, quota, policy, authorization rules)

So that seems reasonable to me. Thanks for explaining everything!

Our release process is funky right now, stuck between the old way for tectonic and the new way for openshift. I made a new PR that will get these changes into openshift: #636

PR #632, opened 2018-12-18
fix(sub): don't skip syncing a subscription if there is no CSV for it

ecordell:


jpeeler:
I think pull #633 contains this fix? Might want to just close this one.

ecordell:
included in #633

PR #633, opened 2018-12-18
Requeue subscriptions on catalogsource change

ecordell:
This PR attempts to clean up several issues:

- Subscriptions shouldn't be set to `AtLatestKnown` if we haven't created a CSV
- Requeue dependent subscriptions when catalogsources change
- Add sourceNamespace to packages subscription


ecordell:
tests are passing in ci, it's just stuck cleaning up the cluster

PR #635, opened 2018-12-18
templates: Update CSV CRD to support "csvs" as a shortname

chancez:
Enables `kubectl get csvs` which is the plural of `kubectl get csv`.
I make this typo a lot, I'm sure others do too so I figured it worthy of
updating.


ecordell:
Fixed in #636

PR #637, opened 2018-12-19
fix(metrics): remove resources that aren't updated

jpeeler:
These resources are updated in the catalog operator, so a new endpoint
needs to be created to scrape them from there instead. This commit
merely removes them from being served from the olm operator.

~(So this is really just a partial fix.)~ Updated to actually serve the new metrics properly.


pkg/metrics/metrics.go
```diff
@@ -129,15 +129,16 @@ var (
...
 		},
 	)
 )
 
 func Register() {
 	prometheus.MustRegister(csvCount)
-	prometheus.MustRegister(installPlanCount)
-	prometheus.MustRegister(subscriptionCount)
-	prometheus.MustRegister(catalogSourceCount)
+	//TODO: these metrics are actually handled in the catalog operator
```
alecmerdler: Can we delete these lines instead of commenting them out?

jpeeler: You're right, I can do better.

cmd/olm/main.go
```diff
@@ -100,12 +100,11 @@ func main() {
...
 		w.WriteHeader(http.StatusOK)
 	})
-	// TODO: both of the following require vendor updates (add k8s.io/apiserver and update prometheus)
-	//healthz.InstallHandler(mux) //(less code)
-	//mux.Handle("/metrics", promhttp.Handler()) //other form is deprecated
-	http.Handle("/metrics", prometheus.Handler())
 	go http.ListenAndServe(":8080", nil)
 
+	http.Handle("/metrics", promhttp.Handler())
+	go http.ListenAndServe(":8081", nil)
```
ecordell: Also serving a nil handler here

cmd/catalog/main.go
```diff
@@ -87,6 +93,10 @@ func main() {
 		log.Panicf("error configuring operator: %s", err.Error())
 	}
 
+	http.Handle("/metrics", promhttp.Handler())
+	go http.ListenAndServe(":8081", nil)
```
ecordell: This is serving a nil handler - did you mean to serve metrics on this port?

cmd/olm/main.go
```diff
@@ -100,12 +100,11 @@ func main() {
 	http.HandleFunc("/healthz", func(w http.ResponseWriter, r *http.Request) {
 		w.WriteHeader(http.StatusOK)
 	})
-	// TODO: both of the following require vendor updates (add k8s.io/apiserver and update prometheus)
-	//healthz.InstallHandler(mux) //(less code)
-	//mux.Handle("/metrics", promhttp.Handler()) //other form is deprecated
-	http.Handle("/metrics", prometheus.Handler())
 	go http.ListenAndServe(":8080", nil)
```
ecordell: Similar issue here, though this has clearly been here a while

cmd/olm/main.go
jpeeler: Yes, nil is intentional, it should use the "DefaultServeMux".

cmd/catalog/main.go
jpeeler: This port change was intentional, though perhaps it's not the best? OLM serves healthz from port 8080, so I wanted the metrics ports to be consistent.

cmd/olm/main.go
jpeeler: See other comments, I'm nearly certain I've confirmed that healthz was being served correctly.

cmd/olm/main.go
ecordell: Gotcha - didn't know about DefaultServeMux - I think go didn't have a muxer last time I used `http`? 😂

ecordell:
@jpeeler I'm optimistic about this merging if you rebase?

jpeeler:
/skip gitlab-ci/pipeline
not sure if this will work...

PR #638, opened 2018-12-19
feat(resolver): take all subscriptions into account when resolving

ecordell:
This is a major change to the resolution algorithm but is backwards compatible in the API:

- the entire namespace of subscriptions and csvs is considered as a unit
- a new generation of subscriptions / csvs is calculated based on available updates and api providers in the catalogs
- if the new generation of subscriptions will apply succesfully, an installplan is generated
- if the installplan is approved, the update goes ahead

This means that installplans can't be manually written (they need resolved steps to be useful). These changes should inform future versions of the API, but for now we're strictly backwards-compatible.

Things left to do:

- [x] Document and explain this behavior (barebones description in docs files)
- [x] Rewrite installplan e2e tests 
- [x] Rewrite subscription unit tests
- [x] Write tests that validate resolving apiservers (only did CRDs for the current batch of tests)
- [x] Write tests that validate extra bundle objects get resolved (CRDs are part of the bundle now, but we should verify we can make other types - I think we may have errors from the `InferGVK` parts)
- [x] Remove inmem catalog code


njhale:
@evan I don't think this is actually blocked on @alecmerdler removing inmem logic from packageserver. As I recall, packageserver has its own trimmed-down implementation of `CatalogSource` configmap parsing.

PR #639, opened 2018-12-20
Consume Operator Registry from PackageManifest Server

alecmerdler:
### Description

Replace the in-memory, `ConfigMap` implementation with the gRPC interface provided by the `operator-registry` server.

Addresses https://jira.coreos.com/browse/ALM-831


pkg/package-server/provider/registry_test.go
```diff
@@ -0,0 +1,69 @@
...
+			description:    "ManySubscribers",
+		},
+	}
+	for _, test := range tests {
+		t.Run(test.description, func(t *testing.T) {
+			prov := &RegistryProvider{Operator: &queueinformer.Operator{}}
+			stopCh := make(chan struct{})
+
+			add, modify, delete, err := prov.Subscribe(test.namespace, stopCh)
+			// TODO(alecmerdler)
```
ecordell: it might be easiest to pull in the `server` from operator-registry and start it up for testing?

PR #649, opened 2019-01-04
update Service Catalog memory & cpu limits

jboyd01:
updated per some better testing and monitoring with @djwhatle


djwhatle:
LGTM, this should prevent the pod crashes under load that we've been seeing in our tests.

PR #650, opened 2019-01-08
Consume Operator Registry from PackageManifest Server

njhale:
### Description

Continuation of [639](https://github.com/operator-framework/operator-lifecycle-manager/pull/639) by @alecmerdler 

>Replace the in-memory, `ConfigMap` implementation with the gRPC interface provided by the `operator-registry` server.

Addresses https://jira.coreos.com/browse/ALM-831


pkg/package-server/provider/manifests/etcd/0.6.1/etcdcluster.crd.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
njhale: Added these for use with registry-server in unit tests.

pkg/package-server/provider/registry.go
```diff
@@ -0,0 +1,472 @@
...
+	}
+
+	return &v1alpha1.PackageManifestList{Items: pkgs}, nil
+}
+
+func (p *RegistryProvider) forward(ctx context.Context, logger *logrus.Entry, out chan v1alpha1.PackageManifest, transform forwarder.Transform) {
+	from := make(chan interface{})
+	to := make(chan interface{})
+
+	go func() {
```
ecordell: isn't this goroutine just an extra `forwarder` layer?

Couldn't you just have:

```go 
func (p *RegistryProvider) forward(ctx context.Context, logger *logrus.Entry, out chan v1alpha1.PackageManifest, transform forwarder.Transform) {
  go func() {
    p.forwarder.Forward(ctx, p.pkgTopic, out, transform)
    // ...
  }
}
```
or am I missing some other purpose?

pkg/package-server/provider/registry.go
```diff
@@ -0,0 +1,472 @@
...
+	delete := make(chan v1alpha1.PackageManifest)
+
+	p.mu.RLock()
+	defer p.mu.RUnlock()
+
+	for _, client := range p.clients {
+		if client.source.GetNamespace() == namespace || client.source.GetNamespace() == p.globalNamespace || namespace == "" {
+			logger.Debugf("found CatalogSource %s", client.source.GetName())
+			// Start up forwarders.
+			p.forward(ctx, logger, add, pkgTransform(addPkg))
```
ecordell: super nit: `pkgTransform` to me suggests that this `Transform` will mutate the events. I'd prefer a naming that suggested that it's just a filter, e.g. `addEventFilter := pkgTransform(addPkg)`

pkg/package-server/server/server.go
```diff
@@ -178,7 +179,9 @@ func (o *PackageServerOptions) Run(stopCh <-chan struct{}) error {
 		return err
 	}
 
-	sourceProvider := provider.NewInMemoryProvider(catsrcSharedIndexInformers, queueOperator, o.GlobalNamespace)
+	// TODO(alecmerdler): Use `operator-registry` as provider instead which watches `CatalogSources` and establishes gRPC connections to each
```
ecordell: I think you can removed these commented lines now?

njhale: yup

pkg/package-server/provider/registry.go
njhale: Unfortunately we need intermediate channels to convert between `interface {}` and `PackageChan`. The `FanoutForwarder` is still useful in that it ensures the 1:n relationship, but if you want I can forego the abstraction and make a `PackageChan` typed implementation of the pattern - then we wouldn't need this step.

pkg/lib/forwarder/fanout.go
```diff
@@ -0,0 +1,67 @@
...
+	// Competing consumers are okay because all stored channels are always forwarded to.
+	for {
+		select {
+		case event := <-from:
+			func() {
+				f.mu.RLock()
+				defer f.mu.RUnlock()
+				for dest, trans := range f.subscribers[from] {
+					if transformed, ok := trans(event); ok {
+						go func(d chan<- interface{}, t interface{}) {
```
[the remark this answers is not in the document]
njhale: To prevent each subscriber write from blocking one another.

pkg/package-server/provider/registry.go
ecordell: Just make a note of it (maybe just a `TODO` comment) and don't worry about it for this PR

pkg/package-server/provider/registry.go
```diff
@@ -0,0 +1,518 @@
...
+	logger.WithField("subscribers", len(subscribers)).Info("found subscribers")
+
+	var subGroup sync.WaitGroup
+	subGroup.Add(len(subscribers) * len(pkgs))
+	for sub, info := range subscribers {
+		subCtx, cancel := context.WithTimeout(info.ctx, defaultEventTimeout)
+		defer cancel()
+
+		// Signal that the subscriber's channel is being written to and is not safe to close.
+		info.wg.Add(1)
```
njhale: Some additional protection against subscriptions ending.

pkg/package-server/provider/registry.go
```diff
@@ -0,0 +1,518 @@
...
+		// Add global packages
+		globalList, err := p.List(catsrc.GetNamespace())
+		if err != nil {
+			logger.Warn("could not list global packages to generate add events")
+			return
+		}
+		pkgs = append(pkgs, globalList.Items...)
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), defaultEventTimeout)
```
njhale: Now we can set a hard timeout on watch events.

pkg/package-server/provider/registry.go
```diff
@@ -0,0 +1,533 @@
...
+	}()
+
+	pkgList, err := p.List(catsrc.GetNamespace())
+	if err != nil {
+		logger.Warn("could not list packages to generate add events")
+		return
+	}
+	pkgs := pkgList.Items
+
+	if catsrc.GetNamespace() != p.globalNamespace {
```
ecordell: Do we need this condition? Shouldn't we just always send the global packages?

ecordell: Do we need this condition? Shouldn't we just always send the global packages?

pkg/package-server/provider/registry.go
```diff
@@ -0,0 +1,533 @@
...
+}
+
+// RegistryProvider aggregates several `CatalogSources` and establishes gRPC connections to their registry servers.
+type RegistryProvider struct {
+	*queueinformer.Operator
+
+	smu             sync.RWMutex
+	cmu             sync.RWMutex
+	globalNamespace string
+	clients         []registryClient
```
ecordell: shouldn't this be a map from a sourcekey (name, namespace) to a registryclient?

pkg/package-server/provider/registry.go
```diff
@@ -0,0 +1,533 @@
...
+	if err != nil {
+		logger.Warn("could not list packages to generate events for")
+		pkgs = &v1alpha1.PackageManifestList{}
+	}
+
+	found := func() bool {
+		p.cmu.Lock()
+		defer p.cmu.Unlock()
+
+		for i, client := range p.clients {
```
ecordell: I think a map would work better here; you won't have to loop through the clients, and you can just `delete()` the key

pkg/package-server/provider/registry.go
njhale: yes - but the way the code is organized here, if the request namespace  _is_ the global one,  you don't want to end up with duplicates. I decided to add this here instead of at the List(...) level because I thought we were playing with the idea of removing the concept of global packages altogether.

PR #653, opened 2019-01-09
fix(deploy): add linux nodeselector

ecordell:
ALM-848


njhale:
Looks good. Do we need to add the nodeselector to the olm and catalog operator deployment templates as well?

ecordell:
@njhale The nodeSelector was already in the templates for those deployments, we just weren't setting the value

PR #656, opened 2019-01-09
[QIP] feat(catalog): add grpc sourcetype to CatalogSources

ecordell:
The grpc sourcetype takes a new `image` field which specifies an image which is expected to serve a registry API over grpc on port 50051.


pkg/controller/operators/catalog/operator.go
```diff
@@ -281,15 +281,29 @@ func (o *Operator) handleDeletion(obj interface{}) {
...
+
+			catsrc, ok = tombstone.Obj.(metav1.Object)
+			if !ok {
+				utilruntime.HandleError(fmt.Errorf("Tombstone contained object that is not a Namespace %#v", obj))
+				return
+			}
+		}
 	}
+	sourceKey := resolver.CatalogKey{Name: catsrc.GetName(), Namespace: catsrc.GetNamespace()}
+	func() {
```
njhale: why does this need to execute in an anonymous function if the outer function returns immediately after this (ie. No real work is being done between the lock and the return)?

pkg/controller/registry/reconciler/reconciler.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/operatorclient"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/operatorlister"
+)
+
+type RegistryReconciler interface {
+	EnsureRegistryServer(catalogSource *v1alpha1.CatalogSource) error
+}
+
+type ReconcilerReconciler interface {
```
njhale: see other comment on `ReconcilerReconciler` name.

pkg/controller/operators/catalog/operator.go
ecordell: For future changes, this keeps the unlock close to to the reason we're locking.

njhale: I understand that - was just commenting on this specific case - it makes sense to be consistent with the rest of the codebase.

PR #663, opened 2019-01-10
 [WIP] Add minimum kube version to CSV & check it against server version

dinhxuanvu:
CSV must contain minimum kube version under spec.minKubeVersion.
OLM will retrieve that info from CSV and check it against server
version to determine its capability. If not compatible, OLM will
not install operator(s).


ecordell:
@dinhxuanvu can you rebase this on master? that should re-kick the tests and also remove the gitlab required check

PR #664, opened 2019-01-10
Use `UpgradeAvailable` Subscription Status for Unapproved InstallPlans

alecmerdler:
### Description

Previously, unapproved `InstallPlans` would cause the status of a `Subscription` to just be `UpgradePending`, which doesn't indicate further action needs to be taken by an admin.

Addresses https://jira.coreos.com/browse/ALM-849


ecordell:
This looks good but, but let's get the tests right before merging.

alecmerdler:
Blocked by https://github.com/operator-framework/operator-lifecycle-manager/pull/679

ecordell:
This should be covered by #741 via conditions.

PR #666, opened 2019-01-14
chore(deploy): reference the operator-registry image built by CI

ecordell:
instead of reference quay images for openshift builds

Also removed a couple of files that I see aren't being used anymore.

We need to verify that the correct image is being picked up and replaced with CI, should be obvious though because the package server won't start if it's not swapped properly (though it will start if the swapping doesn't happen at all)


ecordell:
blocked by https://github.com/openshift/origin/pull/21810

ecordell:
replaced by #693

PR #670, opened 2019-01-15
test(catalog): add e2e test to verify configmap changes are detected

ecordell:
and recorded on the catalogsource object

This was an attempt to replicate an issue marketplace was seeing, but the test appears to pass with no additional changes.


test/e2e/catalog_e2e_test.go
```diff
@@ -97,6 +99,98 @@ func TestDefaultCatalogLoading(t *testing.T) {
...
+	crdPlural := genName("ins-")
+
+	dependentCRD := newCRD(crdPlural)
+	mainCSV := newCSV(mainPackageStable, testNamespace, "", *semver.New("0.1.0"), nil, []apiextensions.CustomResourceDefinition{dependentCRD}, mainNamedStrategy)
+	dependentCSV := newCSV(dependentPackageStable, testNamespace, "", *semver.New("0.1.0"), []apiextensions.CustomResourceDefinition{dependentCRD}, nil, dependentNamedStrategy)
+
+	c := newKubeClient(t)
+	crc := newCRClient(t)
+	defer func() {
+		require.NoError(t, crc.OperatorsV1alpha1().Subscriptions(testNamespace).DeleteCollection(&metav1.DeleteOptions{}, metav1.ListOptions{}))
```
njhale: All `Subscriptions` should be getting deleted by `defer cleaner.NotifyTestComplete(t, true)` if the test succeeds anyway.

ecordell: they should be, but I was seeing that cleaner sometimes flake. adding this helped other tests to pass so I copied it here as well.

ecordell: not needed for this test though, I'll remove

vendor/k8s.io/client-go/pkg/version/base.go
```diff
@@ -55,8 +55,8 @@ var (
 	// NOTE: The $Format strings are replaced during 'git archive' thanks to the
 	// companion .gitattributes file containing 'export-subst' in this same
 	// directory.  See also https://git-scm.com/docs/gitattributes
-	gitVersion   string = "v0.0.0-master+$Format:%h$"
-	gitCommit    string = "$Format:%H$" // sha1 from git, output of $(git rev-parse HEAD)
+	gitVersion   string = "v0.0.0-master+7d04d0e2"
```
jpeeler: This doesn't look right.

ecordell: It looks like client-go updated the way they generate version strings? What looks off to you?

jpeeler: I don't think it has changed? (Not sure which is the source of truth, but they are equal.)
https://github.com/kubernetes/client-go/blob/master/pkg/version/base.go#L58
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/client-go/pkg/version/base.go#L58

ecordell: are we seeing the result of:

> // NOTE: The $Format strings are replaced during 'git archive' thanks to the 
> // companion .gitattributes file containing 'export-subst' in this same
> // directory.  See also https://git-scm.com/docs/gitattributes

I wonder if k8s.io changed to start serving the git archive instead of the git repo?

jpeeler: Maybe so, will assume that's correct.

jpeeler: Nick's PR actually undoes the change in operator-registry, so maybe it was a mistake after all. - https://github.com/operator-framework/operator-registry/pull/19.

pkg/controller/operators/catalog/operator.go
```diff
@@ -276,19 +273,39 @@ func (o *Operator) handleDeletion(obj interface{}) {
...
+	}
+	sourceKey := resolver.CatalogKey{Name: catsrc.GetName(), Namespace: catsrc.GetNamespace()}
+	func() {
+		o.sourcesLock.Lock()
+		defer o.sourcesLock.Unlock()
+		delete(o.sources, sourceKey)
+	}()
+	o.Log.WithField("source", sourceKey).Info("removed client for deleted catalogsource")
+
+	if err := o.catSrcQueueSet.Remove(sourceKey.Name, sourceKey.Namespace); err != nil {
```
njhale: I don't think we want to try to remove the CatalogSource queue for the namespace every time we observe a CatalogSource deletion.

PR #675, opened 2019-01-18
fix resource creation/deletion after operator group config change

jpeeler:
This fixes scenario(s) where CSVs were not being copied.

Commit message:
~This changes the operator group target namespace format specifically
when watching all namespaces. Instead of only including "", the
annotation has been updated to additionally include all the namespaces
seen. The new behavior fixes syncing issues when a new namespace is
created after the initial CSVs have been copied, again only when
watching all namespaces.~
Doesn't change format, just fixes OperatorGroup managed resources.

Requeuing changes:
When namespaces are updated, matching operator groups are requeued.
When operator group annotations change, CSVs in that namspace are
requeued.

ALM-665

### DIFF
--- a/pkg/controller/operators/olm/operator.go
+++ b/pkg/controller/operators/olm/operator.go
@@ -486,47 +486,49 @@ func (a *Operator) syncClusterServiceVersion(obj interface{}) (syncError error) 
 	logger.Info("syncing CSV")
 
 	operatorGroup := a.operatorGroupForActiveCSV(logger, clusterServiceVersion)
-	if operatorGroup == nil {
-		logger.WithField("reason", "no operatorgroup found for active CSV").Info("skipping CSV resource copy to target namespaces")
-		return
-	}
-
 	if clusterServiceVersion.Status.Phase == v1alpha1.CSVPhaseFailed && clusterServiceVersion.Status.Reason == v1alpha1.CSVReasonInterOperatorGroupOwnerConflict {
 		logger.WithField("reason", clusterServiceVersion.Status.Message).Info("skipping CSV resource copy to target namespaces")
-		return
-	}
-	allNamespaces := make([]string, 0)
-	pruneNamespaces := make([]string, 0)
-	targetNamespaces := make([]string, 0)
-	namespaceObjs, err := a.lister.CoreV1().NamespaceLister().List(labels.Everything())
-	if err != nil {
-		return err
-	}
+	} else if operatorGroup != nil {
+		allNamespaces := make([]string, 0)
+		pruneNamespaces := make([]string, 0)
+		targetNamespaces := make([]string, 0)
+		namespaceObjs, err := a.lister.CoreV1().NamespaceLister().List(labels.Everything())
+		if err != nil {
+			return err
+		}
 
-	for _, ns := range namespaceObjs {
-		allNamespaces = append(allNamespaces, ns.GetName())
-	}
-	if len(operatorGroup.Status.Namespaces) == 1 && operatorGroup.Status.Namespaces[0] == corev1.NamespaceAll {
-		targetNamespaces = allNamespaces
-	} else {
-		targetNamespaces = operatorGroup.Status.Namespaces
-		pruneNamespaces = sliceCompare(allNamespaces, targetNamespaces)
-		logger.Debugf("Found namespaces to clean %v", pruneNamespaces)
-	}
+		for _, ns := range namespaceObjs {
+			allNamespaces = append(allNamespaces, ns.GetName())
+		}
+		if len(operatorGroup.Status.Namespaces) == 1 && operatorGroup.Status.Namespaces[0] == corev1.NamespaceAll {
+			targetNamespaces = allNamespaces
+		} else {
+			targetNamespaces = operatorGroup.Status.Namespaces
+			pruneNamespaces = sliceCompare(allNamespaces, targetNamespaces)
+			logger.Debugf("Found namespaces to clean %v", pruneNamespaces)
+		}
 
-	// Check if we need to do any copying / annotation for the operatorgroup
-	if err := a.copyCsvToTargetNamespace(clusterServiceVersion, operatorGroup, targetNamespaces, pruneNamespaces); err != nil {
-		logger.WithError(err).Info("couldn't copy CSV to target namespaces")
-	}
+		// This skips copying non-final phases so that it is not possible for a new CSV that has not yet
+		// been detected as intersecting to overwrite an identical CSV in a different namespace (which
+		// would ultimately potentially cause a valid CSV to get deleted).
+		if clusterServiceVersion.Status.Phase != v1alpha1.CSVPhaseSucceeded && clusterServiceVersion.Status.Phase != v1alpha1.CSVPhaseFailed {
+			logger.Debug("skipping potential CSV copy due to not yet arriving at final phase")
+		} else {
+			// Check if we need to do any copying / annotation for the operatorgroup
+			if err := a.copyCsvToTargetNamespace(clusterServiceVersion, operatorGroup, targetNamespaces, pruneNamespaces); err != nil {
+				logger.WithError(err).Info("couldn't copy CSV to target namespaces")
+			}
 
-	// Ensure operator has access to targetnamespaces
-	if err := a.ensureRBACInTargetNamespace(clusterServiceVersion, operatorGroup, pruneNamespaces); err != nil {
-		logger.WithError(err).Info("couldn't ensure RBAC in target namespaces")
-	}
+			// Ensure operator has access to targetnamespaces
+			if err := a.ensureRBACInTargetNamespace(clusterServiceVersion, operatorGroup, pruneNamespaces); err != nil {
+				logger.WithError(err).Info("couldn't ensure RBAC in target namespaces")
+			}
 
-	// Ensure cluster roles exist for using provided apis
-	if err := a.ensureClusterRolesForCSV(clusterServiceVersion, operatorGroup); err != nil {
-		logger.WithError(err).Info("couldn't ensure clusterroles for provided api types")
+			// Ensure cluster roles exist for using provided apis
+			if err := a.ensureClusterRolesForCSV(clusterServiceVersion, operatorGroup); err != nil {
+				logger.WithError(err).Info("couldn't ensure clusterroles for provided api types")
+			}
+		}
 	}
 
 	outCSV, syncError := a.transitionCSVState(*clusterServiceVersion)
@@ -540,7 +542,7 @@ func (a *Operator) syncClusterServiceVersion(obj interface{}) (syncError error) 
 	}
 
 	// Update CSV with status of transition. Log errors if we can't write them to the status.
-	_, err = a.client.OperatorsV1alpha1().ClusterServiceVersions(clusterServiceVersion.GetNamespace()).UpdateStatus(outCSV)
+	_, err := a.client.OperatorsV1alpha1().ClusterServiceVersions(clusterServiceVersion.GetNamespace()).UpdateStatus(outCSV)
 	if err != nil {
 		updateErr := errors.New("error updating ClusterServiceVersion status: " + err.Error())
 		if syncError == nil {
--- a/pkg/controller/operators/olm/operator.go
+++ b/pkg/controller/operators/olm/operator.go
@@ -495,14 +495,32 @@ func (a *Operator) syncClusterServiceVersion(obj interface{}) (syncError error) 
 		logger.WithField("reason", clusterServiceVersion.Status.Message).Info("skipping CSV resource copy to target namespaces")
 		return
 	}
+	allNamespaces := make([]string, 0)
+	pruneNamespaces := make([]string, 0)
+	targetNamespaces := make([]string, 0)
+	namespaceObjs, err := a.lister.CoreV1().NamespaceLister().List(labels.Everything())
+	if err != nil {
+		return err
+	}
+
+	for _, ns := range namespaceObjs {
+		allNamespaces = append(allNamespaces, ns.GetName())
+	}
+	if len(operatorGroup.Status.Namespaces) == 1 && operatorGroup.Status.Namespaces[0] == corev1.NamespaceAll {
+		targetNamespaces = allNamespaces
+	} else {
+		targetNamespaces = operatorGroup.Status.Namespaces
+		pruneNamespaces = sliceCompare(allNamespaces, targetNamespaces)
+		logger.Debugf("Found namespaces to clean %v", pruneNamespaces)
+	}
 
 	// Check if we need to do any copying / annotation for the operatorgroup
-	if err := a.copyCsvToTargetNamespace(clusterServiceVersion, operatorGroup); err != nil {
+	if err := a.copyCsvToTargetNamespace(clusterServiceVersion, operatorGroup, targetNamespaces, pruneNamespaces); err != nil {
 		logger.WithError(err).Info("couldn't copy CSV to target namespaces")
 	}
 
 	// Ensure operator has access to targetnamespaces
-	if err := a.ensureRBACInTargetNamespace(clusterServiceVersion, operatorGroup); err != nil {
+	if err := a.ensureRBACInTargetNamespace(clusterServiceVersion, operatorGroup, pruneNamespaces); err != nil {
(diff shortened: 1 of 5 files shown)


jpeeler:
/retest
And so it begins.

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -411,9 +435,12 @@ func (a *Operator) updateNamespaceList(op *v1alpha2.OperatorGroup) ([]string, er
 			}
 			namespaceSet[ns] = struct{}{}
 		}
-	} else if selector == nil || selector.Empty() {
-		namespaceSet[corev1.NamespaceAll] = struct{}{}
 	} else {
+		if selector == nil || selector.Empty() {
+			selector = labels.Everything()
+			// set is sorted after returning, so being at the beginning can be relied upon
```
ecordell: not sure I follow this comment - where are we using order of the set? (asking because maps don't retain any ordering)

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -58,16 +58,25 @@ func (a *Operator) syncOperatorGroups(obj interface{}) error {
 	a.Log.Debug("Cluster roles completed")
 
 	for _, csv := range a.csvSet(op.Namespace, v1alpha1.CSVPhaseAny) {
-		origCSVannotations := csv.GetAnnotations()
+		origCSVannotations := a.copyOperatorGroupAnnotations(&csv.ObjectMeta)
```
ecordell: why copy out the just the three annotations and then immediately set just those three below with `addOperatorGroupAnnotations`? If we aren't going to retain the other annotation values, then we can just directly build the set of annotations here and not use these `copyOperatorGroupAnnotations`/`addOperatorGroupAnnotations` methods?

pkg/controller/operators/olm/operator.go
```diff
@@ -310,6 +314,32 @@ func (a *Operator) syncObject(obj interface{}) (syncError error) {
 		a.requeueOwnerCSVs(metaObj)
 	}
 
+	namespace, ok := obj.(*corev1.Namespace)
```
ecordell: I think this may be too aggressive for resyncing.

The cases where we know we need to resync:
 - namespace is added/deleted
 - namespace's labels updated

This makes me think we may want to detect those cases directly by passing in specific handlers for those events.

pkg/controller/operators/olm/operator.go
```diff
@@ -310,6 +314,32 @@ func (a *Operator) syncObject(obj interface{}) (syncError error) {
...
+	if ok {
+		// Check to see if any operator groups are associated with this namespace
+		operatorGroupList, err := a.lister.OperatorsV1alpha2().OperatorGroupLister().OperatorGroups(metav1.NamespaceAll).List(labels.Everything())
+		if err != nil {
+			syncError = fmt.Errorf("lister failed: %v", err)
+			logger.Warn(syncError.Error())
+			return
+		}
+		for _, opGroup := range operatorGroupList {
+			namespaceMap, err := a.getMatchingNamespaces(opGroup)
```
ecordell: We shouldn't need to list all of the namespaces of the operator group to know if the one we're currently syncing should requeue an operatorgroup? Just need to compare the namespace list (in spec of opgroup) or the label selector on the operatorgroup / labels of the namespace.

We probably already have a function somewhere that does this, but don't we just need to do something like `if ShouldContain(opGroup, namespace) { resync opgroup }`?

pkg/api/apis/operators/v1alpha1/clusterserviceversion.go
```diff
@@ -98,7 +98,7 @@ func NewInstallModeSet(modes []InstallMode) (InstallModeSet, error) {
 // the given operatorNamespace and list of target namespaces.
 func (set InstallModeSet) Supports(operatorNamespace string, namespaces []string) error {
 	numNamespaces := len(namespaces)
-	if !set[InstallModeTypeAllNamespaces] && numNamespaces == 1 && namespaces[0] == v1.NamespaceAll {
+	if !set[InstallModeTypeAllNamespaces] && numNamespaces > 0 && namespaces[0] == v1.NamespaceAll {
```
ecordell: Can you explain this change? Why does it help to list additional namespaces for `NamespaceAll`?

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -293,7 +302,7 @@ func (a *Operator) ensureTenantRBAC(operatorNamespace, targetNamespace string, c
 
 func (a *Operator) copyCsvToTargetNamespace(csv *v1alpha1.ClusterServiceVersion, operatorGroup *v1alpha2.OperatorGroup) error {
 	namespaces := make([]string, 0)
-	if len(operatorGroup.Status.Namespaces) == 1 && operatorGroup.Status.Namespaces[0] == corev1.NamespaceAll {
+	if operatorGroup.Status.Namespaces[0] == corev1.NamespaceAll {
```
ecordell: It would be nice if we didn't have to implicitly rely on list ordering here

pkg/controller/operators/olm/operatorgroup.go
jpeeler: Right, I should have specifically said slice instead. It's after being converted from a set to a slice in updateNamespaceList.

pkg/controller/operators/olm/operatorgroup.go
jpeeler: The idea here wasn't to change the logic, but to fix the bug of not detecting annotation changes due to getting a reference instead of a copy. The new `copyOperatorGroupAnnotations` method ensures a copy is done so that the reflection test below works properly.

pkg/controller/operators/olm/operator.go
jpeeler: If we're looking at all three add/update/delete, is having separate handlers for each better? Or perhaps you're referring to something more granular that I don't know about?

pkg/controller/operators/olm/operator.go
jpeeler: The resyncing here is specifically to make an operator group status contain a properly updated namespace list. You are right though that less work is necessary here, so I'll look into fixing that.

pkg/api/apis/operators/v1alpha1/clusterserviceversion.go
jpeeler: The exact scenario that is solved by doing this is after an operator group's status is written that is watching all namespaces, a new namespace creation isn't "noticed" without specifically listing all the namespaces seen.

pkg/controller/operators/olm/operatorgroup.go
jpeeler: I thought this was better than searching the string each time to see if "" was in the list. I couldn't figure out any naming issues with sorting not causing "" to be first.

pkg/controller/operators/olm/operator.go
ecordell: the thing I really want to omit from processing is the case where the namespace is synced periodically, or when there has been some non-label-related change to a namespace

pkg/controller/operators/olm/operator.go
jpeeler: I take back what I said about less work being necessary. I think it should be left as is.

pkg/controller/operators/olm/operator.go
jpeeler: If no namespace processing is done then the scenario I described in https://github.com/operator-framework/operator-lifecycle-manager/pull/675#discussion_r249543359 will only eventually work, not instantly as I was hoping could be the case. Is that acceptable?

pkg/api/apis/operators/v1alpha1/clusterserviceversion.go
njhale: When an operator group is selecting all namespaces, the set of namespaces in its status should be of length = 1 and the first and only element should be `NamespaceAll` (""). There should never be a situation where length > 1 and the first element is "".

pkg/controller/operators/olm/operator.go
```diff
@@ -75,6 +77,7 @@ func NewOperator(logger *logrus.Logger, crClient versioned.Interface, opClient o
 	op := &Operator{
 		Operator:    queueOperator,
 		csvQueueSet: make(map[string]workqueue.RateLimitingInterface),
+		opQueueSet:  make(map[string]workqueue.RateLimitingInterface),
```
njhale: 👍

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -154,7 +163,7 @@ func (a *Operator) ensureRBACInTargetNamespace(csv *v1alpha1.ClusterServiceVersi
 	}
 
 	// if OperatorGroup is global (all namespaces) we generate cluster roles / cluster role bindings instead
-	if len(targetNamespaces) == 1 && targetNamespaces[0] == corev1.NamespaceAll {
+	if targetNamespaces[0] == corev1.NamespaceAll {
```
njhale: should always be length == 1

pkg/controller/operators/olm/operatorgroup.go
njhale: see above comments

jpeeler:
I've reworked this to NOT change the namespace status format in the case of watching all namespaces. I've verified that it works in the additional namespace scenario I've been testing with. However, it does not properly remove a CSV if a namespace is removed from an existing operator group (which I assume does not work properly with the current code either). Perhaps I should add these scenarios as e2es.

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -282,6 +282,16 @@ func (a *Operator) ensureTenantRBAC(operatorNamespace, targetNamespace string, c
 				return err
 			}
 		}
+		for _, ns := range pruneNamespaces {
```
jpeeler: These RBAC changes and the ones below seem to be unnecessary. Why would that be?

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -305,22 +315,37 @@ func (a *Operator) ensureTenantRBAC(operatorNamespace, targetNamespace string, c
...
 }
 
-func (a *Operator) copyCsvToTargetNamespace(csv *v1alpha1.ClusterServiceVersion, operatorGroup *v1alpha2.OperatorGroup) error {
-	namespaces := make([]string, 0)
-	if len(operatorGroup.Status.Namespaces) == 1 && operatorGroup.Status.Namespaces[0] == corev1.NamespaceAll {
-		namespaceObjs, err := a.lister.CoreV1().NamespaceLister().List(labels.Everything())
+func (a *Operator) copyCsvToTargetNamespace(csv *v1alpha1.ClusterServiceVersion, operatorGroup *v1alpha2.OperatorGroup, namespaces []string, pruneNamespaces []string) error {
+	// check for stale CSVs from a different previous operator group configuration
+	for _, ns := range pruneNamespaces {
+		// QUESTION: is there a way to query on annotation?
```
jpeeler: Is there?

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -413,6 +438,22 @@ func (a *Operator) copyOperatorGroupAnnotations(obj *metav1.ObjectMeta) map[stri
 	return copiedAnnotations
 }
 
+// returns items in a that are not in b
+// TODO: combine with namespacesChanged?
```
jpeeler: Maybe the code repetition does not matter for such small functions?

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -58,16 +58,34 @@ func (a *Operator) syncOperatorGroups(obj interface{}) error {
 	a.Log.Debug("Cluster roles completed")
 
 	for _, csv := range a.csvSet(op.Namespace, v1alpha1.CSVPhaseAny) {
-		origCSVannotations := csv.GetAnnotations()
+		origCSVannotations := a.copyOperatorGroupAnnotations(&csv.ObjectMeta)
 		a.addOperatorGroupAnnotations(&csv.ObjectMeta, op, !csv.IsCopied())
 		if reflect.DeepEqual(origCSVannotations, csv.GetAnnotations()) == false {
```
ecordell: Will this do the right thing if there are other annotations on the csv? (Non-operatorgrou-related)

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -287,22 +315,37 @@ func (a *Operator) ensureTenantRBAC(operatorNamespace, targetNamespace string, c
...
+	// check for stale CSVs from a different previous operator group configuration
+	for _, ns := range pruneNamespaces {
+		// QUESTION: is there a way to query on annotation?
+		fetchedCSVs, err := a.lister.OperatorsV1alpha1().ClusterServiceVersionLister().ClusterServiceVersions(ns).List(labels.Everything())
 		if err != nil {
 			return err
 		}
-		for _, ns := range namespaceObjs {
-			namespaces = append(namespaces, ns.GetName())
+		for _, csv := range fetchedCSVs {
```
ecordell: I think we should also verify that the CSV status is “copied” just to be safe

pkg/controller/operators/olm/operatorgroup.go
jpeeler: Ah right, I need to make a getter. The intention was to really only test operator group annotations.

alecmerdler:
@jpeeler What does this PR still need before I give it a review?

jpeeler:
@alecmerdler sorry missed your comment. just pushed a new revision, should be good to go at this point.

pkg/controller/operators/olm/operator.go
```diff
@@ -310,6 +314,34 @@ func (a *Operator) syncObject(obj interface{}) (syncError error) {
 		a.requeueOwnerCSVs(metaObj)
 	}
 
+	namespace, ok := obj.(*corev1.Namespace)
+	if ok {
```
njhale: nit: if you invert this to:
```go
if !ok {
    return nil
}
```
You should be able to un-indent the logic below.

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -58,16 +58,34 @@ func (a *Operator) syncOperatorGroups(obj interface{}) error {
...
-		origCSVannotations := csv.GetAnnotations()
+		origCSVannotations := a.copyOperatorGroupAnnotations(&csv.ObjectMeta)
 		a.addOperatorGroupAnnotations(&csv.ObjectMeta, op, !csv.IsCopied())
 		if reflect.DeepEqual(origCSVannotations, csv.GetAnnotations()) == false {
 			// CRDs don't support strategic merge patching, but in the future if they do this should be updated to patch
 			if _, err := a.client.OperatorsV1alpha1().ClusterServiceVersions(csv.GetNamespace()).Update(csv); err != nil {
 				a.Log.Errorf("Update for existing CSV failed: %v", err)
+			} else {
+				if _, ok := origCSVannotations[v1alpha2.OperatorGroupAnnotationKey]; ok {
+					if err := a.csvQueueSet.Requeue(csv.GetName(), csv.GetNamespace()); err != nil {
```
njhale: Updating `ObjectMeta` on the CSV will enqueue it. I don't think you need to explicitly requeue.

jpeeler: This specifically handles the case when an annotation hasn't occurred, but CSVs still need to be copied to newly created namespaces.

pkg/controller/operators/olm/operator.go
jpeeler: Fixed this.

jpeeler:
I believe the e2e issue I saw earlier has been fixed now (here's for hoping my local success is replicated). Not removing the hold though until I look over https://github.com/operator-framework/operator-lifecycle-manager/pull/701.

jpeeler:
Please focus attention on syncClusterServiceVersion as the CSV copying behavior has been (necessarily) changed due to the rebase changes.

jpeeler:
This needs to be rebased and we talked about instead of only copying CSVs when they enter the final state (in the last commit) to add validation beforehand instead.

sferich888:
@jpeeler can we close this in favor of https://github.com/operator-framework/operator-lifecycle-manager/pull/736

PR #676, opened 2019-01-19
fix(subscriptions): respect startingCSV

njhale:
Fixes `Subscription` resolution to respect the `StartingCSV` field.

Addresses [ALM-851](https://jira.coreos.com/browse/ALM-851).


pkg/controller/registry/resolver/querier_test.go
```diff
@@ -239,7 +239,7 @@ func TestNamespaceSourceQuerier_FindPackage(t *testing.T) {
 			q := &NamespaceSourceQuerier{
 				sources: tt.fields.sources,
 			}
-			got, key, err := q.FindPackage(tt.args.pkgName, tt.args.channelName, tt.args.initialSource)
+			got, key, err := q.FindPackage(tt.args.pkgName, tt.args.channelName, "", tt.args.initialSource)
 			require.Equal(t, tt.out.bundle, got)
```
ecordell: Unit tests for new querier behavior?

test/e2e/subscription_e2e_test.go
```diff
@@ -489,3 +490,79 @@ func TestCreateNewSubscriptionManualApproval(t *testing.T) {
...
+	installPlanName := subscription.Status.Install.Name
+
+	// Wait for InstallPlan to be status: Complete before checking resource presence
+	fetchedInstallPlan, err := fetchInstallPlan(t, crc, installPlanName, buildInstallPlanPhaseCheckFunc(v1alpha1.InstallPlanPhaseComplete, v1alpha1.InstallPlanPhaseFailed))
+	require.NoError(t, err)
+	t.Logf("Install plan %s fetched with status %s", fetchedInstallPlan.GetName(), fetchedInstallPlan.Status.Phase)
+	require.Equal(t, v1alpha1.InstallPlanPhaseComplete, fetchedInstallPlan.Status.Phase)
+
+	// Wait for csvA to be found
+	fetchedCSV, err := fetchCSV(t, crc, csvA.GetName(), testNamespace, buildCSVConditionChecker(v1alpha1.CSVPhaseSucceeded, v1alpha1.CSVPhaseFailed))
```
ecordell: This could be racy, right? What if by the time this happens, we’ve synced again and updated from csvA to csvB?

Maybe we do an manual approval?

njhale: You're totally right - I just pushed up what I had locally. I think a manual approval might work.

pkg/controller/registry/resolver/querier_test.go
njhale: Will do.

pkg/controller/operators/catalog/operator.go
```diff
@@ -530,14 +530,11 @@ func (o *Operator) syncResolvingNamespace(obj interface{}) error {
 	// get the set of sources that should be used for resolution and best-effort get their connections working
 	logger.Debug("resolving sources")
 	resolverSources := o.ensureResolverSources(logger, namespace)
+	querier := resolver.NewNamespaceSourceQuerier(resolverSources)
 
 	logger.Debug("checking if subscriptions need update")
 
 	subs, err := o.lister.OperatorsV1alpha1().SubscriptionLister().Subscriptions(namespace).List(labels.Everything())
-	if err != nil {
```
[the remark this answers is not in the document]
njhale: Good catch - probably happened during the rebase.

pkg/controller/registry/resolver/resolver.go
```diff
@@ -112,7 +111,6 @@ func (r *OperatorsV1alpha1Resolver) ResolveSteps(namespace string, sourceQuerier
 		if subExists && existingSubscription.Status.CurrentCSV != op.Identifier() {
 			existingSubscription.Status.CurrentCSV = op.Identifier()
 			updatedSubs = append(updatedSubs, existingSubscription)
-			continue
```
ecordell: Why are these removed?

njhale: I didn't (and doesn't) look like this `continue` is skipping any logic between here and the next iteration of the loop.

PR #679, opened 2019-01-21
fix(installplan): fix bug where too many installplans can be created

ecordell:
This moves subscription resolution into a namespace queue so that OLM never competes with itself to resolve.


pkg/controller/operators/catalog/operator.go
```diff
@@ -56,16 +56,17 @@ var timeNow = func() metav1.Time { return metav1.NewTime(time.Now().UTC()) }
...
+	client                versioned.Interface
+	lister                operatorlister.OperatorLister
+	namespace             string
+	sources               map[resolver.CatalogKey]resolver.SourceRef
+	sourcesLock           sync.RWMutex
+	sourcesLastUpdate     metav1.Time
+	resolver              resolver.Resolver
+	subQueue              workqueue.RateLimitingInterface
+	catSrcQueueSet        queueinformer.ResourceQueueSet
+	namespaceResolveQueue workqueue.RateLimitingInterface
```
alecmerdler: Fix the unit test panic by adding this to the test setup in `subscriptions_test.go`:
```go
o.namespaceResolveQueue = workqueue.NewRateLimitingQueue(workqueue.DefaultControllerRateLimiter())
```

pkg/controller/operators/catalog/operator.go
```diff
@@ -533,20 +595,35 @@ func (o *Operator) syncSubscriptions(obj interface{}) error {
...
 }
 
+func (o *Operator) syncSubscriptions(obj interface{}) error {
+	sub, ok := obj.(*v1alpha1.Subscription)
+	if !ok {
+		o.Log.Debugf("wrong type: %#v", obj)
+		return fmt.Errorf("casting Subscription failed")
+	}
+
+	o.resolveNamespace(sub.GetNamespace())
```
alecmerdler: Unit test needs to be changed from calling `syncSubscription()` to `syncResolvingNamespace()` since the logic was moved to that method.

pkg/controller/operators/catalog/operator.go
```diff
@@ -209,9 +210,43 @@ func NewOperator(kubeconfigPath string, logger *logrus.Logger, wakeupInterval ti
...
+		// Register queue and QueueInformer
+		queueName := fmt.Sprintf("%s/operatorgroups", namespace)
+		operatorGroupQueue := workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), queueName)
+		operatorGroupQueueInformer := queueinformer.NewInformer(operatorGroupQueue, operatorGroupInformer.Informer(), NoopHandler, nil, queueName, metrics.NewMetricsNil(), logger)
+		op.RegisterQueueInformer(operatorGroupQueueInformer)
+	}
 	return op, nil
 }
 
+func NoopHandler(obj interface{}) error {
```
alecmerdler: Could just use an anonymous function instead since this is only used once.

pkg/controller/operators/catalog/operator.go
ecordell: Just pushed up that change

pkg/controller/operators/catalog/operator.go
ecordell: I kept this around in the git history but I removed the gate on operatorgroup. It didn't end up being related to the fix in this PR (added it to verify it wasn't an issue originally), and I don't think it provides the best UX for that scenario (an error message in a log)

pkg/controller/operators/catalog/operator.go
```diff
@@ -487,36 +506,59 @@ func (o *Operator) syncDependentSubscriptions(logger *logrus.Entry, catalogSourc
...
+		if err != nil {
+			return err
+		}
+
+		// record the current state of the desired corresponding CSV in the status. no-op if we don't know the csv yet.
+		sub, err = o.ensureSubscriptionCSVState(logger, sub)
+		if err != nil {
+			return err
+		}
+		shouldUpdate = shouldUpdate || !o.nothingToUpdate(logger, sub)
```
njhale: nit: I think this can just be `!o.nothingToUpdate(...)`

pkg/controller/operators/catalog/operator.go
```diff
@@ -606,6 +663,44 @@ func (o *Operator) nothingToUpdate(logger *logrus.Entry, sub *v1alpha1.Subscript
...
+func (o *Operator) ensureSubscriptionInstallPlanState(logger *logrus.Entry, sub *v1alpha1.Subscription) (*v1alpha1.Subscription, error) {
+	if sub.Status.Install != nil {
+		return sub, nil
+	}
+
+	logger.Debug("checking for existing installplan")
+
+	// check if there's an installplan that created this subscription (only if it doesn't have a reference yet)
+	// this indicates it was newly resolved by another operator, and we should reference that installplan in the status
+	ips, err := o.client.OperatorsV1alpha1().InstallPlans(sub.GetNamespace()).List(metav1.ListOptions{})
```
njhale: nit: no lister?

ecordell: there wasn't an installplan lister to use, I forgot to go back and add one. I'll update.

pkg/controller/operators/catalog/operator.go
ecordell: if we do `shouldUpdate = !nothingToUpdate`, then a later iteration can "reset" `shouldUpdate` to `false`.

njhale: Ah, I see - missed that this was in a loop. Looks good.

ecordell:
@njhale addressed all comments

PR #681, opened 2019-01-23
Ensure Owner References on ConfigMaps for CatalogSources 

alecmerdler:
### Description

The `CatalogSource` sync handler now ensures that it's backing `ConfigMap` contains an `ownerReference`. This has the useful side effect of causing a resync when the `ConfigMap` is updated.

Addresses https://jira.coreos.com/browse/ALM-878


pkg/controller/operators/catalog/operator.go
```diff
@@ -343,8 +343,9 @@ func (o *Operator) syncCatalogSources(obj interface{}) (syncError error) {
 				ResourceVersion: configMap.GetResourceVersion(),
 			}
 			out.Status.LastSync = timeNow()
-
-			// update status
+			if wasOwned := ownerutil.EnsureOwner(configMap, catsrc); !wasOwned {
```
ecordell: Can you add a quick check (just in one of the existing e2e tests) that configmaps get adopted? (I think the ones we generate for testing get created with owner refs, so we can just remove them to test this)

alecmerdler: Looks like we [already had a test for this](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/test/e2e/catalog_e2e_test.go#L200)?

PR #682, opened 2019-01-23
fix(metrics): add service monitor config

jpeeler:
Please review @openshift/sig-monitoring. <-- I guess that doesn't work since this repo isn't in the openshift namespace?

There are no alerts as the OLM metrics are mostly informational. ~The metrics are reported without TLS, that's ok right?~ (Edit: metrics are served over TLS.) And lastly, just in case the intent is not clear the goal is to scrape two separate endpoints.

(This will need a make release done after final review I believe.)


deploy/chart/templates/0000_50_18-service-monitor.yaml
```diff
@@ -0,0 +1,102 @@
...
+- apiVersion: v1
+  kind: Service
+  metadata:
+    name: catalog-operator
+    namespace: {{ .Values.namespace }}
+    labels:
+      app: catalog-operator
+  spec:
+    ports:
+    - name: http-metrics2
```
ecordell: we don't need to disambiguate since the endpoints are defined in different services. you don't need the name field at all

deploy/chart/templates/0000_50_18-service-monitor.yaml
```diff
@@ -0,0 +1,102 @@
...
+      port: 8081
+      protocol: TCP
+      targetPort: http
+    selector:
+      app: catalog-operator
+---
+- apiVersion: rbac.authorization.k8s.io/v1
```
ecordell: this probably shouldn't be a list?

deploy/chart/templates/0000_50_18-service-monitor.yaml
```diff
@@ -0,0 +1,102 @@
...
+    name: operator-lifecycle-manager-metrics
+  roleRef:
+    apiGroup: rbac.authorization.k8s.io
+    kind: ClusterRole
+    name: operator-lifecycle-manager-metrics
+  subjects:
+  - kind: ServiceAccount
+    name: prometheus-k8s
+    namespace: {{ .Values.namespace }}
+---
```
ecordell: Is something missing here? I don't see the `ClusterRole` that the CRB binds to, and I don't see anything binding to the role. I assume `prometheus-k8s` serviceaccount already exists?

jpeeler: I'm really not sure what this service account should be, but I was assuming this service account existed.

ecordell: Where did the `ClusterRoleBinding` and `Role` come from? Are there docs on integrating with cluster-monitoring? I would have expected rbac pairs of (role, rolebinding), (clusterrole, clusterrolebinding), or (clusterrole, rolebinding), but not (role, clusterrolebinding) - it's not clear to me how these would be used

mxinden:
> The metrics are reported without TLS, that's ok right?

We are trying to secure all metrics endpoints in the monitoring stack. You can find two examples similar to your use case in the [node-exporter](https://github.com/openshift/cluster-monitoring-operator/tree/master/assets/node-exporter) and [kube-state-metrics](https://github.com/openshift/cluster-monitoring-operator/tree/master/assets/kube-state-metrics) manifests. I am not sure this would be a show-stopper though.

deploy/chart/templates/0000_50_18-service-monitor.yaml
```diff
@@ -0,0 +1,102 @@
...
+  namespace: {{ .Values.namespace }}
+  labels:
+    app: olm-operator
+spec:
+  jobLabel: k8s-app
+  endpoints:
+  - port: http-metrics
+    interval: 30s
+    scheme: http
+    bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
```
mxinden: Sending bearer tokens un-encrypted over the wire sounds a bit scary to me. Given that this would be the Prometheus service account token, one could gain quite a bit, impersonating it. What do you think?

deploy/chart/templates/0000_50_18-service-monitor.yaml
jpeeler: There's an internal doc (sorry if anybody else is reading along): https://docs.google.com/document/d/1rCKAYTrYMESjJDyJ0KvNap05NNukmNXVVi6MShISnOw/edit#. I didn't mean to mix cluster/role pairs. I was basing the RBAC permissions off of https://github.com/openshift/cluster-openshift-controller-manager-operator/tree/master/manifests (and didn't see a service account created there).

deploy/chart/templates/0000_50_18-service-monitor.yaml
jpeeler: If this isn't required for http scraping, then I'll certainly remove it.

ecordell:
@jpeeler I think this just needs a rebase and it's good to go, right?

deploy/chart/templates/0000_50_18-service-monitor.yaml
mxinden: @jpeeler whether it is necessary or not depends on how you secure your metrics endpoint on the operator side.

What I was referring to with my comment above was, that you are using `scheme: http`, thereby sending the bearer token in the line below in clear text over the wire. Given that one can impersonate Prometheus once they captures the token, this seems insecure to me.

deploy/chart/templates/0000_50_olm_00-namespace.yaml
```diff
@@ -4,6 +4,7 @@ metadata:
   name: {{ .Values.namespace }}
   labels:
     openshift.io/run-level: "1"
+    openshift.io/cluster-monitoring: "true"
```
ecordell: I don't want to block the PR on this, but these labels probably shouldn't go in upstream releases

deploy/chart/templates/0000_50_olm_02-services.yaml
```diff
@@ -0,0 +1,37 @@
+apiVersion: v1
```
ecordell: Shouldn't these have `ifs` so that they don't appear in upstream charts?

deploy/chart/templates/0000_50_olm_08-catalog-operator.deployment.yaml
```diff
@@ -51,6 +62,14 @@ spec:
           resources:
 {{ toYaml .Values.catalog.resources | indent 12 }}
           {{- end}}
+          volumeMounts:
```
ecordell: Cert mount shouldn't be in upstream

deploy/chart/templates/0000_50_olm_18-upstream-operators.catalogsource.yaml
```diff
@@ -1,3 +1,5 @@
+#! validate-crd: ./deploy/chart/templates/06-catalogsource.crd.yaml
+#! parse-kind: CatalogSource
```
ecordell: I seem to remember these caused problems with the release artifacts, so we removed them.

deploy/chart/templates/0000_50_olm_19-service-monitor.yaml
```diff
@@ -0,0 +1,88 @@
+apiVersion: monitoring.coreos.com/v1
```
ecordell: Don't include upstream :)

deploy/chart/templates/0000_50_olm_00-namespace.yaml
jpeeler: fixed!

deploy/chart/templates/0000_50_olm_02-services.yaml
jpeeler: fixed!

deploy/chart/templates/0000_50_olm_08-catalog-operator.deployment.yaml
jpeeler: fixed!

deploy/chart/templates/0000_50_olm_18-upstream-operators.catalogsource.yaml
jpeeler: As far as I understand it, these lines will get stripped when a release is made. These lines were already there.

deploy/chart/templates/0000_50_olm_19-service-monitor.yaml
jpeeler: fixed!

jpeeler:
I really hate it when github doesn't display the commit order in the order I've specified and uses date order.
If you review by commit:
https://github.com/operator-framework/operator-lifecycle-manager/pull/682/commits/b60eb070b395a65d3b00e3f87ea4ffd0009d9f0e
https://github.com/operator-framework/operator-lifecycle-manager/pull/682/commits/57d01960c6df2a205507f4e8a53e837989aa54a7
https://github.com/operator-framework/operator-lifecycle-manager/pull/682/commits/0887aa7d876109c69cc172d445a9413a0a93d9db
https://github.com/operator-framework/operator-lifecycle-manager/pull/682/commits/cee4a2831b8f84e10276c1b4569036d48d341d6e
https://github.com/operator-framework/operator-lifecycle-manager/pull/682/commits/91003b6c5d4734299165a3d3be1af9afaaec2244
https://github.com/operator-framework/operator-lifecycle-manager/pull/682/commits/f097e51350d3f87967d774976d3c84419c39a756

deploy/chart/templates/0000_50_18-service-monitor.yaml
jpeeler: I've switched everything over to https.

ecordell:
/lgtm

Looks good! Thanks!

PR #686, opened 2019-01-24
fix(subscription): don't adopt subscriptions that have installplan refs

ecordell:
Looks like a small bug here - shouldn't really affect anything but better safe than sorry


njhale:
Superseded by [#698](https://github.com/operator-framework/operator-lifecycle-manager/pull/689).

PR #687, opened 2019-01-25
fix(permissions): Generate unique Names for permissions

ecordell:
Fixes an issue where role and cluster role names can conflict in a global operatorgroup


pkg/controller/registry/resolver/rbac.go
```diff
@@ -6,6 +6,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apiserver/pkg/storage/names"
```
alecmerdler: Cool, we've used this package in our E2E tests so it's probably good.

PR #688, opened 2019-01-27
fix(reconciler): set command in pod spec of registry images

ecordell:
This was a part of #666 but I'm seeing weird issues with the CI, so I split it out into its own PR.


pkg/controller/operators/catalog/operator.go
```diff
@@ -564,28 +580,37 @@ func (o *Operator) syncResolvingNamespace(obj interface{}) error {
...
+		if err := o.updateSubscriptionStatus(namespace, updatedSubs, installPlanReference); err != nil {
+			logger.WithError(err).Debug("error ensuring subscription installplan state")
+			return err
+		}
+		return nil
 	}
 
-	if err := o.updateSubscriptionSetInstallPlanState(namespace, subs, installplanReference); err != nil {
-		logger.WithError(err).Debug("error ensuring subscription installplan state")
+	if err := o.updateSubscriptionStatus(namespace, subs, nil); err != nil {
```
njhale: What case does this cover? It's not immediately apparent to me.

ecordell: this is ensuring that subscriptions get their last update time updated whenever we've fully processed them, so that on a resync we see the time as newer and don't process when we don't need to

PR #690, opened 2019-01-28
fix(olm): Fix CSVs api-servers battle for ownership of APIServices

dinhxuanvu:
Currently, CSVs that install extension api-servers battle for
ownership of APIServices. New CSVs only adopt APIServices that have
OwnerReferences to CSVs in the new CSV's replaces chain.

### DIFF
--- a/pkg/controller/operators/olm/apiservices.go
+++ b/pkg/controller/operators/olm/apiservices.go
@@ -83,7 +83,7 @@ func (a *Operator) checkAPIServiceResources(csv *v1alpha1.ClusterServiceVersion,
 		}
 
 		// Check if the APIService is adoptable
-		if !ownerutil.OwnersIntersect(owners, apiService.GetLabels()) {
+		if !ownerutil.OwnersIntersect(owners, apiService) {
 			err := olmerrors.NewUnadoptableError("", apiServiceName)
 			logger.WithError(err).Warn("found unadoptable apiservice")
 			errs = append(errs, err)
@@ -695,7 +695,7 @@ func (a *Operator) installAPIServiceRequirements(desc v1alpha1.APIServiceDescrip
 			owners = append(owners, replaces)
 		}
 		// check if the APIService is adoptable
-		if !ownerutil.OwnersIntersect(owners, apiService.GetLabels()) {
+		if !ownerutil.OwnersIntersect(owners, apiService) {
 			return nil, fmt.Errorf("pre-existing APIService %s is not adoptable", apiServiceName)
 		}
 	}
--- a/pkg/controller/operators/olm/apiservices.go
+++ b/pkg/controller/operators/olm/apiservices.go
@@ -16,6 +16,7 @@ import (
 
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/certs"
+	olmerrors "github.com/operator-framework/operator-lifecycle-manager/pkg/controller/errors"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/install"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/ownerutil"
 )
@@ -40,8 +41,31 @@ func (a *Operator) shouldRotateCerts(csv *v1alpha1.ClusterServiceVersion) bool {
 	return false
 }
 
+// apiServiceResourceErrorsActionable returns true if OLM can do something about any one
+// of the apiService errors in errs; otherwise returns false
+//
+// This method can be used to determine if a CSV in a failed state due to APIService
+// issues can resolve them by reinstalling
+func (a *Operator) apiServiceResourceErrorsActionable(errs []error) bool {
+	for _, err := range errs {
+		switch err.(type) {
+		case olmerrors.UnadoptableError:
+			return false
+		}
+	}
+
+	return true
+}
+
 // checkAPIServiceResources checks if all expected generated resources for the given APIService exist
-func (a *Operator) checkAPIServiceResources(csv *v1alpha1.ClusterServiceVersion, hashFunc certs.PEMHash) error {
+func (a *Operator) checkAPIServiceResources(csv *v1alpha1.ClusterServiceVersion, hashFunc certs.PEMHash) []error {
+	errs := []error{}
+	owners := []ownerutil.Owner{csv}
+	// Get replacing CSV if exists
+	replacement, _ := a.lister.OperatorsV1alpha1().ClusterServiceVersionLister().ClusterServiceVersions(csv.GetNamespace()).Get(csv.Spec.Replaces)
+	if replacement != nil {
+		owners = append(owners, replacement)
+	}
 	ruleChecker := install.NewCSVRuleChecker(a.lister.RbacV1().RoleLister(), a.lister.RbacV1().RoleBindingLister(), a.lister.RbacV1().ClusterRoleLister(), a.lister.RbacV1().ClusterRoleBindingLister(), csv)
 	for _, desc := range csv.GetOwnedAPIServiceDescriptions() {
 		apiServiceName := fmt.Sprintf("%s.%s", desc.Version, desc.Group)
@@ -51,23 +75,34 @@ func (a *Operator) checkAPIServiceResources(csv *v1alpha1.ClusterServiceVersion,
 			"apiservice": apiServiceName,
 		})
 
+		apiService, err := a.lister.APIRegistrationV1().APIServiceLister().Get(apiServiceName)
+		if err != nil {
+			logger.Warnf("could not retrieve generated APIService")
+			errs = append(errs, err)
+			continue
+		}
+
+		// Check if the APIService is adoptable
+		if !ownerutil.OwnersIntersect(owners, apiService.GetLabels()) {
+			err := olmerrors.NewUnadoptableError("", apiServiceName)
+			logger.WithError(err).Warn("found unadoptable apiservice")
+			errs = append(errs, err)
+			return errs
+		}
+
 		serviceName := APIServiceNameToServiceName(apiServiceName)
 		service, err := a.lister.CoreV1().ServiceLister().Services(csv.GetNamespace()).Get(serviceName)
 		if err != nil {
 			logger.WithField("service", serviceName).Warnf("could not retrieve generated Service")
-			return err
-		}
-
-		apiService, err := a.lister.APIRegistrationV1().APIServiceLister().Get(apiServiceName)
-		if err != nil {
-			logger.Warnf("could not retrieve generated APIService")
-			return err
+			errs = append(errs, err)
+			continue
 		}
 
 		// Check if the APIService points to the correct service
 		if apiService.Spec.Service.Name != serviceName || apiService.Spec.Service.Namespace != csv.GetNamespace() {
 			logger.WithFields(log.Fields{"service": apiService.Spec.Service.Name, "serviceNamespace": apiService.Spec.Service.Namespace}).Warnf("APIService service reference mismatch")
-			return fmt.Errorf("APIService service reference mismatch")
+			errs = append(errs, fmt.Errorf("APIService service reference mismatch"))
+			continue
 		}
 
 		// Check if CA is Active
@@ -75,11 +110,13 @@ func (a *Operator) checkAPIServiceResources(csv *v1alpha1.ClusterServiceVersion,
 		ca, err := certs.PEMToCert(caBundle)
 		if err != nil {
 			logger.Warnf("could not convert APIService CA bundle to x509 cert")
-			return err
+			errs = append(errs, err)
+			continue
 		}
 		if !certs.Active(ca) {
 			logger.Warnf("CA cert not active")
-			return fmt.Errorf("CA cert not active")
+			errs = append(errs, fmt.Errorf("CA cert not active"))
+			continue
 		}
 
 		// Check if serving cert is active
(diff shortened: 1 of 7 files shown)


pkg/lib/ownerutil/util.go
```diff
@@ -123,6 +123,33 @@ func Adoptable(target Owner, owners []metav1.OwnerReference) bool {
...
 
+// APIServiceAdoptable checks whether a resource with the given set of OwnerReferences
+// is "adoptable" by the target OwnerReference. This function returns true if there
+// exists an element in owners referencing the same kind target does and one of
+// the owners is in the new CSV's replaces chain, otherwise it returns false.
+func APIServiceAdoptable(replaceCSV string, target Owner, owners []metav1.OwnerReference) bool {
+	if len(owners) == 0 {
+		// Resources with no owners are not adoptable
+		return false
+	} else if len(owners) == 1 && owners[0].Name == target.GetName() {
```
njhale: What happens when I try to deploy the same CSV in two different namespaces? A different resource of the same name? Could we get away with checking Kinds and UIDs here to be safe?

test/e2e/csv_e2e_test.go
```diff
@@ -1309,6 +1309,208 @@ func TestCreateCSVWithOwnedAPIService(t *testing.T) {
...
+	// Should eventually GC the CSV
+	err = waitForCSVToDelete(t, crc, csv.Name)
+	require.NoError(t, err)
+
+	// Recreate the old CSV
+	cleanupCSV, err := createCSV(t, c, crc, csv, testNamespace, false, true)
+	require.NoError(t, err)
+	defer cleanupCSV()
+
+	_, err = fetchCSV(t, crc, csv.Name, testNamespace, csvFailedChecker)
```
njhale: Can we check the reason too?

dinhxuanvu:
Will close this PR as the fix will be included in #710

pkg/controller/operators/olm/apiservices.go
```diff
@@ -640,8 +640,17 @@ func (a *Operator) installAPIServiceRequirements(desc v1alpha1.APIServiceDescrip
 		}
 		apiService.SetName(apiServiceName)
 	} else {
+		var owners []ownerutil.Owner
+		// Get replacing CSV
+		replaces, err := a.lister.OperatorsV1alpha1().ClusterServiceVersionLister().ClusterServiceVersions(csv.GetNamespace()).Get(csv.Spec.Replaces)
+		if err != nil {
+			owners = append(owners, csv)
+		} else {
+			owners = append(owners, csv)
```
njhale: nit: if you want you can condense this to one line
```go
owners = append(owners, csv, replaces)
```

test/e2e/csv_e2e_test.go
```diff
@@ -1320,6 +1320,380 @@ func TestCreateCSVWithOwnedAPIService(t *testing.T) {
...
+
+	depName := genName("hat-server")
+	mockGroup := fmt.Sprintf("hats.%s.redhat.com", genName(""))
+	version := "v1alpha1"
+	mockGroupVersion := strings.Join([]string{mockGroup, version}, "/")
+	mockKinds := []string{"fedora"}
+	depSpec := newMockExtServerDeployment(depName, mockGroupVersion, mockKinds)
+	apiServiceName := strings.Join([]string{version, mockGroup}, ".")
+
+	// Create CSVs for the package-server
```
njhale: nit: "hat-server"

pkg/lib/ownerutil/util.go
```diff
@@ -123,6 +123,31 @@ func Adoptable(target Owner, owners []metav1.OwnerReference) bool {
...
+// OwnersIntersect checks to see if any member of targets exists in owners
+func OwnersIntersect(targets []Owner, owners []metav1.OwnerReference) bool {
+	if len(owners) == 0 {
+		// Resources with no owners are not adoptable
+		return false
+	}
+
+	for _, target := range targets {
+		// Infer TypeMeta for the target
+		if err := InferGroupVersionKind(target); err != nil {
```
ecordell: Do we really care about the GVK if the `name` and `uid` are the same?

test/e2e/csv_e2e_test.go
```diff
@@ -1320,6 +1320,380 @@ func TestCreateCSVWithOwnedAPIService(t *testing.T) {
...
+	// Create CSV2 to replace CSV
+	cleanupCSV2, err := createCSV(t, c, crc, csv2, testNamespace, false, true)
+	require.NoError(t, err)
+	defer cleanupCSV2()
+
+	_, err = fetchCSV(t, crc, csv2.Name, testNamespace, csvSucceededChecker)
+	require.NoError(t, err)
+
+	// Should create Deployment
+	_, err = c.GetDeployment(testNamespace, depName)
```
njhale: All of this stuff will remain the same since the deployment didn't change. The only things that should get updated are the APIService CABundle and the generated Secret. But they should all still exist, so other than the comments here, this is probably a good check.

test/e2e/csv_e2e_test.go
```diff
@@ -1320,6 +1320,380 @@ func TestCreateCSVWithOwnedAPIService(t *testing.T) {
...
+
+	depName := genName("hat-server")
+	mockGroup := fmt.Sprintf("hats.%s.redhat.com", genName(""))
+	version := "v1alpha1"
+	mockGroupVersion := strings.Join([]string{mockGroup, version}, "/")
+	mockKinds := []string{"fedora"}
+	depSpec := newMockExtServerDeployment(depName, mockGroupVersion, mockKinds)
+	apiServiceName := strings.Join([]string{version, mockGroup}, ".")
+
+	// Create CSVs for the package-server
```
njhale: nit: "hat-server"

test/e2e/csv_e2e_test.go
```diff
@@ -1320,6 +1320,380 @@ func TestCreateCSVWithOwnedAPIService(t *testing.T) {
...
+
+	// Should create a system:auth-delegator Cluster RoleBinding
+	_, err = c.GetClusterRoleBinding(fmt.Sprintf("%s-system:auth-delegator", apiServiceName))
+	require.NoError(t, err, "error getting expected system:auth-delegator ClusterRoleBinding")
+
+	// Should create an extension-apiserver-authentication-reader RoleBinding in kube-system
+	_, err = c.GetRoleBinding("kube-system", fmt.Sprintf("%s-auth-reader", apiServiceName))
+	require.NoError(t, err, "error getting expected extension-apiserver-authentication-reader RoleBinding")
+
+	// Create a new CSV that owns the same API Service and replace the old CSV
```
njhale: nit: remove "and replace the old CSV" from the comment since it looks like you want to generate the same CSV in a different namespace.

test/e2e/csv_e2e_test.go
```diff
@@ -1320,6 +1320,380 @@ func TestCreateCSVWithOwnedAPIService(t *testing.T) {
...
+		},
+	})
+	require.NoError(t, err)
+	defer func() {
+		err = c.KubernetesInterface().CoreV1().Namespaces().Delete(newNamespace, &metav1.DeleteOptions{})
+		require.NoError(t, err)
+	}()
+
+	// Create CSV2 to replace CSV
+	cleanupCSV2, err := createCSV(t, c, crc, csv2, testNamespace, false, true)
```
njhale: I think this is supposed to create the same CSV in `newNamespace`.

test/e2e/csv_e2e_test.go
```diff
@@ -1320,6 +1320,380 @@ func TestCreateCSVWithOwnedAPIService(t *testing.T) {
...
+	require.NoError(t, err)
+	defer cleanupCSV2()
+
+	_, err = fetchCSV(t, crc, csv.Name, newNamespace, csvFailedChecker)
+	require.NoError(t, err)
+
+	// Should eventually GC the CSV
+	err = waitForCSVToDelete(t, crc, csv.Name)
```
njhale: No GCing will occur since the second CSV should transition to FAILED and stick around.

test/e2e/csv_e2e_test.go
```diff
@@ -1320,6 +1320,380 @@ func TestCreateCSVWithOwnedAPIService(t *testing.T) {
...
+		err = c.KubernetesInterface().CoreV1().Namespaces().Delete(newNamespace, &metav1.DeleteOptions{})
+		require.NoError(t, err)
+	}()
+
+	// Create CSV2 to replace CSV
+	cleanupCSV2, err := createCSV(t, c, crc, csv2, testNamespace, false, true)
+	require.NoError(t, err)
+	defer cleanupCSV2()
+
+	_, err = fetchCSV(t, crc, csv.Name, newNamespace, csvFailedChecker)
```
njhale: So I think the CSV will continue to cycle to pending with re-install here, since on failure we [do this check](https://github.com/operator-framework/operator-lifecycle-manager/blob/ed5fd4e28a09a279ea492ee9cfe680a732294c67/pkg/controller/operators/olm/apiservices.go#L44:20) and we should probably have something that checks that we can only transition out of FAILED if the APIService no longer exists, or if the OwnerRerference rules you created in this PR apply again.

pkg/lib/ownerutil/util.go
njhale: @ecordell AFAICT, some resources like `Project` and `Namespace` can share the same name and UID:
```
➜  k get ns openshift-operator-lifecycle-manager -o jsonpath={.metadata.uid}
4b309c1f-33ae-11e9-9f1e-02920433f040
➜  k get project openshift-operator-lifecycle-manager -o jsonpath={.metadata.uid}
4b309c1f-33ae-11e9-9f1e-02920433f040
```

pkg/lib/ownerutil/util.go
```diff
@@ -123,12 +123,51 @@ func Adoptable(target Owner, owners []metav1.OwnerReference) bool {
...
 		ownerRefs = []metav1.OwnerReference{}
 	}
+
+	// Infer TypeMeta for the target
+	if err := InferGroupVersionKind(owner); err != nil {
+		log.Warn(err.Error())
+	}
+	gvk := owner.GetObjectKind().GroupVersionKind()
+
+	for _, item := range ownerRefs {
```
njhale: 👍

pkg/controller/operators/olm/apiservices.go
```diff
@@ -221,6 +221,44 @@ func (a *Operator) areAPIServicesAvailable(descs []v1alpha1.APIServiceDescriptio
 	return true, nil
 }
 
+func (a *Operator) areAPIServicesNotPresentOrAdoptable(csv *v1alpha1.ClusterServiceVersion) bool {
```
njhale: This makes sense, but I wonder if we can avoid duplicating the work. What do you think about doing something like this: https://gist.github.com/njhale/f63eecc64a8a78197b0c43d634667aeb ?

ecordell:
We recently had to remove the ownerreferencs from apiservices entirely to deal with the GC bug. Instead, there are labels on apiservices that contain the same information.

@dinhxuanvu are you good with the changes that need to happen to this PR?

pkg/lib/ownerutil/util.go
```diff
@@ -124,12 +124,44 @@ func Adoptable(target Owner, owners []metav1.OwnerReference) bool {
 	return false
 }
 
+// OwnersIntersect checks to see if any member of targets exists in owners
+func OwnersIntersect(targets []Owner, labels map[string]string) bool {
```
njhale: nit: `targets` and `ownerLabels`

pkg/controller/operators/olm/apiservices.go
```diff
@@ -31,6 +31,11 @@ const (
 	Organization = "Red Hat, Inc."
 )
 
+type UnadoptableError struct {
```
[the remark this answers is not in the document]
njhale: I think we want to remove this extra type definition.

pkg/controller/operators/olm/operator.go
```diff
@@ -793,8 +793,12 @@ func (a *Operator) transitionCSVState(in v1alpha1.ClusterServiceVersion) (out *v
 		}
 
 		// Check if any generated resources are missing
-		if resErr := a.checkAPIServiceResources(out, certs.PEMSHA256); resErr != nil {
-			out.SetPhase(v1alpha1.CSVPhaseFailed, v1alpha1.CSVReasonAPIServiceResourceIssue, resErr.Error(), now)
+		if resErr := a.checkAPIServiceResources(out, certs.PEMSHA256); len(resErr) > 0 {
+			str := []string{}
+			for _, err := range resErr {
```
ecordell: optional: this would be a great place to use the Aggregated Error type from apimachinery:

https://github.com/kubernetes/kubernetes/blob/4992402fa5bee9f7ac6fe931d101d362a4295cea/staging/src/k8s.io/apimachinery/pkg/util/errors/errors.go#L41

You wouldn't have to manually build this string here

pkg/controller/operators/olm/operator.go
```diff
@@ -402,6 +402,18 @@ func (a *Operator) handleClusterServiceVersionDeletion(obj interface{}) {
 			a.csvQueueSet.Requeue(clusterServiceVersion.GetName(), namespace)
 		}
 	}
+
+	if err := a.checkReplacementsAndUpdateStatus(clusterServiceVersion); err == nil {
```
njhale: Maybe a safer approach here is to delete `APIService`s in `Owned` if they have an owner label referencing this CSV.

jpeeler: Ok updated. I guess this is safe since the deletion of a replaced CSV happens after the new CSV is successful.

pkg/controller/operators/olm/operator.go
```diff
@@ -406,6 +406,25 @@ func (a *Operator) handleClusterServiceVersionDeletion(obj interface{}) {
...
+	for _, desc := range clusterServiceVersion.Spec.APIServiceDefinitions.Owned {
+		apiServiceName := fmt.Sprintf("%s.%s", desc.Version, desc.Group)
+		fetched, err := a.lister.APIRegistrationV1().APIServiceLister().Get(apiServiceName)
+		if k8serrors.IsNotFound(err) {
+			continue
+		}
+		if err != nil {
+			logger.WithError(err).Warn("api service get failure")
+		}
+		apiServiceLabels := fetched.GetLabels()
```
njhale: Is `fetched` guaranteed to be non-nil here?

dinhxuanvu: It is. `Get(apiServiceName)` will only return nil for fetched if it is either NotFound or there is an error. The two `if` statements should handle that.

njhale: Reading this code, a non-nil error that __is not__ `NotFound` will allow `GetLabels()` to be called.

dinhxuanvu: I see it now. It should be `continue` on the the second `if` statement I believe.

pkg/controller/operators/olm/operator.go
```diff
@@ -406,6 +406,25 @@ func (a *Operator) handleClusterServiceVersionDeletion(obj interface{}) {
...
 	}
+
+	for _, desc := range clusterServiceVersion.Spec.APIServiceDefinitions.Owned {
+		apiServiceName := fmt.Sprintf("%s.%s", desc.Version, desc.Group)
+		fetched, err := a.lister.APIRegistrationV1().APIServiceLister().Get(apiServiceName)
+		if k8serrors.IsNotFound(err) {
+			continue
+		}
+		if err != nil {
+			logger.WithError(err).Warn("api service get failure")
```
dinhxuanvu: We may need a `continue` here to stop the current iteration execution.

jpeeler:
I could have sworn that when I tested this before that deleting a namespace would cause the CSV deletion handler to not run. But after testing today, it appears that is not the case.

pkg/controller/operators/olm/operator.go
jpeeler: Fixed.

PR #692, opened 2019-01-29
fix(catalogs): Remove global catalogs

dinhxuanvu:
Global catalogs are making it difficult to have private packages.
OperatorGroups remove the need for them. Get rid of global catalogs.


pkg/controller/operators/catalog/operator.go
```diff
@@ -623,7 +623,7 @@ func (o *Operator) ensureResolverSources(logger *logrus.Entry, namespace string)
 				continue
 			}
 			// only resolve in namespace local + global catalogs
```
njhale: nit: update comment

njhale:
I wonder if we even want to support inter-namespace catalog resolution anymore. I believe the original point was to include catalogs from the global namespace in resolution. @ecordell @alecmerdler @jpeeler any thoughts on how also removing inter-namespaces catalog resolution would effect operator-hub users?

jpeeler:
/retest
I think this got hit by the github weirdness?

njhale:
It almost looks like there's a ci cluster namespace sticking around messing these tests up. Could you head to https://console.svc.ci.openshift.org/k8s/cluster/projects to see if there are any namespaces you could delete?

ecordell:
/hold 

We may not want to do this just yet.

ecordell:
We can revisit this in the future, but the current way marketplace interacts with olm means we should hold off for now.

PR #693, opened 2019-01-30
chore(deploy): use downstream image for openshift builds

ecordell:
I did this clean off of master, in case something funky is going on with #666 

Also added operatorstatus to the chart templates


Makefile
```diff
@@ -178,6 +178,7 @@ release:
 	rm -rf manifests
 	mkdir manifests
 	cp -R deploy/ocp/manifests/$(ver)/. manifests
+	find ./manifests -type f -exec sed -i '' '/^#/d' {} \;
```
ecordell: remove comments from manifests

PR #696, opened 2019-01-31
Replace Polling with Watching in E2E Tests

alecmerdler:
### Description

Advantages of `.Watch()` over polling:
- Can catch extra API calls by counting the number of `Modify` events before we see the `status` we want
- Faster tests
- Less test logging noise

Addresses https://jira.coreos.com/browse/ALM-885


ecordell:
This looks good! Holding off the the lgtm because there are lots of small notes to change things still, just let me know when it's ready for a review

njhale:
@ecordell I think this is super valuable, but IIRC we decided that we were going to use watches for all _new_ e2e tests and replace older tests as we drop support for the `v1` APIs. If that is the case, should we close this?

ecordell:
@njhale seems reasonable to me

PR #699, opened 2019-02-04
fix(olm): Remove the "v" prefix in minKubeVersion if presents

dinhxuanvu:
Sometimes users may enter Kubernetes version as "v1.1.1" format
which will cause failture as the code currently only accept "1.1.1"
format. This PR will trim the prefix in advance to avoid this issue.


njhale:
/hold
unit tests should not flake

njhale:
The flaking unit tests are not from this PR and will be addressed in a follow up.
/hold cancel
/lgtm

PR #702, opened 2019-02-06
fix(deploy): Add a component prefix to manifests

smarterclayton:
Parallelism at the CVO level depends on having a consistent prefix
of the form `0000_NN_COMPONENT_*` - give that to OLM manifests.
OLM will now be installed in parallel with other components in
run level 50.


njhale:
We'll add a follow up for `make release` later.

/retest
/lgtm

PR #704, opened 2019-02-08
remove minishift support

leseb:
As per #705 (comment) it has been decided to drop support for minishift.

Signed-off-by: Sébastien Han <seb@redhat.com>


njhale:
@leseb These are good changes, but as we discussed in #705 we want to move away from minishift altogether. `make run-local` uses minikube and builds with `upstream.Dockerfile` which copies the required git objects into the image. If we can repurpose this PR to remove the minishift stuff, that would be incredibly helpful - up to you.

leseb:
@njhale done!

ecordell:
@leseb - we intentionally omitted the git objects to limit the size of our build context. Is there a reason you need or want a full git repo inside of the image?

leseb:
@ecordell I haven't tried again without minishift so I was just assuming this bug was on the other Dockerfile too. Thanks for commenting, let me remove that commit then.

ecordell:
Thanks for this! Could you also remove `e2e-local-shift` from the Makefile?

/approve

leseb:
@njhale done!

leseb:
I believe CI's failure is unrelated to that patch, isn't?

PR #707, opened 2019-02-12
fix(olm): OLM should not attempt to write ClusterStatus by default

dinhxuanvu:
The --writeStatusName flag for writing out a ClusterOperator status
should default to none, instead of currently defaulting to
operator-lifecycle-manager which causes OLM to CrashLoopBackoff in
upstream Kubernetes clusters.

### DIFF
--- a/cmd/olm/main.go
+++ b/cmd/olm/main.go
@@ -31,7 +31,7 @@ import (
 
 const (
 	defaultWakeupInterval = 5 * time.Minute
-	defaultOperatorName   = "operator-lifecycle-manager"
+	defaultOperatorName   = ""
 )
 
 // config flags defined globally so that they appear on the test binary as well


deploy/ocp/manifests/0.8.1/0000_50_10-olm-operator.deployment.yaml
```diff
@@ -26,7 +26,7 @@ spec:
           - /bin/olm
           args:
           - -writeStatusName
-          - operator-lifecycle-manager
```
njhale: We want this set for ocp

deploy/ocp/values.yaml
```diff
@@ -3,7 +3,7 @@ namespace: openshift-operator-lifecycle-manager
 catalog_namespace: openshift-operator-lifecycle-manager
 operator_namespace: openshift-operators
 imagestream: true
-writeStatusName: operator-lifecycle-manager
```
njhale: I think this is the only values file where we want `writeStatusName: operator-lifecycle-manager` set.

dinhxuanvu:
Will close this PR as the fix will be included in #710

PR #709, opened 2019-02-13
feat(catalogsource): allow grpc source types that don't require an image

njhale:
instead, they provide an address directly. OLM will not have visibility
into the resources required for running the grpc catalog for this case,
but will still connect and health check.

Replaces #684

### DIFF
--- a/pkg/api/apis/operators/v1alpha1/catalogsource_types.go
+++ b/pkg/api/apis/operators/v1alpha1/catalogsource_types.go
@@ -18,16 +18,43 @@ const (
 type SourceType string
 
 const (
-	SourceTypeInternal  SourceType = "internal"
+	// SourceTypeInternal (deprecated) specifies a CatalogSource of type SourceTypeConfigmap
+	SourceTypeInternal SourceType = "internal"
+
+	// SourceTypeConfigmap specifies a CatalogSource that generates a configmap-server registry
 	SourceTypeConfigmap SourceType = "configmap"
-	SourceTypeGrpc      SourceType = "grpc"
+
+	// SourceTypeGrpc specifies a CatalogSource that can use an operator registry image to generate a
+	// registry-server or connect to a pre-existing registry at an address.
+	SourceTypeGrpc SourceType = "grpc"
 )
 
 type CatalogSourceSpec struct {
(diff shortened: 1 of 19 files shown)


pkg/api/apis/operators/v1alpha1/catalogsource_types.go
```diff
@@ -18,16 +18,43 @@ const (
...
-	ConfigMap  string     `json:"configMap,omitempty"`
-	Image      string     `json:"image,omitempty"`
-	Secrets    []string   `json:"secrets,omitempty"`
+
+	// ConfigMap is the name of the ConfigMap to be used to back a configmap-server registry.
+	// Only used when SourceType = SourceTypeConfigmap or SourceTypeInternal.
+	// +Optional
+	ConfigMap string `json:"configMap,omitempty"`
+
+	// Address an address that OLM can use to connect to a pre-existing registry.
```
jpeeler: nit: "Address is a host that OLM...". Since one e2e failed, maybe it's worth fixing.

pkg/controller/operators/catalog/operator.go
```diff
@@ -664,7 +666,7 @@ func (o *Operator) ensureResolverSources(logger *logrus.Entry, namespace string)
 		logger.WithField("clientState", client.Conn.GetState()).Debug("source")
 		if client.Conn.GetState() == connectivity.TransientFailure {
 			logger.WithField("clientState", client.Conn.GetState()).Debug("waiting for connection")
-			ctx, _ := context.WithTimeout(context.TODO(), 5*time.Second)
+			ctx, _ := context.WithTimeout(context.TODO(), 2*time.Second)
```
jpeeler: I assume that was an arbitrary time out.

jpeeler:
/lgtm
Ping me if you change that one small nit and need another.

pkg/controller/operators/catalog/operator.go
njhale: I was playing with the timeouts - I'm changing it back, thanks for pointing it out. Yeah it's arbitrary. I really hope over the next sprint we can encapsulate connection logic into a separate module (I'm thinking a `ConnectionManager` in a separate thread).

PR #710, opened 2019-02-13
fix(olm): properly detect cluster operator API

jpeeler:
This was causing vanilla upstream to crash due to attempting to write to
a not present API.

### DIFF
--- a/cmd/olm/main.go
+++ b/cmd/olm/main.go
@@ -12,7 +12,6 @@ import (
 	log "github.com/sirupsen/logrus"
 	v1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/apimachinery/pkg/api/meta"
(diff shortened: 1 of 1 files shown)


jpeeler:
Will test this further while doing other installer work.

cmd/olm/main.go
```diff
@@ -131,84 +132,91 @@ func main() {
 	<-ready
 
 	if *writeStatusName != "" {
```
njhale: Nice, I like this. Would it be useful to us (and other teams) to factor this out into its own module/helper?

jpeeler: Maybe? This particular PR is really only about 5 lines long, but I had to shuffle other code around which makes it look bigger. Do you mean all the changed lines or just the discovery part? My understanding is that in the future we'll be outfitting all of OLM to update these status conditions, not just during start up. So it might be best to wait until later to see how that all shakes out.

jpeeler:
/retest
haven't tested on openshift yet... so this may really be broken

PR #711, opened 2019-02-13
Add Graceful Shutdown to PackageManifest Server

alecmerdler:
### Description

The queue informer provider was sharing the same stop channel as the API server, but the API server continues to accept requests for a certain grace period while the provider stopped immediately. Changed them to use different channels. Also updates the `PackageManifest` server to run with 2 replicas.

Fixes https://jira.coreos.com/browse/ALM-910

### DIFF
--- a/deploy/chart/values.yaml
+++ b/deploy/chart/values.yaml
@@ -28,7 +28,7 @@ catalog:
     beta.kubernetes.io/os: linux
 
 package:
-  replicaCount: 1
+  replicaCount: 2
   image:
     ref: quay.io/coreos/olm:master
     pullPolicy: Always
--- a/pkg/package-server/server/server.go
+++ b/pkg/package-server/server/server.go
@@ -182,8 +182,15 @@ func (o *PackageServerOptions) Run(stopCh <-chan struct{}) error {
 		return err
 	}
 
-	ready, _ := sourceProvider.Run(stopCh)
+	// Ensure that provider stops after the apiserver gracefully shuts down
+	provCh := make(chan struct{})
+	ready, done := sourceProvider.Run(provCh)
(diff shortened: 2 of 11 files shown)


deploy/chart/values.yaml
```diff
@@ -28,7 +28,7 @@ catalog:
     beta.kubernetes.io/os: linux
 
 package:
-  replicaCount: 1
+  replicaCount: 2
```
alecmerdler: We'll have to make a new release for these to take effect, right @njhale.

njhale: Yes, just run `make release` - #712 needs to be merged immediately and has changes to template and manifest prefixes, so when you rebase you'll need to update these again.

pkg/package-server/server/server.go
```diff
@@ -182,8 +182,15 @@ func (o *PackageServerOptions) Run(stopCh <-chan struct{}) error {
...
 
-	ready, _ := sourceProvider.Run(stopCh)
+	// Ensure that provider stops after the apiserver gracefully shuts down
+	provCh := make(chan struct{})
+	ready, done := sourceProvider.Run(provCh)
 	<-ready
 
-	return server.GenericAPIServer.PrepareRun().Run(stopCh)
+	err = server.GenericAPIServer.PrepareRun().Run(stopCh)
+	go func() { provCh <- struct{}{} }()
```
jpeeler: I'm curious why you do this in a goroutine.

alecmerdler: To avoid potential deadlock.

jpeeler: Wouldn't closing the channel be safe and more deterministic?

PR #716, opened 2019-02-18
Fix API Validation for `OperatorGroup` Spec

alecmerdler:
### Description

Previously was not enforcing proper `selector` fields, causing [this issue](https://github.com/openshift/cluster-monitoring-operator/pull/254) with `cluster-monitoring-operator`.

Addresses https://jira.coreos.com/browse/ALM-913

### DIFF
--- a/deploy/chart/templates/0000_50_olm_09-operatorgroup.crd.yaml
+++ b/deploy/chart/templates/0000_50_olm_09-operatorgroup.crd.yaml
@@ -30,35 +30,41 @@ spec:
             selector:
               type: object
               description: Optional label selector to find resources associated with or managed by the operator
-              properties:
-                matchLabels:
-                  type: object
-                  description: Label key:value pairs to match directly
-                matchExpressions:
-                  type: array
-                  description: A set of expressions to match against the resource.
-                  items:
-                    allOf:
-                      - type: object
-                        required:
-                        - key
-                        - operator
-                        - values
-                        properties:
-                          key:
-                            type: string
-                            description: the key to match
-                          operator:
-                            type: string
-                            description: the operator for the expression
-                            enum:
-                            - In
-                            - NotIn
-                            - Exists
-                            - DoesNotExist
-                          values:
-                            type: array
-                            description: set of values for the expression
+              anyOf:
+                - properties:
+                    matchLabels:
+                      type: object
+                      description: Label key:value pairs to match directly
+                  required:
+                    - matchLabels
+                - properties:
+                    matchExpressions:
+                      type: array
+                      description: A set of expressions to match against the resource.
+                      items:
+                        allOf:
+                          - type: object
+                            required:
+                            - key
+                            - operator
+                            - values
+                            properties:
+                              key:
+                                type: string
+                                description: the key to match
+                              operator:
+                                type: string
+                                description: the operator for the expression
+                                enum:
+                                - In
+                                - NotIn
+                                - Exists
+                                - DoesNotExist
+                              values:
+                                type: array
+                                description: set of values for the expression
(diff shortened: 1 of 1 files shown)


deploy/chart/templates/0000_50_olm_09-operatorgroup.crd.yaml
```diff
@@ -70,6 +76,7 @@ spec:
             staticProvidedAPIs:
               type: boolean
               description: If true, OLM will not modify the OperatorGroup's providedAPIs annotation.
+          additionalProperties: false
```
ecordell: `additionalProperties: false` isn't allowed in crd validation: https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#validation

deploy/chart/templates/0000_50_olm_09-operatorgroup.crd.yaml
```diff
@@ -30,35 +30,41 @@ spec:
...
-                            description: the operator for the expression
-                            enum:
-                            - In
-                            - NotIn
-                            - Exists
-                            - DoesNotExist
-                          values:
-                            type: array
-                            description: set of values for the expression
+              anyOf:
```
ecordell: odd that this isn't required for standard kube objects?

```json
  {  
    "io.k8s.apimachinery.pkg.apis.meta.v1.LabelSelector": {
      "description": "A label selector is a label query over a set of resources. The result of matchLabels and matchExpressions are ANDed. An empty label selector matches all objects. A null label selector matches no objects.",
      "properties": {
        "matchExpressions": {
          "description": "matchExpressions is a list of label selector requirements. The requirements are ANDed.",
          "items": {
            "$ref": "#/definitions/io.k8s.apimachinery.pkg.apis.meta.v1.LabelSelectorRequirement"
          },
          "type": "array"
        },
        "matchLabels": {
          "additionalProperties": {
            "type": "string"
          },
          "description": "matchLabels is a map of {key,value} pairs. A single {key,value} in the matchLabels map is equivalent to an element of matchExpressions, whose key field is \"key\", the operator is \"In\", and the values array contains only \"value\". The requirements are ANDed.",
          "type": "object"
        }
      },
      "type": "object"
    },
    "io.k8s.apimachinery.pkg.apis.meta.v1.LabelSelectorRequirement": {
      "description": "A label selector requirement is a selector that contains values, a key, and an operator that relates the key and values.",
      "properties": {
        "key": {
          "description": "key is the label key that the selector applies to.",
          "type": "string",
          "x-kubernetes-patch-merge-key": "key",
          "x-kubernetes-patch-strategy": "merge"
        },
        "operator": {
          "description": "operator represents a key's relationship to a set of values. Valid operators are In, NotIn, Exists and DoesNotExist.",
          "type": "string"
        },
        "values": {
          "description": "values is an array of string values. If the operator is In or NotIn, the values array must be non-empty. If the operator is Exists or DoesNotExist, the values array must be empty. This array is replaced during a strategic merge patch.",
          "items": {
            "type": "string"
          },
          "type": "array"
        }
      },
      "required": [
        "key",
        "operator"
      ],
      "type": "object"
    },
  }
```

alecmerdler: The legacy `selector` field [accepted a list of key/value labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#equality-based-requirement) (now `matchLabels`), so the upstream Kubernetes definition can't use `anyOf` in order to preserve backwards-compatibility. We don't have that requirement, so `anyOf` actually works (tested manually).

PR #717, opened 2019-02-20
feat(csv): allow use verb in permissions

ecordell:
previously this was only allowed in clusterpermissions


jcantrill:
cc @jcantrill

PR #720, opened 2019-02-20
add all-in-one yaml and helper script

dmesser:
add a helper script that takes all manifests generated by helm and concatenates them into a single YAML for simple direct installation using `kubectl -f https://raw.githubusercontent.com/dmesser/operator-lifecycle-manager/quickstart/deploy/upstream/quickstart/olm.yaml`

### DIFF
--- a/deploy/upstream/values.yaml
+++ b/deploy/upstream/values.yaml
@@ -7,21 +7,21 @@ writeStatusName: '""'
 olm:
   replicaCount: 1
   image:
-    ref: quay.io/coreos/olm@sha256:995a181839f301585a0e115c083619b6d73812c58a8444d7b13b8e407010325f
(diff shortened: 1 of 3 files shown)


deploy/upstream/values.yaml
```diff
@@ -7,21 +7,21 @@ writeStatusName: '""'
 olm:
   replicaCount: 1
   image:
-    ref: quay.io/coreos/olm@sha256:995a181839f301585a0e115c083619b6d73812c58a8444d7b13b8e407010325f
+    ref: quay.io/openshift/origin-operator-lifecycle-manager:latest
```
alecmerdler: We probably want Alpine images for upstream.

njhale: do we have any built?

PR #721, opened 2019-02-20
docs(operatorgroups): add more extensive docs

njhale:
Document OperatorGroup purpose and behavior in high fidelity.

### DIFF
--- a/Documentation/design/operatorgroups.md
+++ b/Documentation/design/operatorgroups.md
@@ -1,0 +1,218 @@
+# Operator Multitenancy with OperatorGroups
+
+An `OperatorGroup` is an OLM resource that provides rudimentary multitenant configuration to OLM installed operators.
+
+## OperatorGroup Overview
+
+* An `OperatorGroup` selects a set of target namespaces in which to generate required RBAC access for its member operators.
+* The set of target namespaces is provided via a comma-delimited string stored in the `olm.targetNamespaces` annotation. This annotation is applied to member operator's `ClusterServiceVersion` (CSV) instances and is projected into their deployments. It accessible to operator containers using [The Downward API](https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/#the-downward-api)
+* An operator is said to be a [member of an `OperatorGroup`](#operatorgroup-membership) if its CSV exists in the same namespace as the `OperatorGroup` and its CSV's [`InstallModes` support the set of namespaces targetted by the `OperatorGroup`](#installmodes-and-supported-operatorgroups)
+* In order to transition, a CSV must be an active member of an `OperatorGroup` that has no [provided API conflicts with intersecting `OperatorGroups`](#operatorgroup-intersection)
+
+## OperatorGroup Membership
+
+An operator defined by CSV `csv-a` is said to be a _member_ of `OperatorGroup` `op-a` in namespace `ns-a` if both of the following hold:
+* `op-a` is the only `OperatorGroup` in `ns-a`
+* `csv-a`'s `InstallMode`s support `op-a`'s target namespace set
+
+### TooManyOperatorGroups
+
+If there exists more than one `OperatorGroup` in a single namespace, any CSV created in that namespace will transition to a failure state with reason `TooManyOperatorGroups`. CSVs in a failed state for this reason will transition to pending once the number of `OperatorGroup`s in their namespaces reaches one.
+
+### InstallModes and Supported OperatorGroups
+
+An `InstallMode` consists of an `InstallModeType` field and a boolean `Supported` field. A CSV's spec can contain a set of `InstallModes` of four distinct `InstallModeTypes`: 
+* `OwnNamespace`: If supported, the operator can be a member of an `OperatorGroup` that selects its own namespace
+* `SingleNamespace`: If supported, the operator can be a member of an `OperatorGroup` that selects one namespace
+* `MultiNamespace`: If supported, the operator can be a member of an `OperatorGroup` that selects more than one namespace
+* `AllNamespaces`: If supported, the operator can be a member of an `OperatorGroup` that selects all namespaces (target namespace set is the empty string "")
+
+> Note: If a CSV's spec omit's an entry of `InstallModeType`, that type is considered unsupported unless support can be inferred by an existing entry that implicitly supports it.
+
+### UnsupportedOperatorGroup
+
+If a CSV's `InstallMode`s do not support the target namespace selection of the `OperatorGroup` in its namespace, the CSV will transition to a failure state with reason `UnsupportedOperatorGroup`. CSVs in a failed state for this reason will transition to pending once either the `OperatorGroups`'s target namespace selection changes to a supported configuration, or the CSV's `InstallMode`s are modified to support the `OperatorGroup`'s target namespace selection.
+
+## Target Namespace Selection
+
+Selection of the set of namespaces by specifying a label selector with the `spec.selector` field:
+
+```yaml
+apiVersion: operators.coreos.com/v1alpha2
+kind: OperatorGroup
+metadata:
+  name: my-group
+  namespace: my-namespace
+spec:
+  selector:
+    cool.io/prod: "true"
+```
+
+or by explicitly naming target namespaces with the `spec.targetNamespaces` field:
+
+```yaml
+apiVersion: operators.coreos.com/v1alpha2
+kind: OperatorGroup
+metadata:
+  name: my-group
+  namespace: my-namespace
+  spec:
+    targetNamespaces:
+    - my-namespace
+    - my-other-namespace
+    - my-other-other-namespace
+```
+
+> Note: If both `spec.targetNamespaces` and `spec.selector` are defined, `spec.selector` is ignored.
+
+Additionally, a _global_ `OperatorGroup` (which selects all namespaces) is specified by omitting both `spec.selector` and `spec.targetNamespaces`:
+
+```yaml
+apiVersion: operators.coreos.com/v1alpha2
+kind: OperatorGroup
+metadata:
+  name: my-group
+  namespace: my-namespace
+```
+
+The resolved set of selected namespaces is surfaced in an `OperatorGroup`'s `status.namespaces` field. A global `OperatorGroup`'s `status.namespace` is of length 1 and contains the empty string, `""`, which signals a consuming operator that it should watch all namespaces.
+
+> Note: The consuming operator must know to treat `""` as an all namespace configuration.
+
+## OperatorGroup CSV Annotations
+
+Member CSVs of an `OperatorGroup` get the following annotations:
+* `olm.operatorGroup=<group-name>`
+  * Contains the name of the `OperatorGroup`
+* `olm.operatorGroupNamespace=<group-namespace>`
+  * Contains the namespace of the `OperatorGroup`
+* `olm.targetNamespaces=<target-namespaces>`
+  * Contains a comma-delimited string listing the `OperatorGroup`'s target namespace selection. This annotation is projected onto the pod template of a CSV's deployments and can be consumed by a pod instance via [The Downward API](https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/#the-downward-api)
+
+> Note: All annotations except `olm.targetNamespaces` are included with [copied CSVs](#copied-csvs). Omitting the `olm.targetNamespaces` annotation on copied CSVs prevents the names of target namespaces from being leaked between tenants.
+
+## Provided APIs Annotation
+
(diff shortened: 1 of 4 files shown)


Documentation/design/operatorgroups.md
```diff
@@ -0,0 +1,225 @@
+# Operator Multitenancy with OperatorGroups
+
+An `OperatorGroup` is an OLM resource that provides rudementary multitenant configuration to OLM installed operators.
```
dinhxuanvu: rudimentary?

Documentation/design/operatorgroups.md
```diff
@@ -0,0 +1,225 @@
...
+
+## OperatorGroup CSV Annotations
+
+Member CSVs of an `OperatorGroup` get the following annotations:
+* `olm.operatorGroup=<group-name>`
+  * Contains the name of the `OperatorGroup`
+* `olm.operatorGroupNamespace=<group-namespace>`
+  * Contains the namespace of the `OperatorGroup`
+* `olm.targetNamespaces=<target-namespaces>`
+  * Contains a comma delimited string listing the `OperatorGroup`'s target namespace selection. This annotation is projected onto the pod template of a CSV's deployments and can be consumed by a pod instance via [The Downward API](https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/#the-downward-api)
```
dinhxuanvu: Should be comma-delimited as you use the same type above.

Documentation/design/operatorgroups.md
```diff
@@ -0,0 +1,221 @@
...
+## OperatorGroup CSV Annotations
+
+Member CSVs of an `OperatorGroup` get the following annotations:
+* `olm.operatorGroup=<group-name>`
+  * Contains the name of the `OperatorGroup`
+* `olm.operatorGroupNamespace=<group-namespace>`
+  * Contains the namespace of the `OperatorGroup`
+* `olm.targetNamespaces=<target-namespaces>`
+  * Contains a comma-delimited string listing the `OperatorGroup`'s target namespace selection. This annotation is projected onto the pod template of a CSV's deployments and can be consumed by a pod instance via [The Downward API](https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/#the-downward-api)
+* If the CSV does not [have a Copied status reason](#copied-csvs)
```
ecordell: I think this is missing the end of the sentence?

Documentation/design/operatorgroups.md
```diff
@@ -0,0 +1,221 @@
...
+
+### InstallModes and Supported OperatorGroups
+
+An `InstallMode` consists of an `InstallModeType` field and a boolean `Supported` field. A CSV's spec can contain a set of `InstallModes` of four distinct `InstallModeTypes`: 
+* `OwnNamespace`: If supported, the set of namespaces targetted by an `OperatorGroup` must contain the namespace the Operator is to be installed in.
+* `SingleNamespace`: If supported, the set of namespaces targetted by an `OperatorGroup` can be of length 1.
+* `MultiNamespace`: If supported, the set of namespaces targetted by an `OperatorGroup` can be of length >= 1. Any Operator supporting `MultiNamespace` implicitly supports `SingleNamespace` as well.
+* `AllNamespaces`: If supported, the Operator can support an `OperatorGroup` that selects all namespaces with a targetted set = [""]. Any Operator supporting `AllNamespaces` implicity supports `OwnNamespace` as well.
```
ecordell: I don’t think we would install an operator with ‘AllNamespaces’ into a group watching one namespaces, so I think that this extra comment makes sense

Documentation/design/operatorgroups.md
```diff
@@ -0,0 +1,221 @@
...
+    metadata:
+      creationTimestamp: null
+  targetNamespaces:
+  - local
+status:
+  lastUpdated: 2019-02-19T16:18:28Z
+  namespaces:
+  - local
+```
+
```
ecordell: I think we should introduce the Copied status and the static provided apis around here; they’re referenced frequently in the intersection section.

Documentation/design/operatorgroups.md
njhale: This just means that any CSV supporting `AllNamespaces` must also support `OwnNamespace` - For reference OwnNamespace doesn't imply SingleNamespace.

Documentation/design/operatorgroups.md
njhale: It's been updated to reflect your comments - forgot to push this morning.

Documentation/design/operatorgroups.md
ecordell: But can't I write an operator that starts up and watches all namespaces, but can't be deployed in a configuration that watches only its own namespace?

njhale: In the current implementation, `OwnNamespace` is an additional restriction on an OperatorGroup’s namespace set. More specifically, it means that a CSV can be a member of an OperatorGroup whose target namespace set includes the CSV’s namespace. This rule applies to `Single`, `Multi`, and `All`, rather than just applying to `Single` (see https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/api/apis/operators/v1alpha1/clusterserviceversion.go#L99).

We can change this logic to match your definition (It makes sense to) - but I think a change like that warrants a follow up PR.

Documentation/design/developing.md
```diff
@@ -1,8 +1,8 @@
  # Setting up a namespaced OLM for development
  
  * Disable global olm/catalog if the cluster is only for testing
-    * spin `tectonic-alm-operator` down to 0
-    * delete `alm-operator` and `catalog-operator` deployments from `tectonic-system`
+    * spin `tectonic-OLM-operator` down to 0
```
alecmerdler: RIP Tectonic :cry:  We should change this section to be more generic for upstream Kubernetes/OpenShift.

Documentation/design/operatorgroups.md
```diff
@@ -0,0 +1,221 @@
...
+Member CSVs of an `OperatorGroup` get the following annotations:
+* `olm.operatorGroup=<group-name>`
+  * Contains the name of the `OperatorGroup`
+* `olm.operatorGroupNamespace=<group-namespace>`
+  * Contains the namespace of the `OperatorGroup`
+* `olm.targetNamespaces=<target-namespaces>`
+  * Contains a comma-delimited string listing the `OperatorGroup`'s target namespace selection. This annotation is projected onto the pod template of a CSV's deployments and can be consumed by a pod instance via [The Downward API](https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/#the-downward-api)
+
+> Note: All annotations except `olm.targetNamespaces` are included with [copied CSVs](#copied-csvs).
```
alecmerdler: Could mention this is to avoid leaking namespaces.

Documentation/design/operatorgroups.md
```diff
@@ -0,0 +1,221 @@
...
+* `olm.operatorGroupNamespace=<group-namespace>`
+  * Contains the namespace of the `OperatorGroup`
+* `olm.targetNamespaces=<target-namespaces>`
+  * Contains a comma-delimited string listing the `OperatorGroup`'s target namespace selection. This annotation is projected onto the pod template of a CSV's deployments and can be consumed by a pod instance via [The Downward API](https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/#the-downward-api)
+
+> Note: All annotations except `olm.targetNamespaces` are included with [copied CSVs](#copied-csvs). 
+
+## Provided APIs Annotation
+
+Information about what GVKs are provided by an`OperatorGroup`are surfaced in an `olm.providedAPIs` annotation. The annotation's value is a string consisting of a set of `<Kind>.<version>.<group>`s delimited with commas. The GVKs of CRDs and APIServices provided by all active member CSVs of an`OperatorGroup`are included.
```
alecmerdler: Define `GroupVersionKind` before using GVK abbreviation.

Documentation/design/operatorgroups.md
```diff
@@ -0,0 +1,221 @@
...
+
+Here's an example of an`OperatorGroup`with a single active member CSV providing the PackageManifests resource:
+
+```yaml
+apiVersion: operators.coreos.com/v1alpha2
+kind: OperatorGroup
+metadata:
+  annotations:
+    olm.providedAPIs: PackageManifest.v1alpha1.packages.apps.redhat.com
+  creationTimestamp: 2019-02-19T16:18:28Z
```
alecmerdler: Remove extraneous `metadata` fields.

Documentation/design/operatorgroups.md
```diff
@@ -0,0 +1,221 @@
...
+```
+
+## OperatorGroup Intersection
+
+### OperatorGroup Intersection Terminology
+
+* Two`OperatorGroup`s are said to be _intersecting_ if the intersection of their target namespace sets __is not the empty set__
+* Two`OperatorGroup`s are said to have _intersecting provided APIs_ if they are __intersecting__ and the intersection of their provided API sets (defined by `olm.providedAPIs` annotations) __is not the empty set__
+
+### What can go Wrong?
```
alecmerdler: "What Can Go Wrong?"

Documentation/design/operatorgroups.md
```diff
@@ -0,0 +1,221 @@
+# Operator Multitenancy with OperatorGroups
+
+An `OperatorGroup` is an OLM resource that provides rudimentary multitenant configuration to OLM installed operators.
+
+## OperatorGroup Overview
+
+* An `OperatorGroup` selects a set of target namespaces in which to generate required RBAC access for its member operators.
+* The set of target namespaces is provided via a comma-delimited string stored in the `olm.targetNamespaces` annotation. This annotation is applied to member operator's deployments and is accessible through [The Downward API](https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/#the-downward-api)
+* An operator is said to be a [member of an `OperatorGroup`](#operatorgroup-membership) if its CSV exists in the same namespace as the `OperatorGroup` and its CSV's [`InstallModes` support the set of namespaces targetted by the `OperatorGroup`](#installmodes-and-supported-operatorgroups)
```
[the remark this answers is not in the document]
njhale: Besides being in the same namespace, `InstallModes` are the primary factor in determining membership. I think it makes sense for it to be in the overview. `InstallModes` are tightly coupled to membership so I feel that putting it in a separate bullet will dilute that association. 
The link serves as a pointer to the definition since I don't want to leave people wondering or `ctrl-f`ing.

Documentation/design/operatorgroups.md
```diff
@@ -0,0 +1,221 @@
...
+* In order to transition, a CSV must be an active member of an `OperatorGroup` that has no [provided API conflicts with intersecting `OperatorGroups`](#operatorgroup-intersection)
+
+## OperatorGroup Membership
+
+An operator defined by CSV `csv-a` is said to be a _member_ of `OperatorGroup` `op-a` in namespace `ns-a` if both of the following hold:
+* `op-a` is the only `OperatorGroup` in `ns-a`
+* `csv-a`'s `InstallMode`s support `op-a`'s target namespace set
```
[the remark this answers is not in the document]
njhale: Yes, there is [logic that prevents an `OperatorGroup` from acting on a CSV](https://github.com/operator-framework/operator-lifecycle-manager/blob/1ac9ace63034ea90f31e5ad33435d3f5476c7d85/pkg/controller/operators/olm/operator.go#L621). I'm strictly defining _membership_ by the terms listed here - which align with the criteria for an `OperatorGroup` to act on a CSV.

Documentation/design/operatorgroups.md
njhale: I point that out further into the doc, but i'll stick it in here too!

Documentation/design/developing.md
njhale: Totally agree - but I think we should address doc cleanup in a future PR. I was hoping to keep this one tightly scoped to `OperatorGroup`s. I'm pulling these changes out for now.

Documentation/design/operatorgroups.md
```diff
@@ -0,0 +1,217 @@
...
+
+### InstallModes and Supported OperatorGroups
+
+An `InstallMode` consists of an `InstallModeType` field and a boolean `Supported` field. A CSV's spec can contain a set of `InstallModes` of four distinct `InstallModeTypes`: 
+* `OwnNamespace`: If supported, the set of namespaces targetted by an `OperatorGroup` must contain the namespace the Operator is to be installed in.
+* `SingleNamespace`: If supported, the set of namespaces targetted by an `OperatorGroup` can be of length 1.
+* `MultiNamespace`: If supported, the set of namespaces targetted by an `OperatorGroup` can be of length >= 1. Any Operator supporting `MultiNamespace` implicitly supports `SingleNamespace` as well.
+* `AllNamespaces`: If supported, the set of namespaces targetted by an `OperatorGroup` can be = [""] (which tells operators to watch all namespaces). Any Operator supporting `AllNamespaces` implicity supports `OwnNamespace` as well.
```
SamiSousa: Does this mean that OwnNamespace must be set to true if AllNamespaces is supported?

njhale: That is the current behavior, but not the desired. I'm updating the docs to reflect our _desired_ behavior, in which case OwnNamespace does not need to be set to true if AllNamespaces is supported.

PR #722, opened 2019-02-20
fix(ocp): add csv and deployment for package server

jpeeler:
This is an initial simple approach to allowing CVO upgrades to work properly for the package server. The changes here have been made to only affect OCP installs.

### DIFF
--- a/deploy/chart/templates/0000_50_olm_15-packageserver.clusterserviceversion.yaml
+++ b/deploy/chart/templates/0000_50_olm_15-packageserver.clusterserviceversion.yaml
@@ -1,0 +1,74 @@
+{{ if (.Values.installType) and eq .Values.installType "ocp" }}
+# keep in sync with _packageserver.clusterserviceversion.yaml
+apiVersion: operators.coreos.com/v1alpha1
+kind: ClusterServiceVersion
+metadata:
+  name: packageserver.v{{ .Chart.Version }}
+  namespace: {{ .Values.namespace }}
+spec:
+  displayName: Package Server
+  description: Represents an Operator package that is available from a given CatalogSource which will resolve to a ClusterServiceVersion.
+  minKubeVersion: {{ .Values.minKubeVersion }}
+  keywords: ['packagemanifests', 'olm', 'packages']
+  maintainers:
+  - name: Red Hat
+    email: openshift-operators@redhat.com
+  provider:
+    name: Red Hat
+  links:
+  - name: Package Server
+    url: https://github.com/operator-framework/operator-lifecycle-manager/tree/master/pkg/package-server
+  installModes:
+  - type: OwnNamespace
+    supported: true
+  - type: SingleNamespace
+    supported: true
+  - type: MultiNamespace
+    supported: true
+  - type: AllNamespaces
+    supported: true
+  install:
+    strategy: deployment
+    spec:
+      clusterPermissions:
+      - serviceAccountName: packageserver
+        rules:
+        - apiGroups:
+          - ""
+          resources:
+          - configmaps
+          verbs:
+          - get
+          - list
+          - watch
+        - apiGroups:
+          - operators.coreos.com
+          resources:
+          - catalogsources
+          verbs:
+          - get
+          - list
+          - watch
+        - apiGroups:
+          - packages.apps.redhat.com
+          resources:
+          - packagemanifests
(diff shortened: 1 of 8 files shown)


deploy/chart/templates/0000_50_olm_15-packageserver.clusterserviceversion.yaml
```diff
@@ -0,0 +1,126 @@
+{{ if eq .Values.installType "ocp" }}
```
njhale: `_packageserver.clusterserviceversion.yaml` is a named partial template that you can [use the `include` keyword to pull in](https://docs.helm.sh/chart_template_guide/#named-templates) (then it will always be up to date).

deploy/chart/templates/0000_50_olm_16-packageserver.deployment.yaml
```diff
@@ -0,0 +1,63 @@
+{{ if eq .Values.installType "ocp" }}
```
njhale: Couldn't we use a named partial template for the deployment (`_packageserver.deployment.yaml`) and include it this file and `_packageserver.clusterserviceversion.yaml`? This way we can keep the deployment definition in sync automatically. Although, since we're not expecting to use this for too long, and the packageserver doesn't change very often, maybe just having the note here will be fine.

jpeeler: I'll go ahead and fix this.

deploy/chart/templates/0000_50_olm_15-packageserver.clusterserviceversion.yaml
jpeeler: The reason I didn't include the file is the syntax is slightly different. This file starts off with just `apiVersion...` whereas the _packageserver.clusterserviceversion.yaml file starts off with a list item:
`- apiVersion...`
If you can figure out how to make it come out right with those differences let me know.

njhale: yup, you are totally right - If we were going to fix that I think we would need to remove the list item operator `-` from the sub template and add it to the olm-operators configmap template.

jpeeler:
/retest
didn't work last time.

deploy/chart/templates/0000_50_olm_15-packageserver.clusterserviceversion.yaml
jpeeler: I actually tried what you said before and wasn't able to get it to work FWIW.

ecordell:
This looks good if we've verified it works for an upgrade! @jpeeler did you manage to get it tested?

jpeeler:
/retest
Might as well see if it gets past the current flakes. If it doesn't, will hold off and rebase again later.

deploy/chart/templates/0000_50_olm_15-packageserver.clusterserviceversion.yaml
```diff
@@ -0,0 +1,74 @@
+{{ if (.Values.installType) and eq .Values.installType "ocp" }}
+# keep in sync with _packageserver.clusterserviceversion.yaml
+apiVersion: operators.coreos.com/v1alpha1
+kind: ClusterServiceVersion
+metadata:
+  name: packageserver.v{{ .Chart.Version }}
```
njhale: Nice!

deploy/chart/templates/0000_50_olm_15-packageserver.clusterserviceversion.yaml
```diff
@@ -0,0 +1,74 @@
+{{ if (.Values.installType) and eq .Values.installType "ocp" }}
```
njhale: 👍

deploy/chart/templates/0000_50_olm_15-packageserver.clusterserviceversion.yaml
njhale: Is this changed in the `olm-operators` `ConfigMap` too?

jpeeler: Should be. I didn't actually change _packageserver.clusterserviceversion.yaml, which is included in the configmap.

PR #723, opened 2019-02-22
fix(csv): remove regex on base64 image data

ecordell:


spadgett:
cadvisor flake again :(

/retest

PR #729, opened 2019-02-26
fix(owners): remove cross-namespace and cluster->namespace ownerrefs

ecordell:
this may be triggering a gc bug in kube. where possible, ownerrefs
were replaced with labels so to perform the same function

I verified that we're not calling `AddOwnerReference` between objects in different namespaces by adding a panic to that function and running the tests.


pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -315,54 +331,92 @@ func (a *Operator) ensureSingletonRBAC(operatorNamespace string, csv *v1alpha1.C
...
+			existing.Rules = ownedRole.Rules
+			if _, err := a.OpClient.UpdateRole(existing); err != nil {
 				return err
 			}
+			continue
+		}
+
+		// role doesn't exist, create it
+		ownedRole.SetNamespace(targetNamespace)
+		ownedRole.SetOwnerReferences([]metav1.OwnerReference{ownerutil.NonBlockingOwner(targetCSV)})
```
alecmerdler: Nice, much better to set the `ownerReferences` to the copied CSV in that namespace.

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -315,54 +331,92 @@ func (a *Operator) ensureSingletonRBAC(operatorNamespace string, csv *v1alpha1.C
...
-			// TODO check rules
+		// role binding exists
+		if ok {
+			// TODO: should we check if SA/role has changed?
+			continue
+		}
+
+		// role binding doesn't exist
+		ownedRoleBinding.SetNamespace(targetNamespace)
+		ownedRoleBinding.SetOwnerReferences([]metav1.OwnerReference{ownerutil.NonBlockingOwner(targetCSV)})
```
alecmerdler: Nice, much better to set the `ownerReferences` to the copied CSV in that namespace.

pkg/lib/ownerutil/util.go
```diff
@@ -18,6 +18,7 @@ import (
 const (
 	OwnerKey          = "olm.owner"
 	OwnerNamespaceKey = "olm.owner.namespace"
+	OwnerKind         = "olm.owner.kind"
```
alecmerdler: I'll need to make sure the UI doesn't depend on this.

PR #736, opened 2019-03-02
OperatorGroup expansion/contraction

ecordell:
This takes the commits from #675, and rebases them one at a time to get them working with newer changes to OperatorGroups.

I'd like to double check / write a test for any possible edge cases here. The one that comes to mind is: the same CSV supports AllNamespaces and SingleNamespace, and we accidentally overwrite a "real" CSV with a "copied" CSV. It may be better to go ahead with these and follow up with that test case / fix, though, given that this is blocking.


pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -58,92 +57,121 @@ func (a *Operator) syncOperatorGroups(obj interface{}) error {
...
+	set := a.csvSet(group.Namespace, v1alpha1.CSVPhaseAny)
+	providedAPIsFromCSVs := make(resolver.APISet)
+	for _, csv := range set {
+		// Don't union providedAPIsFromCSVs if the CSV is copied (member of another OperatorGroup)
 		if csv.IsCopied() {
 			logger.Debug("csv is copied. not including in operatorgroup's provided api set")
 			continue
 		}
 
 		// TODO: Throw out CSVs that aren't members of the group due to group related failures?
```
njhale: Not doing this may cause more requeuing than necessary later on.

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -422,18 +449,22 @@ func (a *Operator) ensureTenantRBAC(operatorNamespace, targetNamespace string, c
 	return nil
 }
 
-func (a *Operator) copyCsvToTargetNamespace(csv *v1alpha1.ClusterServiceVersion, operatorGroup *v1alpha2.OperatorGroup) error {
-	namespaces := make([]string, 0)
-	if len(operatorGroup.Status.Namespaces) == 1 && operatorGroup.Status.Namespaces[0] == corev1.NamespaceAll {
-		namespaceObjs, err := a.lister.CoreV1().NamespaceLister().List(labels.Everything())
+func (a *Operator) copyCsvToTargetNamespace(csv *v1alpha1.ClusterServiceVersion, operatorGroup *v1alpha2.OperatorGroup, namespaces []string, pruneNamespaces []string) error {
+	// check for stale CSVs from a different previous operator group configuration
+	for _, ns := range pruneNamespaces {
```
njhale: This should probably be factored out into a `pruneCopiedCSVs(...)` method.

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -422,18 +449,22 @@ func (a *Operator) ensureTenantRBAC(operatorNamespace, targetNamespace string, c
...
+		fetchedCSVs, err := a.lister.OperatorsV1alpha1().ClusterServiceVersionLister().ClusterServiceVersions(ns).List(labels.Everything())
 		if err != nil {
 			return err
 		}
-		for _, ns := range namespaceObjs {
-			namespaces = append(namespaces, ns.GetName())
+		for _, csv := range fetchedCSVs {
+			if csv.IsCopied() && csv.GetAnnotations()[v1alpha2.OperatorGroupAnnotationKey] == operatorGroup.GetName() {
```
njhale: Should check the CSV name and the `olm.operatorNamespace` annotation too since we can have `OperatorGroup`s with the same name in different namespaces with CSVs of the same name.

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -510,6 +541,36 @@ func (a *Operator) addOperatorGroupAnnotations(obj *metav1.ObjectMeta, op *v1alp
...
+			fallthrough
+		case v1alpha2.OperatorGroupTargetsAnnotationKey:
+			copiedAnnotations[k] = v
+		}
+	}
+	return copiedAnnotations
+}
+
+// returns items in a that are not in b
+func sliceCompare(a, b []string) []string {
```
njhale: nit: this is really a set difference.

pkg/controller/operators/olm/operator.go
```diff
@@ -456,48 +492,61 @@ func (a *Operator) syncClusterServiceVersion(obj interface{}) (syncError error)
...
+		targetNamespaces = allNamespaces
+	} else {
+		targetNamespaces = operatorGroup.Status.Namespaces
+		pruneNamespaces = sliceCompare(allNamespaces, targetNamespaces)
+		logger.Debugf("Found namespaces to clean %v", pruneNamespaces)
 	}
 
 	// Check if we need to do any copying / annotation for the operatorgroup
-	if err := a.copyCsvToTargetNamespace(updatedCSV, operatorGroup); err != nil {
+	if err := a.copyCsvToTargetNamespace(clusterServiceVersion, operatorGroup, targetNamespaces, pruneNamespaces); err != nil {
```
njhale: We want to make sure the CSV isn't failed due to an intersecting `OperatorGroup` issue before copying to namespaces with intersecting provided APIs. As it stands, I think this breaks intersection logic because copied CSVs can stomp on good ones.

test/e2e/operator_groups_e2e_test.go
```diff
@@ -900,4 +899,196 @@ func TestStaticProviderOperatorGroup(t *testing.T) {
...
+		RoleRef: rbacv1.RoleRef{
+			Kind: "Role",
+			Name: role.GetName(),
+		},
+	}
+	_, err = c.CreateServiceAccount(serviceAccount)
+	require.NoError(t, err)
+	_, err = c.CreateRole(role)
+	require.NoError(t, err)
+	_, err = c.CreateRoleBinding(roleBinding)
```
njhale: I don't think these get adopted by the CSV, so we may want to defer a delete on them.

test/e2e/operator_groups_e2e_test.go
```diff
@@ -900,4 +899,196 @@ func TestStaticProviderOperatorGroup(t *testing.T) {
...
+	deploymentName := genName("operator-deployment")
+	namedStrategy := newNginxInstallStrategy(deploymentName, permissions, nil)
+
+	aCSV := newCSV(csvName, opGroupNamespace, "", *semver.New("0.0.0"), []apiextensions.CustomResourceDefinition{mainCRD}, nil, namedStrategy)
+	aCSV.Labels = map[string]string{"label": t.Name()}
+	createdCSV, err := crc.OperatorsV1alpha1().ClusterServiceVersions(opGroupNamespace).Create(&aCSV)
+	require.NoError(t, err)
+
+	t.Log("wait for CSV to succeed")
+	err = wait.Poll(pollInterval, pollDuration, func() (bool, error) {
```
njhale: Why not just call ` fetchCSV(t, crc, createdCSV.GetName(), opGroupNamespace, csvSucceededChecker)`?

test/e2e/operator_groups_e2e_test.go
```diff
@@ -900,4 +899,196 @@ func TestStaticProviderOperatorGroup(t *testing.T) {
...
+		if err != nil {
+			return false, err
+		}
+		t.Logf("%s (%s): %s", fetched.Status.Phase, fetched.Status.Reason, fetched.Status.Message)
+		return csvSucceededChecker(fetched), nil
+	})
+	require.NoError(t, err)
+
+	t.Log("Waiting for operator namespace csv to have annotations")
+	err = wait.Poll(pollInterval, pollDuration, func() (bool, error) {
```
njhale: You can make a checker func that both checks for success and the proper annotations and condense this into the previous `fetchCSV(...)` call.

test/e2e/operator_groups_e2e_test.go
```diff
@@ -900,4 +899,196 @@ func TestStaticProviderOperatorGroup(t *testing.T) {
...
+			return false, fetchErr
+		}
+		if checkOperatorGroupAnnotations(fetchedCSV, operatorGroup, true, corev1.NamespaceAll) == nil {
+			return true, nil
+		}
+		return false, nil
+	})
+	require.NoError(t, err)
+
+	csvList, err := crc.OperatorsV1alpha1().ClusterServiceVersions(corev1.NamespaceAll).List(metav1.ListOptions{LabelSelector: "label=TestCSVCopyWatchingAllNamespaces"})
```
njhale: Could use `fmt.Sprintf("label=%s", t.Name())`.

test/e2e/operator_groups_e2e_test.go
```diff
@@ -900,4 +899,196 @@ func TestStaticProviderOperatorGroup(t *testing.T) {
...
+	}
+	_, err = c.KubernetesInterface().CoreV1().Namespaces().Create(&otherNamespace)
+	require.NoError(t, err)
+	defer func() {
+		err = c.KubernetesInterface().CoreV1().Namespaces().Delete(otherNamespaceName, &metav1.DeleteOptions{})
+		require.NoError(t, err)
+	}()
+
+	t.Log("Waiting to ensure copied CSV shows up in other namespace")
+	err = wait.Poll(pollInterval, pollDuration, func() (bool, error) {
```
njhale: You can use `awaitCSV(...)` which will wait until the CSV exists and then apply the given status checker func.

test/e2e/operator_groups_e2e_test.go
```diff
@@ -900,4 +899,196 @@ func TestStaticProviderOperatorGroup(t *testing.T) {
...
+		}
+		return false, nil
+	})
+	require.NoError(t, err)
+
+	t.Log("Re-modifying operator group to be watching all namespaces")
+	currentOperatorGroup, err = crc.OperatorsV1alpha2().OperatorGroups(opGroupNamespace).Get(operatorGroup.Name, metav1.GetOptions{})
+	require.NoError(t, err)
+	currentOperatorGroup.Spec = v1alpha2.OperatorGroupSpec{}
+	_, err = crc.OperatorsV1alpha2().OperatorGroups(opGroupNamespace).Update(currentOperatorGroup)
```
njhale: Even though we fail-fast, this should probably be deferred as soon as the group is contracted.

test/e2e/operator_groups_e2e_test.go
```diff
@@ -900,4 +899,196 @@ func TestStaticProviderOperatorGroup(t *testing.T) {
...
+		}
+		if checkOperatorGroupAnnotations(fetchedCSV, operatorGroup, false, "") == nil {
+			return true, nil
+		}
+		return false, nil
+	})
+	require.NoError(t, err)
+
+	// verify created CSV is cleaned up after operator group is "contracted"
+	t.Log("Modifying operator group to no longer watch all namespaces")
```
njhale: I wonder if we should just generate a new `OperatorGroup` for this test so we avoid modifications to the initial setup.

test/e2e/util_test.go
```diff
@@ -39,7 +39,7 @@ import (
 
 const (
 	pollInterval = 1 * time.Second
-	pollDuration = 3 * time.Minute
+	pollDuration = 4 * time.Minute
```
njhale: I get concerned whenever we wait too long for results since we are likely to run into a resync. Maybe our tests should set our operator's resync interval to 1hr so we don't get fooled into thinking our control loops are reasonably fast.

pkg/controller/operators/olm/operator.go
```diff
@@ -317,6 +318,42 @@ func (a *Operator) syncObject(obj interface{}) (syncError error) {
...
+		a.requeueOwnerCSVs(metaObj)
+	}
+
+	namespace, ok := obj.(*corev1.Namespace)
+	if !ok {
+		return nil
+	}
+
+	// Check to see if any operator groups are associated with this namespace
+	operatorGroupList, err := a.lister.OperatorsV1alpha2().OperatorGroupLister().OperatorGroups(metav1.NamespaceAll).List(labels.Everything())
```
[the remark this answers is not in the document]
ecordell: This finds operatorgroups that would select the current namespace when the current namespace is synced. This catches add/delete events on the namespace for detecting expansion/contraction.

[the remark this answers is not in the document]
njhale: Ah got it, missed the namespace type assertion before. Looks good then.

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -58,92 +57,121 @@ func (a *Operator) syncOperatorGroups(obj interface{}) error {
...
+		if !reflect.DeepEqual(origCSVannotations, a.copyOperatorGroupAnnotations(&csv.ObjectMeta)) {
 			// CRDs don't support strategic merge patching, but in the future if they do this should be updated to patch
 			if _, err := a.client.OperatorsV1alpha1().ClusterServiceVersions(csv.GetNamespace()).Update(csv); err != nil {
 				// TODO: return an error and requeue the OperatorGroup here? Can this cause an update to never happen if there's resource contention?
-				logger.WithError(err).Warnf("update to existing csv failed")
+				logger.WithError(err).Warn("update to existing csv failed")
 				continue
+			} else {
+				if _, ok := origCSVannotations[v1alpha2.OperatorGroupAnnotationKey]; ok {
+					if err := a.csvQueueSet.Requeue(csv.GetName(), csv.GetNamespace()); err != nil {
```
njhale: Why requeue if we just updated the CSV?

pkg/controller/operators/olm/operator.go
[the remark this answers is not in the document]
njhale: Ah, I missed the type assertion above. Looks good then!

pkg/controller/operators/olm/operatorgroup.go
ecordell: Very good question

pkg/controller/operators/olm/operator.go
```diff
@@ -317,6 +318,35 @@ func (a *Operator) syncObject(obj interface{}) (syncError error) {
...
+
+	operatorGroupList, err := a.lister.OperatorsV1alpha2().OperatorGroupLister().OperatorGroups(metav1.NamespaceAll).List(labels.Everything())
+	if err != nil {
+		syncError = fmt.Errorf("lister failed: %v", err)
+		logger.Warn(syncError.Error())
+		return
+	}
+
+	for _, group := range operatorGroupList {
+		_, ok := resolver.NewNamespaceSet(group.Status.Namespaces).Intersection(resolver.NewNamespaceSet([]string{namespace.GetName()}))[namespace.GetName()]
```
njhale: Why generate a new `NamespaceSet` for `namespace` on every iteration of the loop? You should only need to generate a new one each iteration for `group.Status.Namespaces`. Also, rather than checking for `namespace.GetName`'s existence in the set, you can just check if `|intersection| > 0`.

pkg/controller/operators/olm/operator.go
```diff
@@ -415,29 +445,33 @@ func (a *Operator) removeDanglingChildCSVs(csv *v1alpha1.ClusterServiceVersion)
...
-		delete = true
-	} 
+		return a.deleteChild(csv)
+	}
 
-	if delete {
-		if err := a.client.OperatorsV1alpha1().ClusterServiceVersions(csv.GetNamespace()).Delete(csv.GetName(), &metav1.DeleteOptions{}); err != nil {
-			return err
+	if annotations := parent.GetAnnotations(); annotations != nil {
+		if _, ok := resolver.NewNamespaceSet(strings.Split(annotations[v1alpha2.OperatorGroupTargetsAnnotationKey], ","))[csv.GetNamespace()]; !ok {
```
njhale: You should probably use the `Contains(...)` method https://github.com/operator-framework/operator-lifecycle-manager/blob/e1e34a0e3acc16d9c48ff31b5f2d44aea0f6dfbb/pkg/controller/registry/resolver/groups.go#L52

It handles all-namespace edge cases too, which might be a problem here.

pkg/controller/operators/olm/operator.go
njhale: Using the `Contains` method is probably better: https://github.com/operator-framework/operator-lifecycle-manager/blob/e1e34a0e3acc16d9c48ff31b5f2d44aea0f6dfbb/pkg/controller/registry/resolver/groups.go#L52

pkg/controller/operators/olm/operator_test.go
```diff
@@ -3578,7 +3578,7 @@ func TestSyncOperatorGroups(t *testing.T) {
 
 			operatorGroup, err := op.GetClient().OperatorsV1alpha2().OperatorGroups(tt.initial.operatorGroup.GetNamespace()).Get(tt.initial.operatorGroup.GetName(), metav1.GetOptions{})
 			require.NoError(t, err)
-			assert.Equal(t, tt.expectedStatus, operatorGroup.Status)
+			assert.EqualValues(t, tt.expectedStatus, operatorGroup.Status)
```
jpeeler: I know this is annoying to fix, but doesn't this mask the difference between "" vs nil?

ecordell: Doesn't look like  it? https://github.com/stretchr/testify/blob/master/assert/assertions.go#L78:6

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -58,92 +58,115 @@ func (a *Operator) syncOperatorGroups(obj interface{}) error {
...
-	logger.Debug("csv annotation completed")
+	return providedAPIsFromCSVs
+}
 
-	// Don't prune providedAPIs if static
-	if op.Spec.StaticProvidedAPIs {
+func (a *Operator) pruneProvidedAPIs(group *v1alpha2.OperatorGroup, groupProvidedAPIs, providedAPIsFromCSVs resolver.APISet, logger *logrus.Entry) {
+	// Don't prune providedAPIsFromCSVs if static
+	if group.Spec.StaticProvidedAPIs {
 		a.Log.Debug("group has static provided apis. skipping provided api pruning")
```
njhale: Should we use the bound logger variable?

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -422,86 +444,110 @@ func (a *Operator) ensureTenantRBAC(operatorNamespace, targetNamespace string, c
...
+// TODO: do we want to do this? or just let the dangling CSV clean it up
+func (a *Operator) pruneFromNamespace(operatorGroupName, namespace string) error {
+	fetchedCSVs, err := a.lister.OperatorsV1alpha1().ClusterServiceVersionLister().ClusterServiceVersions(namespace).List(labels.Everything())
+	if err != nil {
+		return err
+	}
+
+	errlist := []error{}
+	for _, csv := range fetchedCSVs {
+		if csv.IsCopied() && csv.GetAnnotations()[v1alpha2.OperatorGroupAnnotationKey] == operatorGroupName {
```
njhale: I think for the sake of completeness we should check the `olm.operatorNamespace` annotation as well.

pkg/controller/operators/olm/operator.go
```diff
@@ -452,55 +511,79 @@ func (a *Operator) syncClusterServiceVersion(obj interface{}) (syncError error)
...
+			}
+			syncError = fmt.Errorf("error transitioning ClusterServiceVersion: %s and error updating CSV status: %s", syncError, updateErr)
 		}
-		syncError = fmt.Errorf("error transitioning ClusterServiceVersion: %s and error updating CSV status: %s", syncError, updateErr)
 	}
 
-	operatorGroup := a.operatorGroupForActiveCSV(logger, updatedCSV)
-	if operatorGroup == nil {
-		logger.WithField("reason", "no operatorgroup found for active CSV").Info("skipping CSV resource copy to target namespaces")
+	if outCSV.IsUncopiable() {
```
njhale: I think you also want this check in `syncCopyCSV` since it's also being fed by an informer.

pkg/controller/operators/olm/operator.go
```diff
@@ -1255,3 +1321,51 @@ func (a *Operator) cleanupCSVDeployments(logger *logrus.Entry, csv *v1alpha1.Clu
 		}
 	}
 }
+
+func (a *Operator) ensureDeploymentAnnotations(logger *logrus.Entry, csv *v1alpha1.ClusterServiceVersion) error {
```
ecordell: Added this while looking into why copying was failing - small optimization that skips the walk through Failed if the annotations change in an operatorgroup.

pkg/controller/operators/olm/operator_test.go
```diff
@@ -2067,6 +2085,51 @@ func TestTransitionCSV(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "SingleCSVSucceededToSucceeded/OperatorGroupChanged",
```
ecordell: Verifying we don't have to transition to failed to update the operatorgroup

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -64,102 +66,147 @@ func (a *Operator) syncOperatorGroups(obj interface{}) error {
...
 			if _, err := a.client.OperatorsV1alpha1().ClusterServiceVersions(csv.GetNamespace()).Update(csv); err != nil && !k8serrors.IsNotFound(err) {
-				// TODO: return an error and requeue the OperatorGroup here? Can this cause an update to never happen if there's resource contention?
 				logger.WithError(err).Warnf("error adding operatorgroup annotations")
-				return err
+				updateErrs = append(updateErrs, err)
+				continue
+			}
+		}
+
+		// requeue csvs in original namespaces or in new target namespaces (to capture removed/added namespaces)
```
ecordell: this ended up being the reason contraction wasn't working after the rebase. somewhere in reconciling those commits we lost the code that requeues CSVs from the _old_ set of namespaces, which is important to make sure we pick up CSVs that need to be cleaned up when an operatorgroup is contracted.

Added some abstractions around namespacesets to help make this clearer

njhale: Isn't this effectively done around [line 110](https://github.com/operator-framework/operator-lifecycle-manager/pull/736/files#diff-806841494fa68cbd1620970ba03566ccR110)?

ecordell: Line 110 doesn't capture the "contraction" case, right?

njhale: If you remove the the "not copied" predicate in `getAllCSVsThatProvide(...)`, then I think it would since that returns the set of CSVs across all namespaces that provide the given APIs. As I recall, the input it's given in `syncOperatorGroup` represents the union of APIs in the `providedAPIs` annotation and the CSVs in the `OperatorGroup`'s namespace.

PR #741, opened 2019-03-05
docs(subscriptions): add improved sub status proposal

njhale:
# Improved Subscription Status

The `Subscription` `CustomResource` needs to expose useful information when a failure scenario is encountered. Failures can be encountered throughout a `Subscription`'s existence and can include issues with `InstallPlan` resolution, `CatalogSource` connectivity, `ClusterServiceVersion` (CSV) status, and more. To surface this information, explicit status for `Subscriptions` will be introduced via status conditionswhich will be set by new, read-only sync handlers for dependent resources (`CatalogSource`s, `InstallPlan`s, CSV, etc...).

Note: @ecordell pointed out [statecharts](https://statecharts.github.io/what-is-a-statechart.html), which seem to be a model that fits our use-case pretty well. Modeling our controller logic as statecharts may help us simplify our transitions significantly.

### DIFF
--- a/Documentation/contributors/design-proposals/subscription-status.md
+++ b/Documentation/contributors/design-proposals/subscription-status.md
@@ -1,0 +1,775 @@
+# Improved Subscription Status
+
+Status: Pending
+
+Version: Alpha
+
+Implementation Owner: TBD
+
+## Motivation
+
+The `Subscription` `CustomResource` needs to expose useful information when a failure scenario is encountered. Failures can be encountered throughout a `Subscription`'s existence and can include issues with `InstallPlan` resolution, `CatalogSource` connectivity, `ClusterServiceVersion` (CSV) status, and more. To surface this information, explicit status for `Subscriptions` will be introduced via [status conditions](#status-conditions) which will be set by new, specialized status sync handlers for resources of interest (`Subscriptions`, `InstallPlan`s, `CatalogSource`s and CSVs).
+
+### Following Conventions
+
+In order to design a status that makes sense in the context of kubernetes resources, it's important to conform to current conventions. This will also help us avoid pitfalls that may have already been solved.
+
+#### Status Conditions
+
+The [kube api-conventions docs](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties) state that:
+> Conditions should be added to explicitly convey properties that users and components care about rather than requiring those properties to be inferred from other observations.
+
+A few internal Kubernetes resources that implement status conditions:
+
+- [NodeStatus](https://github.com/kubernetes/kubernetes/blob/6c31101257bfcd47fa53702cea07fe2eedf2ad92/pkg/apis/core/types.go#L3556)
+- [DeploymentStatus](https://github.com/kubernetes/kubernetes/blob/f5574bf62a051c4a41a3fff717cc0bad735827eb/pkg/apis/apps/types.go#L415)
+- [DaemonSetStatus](https://github.com/kubernetes/kubernetes/blob/f5574bf62a051c4a41a3fff717cc0bad735827eb/pkg/apis/apps/types.go#L582)
+- [ReplicaSetStatus](https://github.com/kubernetes/kubernetes/blob/f5574bf62a051c4a41a3fff717cc0bad735827eb/pkg/apis/apps/types.go#L751)
+
+Introducing status conditions will let us have an explicit, level-based view of the current abnormal state of a `Subscription`. They are essentially orthogonal states (regions) of the compound state (`SubscriptionStatus`)¹. A conditionᵢ has a set of sub states [Unknown, True, False] each with sub states of their own [Reasonsᵢ],where Reasonsᵢ contains the set of transition reasons for conditionᵢ. This compound state can be used to inform a decision about performing an operation on the cluster.
+
+> 1. [What is a statechart?](https://statecharts.github.io/what-is-a-statechart.html); see 'A state can have many "regions"'
+
+#### References to Related Objects
+
+The [kube api-convention docs](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#references-to-related-objects) state that:
+> References to specific objects, especially specific resource versions and/or specific fields of those objects, are specified using the ObjectReference type (or other types representing strict subsets of it).
+
+Rather than building our own abstractions to reference managed resources (like `InstallPlan`s), we can take advantage of the pre-existing `ObjectReference` type.
+
+## Proposal
+
+### Changes to SubscriptionStatus
+
+- Introduce a `SubscriptionCondition` type
+  - Describes a single state of a `Subscription` explicity
+- Introduce a `SubscriptionConditionType` field
+  - Describes the type of a condition
+- Introduce a `Conditions` field of type `[]SubscriptionCondition` to `SubscriptionStatus`
+  - Describes multiple potentially orthogonal states of a `Subscription` explicitly
+- Introduce an `InstallPlanRef` field of type [*corev1.ObjectReference](https://github.com/kubernetes/kubernetes/blob/f5574bf62a051c4a41a3fff717cc0bad735827eb/pkg/apis/core/types.go#L3993)
+  - To replace custom type with existing apimachinery type
+- Deprecate the `Install` field
+  - Value will be kept up to date to support older clients until a major version change
+- Introduce a `SubscriptionCatalogStatus` type
+  - Describes a Subscription's view of a CatalogSource's status
+- Introduce a `CatalogStatus` field of type `[]SubscriptionCatalogStatus`
+  - CatalogStatus contains the Subscription's view of its relevant CatalogSources' status
+
+### Changes to Subscription Reconciliation
+
+Changes to `Subscription` reconciliation can be broken into three parts:
+
+1. Phase in use of `SubscriptionStatus.Install` with `SubscriptionStatus.InstallPlanRef`:
+   - Write to `Install` and `InstallPlanRef` but still read from `Install`
+   - Read from `InstallPlanRef`
+   - Stop writing to `Install`
+2. Create independent sync handlers and workqueues for resources of interest (status-handler) that only update specific `SubscriptionStatus` fields and `StatusConditions`:
+   - Build actionable state reactively through objects of interest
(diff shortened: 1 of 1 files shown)


Documentation/contributors/design-proposals/subscription-status.md
```diff
@@ -0,0 +1,186 @@
+# Improved Subscription Status
+
+## Overview
+
+The `Subscription` `CustomResource` needs to expose useful information when a failure scenario is encountered. Failures can be encountered throughout a `Subscription`'s existence and can include issues with `InstallPlan` resolution, `CatalogSource` connectivity, `ClusterServiceVersion` (CSV) status, and more. To surface this information, explicit status for `Subscriptions` will be introduced via [status conditions](status-conditions) which will be set by new, read-only sync handlers for dependent resources (`InstallPlan`s and CSVs).
```
ecordell: `status-conditions` link doesn't seem right

Documentation/contributors/design-proposals/subscription-status.md
```diff
@@ -0,0 +1,188 @@
...
+
+const (
+    // SubscriptionFailed indicates the Subscription has failed progressing the operator installation to the latest version.
+    SubscriptionFailed SubscriptionConditionType = "Failed"
+
+    // SubscriptionProgressing indicates the Subscription is attempting to progress the cluster to the desired state.
+    SubscriptionProgressing SubscriptionConditionType = "Progressing"
+
+    // SubscriptionCatalogSourcesUnhealthy indicates that some or all of the CatalogSources to be used in resolution are unhealthy.
+    SubscriptionCatalogSourcesUnhealthy SubscriptionConditionType = "CatalogSourcesUnhealthy"
```
njhale: Should some of these `SubscriptionConditionType`s be reduced to "XResourceIssue" with the specific details being in the `SubscriptionCondition.Reason` field? From the examples listed here it seems like the answer is no, but the API convention doc defines `Reason` scope as being used for kubectl output and doesn't use a wrapper type for the string.

Documentation/contributors/design-proposals/subscription-status.md
```diff
@@ -0,0 +1,188 @@
...
+
+    // SubscriptionInstallPlanFailed indicates the InstallPlan responsible for installing the current CSV has failed.
+    SubscriptionInstallPlanFailed SubscriptionConditionType = "InstallPlanFailed"
+
+    // SubscriptionInstallPlanMissing indicates the InstallPlan responsible for installing the current CSV is missing.
+    SubscriptionInstallPlanMissing SubscriptionConditionType = "InstallPlanMissing"
+
+    // SubscriptionInstallPlanAwaitingManualApproval indicates the InstallPlan responsible for installing the current CSV is waiting 
+    // for manual approval.
+    SubscriptionInstallPlanAwaitingManua.lApproval SubscriptionConditionType = "InstallPlanAwaitingManualApproval"
```
njhale: Typo :(

Documentation/contributors/design-proposals/subscription-status.md
```diff
@@ -0,0 +1,188 @@
...
+    // +optional
+    LastHeartbeatTime  *metav1.Time  `json:"lastHeartbeatTime,omitempty" description:"last time we got an update on a given condition"`
+
+    // LastTransitionTime is the last time the condition transit from one status to another
+    // +optional
+    LastTransitionTime *metav1.Time  `json:"lastTransitionTime,omitempty" description:"last time the condition transit from one status to another"`
+}
+```
+
+### Changes to Subscription Reconciliation
```
njhale: Is this detailed enough for a proposal, or should there be more specific lines of code pointed out?

Documentation/contributors/design-proposals/subscription-status.md
```diff
@@ -0,0 +1,188 @@
+# Improved Subscription Status
```
njhale: Does it make sense for us to follow the main bullets defined in the [kube proposal template](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/Design_Proposal_TEMPLATE.md)?

Documentation/contributors/design-proposals/subscription-status.md
ecordell: Should we make a rule of thumb for this?

How about: if we use the state to make decisions on actions to perform, then it's a condition, otherwise it's just a reason?

i.e. if we ever find ourselves doing `if x.Condition.Status = Y && x.Condition.Reason = Z`, then the Reason should be it's own Condition?

alecmerdler: By keeping the number of condition "types" small and generic while having many "reasons", we improve future backwards-compatibility by not relying on implementation details (that are still changing). The "type" should be a positive adjective or present progressive verb (`status: 'False'` would indicate negative condition). It would be easier to check for only "false" conditions and act on those according to the "reason".

See `status.conditions` for a failed pod:
```yaml
conditions:
- type: Ready
  status: 'False'
  reason: ContainersNotReady
  message: 'containers with unready status: [hello-openshift]'
  lastProbeTime: '2019-03-11T16:29:09Z'
  lastTransitionTime: '2019-03-11T16:29:09Z'
```

So a smaller set of condition types could contain all our rich reasons:
```go
const (
  // Possible reasons: "CatalogSourceInvalid", "PackageChannelInvalid", "InstallPlanFailed", etc
  SubscriptionConditionResolved  SubscriptionConditionType = "Resolved"
  // Possible reasons: "RequiresApproval", "AtLatest", "InstalledCSVReplacementAvailable", etc
  SubscriptionConditionProgressing  SubscriptionConditionType = "Progressing"
)
```

njhale: @ecordell your rule of thumb sounds appealing - I like having a single, actionable decision point.
 
@alecmerdler I disagree with your definition for "types". I think they should be reserved for abnormal states rather than valid states, since valid should be the level/norm (as state approaches declared). Taking action also only makes sense when there exists an abnormal state, ie. progressing, unhealthy catalogsources, pods not ready, etc... Keep in mind, status has three types, `True`, `False, and `Unknown`, where the last of these indicates we need to check for the abnormal status. See [api-conventions docs](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties) where they talk about "abnormal-true" polarity. The same "check for 'False'" logic applies. 

In the future, as we look into statecharts, I'm thinking that each reasonable set of `Condition`s should get their own concurrent state machine.

I'm also not sure moving the details of what went wrong into `Reason` decouples us from implementation details. If `Reason` determines action, then we still need a state-machine that acts on a combination of "type" and `Reason` - we just split the predicate for state in two.

Documentation/contributors/design-proposals/subscription-status.md
njhale: By "current objects", are you referring to control-loops as well?

Documentation/contributors/design-proposals/subscription-status.md
```diff
@@ -0,0 +1,188 @@
...
+
+Changes to `Subscription` reconciliation can be broken into three parts:
+
+1. Replace the use of `SubscriptionStatus.Install` with `SubscriptionStatus.InstallPlanRef`:
+   - Place the details of the resolved `InstallPlan` in both fields
+   - Utilize only `InstallPlanRef` for decision making
+2. Create independent sync handlers and workqueues for resources of interest (status-handler) that only update specific `SubscriptionStatus` fields and `StatusConditions`:
+   - Build actionable state reactively through objects of interest
+   - Treat omitted `SubscriptionConditionTypes` in `SubscriptionStatus.Conditions` as having `ConditionStatus` "Unknown"
+   - Add new status-handlers with separate informers for:
```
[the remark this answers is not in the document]
njhale: Yup - that makes sense.

Documentation/contributors/design-proposals/subscription-status.md
```diff
@@ -0,0 +1,188 @@
...
+// SubscriptionConditionType indicates an explicit state condition about a Subscription in "abnormal-true"
+// polarity form (see https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties).
+type SusbcriptionConditionType string
+
+const (
+    // SubscriptionFailed indicates the Subscription has failed progressing the operator installation to the latest version.
+    SubscriptionFailed SubscriptionConditionType = "Failed"
+
+    // SubscriptionProgressing indicates the Subscription is attempting to progress the cluster to the desired state.
+    SubscriptionProgressing SubscriptionConditionType = "Progressing"
```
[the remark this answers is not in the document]
njhale: That makes sense but both options are contrary to the "abnormal true" polarity. Maybe `ResolutionFailed`?

But maybe we should sit down and talk amongst the team as to whether we buy this "abnormal true" reasoning.

Documentation/contributors/design-proposals/subscription-status.md
```diff
@@ -0,0 +1,188 @@
...
+    SubscriptionInstallPlanAwaitingManualApproval SubscriptionConditionType = "InstallPlanAwaitingManualApproval"
+
+    // SubscriptionInstalledCSVReplacementAvailable indicates there exists a replacement for the installed CSV.
+    SubscriptionInstalledCSVReplacementAvailable SubscriptionConditionType = "InstalledCSVReplacementAvailable"
+
+    // SubscriptionInstalledCSVMissing indicates the installed CSV is missing.
+    SubscriptionInstalledCSVMissing SubscriptionConditionType = "InstalledCSVMissing"
+
+    // SubscriptionInstalledCSVFailed indicates the installed CSV has failed.
+    SubscriptionInstalledCSVFailed SubscriptionConditionType = "InstalledCSVFailed"
```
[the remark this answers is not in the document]
njhale: We can add that. I just want to keep the number of condition types low - since once we have a condition, it becomes painful to remove if OLM dependents rely on it.

ecordell:
This looks great! Thanks for putting so much thought and research into this. Let's break these into smaller tasks and start working on them asap.

Some reflection on the proposal for future proposals (nothing here needs to change):

I don't think we need quite as much detail in the implementation, but I do think it's hard to know how much detail to go into. There were several functions in this one that could compile and run today, when they could just have comments describing what they do (or a comment elsewhere describing what such a function could do). The goal should be communicating to others what the changes will be and not necessarily writing them out. But these look good!

I'd also like to see some thought around complexity (I see some O(n) api calls in this one) and how we'll manage performance with higher numbers of objects, but we can discuss that type of thing in the actual PRs for this. 

I also think we'll want to start factoring our operators into smaller pieces, but that doesn't need to be included here either.

Great work!


/lgtm

PR #744, opened 2019-03-07
chore(manifests): regenerate manifests to include new anyOf validation

njhale:
Regenerates manifests to include the new anyOf validation for `OperatorGroup` `LabelSelectors`.


deploy/upstream/values.yaml
```diff
@@ -7,21 +7,21 @@ writeStatusName: '""'
 olm:
   replicaCount: 1
   image:
-    ref: quay.io/openshift/origin-operator-lifecycle-manager:latest
+    ref: quay.io/coreos/olm@sha256:995a181839f301585a0e115c083619b6d73812c58a8444d7b13b8e407010325f
```
ecordell: I think we should keep this as the openshift one until we've published a new upstream image

njhale: will be fixed when the next release is cut

PR #749, opened 2019-03-11
Emit `InstallSucceeded` Event for CSV

alecmerdler:
### Description

Switched all `SetPhase()` calls to `SetPhaseWithEvent()` so that these phase transitions have associated Kubernetes events emitted as well.

Addresses https://bugzilla.redhat.com/show_bug.cgi?id=1654977


alecmerdler:
Seeing this error in e2e tests:
> panic: interface conversion: interface {} is v1alpha1.ConditionReason, not string [recovered]

Yet:
```go
type ConditionReason string
```

PR #750, opened 2019-03-11
fix(olm): add deletion monitoring for api services

jpeeler:
If an api service has owner labels, check to make sure that the namespace and owning CSV still exists. If either does not, delete the api service.

This is a follow up to https://github.com/operator-framework/operator-lifecycle-manager/pull/690/commits/79b314dd91cb2f251783800ef4f55eb40eb2dda8. I haven't tested this yet, but maybe the existing e2e is good enough.

### DIFF
--- a/pkg/controller/operators/olm/operator.go
+++ b/pkg/controller/operators/olm/operator.go
@@ -19,6 +19,7 @@ import (
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/client-go/util/workqueue"
+	apiregistrationv1 "k8s.io/kube-aggregator/pkg/apis/apiregistration/v1"
 	kagg "k8s.io/kube-aggregator/pkg/client/informers/externalversions"
 
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
@@ -53,6 +54,7 @@ type Operator struct {
 	*queueinformer.Operator
 	csvQueueSet      *queueinformer.ResourceQueueSet
 	ogQueueSet       *queueinformer.ResourceQueueSet
+	apiSvcQueue      workqueue.RateLimitingInterface
 	client           versioned.Interface
 	resolver         install.StrategyResolverInterface
 	apiReconciler    resolver.APIIntersectionReconciler
@@ -166,17 +168,20 @@ func NewOperator(logger *logrus.Logger, crClient versioned.Interface, opClient o
 
 	// Register APIService QueueInformer
 	apiServiceInformer := kagg.NewSharedInformerFactory(opClient.ApiregistrationV1Interface(), wakeupInterval).Apiregistration().V1().APIServices()
-	op.RegisterQueueInformer(queueinformer.NewInformer(
-		workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "apiservices"),
+	apiServiceQueue := workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "apiservices")
+	apiServiceQueueInformer := queueinformer.NewInformer(
+		apiServiceQueue,
 		apiServiceInformer.Informer(),
-		op.syncObject,
+		op.syncAPIService,
 		&cache.ResourceEventHandlerFuncs{
 			DeleteFunc: op.handleDeletion,
 		},
 		"apiservices",
 		metrics.NewMetricsNil(),
 		logger,
-	))
+	)
+	op.RegisterQueueInformer(apiServiceQueueInformer)
+	op.apiSvcQueue = apiServiceQueue
 	op.lister.APIRegistrationV1().RegisterAPIServiceLister(apiServiceInformer.Lister())
 
 	// Register CustomResourceDefinition QueueInformer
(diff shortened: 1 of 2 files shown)


pkg/controller/operators/olm/operator.go
```diff
@@ -296,6 +301,50 @@ func NewOperator(logger *logrus.Logger, crClient versioned.Interface, opClient o
...
+			return
+		}
+	}
+
+	owningCSVName, owned := serviceLabels[ownerutil.OwnerKey]
+	if owned {
+		_, err := a.lister.OperatorsV1alpha1().ClusterServiceVersionLister().ClusterServiceVersions(owningNamespace).Get(owningCSVName)
+		if k8serrors.IsNotFound(err) {
+			logger.Debug("Deleting api service since owning CSV is not found")
+			syncError = a.OpClient.DeleteAPIService(apiSvc.GetName(), &metav1.DeleteOptions{})
```
ecordell: We also generate some rolebindings in kube-system for APIServices that we should clean up

pkg/controller/operators/olm/operator.go
```diff
@@ -296,6 +301,50 @@ func NewOperator(logger *logrus.Logger, crClient versioned.Interface, opClient o
...
+	}
+
+	logger := a.Log.WithFields(logrus.Fields{
+		"id":     queueinformer.NewLoopID(),
+		"apiSvc": apiSvc.GetName(),
+	})
+	logger.Info("syncing APIService")
+
+	serviceLabels := apiSvc.GetLabels()
+	owningNamespace, owned := serviceLabels[ownerutil.OwnerNamespaceKey]
```
ecordell: we should check if labels are nil before accessing

pkg/controller/operators/olm/operator.go
```diff
@@ -349,6 +398,22 @@ func (a *Operator) namespaceAddedOrRemoved(obj interface{}) {
 			}
 		}
 	}
+	// find any API services owned by this namespace and requeue them
```
ecordell: When a namespace gets deleted with a CSV in it, I'm pretty sure we get `delete` events for CSVs. I think it would be simpler to just queue up apiservices on CSV delete.

pkg/controller/operators/olm/operator.go
```diff
@@ -349,6 +398,22 @@ func (a *Operator) namespaceAddedOrRemoved(obj interface{}) {
...
+	stringSelector := fmt.Sprintf("%v==%v", ownerutil.OwnerNamespaceKey, namespace.GetName())
+	labelSelector, err := labels.Parse(stringSelector)
```
ecordell: this seems like a useful thing to factor out into ownerutil, no?

pkg/controller/operators/olm/operator.go
```diff
@@ -349,6 +398,22 @@ func (a *Operator) namespaceAddedOrRemoved(obj interface{}) {
...
+		logger.WithError(err).Warn("parser failed")
+		return
+	}
+	apiServices, err := a.lister.APIRegistrationV1().APIServiceLister().List(labelSelector)
+	if err != nil {
+		logger.WithError(err).Warn("APIService lister failed")
+		return
+	}
+	for _, svc := range apiServices {
+		logger.Infof("gcing api service %v", svc.GetName())
```
ecordell: nit: trying to migrate towards `logger.WithField("apiservice", svc.GetName()).Info` formatting for logs

test/e2e/csv_e2e_test.go
```diff
@@ -1216,9 +1218,31 @@ func TestCreateCSVWithOwnedAPIService(t *testing.T) {
...
-	cleanupCSV, err := createCSV(t, c, crc, csv, testNamespace, false, true)
+	cleanupCSV, err := createCSV(t, c, crc, csv, testNamespace, false, false)
 	require.NoError(t, err)
-	defer cleanupCSV()
+	defer func() {
+		watcher, err := c.ApiregistrationV1Interface().ApiregistrationV1().APIServices().Watch(metav1.ListOptions{FieldSelector: "metadata.name=" + apiServiceName})
+		require.NoError(t, err)
+
+		deleted := make(chan struct{})
+		go func() {
```
alecmerdler: Using [`Watch()` instead of polling](https://github.com/operator-framework/operator-lifecycle-manager/pull/696).

pkg/controller/operators/olm/operator.go
jpeeler: I still need to do this. But I admit I'm less concerned about it since I've never seen any stray RBAC cause package server install issues.

pkg/controller/operators/olm/operator.go
jpeeler: Refactored to use GetOwnerByKindLabel.

pkg/controller/operators/olm/operator.go
jpeeler: This is covering the case of deleting the namespace that also contains OLM. There already exists code to handle cleaning up api services upon CSV deletion in the "normal" case:
https://github.com/operator-framework/operator-lifecycle-manager/blob/5159b0a1c0dfe2cb76eb706afb4e3cc2ac4447fd/pkg/controller/operators/olm/operator.go#L449
I assume what happens is upon deleting the namespace hosting OLM, OLM doesn't have time to run any clean up.

pkg/controller/operators/olm/operator.go
jpeeler: Well, I'm not sure if the code is even needed. I'm specifically testing for when OLM is being terminated via namespace deletion, which it doesn't look like will allow/guarantee this code to run. Instead, the api service sync loop upon starting up will detect orphaned api services and get rid of them.

jpeeler:
Note that the e2e in this PR will pass without the changes. I've been testing by running TestCreateCSVWithOwnedAPIService once and then again without tearing down the cluster. With the way tests currently run, I'm not sure how to automate such a test.

ecordell:
/retest

^ to see where we stand with the tests now

It might be easier to get #696 merged first, since these tests are written using watches

ecordell:
/hold

jpeeler:
/hold cancel
previous command did not remove the label

Issue #747, opened 2019-03-11
Upstream quickstart and latest manifests should deploy same OLM image


ron1:
Upstream quickstart (https://github.com/operator-framework/operator-lifecycle-manager/blob/master/deploy/upstream/quickstart/olm.yaml) deploys OLM image quay.io/openshift/origin-operator-lifecycle-manager:latest while Upstream manifests (https://github.com/operator-framework/operator-lifecycle-manager/blob/master/deploy/upstream/manifests/0.8.1/0000_50_olm_06-olm-operator.deployment.yaml) deploys OLM image quay.io/coreos/olm@sha256:995a181839f301585a0e115c083619b6d73812c58a8444d7b13b8e407010325f.

These two installation methods should be consistent and deploy the same images.

njhale:
@ron1 Thanks for reporting! #752 and #771 resolve this issue. Closing this out.

PR #752, opened 2019-03-12
refactor(images): use operator-framework namespace for all olm images

njhale:
Add a few changes around upstream images and packaging:
- Use OLM images for all OLM images
- Use existing manifests in `package_quickstart.sh`
- Add quickstart generation to make `release` rule
- Build with go 1.12 in `upstream.Dockerfile`

Addresses part of #747 and [ALM-937](https://jira.coreos.com/browse/ALM-937).


scripts/build_local.sh
```diff
@@ -10,6 +10,6 @@ if [ -z "$NO_MINIKUBE" ]; then
   eval $(minikube docker-env) || { echo 'Cannot switch to minikube docker'; exit 1; }
   kubectl config use-context minikube
 fi
-docker build -f e2e.Dockerfile .
-docker tag $(docker images --filter 'label=stage=olm' --format '{{.CreatedAt}}\t{{.ID}}' | sort -nr | head -n 1 | cut -f2) quay.io/coreos/olm:local
-docker tag $(docker images --filter 'label=stage=builder' --format '{{.CreatedAt}}\t{{.ID}}' | sort -nr | head -n 1 | cut -f2) quay.io/coreos/olm-e2e:local
+docker build -f upstream.Dockerfile .
```
ecordell: Why the switch to upstream here?

njhale: I thought it would be nice to use the same image for local testing - I can put it back if you'd like.

alecmerdler:
Note that we are using an image in `operatorframework` in the [console end-to-end tests for OLM](https://github.com/openshift/console/blob/master/frontend/integration-tests/tests/olm/catalog.scenario.ts#L21) (to add a `CatalogSource` without Marketplace). I pushed the same image to the `operator-framework` org, but haven't made a PR against `openshift/console` to use it yet. Hopefully no one deletes that organization.

ecordell:
@njhale looks like a rebase is in order

njhale:
> Hopefully no one deletes that organization.

@alecmerdler we're using those images in OLM's e2e tests as well. I was planning on opening a PR with the `upstream.Dockerfile` for `operator-registry` today and rigging the Quay build-triggers.

PR #753, opened 2019-03-12
[WIP] docs(installplans): add robust installplans proposal

njhale:
# Robust InstallPlans

`InstallPlan` `CustomResource`s can be redesigned as a generic means of applying Kubernetes manifests to a cluster with robust controls around the lifecycle of each resource generated.

I want to get some feedback on the thoughts here before I add a hard implementation section.

### DIFF
--- a/Documentation/contributors/design-proposals/robust-installplans.md
+++ b/Documentation/contributors/design-proposals/robust-installplans.md
@@ -1,0 +1,184 @@
+# Robust `InstallPlan`s
+
+## Abstract
+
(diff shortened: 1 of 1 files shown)


njhale:
Some in-person discussions have changed my thinking about a generic resource operator (GRO). I no longer think it's worthwhile to pursue. If introduced, a GRO provides questionable value to operators. An operator can manage the resources it's concerned with more flexibly (and directly) without a GRO.

njhale:
Some in-person discussions have changed my thinking about a generic resource operator (GRO). I no longer think it's worthwhile to pursue. If introduced, a GRO provides questionable value to operators. An operator can manage the resources it's concerned with more flexibly (and directly) without a GRO.

PR #756, opened 2019-03-13
Subscriptions step through multiple updates

ecordell:
To start with, this adds an e2e test for validating that subscriptions step through multiple intermediate versions. We have tests for this for CSVs, but not one that validates the flow through subscriptions. This is a pretty big gap.

Unfortunately my laptop will die if I run this to validate it - I expect to see this test fail, and will update this PR with a fix for the (hopefully) failing test.


test/e2e/subscription_e2e_test.go
```diff
@@ -612,3 +614,149 @@ func TestSusbcriptionWithStartingCSV(t *testing.T) {
...
+	_, err = c.KubernetesInterface().CoreV1().ConfigMaps(testNamespace).Update(configMap)
+	require.NoError(t, err)
+
+	// wait for catalog to update
+	_, err = fetchCatalogSource(t, crc, catalogSourceName, testNamespace, func(catalog *v1alpha1.CatalogSource) bool {
+		if catalog.Status.LastSync != fetchedInitialCatalog.Status.LastSync && catalog.Status.ConfigMapResource.ResourceVersion != fetchedInitialCatalog.Status.ConfigMapResource.ResourceVersion {
+			fmt.Println("catalog updated")
+			return true
+		}
+		fmt.Println("waiting for catalog pod to be available")
```
alecmerdler: The gRPC pod may still take some time before it spins up after the `CatalogSource` is synced, right?

ecordell:
/hold

PR #757, opened 2019-03-13
feat(resolver): take minKubeVersion into account when resolving

ecordell:
We won't attempt to upgrade to an operator that says it can't run onthe version of kube we're currently running on.

This is moving some of the checks slightly up the stack into resolution. We still want/need the checks on CSV installation. 

This isn't a perfect solution; ideally we would be able to ask a registry pod for the "latest operator that supports this version of kube". And we'll need something relatively similar for native APIs.

Currently can't run the e2e test to verify it locally, so I might update here if that's needed.


ecordell:
/hold

PR #758, opened 2019-03-14
fix(csv): properly detect apiservice and crd conflicts

njhale:
I noticed that `TestUpdateCSVWithOwnedAPIService` was flaking due to issues with how owner conflicts were being detected and status was being set. This PR introduces the following changes to stop a CSV that has provided APIService conflicts from cycling status endlessly:

- Move APIService resource check before installation check to prevent
transitions from FAILED to PENDING when a CSV has conflicting owners
- Add SetPhaseWithEventIfChanged CSV method
- Use Aggregate error type for APIService resource checks
- Use ownerlabels for APIService conflict detection
- Only detect CRD owner conflict when the potential conflict is not the
syncing CSV or the CSV it replaces

I was hoping to open this PR much earlier but it took a little while to get things _just right_.

### DIFF
--- a/pkg/controller/operators/olm/apiservices.go
+++ b/pkg/controller/operators/olm/apiservices.go
@@ -59,24 +59,28 @@ func (a *Operator) apiServiceResourceErrorActionable(err error) bool {
 
 // checkAPIServiceResources checks if all expected generated resources for the given APIService exist
 func (a *Operator) checkAPIServiceResources(csv *v1alpha1.ClusterServiceVersion, hashFunc certs.PEMHash) error {
+	logger := log.WithFields(log.Fields{
+		"csv":       csv.GetName(),
+		"namespace": csv.GetNamespace(),
+	})
+
 	errs := []error{}
 	owners := []ownerutil.Owner{csv}
+
 	// Get replacing CSV if exists
 	replacing, err := a.lister.OperatorsV1alpha1().ClusterServiceVersionLister().ClusterServiceVersions(csv.GetNamespace()).Get(csv.Spec.Replaces)
-	if err != nil && k8serrors.IsNotFound(err) == false {
-		a.Log.Debugf("Replacement error regarding CSV (%v): %v", csv.GetName(), err)
-		errs = append(errs, err)
-		return utilerrors.NewAggregate(errs)
+	if err != nil && !k8serrors.IsNotFound(err) {
+		logger.WithError(err).Warn("could not get replacement csv")
+		return err
 	}
 	if replacing != nil {
 		owners = append(owners, replacing)
 	}
+
 	ruleChecker := install.NewCSVRuleChecker(a.lister.RbacV1().RoleLister(), a.lister.RbacV1().RoleBindingLister(), a.lister.RbacV1().ClusterRoleLister(), a.lister.RbacV1().ClusterRoleBindingLister(), csv)
 	for _, desc := range csv.GetOwnedAPIServiceDescriptions() {
-		apiServiceName := fmt.Sprintf("%s.%s", desc.Version, desc.Group)
-		logger := log.WithFields(log.Fields{
-			"csv":        csv.GetName(),
-			"namespace":  csv.GetNamespace(),
+		apiServiceName := desc.GetName()
+		logger := logger.WithFields(log.Fields{
 			"apiservice": apiServiceName,
 		})
 
@@ -88,7 +92,7 @@ func (a *Operator) checkAPIServiceResources(csv *v1alpha1.ClusterServiceVersion,
 		}
 
 		// Check if the APIService is adoptable
-		if !ownerutil.OwnersIntersect(owners, apiService) {
+		if !ownerutil.AdoptableLabels(apiService.GetLabels(), true, owners...) {
 			err := olmerrors.NewUnadoptableError("", apiServiceName)
 			logger.WithError(err).Warn("found unadoptable apiservice")
 			errs = append(errs, err)
@@ -253,8 +257,7 @@ func (a *Operator) isAPIServiceAvailable(apiService *apiregistrationv1.APIServic
 
 func (a *Operator) areAPIServicesAvailable(csv *v1alpha1.ClusterServiceVersion) (bool, error) {
 	for _, desc := range csv.Spec.APIServiceDefinitions.Owned {
-		apiServiceName := fmt.Sprintf("%s.%s", desc.Version, desc.Group)
-		apiService, err := a.lister.APIRegistrationV1().APIServiceLister().Get(apiServiceName)
+		apiService, err := a.lister.APIRegistrationV1().APIServiceLister().Get(desc.GetName())
 		if k8serrors.IsNotFound(err) {
 			return false, nil
 		}
@@ -550,7 +553,7 @@ func (a *Operator) installAPIServiceRequirements(desc v1alpha1.APIServiceDescrip
 	existingAuthDelegatorClusterRoleBinding, err := a.lister.RbacV1().ClusterRoleBindingLister().Get(authDelegatorClusterRoleBinding.GetName())
 	if err == nil {
 		// Check if the only owners are this CSV or in this CSV's replacement chain.
-		if ownerutil.AdoptableLabels(csv, existingAuthDelegatorClusterRoleBinding.GetLabels()) {
+		if ownerutil.AdoptableLabels(existingAuthDelegatorClusterRoleBinding.GetLabels(), true, csv) {
 			ownerutil.AddOwnerLabels(authDelegatorClusterRoleBinding, csv)
 		}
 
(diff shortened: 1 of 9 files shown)


pkg/controller/operators/olm/apiservices.go
```diff
@@ -251,8 +255,8 @@ func (a *Operator) isAPIServiceAvailable(apiService *apiregistrationv1.APIServic
 	return false
 }
 
-func (a *Operator) areAPIServicesAvailable(descs []v1alpha1.APIServiceDescription) (bool, error) {
-	for _, desc := range descs {
+func (a *Operator) areAPIServicesAvailable(csv *v1alpha1.ClusterServiceVersion) (bool, error) {
+	for _, desc := range csv.Spec.APIServiceDefinitions.Owned {
 		apiServiceName := fmt.Sprintf("%s.%s", desc.Version, desc.Group)
```
jpeeler: nit: ideally would use desc.GetName() here too

pkg/controller/operators/olm/operator.go
```diff
@@ -855,15 +849,18 @@ func (a *Operator) transitionCSVState(in v1alpha1.ClusterServiceVersion) (out *v
 		}
 
 		// Check for CRD ownership conflicts
-		csvSet := a.csvSet(out.GetNamespace(), v1alpha1.CSVPhaseAny)
-		if syncError = a.crdOwnerConflicts(out, csvSet); syncError != nil {
-			out.SetPhaseWithEvent(v1alpha1.CSVPhaseFailed, v1alpha1.CSVReasonOwnerConflict, fmt.Sprintf("crd owner conflict: %s", syncError), now, a.recorder)
+		if syncError = a.crdOwnerConflicts(out, a.csvSet(out.GetNamespace(), v1alpha1.CSVPhaseAny)); syncError != nil {
+			if syncError == ErrCRDOwnerConflict {
```
jpeeler: Is the idea here to protect against future additional errors being added to crdOwnerConflicts? The same thing is done with apiServiceOwnerConflicts too.

pkg/controller/operators/olm/operator.go
```diff
@@ -901,16 +899,28 @@ func (a *Operator) transitionCSVState(in v1alpha1.ClusterServiceVersion) (out *v
 
 		if installErr := a.updateInstallStatus(out, installer, strategy, v1alpha1.CSVPhaseInstalling, v1alpha1.CSVReasonWaiting); installErr == nil {
 			logger.WithField("strategy", out.Spec.InstallStrategy.StrategyName).Infof("install strategy successful")
+		} else {
+			// Set phase to failed if it's been a long time since the last transition (5 minutes)
+			if metav1.Now().Sub(out.Status.LastTransitionTime.Time) >= 5*time.Minute {
```
jpeeler: Awesome check.

pkg/controller/operators/olm/operator.go
```diff
@@ -989,20 +987,26 @@ func (a *Operator) transitionCSVState(in v1alpha1.ClusterServiceVersion) (out *v
 			return
 		}
 
-		// Check if any generated resources are missing
-		if resErr := a.checkAPIServiceResources(out, certs.PEMSHA256); len(resErr) > 0 {
-			// Check if API services are adoptable. If not, keep CSV as Failed state
-			if a.apiServiceResourceErrorsActionable(resErr) {
-				out.SetPhaseWithEvent(v1alpha1.CSVPhasePending, v1alpha1.CSVReasonAPIServiceResourcesNeedReinstall, resErr[0].Error(), now, a.recorder)
-				return
+		// Check if any generated resources are missing and that OLM can action on them
```
jpeeler: nit: replace and with so

pkg/lib/ownerutil/util.go
```diff
@@ -257,17 +216,24 @@ func IsOwnedByKindLabel(object metav1.Object, ownerKind string) bool {
 	return object.GetLabels()[OwnerKind] == ownerKind
 }
 
-// AdoptableLabels determines if an OLM managed resource is adoptable based on its owner labels
-// Generally used for cross-namespace ownership and for Cluster -> Namespace scope
-func AdoptableLabels(target Owner, labels map[string]string) bool {
+// AdoptableLabels determines if an OLM managed resource is adoptable by any of the given targets based on its owner labels.
+// The checkName perimeter enables an additional check for name equality with the `olm.owner` label.
+// Generally used for cross-namespace ownership and for Cluster -> Namespace scope.
+func AdoptableLabels(labels map[string]string, checkName bool, targets ...Owner) bool {
```
jpeeler: What scenario would you not want to check the name?

pkg/controller/operators/olm/operator.go
```diff
@@ -855,15 +849,18 @@ func (a *Operator) transitionCSVState(in v1alpha1.ClusterServiceVersion) (out *v
 		}
 
 		// Check for CRD ownership conflicts
-		csvSet := a.csvSet(out.GetNamespace(), v1alpha1.CSVPhaseAny)
-		if syncError = a.crdOwnerConflicts(out, csvSet); syncError != nil {
-			out.SetPhaseWithEvent(v1alpha1.CSVPhaseFailed, v1alpha1.CSVReasonOwnerConflict, fmt.Sprintf("crd owner conflict: %s", syncError), now, a.recorder)
+		if syncError = a.crdOwnerConflicts(out, a.csvSet(out.GetNamespace(), v1alpha1.CSVPhaseAny)); syncError != nil {
```
jpeeler: nit: even nicer would be to just generate the csv set in the method.

pkg/controller/operators/olm/apiservices.go
njhale: Agreed. There are a bunch of places that use it. I didn't change older stuff because I wanted to get this in quickly. I think we can phase it in with future PRs. I also wonder if we can just enforce the relation to group and version in our OpenAPI schema validation.

pkg/controller/operators/olm/operator.go
njhale: Doing that makes the function less _pure_ , which makes it slightly more difficult to unit test. We would need to populate the client/informer/listers with test data  instead of just passing CSVs in.

Disclaimer: not sure if I have a unit test for this one.

pkg/lib/ownerutil/util.go
njhale: That's a super good point. This would be useful in situations where all we care about is that the resource was OLM generated or explicitly set to be managed by OLM. The thought process is a carry-over from `OwnerReferences` - we didn't want to adopt a resource that wasn't managed by OLM since we didn't want it to be deleted when the owner was.

pkg/controller/operators/olm/operator.go
```diff
@@ -1177,46 +1182,46 @@ func (a *Operator) parseStrategiesAndUpdateStatus(csv *v1alpha1.ClusterServiceVe
...
-				continue
-			}
-			if csv.OwnsAPIService(name) {
-				owned = true
-			}
-			if owned && in.Spec.Replaces == csvName {
-				return nil
-			}
-		}
+func (a *Operator) apiServiceOwnerConflicts(csv *v1alpha1.ClusterServiceVersion) error {
```
ecordell: why don't we need to check all csvs here?

pkg/lib/ownerutil/util.go
njhale: Looks like there's no place where `false` is being passed in though. Maybe it's better to remove it altogether.

pkg/controller/operators/olm/operator_test.go
```diff
@@ -1236,8 +1238,8 @@ func TestTransitionCSV(t *testing.T) {
 						"0.0.0",
 						"",
 						installStrategy("csv1-dep1", nil, nil),
-						[]*v1beta1.CustomResourceDefinition{crd("c1", "v1", "g1")},
 						[]*v1beta1.CustomResourceDefinition{},
```
ecordell: doesn't this change from owned to required? that seems odd for this test

pkg/controller/operators/olm/operator.go
njhale: What I'm trying to say here is that we're checking two things:
1. Are resources missing?
2. Can OLM perform an action based on the errors returned?

Maybe something like "Check for any actionable APIResource issues" fits better.

pkg/controller/operators/olm/operator_test.go
njhale: Meant to remove this change.

pkg/controller/operators/olm/operator.go
njhale: We're using ownerlabels on the `APIService` itself to determine ownership. If an `APIService` that the given CSV provides is already "owned" by anything other than the given CSV or the CSV it replaces, then a conflict exists.

pkg/controller/operators/olm/operator.go
njhale: Guess I'm already querying for replacements in the method anyway huh.

pkg/controller/operators/olm/operator.go
njhale: Yup, even in the current implementation we can return transient errors that we don't want to fail on.

PR #759, opened 2019-03-14
fix(deployment): Clean up orphaned deployments

dinhxuanvu:

### DIFF
--- a/pkg/api/wrappers/deployment_install_client.go
+++ b/pkg/api/wrappers/deployment_install_client.go
@@ -8,6 +8,7 @@ import (
 	rbacv1 "k8s.io/api/rbac/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/operatorclient"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/operatorlister"
@@ -25,6 +26,7 @@ type InstallStrategyDeploymentInterface interface {
 	DeleteDeployment(name string) error
 	GetServiceAccountByName(serviceAccountName string) (*corev1.ServiceAccount, error)
 	FindAnyDeploymentsMatchingNames(depNames []string) ([]*appsv1.Deployment, error)
+	FindAnyDeploymentsMatchingLabels(label labels.Selector) ([]*appsv1.Deployment, error)
 }
 
 type InstallStrategyDeploymentClientForNamespace struct {
@@ -116,5 +118,15 @@ func (c *InstallStrategyDeploymentClientForNamespace) FindAnyDeploymentsMatching
 			}
 		}
 	}
+	return deployments, nil
+}
+
+func (c *InstallStrategyDeploymentClientForNamespace) FindAnyDeploymentsMatchingLabels(label labels.Selector) ([]*appsv1.Deployment, error) {
+	deployments, err := c.opLister.AppsV1().DeploymentLister().Deployments(c.Namespace).List(label)
+	// Any errors other than !exists are propagated up
+	if err != nil && !apierrors.IsNotFound(err) {
+		return nil, err
+	}
+
 	return deployments, nil
 }
--- a/pkg/controller/install/deployment.go
+++ b/pkg/controller/install/deployment.go
@@ -139,9 +139,15 @@ func (i *StrategyDeploymentInstaller) checkForDeployments(deploymentSpecs []Stra
 		depNames = append(depNames, dep.Name)
 	}
 
-	existingDeployments, err := i.strategyClient.FindAnyDeploymentsMatchingNames(depNames)
+	// Check the owner is a CSV
+	csv, ok := i.owner.(*v1alpha1.ClusterServiceVersion)
+	if !ok {
+		return StrategyError{Reason: StrategyErrReasonComponentMissing, Message: fmt.Sprintf("owner %s is not a CSV", i.owner.GetName())}
+	}
+
+	existingDeployments, err := i.strategyClient.FindAnyDeploymentsMatchingLabels(ownerutil.CSVOwnerSelector(csv))
 	if err != nil {
-		return StrategyError{Reason: StrategyErrReasonComponentMissing, Message: fmt.Sprintf("error querying for %s: %s", depNames, err)}
+		return StrategyError{Reason: StrategyErrReasonComponentMissing, Message: fmt.Sprintf("error querying existing deployments for CSV %s: %s", csv.GetName(), err)}
 	}
 
 	// compare deployments to see if any need to be created/updated
--- a/pkg/controller/install/deployment.go
+++ b/pkg/controller/install/deployment.go
@@ -7,6 +7,7 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	rbac "k8s.io/api/rbac/v1"
 
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/wrappers"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/ownerutil"
 )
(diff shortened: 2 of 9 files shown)


pkg/controller/operators/olm/operator.go
```diff
@@ -904,6 +904,11 @@ func (a *Operator) transitionCSVState(in v1alpha1.ClusterServiceVersion) (out *v
 		if installErr := a.updateInstallStatus(out, installer, strategy, v1alpha1.CSVPhaseInstalling, v1alpha1.CSVReasonWaiting); installErr == nil {
 			logger.WithField("strategy", out.Spec.InstallStrategy.StrategyName).Infof("install strategy successful")
 		}
+
+		// Clean up deployments in case some become orphaned
+		if cleanupErr := a.cleanupOrphanedDeployments(logger, out); cleanupErr == nil {
```
ecordell: This should work, but I think it would make sense to call this over in the deployment installer:

https://github.com/operator-framework/operator-lifecycle-manager/blob/176bf33dd455d5fb18bad3e53d1ae45853c2e329/pkg/controller/install/deployment.go#L104-L121

And looking at this now I think it would be reasonable to:

- skip parsing the previous install strategy like it currently does
- instead of what `cleanupPrevious` is currently doing, let's switch it to what you're doing in `cleanupOrphanedDeployments`
- (as a followup in the future) give the deployment installer access to the deployment cache so we don't have to reach back to the cluster for that

pkg/controller/operators/olm/operator.go
```diff
@@ -1369,3 +1374,43 @@ func (a *Operator) ensureDeploymentAnnotations(logger *logrus.Entry, csv *v1alph
...
+	existingDeployments, err := a.lister.AppsV1().DeploymentLister().Deployments(csv.GetNamespace()).List(ownerutil.CSVOwnerSelector(csv))
+	if err != nil {
+		return err
+	}
+
+	// compare existing deployments to deployments in CSV's spec to see if any need to be deleted
+	for _, d := range existingDeployments {
+		if _, exists := depNames[d.GetName()]; !exists {
+			logger.Infof("found an orphaned deployment %s in namespace %s", d.GetName(), csv.GetNamespace())
+			if err := a.OpClient.DeleteDeployment(csv.GetNamespace(), d.GetName(), &metav1.DeleteOptions{}); err != nil {
```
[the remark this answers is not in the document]
dinhxuanvu: The owner check has already been done in line #1399 above:

https://github.com/operator-framework/operator-lifecycle-manager/pull/759/files#diff-c09af0ae4ca3b7464670dbd80be66c70R1399

The `existingDeployments` should only contain deployments that belong to to a specific CSV `out` based on the `labels.selector` filter.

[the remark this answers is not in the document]
ecordell: That does query by the labels which we use for finding the objects, but for these objects (since they're in the same namespace) we add "real" owner references as well. It's better to be extra safe when GCing

pkg/controller/install/deployment.go
```diff
@@ -111,14 +112,8 @@ func (i *StrategyDeploymentInstaller) Install(s Strategy) error {
 		return err
 	}
 
-	if i.previousStrategy != nil {
-		previous, ok := i.previousStrategy.(*StrategyDetailsDeployment)
-		if !ok {
-			return fmt.Errorf("couldn't parse old install %s strategy with deployment installer", previous.GetStrategyName())
-		}
-		return i.cleanupPrevious(strategy, previous)
```
ecordell: I think you can remove `cleanupPrevious`, right?

pkg/controller/operators/olm/operator.go
```diff
@@ -906,6 +906,10 @@ func (a *Operator) transitionCSVState(in v1alpha1.ClusterServiceVersion) (out *v
 			}
 		}
 
+		// Clean up deployments in case some become orphaned
+		if cleanupErr := a.cleanupOrphanedDeployments(logger, out); cleanupErr == nil {
```
ecordell: This part can be removed now right?

pkg/api/wrappers/deployment_install_client.go
```diff
@@ -25,6 +26,7 @@ type InstallStrategyDeploymentInterface interface {
 	DeleteDeployment(name string) error
 	GetServiceAccountByName(serviceAccountName string) (*corev1.ServiceAccount, error)
 	FindAnyDeploymentsMatchingNames(depNames []string) ([]*appsv1.Deployment, error)
```
ecordell: Can we remove `FindAnyDeploymentsMatchingNames`?

pkg/api/wrappers/deployment_install_client.go
```diff
@@ -118,3 +120,13 @@ func (c *InstallStrategyDeploymentClientForNamespace) FindAnyDeploymentsMatching
 	}
 	return deployments, nil
 }
+
+func (c *InstallStrategyDeploymentClientForNamespace) FindAnyDeploymentsMatchingLabelSelector(label labels.Selector) ([]*appsv1.Deployment, error) {
```
ecordell: Can you add a unit test for this?

pkg/controller/install/deployment.go
```diff
@@ -179,3 +174,39 @@ func (i *StrategyDeploymentInstaller) checkForDeployments(deploymentSpecs []Stra
 	}
 	return nil
 }
+
```
ecordell: I think we want to call `FindAnyDeploymentsMatchingLabels` and `cleanupOrphanedDeployments` in `CheckInstalled` as well?

dinhxuanvu: `checkInstalled` is only used as status check to see if something is missing or unverifiable. I feel it is dubious to run certain cleanup func inside a func that is only intended to verify something. Also, we should only clean up orphaned deployments after new deployments have been installed.

pkg/api/wrappers/deployment_install_client.go
dinhxuanvu: We do use `FindAnyDeploymentsMatchingNames` on `checkForDeployments`.

njhale: @ecordell @dinhxuanvu I'm tempted to say we should be indexing the deployment cache by the labels we care about.

pkg/controller/install/deployment_test.go
```diff
@@ -332,3 +332,115 @@ func TestInstallStrategyDeploymentCheckInstallErrors(t *testing.T) {
...
+	}
+	tests := []struct {
+		description string
+		inputs      inputs
+		setup       setup
+		cleanupMock cleanupMock
+		output      error
+	}{
+		{
+			description: "cleanup successfully",
```
ecordell: we should be testing lots of failure cases here as well, right?

pkg/api/wrappers/deployment_install_client.go
```diff
@@ -118,3 +120,13 @@ func (c *InstallStrategyDeploymentClientForNamespace) FindAnyDeploymentsMatching
 	}
 	return deployments, nil
 }
+
+func (c *InstallStrategyDeploymentClientForNamespace) FindAnyDeploymentsMatchingLabels(label labels.Selector) ([]*appsv1.Deployment, error) {
+	deployments, err := c.opLister.AppsV1().DeploymentLister().Deployments(c.Namespace).List(label)
+	// Any errors other than !exists are propagated up
+	if !apierrors.IsNotFound(err) {
```
njhale: nit: This works, but I think it would be more conventional to do something like this:
```go
if err != nil && !apierrors.IsNotFound(err) {
   return nil, err
}
```

test/e2e/csv_e2e_test.go
```diff
@@ -2556,4 +2555,141 @@ func TestUpdateCSVMultipleVersionCRD(t *testing.T) {
...
+
+	// Should have created new deployment and deleted old
+	depNew, err := c.GetDeployment(testNamespace, strategyNew.DeploymentSpecs[0].Name)
+	require.NoError(t, err)
+	require.NotNil(t, depNew)
+	err = waitForDeploymentToDelete(t, c, strategy.DeploymentSpecs[0].Name)
+	require.NoError(t, err)
+
+	// Should clean up the CSV
+	err = waitForCSVToDelete(t, crc, csv.Name)
```
njhale: This will hang forever since the deferred delete call only executes __after__ the function returns.

njhale: This will hang forever since the deferred delete call only executes __after__ the function returns.

dinhxuanvu:
@ecordell Ready for final review.

pkg/controller/install/deployment.go
ecordell: You don't think it would be reasonable to rely on the label checks for `CheckInstalled` as well? It seems like it would be preferable to finding based on names

test/e2e/csv_e2e_test.go
```diff
@@ -2556,4 +2555,137 @@ func TestUpdateCSVMultipleVersionCRD(t *testing.T) {
...
+	strategyNewRaw, err := json.Marshal(strategyNew)
+	require.NoError(t, err)
+
+	// Fetch the current csv
+	fetchedCSV, err := fetchCSV(t, crc, csv.Name, testNamespace, csvSucceededChecker)
+	require.NoError(t, err)
+
+	// Update csv with same strategy with different deployment's name
+	fetchedCSV.Spec.InstallStrategy.StrategySpecRaw = strategyNewRaw
+
```
ecordell: Shouldn't we also add a deployment here, and check that it is _not_ deleted by the cleanup code?

ecordell:
Looks great, thanks for wrapping this up!

/lgtm
/retest

PR #761, opened 2019-03-15
Subscription steps through multiple upgrades

ecordell:
This pulls out a subset of commits from #756 that omits the additional guards around when we attempt to update (they had farther reaching effects that will take a while to test/verify).


pkg/lib/queueinformer/queueinformer_operator.go
```diff
@@ -140,10 +140,12 @@ func (o *Operator) Run(stopc <-chan struct{}) (ready, done chan struct{}, atLeve
 		o.Log.Info("starting workers...")
 		for _, queueInformer := range o.queueInformers {
 			go o.worker(queueInformer)
+			go o.worker(queueInformer)
 		}
```
jpeeler: Why is this addition and below necessary?

PR #748, opened 2019-03-11
clusteroperator: Report when OLM reaches "level" and check syncs

smarterclayton:
Cluster operators are expected to report the version of the payload
they are included in once they are "deployed", and also to keep the
cluster operator object created. Have the OLM operator keep CO up
to date, report the payload version once it hits available, and use
the count of successful syncs from the queueInformers as a probalistic
measurement of "available" (i.e. is the operator able to retire syncs).
A future change should add a "health over time" metric or a "has
successfully synced all InstallPlans at least once" metric to replace
the current estimation.

PR #762, opened 2019-03-18
Add upstream catalog

ecordell:
This adds the operatorhubio catalog as a default catalog for upstream releases.

Also added in the operatorstatus changes to the chart from #748


deploy/chart/templates/0000_50_olm_17-upstream-operators.catalogsource.yaml
```diff
@@ -0,0 +1,14 @@
...
+#! parse-kind: CatalogSource
+{{ if (.Values.installType) and eq .Values.installType "upstream" }}
+apiVersion: operators.coreos.com/v1alpha1
+kind: CatalogSource
+metadata:
+  name: operatorhubio-catalog
+  namespace: {{ .Values.catalog_namespace }}
+spec:
+  sourceType: grpc
+  image: quay.io/operator-framework/upstream-community-operators:latest
```
njhale: This is good. I just wonder if we want to inject the version/digest as part of the template instead of hard-coding `latest`.

PR #764, opened 2019-03-19
Grant namespace list to global operators

ecordell:
Operators that support both single and all namespace installation should
request the smallest scope of permissions they need. A single-namespace
operator shouldn't require any cluster-level permissions. When placing
it into an operatorgroup selecting all namespaces, it will need list
and watch on namespaces

### DIFF
--- a/test/e2e/operator_groups_e2e_test.go
+++ b/test/e2e/operator_groups_e2e_test.go
@@ -11,6 +11,7 @@ import (
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/ownerutil"
 	"github.com/stretchr/testify/require"
 	appsv1 "k8s.io/api/apps/v1"
+	"k8s.io/api/authorization/v1"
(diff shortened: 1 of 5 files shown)


test/e2e/operator_groups_e2e_test.go
```diff
@@ -1279,6 +1284,21 @@ func TestCSVCopyWatchingAllNamespaces(t *testing.T) {
 	require.EqualValues(t, "rbac.authorization.k8s.io", fetchedRoleBinding.RoleRef.APIGroup)
 	require.EqualValues(t, "ClusterRole", fetchedRoleBinding.RoleRef.Kind)
 
+	t.Log("ensure operator was granted namespace list permission")
+	res, err := c.KubernetesInterface().AuthorizationV1().SubjectAccessReviews().Create(&v1.SubjectAccessReview{
```
alecmerdler: Cool!

PR #765, opened 2019-03-19
docs(git): add changelog

njhale:
After some digging, I noticed that projects like etcd and kubernetes make use of a `CHANGELOG.md` file as the primary source of info for GH releases. Some supplemental googling lead to a wealth of info on the matter (https://keepachangelog.com/en/1.0.0/ and https://github.com/olivierlacan/keep-a-changelog/blob/master/CHANGELOG.md) and even a [tool for generating changelogs](https://github.com/github-changelog-generator/github-changelog-generator).

This PR contains the output of `github-changelog-generator` - Let me know if we can use this as-is, or if the output needs to be better curated.


njhale:
Also, the tool used to generate this assigns content to a header based on labels. This means that we can label merged PRs with labels such as "bug" and they will be listed under the "Fixed Bugs" header. It also has options for custom headers and tags.

CHANGELOG.md
```diff
@@ -0,0 +1,415 @@
+# Change Log
+
+## [0.8.1](https://github.com/operator-framework/operator-lifecycle-manager/tree/0.8.1) (2019-03-18)
```
njhale: Date needs to be updated to the day it was merged.

CHANGELOG.md
```diff
@@ -0,0 +1,415 @@
...
+[Full Changelog](https://github.com/operator-framework/operator-lifecycle-manager/compare/0.7.1...0.8.1)
+
+**Fixed bugs:**
```
njhale: Add the "bug" tag to merged bug fixes and regenerate the changelog.

CHANGELOG.md
```diff
@@ -0,0 +1,415 @@
+# Change Log
+
+## [0.8.1](https://github.com/operator-framework/operator-lifecycle-manager/tree/0.8.1) (2019-03-18)
+[Full Changelog](https://github.com/operator-framework/operator-lifecycle-manager/compare/0.7.1...0.8.1)
+
```
njhale: Add the "enhancement" tag to merged enhancements and regenerate the changelog.

PR #768, opened 2019-03-19
Add support for Windows WSL in run_console_local.sh

leszko:


leszko:
@ecordell @njhale Some test failed, but I believe they are not related to my change. Could you rerun the tests?

PR #771, opened 2019-03-20
chore(manifests): recut 0.8.1

njhale:
Cut 0.8.1 pointing at new upstream image `quay.io/operator-framework/olm`

PR #773, opened 2019-03-20
chore(makefile): fix path for quickstart

jpeeler:
This makes `make release` work.


jpeeler:
I see this is in https://github.com/operator-framework/operator-lifecycle-manager/pull/771.

PR #774, opened 2019-03-20
chore(api): bump operatorgroup and packagemanifest to v1 

ecordell:
also converges on the group `operators.coreos.com`

This changes:

- `OperatorGroups` from `v1alpha2` to `v1`
- `PackageManifests` from `v1alpha1` to `v1`
- `PackageManifests` from `packages.apps.redhat.com` to `packages.operators.coreos.com`


manifests/0000_50_olm_09-operatorgroup.crd.yaml
```diff
@@ -4,7 +4,6 @@ metadata:
   name: operatorgroups.operators.coreos.com
 spec:
   group: operators.coreos.com
-  version: v1alpha2
   versions:
   - name: v1alpha2
```
njhale: I think this needs to be regenerated.

deploy/chart/templates/0000_50_olm_08-catalog-operator.deployment.yaml
```diff
@@ -62,6 +62,11 @@ spec:
             httpGet:
               path: /healthz
               port: {{ .Values.catalog.service.internalPort }}
+          env:
+          {{ if and .Values.installType (eq .Values.installType "ocp") }}
```
njhale: @ecordell have you ever found a way to handle white-space chomping with back-to-back if statements like this? I must have tried every sensible permutation in these situations 🙁

ecordell:
I think if you use dashes in the opening and closing brackets it works out?
I can try in a follow up

deploy/chart/templates/0000_50_olm_10-operatorgroup.crd.yaml
```diff
@@ -4,11 +4,14 @@ metadata:
...
-  version: v1alpha2
+  version: v1
   versions:
-  - name: v1alpha2
-    served: true
-    storage: true
+    - name: v1
+      served: true
+      storage: true
+    - name: v1alpha2
```
njhale: At some point we're going to need to do [something like this](https://v1-12.docs.kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning/#upgrade-existing-objects-to-a-new-stored-version) to completely migrate storage versions - or we could wait for 1.14 to deprecate since there are CRD versioning behavioral differences.

manifests/0000_50_olm_07-olm-operator.deployment.yaml
```diff
@@ -25,10 +25,17 @@ spec:
           args:
           - -writeStatusName
           - operator-lifecycle-manager
+          - -tls-cert
```
njhale: Does this blow up if the certs are missing?

manifests/0000_50_olm_07-olm-operator.deployment.yaml
```diff
@@ -48,6 +55,18 @@ spec:
                 fieldPath: metadata.namespace
           - name: OPERATOR_NAME
             value: olm-operator
+          
+          volumeMounts:
+          - mountPath: /var/run/secrets/serving-cert
+            name: serving-cert
+
```
njhale: nit: I think we want to clean up the white-space chomping in the template in a future PR.

deploy/chart/templates/0000_50_olm_10-operatorgroup.crd.yaml
```diff
@@ -6,9 +6,12 @@ spec:
   group: operators.coreos.com
   version: v1
   versions:
-  - name: v1
-    served: true
-    storage: true
+    - name: v1
```
jpeeler: I assume YAML is forgiving with spacing, but it looks like the dash here isn't aligned as it was before. And the previous aligning is what the convention is.

PR #775, opened 2019-03-20
chore(deploy): set priorityclass on olm pods

ecordell:


deploy/chart/templates/0000_50_olm_06-olm-operator.deployment.yaml
```diff
@@ -39,6 +39,7 @@ spec:
           {{- end }}
           image: {{ .Values.olm.image.ref }}
           imagePullPolicy: {{ .Values.olm.image.pullPolicy }}
+          priorityClassName: "system-cluster-critical"
```
njhale: Does this `PriorityClass` exist by default in kubernetes? If not, should we wrap this with an `installType` check?

PR #791, opened 2019-03-28
fix(packageserver): add struct tags to PackageManifestList

njhale:
This fixes issues encountered in console due to #788 

Here's the console running locally against OLM with this fix:

image

### DIFF
--- a/pkg/package-server/apis/operators/v1/packagemanifest_types.go
+++ b/pkg/package-server/apis/operators/v1/packagemanifest_types.go
@@ -10,10 +10,10 @@ import (
 // PackageManifestList is a list of PackageManifest objects.
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 type PackageManifestList struct {
-	metav1.TypeMeta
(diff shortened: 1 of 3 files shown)


pkg/package-server/apis/operators/v1/packagemanifest_types.go
```diff
@@ -10,10 +10,10 @@ import (
 // PackageManifestList is a list of PackageManifest objects.
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 type PackageManifestList struct {
-	metav1.TypeMeta
-	metav1.ListMeta
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
```
alecmerdler: We don't use `omitempty` on the `ListMeta` for our other API structs, not sure if this will cause problems.

ecordell:
Screen Shot 2019-03-28 at 4 07 05 PM

Double checked and have the same result. Looks great!

/lgtm

PR #792, opened 2019-03-28
fix(annotation): don't annotate deployments that aren't owned by a CSV

ecordell:
Previously this was inadvertently annotating all deployments in a namespace. It now finds deployments via the CSV owner selector.

### DIFF
--- a/pkg/controller/operators/olm/operator.go
+++ b/pkg/controller/operators/olm/operator.go
@@ -8,7 +8,6 @@ import (
 
 	v1 "github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1"
 	"github.com/sirupsen/logrus"
-	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	extinf "k8s.io/apiextensions-apiserver/pkg/client/informers/externalversions"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


pkg/controller/operators/olm/operator.go
```diff
@@ -1435,7 +1435,9 @@ func (a *Operator) ensureDeploymentAnnotations(logger *logrus.Entry, csv *v1alph
 	// compare deployments to see if any need to be created/updated
 	existingMap := map[string]*appsv1.Deployment{}
 	for _, d := range existingDeployments {
```
njhale: Instead of creating two maps, can we use the `depName` map while iterating through the existing deployment list and check if the existing deployment name is in the map?

```go
for _, dep := range existingDeployments {
         if _, ok := depNames[dep.GetName()]; !ok {
               continue
         }
          ...
}
```

pkg/controller/operators/olm/operator_test.go
```diff
@@ -3706,35 +3757,40 @@ func TestSyncOperatorGroups(t *testing.T) {
...
+		case *v1alpha1.ClusterServiceVersion:
+			fetched, err = client.OperatorsV1alpha1().ClusterServiceVersions(namespace).Get(o.GetName(), metav1.GetOptions{})
+		case *v1.OperatorGroup:
+			fetched, err = client.OperatorsV1().OperatorGroups(namespace).Get(o.GetName(), metav1.GetOptions{})
+		default:
+			require.Failf(t, "couldn't find expected object", "%#v", object)
+		}
+		require.NoError(t, err, "couldn't fetch %s %v", namespace, object)
+		require.True(t, reflect.DeepEqual(object, fetched), diff.ObjectDiff(object, fetched))
+		// require.EqualValues(t, object, fetched, "%s in %s not equal", object.GetObjectKind().GroupVersionKind().String(), namespace)
```
njhale: Extra commented line here. Also, is there a reason we're directly using reflect instead of `EqualValues`? Diff output is better?

ecordell: I thought so, I feel like it’s a little bit better

PR #793, opened 2019-03-28
Update Kubernetes dependencies to 1.12.7

jpeeler:
Need to discuss if this is a valid approach. I've not been able to make `go get` download the desired versions without using replace in go.mod. I assume this is because of all the weird non-semver tagging in play. If using replace is okay, then it's probably best to remove the makefile vendoring stuff as it does not work well with the changes here.

The good news is that the necessary changes due to updating were very minimal. One benefit to keeping the dependencies aligned this way is that there won't be go.* file changes during normal OLM development.

OLM-892


jpeeler:
I managed to get the items in the replace block down to three things - logrus, client-go, and api. However, that makes less sense to me than doing what was done here since everything here is permanently in sync and will therefore not break.

The impossible behavior I've observed is that when client-go is fetched, it pulls in the latest k8s.io/api. Fetching other kubernetes dependencies (for whatever reason) downgrades client-go to the previous release. Both of which causes things to break and is why pinning both of them seemed to be absolutely necessary.

PR #794, opened 2019-03-28
fix(olm): generate aggregated clusterroles for ownnamespace operatorgroups correctly

ecordell:
The work is divided by commit:

1. Wrote a test to verify that the bug for generating multiple installplans when using manual approval is fixed
2. Added a SubjectAccessReview for checking permissions in allnamespace operatorgroups
3. Added a new e2e test for checking project admin permissions in OwnNamespace operatorgroups
4. Fixed the clusterrole generation for OwnNamespace operatorgroups

This fixes the issue where a user bound to `admin` in a namespace isn't able to interact with provided apis in that namespace.


njhale:
Looks good - but it's not building properly right now.

PR #796, opened 2019-04-02
fix(build) make run-local must honour NO_MINIKUBE flag

sthaha:
Previously running `NO_MINIKUBE=true make run-local` would still
start minikube as the condition ` [ -z "$NO_MINIKUBE" ] || [ -x "$(command -v minikube)" ]`
gets evaluated to `true`.

This patch fixes it by changing `||` to `&&`

### DIFF
--- a/scripts/build_local.sh
+++ b/scripts/build_local.sh
@@ -5,7 +5,7 @@
 
 set -e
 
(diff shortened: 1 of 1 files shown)


ecordell:
This is fixed in master now, thanks!

PR #799, opened 2019-04-03
chore(makefile): don't pull outdated dependencies

jpeeler:
This probably should have been part of https://github.com/operator-framework/operator-lifecycle-manager/pull/793. It just removes downloading of older kubernetes deps now that we've updated to 1.12 and are using go.mod.


Makefile
```diff
@@ -13,8 +13,6 @@ counterfeiter := $(GOBIN)/counterfeiter
 mockgen := $(GOBIN)/mockgen
 IMAGE_REPO := quay.io/operator-framework/olm
 IMAGE_TAG ?= "dev"
-KUBE_DEPS := api apiextensions-apiserver apimachinery code-generator kube-aggregator kubernetes
-KUBE_RELEASE := release-1.11
```
njhale: Did we want to just switch this to `release-1.12`?

jpeeler: No, I couldn't find a way to have client-go and go modules work together without pinning to a specific release. So no more branch tracking, it's all in go.mod.

jpeeler:
This has been superceded by https://github.com/operator-framework/operator-lifecycle-manager/pull/805.

PR #807, opened 2019-04-10
fix(catalog): recreate registry pods when deleted

njhale:
This PR addresses [ALM-902](https://jira.coreos.com/browse/OLM-902); a bug in which OLM does not recreate registry pods when they are deleted.

### TODO:

- [x] e2e test reproduces bug
- [x] pod deletion informs catalog status directly

### Bonus

As a consequence of updating the `OperatorLister` to include `CatalogSource`s, I needed to regenerate our mocks and fakes. I took the opportunity to upgrade mockgen and counterfeiter to newer versions that support go modules. Subsequently, I updated our mock helpers to reflect the change as well.

### Review Notes

Each commit represents a reasonably standalone unit of work. It will probably make sense to review commit-by-commit since some are entirely vendor updates or generated code.


njhale:
🦀 

/retest

pkg/controller/operators/catalog/operator.go
```diff
@@ -398,9 +444,9 @@ func (o *Operator) syncCatalogSources(obj interface{}) (syncError error) {
 		return fmt.Errorf("no reconciler for source type %s", catsrc.Spec.SourceType)
 	}
 
-	// if registry pod hasn't been created or hasn't been updated since the last configmap update, recreate it
+	// If registry pod hasn't been created or hasn't been updated since the last configmap update, recreate it
 	if catsrc.Status.RegistryServiceStatus == nil || catsrc.Status.RegistryServiceStatus.CreatedAt.Before(&catsrc.Status.LastSync) {
-		logger.Debug("registry server scheduled recheck")
+		logger.Debug("")
```
ecordell: this seems odd?

pkg/controller/operators/catalog/operator.go
```diff
@@ -409,7 +455,7 @@ func (o *Operator) syncCatalogSources(obj interface{}) (syncError error) {
 		logger.Debug("ensured registry server")
 
 		out.Status.RegistryServiceStatus.CreatedAt = timeNow()
-		out.Status.LastSync = timeNow()
+		out.Status.LastSync = out.Status.RegistryServiceStatus.CreatedAt
```
ecordell: 👍

pkg/controller/operators/catalog/operator.go
```diff
@@ -290,21 +297,60 @@ func (o *Operator) handleDeletion(obj interface{}) {
...
+
+	// Check the registry server
+	checker := o.checker.CheckerForSource(catsrc)
+	healthy, err := checker.CheckRegistryServer(catsrc)
+	if err != nil {
+		logger.WithError(err).Warn("error checking registry health")
+	} else if !healthy {
+		logger.Debug("registry server unhealthy, updating catalog source")
+		catsrc.Status.RegistryServiceStatus = nil
+		_, err = o.client.OperatorsV1alpha1().CatalogSources(catsrc.GetNamespace()).UpdateStatus(catsrc)
```
ecordell: So we just clear out the status, so that the reconciler knows it has work to do?

pkg/controller/registry/reconciler/reconciler.go
```diff
@@ -7,20 +7,70 @@ import (
...
+
+// RegistryCheckerFactory is a factory for RegistryCheckers.
+type RegistryCheckerFactory struct {
+	Lister               operatorlister.OperatorLister
+	OpClient             operatorclient.ClientInterface
+	ConfigMapServerImage string
+}
+
+// CheckerForSource returns a RegistryChecker based on the configuration of the given CatalogSource.
+func (r *RegistryCheckerFactory) CheckerForSource(source *v1alpha1.CatalogSource) RegistryChecker {
```
ecordell: I'm not 100% sold on the need for this, but I don't think it's worth blocking this PR

njhale: On the need for a checker, the abstraction, or the factory pattern?

pkg/controller/operators/catalog/operator.go
njhale: yes

pkg/controller/registry/reconciler/reconciler.go
ecordell: The need for a separate factory, since the reconcilers implement both interfaces

pkg/controller/operators/catalog/operator.go
njhale: thanks for catching this

PR #809, opened 2019-04-10
add logging and separate muxer for metrics

jpeeler:
This is to hopefully help debug a strange issue I've seen where https metrics are not being served upon initial pod launch. (However, if the pod is restarted, metrics are served as expected.) These changes might be good enough to keep even if they don't fix the problem, especially the mux change.

(BZ #1689836)

### DIFF
--- a/cmd/catalog/main.go
+++ b/cmd/catalog/main.go
@@ -99,22 +99,35 @@ func main() {
 	if *tlsCertPath != "" && *tlsKeyPath == "" || *tlsCertPath == "" && *tlsKeyPath != "" {
 		logger.Warn("both --tls-key and --tls-crt must be provided for TLS to be enabled, falling back to non-https")
 	} else if *tlsCertPath == "" && *tlsKeyPath == "" {
-		logger.Info("TLS keys not set, using non-https")
+		logger.Info("TLS keys not set, using non-https for metrics")
 	} else {
+		logger.Info("TLS keys set, using https for metrics")
 		useTLS = true
 	}
 
 	// Serve a health check.
-	http.HandleFunc("/healthz", func(w http.ResponseWriter, r *http.Request) {
+	healthMux := http.NewServeMux()
+	healthMux.HandleFunc("/healthz", func(w http.ResponseWriter, r *http.Request) {
 		w.WriteHeader(http.StatusOK)
 	})
-	go http.ListenAndServe(":8080", nil)
+	go http.ListenAndServe(":8080", healthMux)
 
-	http.Handle("/metrics", promhttp.Handler())
+	metricsMux := http.NewServeMux()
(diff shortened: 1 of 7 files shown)


jpeeler:
Ok, definitely leaving in these changes!

```time="2019-04-11T00:09:02Z" level=error msg="Metrics (https) serving failed: open /var/run/secrets/serving-cert/tls.crt: no such file or directory"```

Will try making the secret non-optional tomorrow, and maybe that will cause the pod to fail until the secret is ready. Best as I can tell, the secrets end up in the pod over a minute after OLM is started. Extracted from inside the OLM pod:

```
sh-4.2$ ps aux
USER        PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
1001          1  0.8  0.4  77864 67128 ?        Ssl  00:09   0:10 /bin/olm -writeStatusName operator-lifecycle-manager -tls-cert /var/run/secrets/serving-cert/tls.crt -tls-key /var/run/secrets/serving-cert/tls.key
1001         23  0.0  0.0  11828  2900 pts/0    Ss   00:27   0:00 /bin/sh
1001         34  0.0  0.0  51752  3528 pts/0    R+   00:28   0:00 ps aux
sh-4.2$ ls -l /var/run/secrets/serving-cert --full-time
total 0
lrwxrwxrwx. 1 root root 14 2019-04-11 00:10:29.460383912 +0000 tls.crt -> ..data/tls.crt
lrwxrwxrwx. 1 root root 14 2019-04-11 00:10:29.460383912 +0000 tls.key -> ..data/tls.key
```
```
sh-4.2$ ps -p 1 -o lstart
                 STARTED
Thu Apr 11 00:09:01 2019
```
Confirms metrics aren't being served, (but health is as expected):
```
sh-4.2$ lsof -i -P -n
lsof: no pwd entry for UID 1001
lsof: no pwd entry for UID 1001
COMMAND PID     USER   FD   TYPE DEVICE SIZE/OFF NODE NAME
lsof: no pwd entry for UID 1001
olm       1     1001    7u  IPv6  58483      0t0  TCP *:8080 (LISTEN)
lsof: no pwd entry for UID 1001
olm       1     1001    9u  IPv4 225600      0t0  TCP 10.128.0.9:33000->172.30.0.1:443 (ESTABLISHED)
```

jpeeler:
/retest
e2e-aws passed... looks like e2e-aws-olm had a test issue

ecordell:
Can we use fsnotify to to know when the files have actually been mounted?

jpeeler:
/test unit
Just saw the metrics test pass, so I think this is ready for LGTM again.

jpeeler:
/hold
I just realized I need to slightly modify the metrics test to not fail when metrics are being served over HTTP.

jpeeler:
/retest
CI is not behaving today.

PR #811, opened 2019-04-12
chore(cvo): report progressing=true if the version has updated

ecordell:


pkg/lib/operatorstatus/status.go
```diff
@@ -131,17 +140,24 @@ func MonitorClusterStatus(name string, syncCh chan error, stopCh <-chan struct{}
...
 						Name:    "operator",
-						Version: version,
+						Version: targetOperatorVersion,
 					},
 					{
 						Name:    "operator-lifecycle-manager",
 						Version: olmversion.OLMVersion,
 					},
 				}
+				if targetOperatorVersion != previousOperatorVersion {
```
njhale: Just to check my understanding - on cluster init, `previousOperatorVersion` _should_ be "0.0.1-snapshot" here, right?

ecordell: it should be, yes

PR #814, opened 2019-04-13
fix(unit): fix test flakes in TestTransitionCSV

jpeeler:
I can't get the unit tests in https://github.com/operator-framework/operator-lifecycle-manager/pull/809 to pass. This might not be the right fix, but can hopefully at least serve as inspiration. The other call to op.syncClusterServiceVersion is in a loop presumably to provide determinism, so that is copied here.


jpeeler:
/retest
Now that all the CI issues have been cleared up, see if this passes.

PR #816, opened 2019-04-16
Update catalog if image changes

ecordell:


ecordell:
/hold 
I have additional tests that are based on  #820, so holding until that goes in.

/retest
(to test console tests)

ecordell:
(rebased)

alecmerdler:
Same test has failed twice in a row: `TestSubscriptionUpdatesMultipleIntermediates`.

jeremyeder:
Seems like this PR has done nothing but fail unrelated tests for over a week.  Can anything be done to get around that?

ecordell:
@jeremyeder There is another PR that should fix the flake this PR keeps hitting

jeremyeder:
@alecmerdler can you re-review/re-lgtm this please?

manifests/0000_50_olm_09-aggregated.clusterrole.yaml
```diff
@@ -25,5 +25,8 @@ metadata:
     rbac.authorization.k8s.io/aggregate-to-view: "true"
 rules:
 - apiGroups: ["operators.coreos.com"]
-  resources: ["clusterserviceversions", "catalogsources", "installplans", "subscriptions", "packagemanifests"]
+  resources: ["clusterserviceversions", "catalogsources", "installplans", "subscriptions", "operatorgroups"]
```
ecordell: This fixes the ui issue where project admins can't view the olm ui.

pkg/controller/operators/catalog/operator.go
```diff
@@ -323,22 +323,6 @@ func (o *Operator) handleDeletion(obj interface{}) {
 		return
 	}
 
-	// Check the registry server
```
ecordell: Removed this - all we need is the requeue that's done right below, and the main catalog sync takes care of this.

pkg/controller/operators/catalog/operator.go
```diff
@@ -429,17 +413,22 @@ func (o *Operator) syncCatalogSources(obj interface{}) (syncError error) {
 		logger.Debug("catsrc configmap state good, checking registry pod")
 	}
 
-	reconciler := o.reconciler.ReconcilerForSource(catsrc)
-	if reconciler == nil {
+	srcReconciler := o.reconciler.ReconcilerForSource(catsrc)
```
ecordell: name was shadowing the package name

pkg/controller/operators/catalog/operator.go
```diff
@@ -429,17 +413,22 @@ func (o *Operator) syncCatalogSources(obj interface{}) (syncError error) {
...
+	if srcReconciler == nil {
 		// TODO: Add failure status on catalogsource and remove from sources
 		return fmt.Errorf("no reconciler for source type %s", catsrc.Spec.SourceType)
 	}
 
+	healthy, err := srcReconciler.CheckRegistryServer(catsrc)
```
ecordell: this is the bulk of the feature change :) everything else is fixing flakes in the tests.

pkg/controller/operators/olm/operator.go
```diff
@@ -569,13 +575,16 @@ func (a *Operator) syncClusterServiceVersion(obj interface{}) (syncError error)
...
+				syncError = updateErr
+			} else {
+				syncError = fmt.Errorf("error transitioning ClusterServiceVersion: %s and error updating CSV status: %s", syncError, updateErr)
 			}
-			syncError = fmt.Errorf("error transitioning ClusterServiceVersion: %s and error updating CSV status: %s", syncError, updateErr)
 		}
 	}
 
-	a.copyQueueIndexer.Enqueue(outCSV)
+	if !outCSV.IsUncopiable() {
```
ecordell: check for uncopiable before we enqueue, so we don't have to check on the dequeue

pkg/controller/operators/olm/operator.go
```diff
@@ -551,6 +551,12 @@ func (a *Operator) syncClusterServiceVersion(obj interface{}) (syncError error)
 	})
 	logger.Debug("syncing CSV")
 
+	if clusterServiceVersion.IsCopied() {
+		logger.Debug("skipping copied csv transition, schedule for gc check")
+		a.gcQueueIndexer.Enqueue(clusterServiceVersion)
```
ecordell: moved this here from `transitionCSV`  (b/c we can skip everything if it's copied)

pkg/controller/operators/olm/operator.go
```diff
@@ -1084,39 +1082,27 @@ func (a *Operator) transitionCSVState(in v1alpha1.ClusterServiceVersion) (out *v
 			return
 		}
 
-		// If we are a leaf, we should requeue the replacement for processing
+		// If there is a succeeded replacement, mark this for deletion
```
ecordell: It simplifies everything if we step 1 by 1 instead of marking multiple for deletion at once. This addressed a flakey test and let us delete a chunk of code.

pkg/controller/operators/olm/operator.go
```diff
@@ -1234,11 +1190,11 @@ func (a *Operator) updateInstallStatus(csv *v1alpha1.ClusterServiceVersion, inst
 }
 
 // parseStrategiesAndUpdateStatus returns a StrategyInstaller and a Strategy for a CSV if it can, else it sets a status on the CSV and returns
-func (a *Operator) parseStrategiesAndUpdateStatus(csv *v1alpha1.ClusterServiceVersion) (install.StrategyInstaller, install.Strategy, install.Strategy) {
+func (a *Operator) parseStrategiesAndUpdateStatus(csv *v1alpha1.ClusterServiceVersion) (install.StrategyInstaller, install.Strategy) {
```
ecordell: findIntermediatesForDeletion was the only thing that needed to have both strategies.

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -448,13 +462,15 @@ func (a *Operator) ensureTenantRBAC(operatorNamespace, targetNamespace string, c
 
 		// role doesn't exist, create it
 		// TODO: we can work around error cases here; if there's an un-owned role with a matching name we should generate instead
-		ownedRole.SetNamespace(targetNamespace)
-		ownedRole.SetOwnerReferences([]metav1.OwnerReference{ownerutil.NonBlockingOwner(targetCSV)})
-		if err := ownerutil.AddOwnerLabels(ownedRole, targetCSV); err != nil {
+		targetRole := ownedRole.DeepCopy()
+		targetRole.SetResourceVersion("0")
```
ecordell: the resourceversion being set was causing creates to fail

honestly not sure why this was a flake and not just a flat out failure. I meant to go back and see why this ever worked.

pkg/controller/operators/olm/requirements.go
```diff
@@ -242,7 +242,7 @@ func (a *Operator) requirementStatus(strategyDetailsDeployment *install.Strategy
 }
 
 // permissionStatus checks whether the given CSV's RBAC requirements are met in its namespace
-func (a *Operator) permissionStatus(strategyDetailsDeployment *install.StrategyDetailsDeployment, ruleChecker install.RuleChecker, csvNamespace string) (bool, []v1alpha1.RequirementStatus, error) {
+func (a *Operator) permissionStatus(strategyDetailsDeployment *install.StrategyDetailsDeployment, ruleChecker install.RuleChecker, targetNamespace, serviceAccountNamespace string) (bool, []v1alpha1.RequirementStatus, error) {
```
ecordell: another bug - the targetnamespace and serviceaccount namespace are different for multinamespace. again, not sure why this flaked instead of failing always.

(it looks like it would fail always, but resync would fix, so any test with MultiNamespace would be very slow)

test/e2e/operator_groups_e2e_test.go
```diff
@@ -178,19 +179,35 @@ func TestOperatorGroup(t *testing.T) {
 		},
 	}
 
+	// Create a new NamedInstallStrategy
+	deploymentName := genName("operator-deployment")
+	namedStrategy := newNginxInstallStrategy(deploymentName, permissions, nil)
+
+	aCSV := newCSV(csvName, opGroupNamespace, "", semver.MustParse("0.0.0"), []apiextensions.CustomResourceDefinition{mainCRD}, nil, namedStrategy)
+	createdCSV, err := crc.OperatorsV1alpha1().ClusterServiceVersions(opGroupNamespace).Create(&aCSV)
+	require.NoError(t, err)
```
ecordell: the lack of ownerlabels meant the copying would fail

test/e2e/util_test.go
```diff
@@ -369,24 +369,28 @@ func cleanupOLM(t *testing.T, namespace string) {
 	var err error
 	err = waitForEmptyList(func() (int, error) {
 		res, err := crc.OperatorsV1alpha1().ClusterServiceVersions(namespace).List(metav1.ListOptions{FieldSelector: nonPersistentCSVFieldSelector})
+		t.Logf("%d %s remaining", len(res.Items), "csvs")
```
ecordell: saw some cases where namespace cleanup failed, this helps to find those instances.

pkg/controller/operators/catalog/operator.go
```diff
@@ -914,7 +903,7 @@ func (o *Operator) createInstallPlan(namespace string, subs []*v1alpha1.Subscrip
 func (o *Operator) requeueSubscription(name, namespace string) {
 	// we can build the key directly, will need to change if queue uses different key scheme
 	key := fmt.Sprintf("%s/%s", namespace, name)
-	o.subQueue.AddRateLimited(key)
+	o.subQueue.Add(key)
```
jpeeler: Why was this change made?

pkg/controller/operators/olm/operatorgroup.go
jpeeler: This makes me think we need some kind of "SafeDeepCopy".

pkg/controller/operators/catalog/operator.go
ecordell: We shouldn't really need to rate limit our internal requeues - otherwise we can end up DOSing ourselves. In general we should only requeue when we have work to do

pkg/controller/operators/olm/operatorgroup.go
ecordell: That's a good idea - like a `CopyForCopy` vs. `CopyForUpdate`

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -418,6 +428,10 @@ func (a *Operator) ensureTenantRBAC(operatorNamespace, targetNamespace string, c
 		return err
 	}
 
+	if len(ownedRoles) == 0 {
```
jpeeler: What makes this worth checking, but not the targetCSV? Perhaps this PR obsoletes https://github.com/operator-framework/operator-lifecycle-manager/pull/837/, but the above was the biggest cache issue I saw.

PR #824, opened 2019-04-26
Add node-selector annotation to namespace

ravisantoshgudimetla:
**Why this change?**
When https://github.com/openshift/cluster-kube-apiserver-operator/pull/394 merges, all the pods running openshift cluster will have a [defaultNodeSelector](https://docs.openshift.com/container-platform/3.11/admin_guide/managing_projects.html#setting-the-cluster-wide-default-node-selector) if it has been set by cluster-admin including pods running in `openshift-*` namespace. The main advantage of having that feature is in a multi-tenant environment, any new project created can be steered towards compute nodes(non-master nodes).

**What should I do?**
If you think, the pods in your namespace shouldn't have defaultNodeSelector set, please review this PR. If not feel free to close this. The annotation added as part of this PR lets all the pods created within your project to not have the defaultNodeSelector set.

/cc @deads2k @sjenning

### DIFF
--- a/deploy/ocp/manifests/0.7.2/0000_30_00-namespace.yaml
+++ b/deploy/ocp/manifests/0.7.2/0000_30_00-namespace.yaml
@@ -6,3 +6,5 @@ metadata:
   name: openshift-operator-lifecycle-manager
   labels:
     openshift.io/run-level: "1"
+  annotations:
+    openshift.io/node-selector: ""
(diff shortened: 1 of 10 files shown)


deploy/ocp/manifests/0.7.2/0000_30_00-namespace.yaml
```diff
@@ -6,3 +6,5 @@ metadata:
   name: openshift-operator-lifecycle-manager
   labels:
     openshift.io/run-level: "1"
+  annotations:
+    openshift.io/node-selector: ""
```
ravisantoshgudimetla: @ecordell do we need to update every namespace.yaml for all releases?

ecordell: We should just need latest

ecordell:
/lgtm

we'll follow up with changes to the templates so that this won't get added to upstream releases

PR #826, opened 2019-04-26
chore(cvo): remove ClusterOperator from manifests

ecordell:


wking:
#828 is reverting this.

PR #827, opened 2019-04-26
fix(olm): Fix the issue with missing events due to rate limit

dinhxuanvu:
Event recorder has a rate limit (defaultSpamBurst) 25 events per
10 minutes. After the limit is reached, refill rate reduces to
1 event per 5 minutes which causes some events to get lost.
Currently implementation allows CSV to be stuck in Pending phase
for a long time if there are some missing requirements. As a result,
during sync cycle, CSV continues to spam `CSVReasonRequirementsNotMet`
events. Now, the code is fixed to not send any identical events that
match current state of CSV.

### DIFF
--- a/pkg/controller/operators/olm/operator.go
+++ b/pkg/controller/operators/olm/operator.go
@@ -884,7 +884,7 @@ func (a *Operator) transitionCSVState(in v1alpha1.ClusterServiceVersion) (out *v
 
 		if !met {
 			logger.Info("requirements were not met")
-			out.SetPhaseWithEvent(v1alpha1.CSVPhasePending, v1alpha1.CSVReasonRequirementsNotMet, "one or more requirements couldn't be found", now, a.recorder)
+			out.SetPhaseWithEventIfChanged(v1alpha1.CSVPhasePending, v1alpha1.CSVReasonRequirementsNotMet, "one or more requirements couldn't be found", now, a.recorder)
 			syncError = ErrRequirementsNotMet
 			return
 		}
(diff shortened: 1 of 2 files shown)


ecordell:
Nice work tracking this down!

pkg/controller/operators/olm/operator.go
```diff
@@ -884,23 +884,23 @@ func (a *Operator) transitionCSVState(in v1alpha1.ClusterServiceVersion) (out *v
...
+			out.SetPhaseWithEventIfChanged(v1alpha1.CSVPhasePending, v1alpha1.CSVReasonRequirementsNotMet, "one or more requirements couldn't be found", now, a.recorder)
 			syncError = ErrRequirementsNotMet
 			return
 		}
 
 		// Check for CRD ownership conflicts
 		if syncError = a.crdOwnerConflicts(out, a.csvSet(out.GetNamespace(), v1alpha1.CSVPhaseAny)); syncError != nil {
 			if syncError == ErrCRDOwnerConflict {
-				out.SetPhaseWithEventIfChanged(v1alpha1.CSVPhaseFailed, v1alpha1.CSVReasonOwnerConflict, syncError.Error(), now, a.recorder)
+				out.SetPhaseWithEvent(v1alpha1.CSVPhaseFailed, v1alpha1.CSVReasonOwnerConflict, syncError.Error(), now, a.recorder)
```
ecordell: should they not just all be `IfChanged`?

dinhxuanvu: We should only use `IfChanged` when the current phase is not changing (basically the possibility that we are gonna send the same event again like in the Pending phase case). Otherwise, if the phase is changing, we should the non-`IfChanged` method as intended. I don't think it is necessary to abuse the `IfChanged` method which has additional conditional checks inside that is not needed in many cases.

PR #833, opened 2019-04-29
test(e2e): wait for deployment to exist in csv replacement test

ecordell:
Fixes a flake I noticed on https://github.com/operator-framework/operator-lifecycle-manager/pull/830

Might not be the only one, so I'll append more fixes if needed

### DIFF
--- a/pkg/controller/operators/olm/operator.go
+++ b/pkg/controller/operators/olm/operator.go
@@ -396,7 +396,14 @@ func (a *Operator) handleClusterServiceVersionDeletion(obj interface{}) {
 
 	defer func(csv v1alpha1.ClusterServiceVersion) {
 		logger.Debug("removing csv from queue set")
-		a.csvQueueSet.Remove(csv.GetName(), csv.GetNamespace())
+		if err := a.csvQueueSet.Remove(csv.GetName(), csv.GetNamespace()); err != nil {
+			logger.WithError(err).Debug("error removing from queue")
+		}
+
+		if clusterServiceVersion.IsCopied() {
+			logger.Debug("deleted csv is copied. skipping operatorgroup requeue")
+			return
+		}
 
 		// Requeue all OperatorGroups in the namespace
 		logger.Debug("requeueing operatorgroups in namespace")
--- a/test/e2e/csv_e2e_test.go
+++ b/test/e2e/csv_e2e_test.go
@@ -256,6 +256,19 @@ func awaitCSV(t *testing.T, c versioned.Interface, namespace, name string, check
 	return fetched, err
 }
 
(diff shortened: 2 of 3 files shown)


test/e2e/csv_e2e_test.go
```diff
@@ -2687,6 +2700,9 @@ func TestUpdateCSVModifyDeploymentName(t *testing.T) {
 	_, err = crc.OperatorsV1alpha1().ClusterServiceVersions(testNamespace).Update(fetchedCSV)
 	require.NoError(t, err)
 
+	// Wait for new deployment to exist
+	err = waitForDeployment(t, c, strategyNew.DeploymentSpecs[0].Name)
```
soltysh: Looks like you're missing `require.NoError(t, err)` below it.

ecordell:
/retest

deleting namespaces after a test seems to be hanging :/

pkg/controller/operators/olm/operator.go
```diff
@@ -396,7 +396,14 @@ func (a *Operator) handleClusterServiceVersionDeletion(obj interface{}) {
 
 	defer func(csv v1alpha1.ClusterServiceVersion) {
 		logger.Debug("removing csv from queue set")
-		a.csvQueueSet.Remove(csv.GetName(), csv.GetNamespace())
+		if err := a.csvQueueSet.Remove(csv.GetName(), csv.GetNamespace()); err != nil {
+			logger.WithError(err).Debug("error removing from queue")
+		}
+
+		if clusterServiceVersion.IsCopied() {
```
njhale: This looks good. One structural thing is that we could close over a `var copied *bool` initialized to `false`, set it below when we check `clusterServiceVersion.IsCopied()`, and use that here to avoid checking again (not that it's very resource intensive).

pkg/controller/operators/olm/operator.go
```diff
@@ -396,7 +396,14 @@ func (a *Operator) handleClusterServiceVersionDeletion(obj interface{}) {
 
 	defer func(csv v1alpha1.ClusterServiceVersion) {
 		logger.Debug("removing csv from queue set")
-		a.csvQueueSet.Remove(csv.GetName(), csv.GetNamespace())
+		if err := a.csvQueueSet.Remove(csv.GetName(), csv.GetNamespace()); err != nil {
```
njhale: Nice! It seems like it was missing from the original, but aren't CSV keys in other queues as well (e.g. `csvCopyQueue` and `csvGCQueue`)? Should we handle removing those as well? Could we wrap the delete handlers in our queue abstractions to have them automatically dropped after they return?

ecordell: These are good ideas for a follow up :) 

I was going to add it here but we don't even expose the other queues outside of their indexers, so that would be a bigger change

PR #834, opened 2019-04-30
Support semver ranges of versions to skip in the head of a channel

ecordell:
Define a new annotation for CSVs:

```yaml
olm.skipRange: <semver range>
```
where `<semver range>` has the semver range format supported by the [semver lib](https://github.com/blang/semver#ranges) we're using.

When searching catalogs for updates, if the head of a channel has this annotation _and_ the currently installed operator has a `version` field that falls in the range, we will update to the latest entry in the channel.

The order of precedence is:

1. Channel head in the source specified by `sourceName` on the subscription, if the other criteria for skipping are met
1. The next operator that `replaces` the current one, in the source specified by `sourceName`
1. Channel head in another source that is visible to the subscription, if the other criteria for skipping are met
1. The next operator that `replaces` the current one in any source visible to the subscription

(haven't run the e2e tests locally yet b/c of my current internet connection)

### DIFF
--- a/pkg/controller/registry/resolver/querier.go
+++ b/pkg/controller/registry/resolver/querier.go
@@ -5,10 +5,14 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/blang/semver"
 	"github.com/operator-framework/operator-registry/pkg/client"
 	opregistry "github.com/operator-framework/operator-registry/pkg/registry"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/errors"
 )
+
+const SkipPackageAnnotationKey = "olm.skipRange"
 
 type SourceRef struct {
 	Address     string
@@ -21,7 +25,7 @@ type SourceQuerier interface {
 	FindProvider(api opregistry.APIKey) (*opregistry.Bundle, *CatalogKey, error)
 	FindBundle(pkgName, channelName, bundleName string, initialSource CatalogKey) (*opregistry.Bundle, *CatalogKey, error)
 	FindLatestBundle(pkgName, channelName string, initialSource CatalogKey) (*opregistry.Bundle, *CatalogKey, error)
-	FindReplacement(bundleName, pkgName, channelName string, initialSource CatalogKey) (*opregistry.Bundle, *CatalogKey, error)
+	FindReplacement(currentVersion *semver.Version, bundleName, pkgName, channelName string, initialSource CatalogKey) (*opregistry.Bundle, *CatalogKey, error)
 	Queryable() error
 }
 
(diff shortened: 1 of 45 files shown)


pkg/controller/registry/resolver/querier.go
```diff
@@ -102,24 +106,84 @@ func (q *NamespaceSourceQuerier) FindLatestBundle(pkgName, channelName string, i
...
+			errs = append(errs, err)
+		}
+
+		return nil, nil, errors.NewAggregate(errs)
 	}
 
 	for key, source := range q.sources {
-		bundle, err := source.GetReplacementBundleInPackageChannel(context.TODO(), bundleName, pkgName, channelName)
-		if err == nil {
+		bundle, err := q.findChannelHead(currentVersion, pkgName, channelName, source)
```
njhale: Should we not return the bundle if it's version matches `currentVersion`?

test/e2e/subscription_e2e_test.go
```diff
@@ -77,7 +79,7 @@ var (
 		},
 		Spec: v1alpha1.ClusterServiceVersionSpec{
 			Replaces:       "",
-			Version:        *semver.New("0.1.0"),
+			Version:        version.OperatorVersion{semver.MustParse("0.1.0")},
```
njhale: can we have a wrapper `version.MustParse(...)` function?

test/e2e/subscription_e2e_test.go
```diff
@@ -464,6 +466,70 @@ func TestCreateNewSubscriptionExistingCSV(t *testing.T) {
...
+	defer cleaner.NotifyTestComplete(t, true)
+
+	mainPackageName := genName("nginx-")
+	mainPackageStable := fmt.Sprintf("%s-stable", mainPackageName)
+	updatedPackageStable := fmt.Sprintf("%s-updated", mainPackageName)
+	stableChannel := "stable"
+	mainNamedStrategy := newNginxInstallStrategy(genName("dep-"), nil, nil)
+	mainCSV := newCSV(mainPackageStable, testNamespace, "", semver.MustParse("0.1.0-1556661347"), nil, nil, mainNamedStrategy)
+	updatedCSV := newCSV(updatedPackageStable, testNamespace, "", semver.MustParse("0.1.0-1556661832"), nil, nil, mainNamedStrategy)
+	updatedCSV.SetAnnotations(map[string]string{resolver.SkipPackageAnnotationKey: ">=0.1.0-1556661347 <0.1.0-1556661832"})
```
njhale: Is `0.1.0-1556661832` exclusive to prevent it from being picked up as a replacement after it's installed?

pkg/controller/registry/resolver/querier.go
njhale: Since we're iterating over a map, which is non-deterministic, if we have multiple catalogs with the same package/channel and CSVs at head with `olm.skipRange`s that include `currentVersion`, is it possible that we can have non-deterministic upgrades. 

Could we also get false "upgrade available" detections if the range is inclusive of head?

pkg/controller/registry/resolver/util_test.go
```diff
@@ -207,6 +211,7 @@ func crd(key opregistry.APIKey) *v1beta1.CustomResourceDefinition {
 }
 
 func u(object runtime.Object) *unstructured.Unstructured {
+	fmt.Println(object)
```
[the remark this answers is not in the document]
ecordell: oops! just for tests though

test/e2e/subscription_e2e_test.go
ecordell: yep!

PR #835, opened 2019-05-01
Fix gRPC registry pod recreation

njhale:
Addresses fixes deleted gRPC registry pod recreation; the other half of [OLM-902](https://jira.coreos.com/browse/OLM-902).
- Corrects a bad gRPC registry pod check causing them to not be recreated when missing
- Factors out common fake client decorator options into separate package
- Fixes some gRPC unit tests that were checking for `internal` catalog resources

This steals some test tools from #816 (e.g. k8s fake client decorator, simple name generator option)

### DIFF
--- a/pkg/controller/registry/reconciler/configmap.go
+++ b/pkg/controller/registry/reconciler/configmap.go
@@ -26,26 +26,36 @@ type configMapCatalogSourceDecorator struct {
 	*v1alpha1.CatalogSource
 }
 
+const (
+	// ConfigMapServerPostfix is a postfix appended to the names of resources generated for a ConfigMap server.
+	ConfigMapServerPostfix string = "-configmap-server"
+)
+
 func (s *configMapCatalogSourceDecorator) serviceAccountName() string {
-	return s.GetName() + "-configmap-server"
+	return s.GetName() + ConfigMapServerPostfix
 }
 
 func (s *configMapCatalogSourceDecorator) roleName() string {
-	return s.GetName() + "-configmap-reader"
+	return s.GetName() + ConfigMapServerPostfix
 }
 
 func (s *configMapCatalogSourceDecorator) Selector() map[string]string {
 	return map[string]string{
-		"olm.catalogSource": s.GetName(),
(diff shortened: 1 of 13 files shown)


njhale:
/hold

There's an issue with one of the postfix constant. Will fix.

pkg/controller/registry/reconciler/configmap.go
```diff
@@ -26,26 +26,36 @@ type configMapCatalogSourceDecorator struct {
...
+)
+
 func (s *configMapCatalogSourceDecorator) serviceAccountName() string {
-	return s.GetName() + "-configmap-server"
+	return s.GetName() + ConfigMapServerPostfix
 }
 
 func (s *configMapCatalogSourceDecorator) roleName() string {
-	return s.GetName() + "-configmap-reader"
+	return s.GetName() + ConfigMapServerPostfix
```
njhale: This is wrong. `ConfigMapServerPostfix` ("-configmap-server") should not be used here.

Add a new constant, `ConfigMapReaderPostfix`, set to "-configmap-reader".

pkg/controller/registry/reconciler/configmap.go
```diff
@@ -160,10 +170,15 @@ func (s *configMapCatalogSourceDecorator) Role() *rbacv1.Role {
...
+const (
+	// ConfigMapReaderPostfix is the postfix applied to the name of reader RoleBinding generated for a ConfigMap server.
+	ConfigMapReaderPostfix string = ConfigMapServerPostfix + "-reader"
+)
+
 func (s *configMapCatalogSourceDecorator) RoleBinding() *rbacv1.RoleBinding {
 	rb := &rbacv1.RoleBinding{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      s.GetName() + "-server-configmap-reader",
+			Name:      s.GetName() + ConfigMapReaderPostfix,
```
njhale: This also doesn't match.

njhale:
/hold cancel

Fixed the resource name postfixes.

PR #836, opened 2019-05-01
Set limit on length of Status.Conditions of a csv

tkashem:
Set a maximum length of Status.Conditions of ClusterServiceVersion
object(s). The oldest condition(s) should be removed from the list
as it grows over time to keep it at limit.

The default limit is set to 20.

Jira: https://jira.coreos.com/browse/OLM-963

### DIFF
--- a/pkg/api/apis/operators/v1alpha1/clusterserviceversion_test.go
+++ b/pkg/api/apis/operators/v1alpha1/clusterserviceversion_test.go
@@ -4,9 +4,11 @@ import (
 	"fmt"
 	"testing"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/api/apis/operators/v1alpha1/clusterserviceversion_test.go
```diff
@@ -282,3 +284,111 @@ func TestSupports(t *testing.T) {
 		})
 	}
 }
+
+func TestSetPhaseWithConditionsLengthAtLimit(t *testing.T) {
```
ecordell: These tests look good, but we try to keep a consistent style for tests. Generally we do a "table-drive" test style, where each test case defines expected inputs/outputs.

So for these tests, that would look something like 

```go
func TestSetPhaseLengthLimits() {
  tests := []struct{
     name string
     currentConditions []conditions
     args args
     expected expected
  } {
    // an instance for each test case
  }
for _, tt := range tests {
  t.Run(tt.name, func(t *testing.T) {
    // set csv conditions to tt.currentConditions
    // call csv.SetPhase(args)
    // check csv conditions == expected
  }
}
```

tkashem: @ecordell  converted to table driven, it looks concise and more readable now. Can you take a look?

PR #837, opened 2019-05-01
Refactor to avoid cache races

jpeeler:
This is essentially a fix for unit tests that I noticed locally. It involves making sure a newly created CSV is not attempted to be retrieved using a lister within the same sync loop.

Obsoletes https://github.com/operator-framework/operator-lifecycle-manager/pull/814.

As much as I'd like to say this fixes every flake, this might only fix one of them. But I haven't seen any locally with this fix.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,7 @@ IMAGE_REPO := quay.io/operator-framework/olm
 IMAGE_TAG ?= "dev"
 KUBE_DEPS := api apiserver apimachinery apiextensions-apiserver kube-aggregator code-generator cli-runtime
 KUBE_RELEASE := release-1.12
+SPECIFIC_UNIT_TEST := $(if $(TEST),-run $(TEST),)
 
 .PHONY: build test run clean vendor schema-check \
 	vendor-update coverage coverage-html e2e .FORCE
@@ -24,7 +25,7 @@ all: test build
 test: clean cover.out
 
 unit:
-	go test $(MOD_FLAGS) -v -race -tags=json1 -count=1 ./pkg/...
+	go test $(MOD_FLAGS) $(SPECIFIC_UNIT_TEST) -v -race -tags=json1 -count=1 ./pkg/...
 
 schema-check:
 
--- a/pkg/controller/operators/olm/operator.go
+++ b/pkg/controller/operators/olm/operator.go
@@ -582,6 +582,23 @@ func (a *Operator) syncClusterServiceVersion(obj interface{}) (syncError error) 
 		}
 	}
 
+	operatorGroup := a.operatorGroupFromAnnotations(logger, clusterServiceVersion)
+	if operatorGroup == nil {
+		logger.WithField("reason", "no operatorgroup found for active CSV").Debug("skipping potential RBAC creation in target namespaces")
+		return
+	}
+
+	if len(operatorGroup.Status.Namespaces) == 1 && operatorGroup.Status.Namespaces[0] == operatorGroup.GetNamespace() {
+		logger.Debug("skipping copy for OwnNamespace operatorgroup")
+		return
+	}
+	// Ensure operator has access to targetnamespaces with cluster RBAC
+	// (roles/rolebindings are checked for each target namespace in syncCopyCSV)
+	if err := a.ensureRBACInTargetNamespace(clusterServiceVersion, operatorGroup); err != nil {
+		logger.WithError(err).Info("couldn't ensure RBAC in target namespaces")
+		syncError = err
+	}
+
 	if !outCSV.IsUncopiable() {
 		a.copyQueueIndexer.Enqueue(outCSV)
 	}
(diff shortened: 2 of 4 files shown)


pkg/controller/operators/olm/operator.go
```diff
@@ -568,6 +568,22 @@ func (a *Operator) syncClusterServiceVersion(obj interface{}) (syncError error)
...
+		return
+	}
+
+	if len(operatorGroup.Status.Namespaces) == 1 && operatorGroup.Status.Namespaces[0] == operatorGroup.GetNamespace() {
+		logger.Debug("skipping copy for OwnNamespace operatorgroup")
+		return
+	}
+	// Ensure operator has access to targetnamespaces with cluster RBAC
+	// (roles/rolebindings are checked for each target namespace in syncCopyCSV)
+	if err := a.ensureRBACInTargetNamespace(clusterServiceVersion, operatorGroup); err != nil {
```
ecordell: I'm sure moving this here fixes the unit test flakes, but the reason we needed to move this into the syncCopyCSV in the first place was because of performance issues with large numbers of namespaces (well, not that large - 50 or so).

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -508,25 +486,65 @@ func (a *Operator) ensureCSVsInNamespaces(csv *v1alpha1.ClusterServiceVersion, o
...
 			}
 		}
 	}
 
+	targetNamespaces := operatorGroup.Status.Namespaces
+	if targetNamespaces == nil {
+		a.Log.Errorf("operatorgroup '%v' should have non-nil status", operatorGroup.GetName())
+		return nil
+	}
+	for _, ns := range targetNamespaces {
```
ecordell: Does moving this here fix anything in particular? Don't we call:

- ensureCSVsInNamespaces
- ensureTenantRBAC

so this change is basically folding them into one?

pkg/controller/operators/olm/operator.go
jpeeler: This method has been changed to only handle cluster roles and bindings. Do you still think this would be a problem?

pkg/controller/operators/olm/operatorgroup.go
jpeeler: ensureTenantRBAC has been modified to not use the lister for retrieving the CSV, which is made possible by copyToNamespace returning the CSV.
Also, related to your other point, this keeps RBAC operations for anything other than all namespaces to stay in the syncCopyCSV loop.

jpeeler:
/retest
I didn't see any e2e failures locally, but I'll be slightly concerned if they don't pass this time. (And maybe that speaks for itself...)

jpeeler:
/retest
now that https://github.com/operator-framework/operator-lifecycle-manager/pull/838 has merged, e2e should pass.

jpeeler:
/retest
Cluster didn't come up.

Makefile
```diff
@@ -24,7 +25,7 @@ all: test build
 test: clean cover.out
 
 unit:
-	go test $(MOD_FLAGS) -v -race -tags=json1 -count=1 ./pkg/...
+	go test $(MOD_FLAGS) $(SPECIFIC_UNIT_TEST) -v -race -tags=json1 -count=1 ./pkg/...
```
ecordell: Awesome, thanks!

pkg/controller/operators/olm/operator.go
ecordell: Gotcha, thanks - I think that makes this fine :)

ecordell:
@jpeeler this needs a rebase, and thanks for explaining these changes to me :)

I'd like to see the tests pass and then play with it in a real cluster just to check some of the performance characteristics, but otherwise it looks good to me.

ecordell:
/hold

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -468,7 +468,7 @@ func (a *Operator) ensureTenantRBAC(operatorNamespace, targetNamespace string, c
 		if !ownerutil.IsOwnedBy(ownedRoleBinding, csv) {
 			continue
 		}
-		_, ok := targetRolesByName[ownedRoleBinding.GetName()]
+		_, ok := targetRoleBindingsByName[ownedRoleBinding.GetName()]
```
jpeeler: I'm not sure about this change, but it looks right.

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -583,7 +591,7 @@ func (a *Operator) copyToNamespace(csv *v1alpha1.ClusterServiceVersion, namespac
 			fetchedCSV.SetLabels(utillabels.AddLabel(fetchedCSV.GetLabels(), v1alpha1.CopiedLabelKey, csv.GetNamespace()))
 			// CRs don't support strategic merge patching, but in the future if they do this should be updated to patch
 			logger.Debug("updating target CSV")
-			if _, err := a.client.OperatorsV1alpha1().ClusterServiceVersions(namespace).Update(fetchedCSV); err != nil {
+			if fetchedCSV, err = a.client.OperatorsV1alpha1().ClusterServiceVersions(namespace).Update(fetchedCSV); err != nil {
```
jpeeler: This change and the one below is just getting a more refreshed copy of the CSV. But the way it's written to overwrite the copy from the lister may not be acceptable?

jpeeler:
Dear reviewer, please verify ensureTenantRBAC is right. Otherwise, I think everything else is good.

ecordell:
/lgtm

🎉

PR #838, opened 2019-05-01
fix(resolver): fixes a bug where resolved dependent subscriptions don't

ecordell:
detect updates to catalogs

this occurred because we only check for updates when we have information
about the currently installed CSV on the subscription status, which
was omitted from dependent catalogs.

the resolver now records the csv resolved into the subscription spec,
and it is used to populate the installed csv status when available.


pkg/controller/operators/olm/operator.go
```diff
@@ -1093,6 +1112,10 @@ func (a *Operator) transitionCSVState(in v1alpha1.ClusterServiceVersion) (out *v
 		}
 	case v1alpha1.CSVPhaseDeleting:
 		var immediate int64 = 0
+
+		if err := a.csvQueueSet.Remove(out.GetNamespace(), out.GetName()); err != nil {
```
njhale: The name and namespace parameters are reversed here.

njhale: Doesn't the following delete step result in the key being enqueued again?

ecordell: That explains some logs I saw :) If this is the only thing I will follow up with that fix.

ecordell: I have also fixed in #841

PR #842, opened 2019-05-03
WIP: turn on debugging on all deployments

jpeeler:
This is to debug upgrades.


ecordell:
@jpeeler you don't need this anymore, right?

jpeeler:
nope!

PR #846, opened 2019-05-07
Multiple fixes

ecordell:
This only has one more commit on top of #816, but didn't want to mess up the tests on the other PR.

This includes what I believe to be a fix to the reports of subscriptions occasionally getting stuck. Prior to this change, I could run the `TestInstallPlanWithCSVsAcrossMultipleCatalogSources` 10 times and consistently get a failure with a stuck subscription. After this change, resolution always succeeds. (I ran the test 10 times, twice)

If #816 merged I will stage a PR with just the final commit against master and close this.


pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -544,6 +544,10 @@ func (a *Operator) ensureCSVsInNamespaces(csv *v1alpha1.ClusterServiceVersion, o
 }
 
 func (a *Operator) copyToNamespace(csv *v1alpha1.ClusterServiceVersion, namespace string) error {
+	if csv.GetNamespace() == namespace {
```
jpeeler: This is just defensive, right?

ecordell:
closing because I was able to reproduce again :(

PR #847, opened 2019-05-07
fix(catalog): fix issue where subscriptions sometimes get "stuck"

ecordell:
we were not resetting the client when updating a catalogsource, which
meant it was possible for the client to be stale and never attempt
a reconnect if it didn't go unhealthy "in time" for us to detect and
reconnect.

I ran `InstallPlanWithCSVsAcrossMultipleCatalogSources` 20 times in a row to verify it no longer flakes (previously it would error within ~5 tries)

### DIFF
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -446,6 +446,12 @@ func (o *Operator) syncCatalogSources(obj interface{}) (syncError error) {
 		o.sourcesLastUpdate = timeNow()
 		logger.Debug("registry server recreated")
 
+		func() {
+			o.sourcesLock.Lock()
+			defer o.sourcesLock.Unlock()
+			delete(o.sources, sourceKey)
+		}()
+
 		return nil
 	}
 	logger.Debug("registry state good")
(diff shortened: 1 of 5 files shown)


pkg/controller/operators/catalog/operator.go
```diff
@@ -446,6 +446,12 @@ func (o *Operator) syncCatalogSources(obj interface{}) (syncError error) {
 		o.sourcesLastUpdate = timeNow()
 		logger.Debug("registry server recreated")
 
+		func() {
```
ecordell: this is the fix, everything else is small things I noticed when reviewing

chancez: Any reason you use a closure instead of just lock/unlock without defer?

```
			o.sourcesLock.Lock()
			delete(o.sources, sourceKey)
			o.sourcesLock.Unlock()
```

ecordell: it's purely to keep the lock/unlock always next to each other, so that in the future if we need to do additional work with `sources` here it's harder to make a mistake

chancez: Cool, just curious if I could learn something here, thanks.

eparis:
/retest
hard to imagine this brought down an apiserver and broke etcd   :)

PR #853, opened 2019-05-08
deploy: fix tolerations again

sjenning:
@ravisantoshgudimetla 

/hold
until we make absolutely sure it is fixed this time. manifests are _everywhere_.

### DIFF
--- a/deploy/ocp/manifests/0.9.0/0000_50_olm_07-olm-operator.deployment.yaml
+++ b/deploy/ocp/manifests/0.9.0/0000_50_olm_07-olm-operator.deployment.yaml
@@ -62,18 +62,24 @@ spec:
           volumeMounts:
           - mountPath: /var/run/secrets/serving-cert
             name: serving-cert
-          
-      
       volumes:
       - name: serving-cert
         secret:
           secretName: olm-operator-serving-cert
           optional: true
-      
       nodeSelector:
         beta.kubernetes.io/os: linux
         node-role.kubernetes.io/master: ""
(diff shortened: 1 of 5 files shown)


ecordell:
don't worry about these - they'll get synced from the chart when we cut the next release.

all we need to care about for ocp right now is what's in `/manifests`

sjenning:
soooo... do they not take effect until you cut a release.  i.e. CI the release that claims to capture the fix actually does not?
https://openshift-release.svc.ci.openshift.org/releasestream/4.1.0-0.ci/release/4.1.0-0.ci-2019-05-08-172654

sjenning:
because i deployed a `4.1.0-0.ci-2019-05-08-172654` cluster at the tolerations for `olm-operators` are still wrong
```
$ oc get pod -ojson olm-operators-8rc8q | jq .spec.tolerations
[
  {
    "operator": "Exists"
  }
]
```

ecordell:
our "release cutting" has nothing to do with what the insaller deploys. it's for upstream users.

olm-operators is a pod that is generated by olm, it's not part of the operator. Does that also need to be changed?

if so, it's here:
https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/controller/registry/reconciler/configmap.go#L131-L134
https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/controller/registry/reconciler/grpc.go#L93-L97

but I thought this was supposed to only apply to the operator pods.

PR #856, opened 2019-05-10
test(olm): test role aggregation for aggregate apiservices

ecordell:
We were testing that CRD access was being aggregated to the `admin` cluster role; this adds an additional service to the namespace provided by an aggregated apiserver to test that as well.

### DIFF
--- a/test/e2e/operator_groups_e2e_test.go
+++ b/test/e2e/operator_groups_e2e_test.go
@@ -1,6 +1,7 @@
 package e2e
 
 import (
+	"encoding/json"
 	"fmt"
 	"strings"
 	"testing"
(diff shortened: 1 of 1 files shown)


test/e2e/operator_groups_e2e_test.go
```diff
@@ -496,10 +497,82 @@ func TestOperatorGroupRoleAggregation(t *testing.T) {
...
+				Owned: owned,
+			},
+		},
+	}
+	csvB.SetName(depName)
+
+	// Create the APIService CSV
+	_, err = createCSV(t, c, crc, csvB, nsA, false, true)
+	require.NoError(t, err)
+	// defer cleanupCSV()
```
tkashem: We don't need to clean up the CSV?

ecordell: Whoops!

PR #857, opened 2019-05-10
[WIP] fix(registry): Validate alm-examples from registry

dinhxuanvu:
Validate alm-examples for CSV from registry to ensure it is parsable
and match provided APIs

### DIFF
--- a/pkg/controller/registry/resolver/operators.go
+++ b/pkg/controller/registry/resolver/operators.go
@@ -1,6 +1,7 @@
 package resolver
 
 import (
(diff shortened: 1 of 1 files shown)


ecordell:
Thanks @dinhxuanvu - I think we should probably validate these on the operator-registry side, right?

(we can validate them both places, but if we do it at runtime like this I think the console should do it; we don't actually parse / use the examples in OLM for anything)

dinhxuanvu:
Move the work to operator-registry

PR #860, opened 2019-05-15
(refactor) Move csv set and replace to a package

tkashem:
Promote csv set and replace logic within olm operator to a reusable
package in lib so that it can be reused by other units.

Note:
- No behavior of the operator has changed as part of this refactor.

### DIFF
--- a/pkg/controller/operators/olm/operator_test.go
+++ b/pkg/controller/operators/olm/operator_test.go
@@ -61,6 +61,7 @@ import (
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/ownerutil"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/queueinformer"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/metrics"
+	csvutility "github.com/operator-framework/operator-lifecycle-manager/pkg/lib/csv"
 )
 
 // Fakes
@@ -163,16 +164,24 @@ func NewFakeOperator(clientObjs []runtime.Object, k8sObjs []runtime.Object, extO
 	}
 
 	// Create the new operator
-	queueOperator, err := queueinformer.NewOperatorFromClient(opClientFake, logrus.StandardLogger())
+	logger := logrus.StandardLogger()
+	queueOperator, err := queueinformer.NewOperatorFromClient(opClientFake, logger)
+
+	lister := operatorlister.NewLister()
+	csvSetGenerator := csvutility.NewSetGenerator(logger, lister)
+	csvReplaceFinder := csvutility.NewReplaceFinder(logger, clientFake)
+
 	op := &Operator{
-		Operator:      queueOperator,
-		client:        clientFake,
-		resolver:      strategyResolver,
-		apiReconciler: apiReconciler,
-		lister:        operatorlister.NewLister(),
-		csvQueueSet:   queueinformer.NewEmptyResourceQueueSet(),
-		recorder:      eventRecorder,
-		apiLabeler:    apiLabeler,
+		Operator:         queueOperator,
+		client:           clientFake,
+		resolver:         strategyResolver,
+		apiReconciler:    apiReconciler,
+		lister:           lister,
+		csvQueueSet:      queueinformer.NewEmptyResourceQueueSet(),
+		recorder:         eventRecorder,
+		apiLabeler:       apiLabeler,
+		csvSetGenerator:  csvSetGenerator,
+		csvReplaceFinder: csvReplaceFinder,
 	}
 
 	wakeupInterval := 5 * time.Minute
@@ -216,7 +225,7 @@ func NewFakeOperator(clientObjs []runtime.Object, k8sObjs []runtime.Object, extO
 	}
 	// Register separate queue for copying csvs
 	csvCopyQueue := workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "csvCopy")
-	csvQueueIndexer := queueinformer.NewQueueIndexer(csvCopyQueue, csvIndexes, op.syncCopyCSV, "csvCopy", logrus.StandardLogger(), metrics.NewMetricsNil())
+	csvQueueIndexer := queueinformer.NewQueueIndexer(csvCopyQueue, csvIndexes, op.syncCopyCSV, "csvCopy", logger, metrics.NewMetricsNil())
 	op.RegisterQueueIndexer(csvQueueIndexer)
 	op.copyQueueIndexer = csvQueueIndexer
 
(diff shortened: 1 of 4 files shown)


tkashem:
I forgot to change `NewFakeOperator` and default `&Operator{}` instances in `olm/operator_test.go`. Fixed it now. Can you check again when you have some time @ecordell

tkashem:
> keep trying to think of a better name than "ReplaceFinder"

Yeah, I could not find a better name either. :)

pkg/lib/csv/csvset.go
```diff
@@ -0,0 +1,55 @@
...
+import (
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/operatorlister"
+	"github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/labels"
+)
+
+// NewSetGenerator returns a new instance of SetGenerator.
+func NewSetGenerator(logger *logrus.Logger, lister operatorlister.OperatorLister) SetGenerator {
+	return &csvSet{
```
ecordell: It might make sense to make this an exported type

pkg/lib/csv/replace_finder.go
```diff
@@ -0,0 +1,65 @@
...
+type replace struct {
+	logger *logrus.Logger
+	client versioned.Interface
+}
+
+// IsBeingReplaced returns the corresponding ClusterServiceVersion object that
+// is replacing the given CSV specified.
+//
+// If the corresponding ClusterServiceVersion is not found nil is returned.
+func (r *replace) IsBeingReplaced(in *v1alpha1.ClusterServiceVersion, csvsInNamespace map[string]*v1alpha1.ClusterServiceVersion) (replacedBy *v1alpha1.ClusterServiceVersion) {
```
ecordell: if the CSVSet type were exported, could just reference it here

pkg/controller/operators/olm/operator_test.go
```diff
@@ -163,15 +164,22 @@ func NewFakeOperator(clientObjs []runtime.Object, k8sObjs []runtime.Object, extO
 
 	// Create the new operator
 	queueOperator, err := queueinformer.NewOperatorFromClient(opClientFake, logrus.StandardLogger())
+
+	lister := operatorlister.NewLister()
+	csvSetGenerator := csvutility.NewSetGenerator(logrus.StandardLogger(), lister)
```
ecordell: if you're passing in the logger here, you should construct it ahead of time and pass the same instance to `NewOperator`, `NewSetGenerator`, and `NewReplaceFinder` so that when we pass in configuration (i.e. in the e2e tests) it gets distributed everywhere

pkg/lib/csv/replace_finder.go
```diff
@@ -0,0 +1,65 @@
...
+	for _, csv := range csvsInNamespace {
+		r.logger.Infof("checking %s", csv.GetName())
+		if csv.Spec.Replaces == in.GetName() {
+			r.logger.Infof("%s replaced by %s", in.GetName(), csv.GetName())
+			replacedBy = csv.DeepCopy()
```
ecordell: It would be nice to review where we're calling this and see if the copy is really necessary. This can be quite a lot of objects in a larger cluster.

Another thought: we might want to filter out Copied CSVs here - but again that will require a review of where we call this.

pkg/lib/csv/replace_finder.go
tkashem: You mean, instead of passing the map `csvsInNamespace map[string]*v1alpha1.ClusterServiceVersion`, call `csvSet.WithNamespace` from within the function to get the list of namespaces?  

I thought that the original intention was to pass the map as a parameter so that it does not need to know how to get it, and I wanted to keep that loose coupling.

I also was thinking of the following use case for my next PR on cluster operator
```
selector := labels.SelectorFromSet(labels.Set{
  "olm.clusteroperator.name": "packageserver",
})
related := csvSet.WithNamespaceAndLabels(in.GetNamespace(), v1alpha1.CSVPhaseAny, selector)
replacedBy := replaceFinder.IsBeingReplaced(in, related)
```
Thoughts?

pkg/controller/operators/olm/operator_test.go
tkashem: sounds good, I will make the change. I noticed that `logrus.StandardLogger()` was being called  every time for a logger inside the function so went ahead with that.

pkg/lib/csv/replace_finder.go
tkashem: I was thinking whether deep copy was necessary for this but didn't spend time since I thought it would be outside the scope of this PR.
But I am happy to do the search and see whether we can filter out copied CSVs and not deep copy.

tkashem:
@ecordell please review my comments when you have time.

pkg/lib/csv/replace_finder.go
tkashem: @ecordell I was going to make this change but then came across this - https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/controller/operators/olm/operator_test.go#L4321

pkg/lib/csv/replace_finder.go
```diff
@@ -0,0 +1,70 @@
...
+	client versioned.Interface
+}
+
+// IsBeingReplaced returns the corresponding ClusterServiceVersion object that
+// is replacing the given CSV specified.
+//
+// If the corresponding ClusterServiceVersion is not found nil is returned.
+func (r *replace) IsBeingReplaced(in *v1alpha1.ClusterServiceVersion, csvsInNamespace map[string]*v1alpha1.ClusterServiceVersion) (replacedBy *v1alpha1.ClusterServiceVersion) {
+	for _, csv := range csvsInNamespace {
+		if csv.IsCopied() {
```
tkashem: I checked the call sites of this function, it just requeues both the given CSV and the one replacing it `old -> next` if `next.Status.Phase ~= v1alpha1.CSVPhaseSucceeded`. Let me know if we don't need to requeue copied CSV here.

https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/controller/operators/olm/operator.go#L1094

pkg/lib/csv/replace_finder.go
```diff
@@ -0,0 +1,70 @@
...
+			continue
+		}
+
+		r.logger.Infof("checking %s", csv.GetName())
+
+		if csv.Spec.Replaces == in.GetName() {
+			r.logger.Infof("%s replaced by %s", in.GetName(), csv.GetName())
+			replacedBy = csv
```
tkashem: No more deep copying - I removed `replacedBy = csv.DeepCopy()`. The call sites do not modify the `next` CSV.

PR #863, opened 2019-05-19
Package API server monitoring

tkashem:
Communicate package API server deployment progress back to cluster
version operator (cvo). Accomplish this goal by doing the following:

- Add a label to package API server CSV so that we can identify it.
  olm.clusteroperator.name: operator-lifecycle-manager-packageserver

- olm operator already watches all CSV(s). Add support to tap into
  these reconciliation notifications so that we can inspect all
  ClusterServiceVersion object(s).

- Add logic so that we can identify the CSV we want. Any CSV having
  the label mentioned above is a matching CSV. Whenever we see a
  matching CSV, we find the relevant information and we send it to
  the monitor.

- Write a monitor that runs forever. It receives the notification and
  updates the clusteroperator resource accordingly.


deploy/chart/templates/0000_50_olm_15-operatorstatus.yaml
```diff
@@ -16,4 +16,15 @@ status:
...
+{{- if .Values.writePackageServerStatusName }}
+---
+apiVersion: config.openshift.io/v1
+kind: ClusterOperator
+metadata:
+  name: {{ .Values.writePackageServerStatusName }}
+status:
+  versions:
+    - name: operator
+      version: {{ .Chart.Version }}
```
tkashem: Is this the right version?

pkg/controller/operators/olm/operator.go
```diff
@@ -370,6 +371,14 @@ func (a *Operator) GetReplaceFinder() csvutility.ReplaceFinder {
 	return a.csvReplaceFinder
 }
 
+func (a *Operator) RegisterCSVWatchNotification(csvNotification csvutility.WatchNotification) {
```
tkashem: For now this supports a single client, in the future we can extend it to support multiple clients.

pkg/lib/operatorstatus/builder.go
```diff
@@ -0,0 +1,152 @@
...
+func NewBuilder(clock clock.Clock) *Builder {
+	return &Builder{
+		clock: clock,
+	}
+}
+
+// Builder helps build ClusterOperatorStatus with appropriate
+// ClusterOperatorStatusCondition and OperandVersion.
+type Builder struct {
+	clock  clock.Clock
```
njhale: Nice! This seems cleaner than the `timeNow(...)`. Maybe we should do this for our `Operator`s as well.

pkg/lib/csv/notification.go
```diff
@@ -0,0 +1,17 @@
+package csv
+
+import (
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
+)
+
+// WatchNotification is an sink interface that can be used to get notification
+// of CSV reconciliation request(s) received by the operator.
+type WatchNotification interface {
```
njhale: This seems to overlap with [`cache.ResourceEventHandler`](https://godoc.org/k8s.io/client-go/tools/cache#ResourceEventHandler), multiple of which can be [registered to an informer](https://github.com/kubernetes/client-go/blob/7b18d6600f6b0022e31c46b46875beffd85cc71a/tools/cache/shared_informer.go#L326). Adding a "CVO notifying" `cache.ResourceEventHandler` to our [CSV informers in `NewOperator(...)`](https://github.com/operator-framework/operator-lifecycle-manager/blob/0a281bc71f56fc25cab8f6e318499160652e97b6/pkg/controller/operators/olm/operator.go#L264) might look something like:

```go
// ...
csvInformer := sharedInformerFactory.Operators().V1alpha1().ClusterServiceVersions()
// Add handler for updating ClusterStatus on packageserver CSV changes
csvInformer.Informer().AddEventHandler(op.packageServerClusterStatusHandler)
// ...
```

This gives us a single point of entry for notifying CVO, rather than adding update calls to sync handlers individually. I may not be getting the full picture here though, do you think using `cache.ResourceEventHandler` makes sense?

tkashem: This is the requirement:
We need to be notified of all CSV resource(s) being watched by olm operator. What we need is an interface that injects an `*v1alpha1.ClusterServiceVersion` object whenever it is being reconciled. 

Given this, I thought tapping into `syncClusterServiceVersion` would allow us to send these notification(s) without knowing how the informers/indexers/cache are being set up. This way we can keep the notification logic decoupled from the informers. We also can afford a statically `typed` interface here, the object we receive is of `*v1alpha1.ClusterServiceVersion` type. This also makes it limited to `ClusterServiceVersion` type only as you mentioned.

I am still learning about informer stuffs, but I thought `cache.ResourceEventHandler` funcs sometimes may provide you with the key of the object rather than the object itself.

We can use `cache.ResourceEventHandler` interface to make it applicable to all types. Thinking out loud here, let's say the `Operator` exposes a function `RegisterWatchedResourceNotification(obj interface{}, handlers ResourceEventHandler)`.
`obj` is they type of resources we want to get notification of, it could be `Subscription` or `ClusterServiceVersion`. 

I am happy to make it more generic, thinking whether to do it as part of this PR or a separate one.
Thoughts @njhale?

jpeeler: I was kind of thinking the same thing. I don't know that a statically typed interface outweighs having all the event watcher logic in one place. But I'm fairly certain that a follow up PR for such a change is fine.

njhale: @tkashem I see where you're coming from and I definitely think using a statically typed interface is a good thing for high-level consumers - this also lines up with [some of our thoughts for status updates](https://docs.google.com/document/d/1ROc7kDkZE57tS_Ttx5xf3_aIQLVbMsckg3YSQnVvFvU/edit?usp=sharing).

> I thought cache.ResourceEventHandler funcs sometimes may provide you with the key of the object rather than the object itself.

Correct, on add and update a key is given that does not include the new object. On Delete events, the `cache.ResourceEventHandler` receives a "tombstone" that contains the object - Our current `QueueInformer` abstractions handle [enqueuing keys given to a default event handler onto a workqueue](https://github.com/operator-framework/operator-lifecycle-manager/blob/41821c7a460ea13ee421a57f4d8cc4bb1a356a64/pkg/lib/queueinformer/queueinformer.go#L62) and [consuming keys from that queue](https://github.com/operator-framework/operator-lifecycle-manager/blob/41821c7a460ea13ee421a57f4d8cc4bb1a356a64/pkg/lib/queueinformer/queueinformer_operator.go#L175). They also use those keys to look up resources in the backing cache and [feeds them to a registered `syncHandler` function](https://github.com/operator-framework/operator-lifecycle-manager/blob/41821c7a460ea13ee421a57f4d8cc4bb1a356a64/pkg/lib/queueinformer/queueinformer_operator.go#L205).

I definitely think there's room for improvement in what we currently have. I just want to be careful of introducing more parallel abstractions. I would like to consolidate your ideas, `QueueInformer`, and our other thoughts, and come up with something small, sharp, and simple that we can apply across the board. This can definitely be an incremental thing as you and @jpeeler have suggested.

tkashem: @njhale We are are in agreement. I did see room for improvement on how we currently build our operator instances. I decided not to partake into any refactoring, I just did the minimum for my PR to work :). 

I skimmed through your refactor on the operator builder, I think we are heading in the right direction.

pkg/lib/operatorstatus/monitor.go
```diff
@@ -0,0 +1,221 @@
...
+	Deleted bool
+}
+
+// Monitor is an interface that wraps the Run method.
+//
+// Run runs forever, it reads from an underlying notification channel and
+// updates an clusteroperator resource accordingly.
+// If the specified stop channel is closed the loop must terminate gracefully.
+type Monitor interface {
+	Run(stopCh <-chan struct{})
```
njhale: I wonder if we want to take a [context.Context](https://golang.org/pkg/context/) here instead.

pkg/lib/operatorstatus/monitor.go
```diff
@@ -0,0 +1,221 @@
...
+	for _, name := range m.names {
+		if err := m.init(name); err != nil {
+			m.logger.Errorf("initialization error - %v", err)
+			break
+		}
+	}
+
+	for {
+		select {
+		case context := <-m.notificationCh:
```
njhale: It almost seems like `monitor` itself is another operator. I wonder what this would look like using the `QueueInformer` abstractions. Potentially, this could be another `SyncHandler` registered to the CSV `QueueInformer`. When a `SyncHandler` fails for an `Operator`, the sync is "re-queued" and tried again (up to a certain point). Could that be useful here?

pkg/lib/operatorstatus/monitor.go
```diff
@@ -0,0 +1,221 @@
...
+	}
+
+	select {
+	case m.notificationCh <- context:
+	default:
+		m.logger.Warn("monitor not ready to receive")
+	}
+}
+
+func (m *monitor) Run(stopCh <-chan struct{}) {
```
njhale: nit: Do we want to use [`context.Context`](https://golang.org/pkg/context/) here instead?

pkg/lib/operatorstatus/handler.go
```diff
@@ -0,0 +1,118 @@
...
+// b. If this is the right CSV then send it to the monitor.
+func (h *handler) OnAddOrUpdate(in *v1alpha1.ClusterServiceVersion) {
+	name, matched := h.isMatchingCSV(in)
+	if !matched {
+		return
+	}
+
+	h.logger.Debugf("OnAddOrUpdate - found a matching CSV name=%s phase=%s, sending notification", in.GetName(), in.Status.Phase)
+
+	final := h.getLatestInReplacementChain(in)
```
ecordell: is there utility in tracking whether we're "final" or not?

can we just say "if there's a replacement, progressing=true, else progressing=false"?

pkg/lib/operatorstatus/monitor.go
```diff
@@ -0,0 +1,221 @@
...
+	return monitor, monitor
+}
+
+// NotificationContext contains all necessary information related to a notification.
+type NotificationContext struct {
+	// Name of the clusteroperator resource associated with this CSV.
+	Name string
+
+	// Current is the CSV for which we have received notification.
+	Current *v1alpha1.ClusterServiceVersion
```
ecordell: I think this is probably exposing a bit too much detail from the lower level - we really just need enough information to determine if we're progressing, and to determine which ClusterOperator status to update. (i.e. think about how this will change when we change v1alpha1.ClusterServiceVersion to v1.Operator)

pkg/lib/operatorstatus/monitor.go
```diff
@@ -0,0 +1,221 @@
...
+	// Default size of the notification channel.
+	defaultNotificationChannelSize = 64
+)
+
+// NewMonitor returns a new instance of Monitor that can be used to continuously
+// update a clusteroperator resource and an instance of Sender that can be used
+// to send update notifications to it.
+//
+// The name of the clusteroperator resource to monitor is specified in name.
+func NewMonitor(name string, log *logrus.Logger, opClient operatorclient.ClientInterface, configClient configv1client.ConfigV1Interface) (Monitor, Sender) {
```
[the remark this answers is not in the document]
tkashem: It does not, fixed it.

pkg/lib/operatorstatus/reporter.go
```diff
@@ -0,0 +1,120 @@
...
+func newReporter(releaseVersion string) *reporter {
+	return &reporter{
+		clock:          &clock.RealClock{},
+		releaseVersion: releaseVersion,
+	}
+}
+
+// reporter provides the logic for initialzing ClusterOperator and
+// ClusterOperatorStatus types.
+type reporter struct {
```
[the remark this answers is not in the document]
tkashem: renamed it to `csvStatusReporter`

pkg/lib/operatorstatus/handler.go
tkashem: If a `csv` is being replaced then I set message for `Progressing` to
`Working toward {version of the final csv}"`

Thoughts?

pkg/lib/operatorstatus/monitor.go
tkashem: `main.go` can introduce a parent context and that parent context can be passed into this `Run` method. I think one of your PR(s) introduces a parent context in `main.go`. Once that goes in, I can then use it.

pkg/lib/operatorstatus/monitor.go
tkashem: Let me see if I can make it a little more abstract.

pkg/controller/operators/olm/operator.go
```diff
@@ -614,6 +627,10 @@ func (a *Operator) syncClusterServiceVersion(obj interface{}) (syncError error)
 	})
 	logger.Debug("syncing CSV")
 
+	if a.csvNotification != nil {
+		a.csvNotification.OnAddOrUpdate(clusterServiceVersion)
```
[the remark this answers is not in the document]
tkashem: I can address this in a separate PR once @njhale's PR merges, this way we can avoid code churning. Does that sound good?

pkg/lib/operatorstatus/monitor.go
tkashem: I decoupled `monitor` from csv specifics, so the monitor now accepts the following instead of a csv specific structure -
```
// MutatorFunc accepts an existing status object and appropriately mutates it
// to reflect the observed states.
type MutatorFunc func(existing *configv1.ClusterOperatorStatus) (new *configv1.ClusterOperatorStatus)

// Mutate is a wrapper for MutatorFunc
func (m MutatorFunc) Mutate(existing *configv1.ClusterOperatorStatus) (new *configv1.ClusterOperatorStatus) {
	return m(existing)
}

// NotificationFunc wraps a notification event. it returns the name of the
// cluster operator object associated and a mutator function that will set the
// new status for the cluster operator object.
type NotificationFunc func() (name string, mutator MutatorFunc)

// Get is a wrapper for NotificationFunc.
func (n NotificationFunc) Get() (name string, mutator MutatorFunc) {
	return n()
}
```

This way we can extend this to `subscription` or any other resources. @ecordell let me know your thoughts.

ecordell:
Looks like you're hitting a new requirement:

```
[Feature:Platform] ClusterOperators should define at least one namespace in their lists of related objects [Suite:openshift/conformance/parallel] expand_less
```

ecordell:
/retest
/lgtm

(there were a couple of notes to follow up with after #881, so let's make sure to track that)

tkashem:
@ecordell added a jira ticket to address pending feedback - https://jira.coreos.com/browse/OLM-1119

njhale:
Looks like we're missing some stuff on the `OperatorStatus` resource for packageserver:
> ClusterOperator: operator-lifecycle-manager-packageserver

tkashem:
> Looks like we're missing some fields on CVO resources for packageserver:
> 
> > ClusterOperators should define at least one namespace in their lists of related objects
> > ClusterOperators should define at least one related object that is not a namespace

I thought I would address adding `RelatedObjects` in a separate PR. I didn't realize that the e2e test enforces this. I will make the changes.

PR #864, opened 2019-05-21
chore(deps): bump kube deps to 1.14

ecordell:
/hold

this bumps kube dependencies to 1.14; I was hoping we could get more of our PRs cleared out before doing this so holding for now.


estroz:
@ecordell you'll have to use [`Client.CoreV1(...)`](https://github.com/kubernetes/client-go/blob/ae8359b20417914b73a4b514b7a3d642597700bb/kubernetes/clientset.go#L83) instead of `Client.Core(...)` in [this line](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/lib/operatorclient/serviceaccount.go#L38).

ecordell:
/retest

hit the namespace cleanup issue in the olm tests

PR #868, opened 2019-05-22
docs(proposal): add operator group sa doc

jpeeler:
This is a proposal for having a service account associated with an operator group. The associated service account will serve as a threshold for permissions granted to an operator.

### DIFF
--- a/Documentation/contributors/design-proposals/operator-group-sa.md
+++ b/Documentation/contributors/design-proposals/operator-group-sa.md
@@ -1,0 +1,45 @@
+# Add service accounts for Operator Groups
+
+Status: Pending
+
+Version: v1 (Operator groups are already v1)
+
+Implementation owner: Abu
+
+## Abstract
+
+Proposal to add a new field for storing a service account to use with Operator Groups in order to ensure permissions for installed operators do not exceed given service account.
+
+## Motivation
+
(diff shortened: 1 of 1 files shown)


Documentation/contributors/design-proposals/operator-group-sa.md
```diff
@@ -0,0 +1,45 @@
...
+
+### Method of authorization
+
+These changes are currently written with the intention of utilizing the RBAC authorizer and the surrounding code in OLM.
+
+### Implementation
+
+The Operator Group spec already has a placeholder for the service account to be specified, so no changes are required to the operator group type itself.
+
+Modify NewServiceAccountStepResources, RBACForClusterServiceVersion to include a service account to be used instead of any specified in the CSV. For this proposal, the operator group will be looked up (model from operatorGroupForCSV) to check if a service account has been defined. If a service account is associated with an operator group, permission creation will be disabled in RBACForClusterServiceVersion. Since an operator may have already been deployed, existing permissions that are owned by the CSV should be cleaned up in this scenario as well.
```
ecordell: > If a service account is associated with an operator group, permission creation will be disabled in RBACForClusterServiceVersion. Since an operator may have already been deployed, existing permissions that are owned by the CSV should be cleaned up in this scenario as well.

We still want to create the permissions for the operator, we just want to ensure that the permission granted to the serviceaccount does not exceed the permission granted to the operatorgroup serviceaccount.

jpeeler: I've revised the proposal wording, commit message, github summary, and implementation to reflect this.

jpeeler:
Superceded by https://github.com/operator-framework/operator-lifecycle-manager/pull/880. (I would never have come up with the proposed rest.Config BearerToken handling.)

PR #869, opened 2019-05-23
fix(operatorgroup): Remove namespaces requirement for opgroup status

dinhxuanvu:

### DIFF
--- a/manifests/0000_50_olm_10-operatorgroup.crd.yaml
+++ b/manifests/0000_50_olm_10-operatorgroup.crd.yaml
@@ -91,7 +91,6 @@ spec:
                 type: string
               type: array
           required:
-          - namespaces
           - lastUpdated
           type: object
       required:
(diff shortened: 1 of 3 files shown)


deploy/ocp/manifests/0.10.0/0000_50_olm_10-operatorgroup.crd.yaml
```diff
@@ -93,7 +93,6 @@ spec:
                 type: string
               type: array
           required:
-          - namespaces
```
jpeeler: I don't think you want to modify the versioned directories as they have already been released.

manifests/0000_50_olm_10-operatorgroup.crd.yaml
```diff
@@ -91,7 +91,6 @@ spec:
                 type: string
               type: array
           required:
-          - namespaces
```
jpeeler: This change is probably ok, but should be automatically updated next release.

dinhxuanvu: I ran `make release` given in the past when I changed something in template I was told to run `make release`...

jpeeler: Right, so just get rid of the other changes. Running make release is an easy way to update this directory, but also updates the other directories which we don't want.

PR #870, opened 2019-05-23
feat(catsrc): add podselector to catalogsource

ecordell:
if using address, this will ensure that the pod providing the grpc
connection is in an approved namespace (same as the catalogsource or
the global namespace)

if not using the address field, this will create a service for the pods
matched by the pod selector.


ecordell:
After some further discussion, we will rely on other mechanisms in kube for this (i.e. networkpolicy)

PR #875, opened 2019-05-29
feat(install): Add new install scripts for releases, update release

ecordell:
instructions.

Also removes okd manifests, which are no longer relevant.


deploy/upstream/quickstart/install.sh
```diff
@@ -0,0 +1,41 @@
...
+set -e
+
+if [[ ${#@} -ne 1 ]]; then
+    echo "Usage: $0 version"
+    echo "* version: the github release version"
+    exit 1
+fi
+
+release=$1
+url=https://github.com/operator-framework/operator-lifecycle-manager/releases/download/${release}
```
njhale: 👍

deploy/upstream/quickstart/install.sh
```diff
@@ -0,0 +1,41 @@
...
+namespace=olm
+
+kubectl apply -f ${url}/crds.yaml
```
njhale: Will this block until the CRDs are listed in discovery?

PR #878, opened 2019-05-31
Enable OLM to update CRD when there is only one owner of that CRD

dinhxuanvu:

### DIFF
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -102,17 +102,18 @@ func NewOperator(ctx context.Context, kubeconfigPath string, clock utilclock.Clo
 
 	// Allocate the new instance of an Operator.
 	op := &Operator{
-		Operator:       queueOperator,
-		logger:         logger,
-		clock:          clock,
-		opClient:       opClient,
-		client:         crClient,
-		lister:         lister,
-		namespace:      operatorNamespace,
-		sources:        make(map[resolver.CatalogKey]resolver.SourceRef),
-		resolver:       resolver.NewOperatorsV1alpha1Resolver(lister),
-		catsrcQueueSet: queueinformer.NewEmptyResourceQueueSet(),
-		subQueueSet:    queueinformer.NewEmptyResourceQueueSet(),
+		Operator:               queueOperator,
+		logger:                 logger,
+		clock:                  clock,
+		opClient:               opClient,
+		client:                 crClient,
+		lister:                 lister,
+		namespace:              operatorNamespace,
+		sources:                make(map[resolver.CatalogKey]resolver.SourceRef),
+		resolver:               resolver.NewOperatorsV1alpha1Resolver(lister),
+		catsrcQueueSet:         queueinformer.NewEmptyResourceQueueSet(),
+		subQueueSet:            queueinformer.NewEmptyResourceQueueSet(),
+		csvProvidedAPIsIndexer: map[string]cache.Indexer{},
 	}
 	op.reconciler = reconciler.NewRegistryReconcilerFactory(lister, opClient, configmapRegistryImage, op.now)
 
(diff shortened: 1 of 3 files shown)


pkg/controller/operators/catalog/operator.go
```diff
@@ -1056,6 +1057,19 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
 				// Attempt to create the CRD.
 				_, err = o.OpClient.ApiextensionsV1beta1Interface().ApiextensionsV1beta1().CustomResourceDefinitions().Create(&crd)
 				if k8serrors.IsAlreadyExists(err) {
+					currentCRD, _ := o.OpClient.ApiextensionsV1beta1Interface().ApiextensionsV1beta1().CustomResourceDefinitions().Get(crd.GetName(), metav1.GetOptions{})
+					// Compare 2 CRDs to see if it needs to be updatetd
+					if !reflect.DeepEqual(crd, *currentCRD) {
+						// Verify CRD ownership, only attempt to update if
+						// CRD has only one owner
+						if len(existingCRDOwners[currentCRD.GetName()]) == 1 {
```
ecordell: Unfortunately, existingCRDOwners only tells you about owners in the same namespace:
https://github.com/operator-framework/operator-lifecycle-manager/blob/07dd9447712d5d3d21975f3275c9fa51185a29de/pkg/controller/operators/catalog/operator.go#L1399-L1420

For upgrade safety for CRDs, we need to be sure that there's no owner in _all_ namespaces.

The simplest thing to do would be to have a similar function but which collects from all namespace.

But, the nicest way to do this would be to add an extra IndexFunc to our CSV cache to key off of provided APIs. If we do that, then this becomes a very simple query into the cache. We're already storing all CSVs in the cache so adding the extra index should be very low overhead.

jpeeler: @ecordell Is this what you had in mind? I'm not completely sure if the indexes are correct.

```
	sharedInformerFactory := externalversions.NewSharedInformerFactory(crClient, wakeupInterval)
	csvInformer := sharedInformerFactory.Operators().V1alpha1().ClusterServiceVersions()
	op.lister.OperatorsV1alpha1().RegisterClusterServiceVersionLister(metav1.NamespaceAll, csvInformer.Lister())
	csvInformer.Informer().AddIndexers(cache.Indexers{"providedAPIs": func(obj interface{}) ([]string, error) {
		indicies := []string{}

		csv, ok := obj.(*v1alpha1.ClusterServiceVersion)
		if !ok {
			return indicies, fmt.Errorf("invalid object of type: %T", obj)
		}

		for _, crd := range csv.Spec.CustomResourceDefinitions.Owned {
			indicies = append(indicies, fmt.Sprintf("%s=%s", crd.Name, csv.GetName()))
		}
		for _, api := range csv.Spec.APIServiceDefinitions.Owned {
			indicies = append(indicies, fmt.Sprintf("%s=%s", api.Group, csv.GetName()))
		}

		return indicies, nil
	}})
	op.csvProvidedAPIIndexer = csvInformer.Informer().GetIndexer()
```

jpeeler: I just realized that the code here is in the catalog operator. Given that, I think it's probably best to just extend getExistingApiOwners as necessary (or create something new).

ecordell: @jpeeler Yep, something like that! I think the current way the informers are set up you may need one per namespace, IIRC (that may change with @njhale 's prs)

The index itself should be something we can query:

```go
fmt.Sprintf("provided=%s/%s/%k", csvGroup, csvVersion, csvKind);
```
That way you can get the list of CSVs that provide an api with `provided=database.coreos.com/v1alpha1/EtcdCluster`, etc

pkg/controller/operators/catalog/operator.go
```diff
@@ -1056,6 +1067,25 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
 				// Attempt to create the CRD.
 				_, err = o.OpClient.ApiextensionsV1beta1Interface().ApiextensionsV1beta1().CustomResourceDefinitions().Create(&crd)
 				if k8serrors.IsAlreadyExists(err) {
+					currentCRD, _ := o.OpClient.ApiextensionsV1beta1Interface().ApiextensionsV1beta1().CustomResourceDefinitions().Get(crd.GetName(), metav1.GetOptions{})
+					// Compare 2 CRDs to see if it needs to be updatetd
+					if !reflect.DeepEqual(crd, *currentCRD) {
+						// Verify CRD ownership, only attempt to update if
+						// CRD has only one owner
+						// Example: provided=database.coreos.com/v1alpha1/EtcdCluster
+						apiKey := fmt.Sprintf("provided=%s/%s", crd.APIVersion, crd.Kind)
```
jpeeler: This is not the correct format for the key (look at what was chosen in api.go). The example text should look like:
provided=packages.operators.coreos.com/v1/PackageManifest provided=cluster.com/v1alpha1/ins-mcvbv...
At first I thought it might just be easier to change the key format to use the CRD name without splitting it, but maybe for consistency with the api services it should stay as is.

pkg/controller/operators/catalog/operator.go
```diff
@@ -1056,6 +1067,25 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
...
+						matchedCSV, err := index.APIsIndexValues(o.csvProvidedAPIsIndexer, apiKey)
+						if err != nil {
+							return errorwrap.Wrapf(err, "error find matched CSV: %s", step.Resource.Name)
+						}
+						if len(matchedCSV) == 1 {
+							// Attempt to update CRD
+							_, err = o.OpClient.ApiextensionsV1beta1Interface().ApiextensionsV1beta1().CustomResourceDefinitions().Update(&crd)
```
jpeeler: This update will fail because the resource version is empty. You can just copy it from the retrieved version like:
`crd.SetResourceVersion(currentCRD.GetResourceVersion())`.

dinhxuanvu:
Ready for final review and approval for merging :)

test/e2e/installplan_e2e_test.go
```diff
@@ -641,6 +642,176 @@ func TestCreateInstallPlanWithPreExistingCRDOwners(t *testing.T) {
 	})
 }
 
+func TestUpdateInstallPlan(t *testing.T) {
+	defer cleaner.NotifyTestComplete(t, true)
+	t.Run("UpdateSingleExistingCRDOwner", func(t *testing.T) {
```
njhale: Nice! I think we also want a negative test-case; i.e. CRD upgrade is blocked from a pre-existing owner.

test/e2e/installplan_e2e_test.go
```diff
@@ -641,6 +642,335 @@ func TestCreateInstallPlanWithPreExistingCRDOwners(t *testing.T) {
...
+
+		// Wait for csv to update
+		_, err = awaitCSV(t, crc, testNamespace, updatedCSV.GetName(), csvAnyChecker)
+		require.NoError(t, err)
+
+		// Get the CRD to see if it is updated
+		fetchedCRD, err := c.ApiextensionsV1beta1Interface().ApiextensionsV1beta1().CustomResourceDefinitions().Get(crdName, metav1.GetOptions{})
+		require.NoError(t, err)
+
+		for _, version := range fetchedCRD.Spec.Versions {
```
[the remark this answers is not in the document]
njhale: `require.EqualElements` ignores ordering if that's needed.

PR #880, opened 2019-05-31
(proposal) Add user defined service account

tkashem:
Put forth a proposal for the following story:
As a cluster administrator, I should be able to require that all
operator installs and upgrades be and run under a ServiceAccount,
so that I can ensure no user can install an operator with
greater permissions than their own.


tkashem:
> The only thing I'm thinking is that we _may_ want to go ahead and use the serviceaccount for deploying the operator as well (deployment, etc). I don't see any reason not to, and it may help with other features in the future. But otherwise this looks good!

Sounds good to me!

tkashem:
> /lgtm
> 
> The only thing I'm thinking is that we _may_ want to go ahead and use the serviceaccount for deploying the operator as well (deployment, etc). I don't see any reason not to, and it may help with other features in the future. But otherwise this looks good!

I added this to the scope. @ecordell

PR #881, opened 2019-06-03
Subscription CatalogSource Status

njhale:
This PR adds Subscription CatalogSource status reconciliation by:
- Adding a Subscription-scoped operator
- Introducing typestates for subscriptions
- Introducing a typestate Reconciler abstraction

I've aimed for this to be a  _best of_ compilation of ideas/refactoring from #825 and #872, while still leaving the door open for future iterations (such as [this]( https://gist.github.com/njhale/5e20a68102cd1e4b0f142a165ae19113)).


pkg/api/apis/operators/subscription_types.go
```diff
@@ -35,6 +37,56 @@ type SubscriptionSpec struct {
...
+	AllCatalogSourcesHealthy string = "AllCatalogSourcesHealthy"
+
+	// CatalogSourceUpdated is a reason string for Subscriptions that transitioned due to a CatalogSource being updated.
+	CatalogSourceUpdated string = "CatalogSourceUpdated"
+
+	// CatalogSourceRemoved is a reason string for Subscriptions that transitioned due to a CatalogSource being removed.
+	CatalogSourceRemoved string = "CatalogSourceRemoved"
+
+	// UnhealthyCatalogSourceFound is a reason string for Subscriptions that transitioned because an unhealthy catalog source was found.
+	UnhealthyCatalogSourceFound string = "UnhealthyCatalogSourceFound"
```
ecordell: I haven't made it this far yet, but ideally we use this in conjunction with a message that indicates whether that matters for the subscription (i.e. it's explicitly referenced in the Subscription spec)

pkg/controller/operators/catalog/subscription/operator.go
```diff
@@ -0,0 +1,165 @@
...
+type Operator interface {
+	Sync(ctx context.Context, eventType ResourceEventType, obj interface{}) error
+}
+
+type subscriptionOperator struct {
+	logger            *logrus.Logger
+	clock             clock.Clock
+	chain             ReconcilerChain
+	subscriptionCache cache.Indexer
+	// reconcilerFactory is used to check the health of catalogs.
```
ecordell: this comment describes a field that's not there?

pkg/controller/operators/catalog/subscription/operator.go
```diff
@@ -0,0 +1,165 @@
...
+		"event":       eventType,
+	})
+
+	// Enter initial state based on subscription and event type
+	initial := NewSubscriptionState(ext.DeepCopy())
+	switch eventType {
+	case ResourceAdded:
+		initial = initial.Add()
+	case ResourceUpdated:
+		initial = initial.Update()
```
ecordell: we shouldn't be relying on this state, right?

pkg/controller/operators/catalog/subscription/operator.go
```diff
@@ -0,0 +1,165 @@
...
+			now:                       o.now,
+			subscriptionClient:        config.subscriptionClient,
+			catalogLister:             listers.NewCatalogSourceLister(config.catalogInformer.GetIndexer()),
+			registryReconcilerFactory: config.registryReconcilerFactory,
+			globalCatalogNamespace:    config.globalCatalogNamespace,
+		},
+		// TODO: Register other reconcilers
+		// TOOD: Register delete reconciler(s)
+	}
+	o.chain = append(defaultChain, o.chain)
```
ecordell: `o.chain...`?

pkg/controller/operators/catalog/subscription/reconciler.go
```diff
@@ -0,0 +1,163 @@
...
+
+func (c *catalogHealthReconciler) healthy(catalog *operators.CatalogSource) (bool, error) {
+	// TODO: remove conversion when reconciler takes internal type
+	var ext *v1alpha1.CatalogSource
+	err := scheme.Convert(catalog, ext, 0)
+	if err != nil {
+		return false, err
+	}
+
+	return c.registryReconcilerFactory.ReconcilerForSource(ext).CheckRegistryServer(ext)
```
ecordell: non-blocking: we should probably memoize `ReconcilerForSource` so that we only have to construct each reconciler once

pkg/controller/operators/catalog/subscription/operator.go
```diff
@@ -0,0 +1,165 @@
...
+	config.subscriptionInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		DeleteFunc: func(obj interface{}) {
+			o.Sync(ctx, ResourceDeleted, obj)
+		},
+	})
+
+	// Add dependency notifications
+	config.catalogInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			// TODO: extract catalog key and notify dependents with an update event
```
ecordell: this is important for this PR, correct? (just checking if I missed something)

pkg/controller/operators/catalog/subscription/reconciler.go
```diff
@@ -0,0 +1,163 @@
...
+type Reconciler interface {
+	Reconcile(ctx context.Context, in State) (out State, err error)
+}
+
+type ReconcilerChain []Reconciler
+
+func (r ReconcilerChain) Reconcile(ctx context.Context, in State) (out State, err error) {
+	out = in
+	for _, rec := range r {
+		if out, err = rec.Reconcile(ctx, out); out.Terminal() || err != nil {
```
ecordell: 👍

pkg/controller/operators/catalog/subscription/reconciler.go
```diff
@@ -0,0 +1,163 @@
...
+	// When CatalogSource status conditions become robust enough, we should swap this
+	// for typestate based check.
+	registryReconcilerFactory reconciler.RegistryReconcilerFactory
+	globalCatalogNamespace    string
+}
+
+func (c *catalogHealthReconciler) Reconcile(ctx context.Context, in State) (out State, err error) {
+	next := in
+	// loop until this state can no longer transition.
+	for !next.Terminal() && err == nil && out == nil {
```
ecordell: how can we reason that this will terminate?

- context has a timeout?
- next can't return the same state without returning out or err?

pkg/controller/operators/olm/operator.go
```diff
@@ -47,314 +49,354 @@ var (
...
+	return newOperatorWithConfig(ctx, config)
+}
+
+func newOperatorWithConfig(ctx context.Context, config *operatorConfig) (*Operator, error) {
+	if err := config.validate(); err != nil {
+		return nil, err
 	}
 
-	queueOperator, err := queueinformer.NewOperatorFromClient(opClient, logger)
+	// // Create a new client for OLM types (CRs)
```
ecordell: some old code here

pkg/lib/queueinformer/queueinformer_operator.go
```diff
@@ -1,273 +1,300 @@
...
 	}
-	if !exists {
-		// For now, we ignore the case where an object used to exist but no longer does
-		logger.Info("couldn't get from queue")
-		logger.Debugf("have keys: %v", loop.informer.GetIndexer().ListKeys())
-		return nil
+
+	o.logger.Info("starting workers...")
+	for _, queueInformer := range o.queueInformers {
+		go o.worker(ctx, queueInformer)
```
ecordell: the number of workers should probably be configurable?

pkg/lib/queueinformer/queueinformer_operator.go
```diff
@@ -1,273 +1,300 @@
...
+		utilruntime.HandleError(errors.Wrap(err, fmt.Sprintf("sync %q failed", item)))
+		queue.AddRateLimited(item)
 		return true
 	}
-	queue.Forget(key)
-	if err := loop.HandleMetrics(); err != nil {
-		o.Log.Error(err)
+	queue.Forget(item)
+
+	select {
```
ecordell: What is this select for?

pkg/controller/operators/catalog/subscription/reconciler.go
```diff
@@ -0,0 +1,180 @@
...
+	globalCatalogNamespace    string
+}
+
+// Reconcile reconciles subscription catalog health conditions.
+func (c *catalogHealthReconciler) Reconcile(ctx context.Context, in kubestate.State) (out kubestate.State, err error) {
+	next := in
+	var prev kubestate.State
+
+	// loop until this state can no longer transition
+	for err == nil && out == nil && next != nil && !next.Terminal() && prev != next {
```
ecordell: shouldn't we be able to transition from one state to the same state?

njhale: Yes, we can. The idea here is to use that as a condition for ending this reconciliation cycle.

pkg/lib/queueinformer/queueinformer_operator.go
njhale: This was in the original `QueueInformer` code. It looks like it makes the channel send non-blocking, so the sync can exit without a receiver on the other side (or it's a full buffered channel).

pkg/lib/queueinformer/queueinformer_operator.go
njhale: Yup, would a follow up PR be okay?

pkg/controller/operators/catalog/operator.go
```diff
@@ -942,6 +915,18 @@ func (o *Operator) syncInstallPlans(obj interface{}) (syncError error) {
 
 	logger.Info("syncing")
 
+	defer func() {
```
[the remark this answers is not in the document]
njhale: This should probably be moved down. I moved this above the plan length check because I was seeing issues around subscriptions getting wedged.

PR #883, opened 2019-06-03
Add Visibility Label for Included CatalogSources

alecmerdler:
### Description

Used as a solution to filtering out `CatalogSources` created by Marketplace Operator for the OperatorHub UI.

### DIFF
--- a/deploy/chart/templates/0000_50_olm_12-olm-operators.catalogsource.yaml
+++ b/deploy/chart/templates/0000_50_olm_12-olm-operators.catalogsource.yaml
@@ -5,6 +5,8 @@ kind: CatalogSource
 metadata:
   name: olm-operators
   namespace: {{ .Values.catalog_namespace }}
+  labels: 
+    olm-visibility: hidden
 spec:
   sourceType: internal
   configMap: olm-operators
--- a/deploy/chart/templates/0000_50_olm_18-upstream-operators.catalogsource.yaml
+++ b/deploy/chart/templates/0000_50_olm_18-upstream-operators.catalogsource.yaml
@@ -4,6 +4,8 @@ kind: CatalogSource
 metadata:
   name: operatorhubio-catalog
   namespace: {{ .Values.catalog_namespace }}
+  labels: 
+    olm-visibility: visible
 spec:
   sourceType: grpc
   image: quay.io/operator-framework/upstream-community-operators:latest


ecordell:
Why is this needed? If the global namespace is `openshift-marketplace`, then olm-operators will only be visible when you directly to openshift-operator-lifecycle-manager.

alecmerdler:
@ecordell The only label selector which will filter "installed" `CatalogSources` (from Marketplace Operator), but keep the "app registry" ones is `olm-visibility: Exists`. Which means any `CatalogSources` we want to be visible in OperatorHub need to have this label.

alecmerdler:
@njhale Does this look good to you?

alecmerdler:
Is there a `labelSelector` that will satisfy the filter requirements for the following `CatalogSources`?

**"Installed" `CatalogSource` (hide):**
```yaml
kind: CatalogSource
metadata:
  labels:
    csc-owner-name: installed-community-olm-dev-old
    csc-owner-namespace: openshift-marketplace
  name: installed-community-olm-dev-old
  uid: 8d06d8b9-87be-11e9-aeff-021ae8451416
spec:
 ...
status:
  ...
```
**"OCP Included" `CatalogSource` (show):**
```yaml
kind: CatalogSource
metadata:
  labels:
    csc-owner-name: certified-operators
    csc-owner-namespace: openshift-marketplace
    olm-visibility: hidden
    openshift-marketplace: 'true'
    opsrc-datastore: 'true'
    opsrc-owner-name: certified-operators
    opsrc-owner-namespace: openshift-marketplace
    opsrc-provider: certified
  name: certified-operators
spec:
  ...
status:
  ...
```
**"OperatorHub.io" `CatalogSource` (show):**
```yaml
kind: CatalogSource
metadata:
  name: operatorhubio-catalog
spec:
  ...
```

ecordell:
What about:

hide if:
```
csc.owner.name && !openshift-marketplace
```
?

alecmerdler:
Label selectors are for _matching_ expressions to include, not exclude. The negation of that would be:
```
csc-owner-name: DoesNotExist || openshift-marketplace: Exists
```
Like I said, you can't do OR statements :cry:

aravindhp:
If only [set based requirements](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#resources-that-support-set-based-requirements) worked for CRDs...

ecordell:
How difficult would it be to make two api calls? `openshift-marketplace=true` and `!csc-owner-name` should find everything we want to display?

galletti94:
In the installation of an operator you could set the `openshift-marketplace` label to `false`. Then your label selector would just be based on that label's value? From what I remember, this was the original intent of that label.

galletti94:
Nvm I just saw your comment about exclusion.

alecmerdler:
Ended up [using two queries for `PackageManifest`](https://github.com/openshift/console/pull/1627/files#diff-678de420ac7711f131c2d103e91e969fR101): one for Marketplace items and another for local `CatalogSources`.

PR #885, opened 2019-06-04
repair run_console_run.sh script

madorn:
Script wasn't able to run due to getting stuck on exit status check on running container in foreground. Added a few extra functions to @camilamacedo86's original additions/features.

### DIFF
--- a/scripts/run_console_local.sh
+++ b/scripts/run_console_local.sh
@@ -3,7 +3,16 @@
 # Colors definition
 readonly RED=$(tput setaf 1)
 readonly RESET=$(tput sgr0)
-readonly BLUE=$(tput setaf 2)
+readonly GREEN=$(tput setaf 2)
+
+# Check if Podman binary exists
+verify_podman_binary() {
+    if hash podman 2>/dev/null; then
+        POD_MANAGER="podman"
+    else
+        POD_MANAGER="docker"
+    fi
+}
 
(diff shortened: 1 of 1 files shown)


scripts/run_console_local.sh
```diff
@@ -14,38 +23,49 @@ add_host_port_arg (){
...
       -e BRIDGE_K8S_AUTH="bearer-token" \
       -e BRIDGE_K8S_AUTH_BEARER_TOKEN=$(kubectl get secret "$secretname" --namespace=kube-system -o template --template='{{.data.token}}' | base64 --decode) \
       quay.io/openshift/origin-console:latest &> /dev/null
+}
 
-    docker_exists=${?}; if [[ ${docker_exists} -ne 0 ]]; then
-        echo -e "${BLUE}The OLM is accessible via web console at:${RESET}"
-        echo -e "${BLUE}https://localhost:9000/${RESET}"
+verify_ocp_console_image (){
+  while true; do
```
ecordell: What's the purpose of the loop? when I tried this on mac it succeeded at first, then slept 10s and hit the else block and exited.

madorn: maybe I'm missing something but it shouldn't be hitting the else - also tested on mac and linux:
https://asciinema.org/a/WLQ5sQToBod41aQTeTEWPnh13

loops message so that session is blocked with cleanup on quit vs. script runs, exits, and user forgets to cleanup manually.

PR #871, opened 2019-05-23
Change Global Catalog Namespace for OpenShift

alecmerdler:
### Description

Changes the `-namespace` and `-global-namespace` flags for `catalog-operator` and `packageserver` respectively. This allows creating `Subscriptions` directly to `CatalogSources` in the `openshift-marketplace` namespace without the indirection of creating a `CatalogSourceConfig`.

Addresses https://jira.coreos.com/browse/OLM-1080
Unblocks https://jira.coreos.com/browse/MKTPLC-367

PR #889, opened 2019-06-06
make release 0.10.1

dinhxuanvu:


manifests/0000_50_olm_12-olm-operators.catalogsource.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: operators.coreos.com/v1alpha1
 kind: CatalogSource
 metadata:
   name: olm-operators
-  namespace: openshift-operator-lifecycle-manager
+  namespace: openshift-marketplace
```
ecordell: I don't think this should've changed to openshift-marketplace, I think it should stay in the olm namespace.

This looks like it's causing the e2e failures as well.

alecmerdler: That's probably from #871. We might need another variable for that.

dinhxuanvu: I will revert that change my PR.

dinhxuanvu: On the second thought, there are changes like this across other files so I need to know if we are gonna switch all of them back to olm namespace or just this one specifically.

ecordell: it should just be this one that needs to change, from what I can recall

PR #893, opened 2019-06-07
Handle Non-Existent Namespace Queries for PackageManifest

alecmerdler:
### Description

Return empty list for `List` requests and `404 Not Found` for `Get` requests.

Addresses https://jira.coreos.com/browse/OLM-1103


deploy/chart/templates/_packageserver.clusterserviceversion.yaml
```diff
@@ -48,6 +48,7 @@ spec:
           - ""
           resources:
           - configmaps
+          - namespaces
```
alecmerdler: The `package-server` needs RBAC to list namespaces in order to determine if the requested namespace exists (if it doesn't, return empty list like native Kubernetes resources).

alecmerdler:
@ecordell Ready for another review.

pkg/package-server/provider/registry.go
```diff
@@ -242,6 +246,12 @@ func (p *RegistryProvider) List(namespace string) (*operators.PackageManifestLis
 	defer p.mu.RUnlock()
 
 	pkgs := []operators.PackageManifest{}
+	if namespace != metav1.NamespaceAll {
```
ecordell: Instead of querying the cluster, can't we just loop through `p.clients` and return an empty list if there's no client for that namespace?

alecmerdler: A request for a namespace that doesn't exist would still return `PackageManifests` from the "global" namespace.

njhale:
Going to close this out since the underlying code has changed so much and @alecmerdler has moved onto another team.

I'm going to open a BZ and new fix for this.

PR #900, opened 2019-06-10
bug:1691546 fix(catalog): do not add owner references to clusterroles or crbs

ecordell:
Operators requesting Cluster Permissions could hit a kube GC bug

### DIFF
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -1147,16 +1147,11 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
 					return errorwrap.Wrapf(err, "error parsing step manifest: %s", step.Resource.Name)
 				}
 
-				// Update UIDs on all CSV OwnerReferences
-				updated, err := o.getUpdatedOwnerReferences(cr.OwnerReferences, plan.Namespace)
-				if err != nil {
-					return errorwrap.Wrapf(err, "error generating ownerrefs for clusterrole %s", cr.GetName())
-				}
-				cr.OwnerReferences = updated
-
 				// Attempt to create the ClusterRole.
 				_, err = o.opClient.KubernetesInterface().RbacV1().ClusterRoles().Create(&cr)
 				if k8serrors.IsAlreadyExists(err) {
+					// if we're updating, point owner to the newest csv
+					cr.Labels[ownerutil.OwnerKey] = step.Resolving
 					_, err = o.opClient.UpdateClusterRole(&cr)
 					if err != nil {
 						return errorwrap.Wrapf(err, "error updating clusterrole %s", cr.GetName())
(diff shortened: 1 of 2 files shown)


pkg/controller/operators/catalog/operator.go
```diff
@@ -1168,16 +1168,11 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
...
-				if err != nil {
-					return errorwrap.Wrapf(err, "error generating ownerrefs for clusterrole %s", cr.GetName())
-				}
-				cr.OwnerReferences = updated
-
 				// Attempt to create the ClusterRole.
 				_, err = o.OpClient.KubernetesInterface().RbacV1().ClusterRoles().Create(&cr)
 				if k8serrors.IsAlreadyExists(err) {
+					// if we're updating, point owner to the newest csv
+					cr.Annotations[ownerutil.OwnerKey] = step.Resolving
```
njhale: Isn't this supposed to be a label?

ecordell: It should be a label, yes. 

We're not currently GCing at all, so switching it should be fine as well. We do need to have an api similar to ownerrefs though, and GC these ourselves (or implement GC in kube)

pkg/controller/operators/catalog/operator.go
```diff
@@ -1198,17 +1193,11 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
...
-				updated, err := o.getUpdatedOwnerReferences(rb.OwnerReferences, plan.Namespace)
-				if err != nil {
-					return errorwrap.Wrapf(err, "error generating ownerrefs for clusterrolebinding %s", rb.GetName())
-				}
-				rb.OwnerReferences = updated
-
 				// Attempt to create the ClusterRoleBinding.
 				_, err = o.OpClient.KubernetesInterface().RbacV1().ClusterRoleBindings().Create(&rb)
 				if k8serrors.IsAlreadyExists(err) {
-					rb.SetNamespace(plan.Namespace)
```
jpeeler: Why was this line removed?

ecordell: ClusterRoleBindings don't have namespaces :)

jpeeler: Ah right!

pkg/controller/operators/catalog/operator.go
```diff
@@ -1168,16 +1168,11 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
...
-				if err != nil {
-					return errorwrap.Wrapf(err, "error generating ownerrefs for clusterrole %s", cr.GetName())
-				}
-				cr.OwnerReferences = updated
-
 				// Attempt to create the ClusterRole.
 				_, err = o.OpClient.KubernetesInterface().RbacV1().ClusterRoles().Create(&cr)
 				if k8serrors.IsAlreadyExists(err) {
+					// if we're updating, point owner to the newest csv
+					cr.Labels[ownerutil.OwnerKey] = step.Resolving
```
tkashem: question: how is this label `ownerutil.OwnerKey` used?

ecordell: it's a label that points to the owner for cases where we can't use real ownerrefs

PR #902, opened 2019-06-11
Allow CRD updates with multiple owners

jpeeler:
The purpose of this PR is to allow CRD updates for existing CRDs, which isn't currently done. This enables that functionality in the cases of:
- New CRD is a different non-existing version
- CRD is the same version without an openapi schema change

~This is rebased on top of https://github.com/operator-framework/operator-lifecycle-manager/pull/878, so that should merge first.~

### DIFF
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -1021,6 +1021,44 @@ func (o *Operator) ResolvePlan(plan *v1alpha1.InstallPlan) error {
 	return nil
 }
 
+// TODO: in the future this will be extended to verify more than just whether or not the schema changed
+func checkCRDSchemas(oldCRD *v1beta1ext.CustomResourceDefinition, newCRD *v1beta1ext.CustomResourceDefinition) bool {
+	if reflect.DeepEqual(oldCRD.Spec.Validation, newCRD.Spec.Validation) {
+		return true
+	}
+
+	return false
+}
+
+func safeToUpdate(oldCRD *v1beta1ext.CustomResourceDefinition, newCRD *v1beta1ext.CustomResourceDefinition) error {
+	shouldCheckSchema := len(oldCRD.Spec.Versions) == len(newCRD.Spec.Versions) // no version bump
+
+	// 1) if there's no version change, verify that schemas match 2) ensure all old versions are present
+	for _, oldVersion := range oldCRD.Spec.Versions {
+		var versionPresent bool
+		for _, newVersion := range newCRD.Spec.Versions {
+			if oldVersion.Name == newVersion.Name {
+				if shouldCheckSchema && !checkCRDSchemas(oldCRD, newCRD) {
+					return fmt.Errorf("not allowing CRD (%v) update with multiple owners with schema change on version %v", newCRD.GetName(), oldVersion)
+				}
+				versionPresent = true
+			}
+		}
+		if !versionPresent {
+			return fmt.Errorf("not allowing CRD (%v) update with unincluded version %v", newCRD.GetName(), oldVersion)
+		}
+	}
+
+	// ensure a CRD with multiple versions isn't checked
+	if newCRD.Spec.Version != "" {
+		if !checkCRDSchemas(oldCRD, newCRD) {
+			return fmt.Errorf("not allowing CRD (%v) update with multiple owners with schema change on (single) version %v", newCRD.GetName(), newCRD.Spec.Version)
+		}
+	}
+	return nil
+
+}
+
 // ExecutePlan applies a planned InstallPlan to a namespace.
 func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
 	if plan.Status.Phase != v1alpha1.InstallPlanPhaseInstalling {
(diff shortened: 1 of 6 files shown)


ecordell:
This looks good but waiting on #878 !

jpeeler:
Since I branched off of Vu's PR, I think this commit will merge properly without me rebasing.

test/e2e/installplan_e2e_test.go
```diff
@@ -498,148 +498,177 @@ func TestCreateInstallPlanWithPreExistingCRDOwners(t *testing.T) {
 		// Should have removed every matching step
 		require.Equal(t, 0, len(expectedSteps), "Actual resource steps do not match expected")
 	})
+}
 
-	t.Run("PreExistingCRDOwnerIsReplaced", func(t *testing.T) {
```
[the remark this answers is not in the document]
jpeeler: Wow, will make sure to put that back.

test/e2e/installplan_e2e_test.go
```diff
@@ -498,148 +498,177 @@ func TestCreateInstallPlanWithPreExistingCRDOwners(t *testing.T) {
...
+		expectedPhase v1alpha1.InstallPlanPhase
+	}{
+		{
+			name:          "CRD same version",
+			version:       "v1alpha1",
+			expectedPhase: v1alpha1.InstallPlanPhaseFailed,
+		},
+		{
+			name:          "CRD different version",
+			version:       "v1alpha2",
```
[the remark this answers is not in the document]
jpeeler: I thought that's what I had done. If a new CRD comes along with the same version, but with a different schema the update will be rejected. If a new version comes along and the schema hasn't changed, the update will be allowed.

I didn't check if a version has been removed, will revise.

pkg/controller/operators/catalog/operator.go
```diff
@@ -1067,9 +1077,49 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
...
+							// ensure all old versions are present
+							for _, oldVersion := range currentCRD.Spec.Versions {
+								var versionPresent bool
+								for _, newVersion := range crd.Spec.Versions {
+									if oldVersion.Name == newVersion.Name {
+										versionPresent = true
+									}
+								}
+								if !versionPresent {
+									return fmt.Errorf("not allowing CRD (%v) update with unincluded version %v", crd.GetName(), oldVersion)
```
ecordell: if we're exiting here, why not move this to line 1095?

pkg/controller/operators/catalog/operator.go
```diff
@@ -1067,9 +1077,49 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
...
+								}
+							}
+
+							// no version bump
+							if len(currentCRD.Spec.Versions) == len(crd.Spec.Versions) {
+								for _, oldVersion := range currentCRD.Spec.Versions {
```
ecordell: since we have to do multiple passes over these versions, why not pull them into maps? e.g.  `oldVersions := map[string]json.Schema` (or whatever the schema type is)

pkg/controller/operators/catalog/operator.go
jpeeler: Because the entire set needs to be checked, so returning immediately on the first mismatch wouldn't necessarily be accurate.

pkg/controller/operators/catalog/operator.go
jpeeler: Your above suggestion led me to refactoring and removing this looping. I think for loops are clearer for what's happening, but see what you think.

test/e2e/installplan_e2e_test.go
```diff
@@ -642,6 +642,203 @@ func TestCreateInstallPlanWithPreExistingCRDOwners(t *testing.T) {
...
+
+			// Wait for InstallPlan to be status: Complete or Failed before checking resource presence
+			fetchedInstallPlan, err = fetchInstallPlan(t, crc, installPlanName, buildInstallPlanPhaseCheckFunc(v1alpha1.InstallPlanPhaseComplete, v1alpha1.InstallPlanPhaseFailed))
+			require.NoError(t, err)
+			fmt.Printf("Phase = %v", fetchedInstallPlan.Status.Phase)
+			t.Logf("Install plan %s fetched with status %s", fetchedInstallPlan.GetName(), fetchedInstallPlan.Status.Phase)
+
+			require.Equal(t, tt.expectedPhase, fetchedInstallPlan.Status.Phase)
+
+			// Fetch installplan again to check for unnecessary control loops
```
jpeeler: I believe I need to remove this part.

pkg/controller/operators/catalog/operator.go
```diff
@@ -1067,9 +1077,40 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
 						if err != nil {
 							return errorwrap.Wrapf(err, "error find matched CSV: %s", step.Resource.Name)
 						}
+						crd.SetResourceVersion(currentCRD.GetResourceVersion())
 						if len(matchedCSV) == 1 {
-							// Attempt to update CRD
-							crd.SetResourceVersion(currentCRD.GetResourceVersion())
+							o.logger.Debugf("Found one owner for CRD %v", crd)
```
ecordell: Don't block the PR on this, but I think there's some room for improving the readability here. Even something simple like replacing this block (~1082-1112) with a function `safeToUpdate(oldCrd, newCrd)` that we could test independently of the operator (and expand as we enable more methods).

test/e2e/installplan_e2e_test.go
```diff
@@ -642,6 +642,203 @@ func TestCreateInstallPlanWithPreExistingCRDOwners(t *testing.T) {
...
+				}
+				mainCRD.Spec.Version = ""
+				mainCRD.Spec.Versions[len(tt.multipleVersions)-1].Storage = true
+			} else {
+				mainCRD.Spec.Version = tt.version
+			}
+
+			var min float64 = 2
+			var max float64 = 256
+			if tt.changeSchema == true {
```
ecordell: also non-blocking - this may just be a personal thing, but I find it easier to reason about table-driven tests when there are minimal conditionals in the test. 

e.g. instead of 

```go
test := {
   branch: true
}

t.Run() {
  if branch {
    do(thing)
  }
  do(other thing)
}
```

doing something like:

```go 
func branch(thing) {
  return other thing
}

tests := {
   {
   obj: branch(thing)
  },
  {
   obj: thing
  }
}

t.Run() {
  do(tt.obj)
}
```

test/e2e/installplan_e2e_test.go
ecordell: If it's not flaking you don't have to, but you can remove it if you want.

pkg/controller/operators/catalog/operator.go
ecordell: Followup - I'm a little concerned that there's no unit test for this (just an e2e) - so when you follow up with a unit test, refactoring like this might be an easy way to get that done quickly (because you can test a single, pure (w.r.t. fake clients) function)

test/e2e/installplan_e2e_test.go
jpeeler: No, you're right. I've fixed it now.

pkg/controller/operators/catalog/operator.go
jpeeler: Added unit test and refactored to avoid using the clients and such.

PR #904, opened 2019-06-11
Fix PackageManifest Server Crashes

alecmerdler:
### Description

Didn't realize that `GetLabels()` returns a reference, not a copy.

Addresses https://jira.coreos.com/browse/OLM-1111

### DIFF
--- a/pkg/package-server/provider/registry.go
+++ b/pkg/package-server/provider/registry.go
@@ -13,10 +13,10 @@ import (
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/connectivity"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/labels"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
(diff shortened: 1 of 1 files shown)


pkg/package-server/provider/registry.go
```diff
@@ -289,9 +289,11 @@ func toPackageManifest(pkg *api.Package, client registryClient) (*operators.Pack
 	catsrc := client.source
 	manifest := &operators.PackageManifest{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:              pkg.GetName(),
-			Namespace:         catsrc.GetNamespace(),
-			Labels:            catsrc.GetLabels(),
+			Name:      pkg.GetName(),
+			Namespace: catsrc.GetNamespace(),
+			Labels: labels.CloneAndAddLabel(
```
ecordell: Why clone twice?

alecmerdler: So it can all be done in-place without extra variables. There isn't a `CloneAndAddLabels()` function.

PR #906, opened 2019-06-12
GRPC source manager

ecordell:
This refactors grpc source management into a component that watches for connection changes and notifies a subscriber of changes. 

Currently based off of #881


scripts/run_e2e_bare.sh
```diff
@@ -14,4 +14,4 @@ fi
 
 echo "${test_flags}"
 go test -c -tags=bare -mod=vendor -v -o e2e-bare github.com/operator-framework/operator-lifecycle-manager/test/e2e
-./e2e-bare  -test.v -test.timeout 20m ${test_flags} -kubeconfig=${KUBECONFIG:-minikube.kubeconfig} -namespace=$(cat e2e.namespace) -olmNamespace=operator-lifecycle-manager
+./e2e-bare  -test.v -test.timeout 20m ${test_flags} -kubeconfig=${KUBECONFIG:-minikube.kubeconfig} -namespace=$(cat e2e.namespace) -olmNamespace=operator-lifecycle-manager -dummyImage=hang:10
```
jpeeler: Does a hang.Dockerfile need to be built here so the image exists for the baremetal run?

pkg/controller/registry/grpc/source_test.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+}
+
+func TestConnectionEvents(t *testing.T) {
+	type testcase struct {
+		name            string
+		expectedHistory map[resolver.CatalogKey]map[int]connectivity.State
+	}
+
+	test := func(tt testcase) func(t *testing.T) {
```
njhale: new style 👍

pkg/controller/registry/grpc/source_test.go
```diff
@@ -0,0 +1,119 @@
...
+			defer stop()
+
+			syncer := NewFakeSourceSyncer(len(tt.expectedHistory[resolver.CatalogKey{Name: "test", Namespace: "test"}]))
+			sources := NewSourceStore(logrus.New(), syncer.sync)
+
+			ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+			defer cancel()
+			sources.Start(ctx)
+
+			_, err := sources.Add(resolver.CatalogKey{Name: "test", Namespace: "test"}, "localhost:50051")
```
njhale: It doesn't look like this test is complete. Should this call `Add` for each key in `tt.expectedHistory`?

pkg/lib/queueinformer/queueinformer_operator.go
```diff
@@ -144,6 +144,7 @@ func (o *operator) addHasSynced(hasSynced cache.InformerSynced) {
 	}
 }
 
+// TODO: generalize over func(ctx) so this can start informers + source watcher
```
njhale: Alternatively, you could create a method on the catalog `Operator` that starts the source watcher after/before calling this.

scripts/build_local.sh
```diff
@@ -6,12 +6,13 @@
...
 if [ -z "$NO_MINIKUBE" ]; then
-  ps x | grep -q [m]inikube || minikube start --kubernetes-version="v1.13.0" --extra-config=apiserver.v=4 || { echo 'Cannot start minikube.'; exit 1; }
-  eval $(minikube docker-env) || { echo 'Cannot switch to minikube docker'; exit 1; }
+  pgrep -qf "[m]inikube" || minikube start --kubernetes-version="v1.14.2" --extra-config=apiserver.v=4 || { echo 'Cannot start minikube.'; exit 1; }
+  eval "$(minikube docker-env)" || { echo 'Cannot switch to minikube docker'; exit 1; }
   kubectl config use-context minikube
 fi
 
 docker build -f local.Dockerfile -t quay.io/operator-framework/olm:local -t quay.io/operator-framework/olm-e2e:local ./bin
+docker build -f test/e2e/hang.Dockerfile -t hang:10 ./bin
```
njhale: 👍

pkg/controller/operators/catalog/operator.go
```diff
@@ -301,6 +311,19 @@ func (o *Operator) now() metav1.Time {
...
 }
 
+func (o *Operator) syncSourceState(state grpc.SourceState) {
+	o.sourcesLastUpdate = o.now()
+
+	switch state.State {
+	case connectivity.Ready:
+		o.resolveNamespace(state.Key.Namespace)
+	default:
+		if err := o.catsrcQueueSet.Requeue(state.Key.Namespace, state.Key.Name); err != nil {
```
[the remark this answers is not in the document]
ecordell: This only gets triggered if the source state changes, and the grpc connection manager does back off.

pkg/controller/operators/catalog/operator.go
```diff
@@ -468,124 +484,66 @@ func (o *Operator) syncCatalogSources(obj interface{}) (syncError error) {
...
 
-			if s, ok := o.sources[sourceKey]; ok {
-				if err := s.Client.Close(); err != nil {
-					logger.WithError(err).Debug("error closing client connection")
-				}
-			}
-			delete(o.sources, sourceKey)
-			o.sourcesLastUpdate = out.Status.LastSync
+		logger.Debug("updating catsrc status")
+		if _, err := o.client.OperatorsV1alpha1().CatalogSources(out.GetNamespace()).UpdateStatus(out); err != nil {
```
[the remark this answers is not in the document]
ecordell: Just in case!

pkg/controller/registry/grpc/source.go
```diff
@@ -0,0 +1,203 @@
...
+	}
+}
+
+func (s *SourceStore) Start(ctx context.Context) {
+	s.logger.Debug("starting source manager")
+	go func() {
+		for {
+			select {
+			case <-ctx.Done():
+				s.logger.Debug("closing source manager")
```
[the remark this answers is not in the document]
ecordell: I can add that - wasn't too worried since this has the lifetime of the program

pkg/controller/registry/grpc/source.go
```diff
@@ -0,0 +1,203 @@
...
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		default:
+			timeout := 10 * time.Second
+			if state == connectivity.Ready {
+				timeout = 10 * time.Minute
+			}
+			timer, _ := context.WithTimeout(context.Background(), timeout)
```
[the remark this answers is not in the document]
ecordell: That makes sense, just trying to remember if there was some reason I didn't do that in the first place

pkg/controller/registry/grpc/source.go
```diff
@@ -0,0 +1,203 @@
...
+				s.notify <- SourceState{Key: key, State: newState}
+			}
+		}
+	}
+}
+
+func (s *SourceStore) Remove(key resolver.CatalogKey) error {
+	s.sourcesLock.RLock()
+	source, ok := s.sources[key]
+	s.sourcesLock.RUnlock()
```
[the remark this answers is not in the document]
ecordell: Not sure I follow - there is a separate lock taken out for the delete?

pkg/controller/registry/grpc/source_test.go
ecordell: 🤷‍♂ just trying something new.

One of my frustrations with "normal" table tests is scrolling back and forth between the testcase struct and the actual test code, with a ton of "table entries" in between. this keeps it all together

scripts/run_e2e_bare.sh
ecordell: It's built in `build_local.sh`

pkg/lib/queueinformer/queueinformer_operator.go
ecordell: that's essentially how it's configured now - did you have something else in mind?

pkg/controller/registry/grpc/source.go
```diff
@@ -0,0 +1,203 @@
...
+			if state == connectivity.Ready {
+				timeout = 10 * time.Minute
+			}
+			timer, _ := context.WithTimeout(context.Background(), timeout)
+			if source.Conn.WaitForStateChange(timer, state) {
+				newState := source.Conn.GetState()
+				state = newState
+
+				// update connection state
+				if src := s.Get(key); src != nil {
```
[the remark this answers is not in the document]
ecordell: I can't call `source.cancel` if I can't find `source` - but I can exit this loop here so the goroutine gets cleaned up.

pkg/lib/queueinformer/queueinformer_operator.go
njhale: No - just making 2 calls -> 1 from from the `catalog.Operator` client perspective.

pkg/controller/operators/catalog/operator_test.go
```diff
@@ -704,17 +705,18 @@ func NewFakeOperator(ctx context.Context, namespace string, watchedNamespaces []
 				// 1 qps, 100 bucket size.  This is only for retry speed and its only the overall factor (not per item)
 				&workqueue.BucketRateLimiter{Limiter: rate.NewLimiter(rate.Limit(1), 100)},
 			), "resolver"),
-		sources:               make(map[resolver.CatalogKey]resolver.SourceRef),
-		resolver:              config.resolver,
+		resolver:              &fakes.FakeResolver{},
```
njhale: I think `config.resolver` is the value we want here. It's defaulted to `&fakes.FakeResolver{}` above and allows test authors to specify a specific fake using the `withResolver(...)` option.

pkg/controller/registry/grpc/source.go
```diff
@@ -0,0 +1,214 @@
...
+		syncFn:       sync,
+		logger:       logger,
+		timeout:      timeout,
+		readyTimeout: readyTimeout,
+	}
+}
+
+func (s *SourceStore) Start(ctx context.Context) {
+	s.logger.Debug("starting source manager")
+	go func() {
```
njhale: nit: maybe use `sync.Once` to ensure the source manager goroutine runs as a singleton.

pkg/controller/registry/grpc/source_test.go
```diff
@@ -0,0 +1,152 @@
...
+	stop := func() {
+		s.Stop()
+	}
+
+	return serve, stop
+}
+
+type FakeSourceSyncer struct {
+	// using a map[int] to preserve order
+	History map[resolver.CatalogKey]map[int]connectivity.State
```
njhale: Question: This might be a dumb question, but for your use-case, what does a `map[int]<t>` give you that a `[]<t>` doesn't?

pkg/controller/registry/grpc/source.go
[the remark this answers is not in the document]
njhale: I just thought that you could use the source passed in as an argument - returning works fine too. Thanks!

pkg/controller/operators/catalog/operator_test.go
ecordell: Ah, yep - this flipped when I rebased

pkg/controller/registry/grpc/source.go
ecordell: Seems reasonable - but I don't think we do that for our other long-running goroutines do we?

pkg/controller/registry/grpc/source_test.go
ecordell: Nothing - no idea why I did this 😄

pkg/controller/registry/grpc/source.go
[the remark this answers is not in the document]
jpeeler: Would it not be possible for a source removal to occur after the ok has been checked, but before the delete lock occurs? I see now though that a delete is a nop if the element doesn't exist, so I guess it's ok.

pkg/controller/registry/grpc/source.go
njhale: I added that to the refactored `queueinformer.Operator` code as a safety net against successive calls that start its informers. Not saying that we _need_ to add this here - just wanted to have some feedback about whether the idea was pragmatic.

pkg/controller/registry/grpc/source.go
```diff
@@ -0,0 +1,217 @@
...
+		sources:      make(map[resolver.CatalogKey]SourceConn),
+		notify:       make(chan SourceState),
+		syncFn:       sync,
+		logger:       logger,
+		timeout:      timeout,
+		readyTimeout: readyTimeout,
+	}
+}
+
+func (s *SourceStore) Start(ctx context.Context) {
```
njhale: nit: Extra "empty" goroutines can be created. Not sure if the compiler will optimize them away, but you could always create the goroutine _inside_ the `Do` function and avoid this all together:
  
```suggestion
func (s *SourceStore) Start(ctx context.Context) {
    s.logger.Debug("starting source manager")
    s.Do(func() {
        // Only start one source manager goroutine in the lifetime of s
	go func() {
	    for {
                select {
                case <-ctx.Done():
	           s.logger.Debug("closing source manager")
	           return
	        case e := <-s.notify:
                    s.logger.Debugf("Got source event: %#v", e)
	            s.syncFn(e)
	        }
	    }
	}()
    }
}
```

ecordell: The “extra” goroutines would immediately exit

ecordell:
/hold

I see what might be flakes caused by this PR in e2e

pkg/controller/registry/grpc/source.go
njhale: True - not a big deal at all - but why bother creating them in the first place?

ecordell:
the first two commits are included in #974 

The final commit will be staged as a separate pr.

PR #911, opened 2019-06-18
Add config options to QueueInformer constructors

njhale:
This PR adds config options to QueueInformers and follows up on some unresolved comments left in #881

- Add variadic config options to QueueInformer constructors
- Make number of queue workers configurable (default to 2)
- Ensure InstallPlan sync only triggers resolution *after* a change
- Remove empty `init()`
- Switch log level back to debug
- Change verbose permissions requirement to `Tracef(...)`

### DIFF
--- a/pkg/lib/queueinformer/config.go
+++ b/pkg/lib/queueinformer/config.go
@@ -3,6 +3,7 @@ package queueinformer
 import (
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
+	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/workqueue"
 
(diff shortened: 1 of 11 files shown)


pkg/lib/queueinformer/config.go
```diff
@@ -140,3 +141,76 @@ func WithSyncer(syncer kubestate.Syncer) Option {
...
+		option(c)
+	}
+}
+
+func newInvalidOperatorConfigError(msg string) error {
+	return errors.Errorf("invalid queue informer operator config: %s", msg)
+}
+
+// WithOperatorLogger sets the logger used by an Operator.
+func WithOperatorLogger(logger *logrus.Logger) OperatorOption {
```
ecordell: nbd, but why not `WithLogger`?

PR #912, opened 2019-06-18
fix(build): don't use helm in src image

ecordell:
We lose our manifest validation with this change, but it's currently an issue for the build pipeline.


Makefile
```diff
@@ -172,6 +175,9 @@ verify-manifests:
 	find ./manifests -type f -exec sed -i "1{/---/d}" {} \;
 	git diff --exit-code
 
+
```
jpeeler: extra space?

PR #913, opened 2019-06-18
Add webhook proposal

jpeeler:
This is my current view of how webhooks should be added into OLM. Feedback/discussion encouraged.


Documentation/contributors/design-proposals/webhooks.md
```diff
@@ -0,0 +1,115 @@
...
+            - apiGroups:
+              - "operators.olm.com"
+              apiVersions:
+              - v1alpha1
+              operations:
+              - CREATE
+              resources:
+              - clusterserviceversions
+          clientConfig:
+            caBundle: ## have this automatically created if not specified ##
```
njhale: For APIServices, OLM always generates CA bundles. We could allow users to specify them, but we would also need to provide a way to specify the webhook's serving cert (or the CA's private key to generate it). We can let them reference secrets, but they would need to be created outside of OLM's control, or provided by a CatalogSource (the latter has its own set of issues).

Documentation/contributors/design-proposals/webhooks.md
```diff
@@ -0,0 +1,115 @@
...
+                        fieldPath: metadata.annotations['olm.targetNamespaces']
+                  - name: MY_POD_NAME
+                    valueFrom:
+                      fieldRef:
+                        fieldPath: metadata.name
+                volumes:
+                - name: tls-keys
+                    secretName: service-tls
+      services:
+      - name: mywebhook-service
```
njhale: Can this be inferred from the webhooks section?

Documentation/contributors/design-proposals/webhooks.md
jpeeler: I pushed before I saw these comments. I didn't realize that the caBundle field would be "filled in" from the apiserver if unspecified. This to me seems like the behavior we'd want for an OLM-managed webhook.

Documentation/contributors/design-proposals/webhooks.md
jpeeler: I try to address this as a possibility in the v2 section.

Documentation/contributors/design-proposals/webhooks.md
```diff
@@ -0,0 +1,213 @@
...
+            service:
+              name: mywebhook-service
+              namespace: test-ns
+              path: "/admit"
+
+Note that the CABundle field in the clientConfig is set to use the general CA from the Kubernetes apiserver.
+
+### Description for v2
+
+The following example is a different take which would remove control from the user and have OLM manage the missing bits that aren't specified. The missing fields and objects would be created by OLM on the fly.
```
jpeeler: "on the fly" - I'm not quite sure how. I was originally thinking of this sort of thing happening during resolving, but in out of band discussions adding data to the bundles has been mentioned which I don't fully understand what that'd look like.

ecordell: This would be handled during CSV creation, similar to how we handle apiservices. See: https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/controller/operators/olm/operator.go#L1031-L1036

@njhale also made the point that we're not using CSRs there, but we probably should be

Documentation/contributors/design-proposals/webhooks.md
```diff
@@ -0,0 +1,213 @@
...
+              - clusterserviceversions
+          clientConfig:
+            service:
+              name: mywebhook-service
+              namespace: test-ns
+              path: "/admit"
+
+Note that the CABundle field in the clientConfig is set to use the general CA from the Kubernetes apiserver.
+
+### Description for v2
```
ecordell: I like this better for the case where we have it embedded in the CSV 👍

Documentation/contributors/design-proposals/webhooks.md
```diff
@@ -0,0 +1,213 @@
...
+          - name: csv.operators.olm.com
+          rules:
+            - apiGroups:
+              - "operators.olm.com"
+              apiVersions:
+              - v1alpha1
+              operations:
+              - CREATE
+              resources:
+              - clusterserviceversions
```
ecordell: In the absence of the client config, I think we need to also take a pod selector, so that we can generate the service correctly. right?

Documentation/contributors/design-proposals/webhooks.md
```diff
@@ -0,0 +1,213 @@
...
+      #- name: mywebhook-service
+      #  namespace: test-ns
+      #  spec:
+      #    selector:
+      #      name: mywebhook
+      #    ports:
+      #    - protocol: TCP
+      #      port 443
+      #      targetPort: 443
+      validatingWebhooks:
```
ecordell: Pros/cons of something like this?
```yaml
webhooks:
- name: olm-admission
   type: ValidatingAdmissionWebhook
   spec:
- name: olm-mutate
   type: MutatingAdmissionWebhook
```
?

jpeeler: The only drawback to this approach is if you had several webhooks of the same type, the previous format would allow you to add them as another list item rather than creating a new section in webhooks as your example above. Given though that it's probably rare that multiple webhooks of the same type are present, I'm going to reframe the proposal using what is proposed here.

Documentation/contributors/design-proposals/webhooks.md
```diff
@@ -0,0 +1,125 @@
...
+
+The new types can be added to the existing [StrategyDetailsDeployment](https://github.com/operator-framework/operator-lifecycle-manager/blob/4de8dd1a71001df91a5bd9517435a5aeba240920/pkg/controller/install/deployment.go#L33) type given that a webhook requires a deployment for operation.
+
+The [validating webhook](https://github.com/kubernetes/kubernetes/blob/ae3c44d043769305c6add665cd81c16319479ba7/pkg/apis/admissionregistration/types.go#L169) and [mutating webhook](https://github.com/kubernetes/kubernetes/blob/ae3c44d043769305c6add665cd81c16319479ba7/pkg/apis/admissionregistration/types.go#L293) types are extremely similar. Because of this similarity and due to the user not needing to provide service details new OLM specific types are to be created that include the critical fields.
+
+### CSV webhook configuration example
+
+(The following example has standard unaffected parts of configuration abbreviated.)
+
+...
```
ecordell: nit: ```yaml here instead so it shows nicely?

Documentation/contributors/design-proposals/webhooks.md
```diff
@@ -0,0 +1,125 @@
...
+        ...
+
+Notable omissions in the above example include:
+- volume and volumeMounts in the deployment for the certificates
+- services
+- clientConfig in the webhooks section (path was retained via webhookPath)
+
+### Securing webhook via certificates
+
+It is planned for OLM to handle certificate creation for securing webhook connection via CSRs (`CertificateSigningRequest`). When a CSR is issued, the API server uses its CA certificate to sign a certificate for consumption and in this case will be used for the webhook. By default the `ClusterSigningDuration` is set to be [one year](https://github.com/kubernetes/kubernetes/blob/c48ed1c5aca6e06b6a3d70d5f189216df3150379/pkg/controller/certificates/signer/config/v1alpha1/defaults.go#L44) and is set via a flag to the kube-controller-manager called "--experimental-cluster-signing-duration duration".
```
ecordell: What happens to the cert after the duration? Does olm need to create another CSR, or is there a mechanism to auto-request a new one?

jpeeler: (This is why this document is still marked as WIP.) OLM will need to create another CSR and there is no auto-renew mechanism like there is for kubelet certificates. In addition, we'll have to delete the old CSR and approve the new CSR as well.

jpeeler: I've removed this section after changing directions on CSR usage.

Documentation/contributors/design-proposals/webhooks.md
```diff
@@ -0,0 +1,114 @@
...
+          - apiGroups:
+            - "operators.olm.com"
+            apiVersions:
+            - v1alpha1
+            operations:
+            - create
+            resources:
+            - clusterserviceversions
+          selector:
+            name: mywebhook
```
jpeeler: Is this the right desired syntax or should this actually be a match label?

rcernich:
Regarding certificate management, have you thought about relying on cert-manager or OpenShift's service-ca?

PR #918, opened 2019-06-19
fix(build): set GO111MODULE=on

jpeeler:
This ensures builds that are done within GOPATH use modules and avoids this error message:
"build flag -mod=vendor only valid when using modules"


Makefile
```diff
@@ -15,6 +15,7 @@ IMAGE_TAG ?= "dev"
 KUBE_DEPS := api apiserver apimachinery apiextensions-apiserver kube-aggregator code-generator cli-runtime
 KUBE_RELEASE := release-1.14
 SPECIFIC_UNIT_TEST := $(if $(TEST),-run $(TEST),)
+export GO111MODULE=on
```
njhale: You could probably do something like:
```
export GO111MODULE=$(shell (go version | grep -q -E "1\.(11|12)") && echo on || echo auto)
```

jpeeler:
Verify is broken because it requires NOT using go modules...

PR #921, opened 2019-06-20
Add a target for compiling bare e2e

jpeeler:
This is just an idea of how to do this and will likely not go in.

Initially I tried using the $(CMDS) target, refactored it, and then ended up leaving it since it looked simpler. But maybe it's not worth changing...


jpeeler:
/retest
Ok given this is going in, we still have to call the new make target somewhere to make breakage show up. I didn't add it to any of the existing targets here and I don't think it should be really.

PR #923, opened 2019-06-22
Add Subscription InstallPlan Status

njhale:
Adds Subscription status conditions for referenced InstallPlans. 

Conditions added:
- `SubscriptionInstallPlanMissing`
- `SubscriptionInstallPlanPending`
- `SubscriptionInstallPlanFailed`


pkg/controller/operators/catalog/subscription/reconciler.go
```diff
@@ -96,6 +125,11 @@ func (c *catalogHealthReconciler) catalogHealth(namespace string) ([]v1alpha1.Su
 		catalogs = append(catalogs, globals...)
 	}
 
+	// Sort to ensure ordering
+	sort.Slice(catalogs, func(i, j int) bool {
```
ecordell: we only really need this for testing, right? is there any way we can just sort for tests and not for prod?

pkg/controller/operators/catalog/subscription/reconciler.go
```diff
@@ -151,30 +185,71 @@ func (c *catalogHealthReconciler) healthy(catalog *v1alpha1.CatalogSource) (bool
...
-		case SubscriptionDeletedState:
-			if onDelete != nil {
-				onDelete(s.Subscription())
-			}
-		case SubscriptionState:
-			if sync != nil {
-				err = sync(s.Subscription())
-			}
+// installPlanReconciler reconciles InstallPlan status for Subscriptions.
+type installPlanReconciler struct {
```
ecordell: installPlanStatusReconciler?

PR #924, opened 2019-06-25
Operator validation library project proposal

galletti94:
Please let us know your thoughts!

FYI @dweepgogia is the intern working on this project this summer

/cc @estroz @ecordell @jpeeler @tkashem

### DIFF
--- a/Documentation/contributors/design-proposals/operator-validation-library.md
+++ b/Documentation/contributors/design-proposals/operator-validation-library.md
@@ -1,0 +1,55 @@
+# Operator Validation Library
+
+### Problem Statement
+
+This project aims to create a validation library within OLM for operators managed by OLM. Such operators are manageable by creating a manifest bundle composed of [ClusterServiceVersions](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/Documentation/design/building-your-csv.md), [CustomResourceDefinitions](https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/), and [Package Manifest](https://github.com/operator-framework/operator-lifecycle-manager#discovery-catalogs-and-automated-upgrades) yamls. The steps to creating these files are error-prone and validation of this bundle happens in scattered places that are not always up to date with the latest changes to the manifest&#39;s definition.
+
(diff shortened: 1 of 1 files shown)


dweepgogia:
To review the progress so far, please see this PR: https://github.com/dweepgogia/new-manifest-verification/pull/3.

estroz:
The doc looks good. I would like to see discussion of how versioning will be handled before merging this. @ecordell correct me if I'm wrong but CSV versions don't change (often) but fields/requirements do within the same version. We could simply enforce whatever field requirements a version, ex. `v1alpha1`, has at the time.

Documentation/contributors/design-proposals/operator-validation-library.md
```diff
@@ -0,0 +1,55 @@
...
+
+In the future, we aim to extend this library to contain runtime tests and apis to be called by runtime tools on cluster.
+
+### Requirements
+
+- Methods of this library should adapt to any changes in OLM&#39;s type. There should be minimal to no changes required when there is a change in OLM&#39;s type and/or version. Or if this is not possible, methods of the library need to be part of the build/test pipeline of OLM.
+- Better error reporting by informing the user of all errors found instead of exiting after the first occurrence. We should define custom error types easily digestible by users of these APIs.
+- APIs should expose verification methods for individual yaml files as well as for the bundle as a whole.
+- This library should expose a validator interface for users to define their own custom validators (ex: operatorhub.io specific validators) while obeying the same overarching structure.
```
ecordell: How will we take advantage in practice? Have an operatorhub.io version of the validator tool? or can you load in additional validators at runtime?

galletti94: What we were thinking was that operatorhub.io folks would write a validator using the interface exposed in this library. Then whatever command line tool (that replaces what courier is doing) would call the apis from the library **and** load those validators defined by operatorhub.io (wherever those live). That tool could easily digest any number of such validators since they would have the same interface.

galletti94:
@ecordell we can get started on those points you raised. Should we just add those to the doc itself?

Just to clarify on 1), it's more important to us that we enable developers to easily write a CLI than actually writing the CLI ourselves if that makes sense. Do you think we should include such a CLI in the OLM code base as well or are you suggesting this for the purposes of flushing out requirements for the library's apis?

Documentation/contributors/design-proposals/operator-validation-library.md
```diff
@@ -0,0 +1,298 @@
...
+
+Here `path/to/bundle` is a directory structure as per the [operator manifest format](https://github.com/operator-framework/operator-registry#manifest-format). This command reports errors and/or warnings for each file in the bundle.
+
+Using the `help` flag with `manifest` command we get,
+
+```text
+Validates the manifest bundle as a whole, in addition to validating individual operator files. `manifest` reports errors/warnings for each file in the bundle. It works both with a directory structure (as per operator manifest format) and an operator image. `manifest` can also validate only the CSVs or CRDs present in a bundle. See flags for more information. 
+
+Usage:
+  operator-verify verify [flags]
```
galletti94: @dweepgogia I think this should be `operator-verify manifest [flags]` no?

dweepgogia: Yes, you're correct! Good catch, thanks!

dweepgogia:
@estroz @ecordell @njhale @jpeeler @tkashem @alecmerdler - Please review. Thanks!

estroz:
@dweepgogia Is this PR still valid?

dweepgogia:
> @dweepgogia Is this PR still valid?

Yes, it is. You can take a look at the  `fake readme` section to get a sense of how the final cli interface/apis would look like.

ecordell:
/lgtm

Sorry for the delay!

PR #925, opened 2019-06-26
Bug 1723818: CSV name change should not cause upgrade to fail

tkashem:
If a head CSV has a different name than the CSV it replaces, upgrade
fails because olm fails to transfer ownership of the APIService from
the previous CSV to the head CSV.

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1723818

### DIFF
--- a/pkg/controller/operators/olm/apiservices.go
+++ b/pkg/controller/operators/olm/apiservices.go
@@ -1,6 +1,7 @@
 package olm
 
 import (
+	"errors"
 	"fmt"
 	"strings"
 	"time"
@@ -65,18 +66,6 @@ func (a *Operator) checkAPIServiceResources(csv *v1alpha1.ClusterServiceVersion,
 	})
 
 	errs := []error{}
-	owners := []ownerutil.Owner{csv}
-
-	// Get replacing CSV if exists
-	replacing, err := a.lister.OperatorsV1alpha1().ClusterServiceVersionLister().ClusterServiceVersions(csv.GetNamespace()).Get(csv.Spec.Replaces)
-	if err != nil && !k8serrors.IsNotFound(err) {
-		logger.WithError(err).Warn("could not get replacement csv")
-		return err
-	}
-	if replacing != nil {
-		owners = append(owners, replacing)
-	}
-
 	ruleChecker := install.NewCSVRuleChecker(a.lister.RbacV1().RoleLister(), a.lister.RbacV1().RoleBindingLister(), a.lister.RbacV1().ClusterRoleLister(), a.lister.RbacV1().ClusterRoleBindingLister(), csv)
 	for _, desc := range csv.GetOwnedAPIServiceDescriptions() {
 		apiServiceName := desc.GetName()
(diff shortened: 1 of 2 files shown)


pkg/controller/operators/olm/operator.go
```diff
@@ -1416,6 +1416,73 @@ func (a *Operator) apiServiceOwnerConflicts(csv *v1alpha1.ClusterServiceVersion)
...
+		ownerNamespace := labels[ownerutil.OwnerNamespaceKey]
+
+		if ownerKind == "" || ownerNamespace == "" || ownerName == "" {
+			// If owner is set, do we always expect all three labels to be set?
+			// Question: Can a cluster scoped object ever own an APIService?
+			return ErrAPIServiceOwnerConflict
+		}
+
+		// Is it owned by this csv or the one it replaces.
+		if ownerutil.AdoptableLabels(apiService.GetLabels(), true, owners...) {
```
njhale: If you check the owner for existence before this, and add it to the `owners` slice _if it's not found_, then you should be able to get rid of the rest of this function.

pkg/controller/operators/olm/operator.go
```diff
@@ -1416,6 +1416,73 @@ func (a *Operator) apiServiceOwnerConflicts(csv *v1alpha1.ClusterServiceVersion)
...
+			if err != nil {
+				if !k8serrors.IsNotFound(err) && !k8serrors.IsGone(err) {
+					return err
+				}
+				
+				// The current owner CSV no longer exists, so we can adopt the APIService
+				continue
+			}
+
+			if currentOwnerCSV.Status.Phase == v1alpha1.CSVPhaseReplacing ||
```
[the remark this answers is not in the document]
tkashem: `CSV-C` either currently owns `APIService-A` or is trying to own it. If `CSV-C` currently owns it and is not being replaced or being deleted then the function will not allow `CSV-B` to take ownership. In the latter case where `CSV-C` is also trying to own it then it's a race to the ownership.

pkg/controller/operators/olm/apiservices.go
```diff
@@ -759,3 +750,54 @@ func APIServiceNameToServiceName(apiServiceName string) string {
...
+		err = ownerGetErr
+		return
+	}
+
+	owners := []ownerutil.Owner{target}
+	if replacing != nil {
+		owners = append(owners, replacing)
+	}
+	if currentOwnerCSV != nil && (
+		currentOwnerCSV.Status.Phase == v1alpha1.CSVPhaseReplacing || currentOwnerCSV.Status.Phase == v1alpha1.CSVPhaseDeleting) {
```
ecordell: I believe there's a set of phases that include these that you can check for membership

smarterclayton:
Looks like it’s still failing:

Jun 29 02:57:43.223: INFO: cluster upgrade is Progressing: Unable to apply 4.2.0-0.ci-2019-06-28-191017: the cluster operator operator-lifecycle-manager-packageserver has not yet successfully rolled out

https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/logs/release-openshift-origin-installer-e2e-aws-upgrade-4.1-to-4.2/131

PR #926, opened 2019-06-26
[WIP] Verify CRD validation schema to be structural before updating

dinhxuanvu:
CRD schema needs to be structural as a requirement for CRD to
be eligible for updating.

### DIFF
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -39,6 +39,7 @@ import (
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/operatorlister"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/ownerutil"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/queueinformer"
+	schema "github.com/operator-framework/operator-lifecycle-manager/pkg/lib/schema"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/metrics"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/controller/operators/catalog/operator.go
```diff
@@ -39,6 +39,7 @@ import (
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/operatorlister"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/ownerutil"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/queueinformer"
+	schema "github.com/operator-framework/operator-lifecycle-manager/pkg/lib/schema"
```
ecordell: Can we vendor the kube lib for this instead of copying it into our libs? Do we need to update our dependencies to 1.15?

dinhxuanvu: Ideally, we would update to 1.15 to take advantage of structural schema feature as even with this carried-over code, I had to strip down couple things due to missing fields in CRD. However, we are still using kube 1.12 for OLM so I wonder if a big jump kube-version-wise will be acceptable. I put up this PR to facilitate the discussion if structural schema should be a hard requirement for CRD update given non-structural schema is still supported.

njhale:
@dinhxuanvu Now that master uses kube 1.16, is this easier to get in? Is it obsolete?

dinhxuanvu:
@njhale I need to rebase this PR but yes it will be a small change to get this in now that master is in 1.16.

dinhxuanvu:
This PR is outdated and no longer needed. Closed.

PR #931, opened 2019-06-27
feat(csv): detect changes to a deployment and persist them

ecordell:
Prior to this PR, changes made to a CSV's deployment specs would not affect the running deployments. Instead, you would need to delete and recreate the CSV, or create a new one that replaces the existing one, to test changes to a deployment.

This is obviously a UX improvement for developing CSVs, but this also has an advantage for simplifying our deployment. We are looking at switching to kustomize instead of helm, which means we need to limit the templating we do, and specifically the lack of partials in kustomize means that it's difficult to support our current deployment model. With this change we can just include the packageserver CSV (instead of CSV + Deployment + CatalogSource + ConfigMap) in our release artifacts.

### DIFF
--- a/pkg/controller/install/deployment.go
+++ b/pkg/controller/install/deployment.go
@@ -6,6 +6,8 @@ import (
 	log "github.com/sirupsen/logrus"
 	appsv1 "k8s.io/api/apps/v1"
 	rbac "k8s.io/api/rbac/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
+	"k8s.io/apimachinery/pkg/util/diff"
 
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/wrappers"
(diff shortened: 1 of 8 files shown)


test/e2e/csv_e2e_test.go
```diff
@@ -2393,6 +2393,137 @@ func TestUpdateCSVMultipleIntermediates(t *testing.T) {
...
+
+	// Wait for updated CSV to succeed
+	updatedCSV, err := fetchCSV(t, crc, csv.Name, testNamespace, csvSucceededChecker)
+	require.NoError(t, err)
+
+	// Should have updated existing deployment
+	depUpdated, err := c.GetDeployment(testNamespace, strategyNew.DeploymentSpecs[0].Name)
+	require.NoError(t, err)
+	require.NotNil(t, depUpdated)
+	require.Equal(t, depUpdated.Spec.Template.Spec.Containers[0].Name, strategyNew.DeploymentSpecs[0].Spec.Template.Spec.Containers[0].Name)
```
njhale: Would it also help to have negative test cases (e.g. changes to the request field are not detected)?

test/e2e/csv_e2e_test.go
```diff
@@ -2393,6 +2393,137 @@ func TestUpdateCSVMultipleIntermediates(t *testing.T) {
...
+						Kind:        crdPlural,
+						DisplayName: crdName,
+						Description: "In the cluster",
+					},
+				},
+			},
+		},
+	}
+
+	// Don't need to cleanup this CSV, it will be deleted by the upgrade process
```
njhale: nit: left-over comment?

test/e2e/csv_e2e_test.go
```diff
@@ -2393,6 +2393,137 @@ func TestUpdateCSVMultipleIntermediates(t *testing.T) {
...
+	_, err = createCSV(t, c, crc, csv, testNamespace, false, false)
```
njhale: nit: the test cleaner will most likely delete the CSV, but I think we should explicitly delete every where it makes sense to, in case we decide to get rid of it later.

test/e2e/csv_e2e_test.go
```diff
@@ -2393,6 +2393,137 @@ func TestUpdateCSVMultipleIntermediates(t *testing.T) {
...
+				Spec: newNginxDeployment(nginxName),
+			},
+		},
+	}
+	strategyNewRaw, err := json.Marshal(strategyNew)
+	require.NoError(t, err)
+
+	fetchedCSV.Spec.InstallStrategy.StrategySpecRaw = strategyNewRaw
+
+	// Update CSV directly
```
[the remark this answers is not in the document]
ecordell: Good point!

test/e2e/csv_e2e_test.go
```diff
@@ -2393,6 +2393,163 @@ func TestUpdateCSVMultipleIntermediates(t *testing.T) {
...
+	_, err = crc.OperatorsV1alpha1().ClusterServiceVersions(testNamespace).Update(fetchedCSV)
+	require.NoError(t, err)
+
+	// Wait for updated CSV to succeed
+	_, err = fetchCSV(t, crc, csv.Name, testNamespace, csvSucceededChecker)
```
njhale: I think this can race with OLM's update of the CSV's status. Its deployment may not always be updated when this passes.

pkg/controller/install/deployment.go
```diff
@@ -179,10 +183,44 @@ func (i *StrategyDeploymentInstaller) checkForDeployments(deploymentSpecs []Stra
 				return StrategyError{Reason: StrategyErrReasonAnnotationsMissing, Message: fmt.Sprintf("annotations on deployment don't match. couldn't find %s: %s", key, value)}
 			}
 		}
+
+		// check equality
+		calculated := i.deploymentForSpec(spec.Name, spec.Spec)
+		if !i.equalDeployments(&calculated.Spec, &dep.Spec) {
+			return StrategyError{Reason: StrategyErrDeploymentUpdated, Message: fmt.Sprintf("deployment changed, rolling update with patch: %s\n%#v\n%#v", diff.ObjectDiff(dep.Spec.Template.Spec, calculated.Spec.Template.Spec), calculated.Spec.Template.Spec, dep.Spec.Template.Spec)}
```
tkashem: Question: I assume this error will cause the csv to go to a `Failed` state for a retry - in the next sync loop it will be set to `Pending` and redeployment will take place from there, correct?

ecordell: IIRC, we don't take a step through Failed if we know that a re-apply can fix it. It should go straight back to `Pending` for re-install.

test/e2e/csv_e2e_test.go
ecordell: fixed this

test/e2e/csv_e2e_test.go
```diff
@@ -148,7 +148,7 @@ func newNginxDeployment(name string) appsv1.DeploymentSpec {
 				Containers: []corev1.Container{
 					{
 						Name:  genName("nginx"),
-						Image: "bitnami/nginx:latest",
+						Image: "redis",
```
njhale: 🙂 
Note: It would be nice not to worry about this in the future. Maybe, in the future, we could use a standard base image configured to run a "generic test operator" script.

PR #932, opened 2019-06-28
Disable clusteroperator for package server

tkashem:
Disable clusteroperator for package server while we fix https://bugzilla.redhat.com/show_bug.cgi?id=1723818.

### DIFF
--- a/manifests/0000_50_olm_07-olm-operator.deployment.yaml
+++ b/manifests/0000_50_olm_07-olm-operator.deployment.yaml
@@ -26,8 +26,6 @@ spec:
           args:
           - -writeStatusName
           - operator-lifecycle-manager
-          - -writePackageServerStatusName
-          - operator-lifecycle-manager-packageserver
           - -tls-cert
           - /var/run/secrets/serving-cert/tls.crt
           - -tls-key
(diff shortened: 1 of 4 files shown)


smarterclayton:
This is likely to fail upgrade because it leaves a stranded operator version.

I would recommend reverting the change that broke the upgrade rather than completely disabling the operator (which is likely to break others who depend on it for testing)

smarterclayton:
(If this doesn’t fail upgrade tests that’s probably a bug in the upgrade tests...)

tkashem:
> This is likely to fail upgrade because it leaves a stranded operator version.
> 
> I would recommend reverting the change that broke the upgrade rather than completely disabling the operator (which is likely to break others who depend on it for testing)

@smarterclayton We have recently introduced this `clusteroperator` resource to track the deployment progress of `packageserver` component within olm. I don't believe this `clusteroperator` resource would exist in a 4.1 cluster.
No other test is depending on this newly introduced clusteroperator object yet, I believe.

I can definitely revert the PR that that introduced this clusteroperator object as well if that is desired. But I think it would have the same effect.

smarterclayton:
Upgrade tests run against both 4.1 and recent builds of 4.2, so the e2e-aws-upgrade job should be failing because you removed an operator it knows about

ecordell:
Keeping this staged just in case, but #925 should have fixed the underlying bug, making this unnecessary.

ecordell:
Fixed the underlying issue

PR #933, opened 2019-06-28
Validate existing Custom Resources against new schema prior updating

dinhxuanvu:
Allow CRD update if:

1. Now existing CRs found
2. Existing CRs is validated against new CRD schema

### DIFF
--- a/Documentation/design/dependency-resolution.md
+++ b/Documentation/design/dependency-resolution.md
@@ -2,14 +2,22 @@
 
 OLM manages the dependency resolution and upgrade lifecycle of running operators. In many ways, thes problems OLM faces are similar to other OS package managers like `apt`/`dkpg` and `yum`/`rpm`.
 
-However, there is one constraint that similar systems don't generally have that OLM does: because operators are always running, OLM attempts to ensure that at no point in time are you left with a set of operators that do not work with each other. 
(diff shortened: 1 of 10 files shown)


Documentation/design/dependency-resolution.md
```diff
@@ -2,15 +2,23 @@
...
 This means that OLM needs to never:
- 
+
  - install a set of operators that require APIs that can't be provided
  - update an operator in a way that breaks another that depends upon it
- 
+
 The following examples motivate why OLM's dependency resolution and upgrade strategy works as it does, followed by a description of the current algorithm.
 
+## CustomResourceDefinition (CRD) Upgrade
```
ecordell: 👍

pkg/controller/operators/catalog/operator.go
```diff
@@ -114,6 +125,7 @@ func NewOperator(ctx context.Context, kubeconfigPath string, clock utilclock.Clo
 		logger:                 logger,
 		clock:                  clock,
 		opClient:               opClient,
+		dynamicClient:          dynamicClient,
```
ecordell: 👍 as a bonus, this will let us support "any" bundle object in the future and clean up ExecutePlan

PR #936, opened 2019-06-29
Add skip range to packageserver templates and manifests

njhale:
Adds an `olm.skipRange` annotation to packageserver CSVs. The range includes any version less than the most current (i.e. `<0.10.1`). This should let OLM upgrade packageserver to $CHANNEL_HEAD from any previous version.


njhale:
superseded by #937

PR #937, opened 2019-06-29
1723818: package server 4.1 to 4.2 upgrade should not fail

tkashem:
packageserver v0.10.1 `replaces` v0.9.0. This should kick off a
trigger that sets v0.9.0 to Replacing phase. The head CSV then
can take ownership of the APIService object.

TODO:
- I directly modified the manifests folder. In a follow up PR
  I will add the changes to the chart templates.

### DIFF
--- a/manifests/0000_50_olm_11-olm-operators.configmap.yaml
+++ b/manifests/0000_50_olm_11-olm-operators.configmap.yaml
@@ -9,6 +9,127 @@ data:
     - apiVersion: operators.coreos.com/v1alpha1
       kind: ClusterServiceVersion
       metadata:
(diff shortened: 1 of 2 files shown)


manifests/0000_50_olm_11-olm-operators.configmap.yaml
```diff
@@ -6,6 +6,127 @@ metadata:
 data:
   customResourceDefinitions: |-
   clusterServiceVersions: |-
```
tkashem: I changed configmap.yaml to keep both records in sync, not sure if we will run into validation error from registry otherwise.

PR #938, opened 2019-07-01
test(e2e): remove extra resource comparison

njhale:
Remove extra resource comparison at the end of two tests since
eventually consistent orthogonal conditions sometimes cause a false
negative.


jpeeler:
/lgtm
Honestly I wondered why this was here and need to remove it from some of my copy/pasting.

alecmerdler:
I had added these long ago to catch runaway OLM control loops causing the UI to break (impossible to update a resource using YAML editor). We can catch unnecessary updates to objects using the `Watch()` PR once that is merged.

PR #939, opened 2019-07-01
(proposal) improved csv status reporting

tkashem:
https://jira.coreos.com/browse/OLM-1053


ecordell:
/lgtm

This is a good starting point for conditions for CSVs. When we get to implementing this we can bikeshed about the actual fields/etc 😄

PR #940, opened 2019-07-02
Rebase k8s 1.15.0

dinhxuanvu:


shawn-hurley:
/hold

Just to make sure that this doesn't accidentally merge

ecordell:
This has diverged quite a bit, and we'll need to do it roughly in-sync with openshift (until we split into an upstream/downstream)

PR #942, opened 2019-07-03
Remove package server subscription 

ecordell:
This PR includes #931 but also simplifies our deploy manifests by removing the catalog and subscription for package-server.


njhale:
`e2e-aws` is failing because [an OLM OpenShift conformance test](https://github.com/openshift/origin/blob/527734c686f04aaf3e9b33ecb5875f6044faa1b3/test/extended/operators/olm.go#L36) expects that at least one instance of each OLM kind exists on a given cluster. We could potentially ask @jianzhangbjz to remove or modify this expectation in [openshift/origin #223224](https://github.com/openshift/origin/pull/23224/files).

manifests/0000_50_olm_15-packageserver.clusterserviceversion.yaml
```diff
@@ -9,7 +9,6 @@ spec:
   displayName: Package Server
   description: Represents an Operator package that is available from a given CatalogSource which will resolve to a ClusterServiceVersion.
   minKubeVersion: 1.11.0
-  replaces: packageserver.v0.9.0
```
njhale: Why remove this? During an OpenShift upgrade, how will OLM know to clean up the `replaced` CSV, or to change ownership of the `APIService`?

ecordell: I updated the init to delete `packageserver.v0.9.0` and `packageserver.v0.10.0` - so whatever was there before, we're now moving to just having a single `packageserver` that we can update.

njhale: are there clusters that still have 0.10.1?

PR #943, opened 2019-07-04
Add simple benchmark for namespaced subscription resolution

njhale:
Adds a simple benchmark test for the `syncResolvingNamespace` method. 

I initially made this to catch some memory utilization issues, but thought it could also be useful as a reference for other benchmarks we may write.


jpeeler:
Cool idea. What do you think about adding in a generous threshold that triggers failure?

jpeeler:
/lgtm
Just a timeout change since I last looked over it.

PR #946, opened 2019-07-09
proposal(relatedImages): add a proposal for specifying related images in CSVs

ecordell:
There is still plenty of room for improvement, but I wanted to get a strawman out for discussion.

This proposal places the info as a new field on the `CSV` `spec`. This is so that we can take advantage of validation from kubernetes for these fields, and so that we can (in a future iteration) take advantage of this information at runtime. The two use cases I could think of were: 

- specifying that some image availability should prevent installation of the operator
- projecting the relatedimages down into the operator so that the value can be used at runtime

### DIFF
--- a/Documentation/contributors/design-proposals/related-images.md
+++ b/Documentation/contributors/design-proposals/related-images.md
@@ -1,0 +1,192 @@
+# Related Images
+
+Status: Pending
+
+Version: Alpha
+
+Implementation Owner: ecordell 
+
+## Motivation
+
+Operators often need to make use of other container images to perform their functions as operators. 
+
+## Proposal
+
+Introduce a new field `relatedImages` to the `ClusterServiceVersion` spec. 
+
+### ClusterServiceVersion Spec Changes
+
+A new section `relatedImages` is added to the ClusterServiceVersionSpec.
+
+```yaml
+kind: ClusterServiceVersion 
+metadata:
+  name: etcd-operator
+spec:
+  relatedImages:
+  - name: default
+    image: quay.io/coreos/etcd@sha256:12345 
+    annotation: default
+  - name: etcd-2.1.5
+    image: quay.io/coreos/etcd@sha256:12345 
+  - name: etcd-3.1.1
+    image: quay.io/coreos/etcd@sha256:12345 
+```
+
+These will be made available as annotations on the operator deployments, so that they can be used via downward API if desired. This may be particularly useful for operators that are tightly coupled to another particular image.
+
+```yaml
+kind: Deployment
+metadata:
+  name: etcd-operator
+  annotations:
+    default: quay.io/coreos/etcd@sha256:12345 
+    olm.relatedImage.etcd-2.1.5: quay.io/coreos/etcd@sha256:12345 
+    olm.relatedImage.etcd-3.1.1: quay.io/coreos/etcd@sha256:12345 
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      name: etcd-operator
+  template:
+    metadata:
+      name: etcd-operator
+      labels:
+        name: etcd-operator
+    spec:
+      serviceAccountName: etcd-operator
+      containers:
+      - name: etcd-operator
+        command:
+        - etcd-operator
+        - --create-crd=false
+        - --defaultImage=${DEFAULT}
+        image: quay.io/coreos/etcd-operator@sha256:c0301e4686c3ed4206e370b42de5a3bd2229b9fb4906cf85f3f30650424abec2
+        env:
+        - name: NAMESPACE
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.namespace
+        - name: DEFAULT
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.annotations['default']
+```
+
+### Implementation
+
+#### ClusterServiceVersion Spec and Status
+
+Spec needs to be updated to include the fields described above, and the openapi validation should be updated as well.
+
+Note that the `name` field for related images must satisfy the annotation name conventions.
+
+#### Install Strategy
+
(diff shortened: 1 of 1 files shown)


Documentation/contributors/design-proposals/related-images.md
```diff
@@ -0,0 +1,192 @@
...
+        name: etcd-operator
+    spec:
+      serviceAccountName: etcd-operator
+      containers:
+      - name: etcd-operator
+        command:
+        - etcd-operator
+        - --create-crd=false
+        - --defaultImage=${DEFAULT}
+        image: quay.io/coreos/etcd-operator@sha256:c0301e4686c3ed4206e370b42de5a3bd2229b9fb4906cf85f3f30650424abec2
```
dmesser: Should this not actually be an example where the downward-API is used to retrieve the value for `image`?

ecordell: This is the actual operator image; we can't set that via downwardAPI

ralphbean:
Thanks for this!  My use cases here are mostly about security metadata.

* Outside of the context of OLM, we can use relatedImages to drive automated CVE rebuilds.  i.e., if an operand is affected by a vulnerability and we know this, then we can respin it producing a new digest, and then traverse to rebuild any operator(s) that previously referenced its old digest with the new digest.
* Same goes for displaying vulnerability scan data for an operator as a whole.   With the relatedImages listed explicitly here, we can build tooling to display scan results for the operator image and all of the operands together.

Documentation/contributors/design-proposals/related-images.md
```diff
@@ -0,0 +1,192 @@
...
```
ralphbean: Is the `etcd-operator` image intentionally not listed among the `relatedImages`?

ralphbean: Oh, looks like @dmesser had [the same question](https://github.com/operator-framework/operator-lifecycle-manager/pull/946/files#r301639434).

ralphbean: On thinking about this a bit more, I think I may need the operator image to appear in the `relatedImages` list, even if it cannot be usefully used by OLM.  Needed for the security manifest use cases mentioned above.

ralphbean:
An idea (not sure if it is sane):  can OLM create a `ImageContentSourcePolicy` to enforce that the operator only use those images listed as `relatedImages`?

Documentation/contributors/design-proposals/related-images.md
shawn-hurley: I am also wondering if some other value should be more readily available for the operator's image in the spec. This would make it so that others would only have to look at something like `spec.images` or something to get all the images that a CSV would need/deploy?

njhale:
Adding a note here that we discussed using image labels as an alternative solution.

Documentation/contributors/design-proposals/related-images.md
```diff
@@ -0,0 +1,192 @@
...
+  name: etcd-operator
+spec:
+  relatedImages:
+  - name: default
+    image: quay.io/coreos/etcd@sha256:12345 
+    annotation: default
+  - name: etcd-2.1.5
+    image: quay.io/coreos/etcd@sha256:12345 
+  - name: etcd-3.1.1
+    image: quay.io/coreos/etcd@sha256:12345
```
ralphbean: WDYT about listing these as objects instead of string?  Something like:

```
- name: etcd-3.1.1
  image:
    registry: quay.io
    repository: coreos/etcd
    reference: sha256:12345
```

In one of my use cases, I'd like to rebuild operator metadata when automatically rebuilding one of the operands.  It would be nice to not have to *parse* the `image` string in order to substitute in a new sha256 value automatically.

ecordell:
@shawn-hurley @dmesser @ralphbean I've updated the proposal to separate the image metadata from the CSV for this first step. I'm curious what you all think of this approach.

Documentation/contributors/design-proposals/related-images.md
```diff
@@ -0,0 +1,129 @@
...
+  newTag: 1.8.0
+- name: my-demo-app
+  newName: my-app
+- name: alpine
+  digest: sha256:24a0c4b4a4c0eb97a1aabb8e29f18e917d05abfe1b7a7c07857230879ce7d3d3
+```
+
+([image type](https://github.com/kubernetes-sigs/kustomize/blob/master/pkg/image/image.go))
+
+The images in the `image` list will be used to determine the set of related images that are required for the operator.
```
ralphbean: Typo, "`images` list"

Documentation/contributors/design-proposals/related-images.md
```diff
@@ -0,0 +1,129 @@
...
+The list of images will then be:
+
+```
+quay.io/coreos/etcd-operator:latest
+quay.io/coreos/etcd:3.0.5
+quay.io/coreos/etcd@sha256:24a0c4b4a4c0eb97a1aabb8e29f18e917d05abfe1b7a7c07857230879ce7d3d3
+```
+
+Note that `quay.io/coreos/etcd-operator@sha256:66a37fd61a06a43969854ee6d3e21087a98b93838e284a6086b13917f96b0d9b` is not
+included, since it would be replaced with `:latest` if the `kustomization.yaml` were applied.
```
jmazzitelli: Just a comment - the relationship between the key names in **bundle/csv.yaml** and **bundle/kustomization.yaml**
 are confusing because I can't tell if `name` in **bundle/kustomization.yaml** refers to the `name` or `image` in **bundle/csv.yaml** ... you would think `name` since that is the same key-name in both, but here it seems to be saying, no, `name` in **bundle/kustomization.yaml** refers to `image` in **bundle/csv.yaml** (or at least refers to a part of the `image` value).

ecordell:
Closing as we work on a new set of proposals for collecting and transmitting this type of metadata.

johnfriz:
@ecordell what's the next step here in terms of moving from proposal to POC / implementation? And idea on timeframe as to when this might land?

PR #948, opened 2019-07-09
(status) update ClusterOperatorStatus builder

awgreene:
Problem: The ClusterOperatorStatus builder does not provide a way to
report a reason when degraded. When an operator reports degraded,
OpenShift telemetry only provided the status (degraded == true) and
the reason. Also - the builder does not include the Upgradeable
condition.

Solution: Create and update the functions provided by the builder.


awgreene:
Earlier today @tkashem asked me to add a builder test before this goes through.

/hold

ecordell:
Another PR introduced `WithUpgradeable`, so this just needs the reason changes and it should be good to go.

ecordell:
Closing for now, we can re-assess if we need these changes later.

PR #949, opened 2019-07-11
proposal(api): add installed operator api proposal

njhale:
Adds a proposal for a new API that acts as a replacement for copied CSVs.

### DIFF
--- a/Documentation/contributors/design-proposals/installed-api.md
+++ b/Documentation/contributors/design-proposals/installed-api.md
@@ -1,0 +1,282 @@
+# Installed Operator API
+
+Status: Pending
+
+Version: Alpha
+
+Implementation Owner: TBD
+
+## Motivation
+
+Today, information about operators configured to watch a given namespace is surfaced via "copied" `ClusterServiceVersion`s (`CSV`s). This means that for each __<CSV, watched namespace>__ pair, OLM must manage a copied CSV instance. Additionally, `CSV`s often contain large amounts of metadata, most notably a [base64 encoded icon](https://github.com/operator-framework/operator-lifecycle-manager/blob/18b6c0d2edba2534b4726138b225fff066aae99a/manifests/0000_50_olm_03-clusterserviceversion.crd.yaml#L114). As of OpenShift 4.2, there are more than a dozen namespaces that globally installed operators have their `CSV`s copied to, which causes OLM's `CSV` cache (and memory consumption) to grow large (~150MB) with a fairly low number of operators installed (~5). 
+
+Due to these factors, an additional means of providing this information is required.
+
+## Proposal
+
+Define a new read-only API resource, `installed.packages.operators.coreos.com`, that surfaces metadata about an operator installed and configured to watch a namespace. Back that API by an extension api-server that _synthesizes_ those resources on request.
+
+### Naming
+
+The resource will share a name with the `CSV` of the installed operator it's tracking.
+
+### Namespaces
+
+The resource will be namespaced. Additionaly, it will exist only when an operator that provides a package matching its `metadata.name` is installed and configured to watch its namespace.
+
+### Spec and Status
+
+The resource __will not__ define a `spec` field because it's intended to aggregate the state of existing resources. In other words, there is no meaning in declaring a _desired state_. However, It __will__ define a `status` field that embeds the latest `CSV` and `Subscription`. Embedding entire resources instead of their spec or status directly allows tracked resources to be added or removed more easily in subsequent API changes.
+
+### Labels and Annotations
+
+Labels of the format `operators.coreos.com/<short>: <name>` will be added for each resource tracked in the `status` (see the [k8s docs](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set) for the inspiration behind the key prefix). For tracking `CSV`s and `Subscription`s, the label keys would include `operators.coreos.com/csv` and `operators.coreos.com/sub`.
+
+Labels on `Installed` resources must provide enough information to be queriable by the [openshift console](https://github.com/openshift/console). As such, labels may need to be projected from tracked resources or added as needed.
+
+### The `Installed` Resource
+
+Example `Installed` resource for the `etcdoperator.v0.9.4-clusterwide` `CSV`:
+
+```yaml
+apiVersion: v1alpha2/packages.operators.coreos.com
+kind: Installed
+metadata:
+  labels:
+    # Names of related resources
+    operators.coreos.com/sub: etcd
+    operators.coreos.com/csv: etcdoperator.v0.9.4-clusterwide
+    # TODO: Project labels from CSV/Subscription?
+  # Named after the package the installed operator belongs to
+  name: etcdoperator.v0.9.4-clusterwide
+  namespace: local
+status:
+    # Entire Subscription responsible for installation
+    subscription:
+        apiVersion: operators.coreos.com/v1alpha1
+        kind: Subscription
+        metadata:
+            name: etcd
+            namespace: local-operators
+        spec:
+            # <subscription spec>
+        status:
+            # <subscription status>
+
+    # Sanitized CSV responsible for installation
+    clusterServiceVersion:
+        apiVersion: operators.coreos.com/v1alpha1
+        kind: ClusterServiceVersion
+        metadata:
+            annotations:
+                # Annotations are sanitized to remove sensitive info such as `olm.targetNamespaces`
+                tectonic-visibility: ocs
+                repository: 'https://github.com/coreos/etcd-operator'
+                alm-examples: |
+                    # <alm examples>
+                capabilities: Full Lifecycle
+                olm.operatorNamespace: operators
+                containerImage: >-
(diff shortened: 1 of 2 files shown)


Documentation/contributors/design-proposals/installed-api.md
```diff
@@ -0,0 +1,282 @@
...
+
+## Motivation
+
+Today, information about operators configured to watch a given namespace is surfaced via "copied" `ClusterServiceVersion`s (`CSV`s). This means that for each __<CSV, watched namespace>__ pair, OLM must manage a copied CSV instance. Additionally, `CSV`s often contain large amounts of metadata, most notably a [base64 encoded icon](https://github.com/operator-framework/operator-lifecycle-manager/blob/18b6c0d2edba2534b4726138b225fff066aae99a/manifests/0000_50_olm_03-clusterserviceversion.crd.yaml#L114). As of OpenShift 4.2, there are more than a dozen namespaces that globally installed operators have their `CSV`s copied to, which causes OLM's `CSV` cache (and memory consumption) to grow large (~150MB) with a fairly low number of operators installed (~5). 
+
+Due to these factors, an additional means of providing this information is required.
+
+## Proposal
+
+Define a new read-only API resource, `installed.packages.operators.coreos.com`, that surfaces metadata about an operator installed and configured to watch a namespace. Back that API by an extension api-server that _synthesizes_ those resources on request.
```
ecordell: I’m wary of mixing up our terms here - package is the entity in a catalog that contains many csvs / operators. 

So I would think that, for example, if we keep this as “packages” we would include much more information than the CSV, like updates available in other channels, etc

Documentation/contributors/design-proposals/installed-api.md
```diff
@@ -0,0 +1,282 @@
...
+
+The resource will share a name with the `CSV` of the installed operator it's tracking.
+
+### Namespaces
+
+The resource will be namespaced. Additionaly, it will exist only when an operator that provides a package matching its `metadata.name` is installed and configured to watch its namespace.
+
+### Spec and Status
+
+The resource __will not__ define a `spec` field because it's intended to aggregate the state of existing resources. In other words, there is no meaning in declaring a _desired state_. However, It __will__ define a `status` field that embeds the latest `CSV` and `Subscription`. Embedding entire resources instead of their spec or status directly allows tracked resources to be added or removed more easily in subsequent API changes.
```
ecordell: You’ve read over the kube api conventions more recently than I - isn’t it fine to omit spec and status in cases like this? Should we just not have them?

Documentation/contributors/design-proposals/installed-api.md
ecordell: Ah, I now that I’ve read further I see that it does contain more than the CSV :+1:

Documentation/contributors/design-proposals/installed-api.md
```diff
@@ -0,0 +1,282 @@
...
+    # Names of related resources
+    operators.coreos.com/sub: etcd
+    operators.coreos.com/csv: etcdoperator.v0.9.4-clusterwide
+    # TODO: Project labels from CSV/Subscription?
+  # Named after the package the installed operator belongs to
+  name: etcdoperator.v0.9.4-clusterwide
+  namespace: local
+status:
+    # Entire Subscription responsible for installation
+    subscription:
```
ecordell: I think it’s worth exploring a variant that doesn’t have the objects nested, but just has the fields we want. 

The way you have here could be a. Easier transition for the UI, though

Documentation/contributors/design-proposals/installed-api.md
```diff
@@ -0,0 +1,282 @@
...
+
+The `Installed` resource belongs to the same API group (`packages.operators.coreos.com`) as `PackageManifest`, but at different versions (`v2alpha1` and `v1` respectively). It follows that they should share the internal types and conversion logic, which makes `packageserver` an ideal place to add this API. However, due to [current limitations in OLM](https://github.com/operator-framework/operator-lifecycle-manager/issues/727) a separate `Deployment` must be introduced into the `packageserver` `CSV` to support the new version (`v2alpha1`). To this end, the names of each deployment will be of the format `packageserver.<version>` and will be referenced in their respective `owned` entry:
```
ecordell: I’m not against putting this in packageserver but I don’t really see any particular reason to. 

I don’t believe there’s any shared functionality between this proposed apiserver and the packageserver?

Documentation/contributors/design-proposals/installed-api.md
```diff
@@ -0,0 +1,282 @@
...
+
+### Deprecating Copied `CSV`s
+
+In the long-term, all logic generating copied `CSV`s will _eventually_ be removed.
+
+Changes removing copied `CSV`s from OLM:
+
+- Remove `CSV` copying during `OperatorGroup` reconciliation
+- Remove copied `CSV` queues, sync functions, and garbage collection
+- Update garbage collection for target namespace RBAC (can no longer add `OwnerReference`s to target `CSV`s)
```
ecordell: Can’t we have the ownerrefs point to the installed object? 

We might need to be careful to generate uids. I think CSV uid mangled with target namespace would work

Documentation/contributors/design-proposals/installed-api.md
```diff
@@ -0,0 +1,282 @@
...
+  - Add surrogate resource in target namespace to use for `OwnerReference` and garbage collection
```
ecordell: If we still need a surrogate resource, that resource will have need almost exactly the csv copying logic (modulo status updates) so we might not want to throw that out. 

But I would rather use ownerefs to this new API unless that is impossible for some reason

Documentation/contributors/design-proposals/installed-api.md
njhale: That's fair. The reason I went with nesting is because I thought it would provide an easier path when our APIs become unified - though I could be wrong.

Documentation/contributors/design-proposals/installed-api.md
njhale: If they share an API group, it behooves us from an internal types/codegen perspective to have versioned structures exist in the same go package. The servers themselves could be separate and live in separate packages.

Documentation/contributors/design-proposals/installed-api.md
njhale: I just want to be cautious when it comes to adding ownerrefs to resources that aren't actually stored in etcd (e.g. what happens if the aggregated API is unavailable? I'm guessing GC runs on the children).

ecordell: We can review the GC code again, but IIRC GC controller won't GC if the apiserver is unavailable. The API needs to be up and respond correctly for it to process.

Documentation/contributors/design-proposals/installed-api.md
njhale: Here, I felt that `status` would imply intent; that it would be easier for users to understand that they can't modify the listed resources. After re-reading the api conventions' [section on subresources](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#types-kinds), I think omitting the `status` and pulling the tracked resources to the top-level (as I think you are alluding to) is fine as well.

Do you have a preference?

Documentation/contributors/design-proposals/installed-api.md
njhale: I took a look and I think [you're right](https://github.com/kubernetes/kubernetes/blob/aaad86c2105bfbf98f26745bf8262ab8c357e8d3/pkg/controller/garbagecollector/garbagecollector.go#L340). 

I also agree it would be cleaner for the `Installed` resource to be the referent. There a couple of things I think we need to watch out for:
-  For GC to be responsive, we must [support `delete`, `list`, and `watch`](https://github.com/kubernetes/kubernetes/blob/aaad86c2105bfbf98f26745bf8262ab8c357e8d3/pkg/controller/garbagecollector/garbagecollector.go#L650) verbs for `Installed` resources
- To prevent GC from cleaning up the referring resources when the aggregated apiserver isn't fully initialized (serving discovery, but synthesized resources aren't yet populated), we should prevent readiness until synthetic resource caches have been populated at least once.
- Generate deterministic UIDs (your suggestion would work)

Let me know if this makes sense so I can get started amending the proposal.

ecordell: Yeah that sounds great - but how can we support `delete`?

Documentation/contributors/design-proposals/installed-api.md
ecordell: I think omitting sounds good but it might be one of those things where you try it and see how it looks?

Documentation/contributors/design-proposals/installed-api.md
ecordell: Since we've been talking about moving away from / combining the two, so I thought not having the names there would make it simpler?

Documentation/contributors/design-proposals/installed-api.md
ecordell: Why should they or should they not share an apigroup?

Documentation/contributors/design-proposals/installed-api.md
njhale: make it a no-op?

ecordell:
@njhale closing because I believe this will be superseded. Correct me if I'm wrong.

PR #950, opened 2019-07-11
Pod configuration

tkashem:
As an OLM user, I would like to set configuration for operator pods, so that they are persisted between operator upgrades.
https://jira.coreos.com/browse/OLM-1163

Pending issue:
* `Proxy.Status` is not populated on the cluster yet. We read proxy env variable(s) off of Proxy.Status as per instruction.

TODO(s):
- [ ] Pin to `k8s 1.14` libs. Once https://github.com/operator-framework/operator-lifecycle-manager/pull/954 is merged then do a rebase. 
- [ ] From the `resolver` we get the owner ( this is the csv that owns the deployment ). I need to get to the subscription that owns this csv to get the pod configuration. Right now, we are using a subscription lister to list all subscription(s) in the same namespace and then find the owner subscription by matching `subscription.Status.InstalledCSV`. We can use `annotations/label` to get to the subscription. This can be done in a separate PR as well.
- [ ] Subscription status is not updated with the result of env injection operation. We can address this in a follow up PR.
- [x] Add e2e test to validate that env variable(s) in 'cluster' Proxy object propagates to the deployment.
- [ ] OpenAPI validation for pod configuration
- [ ] Add support for `EnvFrom`


pkg/controller/install/resolver.go
```diff
@@ -47,7 +49,13 @@ func (r *StrategyResolver) InstallerForStrategy(strategyName string, opClient op
 	switch strategyName {
 	case InstallStrategyNameDeployment:
 		strategyClient := wrappers.NewInstallStrategyDeploymentClient(opClient, opLister, owner.GetNamespace())
-		return NewStrategyDeploymentInstaller(strategyClient, annotations, owner, previousStrategy)
+
+		initializers := []DeploymentInitializerFunc{}
+		if r.ProxyInjectorBuilder != nil {
+			initializers = append(initializers, r.ProxyInjectorBuilder(owner))
+		}
+
```
tkashem: This is a little bit of a hack for now, I am leaving it as a todo.

pkg/lib/proxy/syncer.go
```diff
@@ -0,0 +1,87 @@
...
+// returns the proxy env variable(s) to the user.
+func (w *Syncer) QueryProxyConfig() (proxy []corev1.EnvVar, err error) {
+	global, getErr := w.lister.Get(globalProxyName)
+	if getErr != nil {
+		if !k8serrors.IsNotFound(getErr) {
+			err = getErr
+			return
+		}
+
+		err = fmt.Errorf("global Proxy configuration not defined in '%s' - %v", globalProxyName, getErr)
```
tkashem: If an openshift cluster does not have the `cluster` proxy object then deployment initialization will fail and the operator install will fail as well.

pkg/controller/operators/olm/envvar/config.go
```diff
@@ -0,0 +1,45 @@
...
+	return
+}
+
+func findOwner(list []*v1alpha1.Subscription, ownerCSV ownerutil.Owner) *v1alpha1.Subscription {
+	for i := range list {
+		sub := list[i]
+		if sub.Status.InstalledCSV == ownerCSV.GetName() {
+			return sub
+		}
+	}
```
tkashem: It's a little hacky too! there should be a more straightforward way to get to the subscription from the csv. I am tracking it as a todo.

pkg/lib/proxy/syncer.go
ecordell: I don't think it's required that an openshift cluster have a cluster proxy object, right? Just that if it is there, we should respect it?

pkg/controller/operators/olm/envvar/inject.go
```diff
@@ -0,0 +1,61 @@
...
+		container := &podSpec.Containers[i]
+		container.Env = merge(container.Env, envVars)
+	}
+
+	return nil
+}
+
+func merge(containerEnvVars []corev1.EnvVar, newEnvVars []corev1.EnvVar) (merged []corev1.EnvVar) {
+	merged = containerEnvVars
+
```
tkashem: Thoughts?
- If an `env` variable that is being injected is already defined in the `csv` deployment spec, it will be overwritten.
- For proxy env var, empty means unset and will not result in an env var. Can we say the same for all other env variable(s) in pod configuration?

pkg/lib/proxy/syncer.go
tkashem: Yes, this is much safer. And for a scenario where there was a `cluster` object and then it gets removed later - we will not find a `cluster` proxy object and generate a pod spec without the proxy env vars. This will cause a mismatch of the desired and observed and eventually update deployment(s) (originally created with the proxy env variable(s) )  to not have them any longer, correct?

pkg/controller/operators/olm/envvar/config.go
ecordell: This is fine for now - we will eventually start storing the CSV data on the subscription so this will get fixed

pkg/controller/operators/olm/operator.go
```diff
@@ -212,6 +214,19 @@ func newOperatorWithConfig(ctx context.Context, config *operatorConfig) (*Operat
 			return nil, err
 		}
 
+		subInformer := extInformerFactory.Operators().V1alpha1().Subscriptions()
```
ecordell: Now that we've done this, we need to get moving on combining subscription/csv so we don't have to cache both objects in both operators

pkg/controller/operators/olm/operator.go
```diff
@@ -392,13 +407,71 @@ func newOperatorWithConfig(ctx context.Context, config *operatorConfig) (*Operat
 		return nil, err
 	}
 
+
+	// setup proxy env var injection policies
+	discovery := config.operatorClient.KubernetesInterface().Discovery()
+	proxyAPIExists, err := proxy.IsAPIAvailable(discovery)
+	if err != nil {
+		op.logger.Errorf("error happened while probing for Proxy API support - %v", err)
+		return nil, err
```
ecordell: this shouldn't error out if we're not on an openshift cluster, right?

pkg/controller/operators/olm/operator.go
```diff
@@ -392,13 +407,71 @@ func newOperatorWithConfig(ctx context.Context, config *operatorConfig) (*Operat
 		return nil, err
 	}
 
+
+	// setup proxy env var injection policies
+	discovery := config.operatorClient.KubernetesInterface().Discovery()
+	proxyAPIExists, err := proxy.IsAPIAvailable(discovery)
```
ecordell: For the cluster operator API, we poll for a while to see if the api is ready. We should think about doing that here too.

pkg/lib/proxy/querier.go
```diff
@@ -0,0 +1,26 @@
+package proxy
+
+import (
+	corev1 "k8s.io/api/core/v1"
+)
+
+// DefaultQuerier does...
+func DefaultQuerier() Querier {
```
ecordell: Default is fine, I think we use `Noop` for this type of object elsewhere...

test/e2e/subscription_e2e_test.go
```diff
@@ -1089,6 +1115,132 @@ func TestSubscriptionInstallPlanStatus(t *testing.T) {
 	}
 }
 
+func TestCreateNewSubscriptionWithPodConfig(t *testing.T) {
+	defer cleaner.NotifyTestComplete(t, true)
+
+	newConfigClient := func(t *testing.T) configv1client.ConfigV1Interface {
+		config, err := clientcmd.BuildConfigFromFlags("", *kubeConfigPath)
+		// require.NoErrorf(t, err, "could not create rest config: %s", err.Error())
```
ecordell: nit: commented code

test/e2e/subscription_e2e_test.go
```diff
@@ -1089,6 +1115,132 @@ func TestSubscriptionInstallPlanStatus(t *testing.T) {
...
+	require.NotNil(t, subscription)
+
+	csv, err := fetchCSV(t, crc, subscription.Status.CurrentCSV, testNamespace, buildCSVConditionChecker(v1alpha1.CSVPhaseSucceeded))
+	require.NoError(t, err)
+
+	proxyEnv := proxyEnvVarFunc(t, config)
+	expected := podEnv
+	expected = append(expected, proxyEnv...)
+
+	checkDeploymentWithPodConfiguration(t, c, csv, podConfig.Env)
```
ecordell: 👍

pkg/controller/install/resolver.go
```diff
@@ -28,7 +28,9 @@ type StrategyResolverInterface interface {
 	InstallerForStrategy(strategyName string, opClient operatorclient.ClientInterface, opLister operatorlister.OperatorLister, owner ownerutil.Owner, annotations map[string]string, previousStrategy Strategy) StrategyInstaller
 }
 
-type StrategyResolver struct{}
+type StrategyResolver struct {
+	ProxyInjectorBuilder DeploymentInitializerFuncBuilder
```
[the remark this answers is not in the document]
tkashem: It might, depending on the pod configuration. We could also decide a static set of `initializers` some of which could be noop. Today there is only one for env.
Ideally we should have a layer that returns a chain of deployment `initializer` func based on the context. What we have today is a quick fix, we can improve the strategy resolver and deployment `initializer`. I have a task that tracks it.

pkg/controller/operators/olm/envvar/inject.go
```diff
@@ -0,0 +1,61 @@
+package envvar
```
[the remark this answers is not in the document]
tkashem: I thought the `envvar` package was specific to `olm` only. I have a `proxy` package that deals with openshift `proxy` type that is in `pkg/lib`.

pkg/lib/proxy/syncer.go
ecordell: Yep, that sounds right

pkg/controller/operators/olm/operator.go
tkashem: It should not, `err` is set only when there is a connectivity error. check - https://github.com/operator-framework/operator-lifecycle-manager/pull/950/commits/35a159ed8e1b1d7931a09f93a68a7c74c90540ff#diff-eb26d569f984271521cd371191e0b72bR30

pkg/controller/operators/olm/envvar/inject.go
tkashem: yes, I agree on allowing the user to overwrite a "set" env var to ""
proxy env is an exception, if they are empty they are not injected. and if the use specifies any in subscription pod ( empty or not ) it overrides the proxy, we won't even query `cluster` proxy at that point.

tkashem: Given this, we need to treat the proxy env differently, I will add that change ( maybe in the followup PR I am planning if I don't have enough time )

pkg/controller/operators/olm/operator.go
tkashem: yeah, in `monitor` we wait indefinitely since it'a an independently running go routine. For the operator we can retry - is 6 tries, every 10-second and then giving up a good default?

ecordell:
/lgtm

None of my comments above should block :)

pkg/controller/operators/olm/operator.go
ecordell: Seems reasonable to me - we probably want to go Degraded if this fails (and we're on openshift), but I think that can be a followup in the future.

PR #951, opened 2019-07-11
feat(metrics): record sync count for Subscriptions, labeled with name and installedCSV

ecordell:
This increments a `count` whenever we sync a CSV, labelled with the name and phase of the CSV. This should let us see which operators are being installed and what their lifecycle generally looks like.

Open to suggestions for better ways to report; the count seemed like the simplest thing. If OLM should be syncing one of them long enough to max the count, it should wrap back around.

Example queries:

**Sync counts by name**
```
sum(subscription_sync_total) by (name, installed)
```

Element | Value
-- | --
{installed="postgresoperator.v4.0.0-rc10",name="crunchy-postgres-operator"} | 4
{name="crunchy-postgres-operator"} | 7

(in grafana you can do `label_values(subscription_sync_total, name)` to get just a list of names)

### DIFF
--- a/pkg/controller/operators/olm/operator.go
+++ b/pkg/controller/operators/olm/operator.go
@@ -7,9 +7,9 @@ import (
 	"strings"
 	"time"
 
-	log "github.com/sirupsen/logrus"
 	v1 "github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1"
 	"github.com/sirupsen/logrus"
+	log "github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
 	extinf "k8s.io/apiextensions-apiserver/pkg/client/informers/externalversions"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
@@ -678,6 +678,8 @@ func (a *Operator) syncClusterServiceVersion(obj interface{}) (syncError error) 
 		return
 	}
 
+	a.recordMetrics(logger, clusterServiceVersion)
+
 	outCSV, syncError := a.transitionCSVState(*clusterServiceVersion)
 
 	if outCSV == nil {
@@ -775,6 +777,18 @@ func (a *Operator) syncGcCsv(obj interface{}) (syncError error) {
 		return
 	}
 	return
+}
+
+func (a *Operator) recordMetrics(logger *logrus.Entry, csv *v1alpha1.ClusterServiceVersion) {
+	if csv.IsCopied() {
+		// don't record for copied CSVs
+		return
+	}
+	counter, err := metrics.CounterForCSV(csv.GetName(), string(csv.Status.Phase))
+	if err != nil {
+		logger.WithError(err).Warn("could not record metrics")
+	}
+	counter.Inc()
 }
 
 // operatorGroupFromAnnotations returns the OperatorGroup for the CSV only if the CSV is active one in the group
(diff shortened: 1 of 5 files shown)


ecordell:
/hold 

I want to test in a real cluster first, but otherwise fine to review

pkg/controller/operators/olm/operator.go
```diff
@@ -777,6 +779,18 @@ func (a *Operator) syncGcCsv(obj interface{}) (syncError error) {
 	return
 }
 
+func (a *Operator) recordMetrics(logger *logrus.Entry, csv *v1alpha1.ClusterServiceVersion) {
+	if csv.IsCopied() {
+		// don't record for copied CSVs
+		return
+	}
+	counter, err := metrics.CounterForCSV(csv.GetName(), string(csv.Status.Phase))
```
jwforres: isn't the CSV name unique for every version of the operator, that will make it hard for us to roll up counts that include all versions of an operator

i think what i would want would be the unique identifier for the operator as one label, the version is another, and then the phase

ecordell: We can report `Subscriptions` as well, labelled by package, which would give the identifier for which operators are installed sans version?

pkg/metrics/metrics.go
```diff
@@ -132,15 +137,46 @@ var (
...
 
 func RegisterCatalog() {
 	prometheus.MustRegister(installPlanCount)
 	prometheus.MustRegister(subscriptionCount)
 	prometheus.MustRegister(catalogSourceCount)
+	prometheus.MustRegister(SubscriptionSyncCount)
+}
+
+func CounterForCSV(name, phase string) (prometheus.Counter, error) {
+	return CSVSyncCount.GetMetricWith(map[string]string{
```
brancz: I personally find `.WithLabelValues` more ergonomic to use for this https://github.com/prometheus/client_golang/blob/ff1d4e21c12e4c0b190a166d96507786c8f989d6/prometheus/counter.go#L203

pkg/controller/operators/olm/operator.go
brancz: I think that might be more aligned with our use of these metrics.

pkg/metrics/metrics.go
ecordell: I don't really want OLM to panic if this fails though, right?

pkg/controller/operators/olm/operator.go
jwforres: so with subscriptions we don't know the currently installed version on that object do we, only sometimes the starting version? Is there a way (in followup work later) for us to report a status back to the subscription that says what is currently installed

not that we'd try to do that as part of this PR, just trying to brainstorm how we get to the end goal

pkg/metrics/metrics.go
brancz: if you pass the wrong labels your instrumentation is still going to be wrong, but it's a style choice, you're free to choose yours :) we prefer `.WithLabelValues` generally

pkg/controller/operators/olm/operator.go
ecordell: Subscriptions report the `currentCSV` in their status, but that doesn't give you information about the health of the CSV. 

I can switch to _only_ reporting metrics for subscriptions, with `name` and `currentCSV` labels?

pkg/controller/operators/catalog/operator.go
```diff
@@ -697,11 +697,16 @@ func (o *Operator) syncSubscriptions(obj interface{}) error {
 		return fmt.Errorf("casting Subscription failed")
 	}
 
+	o.recordMetrics(sub)
```
njhale: You may want to record metrics from the [subscriptionSyncer's `Sync` function](https://github.com/operator-framework/operator-lifecycle-manager/blob/e9ac62be99f467ccf201bd2af73cacea91ec714c/pkg/controller/operators/catalog/subscription/syncer.go#L46) since this sync function will only be called _after_ all others have successfully processed (i.e. you'll only get a fraction of total sync events).

pkg/controller/operators/olm/operator.go
jeremyeder: Just leaving this here for future reference; "we have a cluster_operator_conditions metrics that tells us if any CVO components are degraded.  is there an equivalent metric for tracking non CVO operators? something planned for OLM?"

PR #955, opened 2019-07-18
feat(deploy): add resource requests to all pods

jpeeler:
CPU has been set to 10m for all pods. The following are the memory requests:
olm-operator     160Mi
catalog-operator 80Mi
packageserver    50Mi
catalog sources  50Mi

This is for https://bugzilla.redhat.com/show_bug.cgi?id=1711070 (OLM-1130).


jpeeler:
The unit test failure here is real, unfortunately.

PR #957, opened 2019-07-19
fix(manifest): fix env var projection in command

ecordell:
the wrong brackets was causing the "cleanup" namespace to be set to ""
instead of the namespace where olm is deployed. this caused upgrade
failures when old artifacts weren't properly cleaned on startup.

### DIFF
--- a/deploy/chart/templates/0000_50_olm_07-olm-operator.deployment.yaml
+++ b/deploy/chart/templates/0000_50_olm_07-olm-operator.deployment.yaml
@@ -27,7 +27,7 @@ spec:
           - /bin/olm
           args:
           - -namespace
-          - ${OPERATOR_NAMESPACE}
+          - $(OPERATOR_NAMESPACE)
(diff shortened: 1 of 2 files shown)


deploy/chart/templates/0000_50_olm_07-olm-operator.deployment.yaml
```diff
@@ -27,7 +27,7 @@ spec:
           - /bin/olm
           args:
           - -namespace
-          - ${OPERATOR_NAMESPACE}
+          - $(OPERATOR_NAMESPACE)
```
njhale:
```suggestion
          - "$(OPERATOR_NAMESPACE)"
```

The [1.14 docs](https://v1-14.docs.kubernetes.io/docs/tasks/inject-data-application/define-environment-variable-container/#using-environment-variables-inside-of-your-config) wrap the expansion with quotes. Not sure if it makes a difference though - waiting on CI.

PR #958, opened 2019-07-22
Bug 1732613: Follow up for pod configuration

tkashem:
* When a subscription changes, if it has an installed CSV then requeue
it so that any changes in pod configuration can take effect
immediately.

* Empty proxy env var should result in unset  …
*  If an env variable specified in pod configuration is empty, it will
  override the target PodSpec. For example, if 'foo=""' in pod
  configuration, olm will reset the 'foo' env variable in the target
  PodSpec to 'foo=""'
*  However, proxy env var is an exception to the above. An empty proxy
  env var means that it should not be set or it should be unset if it
  is already set in the target PodSpec.

* Rename and remove commented code

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1732613

### DIFF
--- a/pkg/controller/operators/olm/envvar/initializer.go
+++ b/pkg/controller/operators/olm/envvar/initializer.go
@@ -63,7 +63,7 @@ func (d *DeploymentInitializer) initialize(ownerCSV ownerutil.Owner, deployment 
 	merged = append(podConfigEnvVar, proxyEnvVar...)
 
 	if len(merged) == 0 {
-		d.logger.Debugf("no env var to inject into csv=%s", ownerCSV.GetName())
+		d.logger.WithField("csv", ownerCSV.GetName()).Debug("no env var to inject into csv")
 	}
 
 	podSpec := deployment.Spec.Template.Spec
--- a/pkg/controller/operators/olm/envvar/initializer.go
+++ b/pkg/controller/operators/olm/envvar/initializer.go
@@ -56,6 +56,8 @@ func (d *DeploymentInitializer) initialize(ownerCSV ownerutil.Owner, deployment 
 			err = fmt.Errorf("failed to query cluster proxy configuration - %v", err)
 			return err
 		}
+
+		proxyEnvVar = dropEmptyProxyEnv(proxyEnvVar)
 	}
 
 	merged = append(podConfigEnvVar, proxyEnvVar...)
@@ -70,4 +72,17 @@ func (d *DeploymentInitializer) initialize(ownerCSV ownerutil.Owner, deployment 
 	}
 
 	return nil
+}
+
+func dropEmptyProxyEnv(in []corev1.EnvVar) (out []corev1.EnvVar) {
+	out = make([]corev1.EnvVar, 0)
+	for i := range in {
+		if in[i].Value == "" {
+			continue
+		}
+
+		out = append(out, in[i])
+	}
+
+	return
 }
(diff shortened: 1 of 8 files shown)


pkg/controller/operators/olm/envvar/initializer.go
```diff
@@ -58,14 +59,19 @@ func (d *DeploymentInitializer) initialize(ownerCSV ownerutil.Owner, deployment
 		}
 	}
 
-	merged = append(podConfigEnvVar, proxyEnvVar...)
+	// In pod configuration, an empty env var will allow us to set target to empty
+	// On the other hand, for proxy env vars, empty means that the env variable
+	// will not be set.
+	podConfigEnv := ToOverrideList(false, podConfigEnvVar...)
+	proxyEnv := ToOverrideList(true, proxyEnvVar...)
+	overrides = append(podConfigEnv, proxyEnv...)
```
ecordell: Can we safely remove the `Override` abstraction? along the lines of:

```go
func toMap(env []corev1.EnvVar) map[string]corev1.EnvVar {
	envMap := map[string]corev1.EnvVar{}
	for i := range env {
		e := env[i]
		envMap[e.Name] = e
	}
	return envMap
}

func overrideIgnoreEmpty(proxyEnv map[string]corev1.EnvVar, []corev1.EnvVar) {
  for _, e := proxyConfigEnv {
    if e.Value != "" {
        proxyEnv[e.Name] = e
     }
  }
}

merged := overrideIgnoreEmpty(toMap(proxyEnv), proxyConfigEnv)
```

tkashem: I was thinking of the scenario where these proxy env variables are already set on the deployment `PodSpec`. If we remove them ( if they are empty ) here, we will not have enough context to unset them from the target `PodSpec`. But I assume we will never have that scenario, correct? Our target `PodSpec` is always generated/calculated and never from an existing `PodSpec`. Also, we won't have users specif proxy env vars in `csv`, I assume. 

I can go back to what we had before and simply drop the empty proxy env vars here.

ecordell: > Our target PodSpec is always generated/calculated and never from an existing PodSpec

Yes, we're always calculating what the pod spec should be and then checking to ensure the on-cluster one matches.

> Also, we won't have users specify proxy env vars in csv, I assume.

It seems unlikely - but if so I would think that anything in the config would overwrite them?

tkashem: > but if so I would think that anything in the config would overwrite them?

Yes, anything in the subscription pod config will overwrite them. But not in the case where subscription pod config is `nil` and the `cluster` `Proxy` object has empty proxy env vars, then if we drop them here we lose this contextual information. 
I am assuming the order of precedence is, ( from high to low ) subscription pod config -> `cluster` proxy env var(s) -> `csv` spec.

To summarize, if we drop `UnsetIfEmpty` ( this is a package level abstraction not at the OLM API level ), this is what we have:
- Empty value in subscription pod `config` will overwrite target.
- We drop empty env var from `cluster` proxy object.
- We expect that the target `PodSpec` will never have proxy variables set.

If this sounds good, I will go ahead and remove  `UnsetIfEmpty`

ecordell: That sounds good to me

PR #959, opened 2019-07-22
Bug 1732302: Fix panic when binding already exists

cblecker:
Fix the following panic when the RoleBinding/ClusterRoleBinding already exists.

Note this will need to be backported, as it exists in the release-4.x branches as well.

```
panic: assignment to entry in nil map

goroutine 183 [running]:
github.com/operator-framework/operator-lifecycle-manager/pkg/controller/operators/catalog.(*Operator).ExecutePlan(0xc4204bec00, 0xc421924000, 0x1, 0x1)
        /go/src/github.com/operator-framework/operator-lifecycle-manager/pkg/controller/operators/catalog/operator.go:1187 +0x4243
```

PR #964, opened 2019-07-24
Bug 1732214: Fix panic when binding already exists

ecordell:
This is a manual cherry-pick of https://github.com/operator-framework/operator-lifecycle-manager/pull/959

@cblecker

### DIFF
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -1160,6 +1160,9 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
 				_, err = o.OpClient.KubernetesInterface().RbacV1().ClusterRoles().Create(&cr)
 				if k8serrors.IsAlreadyExists(err) {
 					// if we're updating, point owner to the newest csv
+					if cr.ObjectMeta.Labels == nil {
+						cr.ObjectMeta.Labels = map[string]string{}
+					}
 					cr.Labels[ownerutil.OwnerKey] = step.Resolving
 					_, err = o.OpClient.UpdateClusterRole(&cr)
 					if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/controller/operators/catalog/operator.go
```diff
@@ -1160,6 +1160,9 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
 				_, err = o.OpClient.KubernetesInterface().RbacV1().ClusterRoles().Create(&cr)
 				if k8serrors.IsAlreadyExists(err) {
 					// if we're updating, point owner to the newest csv
+					if cr.ObjectMeta.Labels == nil {
+						cr.ObjectMeta.Labels = map[string]string{}
+					}
 					cr.Labels[ownerutil.OwnerKey] = step.Resolving
```
cblecker: This line also needs to be changed to:
```suggestion
					cr.ObjectMeta.Labels[ownerutil.OwnerKey] = step.Resolving
```

pkg/controller/operators/catalog/operator.go
```diff
@@ -1184,6 +1187,9 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
 				// Attempt to create the ClusterRoleBinding.
 				_, err = o.OpClient.KubernetesInterface().RbacV1().ClusterRoleBindings().Create(&rb)
 				if k8serrors.IsAlreadyExists(err) {
+					if rb.ObjectMeta.Labels == nil {
+						rb.ObjectMeta.Labels = map[string]string{}
+					}
 					rb.Labels[ownerutil.OwnerKey] = step.Resolving
```
cblecker:
```suggestion
					rb.ObjectMeta.Labels[ownerutil.OwnerKey] = step.Resolving
```

pkg/controller/operators/catalog/operator.go
ecordell: These are the same thing (`ObjectMeta` is embedded in the `ClusterRole`) but I will commit this change so that it matches the other PR.

cblecker:
/lgtm

@ecordell I think there was two BZs opened.. the original (https://bugzilla.redhat.com/show_bug.cgi?id=1732214) and the one @jianzhangbjz opened (https://bugzilla.redhat.com/show_bug.cgi?id=1732302). I don't know if that makes a difference in resolving the invalid-bug label

eparis:
@cblecker FYI Evan's got the bugs and PRs all set up perfectly now. We have a bug with target release 4.2 (1732302) which 'blocks' a bug with target release 4.1 (1733324).   And now the bot is telling us that 1733324 (4.2) is not yet verified by QE.  Once that bug goes VERIFIED we should run `/bugzilla refresh` again.

Thank you both!

PR #969, opened 2019-07-26
(doc) add doc for scoped operator install

tkashem:


tkashem:
> Generally looks good. A couple of thoughts:
> 
> * change the perspective to that of the user and refrain from using internal implementation, e.g.
>   
>   * OLM controllers run with cluster-admin privileges vs. OLM runs with cluster-admin privileges
>   * OLM uses a client bound to a serviceaccount vs. the Operator can request any permissions
> * use consistent nomenclature, e.g. `serviceaccount` vs. `ServiceAccount` and `olm` vs. `OLM`
> * use of colloquial language: "OLM does not spit out..."

@dmesser addressed your feedback, please take a look when you have time.

Documentation/design/scoped-operator-install.md
```diff
@@ -0,0 +1,198 @@
+## Background
+OLM runs with `cluster-admin` privileges. By default, operator author can specify any set of permission(s) in the `csv` and OLM will consequently grant it to the operator. In effect, an operator can achieve `cluster-scoped` privilege(s) which may not always be desired.
```
dmesser:
```suggestion
OLM runs with `cluster-admin` privileges. By default, operator author can specify any set of permission(s) in the `CSV` and OLM will consequently grant it to the operator. In effect, an operator can achieve `cluster-scoped` privilege(s) which may not always be desired.
```

Documentation/design/scoped-operator-install.md
```diff
@@ -0,0 +1,198 @@
...
+
+OLM introduced a new feature that applies principle of attenuation to the operator being installed. The administrator can specify a service account with a set of privilege(s) granted to it. OLM will ensure that when an operator is installed its privileges are confined to that of the service account specified.
```
dmesser:
```suggestion
OLM introduced a new feature that applies principle of attenuation to the operator being installed. The administrator can specify a service account with a set of privilege(s) granted to it. OLM will ensure that when an operator is installed its privileges are confined to that of the service account specified.

As a result a `cluster-admin` can limit an Operator to a pre-defined set of RBAC rules. The Operator will not be able to do anything that is not explicitly permitted by those. This enables self-sufficient installation of Operators by non-`cluster-admin` users with a limited scope.
```

PR #970, opened 2019-07-29
Bug 1743345: clean up service account, cluster roles, and cluster role bindings after CSV deletion

jpeeler:
The approach I have here works, but takes a while as the resync interval has to pass. I was trying avoid creating a new queue for requeueing, though that may not have worked anyway since really it'd need to be requeued after CSV deletion. The commented out section did not work due to CSV requeueing and in progress CSV processing.

(The above is now outdated.)

### DIFF
--- a/pkg/controller/operators/olm/operator.go
+++ b/pkg/controller/operators/olm/operator.go
@@ -8,9 +8,11 @@ import (
 	"time"
 
 	v1 "github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/kubestate"
 	"github.com/sirupsen/logrus"
 	log "github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
 	extinf "k8s.io/apiextensions-apiserver/pkg/client/informers/externalversions"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -32,6 +34,7 @@ import (
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/informers/externalversions"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/certs"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/install"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/operators/olm/envvar"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/registry/resolver"
 	csvutility "github.com/operator-framework/operator-lifecycle-manager/pkg/lib/csv"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/event"
@@ -40,10 +43,9 @@ import (
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/operatorclient"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/operatorlister"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/ownerutil"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/proxy"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/queueinformer"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/scoped"
-	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/proxy"
-	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/operators/olm/envvar"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/metrics"
 )
 
@@ -65,6 +67,7 @@ type Operator struct {
 	csvQueueSet           *queueinformer.ResourceQueueSet
 	csvCopyQueueSet       *queueinformer.ResourceQueueSet
 	csvGCQueueSet         *queueinformer.ResourceQueueSet
+	objGCQueueSet         *queueinformer.ResourceQueueSet
 	apiServiceQueue       workqueue.RateLimitingInterface
 	csvIndexers           map[string]cache.Indexer
 	recorder              record.EventRecorder
@@ -118,6 +121,7 @@ func newOperatorWithConfig(ctx context.Context, config *operatorConfig) (*Operat
 		csvQueueSet:           queueinformer.NewEmptyResourceQueueSet(),
 		csvCopyQueueSet:       queueinformer.NewEmptyResourceQueueSet(),
 		csvGCQueueSet:         queueinformer.NewEmptyResourceQueueSet(),
+		objGCQueueSet:         queueinformer.NewEmptyResourceQueueSet(),
 		apiServiceQueue:       workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "apiservice"),
 		resolver:              config.strategyResolver,
 		apiReconciler:         config.apiReconciler,
@@ -215,7 +219,7 @@ func newOperatorWithConfig(ctx context.Context, config *operatorConfig) (*Operat
 		}
 
 		subInformer := extInformerFactory.Operators().V1alpha1().Subscriptions()
-		op.lister.OperatorsV1alpha1().RegisterSubscriptionLister(namespace, subInformer.Lister())		
+		op.lister.OperatorsV1alpha1().RegisterSubscriptionLister(namespace, subInformer.Lister())
 		subQueueInformer, err := queueinformer.NewQueueInformer(
 			ctx,
 			queueinformer.WithLogger(op.logger),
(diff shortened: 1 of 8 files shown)


pkg/controller/registry/resolver/rbac.go
```diff
@@ -116,7 +116,11 @@ func RBACForClusterServiceVersion(csv *v1alpha1.ClusterServiceVersion) (map[stri
 	for _, permission := range strategyDetailsDeployment.ClusterPermissions {
 		// Create ServiceAccount if necessary
 		if _, ok := permissions[permission.ServiceAccountName]; !ok {
-			serviceAccount := &corev1.ServiceAccount{}
+			serviceAccount := &corev1.ServiceAccount{
```
jpeeler: Can remove these labels if they don't end up getting used.

pkg/controller/operators/olm/operator.go
```diff
@@ -550,14 +567,66 @@ func (a *Operator) syncObject(obj interface{}) (syncError error) {
...
+		return
 	}
+	logger := a.logger.WithFields(logrus.Fields{
+		"name":      metaObj.GetName(),
+		"namespace": metaObj.GetNamespace(),
+		"self":      metaObj.GetSelfLink(),
+	})
 
 	// Requeues objects that can't have ownerrefs (cluster -> namespace, cross-namespace)
 	if ownerutil.IsOwnedByKindLabel(metaObj, v1alpha1.ClusterServiceVersionKind) {
```
njhale: There seems to be some duplicated work. I wonder if we can rearrange things to make it a little cleaner. Does it make sense to move some of this to the `requeueOwnerCSVs` method; particularly the existence check and GC enqueuing?

pkg/lib/queueinformer/queueinformer.go
```diff
@@ -105,6 +105,13 @@ func (q *QueueInformer) metricHandlers() *cache.ResourceEventHandlerFuncs {
 	}
 }
 
+func NewQueue(ctx context.Context, options ...Option) (*QueueInformer, error) {
```
njhale: Having a different constructor is fine, but it doesn't look like this is doing anything different from `NewQueueInformer(...)`. If the goal is to have a QueueInformer that doesn't have an informer behind it, then I think what you want to do is remove the [non-nil informer/indexer test](https://github.com/operator-framework/operator-lifecycle-manager/blob/86bac67214a58e64a3000f6aae50919380c3990a/pkg/lib/queueinformer/config.go#L54) in the `queueInformerConfig.validate()` method, or create a new validate method that ignores that check. Additionally, you would need to detect a [nil indexer](https://github.com/operator-framework/operator-lifecycle-manager/blob/86bac67214a58e64a3000f6aae50919380c3990a/pkg/lib/queueinformer/queueinformer_operator.go#L251) and attempt to use the resource nested in the event, rather than getting it from the indexer.

pkg/controller/operators/olm/operator.go
```diff
@@ -319,6 +322,21 @@ func newOperatorWithConfig(ctx context.Context, config *operatorConfig) (*Operat
 		if err := op.RegisterQueueInformer(serviceAccountQueueInformer); err != nil {
 			return nil, err
 		}
+
+		objGCQueue := workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), fmt.Sprintf("%s/obj-gc", namespace))
+		objGCQueueInformer, err := queueinformer.NewQueue(
+			ctx,
+			queueinformer.WithLogger(op.logger),
+			queueinformer.WithQueue(objGCQueue),
+			queueinformer.WithSyncer(queueinformer.LegacySyncHandler(op.syncGCObject).ToSyncer()),
```
njhale: Is using the legacy adapter more convenient than implementing a new `Syncer`? If that's the case then we may want to remove that interface going forward.

pkg/controller/operators/olm/operator.go
```diff
@@ -550,14 +567,66 @@ func (a *Operator) syncObject(obj interface{}) (syncError error) {
...
+			logger.Error("unexpected owner label retrieval failure")
+		}
+		_, err := a.lister.OperatorsV1alpha1().ClusterServiceVersionLister().ClusterServiceVersions(ns).Get(name)
+		if !k8serrors.IsNotFound(err) {
+			logger.Debug("requeueing owner csvs from owner label")
+			a.requeueOwnerCSVs(metaObj)
+		} else {
+			switch metaObj.(type) {
+			case *corev1.ServiceAccount, *rbacv1.ClusterRole, *rbacv1.ClusterRoleBinding:
+				a.objGCQueueSet.Requeue(ns, metaObj.GetName())
```
njhale: If the goal is to send a `ResourceEvent`, this will not work as intended since it constructs a key manually. It may be helpful to add a `RequeueEvent` method to the `ResourceQueue` type.

pkg/controller/operators/olm/operator.go
jpeeler: Since I'm familiar with the "old" way, I was going to implement that first.

pkg/controller/operators/olm/operator.go
jpeeler: Most of the code here was already present. The goal was to avoid requeueing CSVs in a deletion scenario. Maybe I can circle back to this later.

pkg/lib/queueinformer/queueinformer.go
jpeeler: I ended up creating a ResourceEvent (with type deleted) and putting that on the queue as you suggested out of band.

pkg/lib/queueinformer/resourcequeue.go
```diff
@@ -34,6 +34,18 @@ func (r *ResourceQueueSet) Set(key string, queue workqueue.RateLimitingInterface
 	r.queueSet[key] = queue
 }
 
+func (r *ResourceQueueSet) RequeueEvent(namespace string, resourceEvent kubestate.ResourceEvent) error {
+	r.mutex.RLock()
+	defer r.mutex.RUnlock()
+
+	if queue, ok := r.queueSet[namespace]; ok {
+		queue.AddRateLimited(resourceEvent)
```
jpeeler: Really what I wanted to do here is `queue.AddAfter(resourceEvent, 10 * time.Second)`, but I think what is here works okay at least. I figured a timeout wouldn't be allowed, but it would avoid some potential requeueing due to lingering CSVs.

pkg/controller/operators/olm/operator.go
```diff
@@ -319,6 +323,38 @@ func newOperatorWithConfig(ctx context.Context, config *operatorConfig) (*Operat
...
+			return nil, err
+		}
+		if err := op.RegisterQueueInformer(objGCQueueInformer); err != nil {
+			return nil, err
+		}
+
+	}
+
+	// add queue for all namespaces as well
+	objGCQueue := workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), fmt.Sprintf("%s/obj-gc", ""))
```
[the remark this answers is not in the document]
jpeeler: Possibly, but it does support namespaced resources too (originally, I was using it on service accounts too).

pkg/controller/operators/olm/operator.go
```diff
@@ -319,6 +323,38 @@ func newOperatorWithConfig(ctx context.Context, config *operatorConfig) (*Operat
 		if err := op.RegisterQueueInformer(serviceAccountQueueInformer); err != nil {
 			return nil, err
 		}
+
+		objGCQueue := workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), fmt.Sprintf("%s/obj-gc", namespace))
```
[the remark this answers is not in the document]
jpeeler: Not right now, but maybe later? I figured there wasn't much overhead in a queue that isn't used.

pkg/lib/queueinformer/config.go
```diff
@@ -43,15 +43,15 @@ func (c *queueInformerConfig) complete() {
 }
 
 // validate returns an error if the config isn't valid.
-func (c *queueInformerConfig) validate() (err error) {
+func (c *queueInformerConfig) validate(skipIndexerValidation bool) (err error) {
```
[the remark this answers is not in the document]
jpeeler: That's a good suggestion.

pkg/controller/registry/resolver/rbac.go
```diff
@@ -117,6 +117,7 @@ func RBACForClusterServiceVersion(csv *v1alpha1.ClusterServiceVersion) (map[stri
 		// Create ServiceAccount if necessary
 		if _, ok := permissions[permission.ServiceAccountName]; !ok {
 			serviceAccount := &corev1.ServiceAccount{}
+			ownerutil.AddOwner(serviceAccount, csv, false, false)
```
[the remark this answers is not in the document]
jpeeler: I see now that this differs from https://github.com/operator-framework/operator-lifecycle-manager/pull/970/files#diff-64691a309491514cf415b817a09535c5L77, but does that mean both instances are in the wrong place?

shawn-hurley:
Please add a bug to the title that this is fixing

pkg/controller/registry/resolver/rbac.go
jpeeler: Ended up handling this at the owner reference level instead of with owner labels.

pkg/lib/queueinformer/queueinformer_operator.go
```diff
@@ -247,17 +247,24 @@ func (o *operator) processNextWorkItem(ctx context.Context, loop *QueueInformer)
...
-			logger.WithError(err).Error("cache get failed")
-			queue.Forget(item)
-			return true
-		}
-		if !exists {
-			logger.WithField("existing-cache-keys", loop.indexer.ListKeys()).Debug("cache get failed, key not in cache")
-			queue.Forget(item)
-			return true
+		var resource interface{}
+		if loop.indexer == nil {
```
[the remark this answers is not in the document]
jpeeler: I wasn't able to.

pkg/controller/registry/resolver/rbac.go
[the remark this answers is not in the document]
ecordell: Ah - this is fine. They get added here, but updated by ExecutePlan before being applied to the cluster. See: https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/controller/operators/catalog/operator.go#L1418-L1424

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -253,7 +253,7 @@ func (a *Operator) ensureProvidedAPIClusterRole(operatorGroup *v1.OperatorGroup,
 		},
 		Rules: []rbacv1.PolicyRule{{Verbs: verbs, APIGroups: []string{group}, Resources: []string{resource}, ResourceNames: resourceNames}},
 	}
-	err := ownerutil.AddOwnerLabels(clusterRole, operatorGroup)
```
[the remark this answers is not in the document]
jpeeler: Thanks!

PR #971, opened 2019-07-30
Bug 1727032 Report `Upgradeable` in ClusterOperator status

kevinrizza:
Update status reporting for Openshift `ClusterOperator` status to report
the `Upgradeable` condition in the following context:
* When upgrading (`Progressing == true` and `Available == false`) set
`Upgradeable` to `false`
* Once upgrade is complete, set `Upgradeable` to `true`

This PR resolves the following bugs:
https://bugzilla.redhat.com/show_bug.cgi?id=1727032
https://bugzilla.redhat.com/show_bug.cgi?id=1727042


ecordell:
/retest
/lgtm

long term, I think we want to be able to signal not-upgradeable based on the state of a set of installed operators

PR #973, opened 2019-08-01
Bug 1732914: Operator upgrades fail when versions field is not set

dinhxuanvu:
1. Fix the ensureCRDVersions to account for `version` field
2. Add test cases for `version` scenarios

### DIFF
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -1064,23 +1064,35 @@ func (o *Operator) ResolvePlan(plan *v1alpha1.InstallPlan) error {
 	return nil
 }
 
-// Ensure all existing versions are present in new CRD
+// Ensure all existing served versions are present in new CRD
 func ensureCRDVersions(oldCRD *v1beta1ext.CustomResourceDefinition, newCRD *v1beta1ext.CustomResourceDefinition) error {
-	for _, oldVersion := range oldCRD.Spec.Versions {
-		var versionPresent bool
-		for _, newVersion := range newCRD.Spec.Versions {
-			if oldVersion.Name == newVersion.Name {
-				versionPresent = true
-			}
-		}
-		if !versionPresent {
-			return fmt.Errorf("not allowing CRD (%v) update with unincluded version %v", newCRD.GetName(), oldVersion)
-		}
+	newCRDVersions := map[string]struct{}{}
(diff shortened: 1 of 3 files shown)


pkg/controller/operators/catalog/operator.go
```diff
@@ -1066,15 +1066,25 @@ func (o *Operator) ResolvePlan(plan *v1alpha1.InstallPlan) error {
 
 // Ensure all existing versions are present in new CRD
 func ensureCRDVersions(oldCRD *v1beta1ext.CustomResourceDefinition, newCRD *v1beta1ext.CustomResourceDefinition) error {
-	for _, oldVersion := range oldCRD.Spec.Versions {
-		var versionPresent bool
-		for _, newVersion := range newCRD.Spec.Versions {
-			if oldVersion.Name == newVersion.Name {
-				versionPresent = true
-			}
+	newCRDVersions := map[string]struct{}{}
```
ecordell: Is the behavior that we want here actually something like:

1. There must be a nonempty intersection of oldCRDVersions and newCRDVersions
2. For any version in both oldCRD and newCRD, ensure the validation is unchanged (or has only backwards-compatible changes)

I'm thinking that this used as-is will prevent anyone from ever deprecating and older api version. What do you think?

dinhxuanvu: Developers can deprecate/disable a version of CRD by simply changing `Served` field into `false`. The reason I think we shouldn't allow developers to remove a version straight out without at least disabling it first to avoid the confusion that a CSV may reference that version or an existing CR with that version but the new CRD has no record of that version. Also, if a CSV is still referencing to a version that is not present in new CRD, the `InstallPlan` won't include a step to upgrade that CRD.

rthallisey:
will this change be back ported to 4.2?

dinhxuanvu:
@rthallisey This should be 4.2 bug fix so it should be in 4.2.

pkg/controller/operators/catalog/operator.go
shawn-hurley: I would like to make sure that this is documented. Can we add some doc's in this PR?

dinhxuanvu: @shawn-hurley The doc PR is referenced in this PR. It is #984.

PR #974, opened 2019-08-01
Bug 1737081: catsrc status should show error 

tkashem:
Currently Status of CatalogSource does not show any information to the
user on current state(s). For example, if the registry pod backed by
the CatalogSource is crash looping, the status shows no information.

Add the following fields to Status:
- Message: It will display the error last observed.
- Reason: It will display the reason, for now we have two reasons:
    - ConfigMapError
    - RegistryServerError

Change the sync handler to set the error to status. This is a stop gap measure, for the longer term, we would like to have `Conditions` that comply to kubernetes conventions.

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1737081
Jira: https://jira.coreos.com/browse/OLM-929


cmd/olm/cleanup.go
```diff
@@ -37,6 +37,10 @@ func cleanup(logger *logrus.Logger, c operatorclient.ClientInterface, crc versio
 		logger.WithError(err).Fatal("couldn't clean previous release")
 	}
 
+	if err := waitForDelete(checkClusterServiceVersion(crc, "packageserver.v0.10.1"), deleteClusterServiceVersion(crc, "packageserver.v0.10.0")); err != nil {
```
jpeeler: Nit: would be nicer to loop over the package server version names here.

pkg/api/apis/operators/v1alpha1/catalogsource_types.go
```diff
@@ -73,18 +78,35 @@ func (s *RegistryServiceStatus) Address() string {
...
+	Name            string      `json:"name"`
+	Namespace       string      `json:"namespace"`
+	UID             types.UID   `json:"uid,omitempty"`
+	ResourceVersion string      `json:"resourceVersion,omitempty"`
+	LastUpdateTime  metav1.Time `json:"lastUpdateTime,omitempty"`
+}
 
-	UID             types.UID `json:"uid,omitempty"`
-	ResourceVersion string    `json:"resourceVersion,omitempty"`
+func (r *ConfigMapResourceReference) IsAMatch(object *metav1.ObjectMeta) bool {
```
jpeeler: This might be better placed in the internal versions file.

PR #976, opened 2019-08-02
Bug 1737156: feat(metrics): record sync count for Subscriptions, labeled with name and installedCSV

ecordell:
cherry pick of #951

I had to move `recordMetrics` from the syncer to the operator, sync the syncer does not exist in 4.1


tkashem:
/lgtm

Question: if I want to run this locally on minikube and want to see the metrics on a dashboard, I assume it is possible?

PR #977, opened 2019-08-05
Bug 1740174: fix(catalog): re-install resources in existing installplan

jpeeler:
This specifically fixes the scenario where a subscription is deleted and recreated in a namespace with an additional subscription. When the subscription of interest is deleted, the associated installplan remains due to the other subscription's owner reference. Because the existing installplan manifests match, no additional work is done. This is a problem if resources that were originally installed have been deleted.

Now after a subscription is recreated, the proper owner references are re-added onto the install plan as well as all the resources are checked that they are installed.

(Cherry pick of https://github.com/operator-framework/operator-lifecycle-manager/pull/965.)

### DIFF
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -844,6 +844,26 @@ func (o *Operator) ensureInstallPlan(logger *logrus.Entry, namespace string, sub
 	for _, installPlan := range installPlans {
 		if installPlan.Status.CSVManifestsMatch(steps) {
 			logger.Infof("found InstallPlan with matching manifests: %s", installPlan.GetName())
+
+			ownerWasAdded := false
+			for _, sub := range subs {
+				ownerWasAdded = ownerWasAdded || !ownerutil.EnsureOwner(installPlan, sub)
+			}
+			if ownerWasAdded {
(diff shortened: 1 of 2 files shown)


crawford:
I don't see an associated BZ for the [parent PR](https://github.com/operator-framework/operator-lifecycle-manager/pull/965). That needs to be created and then verified by QE before I can merge this.

ecordell:
@crawford The original BZ was filed only against 4.1, should we clone the bug so there's one for 4.1 and 4.2?

crawford:
@ecordell You need a bug filed against 4.1.z and 4.2. Typically, the bug against 4.2 gets filed first and then the 4.1.z bug is cloned from that. Once QE has verified the 4.2 bug on the master branch, we can cherry-pick it into the release branch.

ecordell:
@jpeeler we're waiting for QE to verify the parent bug against 4.2, then this will be ready to merge.

jpeeler:
QE has verified the 4.2 fix now (https://bugzilla.redhat.com/show_bug.cgi?id=1714140).

jpeeler:
@crawford should be good to go now

jpeeler:
@derekwaynecarr I think you're handling backport approvals now?

PR #978, opened 2019-08-06
Bug 1737057: Sub CRD does not validate config

tkashem:
Subscription doesn't validate the new `config` block. Add API
validation to Subscription CRD for `Env` and `EnvFrom`.
Use the kubernetes API specification as reference
https://raw.githubusercontent.com/kubernetes/kubernetes/master/api/openapi-spec/swagger.json

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1737057
Jira: https://jira.coreos.com/browse/OLM-1198

### DIFF
--- a/deploy/chart/templates/0000_50_olm_05-subscription.crd.yaml
+++ b/deploy/chart/templates/0000_50_olm_05-subscription.crd.yaml
@@ -71,3 +71,114 @@ spec:
               enum:
               - Manual
               - Automatic
+            config:
+              type: object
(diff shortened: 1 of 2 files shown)


tkashem:
> /lgtm
> 
> some of the descriptions are a little funky here, but it's probably better to keep in sync with upstream (this is what you would see if we were able to reference kube definitions directly)

Yeah, I copied the description verbatim from upstream.

PR #983, opened 2019-08-15
Bug 1746270: Remove deprecated CRD's stored versions to allow CRD update

dinhxuanvu:
Attempt to remove deprecated CRD versions from stored versions in
current CRD. Then, the CRD update can be proceeded.

### DIFF
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -1128,16 +1128,22 @@ func (o *Operator) ResolvePlan(plan *v1alpha1.InstallPlan) error {
 	return nil
 }
 
+func getCRDVersionsMap(crd *v1beta1ext.CustomResourceDefinition) map[string]struct{} {
+	versionsMap := map[string]struct{}{}
+
+	for _, version := range crd.Spec.Versions {
+		versionsMap[version.Name] = struct{}{}
+	}
+	if crd.Spec.Version != "" {
+		versionsMap[crd.Spec.Version] = struct{}{}
+	}
+
+	return versionsMap
+}
+
 // Ensure all existing served versions are present in new CRD
 func ensureCRDVersions(oldCRD *v1beta1ext.CustomResourceDefinition, newCRD *v1beta1ext.CustomResourceDefinition) error {
-	newCRDVersions := map[string]struct{}{}
-
-	for _, newVersion := range newCRD.Spec.Versions {
-		newCRDVersions[newVersion.Name] = struct{}{}
-	}
-	if newCRD.Spec.Version != "" {
-		newCRDVersions[newCRD.Spec.Version] = struct{}{}
-	}
+	newCRDVersions := getCRDVersionsMap(newCRD)
 
 	for _, oldVersion := range oldCRD.Spec.Versions {
 		if oldVersion.Served {
@@ -1203,8 +1209,37 @@ func (o *Operator) validateExistingCRs(gvr schema.GroupVersionResource, newCRD *
 			return fmt.Errorf("error validating custom resource against new schema %#v: %s", newCRD.Spec.Validation, err)
 		}
 	}
-
 	return nil
+}
+
+// Attempt to remove stored versions that have been deprecated before allowing
+// those versions to be removed from the new CRD.
+// The function may not always succeed as storedVersions requires at least one
+// version. If there is only stored version, it won't be removed until a new
+// stored version is added.
+func removeDeprecatedStoredVersions(oldCRD *v1beta1ext.CustomResourceDefinition, newCRD *v1beta1ext.CustomResourceDefinition) *v1beta1ext.CustomResourceDefinition {
+	// StoredVersions requires to have at least one version.
+	if len(oldCRD.Status.StoredVersions) <= 1 {
+		return nil
+	}
+
+	updatedCRD := oldCRD.DeepCopy()
+
+	newCRDVersions := getCRDVersionsMap(newCRD)
+	newStoredVersions := []string{}
+	for _, v := range oldCRD.Status.StoredVersions {
+		_, ok := newCRDVersions[v]
+		if ok {
+			newStoredVersions = append(newStoredVersions, v)
+		}
+	}
+
+	if len(newStoredVersions) < 1 {
+		return nil
+	} else {
+		updatedCRD.Status.StoredVersions = newStoredVersions
+		return updatedCRD
+	}
 }
 
 // ExecutePlan applies a planned InstallPlan to a namespace.
@@ -1268,6 +1303,14 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
 						if len(matchedCSV) == 1 {
 							o.logger.Debugf("Found one owner for CRD %v", crd)
 
+							crdStatus := removeDeprecatedStoredVersions(currentCRD, &crd)
+							if crdStatus != nil {
+								_, err = o.opClient.ApiextensionsV1beta1Interface().ApiextensionsV1beta1().CustomResourceDefinitions().UpdateStatus(crdStatus)
+								if err != nil {
+									return errorwrap.Wrapf(err, "error updating CRD's status: %s", step.Resource.Name)
+								}
+							}
+
 							_, err = o.opClient.ApiextensionsV1beta1Interface().ApiextensionsV1beta1().CustomResourceDefinitions().Update(&crd)
 							if err != nil {
 								return errorwrap.Wrapf(err, "error updating CRD: %s", step.Resource.Name)
@@ -1282,6 +1325,14 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
 
 							if err = o.validateCustomResourceDefinition(currentCRD, &crd); err != nil {
 								return errorwrap.Wrapf(err, "error validating existing CRs agains new CRD's schema: %s", step.Resource.Name)
+							}
+
+							crdStatus := removeDeprecatedStoredVersions(currentCRD, &crd)
+							if crdStatus != nil {
+								_, err = o.opClient.ApiextensionsV1beta1Interface().ApiextensionsV1beta1().CustomResourceDefinitions().UpdateStatus(crdStatus)
+								if err != nil {
+									return errorwrap.Wrapf(err, "error updating CRD's status: %s", step.Resource.Name)
+								}
 							}
 
 							_, err = o.opClient.ApiextensionsV1beta1Interface().ApiextensionsV1beta1().CustomResourceDefinitions().Update(&crd)
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -1217,16 +1217,14 @@ func (o *Operator) validateExistingCRs(gvr schema.GroupVersionResource, newCRD *
 // The function may not always succeed as storedVersions requires at least one
 // version. If there is only stored version, it won't be removed until a new
 // stored version is added.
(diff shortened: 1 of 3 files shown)


test/e2e/installplan_e2e_test.go
```diff
@@ -823,6 +823,39 @@ func TestInstallPlanWithCRDSchemaChange(t *testing.T) {
 				return &newCRD
 			}(),
 		},
+		{
+			name:          "existing version is present in new CRD (decapitated field)",
```
jpeeler: off with that field's head! :)

pkg/controller/operators/catalog/operator.go
```diff
@@ -1125,10 +1145,46 @@ func (o *Operator) validateExistingCRs(gvr schema.GroupVersionResource, newCRD *
 			return fmt.Errorf("error validating custom resource against new schema %#v: %s", newCRD.Spec.Validation, err)
 		}
 	}
-
 	return nil
 }
 
+// Attempt to remove stored versions that have been deprecated before allowing
+// those versions to be removed from the new CRD.
+// The functiona may not always succeed as storedVersions requires at least one
```
jpeeler: function

pkg/controller/operators/catalog/operator.go
```diff
@@ -1064,23 +1064,41 @@ func (o *Operator) ResolvePlan(plan *v1alpha1.InstallPlan) error {
...
+		if oldVersion.Served {
+			_, ok := newCRDVersions[oldVersion.Name]
+			if !ok {
+				return fmt.Errorf("New CRD (%s) must contain existing served versions (%s)", oldCRD.Name, oldVersion.Name)
 			}
 		}
-		if !versionPresent {
-			return fmt.Errorf("not allowing CRD (%v) update with unincluded version %v", newCRD.GetName(), oldVersion)
+	}
+	if oldCRD.Spec.Version != "" {
```
jpeeler: Shouldn't the served boolean be checked here too?

dinhxuanvu: The `version` field is just a string. There is no served field.

pkg/controller/operators/catalog/operator.go
```diff
@@ -1125,10 +1145,46 @@ func (o *Operator) validateExistingCRs(gvr schema.GroupVersionResource, newCRD *
...
+	for i, v := range oldCRD.Status.StoredVersions {
+		storedVersionsMap[v] = i
+	}
+
+	newCRDVersions := getCRDVersionsMap(newCRD)
+
+	for _, version := range oldCRD.Status.StoredVersions {
+		_, ok := newCRDVersions[version]
+		if !ok {
+			delete(storedVersionsMap, version)
```
jpeeler: Instead of doing this delete, you could combine with the above for loop and just insert when the old version is also present in the newCRDVersions map.

pkg/controller/operators/catalog/operator.go
```diff
@@ -1091,11 +1109,13 @@ func (o *Operator) validateCustomResourceDefinition(oldCRD *v1beta1ext.CustomRes
 	if err := v1beta1ext.Convert_v1beta1_CustomResourceDefinition_To_apiextensions_CustomResourceDefinition(newCRD, convertedCRD, nil); err != nil {
 		return err
 	}
-	for _, oldVersion := range oldCRD.Spec.Versions {
-		gvr := schema.GroupVersionResource{Group: oldCRD.Spec.Group, Version: oldVersion.Name, Resource: oldCRD.Spec.Names.Plural}
-		err := o.validateExistingCRs(gvr, convertedCRD)
-		if err != nil {
-			return err
+	for _, version := range oldCRD.Spec.Versions {
+		if version.Served {
```
ecordell: nit: 
```
if !versioned.Served {
  continue
}
```

pkg/controller/operators/catalog/operator.go
```diff
@@ -1106,7 +1126,7 @@ func (o *Operator) validateCustomResourceDefinition(oldCRD *v1beta1ext.CustomRes
 			return err
 		}
 	}
-
+	o.logger.Debugf("Successfully validate CRD %s\n", newCRD.Name)
```
ecordell: nit: `validated`

pkg/controller/operators/catalog/operator.go
```diff
@@ -1125,10 +1145,37 @@ func (o *Operator) validateExistingCRs(gvr schema.GroupVersionResource, newCRD *
...
-
 	return nil
 }
 
+// Attempt to remove stored versions that have been deprecated before allowing
+// those versions to be removed from the new CRD.
+// The function may not always succeed as storedVersions requires at least one
+// version. If there is only stored version, it won't be removed until a new
+// stored version is added.
+func removeDeprecatedStoredVersions(oldCRD *v1beta1ext.CustomResourceDefinition, newCRD *v1beta1ext.CustomResourceDefinition) *v1beta1ext.CustomResourceDefinition {
```
ecordell: this is taking a reference to `oldCRD` and modifying it, and also returning that same modified one. I think we should pick one or the other - either pass a ref and modify it (and don't return it) or make a copy and return the copy

jpeeler:
That unit test failure is real.

pkg/controller/operators/catalog/operator.go
```diff
@@ -1153,16 +1153,14 @@ func (o *Operator) validateExistingCRs(gvr schema.GroupVersionResource, newCRD *
 // The function may not always succeed as storedVersions requires at least one
 // version. If there is only stored version, it won't be removed until a new
 // stored version is added.
-func removeDeprecatedStoredVersions(oldCRD *v1beta1ext.CustomResourceDefinition, newCRD *v1beta1ext.CustomResourceDefinition) *v1beta1ext.CustomResourceDefinition {
+func removeDeprecatedStoredVersions(oldCRD *v1beta1ext.CustomResourceDefinition, newCRD *v1beta1ext.CustomResourceDefinition) []string {
```
jpeeler: Nice change

pkg/controller/operators/catalog/operator.go
```diff
@@ -1191,21 +1237,42 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
...
 							o.logger.Debugf("Found one owner for CRD %v", crd)
 
+							crdStatus := removeDeprecatedStoredVersions(currentCRD, &crd)
+							if crdStatus != nil {
+								currentCRD.Status.StoredVersions = crdStatus
+								resultCRD, err := o.opClient.ApiextensionsV1beta1Interface().ApiextensionsV1beta1().CustomResourceDefinitions().UpdateStatus(currentCRD)
+								if err != nil {
+									return errorwrap.Wrapf(err, "error updating CRD's status: %s", step.Resource.Name)
+								}
+								crd.SetResourceVersion(resultCRD.GetResourceVersion())
```
jpeeler: I believe I misadvised in the past on this. Instead of pulling out the resource version, it's better to save off the result of the create here: https://github.com/operator-framework/operator-lifecycle-manager/pull/983/files#diff-efdf1c2d694e58ab3f44e9e02a4900b9R1224.

pkg/controller/operators/catalog/operator.go
```diff
@@ -1191,21 +1237,42 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
...
 							}
 
+							crdStatus := removeDeprecatedStoredVersions(currentCRD, &crd)
+							if crdStatus != nil {
+								currentCRD.Status.StoredVersions = crdStatus
+								resultCRD, err := o.opClient.ApiextensionsV1beta1Interface().ApiextensionsV1beta1().CustomResourceDefinitions().UpdateStatus(currentCRD)
+								if err != nil {
+									return errorwrap.Wrapf(err, "error updating CRD's status: %s", step.Resource.Name)
+								}
+								crd.SetResourceVersion(resultCRD.GetResourceVersion())
```
jpeeler: The above suggested change should make this go away too.

test/e2e/installplan_e2e_test.go
```diff
@@ -823,6 +845,39 @@ func TestInstallPlanWithCRDSchemaChange(t *testing.T) {
...
+			expectedPhase: v1alpha1.InstallPlanPhaseComplete,
+			oldCRD: func() *apiextensions.CustomResourceDefinition {
+				oldCRD := newCRD(mainCRDPlural + "d")
+				oldCRD.Spec.Version = "v1alpha1"
+				return &oldCRD
+			}(),
+			newCRD: func() *apiextensions.CustomResourceDefinition {
+				newCRD := newCRD(mainCRDPlural + "d")
+				newCRD.Spec.Version = "v1alpha1"
+				newCRD.Spec.Validation = &apiextensions.CustomResourceValidation{
```
jpeeler: Did you try by any chance try and add this validation to newCRD?

test/e2e/installplan_e2e_test.go
```diff
@@ -960,6 +1015,260 @@ func TestInstallPlanWithCRDSchemaChange(t *testing.T) {
...
+	// excluded: new CRD, same version, same schema - won't trigger a CRD update
+	tests := []struct {
+		name            string
+		expectedPhase   v1alpha1.InstallPlanPhase
+		oldCRD          *apiextensions.CustomResourceDefinition
+		intermediateCRD *apiextensions.CustomResourceDefinition
+		newCRD          *apiextensions.CustomResourceDefinition
+	}{
+		{
+			name:          "upgrade CRD with deprecated version",
```
jpeeler: Since you only have one scenario, do you think additional tests should be added in the future? If so, how about add them now? Since this e2e is kind of lengthy, I wonder if "smooshing" it all together might make it a little shorter.

dinhxuanvu: This is a test case for standard deprecating a CRD version flow that is recommended on the doc. So it is only one case to properly remove a CRD version. I can add a failure case but it is a failure on k8s side as users don't follow the k8s doc. So the new `removeDeprecatedStoredVersions` function won't do anything so the failure will be on the k8s `Update` func. I don't think it is worth a test case to basically test a k8s func. The unit test case is to test the those rejected conditions for `removeDeprecatedStoredVersions`. That is faster and more efficient given the e2e test case is as long as it is, not just code-wise but steps-wise.

dinhxuanvu: There are sone duplicated codes so I can group them into some functions to reduce the length.

pkg/controller/operators/catalog/operator.go
jpeeler: @dinhxuanvu pointed out to me that the create has already failed at this point. Instead I propose doing a get on the CRD to get the proper fields filled in.

test/e2e/installplan_e2e_test.go
dinhxuanvu: There is a comparison to determine if the new and old CRDs are different before going ahead of upgrade process. The validation part is here to make sure they are different.

pkg/controller/operators/catalog/operator.go
```diff
@@ -1190,26 +1238,32 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
...
 							}
 
 							if err = o.validateCustomResourceDefinition(currentCRD, &crd); err != nil {
 								return errorwrap.Wrapf(err, "error validating existing CRs agains new CRD's schema: %s", step.Resource.Name)
 							}
-
-							_, err = o.opClient.ApiextensionsV1beta1Interface().ApiextensionsV1beta1().CustomResourceDefinitions().Update(&crd)
+						}
+						// Remove deprecated version in CRD storedVersions
+						crdStatus := removeDeprecatedStoredVersions(currentCRD, &crd)
```
jpeeler: nit: I think this could be better named as storedVersions.

PR #988, opened 2019-08-16
(doc) How to update operators documentation

dinhxuanvu:


mhrivnak:
Can we move this forward soon and get it merged? I would like to contribute additional docs about z-stream management before I forget what we settled on; that'll be a lot easier with this in-place. Thanks!

dinhxuanvu:
I will get this PR updated to address some of the feedbacks and will get it moving forward and merged soon. Thanks.

ecordell:
Quick note - docs now live in `/docs`, so these should be moved as well.

ecordell:
doc-only, merging

PR #989, opened 2019-08-16
Bug 1727042: feat(operatorstatus): report packageserver as upgradeable

ecordell:
The PR that added Upgradeable for olm and catalog missed packageserver.


kevinrizza:
/lgtm cancel

Didn't know that would actually add the label :)

PR #985, opened 2019-08-15
Proposal for PackageManifest Icon Subresource

alecmerdler:
### Description

Addressing the tech debt of including base64 image data in the body of the `PackageManifest` object by creating a new subresource which serves the file data. Needed to address [poor load times for **OperatorHub**](https://bugzilla.redhat.com/show_bug.cgi?id=1727800).

Proposal for https://jira.coreos.com/browse/OLM-1211

PR #990, opened 2019-08-16
Bug 1727800: Icon Subresource for PackageManifest

alecmerdler:
### Description

Adds `packagemanifests/icon` subresource, which serves logo icons for a given `PackageManifest`.

Caching will be accomplished in this way:
1. Console requests `/packagemanifests/<pkg>/icon?resourceVersion=<pkg>.<channel>.<csvName>`
2. Server gets bundle from gRPC `operator-registry` and sends the image data with an ETag header
3. Next time browser requests the icon endpoint, sends `If-None-Match` header containing the ETag
4. Server checks the existence of ETag and sends `HTTP 304`
5. When new `ClusterServiceVersion` is added to the package, console will change the `resourceVersion` in the request URL and therefore not send an `If-None-Match` header.
6. User can always perform hard refresh or clear cache to retrieve fresh icons.

### Resources Used

- https://sourcegraph.com/github.com/kubernetes/kubernetes/-/blob/pkg/registry/core/pod/storage/storage.go#L69
- https://sourcegraph.com/github.com/kubernetes/kubernetes@master/-/blob/pkg/registry/core/rest/storage_core.go#L234
- https://sourcegraph.com/github.com/kubernetes/kubernetes/-/blob/pkg/registry/core/pod/rest/subresources.go#L134

Implements https://github.com/operator-framework/operator-lifecycle-manager/pull/985
Addresses https://jira.coreos.com/browse/OLM-1211

### DIFF
--- a/pkg/package-server/apis/operators/packagemanifest.go
+++ b/pkg/package-server/apis/operators/packagemanifest.go
@@ -11,11 +11,17 @@ func CreateCSVDescription(csv *operatorsv1alpha1.ClusterServiceVersion) CSVDescr
 			Name: csv.Spec.Provider.Name,
 			URL:  csv.Spec.Provider.URL,
 		},
-		Annotations:               csv.GetAnnotations(),
-		LongDescription:           csv.Spec.Description,
-		InstallModes:              csv.Spec.InstallModes,
-		CustomResourceDefinitions: csv.Spec.CustomResourceDefinitions,
-		APIServiceDefinitions:     csv.Spec.APIServiceDefinitions,
+		Annotations:     csv.GetAnnotations(),
+		LongDescription: csv.Spec.Description,
+		InstallModes:    csv.Spec.InstallModes,
+		CustomResourceDefinitions: operatorsv1alpha1.CustomResourceDefinitions{
+			Owned:    descriptionsForCRDs(csv.Spec.CustomResourceDefinitions.Owned),
+			Required: descriptionsForCRDs(csv.Spec.CustomResourceDefinitions.Required),
+		},
+		APIServiceDefinitions: operatorsv1alpha1.APIServiceDefinitions{
+			Owned:    descriptionsForAPIServices(csv.Spec.APIServiceDefinitions.Owned),
+			Required: descriptionsForAPIServices(csv.Spec.APIServiceDefinitions.Required),
+		},
 	}
 
 	icons := make([]Icon, len(csv.Spec.Icon))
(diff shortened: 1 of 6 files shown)


pkg/package-server/apis/operators/packagemanifest.go
```diff
@@ -32,3 +38,34 @@ func CreateCSVDescription(csv *operatorsv1alpha1.ClusterServiceVersion) CSVDescr
 
 	return desc
 }
+
+// descriptionsForCRDs filters certain fields from provided API descriptions to reduce response size.
+func descriptionsForCRDs(crds []operatorsv1alpha1.CRDDescription) []operatorsv1alpha1.CRDDescription {
```
alecmerdler: Strip out `spec` and `status` descriptors, we don't need them.

dinhxuanvu:
lgtm

alecmerdler:
@ecordell Is this good to merge?

pkg/package-server/apis/operators/packagemanifest.go
ecordell: We really shouldn't remove fields that we've already published. Is there an actual bug if they remain?

pkg/package-server/storage/subresources.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+// New satisfies the Storage interface
+func (s *LogoStorage) New() runtime.Object {
+	return &operators.PackageManifest{}
+}
+
+// Connect satisfies the Connector interface and returns the image icon file for a given `PackageManifest`
+func (s *LogoStorage) Connect(ctx context.Context, name string, options runtime.Object, responder rest.Responder) (http.Handler, error) {
+	var handler http.HandlerFunc = func(w http.ResponseWriter, r *http.Request) {
```
ecordell: nit: why even have this variable?

pkg/package-server/storage/subresources.go
```diff
@@ -0,0 +1,92 @@
...
+		if match := r.Header.Get("If-None-Match"); match != "" && r.URL.Query().Get("resourceVersion") != "" {
+			w.WriteHeader(http.StatusNotModified)
+			return
+		}
+
+		namespace := genericreq.NamespaceValue(ctx)
+		pkg, err := s.prov.Get(namespace, name)
+		if err != nil || pkg == nil || len(pkg.Status.Channels[0].CurrentCSVDesc.Icon) == 0 {
```
ecordell: can we make this section more defensive? don't assume there are any channels, don't assume that Icon slice is not nil, etc.

pkg/package-server/apis/operators/packagemanifest.go
alecmerdler: Anything we can do to decrease the JSON response size. Perhaps it would be better to add a query parameter for filtering these fields (since it's an aggragated apiserver so we have that control).

dinhxuanvu: I recall there was a question on forum-operator about these 2 fields and we did say it is legacy-only with intention to remove those fields.

pkg/package-server/storage/subresources.go
alecmerdler: Needed for type coercion in order to use [regular functions as an `http.Handler`](https://godoc.org/net/http#HandlerFunc). Get this error if you try and return just the func itself:
```
cannot use (func(w http.ResponseWriter, r *http.Request) literal) (value of type func(w http.ResponseWriter, r *http.Request)) as http.Handler value in return statement: missing method ServeHTTPLSP

```

pkg/package-server/storage/subresources.go
alecmerdler: Added check for no channels, but `len(nil)` will return 0 in Go.

pkg/package-server/apis/operators/packagemanifest.go
ecordell: If we want to actually remove fields we need to bump the version of the api.

ecordell: If the fields remain but are nil, this is fine.

PR #994, opened 2019-08-19
feat(scripts): add shell lib with some common helper funcs

njhale:
Make rolling out CSVs and OLM easier by adding a shell library with some utility functions.


ecordell:
/hold cancel
/lgtm

this doesn't touch code, so no reason to hold it.

PR #997, opened 2019-08-19
[WIP] check the e2e test behavior with sync concurrency set to 1

tkashem:


ecordell:
@tkashem Do we still want or need this?

PR #1000, opened 2019-08-20
[WIP] [e2e] Reproducer for duplicate InstallPlans

gallettilance:
When two subscriptions are created in rapid succession, two installplans can get created for the same subscription.

### DIFF
--- a/test/e2e/subscription_e2e_test.go
+++ b/test/e2e/subscription_e2e_test.go
@@ -389,6 +389,29 @@ func buildSubscriptionCleanupFunc(t *testing.T, crc versioned.Interface, subscri
 	}
 }
 
+func scheduleCreateSubscription(t *testing.T, crc versioned.Interface, namespace, name, catalog, packageName, channel, startingCSV string, approval v1alpha1.Approval) error {
+	subscription := &v1alpha1.Subscription{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       v1alpha1.SubscriptionKind,
+			APIVersion: v1alpha1.SubscriptionCRDAPIVersion,
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: namespace,
+			Name:      name,
+		},
+		Spec: &v1alpha1.SubscriptionSpec{
+			CatalogSource:          catalog,
+			CatalogSourceNamespace: namespace,
+			Package:                packageName,
+			Channel:                channel,
+			InstallPlanApproval:    approval,
+		},
+	}
+
+	subscription, err := crc.OperatorsV1alpha1().Subscriptions(namespace).Create(subscription)
+	return err
+}
+
 func createSubscription(t *testing.T, crc versioned.Interface, namespace, name, packageName, channel string, approval v1alpha1.Approval) cleanupFunc {
 	subscription := &v1alpha1.Subscription{
 		TypeMeta: metav1.TypeMeta{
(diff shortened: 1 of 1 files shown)


test/e2e/subscription_e2e_test.go
```diff
@@ -1287,3 +1310,103 @@ func updateInternalCatalog(t *testing.T, c operatorclient.ClientInterface, crc v
...
+	channelNames := []string{}
+	crds := []apiextensions.CustomResourceDefinition{}
+	csvs := []v1alpha1.ClusterServiceVersion{}
+	manifests := []registry.PackageManifest{}
+	for i := 0; i < Numpkgs; i++ {
+		pkgName := genName("pkg-")
+		channelName := genName("channel-")
+		strategy := newNginxInstallStrategy(pkgName, nil, nil)
+		crd := newCRD(pkgName)
+		csv := newCSV(pkgName, ns.GetName(), "", semver.MustParse("0.1.0"), []apiextensions.CustomResourceDefinition{crd}, nil, strategy)
```
ecordell: this is putting n "0.1.0" releases in the configmap; it might be nice to increment it to simulate a real update chain.

test/e2e/subscription_e2e_test.go
```diff
@@ -1287,3 +1310,103 @@ func updateInternalCatalog(t *testing.T, c operatorclient.ClientInterface, crc v
...
+	defer func() {
+		require.NoError(t, c.KubernetesInterface().CoreV1().Namespaces().Delete(ns.GetName(), &metav1.DeleteOptions{}))
+	}()
+
+	Numpkgs := 2
+
+	// Create a CatalogSource, cs, in ns with Numpkgs pkgs
+	pkgNames := []string{}
+	channelNames := []string{}
+	crds := []apiextensions.CustomResourceDefinition{}
```
ecordell: this seems like a good helper function to pull out and share with other tests (this section generating catalogs)

test/e2e/subscription_e2e_test.go
```diff
@@ -1287,3 +1310,103 @@ func updateInternalCatalog(t *testing.T, c operatorclient.ClientInterface, crc v
...
+		crds = append(crds, crd)
+	}
+	catalogName := genName("catalog-")
+	_, cleanupCatalogSource := createInternalCatalogSource(t, c, crc, catalogName, ns.GetName(), manifests, crds, csvs)
+	defer cleanupCatalogSource()
+	_, err = fetchCatalogSource(t, crc, catalogName, ns.GetName(), catalogSourceRegistryPodSynced)
+	require.NoError(t, err)
+
+	// Create Subscription to a package of cs in ns, sub
+	subCreated := make(chan string)
```
ecordell: make a buffered channel with of size Numpkgs?

test/e2e/subscription_e2e_test.go
```diff
@@ -1287,3 +1310,103 @@ func updateInternalCatalog(t *testing.T, c operatorclient.ClientInterface, crc v
...
+			Name: genName("ns-"),
+		},
+	}
+	_, err := c.KubernetesInterface().CoreV1().Namespaces().Create(ns)
+	require.NoError(t, err)
+	defer func() {
+		require.NoError(t, c.KubernetesInterface().CoreV1().Namespaces().Delete(ns.GetName(), &metav1.DeleteOptions{}))
+	}()
+
+	Numpkgs := 2
```
ecordell: nit: lowercase or make a const with PascalCase

test/e2e/subscription_e2e_test.go
galletti94: I'm not sure I understand why we want to simulate a real update chain - could you help clarify?

ecordell: This loop generates a set of csvs, crds, and packages, but they all have the same content. When we create the internal catalog source, they're all getting collapsed/ignored into 1 csv, 1 crd, and 1 package.

To rephrase my original comment - let's either remove the loop here (so there's really only 1 package/csv/crd), or make the loop generate content that can actually get loaded by a catalog by incrementing the versions, changing the names, setting up replacement chains, etc.

Unless I'm missing why this is the way it is?

njhale:
/retest

@gallettilance is this still a WIP?

ecordell:
This was covered by another recent PR.

PR #1001, opened 2019-08-20
Bug 1744245: fix e2e failure

tkashem:
Back to back delete and recreate of a subscription object causes
operator install to fail.

How to reproduce:
- Create a CatalogSource object
- Create a subscription that refers to the CatalogSource above.
- Wait for the operator to install successfully.
- Update the CatalogSource
- Wait for the CatalogSource to become healthy
- Delete the Subscription object ( from above ).
- Create the Subscription object ( no time delay between delete
  and create ). Delete and Create can be done one after another,
  there is no need to make them concurrent.

The operator install will fail, Subscription status will have an error
condition `ReferencedInstallPlanNotFound`. The new install plan object
created by OLM gets deleted by GC.

Root cause:
- OLM uses a lister to get the list of Subscription(s) in a given
  namespace and sets the relevant subscriptions(s) found in the list as
  owner of the installplan object(s).
- Because lister uses cache, it will return a deleted subscription
  until the cache is synced.
- The new installplan object may get an owner ref that points to the
  deleted subscription.
- GC garbage collects the deleted subscription and consequently
  deletes the new InstallPlan.
- Subscription reconciler reports that the new InstallPlan object is
  missing and moves the Subscription to a Failed state.

The api audit log has entries that validates that GC is rightfully
"deleting" the new InstallPlan object.

Fix:
- For now, use a direct non-cached client to retrieve the list of
  Subscription.

### DIFF
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -131,7 +131,7 @@ func NewOperator(ctx context.Context, kubeconfigPath string, clock utilclock.Clo
 		client:                 crClient,
 		lister:                 lister,
 		namespace:              operatorNamespace,
-		resolver:               resolver.NewOperatorsV1alpha1Resolver(lister),
+		resolver:               resolver.NewOperatorsV1alpha1Resolver(lister, crClient),
 		catsrcQueueSet:         queueinformer.NewEmptyResourceQueueSet(),
 		subQueueSet:            queueinformer.NewEmptyResourceQueueSet(),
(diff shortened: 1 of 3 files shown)


ecordell:
Can we pull in @galletti94 's test case and get both in this PR?

tkashem:
> Can we pull in @galletti94 's test case and get both in this PR?
Do you want both fixes in the same PR? This might delay other PR(s) blocked. Can we get this merged and unblock folks and then we can focus on the issue @galletti94 is working?

PR #1002, opened 2019-08-21
Bug 1727800: Remove Icon from `PackageManifest` Resource

alecmerdler:
### Description

Reduces the large HTTP size of the `PackageManifest` resource by removing the base64-encoded icon data. Clients should use the [`packagemanifests/icon` subresource](https://github.com/operator-framework/operator-lifecycle-manager/pull/990) instead. This field is **optional**, so clients should have been checking for `nil` values and won't be broken by this change.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1727800
Addresses https://jira.coreos.com/browse/OLM-1211

### DIFF
--- a/pkg/package-server/storage/reststorage.go
+++ b/pkg/package-server/storage/reststorage.go
@@ -85,6 +85,12 @@ func (m *PackageManifestStorage) List(ctx context.Context, options *metainternal
 			filtered = append(filtered, manifest)
 		}
 	}
+	// Strip logo icons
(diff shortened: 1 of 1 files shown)


ecordell:
@alecmerdler is there a pr to track for when console no longer uses the icon field?

alecmerdler:
@ecordell The console PR has [already been merged](https://github.com/openshift/console/pull/2425).

PR #1004, opened 2019-08-22
Bug 1738203: Add relatedObjects to cluster operator status 

dinhxuanvu:

### DIFF
--- a/cmd/catalog/main.go
+++ b/cmd/catalog/main.go
@@ -151,7 +151,7 @@ func main() {
 	<-ready
 
 	if *writeStatusName != "" {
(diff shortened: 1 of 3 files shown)


eparis:
/retitle Bug 1745272: Add relatedObjects to cluster operator status

eparis:
There was no 4.2 bugzilla. I created it. We need to wait for QE to verify this is fixed in master. Then run `/bugzilla refresh`

ecordell:
/hold

This will need to be replaced based on newer understanding of `relatedObjects` (see #1018)

PR #1006, opened 2019-08-22
Bug 1740332: OLM should resume operator install

tkashem:
OLM should automatically resume operator install when a user grants
proper permission(s).

Currently, a user has to manually delete the subscription and recreate
it in order to trigger a reinstall of the operator.

Do the following to trigger reinstall:
- Add a new field 'AttenuatedServiceAccountRef' to status of
  InstallPlan. We use this to refer to the ServiceAccount that will be
  used to do attenuated scoped install of the operator.
- Watch on Role(Binding), ServiceAccount resources. When these RBAC
  resources are added/updated find the target InstallPlan object.
- Update the status.phase of the InstallPlan object to Installing.
  This will trigger a sync of the InstallPlan.

Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1740332
Jira: https://jira.coreos.com/browse/OLM-1244

### DIFF
--- a/pkg/controller/operators/catalog/installplan_sync.go
+++ b/pkg/controller/operators/catalog/installplan_sync.go
@@ -1,0 +1,100 @@
+package catalog
+
+import (
+	"errors"
+
+	"github.com/sirupsen/logrus"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/ownerutil"
+)
+
+// When a user adds permission to a ServiceAccount by creating or updating
+// Role/RoleBinding then we expect the InstallPlan that refers to the
+// ServiceAccount to be retried if it has failed to install before due to
+// permission issue(s).
+func (o *Operator) triggerInstallPlanRetry(obj interface{}) (syncError error) {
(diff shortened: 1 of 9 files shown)


jpeeler:
I don't think approach will work (well) since syncObject will constantly attempt reinstalling without any change in the cluster. Since install plans aren't reconciled constantly by design, I'm thinking the best way to proceed is to watch newly created roles and see if they apply to existing install plans (or perhaps check the status of failed install plans?)

pkg/controller/operators/catalog/operator.go
```diff
@@ -1027,6 +1027,28 @@ func (o *Operator) syncInstallPlans(obj interface{}) (syncError error) {
...
+		out := plan.DeepCopy()
+		out.Status.AttenuatedServiceAccountRef = reference
+
+		if !reflect.DeepEqual(plan, out) {
+			if _, updateErr := o.client.OperatorsV1alpha1().InstallPlans(out.GetNamespace()).UpdateStatus(out); err != nil {
+				syncError = fmt.Errorf("failed to attach attenuated ServiceAccount to status - %v", updateErr)
+				return
+			}
+						
+			logger.Infof("successfully attached attenuated ServiceAccount to status - %s", reference.Name)
```
ecordell: nit: could use a structured logging field for the `reference.Name` here

pkg/controller/operators/catalog/installplan_sync.go
```diff
@@ -0,0 +1,100 @@
...
+
+		logger := o.logger.WithFields(logrus.Fields{
+			"ip":        ip.GetName(),
+			"namespace": ip.GetNamespace(),
+			"phase":     ip.Status.Phase,
+		})
+
+		if updateErr := update(ip); updateErr != nil {
+			errs = append(errs, updateErr)
+			logger.Warnf("failed to kick off InstallPlan retry - %v", updateErr)
```
ecordell: nit `logger.WithError()`

PR #1007, opened 2019-08-23
Bug 1727042: Ensure status monitor always reports upgradeable

kevinrizza:
Currently on startup the initial status reporting sets the upgradeable
field. However, because of the way the status monitor builder recreates
the entire status block on every report, we also need to set upgradeable
every time that object is reported.

For now, report that upgradeable is always true as long as nothing is in
a degraded state.

Pertaining to these bugzillas:
https://bugzilla.redhat.com/show_bug.cgi?id=1727042
https://bugzilla.redhat.com/show_bug.cgi?id=1727032

### DIFF
--- a/pkg/lib/operatorstatus/builder.go
+++ b/pkg/lib/operatorstatus/builder.go
@@ -71,6 +71,21 @@ func (b *Builder) WithAvailable(status configv1.ConditionStatus, message string)
 	return b
 }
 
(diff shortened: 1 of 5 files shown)


kevinrizza:
Closing in favor of https://github.com/operator-framework/operator-lifecycle-manager/pull/989

PR #1014, opened 2019-08-28
[CatalogSource] Set Global Catalog NS with EnvVar

awgreene:
This commit introduces a change that makes it possible to configure the
Global Catalog Namespace by adding the CATALOG_NAMESPACE environment
variable to the catalog-operator deployment.


jpeeler:
/hold
guess it only works on a line by itself

awgreene:
@ecordell @jpeeler please let me know if you have any other change requests.

PR #1019, opened 2019-08-30
Bug 1748914: Drop the hard-coded DNS suffix

skitt:
Having the catalog source hard-code its cluster.local suffix makes it
impossible to deploy in environments which use different suffixes.
This patch drops it, relying on the default search which K8s setups
expect to be set correctly anyway.

(For context, we use multiple clusters with different suffixes in our Submariner multi-cluster end-to-end tests.)


ecordell:
/bugzilla refresh

Thanks @skitt! I verified this still locally as well.

PR #1020, opened 2019-08-30
Allow building with older Docker versions

skitt:
Older Docker versions (as available in many distributions) require
their descriptor files to be inside their work directory. This is the
only impediment to using these versions of Docker with the OLM local
build infrastructure, so this patch copies the descriptors into the
work directory (bin, which is deleted on clean-up) and references the
copies.

Signed-off-by: Stephen Kitt <skitt@redhat.com>

### DIFF
--- a/scripts/build_local.sh
+++ b/scripts/build_local.sh
@@ -11,8 +11,9 @@ if [ -z "$NO_MINIKUBE" ]; then
   kubectl config use-context minikube
 fi
 
-docker build -f local.Dockerfile -t quay.io/operator-framework/olm:local -t quay.io/operator-framework/olm-e2e:local ./bin
-docker build -f test/e2e/hang.Dockerfile -t hang:10 ./bin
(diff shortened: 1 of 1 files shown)


jpeeler:
I'm not sure this is worth the investment. I believe there are other places that aren't compatible with the current packaged version of Docker (assuming you're using docker in Fedora), such as the e2e builds.

Soon Fedora won't have a docker v1.13 based package any more: https://src.fedoraproject.org/rpms/docker/blob/master/f/dead.package

skitt:
> I'm not sure this is worth the investment. I believe there are other places that aren't compatible with the current packaged version of Docker (assuming you're using docker in Fedora), such as the e2e builds.
> 
> Soon Fedora won't have a docker v1.13 based package any more: https://src.fedoraproject.org/rpms/docker/blob/master/f/dead.package

Fair enough, I’ll drop this (and finish switching to `podman-docker`).

PR #1025, opened 2019-09-03
Bug 1746159: Cleanup leftover cross-namespace OwnerReferences

njhale:
Adds cross-namespace OwnerReference cleanup logic to OLM startup.

Cross-namespace and cluster-to-namespace scoped OwnerReferences may cause sibling resources in the owner namespace to be deleted sporadically (see [CVE-2019-3884]( https://access.redhat.com/security/cve/cve-2019-3884)). Older versions of OLM use both types of OwnerReference, and in cases where OLM is updated, they must be removed to prevent erroneous deletion of OLM's self-hosted components.

### DIFF
--- a/cmd/olm/cleanup.go
+++ b/cmd/olm/cleanup.go
@@ -4,10 +4,14 @@ import (
 	"time"
 
 	"github.com/sirupsen/logrus"
+	rbacv1 "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/util/retry"
 
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/clientset/versioned"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/operatorclient"
 )
@@ -19,6 +23,7 @@ const (
 
 type checkResourceFunc func() error
 type deleteResourceFunc func() error
+type mutateMeta func(obj metav1.Object) (mutated bool)
 
 func cleanup(logger *logrus.Logger, c operatorclient.ClientInterface, crc versioned.Interface) {
 	if err := waitForDelete(checkCatalogSource(crc, "olm-operators"), deleteCatalogSource(crc, "olm-operators")); err != nil {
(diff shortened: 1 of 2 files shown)


cmd/olm/cleanup.go
```diff
@@ -119,3 +127,115 @@ func deleteCatalogSource(crc versioned.Interface, name string) deleteResourceFun
...
+		case *rbacv1.Role:
+			update = func() error {
+				_, err = c.KubernetesInterface().RbacV1().Roles(v.GetNamespace()).Update(v)
+				return err
+			}
+		case *rbacv1.RoleBinding:
+			update = func() error {
+				_, err = c.KubernetesInterface().RbacV1().RoleBindings(v.GetNamespace()).Update(v)
+				return err
+			}
```
jpeeler: Why doesn't you make the default case log a warning message and then return? Then you can remove the empty update function assignment above.

cmd/olm/cleanup.go
```diff
@@ -119,3 +127,115 @@ func deleteCatalogSource(crc versioned.Interface, name string) deleteResourceFun
...
+	for _, obj := range objs {
+		if !removeBadRefs(obj) {
+			continue
+		}
+
+		update := func() error {
+			// If this is not a type we care about, do nothing
+			return nil
+		}
+		switch v := obj.(type) {
```
jpeeler: Super paranoid comment here (and slightly contradicts my comment below), but this sort of change almost makes me ponder a v1alpha2 version bump. Then you could easily filter out potential resources that might be problematic. I'm wondering if it's possible that a future backport ends up pulling in a new resource and this code doesn't get updated. I might be way off base here.

cmd/olm/cleanup.go
```diff
@@ -119,3 +127,115 @@ func deleteCatalogSource(crc versioned.Interface, name string) deleteResourceFun
...
+
+	return nil
+}
+
+func crossNamespaceOwnerReferenceRemoval(kind string, uidNamespaces map[types.UID]string) mutateMeta {
+	return func(obj metav1.Object) (mutated bool) {
+		var cleanRefs []metav1.OwnerReference
+		objNamespace := obj.GetNamespace()
+		for _, ref := range obj.GetOwnerReferences() {
+			if ref.Kind == kind {
```
ecordell: nit: 
```
if ref.Kind != kind {
  continue
}
```

cmd/olm/cleanup.go
ecordell: Hopefully we don't add cross-namespace ownerrefs in newer versions, I don't think we have to worry too much about that.

cmd/olm/cleanup.go
njhale: SGTM. We could also panic since it's probably a programmer error. This already merged, so I may try to sneak this in with another PR at some point.

Issue #1032, opened 2019-09-09
Build fails because of coreos/etcd checksum mismatch


skitt:
## Bug Report

**What did you do?**

I cloned the OLM repository, and tried running `make` inside it.

**What did you expect to see?**

I expected the tests to pass and the build to succeed.

**What did you see instead? Under which circumstances?**

```
go: downloading github.com/coreos/etcd v3.3.10+incompatible
go: downloading k8s.io/kubernetes v1.14.5-beta.0.0.20190708100021-7936da50c68f
go: downloading google.golang.org/genproto v0.0.0-20181016170114-94acd270e44e
go: downloading github.com/go-openapi/validate v0.19.2
go: extracting github.com/go-openapi/validate v0.19.2
go: downloading github.com/go-openapi/loads v0.19.2
go: extracting google.golang.org/genproto v0.0.0-20181016170114-94acd270e44e
go: downloading github.com/go-openapi/runtime v0.19.0
go: extracting github.com/go-openapi/loads v0.19.2
go: downloading github.com/go-openapi/analysis v0.19.2
verifying github.com/coreos/etcd@v3.3.10+incompatible: checksum mismatch
	downloaded: h1:jFneRYjIvLMLhDLCzuTuU4rSJUjRplcJQ7pD7MnhC04=
	go.sum:     h1:KjVWqrZ5U0wa3CxY2AxlH6/UcB+PK2td1DcsYhA+HRs=

SECURITY ERROR
This download does NOT match an earlier download recorded in go.sum.
The bits may have been replaced on the origin server, or an attacker may
have intercepted the download attempt.

For more information, see 'go help module-auth'.
go test  -v -race -tags=json1 -coverprofile=cover.out -covermode=atomic \
	-coverpkg ./pkg/controller/... ./pkg/...
go: downloading k8s.io/kubernetes v1.14.5-beta.0.0.20190708100021-7936da50c68f
go: downloading github.com/coreos/etcd v3.3.10+incompatible
go: downloading github.com/mattn/go-sqlite3 v1.10.0
go: downloading github.com/go-openapi/runtime v0.19.0
go: extracting github.com/go-openapi/runtime v0.19.0
go: downloading github.com/go-openapi/analysis v0.19.2
go: extracting github.com/go-openapi/analysis v0.19.2
verifying github.com/coreos/etcd@v3.3.10+incompatible: checksum mismatch
[...]

PR #1035, opened 2019-09-10
Restore coreos/etcd h1 checksum

skitt:
This was changed for some reason in
b4f2c08a3ca5c53400d1a534cf119d0776d14840 to an invalid value. This
patch restores the previous value, which is the currently-valid value
too.

Closes: #1032

Signed-off-by: Stephen Kitt <skitt@redhat.com>


ecordell:
Hi @skitt, could you check on this before we merge this? https://github.com/operator-framework/operator-lifecycle-manager/issues/1032#issuecomment-530378024

PR #1043, opened 2019-09-20
Bug 1740857: Backport grpc sourcemanager to 4.1

ecordell:
This backports the grpc sourcemanager to 4.1

The sourcemanager will fix an issue in some 4.1 clusters which see high memory usage. 

**Reviewer Checklist**


pkg/controller/operators/catalog/operator.go
```diff
@@ -211,6 +211,7 @@ func NewOperator(kubeconfigPath string, logger *logrus.Logger, wakeupInterval ti
 		op.lister.CoreV1().RegisterPodLister(namespace, podInformer.Lister())
 		op.lister.CoreV1().RegisterConfigMapLister(namespace, configMapInformer.Lister())
 	}
+	op.sources = grpc.NewSourceStore(logger, 10*time.Second, 10*time.Minute, op.syncSourceState)
```
jpeeler: I don't think this line is necessary, is it?

ecordell: Oh, that's weird! I will fix

PR #1054, opened 2019-10-01
[WIP] (proposal) Make bundle accessible to a cluster

tkashem:
Make bundle accessible to a cluster

### DIFF
--- a/doc/contributors/design-proposals/pull-bundle-on-a-cluster.md
+++ b/doc/contributors/design-proposals/pull-bundle-on-a-cluster.md
@@ -1,0 +1,80 @@
+# Pull Bundle on a Cluster
+
+## Store a bundle in a `ConfigMap`
+Below is the directory layout of the operator bundle inside the image.
+```bash
+$ tree
+/
+├── manifests
+│   ├── testbackup.crd.yaml
+│   ├── testcluster.crd.yaml
+│   ├── testoperator.v0.1.0.clusterserviceversion.yaml
+│   └── testrestore.crd.yaml
+└── metadata
+    └── annotations.yaml
+    
+$ cat /annotations.yaml
+annotations:
+  operators.coreos.com.bundle.resources: "manifests+metadata"
+  operators.coreos.com.bundle.mediatype: "operator-registry+v1"
+```
+
+The following `ConfigMap` maps to the above operator bundle
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: test
+  namespace: test
+  annotations:
+    operators.coreos.com.bundle.resources: "manifests+metadata"
+    operators.coreos.com.bundle.mediatype: "registry+v1"
+data:
+  testbackup.crd.yaml: content of testbackup.crd.yaml
+  testcluster.crd.yaml: content of testcluster.crd.yaml
+  testoperator.v0.1.0.clusterserviceversion.yaml: content oftestoperator.v0.1.0.clusterserviceversion.yaml
+  testrestore.crd.yaml: content of testrestore.crd.yaml
+```
+
+The `key` of a `ConfigMap` has the following format
+```go
+	// Data contains the configuration data.
+	// Each key must consist of alphanumeric characters, '-', '_' or '.'.
+	// Values with non-UTF-8 byte sequences must use the BinaryData field.
+	// The keys stored in Data must not overlap with the keys in
+	// the BinaryData field, this is enforced during validation process.
+	// +optional
+	Data map[string]string `json:"data,omitempty" protobuf:"bytes,2,rep,name=data"`
+```
+
+Notes:
+* The resource file name needs to be manipulated if it contains special characters.
+* The consumer of the `ConfigMap` does will not use the key name in `Data` section to identify the type of resource. It should inspect the content.
+* The consumer will iterate through the `Data` section and and add each resource to the bundle.
+* The annotations from the `annotations.yaml` file is copied to `metadata.annotations` to the `ConfigMap`
+* The `ConfigMap` may have a resource that contains a `PackageManifest` resource. The consumer needs to handle this properly.
+
+
+## Build a Bundle from ConfigMap
+```go
+import (
+	"github.com/operator-framework/operator-registry/pkg/registry"
+	corev1 "k8s.io/api/core/v1"
+)
+
+// Manifest contains a bundle and a PackageManifest.
+type Manifest struct {
+	Bundle          *registry.Bundle
+	PackageManifest *registry.PackageManifest
+}
+
+type Loader interface {
+	Load(cm *corev1.ConfigMap) (manifest *Manifest, err error)
+}
+```
+
+## Managing Lifecycle of the `ConfigMap`
+
+## Things that Happen after reading the configmap
+
+
--- a/doc/contributors/design-proposals/pull-bundle-on-a-cluster.md
+++ b/doc/contributors/design-proposals/pull-bundle-on-a-cluster.md
@@ -1,5 +1,56 @@
 # Pull Bundle on a Cluster
 
+The purpose of this proposal is to allow bundle data to be pulled at install time. At a high level, this starts with launching the bundle inside a pod and then writing the data found inside of the container into a configmap.
+
+## Launching a bundle image for data extraction
+
+The function to call for launching the bundle looks like:
+
+LaunchBundleImage(kubeclient kubernetes.Interface, bundleImage, initImage, namespace string) (*corev1.ConfigMap, *batchv1.Job, error)
+
+Here the bundle image, init image, and namespace are the core parameters. Note that it is a requirement that the service account have permissions to update configmaps in the specified namespace. In the expected usage however, the namespace of OLM will be used and permissions will not be a problem. The configmap and job are returned for the caller to delete when done.
+
+Within the launch function a job is called using a spec simliar to what is described below. Unique random names are able to be depended upon by using generateName for both the job and configmap.
+
+```yaml
+apiVersion: batch/v1
+kind: Job
+metadata:
+  generateName: deploy-bundle-image-
+spec:
+  #ttlSecondsAfterFinished: 0 # alpha feature - https://github.com/kubernetes/enhancements/issues/592, https://github.com/kubernetes/kubernetes/pull/82082
+  template:
+    metadata:
+      name: bundle-image
+    spec:
+      containers:
+      - name: bundle-image
+        image: &image bundle-image
+        command: ['/injected/operator-registry', 'serve']
+        env:
+          - name: CONTAINER_IMAGE
+            value: *image
+        volumeMounts:
+        - name: copydir
+          mountPath: /injected
+      initContainers:
+      - name: copy-binary
+        image: init-operator-manifest
+        imagePullPolicy: Never
+        command: ['/bin/cp', '/operator-registry', '/copy-dest']
+        volumeMounts:
+        - name: copydir
+          mountPath: /copy-dest
+      volumes:
+      - name: copydir
+        emptyDir: {}
+      restartPolicy: OnFailure
+```
+
+## Serving the bundle data
+
+A new command will be made in operator-registry to provide functionality for traversing the directies of the bundle image for writing to a configmap (example above is "operator-registry serve"). The configmap format is described in more detail in the next section. The pre-generated configmap name is specified to be updated with the bundle data. As previously noted, after the configmap data has been read it is the responsibility of the caller to delete both the job and configmap. As a sanity check, the configmap is also updated with the image name as an annotation (olm.imageSource).
+
 ## Store a bundle in a `ConfigMap`
 Below is the directory layout of the operator bundle inside the image.
 ```bash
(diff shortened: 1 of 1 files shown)


doc/contributors/design-proposals/pull-bundle-on-a-cluster.md
```diff
@@ -0,0 +1,80 @@
...
+├── manifests
+│   ├── testbackup.crd.yaml
+│   ├── testcluster.crd.yaml
+│   ├── testoperator.v0.1.0.clusterserviceversion.yaml
+│   └── testrestore.crd.yaml
+└── metadata
+    └── annotations.yaml
+    
+$ cat /annotations.yaml
+annotations:
```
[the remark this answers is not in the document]
jpeeler: We discussed this will remain in yaml format to allow extending with additional sections later on.

doc/contributors/design-proposals/pull-bundle-on-a-cluster.md
```diff
@@ -0,0 +1,80 @@
...
+	// Values with non-UTF-8 byte sequences must use the BinaryData field.
+	// The keys stored in Data must not overlap with the keys in
+	// the BinaryData field, this is enforced during validation process.
+	// +optional
+	Data map[string]string `json:"data,omitempty" protobuf:"bytes,2,rep,name=data"`
+```
+
+Notes:
+* The resource file name needs to be manipulated if it contains special characters.
+* The consumer of the `ConfigMap` does will not use the key name in `Data` section to identify the type of resource. It should inspect the content.
```
awgreene: `does will not` seems to be a grammatical error.

doc/contributors/design-proposals/pull-bundle-on-a-cluster.md
```diff
@@ -0,0 +1,80 @@
...
+* The consumer will iterate through the `Data` section and and add each resource to the bundle.
+* The annotations from the `annotations.yaml` file is copied to `metadata.annotations` to the `ConfigMap`
```
awgreene: Could you explain what you meant by `...is copied to metadata.annotations to the ConfigMap`

awgreene: You are also missing a `.` here.

doc/contributors/design-proposals/pull-bundle-on-a-cluster.md
```diff
@@ -0,0 +1,80 @@
+# Pull Bundle on a Cluster
```
awgreene: Could you include why the proposed change is being made?

doc/contributors/design-proposals/pull-bundle-on-a-cluster.md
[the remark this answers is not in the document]
awgreene: > to allow extending with additional sections later on.

Could you elaborate what you mean by this?

doc/contributors/design-proposals/pull-bundle-on-a-cluster.md
```diff
@@ -0,0 +1,80 @@
...
+  namespace: test
+  annotations:
+    operators.coreos.com.bundle.resources: "manifests+metadata"
+    operators.coreos.com.bundle.mediatype: "registry+v1"
+data:
+  testbackup.crd.yaml: content of testbackup.crd.yaml
+  testcluster.crd.yaml: content of testcluster.crd.yaml
+  testoperator.v0.1.0.clusterserviceversion.yaml: content oftestoperator.v0.1.0.clusterserviceversion.yaml
+  testrestore.crd.yaml: content of testrestore.crd.yaml
+```
```
awgreene: Is this generated by the user?

doc/contributors/design-proposals/pull-bundle-on-a-cluster.md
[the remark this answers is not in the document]
jpeeler: What I was led to believe is another section like:
annotations:
  key: value
moredata:
...

But I don't understand how that makes sense in the context of a file called "annotations.yaml".

doc/contributors/design-proposals/pull-bundle-on-a-cluster.md
jpeeler: This configmap is generated by some new code, which I'm going to describe at the beginning of this document in a bit (today).

doc/contributors/design-proposals/pull-bundle-on-a-cluster.md
jpeeler: The annotations here are referring to lines 30-31.

doc/contributors/design-proposals/pull-bundle-on-a-cluster.md
jpeeler: I added a little bit, but maybe more is necessary.

doc/contributors/design-proposals/pull-bundle-on-a-cluster.md
```diff
@@ -0,0 +1,127 @@
...
+          mountPath: /copy-dest
+      volumes:
+      - name: copydir
+        emptyDir: {}
+      restartPolicy: OnFailure
+```
+
+## Serving the bundle data
+
+New code will be made in operator-registry to provide functionality for traversing the directies of the bundle image for writing to a configmap (example above is "operator-registry serve"), the format for which is discussed in more detail below. The configmap will also use a generated name to avoid collisions and will be labeled to match the requested bundle image. It will be the responsibility of the caller to 1) launch the job, 2) watch for the target configmap to be created 3) to delete the job (until ttlSecondsAfterFinished is available) 4) delete the configmap after reading.
```
jpeeler: Alternatively, instead of generating a name for the configmap, could use a sanitized version of the bundle image name and use that (can't use colons, but maybe that's the only unacceptable character. also 253 character limit might be worth worrying about). Either way, I don't think specifying the name of the configmap by the caller is the correct approach.

jpeeler: Hrm, labels are actually restricted to 63 characters. Docker registry names are limited to 256 characters. And quay names seem to be even potentially longer (https://coreos.com/quay-enterprise/releases/#2.0.4).

doc/contributors/design-proposals/pull-bundle-on-a-cluster.md
```diff
@@ -0,0 +1,127 @@
...
+  generateName: deploy-bundle-image
+spec:
+  #ttlSecondsAfterFinished: 0 # alpha feature - https://github.com/kubernetes/enhancements/issues/592, https://github.com/kubernetes/kubernetes/pull/82082
+  template:
+    metadata:
+      name: bundle-image
+    spec:
+      containers:
+      - name: bundle-image
+        image: &image bundle-image
```
ecordell: This seems odd as an example?

doc/contributors/design-proposals/pull-bundle-on-a-cluster.md
```diff
@@ -0,0 +1,127 @@
...
+        command: ['/injected/operator-registry', 'serve']
+        env:
+          - name: CONTAINER_IMAGE
+            value: *image
```
ecordell: This also seems like an odd example

doc/contributors/design-proposals/pull-bundle-on-a-cluster.md
```diff
@@ -0,0 +1,127 @@
...
+        - name: copydir
+          mountPath: /injected
+      initContainers:
+      - name: copy-binary
+        image: init-operator-manifest
+        imagePullPolicy: Never
+        command: ['/bin/cp', '/operator-registry', '/copy-dest']
+        volumeMounts:
+        - name: copydir
+          mountPath: /copy-dest
```
ecordell: is there a reason to do this instead of, say, `/bin`?

doc/contributors/design-proposals/pull-bundle-on-a-cluster.md
```diff
@@ -0,0 +1,127 @@
...
+	// Each key must consist of alphanumeric characters, '-', '_' or '.'.
+	// Values with non-UTF-8 byte sequences must use the BinaryData field.
+	// The keys stored in Data must not overlap with the keys in
+	// the BinaryData field, this is enforced during validation process.
+	// +optional
+	Data map[string]string `json:"data,omitempty" protobuf:"bytes,2,rep,name=data"`
+```
+
+Notes:
+* The resource file name needs to be manipulated if it contains special characters.
```
ecordell: What if we made restrictions on the bundle format side?

doc/contributors/design-proposals/pull-bundle-on-a-cluster.md
njhale: You could always create a shortened hash of the fully qualified image name -- that would help you meet both the length and character-set restrictions.

njhale: > 3) to delete the job (until ttlSecondsAfterFinished is available)

Is there any reason the caller can't just delete the `job` when it reaches `Completed`? Why wait?

> 4) delete the configmap after reading.

I think we want to keep the `ConfigMap` around as an on-cluster cache.

doc/contributors/design-proposals/pull-bundle-on-a-cluster.md
```diff
@@ -0,0 +1,127 @@
...
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: test
+  namespace: test
+  annotations:
+    operators.coreos.com.bundle.resources: "manifests+metadata"
+    operators.coreos.com.bundle.mediatype: "registry+v1"
+data:
+  testbackup.crd.yaml: content of testbackup.crd.yaml
```
njhale: I think we should shard files across multiple `ConfigMaps` to avoid hitting the resource size limit (RSL). A simple strategy we could use is to generate a `ConfigMap` per file (assume |single file| < RSL - |metadata+kind+etc...|).

doc/contributors/design-proposals/pull-bundle-on-a-cluster.md
```diff
@@ -0,0 +1,127 @@
...
+	// +optional
+	Data map[string]string `json:"data,omitempty" protobuf:"bytes,2,rep,name=data"`
+```
+
+Notes:
+* The resource file name needs to be manipulated if it contains special characters.
+* The consumer of the `ConfigMap` does not use the key name in `Data` section to identify the type of resource. It should inspect the content.
+* The consumer will iterate through the `Data` section and and add each resource to the bundle.
+* The annotations from the `annotations.yaml` file is copied to `metadata.annotations` to the `ConfigMap`.
+* The `ConfigMap` may have a resource that contains a `PackageManifest` resource. The consumer needs to handle this properly.
```
njhale: Can you elaborate on this for me? When does this happen and is it something consumers would trip on?

doc/contributors/design-proposals/pull-bundle-on-a-cluster.md
ecordell: I think (3) is:

- the caller will be responsible for deleting the job
- when ttlSecondsAfterfinished is available, the caller will no longer be required to delete the job.

jpeeler: That's a good idea, however, I pivoted a little bit.
I ended up retracting my original aversion to specifying a configmap, at least within the bundle image. Rather than worrying about character limits I ended up putting the configmap creation code in the "launch" function, so that the caller doesn't have to worry about configmap naming. I'll be updated the proposal here shortly that hopefully fully clarifies the new/final direction.

doc/contributors/design-proposals/pull-bundle-on-a-cluster.md
jpeeler: Were you wanting a more realistic pull spec? It's what I used for testing since I was using local images.

doc/contributors/design-proposals/pull-bundle-on-a-cluster.md
jpeeler: I thought it was a clever usage of repeated nodes, but the implementation won't use this.

doc/contributors/design-proposals/pull-bundle-on-a-cluster.md
jpeeler: It reduces the likelihood of conflicts.

doc/contributors/design-proposals/pull-bundle-on-a-cluster.md
jpeeler: That would be nice, but given that we aren't actually remounting the configmap data (yet) I think handling the translation instead of just failing is best. What do you think?

doc/contributors/design-proposals/pull-bundle-on-a-cluster.md
jpeeler: Yeah, I was referring to the ttlSecondAfterFinished feature being non-alpha. I've tried to rewrite the proposal in such a way that makes it clear that deletion is up to the caller.

doc/contributors/design-proposals/pull-bundle-on-a-cluster.md
jpeeler: Interesting idea. At a minimum the data usage should be checked. If this is a huge concern though, should configmaps even be used for storing the data?

doc/contributors/design-proposals/pull-bundle-on-a-cluster.md
```diff
@@ -0,0 +1,131 @@
...
+```
+
+The `key` of a `ConfigMap` has the following format
+```go
+	// Data contains the configuration data.
+	// Each key must consist of alphanumeric characters, '-', '_' or '.'.
+	// Values with non-UTF-8 byte sequences must use the BinaryData field.
+	// The keys stored in Data must not overlap with the keys in
+	// the BinaryData field, this is enforced during validation process.
+	// +optional
```
awgreene: We could use a [validation webhook](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/) to do this.

doc/contributors/design-proposals/pull-bundle-on-a-cluster.md
ecordell: Consumers being OLM applying the manifests to a cluster?

ecordell:
/lgtm

We may make some minor changes to this over time (esp. w.r.t. sharding) but I think this proposal is good to go, and we should merge it.

Doc-only, merging without tests.

PR #1055, opened 2019-10-01
feat(metrics) record CSV failures

awgreene:
This commit introduces a change so OLM will exposed metrics about a CSV
whenever either of the following conditions are met:
- The CSV enters the failed phase.
- The CSV is already in the failed phase, but the reason has changed.

The metrics are recorded with name, version, and reason labels.

### DIFF
--- a/pkg/controller/operators/olm/operator.go
+++ b/pkg/controller/operators/olm/operator.go
@@ -930,6 +930,8 @@ func (a *Operator) syncClusterServiceVersion(obj interface{}) (syncError error) 
 	})
 	logger.Debug("syncing CSV")
 
+	metrics.EmitCSVMetric(clusterServiceVersion)
+
 	if a.csvNotification != nil {
 		a.csvNotification.OnAddOrUpdate(clusterServiceVersion)
 	}
--- a/test/e2e/metrics_e2e_test.go
+++ b/test/e2e/metrics_e2e_test.go
@@ -6,34 +6,72 @@ import (
 	"testing"
 
 	log "github.com/sirupsen/logrus"
+	"github.com/stretchr/testify/require"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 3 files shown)


test/e2e/metrics_e2e_test.go
```diff
@@ -6,18 +6,59 @@ import (
...
+	failingCSV := v1alpha1.ClusterServiceVersion{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       v1alpha1.ClusterServiceVersionKind,
+			APIVersion: v1alpha1.ClusterServiceVersionAPIVersion,
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name: genName("failing-csv-test-"),
+		},
+		Spec: v1alpha1.ClusterServiceVersionSpec{
+			InstallStrategy: v1alpha1.NamedInstallStrategy{
```
awgreene: This seemed to be the most generic `InstallStrategy` that I could find, if a better alternative exists please let me know.

pkg/controller/operators/olm/operator.go
```diff
@@ -952,6 +952,13 @@ func (a *Operator) syncClusterServiceVersion(obj interface{}) (syncError error)
 		outCSV.Status.Reason == clusterServiceVersion.Status.Reason &&
 		outCSV.Status.Message == clusterServiceVersion.Status.Message) {
 
+		// If the CSV has transitioned into a failed state or is in the failed state for a new reason, record a CSV failure.
+		if outCSV.Status.Phase == v1alpha1.CSVPhaseFailed &&
```
awgreene: We may want to add a check for the version here as well, thoughts?

test/e2e/metrics_e2e_test.go
ecordell: This is good!

ecordell:
/lgtm

Giving a pass on missing docs since we're restructuring those at the moment 😄

PR #1056, opened 2019-10-01
Bug: Fix olm chart

awgreene:
This commit introduces a change that removes some metadata from a CRD
that was causing `make e2e-local` to fail.


tkashem:
@awgreene thanks for catching this!
/lgtm

PR #1058, opened 2019-10-03
feat(subscription) Config Volume and VolumeMounts

awgreene:
This commit introduces a number of changes that allow OLM users to configure an operator's Volumes and VolumeMounts via its subscription.

### DIFF
--- a/doc/design/subscription-config.md
+++ b/doc/design/subscription-config.md
@@ -1,0 +1,34 @@
+# Subscription Config
+
+## Configuring Operators deployed by OLM
+
+It is possible to configure how OLM deploys an operator via the `config` field in the [subscription](../../pkg/api/apis/operators/subscription_types.go) object.
+
+Currently, the OLM supports the following configurations:
+
+### Env
+
+The `Env` field defines a list of environment variables that must exist in all containers in the `pod` created by OLM.
+
+> Note: Values defined here will overwrite existing environment varaibles of the same name.
+
+### EnvFrom
+
+The `EnvFrom` field defines a list of sources to populate environment variables in the container. 
+The keys defined within a source must be a C_IDENTIFIER. 
+All invalid keys will be reported as an event when the container is starting.
+When a key exists in multiple sources, the value associated with the last source will take precedence.
+
+> Note: Values defined by an Env with a duplicate key will take precedence.
+
+### Volumes
+
+The `Volumes` field defines a list of [volumes](https://kubernetes.io/docs/concepts/storage/volumes/) that must exist on the `pod` created by OLM.
+
+> Note: Volumes defined here will overwrite existing Volumes of the same name.
+
+### VolumeMounts
+
+The `VolumeMounts` field defines a list of [volumeMounts](https://kubernetes.io/docs/concepts/storage/volumes/) that must exist in all containers in the `pod` created by OLM. If a `volumeMount` references a `volume` that does not exist, OLM will fail to deploy the operator.
+
+> Note: VolumeMounts defined here will overwrite existing VolumeMounts of the same name.
--- a/manifests/0000_50_olm_05-subscription.crd.yaml
+++ b/manifests/0000_50_olm_05-subscription.crd.yaml
@@ -182,3 +182,1282 @@ spec:
                             type: boolean
                             description: Specify whether the Secret must be defined
 
+                volumeMounts:
+                  description: List of VolumeMounts to set in the container.
+                  items:
+                    description: VolumeMount describes a mounting of a Volume within
+                      a container.
+                    properties:
+                      mountPath:
+                        description: Path within the container at which the volume
+                          should be mounted.  Must not contain ':'.
+                        type: string
+                      mountPropagation:
+                        description: mountPropagation determines how mounts are
+                          propagated from the host to container and the other way
+                          around. When not set, MountPropagationNone is used. This
+                          field is beta in 1.10.
+                        type: string
+                      name:
+                        description: This must match the Name of a Volume.
+                        type: string
+                      readOnly:
+                        description: Mounted read-only if true, read-write otherwise
+                          (false or unspecified). Defaults to false.
+                        type: boolean
+                      subPath:
+                        description: Path within the volume from which the container's
+                          volume should be mounted. Defaults to "" (volume's root).
+                        type: string
+                      subPathExpr:
+                        description: Expanded path within the volume from which
+                          the container's volume should be mounted. Behaves similarly
+                          to SubPath but environment variable references $(VAR_NAME)
+                          are expanded using the container's environment. Defaults
+                          to "" (volume's root). SubPathExpr and SubPath are mutually
+                          exclusive. This field is beta in 1.15.
+                        type: string
+                    required:
+                      - mountPath
+                      - name
+                    type: object
+                  type: array
+                volumes:
+                  description: List of Volumes to set in the podSpec.
+                  items:
+                    description: Volume represents a named volume in a pod that
+                      may be accessed by any container in the pod.
+                    properties:
+                      awsElasticBlockStore:
+                        description: 'AWSElasticBlockStore represents an AWS Disk
+                                          resource that is attached to a kubelet''s host machine
+                                          and then exposed to the pod. More info: https://kubernetes.io/docs/concepts/storage/volumes#awselasticblockstore'
+                        properties:
+                          fsType:
+                            description: 'Filesystem type of the volume that you
+                                              want to mount. Tip: Ensure that the filesystem type
+                                              is supported by the host operating system. Examples:
+                                              "ext4", "xfs", "ntfs". Implicitly inferred to be "ext4"
+                                              if unspecified. More info: https://kubernetes.io/docs/concepts/storage/volumes#awselasticblockstore
+                                              TODO: how do we prevent errors in the filesystem from
+                                              compromising the machine'
+                            type: string
+                          partition:
+                            description: 'The partition in the volume that you want
+                                              to mount. If omitted, the default is to mount by volume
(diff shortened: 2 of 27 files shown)


awgreene:
Hoping to get the `code / feature` reviewed - I will work on adding documentation and an `E2E` test.

pkg/controller/operators/olm/initializer/inject.go
```diff
@@ -0,0 +1,151 @@
...
+
+	for i := range podSpec.Containers {
+		container := &podSpec.Containers[i]
+		container.Env = merge(container.Env, envVars)
+	}
+
+	return nil
+}
+
+func merge(containerEnvVars []corev1.EnvVar, newEnvVars []corev1.EnvVar) (merged []corev1.EnvVar) {
```
awgreene: I will make sure to rename this to `mergeEnvVars`.

pkg/controller/operators/olm/initializer/inject.go
```diff
@@ -0,0 +1,151 @@
...
+		merged = append(merged, corev1.EnvVar{
+			Name:  newEnvVar.Name,
+			Value: newEnvVar.Value,
+		})
+	}
+
+	return
+}
+
+func find(proxyEnvVar []corev1.EnvVar, name string) (envVar *corev1.EnvVar, found bool) {
```
awgreene: I will make sure to rename this to `findEnvVars`.

doc/design/subscription-config.md
```diff
@@ -0,0 +1,50 @@
+# Subscription Config
+
+## Configuring Operators deployed by OLM
+
+It is possible to configure how OLM deploys an operator via the `config` field in the [subscription](../../pkg/api/apis/operators/subscription_types.go) object.
+
+Currently, the `SubscriptionConfig` object allows users to configure the following features on Deployments created by OLM:
```
ecordell: nit: maybe just `the Subscription Config ` so that it doesn't seem like SubscriptionConfig is another `Kind`?

doc/design/subscription-config.md
```diff
@@ -0,0 +1,50 @@
...
+
+### Selector
+The `selector` field defines a list of `Label` selectors for pods. Existing ReplicaSets whose pods are selected by this will be the ones affected by this deployment.
```
ecordell: nit: reword the second sentence to remove passive voice?

doc/design/subscription-config.md
```diff
@@ -0,0 +1,50 @@
...
+> Note: It must match the pod template's labels.
+### NodeSelector
```
ecordell: Should we document these before they have been implemented?

pkg/api/apis/operators/v1alpha1/subscription_types.go
```diff
@@ -76,6 +76,14 @@ type SubscriptionConfig struct {
 	// +patchStrategy=merge
 	// +optional
 	Env []corev1.EnvVar `json:"env,omitempty" patchMergeKey:"name" patchStrategy:"merge"`
+	
+	// List of Volumes to set in the podSpec.
+	// +optional
+	Volumes []corev1.Volume `json:"volumes,omitempty"`
```
ecordell: Has the CRD schema been updated to match?

pkg/controller/operators/olm/initializer/config.go
```diff
@@ -16,7 +16,7 @@ type operatorConfig struct {
 	logger *logrus.Logger
 }
 
-func (o *operatorConfig) GetOperatorConfig(ownerCSV ownerutil.Owner) (overrides []corev1.EnvVar, err error) {
+func (o *operatorConfig) GetOperatorConfig(ownerCSV ownerutil.Owner) (overrides []corev1.EnvVar, volumes []corev1.Volume, volumeMounts []corev1.VolumeMount, err error) {
```
ecordell: since we know we need to expand the config in the future, we could return some struct `overrides *ConfigOverrides`

pkg/controller/operators/olm/initializer/inject.go
```diff
@@ -0,0 +1,151 @@
...
+			envVar = &volumes[i]
+
+			break
+		}
+	}
+
+	return
+}
+
+// InjectVolumeMountsIntoDeployment injects the provides VolumeMounts
```
ecordell: nit: `provided`

pkg/controller/operators/olm/initializer/inject.go
```diff
@@ -0,0 +1,151 @@
...
+// into the given PodSpec.
+//
+// If the PodSpec already defines a VolumeMount of the same name
+// as any of the provided VolumeMounts then it will be overwritten.
+func InjectVolumeMountsIntoDeployment(podSpec *corev1.PodSpec, volumeMounts []corev1.VolumeMount) error {
```
ecordell: Should we check that that `Volume`s referenced by the `volumeMounts` exist?

pkg/controller/operators/olm/initializer/config.go
awgreene: That's a good idea, I'll also update the variables returned to `envVarOverrides`, `volumeOverrides`, and `volumeMountOverrides`.

doc/design/subscription-config.md
awgreene: Good point - I had only added them because they are defined in the [subscription_types.go](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/api/apis/operators/subscription_types.go) file.

awgreene: I will update this doc to only include:

- EnvFrom
- Env
- Volumes
- VolumeMounts

It does not appear that any of the other fields are supported yet.

pkg/controller/operators/olm/initializer/inject.go
awgreene: Sure thing!

awgreene: @ecordell and I discussed this offline and have decided not to include this feature in this PR. Instead, we are going to investigated validating the subscription upfront with an admission webhook for Subscriptions

pkg/api/apis/operators/v1alpha1/subscription_types.go
awgreene: The CRD will be updated in an upcoming change to this PR.

doc/design/subscription-config.md
awgreene: Resolved by https://github.com/operator-framework/operator-lifecycle-manager/pull/1058#discussion_r332242758

pkg/api/apis/operators/v1alpha1/subscription_types.go
awgreene: The CRD will be updated in the next version of the PR.

awgreene:
@ecordell existing comments have been addressed.

manifests/0000_50_olm_05-subscription.crd.yaml
```diff
@@ -9,33 +9,33 @@ spec:
   group: operators.coreos.com
   version: v1alpha1
   versions:
-  - name: v1alpha1
-    served: true
-    storage: true
+    - name: v1alpha1
```
dinhxuanvu: Something doesn't look right with this file with all extra spaces at the start of these lines. They shouldn't be there.

manifests/0000_50_olm_05-subscription.crd.yaml
```diff
@@ -181,4 +181,1282 @@ spec:
...
+                  type: array
+                volumes:
+                  description: List of Volumes to set in the podSpec.
+                  items:
+                    description: Volume represents a named volume in a pod that
+                      may be accessed by any container in the pod.
+                    properties:
+                      awsElasticBlockStore:
+                        description: 'AWSElasticBlockStore represents an AWS Disk
+                                          resource that is attached to a kubelet''s host machine
```
dinhxuanvu: YAML lint is complaining this is not valid YAML file due to double 'single quote' on this line.

awgreene: The error you mentioned above seems to be specific to the linter. Ultimately, the CRD is accepted by k8s so this shouldn't be an issue.

doc/design/subscription-config.md
```diff
@@ -0,0 +1,30 @@
+# Subscription Config
+
+## Configuring Operators deployed by OLM
+
+It is possible to configure how OLM deploys an operator via the `config` field in the [subscription](../../pkg/api/apis/operators/subscription_types.go) object.
+
+Currently, the `Subscription` object has a `config` field that allows users to configure the following features on Deployments created by OLM:
+
+### EnvFrom
```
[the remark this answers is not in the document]
awgreene: Sure thing!

pkg/controller/operators/olm/overrides/inject.go
```diff
@@ -0,0 +1,151 @@
...
+
+	return nil
+}
+
+func mergeVolumes(podSpecVolumes []corev1.Volume, newVolumes []corev1.Volume) (merged []corev1.Volume) {
+	merged = podSpecVolumes
+
+	for _, newVolume := range newVolumes {
+		existing, found := findVolume(podSpecVolumes, newVolume.Name)
+		if found {
```
jpeeler: I'm wondering with such PodPreset-ish behavior why overwriting was chosen instead of bailing out? Just curious - https://github.com/kubernetes/kubernetes/blob/4d6deea24179a72f36f252664937d2585598ee30/plugin/pkg/admission/podpreset/admission.go#L394.

jpeeler: Yeah, maybe it's better to support the feature of overwriting.

PR #1059, opened 2019-10-04
Bug 1758008: Manual approval strategy ignored for subsequent releases

dinhxuanvu:
For an subscription with manual approval, only the initial installplan
respects the approval requires. Subsequent installplans ignore the
fact the approval status is still false and install the operator
regardless. This issue is due to catalog incorrectly transitions
the installplan into install phase without checking for approval
condition.

**Reviewer Checklist**


dinhxuanvu:
PTAL

ecordell:
@dinhxuanvu The fix looks good - can you write a test for this so we don't regress?

ecordell:
@dinhxuanvu Can this be triggered with an e2e test similar to the ones testing the OperatorGroup serviceaccount scope?

i.e. create a manual installplan that would fail because of scope, then give the operatorgroup enough permission to install it, and then see it install anyway (even though it's manual)?

dinhxuanvu:
Test case included. @ecordell Ready for final review.

PR #1069, opened 2019-10-10
Proposal: RelatedImages

ecordell:
This is a resurrection of #946 

We need to be able to identify related images for mirroring into disconnected environments.


ecordell:
Verbal approval from @shawn-hurley and I want to reference this doc, so merging now and if we need to change something we can make a new PR.

rcernich:
Not sure how to comment on the proposal, so adding my two cents here...

Would it be better to put this information into a separate file in the metadata bundle?  I think this would be a simpler solution than modifying the csv file every time an image used within it is rebuilt.

AndrienkoAleksandr:
Hello, @ecordell . Sorry, was this proposal implemented about olm.relatedImage in the annotation? I see it was marked like future work...

AndrienkoAleksandr:
@ecordell https://github.com/operator-framework/operator-lifecycle-manager/commit/840a709774ca36dcf3cf70df4384231af80650cc#diff-4ada87752d32e1f3f585474c722d2633R44 we have short digest, because annotation has limitation in 63 characters and such long image can't be applied: docker.io/eclipse/che-theia-endpoint-runtime-binary@sha256:14041e6e51f47248e1018a2dacc8e70acad08f2a2dd7dc67c83a1b43d2cb73b7 in a full form?

PR #1073, opened 2019-10-12
Lock down package server runtime environment

javanthropus:
This PR modifes the package server component's deployment manifest's pods to:

* Run as non-root (uid 1000) by default
* Mount a emptyDir volume to /tmp so that the root filesystem can be read-only

At Indeed we're using restrictive pod security policies that among other things prevent pods running as root and marks them to have read-only root filesystems.

**Reviewer Checklist**

### DIFF
--- a/deploy/chart/templates/_packageserver.deployment-spec.yaml
+++ b/deploy/chart/templates/_packageserver.deployment-spec.yaml
@@ -60,5 +60,15 @@ spec:
         {{- if .Values.package.resources }}
         resources:
 {{ toYaml .Values.package.resources | indent 10 }}
-        {{- end}}
+        {{- end }}
+        {{- if .Values.package.securityContext }}
+        securityContext:
+          runAsUser: {{ .Values.package.securityContext.runAsUser }}
+        {{- end }}
+        volumeMounts:
+        - name: tmpfs
+          mountPath: /tmp
+      volumes:
+      - name: tmpfs
+        emptyDir: {}
 {{- end -}}
--- a/deploy/upstream/values.yaml
+++ b/deploy/upstream/values.yaml
@@ -27,5 +27,7 @@ package:
     pullPolicy: Always
   service:
     internalPort: 5443
+  securityContext:
+    runAsUser: 1000
 catalog_sources:
 - rh-operators
--- a/manifests/0000_50_olm_15-packageserver.clusterserviceversion.yaml
+++ b/manifests/0000_50_olm_15-packageserver.clusterserviceversion.yaml
@@ -123,6 +123,12 @@ spec:
                   requests:
                     cpu: 10m
                     memory: 50Mi
+                volumeMounts:
+                - name: tmpfs
+                  mountPath: /tmp
+              volumes:
+              - name: tmpfs
+                emptyDir: {}
   maturity: alpha
   version: 0.14.1
   apiservicedefinitions:


deploy/chart/templates/_packageserver.deployment-spec.yaml
```diff
@@ -60,5 +60,13 @@ spec:
         {{- if .Values.package.resources }}
         resources:
 {{ toYaml .Values.package.resources | indent 10 }}
-        {{- end}}
+        {{- end -}}
+        securityContext:
```
ecordell: Could you make this block conditional on the presence of `securityContext` in values.yaml?

javanthropus: I can do that, but do you also want me to remove the default `runAsUser` setting in `values.yaml`?  If I don't, `securityContext` will always be set, and the conditional will be a bit meaningless.

ecordell:
Thanks @javanthropus ! 

/lgtm

ecordell:
/hold

Sorry, I thought you had removed the default runAsUser. Could you remove that as well?

If you can explain more about your deploy target we can maybe find a better place to put those values?

javanthropus:
The clusters we're building have a default pod security policy applied such that pods are not allowed to run as root and their root filesystems are mounted read-only.  Another way to address this problem would have been to apply a pod security policy and corresponding role and role binding for packageserver to use; however, since pod security policies are still a beta resource, I figured that locking down packageserver was the safer bet.  Plus, running it this way is a bit safer should there be a vulnerability in it.

javanthropus:
I've been doing more testing against our clusters, and it actually looks like this change may not be necessary in the first place.  The package server pods run with the `olm-operator-serviceaccount` service account, and that service account is basically granted full admin access over the cluster, which allows it to use all available pod security policies.

At this point, I'm not sure anymore what was broken in my test cluster that made me think I needed to make this change.  I'm going to do more testing, and I'll close this PR if it really proves to be unnecessary, unless you would like to keep the changes for the security benefits alone.

javanthropus:
I tested things out with the stock 0.12.0 release on a fresh cluster, and none of my changes were necessary to work around the pod security policies.  Sadly, our most permissive policy is used, but that's a result of how the policy admission controller works and the broad permissions granted to the service account used by packageserver.

I would still like to see this MR accepted since it still helps reduce the attack surface of packageserver.  If you agree, let me know where to move that default `runAsUser` setting; otherwise, please feel free to close this out.

deploy/chart/values.yaml
```diff
@@ -51,3 +51,5 @@ package:
     requests:
      cpu: 10m
      memory: 50Mi
+  securityContext:
```
ecordell: Can you move this to `deploy/upstream/values.yaml`? Then we should be good to merge. Thanks!

javanthropus: Done!

njhale:
@javanthropus Sorry for taking so long to get back to you on this. I think this is a good thing to add regardless of whether it was causing the reported issues.

Before merging this, could I ask you to rebase, run `make release`, and add only the changes to the `./manifests` directory in a new commit? This way, our CI can actually deploy and test packageserver with this patch in effect.

Holding for now, since we're currently only accepting doc and bugfix PRs

/hold

javanthropus:
FYI, the `make release` process is incompatible with helm v3.  Once I remembered that I had that installed, I was able to downgrade to v2 in order to get things working. :-)

javanthropus:
Please stop the automatic re-tries.  The failing tests are clearly not going to pass without manual intervention.  These failure notices are becoming a problem.

PR #1075, opened 2019-10-14
Bug 1760608: Add resource limits to all pods

awgreene:
Introduces resource limits to prevent OLM from causing a memory leak.

### DIFF
--- a/deploy/chart/values.yaml
+++ b/deploy/chart/values.yaml
@@ -19,9 +19,12 @@ olm:
   nodeSelector:
     beta.kubernetes.io/os: linux
   resources:
+    limits:
+      cpu: 20m
+      memory: 320Mi
     requests:
-     cpu: 10m
-     memory: 160Mi
+      cpu: 10m
+      memory: 160Mi
 
 catalog:
   replicaCount: 1
@@ -34,9 +37,12 @@ catalog:
   nodeSelector:
     beta.kubernetes.io/os: linux
   resources:
+    limits:
+      cpu: 20m
+      memory: 160Mi
     requests:
-     cpu: 10m
-     memory: 80Mi
+      cpu: 10m
+      memory: 80Mi
 
 package:
   replicaCount: 2
@@ -48,6 +54,9 @@ package:
   nodeSelector:
     beta.kubernetes.io/os: linux
   resources:
+    limits:
+      cpu: 20m
+      memory: 100Mi
     requests:
      cpu: 10m
      memory: 50Mi
--- a/deploy/ocp/manifests/0.12.0/0000_50_olm_07-olm-operator.deployment.yaml
+++ b/deploy/ocp/manifests/0.12.0/0000_50_olm_07-olm-operator.deployment.yaml
@@ -64,6 +64,9 @@ spec:
           - name: OPERATOR_NAME
             value: olm-operator
           resources:
+            limits:
+              cpu: 20m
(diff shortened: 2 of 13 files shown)


ecordell:
I'd like to collect a bit of data before we commit to these memory numbers - since we know we build a large cache in some cases, we want to be careful.

deploy/chart/values.yaml
```diff
@@ -19,9 +19,12 @@ olm:
   nodeSelector:
     beta.kubernetes.io/os: linux
   resources:
+    limits:
+      cpu: 20m
```
njhale: 1m = 1/1000 CPU core. Are we sure that our operators top out under 1/50th of a CPU core?

Denton24646:
I have put together some slides showing memory and CPU usage for all OLM namespace pods (catalog operator, OLM operator, package server) after conducting a small load test on a CI cluster running 4.3 OpenShift. The data is coming from Prometheus.  

https://docs.google.com/presentation/d/1qSB50bkZJaRbPAKv2aIm-26mRyivMYcaUvOxZ3z_sZo/edit?usp=sharing

I installed 10 operators at once at 45, then deleted them and reinstalled them at 50. The first 6 slides show the close up usage over 15 minutes, the last 6 show the usage over the full two hours. 
There are some interesting takeaways here:

- CPU pops up after the installation but the settles back down to steady levels afterwards
- Memory trends upwards for both catalog and OLM pods, gapping upwards upon installation
- The statistics don't reflect memory being freed (Prometheus metrics show both virtual and resident memory in use, ie all memory) 
- The package server does not increases resource usage as a result of installation

Based on the numbers and these findings, I'm willing to propose the following limits based on the usage per operator per namespace. We should keep in mind that memory and CPU limits behave differently: for CPU, the container is throttled to use that amount of CPU only (it can't use more) but there is no hard stop if it goes over. For memory, if the pod is using consistently more memory than the limit, it may be OOMKilled. Since the limits will differ than the requests these pods will be in the `Burstable` QoS class. 

Here are some variables used in the equation to find the limits:
S - scale up factor. For the load test we installed 10 operators over 20 namespaces (=200). For a high performance cluster we may see installations of 40 operators over 50 namespaces (=2000). So S will be equal to 10. 
K - safety factor. This is a fudge factor that get the limits up higher. K is equal to 2 for CPU limits and 5 for memory limits. 
C - constant factor. This is the usage under constant terms (no load) and will be added on a per pod basis. 

Limit = (Resource utilization increase * S) * K + C

Catalog Operator Pod
- CPU went up 6m so limit equals (6 * 10) * 2 + 5 = 125m
- Memory went up 4M so limit equals (4 * 10) * 5 + 70 = 270M 

OLM Operator Pod
- CPU went up 15m so limit equals (15 * 10) * 2 + 5 = 305m
- Memory went up 4M so limit equals (4 * 10) * 5 + 100 = 300M 

Marketplace
- CPU Limit = 20m
- Memory Limit = 100M

njhale:
These seem pretty reasonable for OCP. I would round up to the nearest hundred (e.g. 125 -> 200). I think we may want to have lower or no limits for upstream manifests since local clusters may be tight on memory to begin with.

Denton24646:
Sounds good - I'm going to open a new PR to implement these limits and keep things simple.

Denton24646:
Continued in #1142

PR #1085, opened 2019-10-25
Fix log message in resolver

ecordell:
This was masking the actual errors coming from the step generator

Fixes an issue with the logs - actual errors were being masked instead of wrapped.

Debugging a failing operator step generation.

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/registry/resolver/resolver.go
+++ b/pkg/controller/registry/resolver/resolver.go
@@ -96,7 +96,7 @@ func (r *OperatorsV1alpha1Resolver) ResolveSteps(namespace string, sourceQuerier
 		if op.Bundle() != nil {
 			bundleSteps, err := NewStepResourceFromBundle(op.Bundle(), namespace, op.Replaces(), op.SourceInfo().Catalog.Name, op.SourceInfo().Catalog.Namespace)
 			if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/controller/registry/resolver/resolver.go
```diff
@@ -96,7 +96,7 @@ func (r *OperatorsV1alpha1Resolver) ResolveSteps(namespace string, sourceQuerier
 		if op.Bundle() != nil {
 			bundleSteps, err := NewStepResourceFromBundle(op.Bundle(), namespace, op.Replaces(), op.SourceInfo().Catalog.Name, op.SourceInfo().Catalog.Namespace)
 			if err != nil {
-				return nil, nil, fmt.Errorf("failed to turn bundle into steps")
+				return nil, nil, fmt.Errorf("failed to turn bundle into steps %s", err.Error())
```
dinhxuanvu: It would be nice to add a colon to separate the generic error message from actual error message: `return nil, nil, fmt.Errorf("failed to turn bundle into steps: %s", err.Error())`.

ecordell: done

kevinrizza:
lgtm

PR #1091, opened 2019-10-28
Bug 1763749: Prioritize APIs from same CatSrc

awgreene:
Problem: When OLM installs an operator that requires dependencies that
are provided via multiple operators in different CatalogSources, the API
is pulled from any of the CatalogSources that provide the API.

Solution: This commit introduces a change so that OLM will generate a
list of operators that depend on the API, randomly select one of their
sources, and prioritize checking in that CatalogSource for the API prior
to checking the remaining CatalogSource for the API.

**This commit is different than the 4.3 fix because of a method signature in the e2e test.**

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/registry/resolver/evolver.go
+++ b/pkg/controller/registry/resolver/evolver.go
@@ -112,8 +112,15 @@ func (e *NamespaceGenerationEvolver) queryForRequiredAPIs() error {
 		}
 		e.gen.MarkAPIChecked(*api)
 
+		// identify the initialSource
+		initialSource := CatalogKey{}
+		for _, operator := range e.gen.MissingAPIs()[*api] {
+			initialSource = operator.SourceInfo().Catalog
+			break
+		}
+
 		// attempt to find a bundle that provides that api
-		if bundle, key, err := e.querier.FindProvider(*api); err == nil {
+		if bundle, key, err := e.querier.FindProvider(*api, initialSource); err == nil {
 			// add a bundle that provides the api to the generation
 			o, err := NewOperatorFromBundle(bundle, "", "", *key)
 			if err != nil {
--- a/pkg/controller/registry/resolver/querier_test.go
+++ b/pkg/controller/registry/resolver/querier_test.go
@@ -106,13 +106,25 @@ func TestNamespaceSourceQuerier_Queryable(t *testing.T) {
 
 func TestNamespaceSourceQuerier_FindProvider(t *testing.T) {
 	fakeSource := fakes.FakeInterface{}
+	fakeSource2 := fakes.FakeInterface{}
 	sources := map[CatalogKey]client.Interface{
 		CatalogKey{"test", "ns"}: &fakeSource,
+		CatalogKey{"test2", "ns"}: &fakeSource2,
 	}
 
 	bundle := opregistry.NewBundle("test", "testPkg", "testChannel")
+	bundle2 := opregistry.NewBundle("test2", "testPkg2", "testChannel2")
 	fakeSource.GetBundleThatProvidesStub = func(ctx context.Context, group, version, kind string) (*opregistry.Bundle, error) {
+		if group != "group" || version != "version" || kind != "kind" {
+			return nil, fmt.Errorf("Not Found")
+		}
 		return bundle, nil
+	}
+	fakeSource2.GetBundleThatProvidesStub = func(ctx context.Context, group, version, kind string) (*opregistry.Bundle, error) {
+		if group != "group2" || version != "version2" || kind != "kind2" {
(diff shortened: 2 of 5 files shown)


pkg/controller/registry/resolver/evolver.go
```diff
@@ -112,8 +112,15 @@ func (e *NamespaceGenerationEvolver) queryForRequiredAPIs() error {
 		}
 		e.gen.MarkAPIChecked(*api)
 
+		// identify the initialSource
+		initialSource := CatalogKey{}
+		for _, operator := range e.gen.MissingAPIs()[*api] {
+			initialSource = operator.SourceInfo().Catalog
```
njhale: Correct me if I'm wrong, but doesn't this break down when there exists two or more concurrent `Subscriptions` that require the same API, but target different sources that contain operators providing that API?

pkg/controller/registry/resolver/querier_test.go
```diff
@@ -151,23 +165,52 @@ func TestNamespaceSourceQuerier_FindProvider(t *testing.T) {
...
 				api: opregistry.APIKey{"group", "version", "kind", "plural"},
+				catalogKey: CatalogKey{},
 			},
 			out: out{
 				bundle: nil,
 				key:    nil,
 				err:    fmt.Errorf("group/version/kind (plural) not provided by a package in any CatalogSource"),
 			},
 		},
+		{
```
njhale: nit: It would be nice to add descriptive names to each test case.

pkg/controller/registry/resolver/querier_test.go
```diff
@@ -151,23 +165,52 @@ func TestNamespaceSourceQuerier_FindProvider(t *testing.T) {
...
+			args: args{
+				api: opregistry.APIKey{"group2", "version2", "kind2", "plural2"},
+				catalogKey: CatalogKey{Name: "test3", Namespace: "ns"},
+			},
+			out: out{
+				bundle: bundle2,
+				key:    &CatalogKey{Name: "test2", Namespace: "ns"},
+				err:    nil,
+			},
+		},
```
njhale: I think we should have a test to ensure that the correct bundle is returned when more than one source containing a provider is given along with the preferred source key.

test/e2e/subscription_e2e_test.go
```diff
@@ -1209,6 +1209,47 @@ func TestCreateNewSubscriptionWithPodConfig(t *testing.T) {
...
+		defer duplicateCatSrcCleanup()
+
+		// Ensure that the catalog source is resolved before we create a subscription.
+		_, err = fetchCatalogSource(t, crClient, duplicateCatsrc.GetName(), testNamespace, catalogSourceRegistryPodSynced)
+		require.NoError(t, err)
+	}
+
+	// Create a subscription that has a dependency
+	subscriptionName := genName("podconfig-sub-")
+	cleanupSubscription := createSubscriptionForCatalogWithSpec(t, crClient, testNamespace, subscriptionName, subSpec)
```
njhale: Could we add concurrent `Subscriptions` pointing to each catalog to test the aforementioned "break down"?

pkg/controller/registry/resolver/evolver.go
awgreene: You are correct that the current approach can introduce some randomness when the following conditions are met:

* `Operator A` defines dependencies on APIs provided by`Operator B` and `Operator C`
* `Operator B` and `Operator C` come from different `CatalogSources`
* Both `Operator B` and `Operator C` define dependencies on the same API provided by `Operator D`.

In this situation - the existing solution does not guarantee that a specific `CatalogSource` will be prioritized consistently.

This is a known issue with the short-term solution and is expected to be addressed in a future solution. We are comfortable with supporting the existing approach due to the fact that this "bug" only hits a small number of use cases.

pkg/controller/registry/resolver/querier_test.go
awgreene: Will do.

pkg/controller/registry/resolver/querier_test.go
awgreene: This is covered in the E2E testcase. I can create a Unit test as well.

njhale: It's probably best to add it to `master`.

ecordell:
/lgtm
/approve

we need this in the next 4.1, comments can be addressed in a followup in master

crawford:
AWS throttled our API requests.

/retest

crawford:
> error simulating policy: SerializationError: failed to decode query XML error response\n\tstatus code: 500, request id: \ncaused by: EOF

Cool.

/retest

PR #1097, opened 2019-10-29
doc: add internal image proposal

robszumski:
This PR adds a proposal for marking a subset of CRDs as "internal", meaning they are not for end-user consumption or manipulation.

Many Operators have internal concepts that are confusing to end-users when they show up in user interfaces or CLIs like operatorhub.io. This proposal solidifies this concept into a standard convention that others can follow to hide these CRDs and downstream user interfaces by hide them if deemed necessary for their user base.


joelanford:
@robszumski Along these same lines, there is also a distinction between workload/application CRDs and data-only CRDs. Such a distinction would make running the SDK scorecard more reliable and may be helpful metadata for a user. 

Thoughts on that? And if you agree, would it make sense to include this concept in this proposal as well?

sbose78:
+1 !

We were looking to see if there's a good way to hide some of the CRDs which the users don't need to , on the UI ( and other places where we want to hide out details/abstractions ) :

This sounds like a good idea.

doc/contributors/design-proposals/internal-images.md
```diff
@@ -0,0 +1,51 @@
...
+
+The behavior is straightforward, when a CRD is managed as part of the Operator installation, it can be marked with an annotation which is available for downstream tools to read and hide the CRD where applicable. This should be backwards compatible as a no-op, so it can be considered progressive enhancement.
+
+```
+kind: CustomResourceDefinition
+apiVersion: apiextensions.k8s.io/v1beta1
+metadata:
+  name: hivetables.metering.openshift.io
+  annotations:
+    operators.coreos.com/internal-object: true
```
sbose78: Let's make it `apps.kubernetes.io/internal-object:true` ? :)

sbose78:
That'll be good, @robszumski .

robszumski:
PTAL again, updated title, filename and added the new annotations.

ecordell:
Unfortunately, we can't use the k8s namespace without upstream approval. 

Can we use `operators.operatorframework.io` prefix like we do for our bundles? 

`operators.operatorframework.io.bundle.mediatype.v1: "registry+v1"` for example

doc/contributors/design-proposals/internal-objects.md
```diff
@@ -0,0 +1,55 @@
...
+
+The behavior is straightforward, when a CRD is managed as part of the Operator installation, it can be marked with an annotation which is available for downstream tools to read and hide the CRD where applicable. This should be backwards compatible as a no-op, so it can be considered progressive enhancement.
+
+```
+kind: CustomResourceDefinition
+apiVersion: apiextensions.k8s.io/v1beta1
+metadata:
+  name: hivetables.metering.openshift.io
+  annotations:
+    apps.kubernetes.io/internal-object:true
```
shawn-hurley: apps.operatorframework.io?

bbrowning:
Would this annotation be appropriate to hide things that should not show up to end-users in things like the Developer Catalog but instead can only be created by cluster admins? Or will this annotation also hide it from cluster admins in screens outside of just the Developer Catalog?

In my case, I want admins to see the option to create a `KnativeServing` CR from their Installed Operators screen. But I don't want developers to be offered to create a `KnativeServing` CR from their Developer Catalog. We have RBAC that limits who can create them, but last I checked the Dev Catalog doesn't inspect the RBAC rules for CRs when deciding what to show.

christianvogt:
@robszumski what about supporting these annotations on the CSV. Looks like the console will have an issue:
https://github.com/openshift/console/pull/3170#issuecomment-553071473

robszumski:
@christianvogt Are you thinking it goes on to `spec.customresourcedefinitions.owned`? No issues on that for me other than it's now a deeper OLM spec change not an optional field to be read (or not).

spadgett:
The problem here for console is that normal users can't list CRDs, so there's no way to console read the annotation. We currently look at `spec.customresourcedefinitions.owned` in the CSV. So this would either need to be a new field there or perhaps an annotation on the CSV itself.

christianvogt:
@sbose78 had suggested an annotation on the CSV which was originally a whitelist array of CRDs kinds I believe. A agree though that a blacklist approach is better.

sbose78:
My initial suggestion was to have  :

```
kind: ClusterServiceVersion
metadata:
  annotations:
    apps.operatorframework.io/primary-objects: [....] 
```

- This avoids the need to make a deeper OLM CSV schema change
- The typical RBAC design for consumption of OLM lets non-admin users read CSVs compared to reading CRDs. Hence, having it on the CSV might help a bigger audience consume this information.
- Downside: Doesn't make it OLM agnostic.

robszumski:
@sbose78 Turning that into a blacklist seems like a good solution to me

christianvogt:
An annotation on CSV whose value is an array of names `spec.owned.customresourcedefinitions.name`.
Blacklist and renamed @sbose78's proposal to align with the previous naming:
`operators.operatorframework.io/internal-objects: [...]`

sbose78:
@christianvogt & @robszumski :

Sounds good.

- the blacklist, looks good to me 👍
-  use of `spec.owned.customresourcedefinitions.name` in the array provides uniqueness, so 👍 .

christianvogt:
@robszumski @sbose78 
Dev console is going to go ahead with aligning on this proposal noted above.

sbose78:
Sounds good.

abhinandan13jan:
FYI https://github.com/openshift/console/pull/3518

abhinandan13jan:
@sbose78 @christianvogt Please let me know if this is to be changed
> @christianvogt & @robszumski :
> 
> Sounds good.
> 
> * the blacklist, looks good to me +1
> * use of `spec.owned.customresourcedefinitions.name` in the array provides uniqueness, so +1 .
> 
> ```
> apiVersion: operators.coreos.com/v1alpha1
> kind: ClusterServiceVersion
> metadata:
>   annotations:
>     apps.operatorframework.io/internal-objects: [....] 
> ```

As part of CR the annotation considered is `operators.operatorframework.io/internal-objects` and not `apps.operatorframework.io/internal-objects` mentioned in the quoted comment.

christianvogt:
@sbose78 @robszumski 
Siince the annotation must be a string. I just want to confirm that this will be a stringified JSON array value: eg `"[\"foo\",\"bar\"]"`

ecordell:
@robszumski do you have time to fix this proposal up to match the decisions made in the comments here?

matzew:
Is this still planed to be able to hide some CRDs ?

ecordell:
@matzew This is already implemented in console

ecordell:
doc-only, merging without tests.

PR #1099, opened 2019-10-30
feat(metrics) Limit Cardinality of CSV metrics

awgreene:
This commit introduces a change that limits the number of metrics that
an OLM cluster reports at any given time for a CSV.

The first metric introduced is called csv_succeeded, which tracks CSVs that
have reached the succeeded phase. The following information is
provided about the CSV via labels: name, version. The value of this
metric will always be 0 or 1.

The second metric introduced is called csv_abnormal, which is reported
whenever the CSV is updated and has not reached the succeeded phase. The
following information is provided about the CSV via labels: name,
version, phase, reason. Whenever a CSV is updated, the existing
timeseries is deleted and replaced by an updated version.

**Reviewer Checklist**

### DIFF
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -5,17 +5,17 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 
+	olmv1alpha1 "github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/clientset/versioned"
 	v1alpha1 "github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/listers/operators/v1alpha1"
-	olmv1alpha1 "github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
-
 )
 
 const (
 	NAME_LABEL      = "name"
 	INSTALLED_LABEL = "installed"
+	NAMESPACE_LABEL = "namespace"
 	VERSION_LABEL   = "version"
-	PHASE_LABEL    = "phase"
+	PHASE_LABEL     = "phase"
 	REASON_LABEL    = "reason"
 )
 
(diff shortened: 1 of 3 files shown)


awgreene:
# Example Cases:


## When a CSV is updated but not in the succeeded phase
```
# HELP csv_abnormal Successful CSV Install
# TYPE csv_abnormal gauge
csv_abnormal{name="etcd-operator.v1.0.0",phase="Failed",reason="UnsupportedOperatorGroup",version="1.0.0"} 1.0
...
...
...
# HELP csv_up Successful CSV Install
# TYPE csv_up gauge
csv_up{name="packageserver",version="1.0.0"} 1.0
csv_up{name="etcd-operator.v1.0.0",version="1.0.0"} 0.0
```

>Note: When a CSV is updated, the old metric is deleted.

## When a CSV reaches the succeeded phase

```
# HELP csv_up Successful CSV Install
# TYPE csv_up gauge
csv_up{name="packageserver",version="1.0.0"} 1.0
csv_up{name="etcd-operator.v1.0.0",version="1.0.0"} 1.0
```

>Note: Notice that the `csv_abnormal` timeseries was removed

pkg/metrics/metrics.go
```diff
@@ -177,6 +185,21 @@ func CounterForSubscription(name, installedCSV string) prometheus.Counter {
...
 
-func EmitCSVMetric(csv *olmv1alpha1.ClusterServiceVersion){
-	csvSyncCounter.WithLabelValues(csv.Name, csv.Spec.Version.String(), string(csv.Status.Phase), string(csv.Status.Reason)).Inc()
+func EmitCSVMetric(oldCSV *olmv1alpha1.ClusterServiceVersion, newCSV *olmv1alpha1.ClusterServiceVersion) {
+	if oldCSV == nil || newCSV == nil {
+		return
+	}
+
+	// Delete the old CSV metrics
+	csvAbnormal.DeleteLabelValues(oldCSV.Name, oldCSV.Spec.Version.String(), string(oldCSV.Status.Phase), string(oldCSV.Status.Reason))
```
ecordell: I think we need to include the namespace here so that we don't conflict if the same operator is installed multiple times.

I think we also want to avoid emitting for `copied` csvs, right?

awgreene: Good points, both are addressed in the latest version of the PR.

ecordell:
/hold 

while I spin up a cluster to look at this :)

awgreene:
> /hold
> 
> while I spin up a cluster to look at this :)

Sure - let me know if you find anything interesting!

ecordell:
Checking CSV succeeded: 
blue is packageserver, which is always good
green is etcd-operator - I killed its deployment to watch it go unhealthy

Screen Shot 2019-10-31 at 8 58 56 AM

This is csv_abnormal - I just see it for `installready`, when olm detected the deployment was deleted and reconciles from the CSV. 
Screen Shot 2019-10-31 at 8 59 16 AM

This looks good, but I did expect to see more values for "abnormal". I think what's happening is that we're deleting the abnormal states from prometheus before prometheus can scrape them. 

Knowing when to delete them is tricky, since the scrape time is configurable. I think it's 20s by default. We may need to do something where we keep the "old" timeseries around for a while before deleting them.

kevinrizza:
> Checking CSV succeeded:
> blue is packageserver, which is always good
> green is etcd-operator - I killed its deployment to watch it go unhealthy
> 
> This is csv_abnormal - I just see it for `installready`, when olm detected the deployment was deleted and reconciles from the CSV.
> 
> This looks good, but I did expect to see more values for "abnormal". I think what's happening is that we're deleting the abnormal states from prometheus before prometheus can scrape them.
> 
> Knowing when to delete them is tricky, since the scrape time is configurable. I think it's 20s by default. We may need to do something where we keep the "old" timeseries around for a while before deleting them.

@ecordell Maybe this is a silly question, but what value do we get from seeing every step here rather than just seeing "it's in a good state" or "it's not in a good state" ? Is there some unique knowledge we would gain from seeing all of those steps? I was thinking that we care more about "what step are we stuck in" in the same way that I can learn that by `kubectl get`ting the CSV

My concern is that being aware of the state of these things and deleting them later seems like it adds a medium amount of complexity to managing the metric endpoint's state for the value that it adds.

ecordell:
That's a good point, and I do think it would be reasonable to merge this as-is and accept that not all states will necessarily get picked up by prometheus.

It would be nice to see the full lifecycle of operators reported if possible though. Perhaps we do that as a follow up.

PR #1101, opened 2019-10-31
feat(resolver): use provided and required apis from grpc

ecordell:

This uses the required/provided apis from the registry api response to do dependency resolution. 

The change itself is fairly small, but the diff is slightly larger because I switched the resolver to work directly with the api objects instead of the bundle abstraction (which is less useful if we don't unpack/cache the contents).

Previously, dependency information was extracted from CSVs that were included in the registry response. With bundle images, we will not always have the CSV available. 

Instead, the registry extracts this information (it was already extracting provided apis) and sends it as part of the bundle response. OLM trusts this response to know which operators to resolve.

**Reviewer Checklist**


go.mod
```diff
@@ -54,6 +52,8 @@ replace (
 	github.com/openshift/api => github.com/openshift/api v3.9.1-0.20190924102528-32369d4db2ad+incompatible
 	github.com/openshift/client-go => github.com/openshift/client-go v0.0.0-20190923180330-3b6373338c9b
 
+	github.com/operator-framework/operator-registry => ../operator-registry
```
jpeeler: I assume this was for dev only.

jpeeler: It's back!

ecordell:
This is passing all tests but gcp, which seems to broken everywhere. Merging manually.

PR #1102, opened 2019-10-31
WIP: Webhook Admissions

awgreene:

**Reviewer Checklist**


pkg/api/apis/operators/v1alpha1/clusterserviceversion_types.go
```diff
@@ -147,6 +162,7 @@ type ClusterServiceVersionSpec struct {
 	Maturity                  string                    `json:"maturity,omitempty"`
 	CustomResourceDefinitions CustomResourceDefinitions `json:"customresourcedefinitions,omitempty"`
 	APIServiceDefinitions     APIServiceDefinitions     `json:"apiservicedefinitions,omitempty"`
+	WebhookDefinitions        WebhookDefinitions        `json:"webhookdefinitions,omitempty"`
```
ecordell: I think this can just be a `[]WebhookDefinition` - not sure the "owned" distinction is useful here.

pkg/lib/operatorclient/client.go
```diff
@@ -127,6 +129,18 @@ type DeploymentClient interface {
 	ListDeploymentsWithLabels(namespace string, labels labels.Set) (*appsv1.DeploymentList, error)
 }
 
+type WebhookClient interface {
```
ecordell: don't block on this, but we've been moving away from operatorclient and towards the standard kube client, so this isn't really needed here.

pkg/api/apis/operators/v1alpha1/clusterserviceversion_types.go
awgreene: I agree - I'll get around to that next.

awgreene: The `owned`  field has been removed.

pkg/lib/operatorclient/client.go
awgreene: Updated the code to use the kube client.

ecordell:
closing in favor of #1436

PR #1103, opened 2019-10-31
[WIP] feat: add CatalogSource polling interval for new catalog sources

Denton24646:
Add a Poll Interval to the Spec of Catalog Sources in order to enable the catalog operator to automatically check for new catalog source versions. Part of https://github.com/openshift/enhancements/blob/master/enhancements/olm/operator-registry.md

Have the catalog operator be responsible keeping catalog sources up to date via the container runtime. 

**Reviewer Checklist**


pkg/controller/registry/reconciler/grpc.go
```diff
@@ -168,6 +169,78 @@ func (c *GrpcRegistryReconciler) ensureService(source grpcCatalogSourceDecorator
...
+	interval := source.Spec.Poll.Interval.Duration
+
+	// check poll value is not zero (default poll value)
+	// if polling interval is zero polling will not be done
+	if interval == time.Duration(0) {
+		return false
+	}
+
+	// check digest every poll interval
+	if time.Now().After(lastCheckedTimestamp.Add(interval)) &&
```
dinhxuanvu: If we keep compare the polling time against creation time without anyway to keep track when is the last time we check for the update, then this if statement will return `true` after creation-time + poll-interval. I would add a new field to `CatalogSource` to keep track of `LastUpdateTime` and also a func to determine if the time is up for an update attempt based on Poll. Essentially, that function will trigger this `updateRegistryPodByDigest` as a part of sync `CatalogSource` process that is controlled by catalog operator. See my [PR](https://github.com/Denton24646/operator-lifecycle-manager/pull/1/).

pkg/controller/registry/reconciler/grpc.go
```diff
@@ -168,6 +169,78 @@ func (c *GrpcRegistryReconciler) ensureService(source grpcCatalogSourceDecorator
...
+		return
+	}
+
+	var newCatalogSourceImage string
+	// check pod to get container details
+	// we are only interested in the container image id, to see if it matches the old version
+	for i := 0; i < 10; i++ {
+		newCatalogSourceImage = pod.Status.ContainerStatuses[0].ImageID
+		if newCatalogSourceImage == "" {
+			time.Sleep(10 * time.Second)
```
dinhxuanvu: Even though this asynchronous behavior is fine, I think there is potentially a blocker + some weird behavior of go routine when is in combo with `CatalogSource` sync process in catalog operator. We should just simply skip this poll cycle and move on. Given there is a log message below to report this, we should be fine in term of keeping a record of this skipping cycle. See my [PR](https://github.com/Denton24646/operator-lifecycle-manager/pull/1/).

pkg/controller/registry/reconciler/grpc.go
```diff
@@ -168,6 +169,78 @@ func (c *GrpcRegistryReconciler) ensureService(source grpcCatalogSourceDecorator
...
+	if err != nil {
+		logrus.WithField("pod", "catalog-source-pod").Warn("couldn't create new catalog source pod")
+		return
+	}
+
+	var newCatalogSourceImage string
+	// check pod to get container details
+	// we are only interested in the container image id, to see if it matches the old version
+	for i := 0; i < 10; i++ {
+		newCatalogSourceImage = pod.Status.ContainerStatuses[0].ImageID
```
dinhxuanvu: I think we should check `pod.Status.ContainerStatuses` to ensure it is not `nil` given `ContainerStatuses` is actually `omitempty`. See my [PR](https://github.com/Denton24646/operator-lifecycle-manager/pull/1/)

dinhxuanvu:
You probably need to run some code-gen to ensure this PR passing CI but we can deal with it after the code is completed.

pkg/controller/registry/reconciler/grpc.go
```diff
@@ -168,6 +169,78 @@ func (c *GrpcRegistryReconciler) ensureService(source grpcCatalogSourceDecorator
...
+	}
+
+	// check digest every poll interval
+	if time.Now().After(lastCheckedTimestamp.Add(interval)) &&
+		source.CreationTimestamp.Add(interval).Before(time.Now()) {
+		lastCheckedTimestamp = time.Now()
+		go c.getPodDigest(source, podDigestChan)
+	}
+
+	select {
```
ecordell: I don't think this will have the desired effect - select won't block and wait; there's a default case that will immediately return (and once this function returns, nothing is reading from the channel any more).

Denton24646: Yes, I see what you mean - my original idea was since `getPodDigest` can potentially take a long time this function would exit and on the next reconcile loop check the channel again for a value. But since I declare the channel within the scope of the function its a local variable that gets GC'ed. I think if I declare the `podDigestChan` at a higher level this idea makes more sense.

Denton24646: I'm thinking about adding that channel to the `GrpcRegistryReconciler` struct itself

pkg/controller/registry/reconciler/grpc.go
```diff
@@ -168,6 +168,62 @@ func (c *GrpcRegistryReconciler) ensureService(source grpcCatalogSourceDecorator
...
+		return false
+	}
+}
+
+// getPodDigest is an internal method that creates a pod using the latest catalog source.
+// Once the pod comes up, it puts the container image digest onto a channel.
+func (c *GrpcRegistryReconciler) getPodDigest(source grpcCatalogSourceDecorator, podChan chan string) {
+	// remove label from pod to ensure service does accidentally route traffic to the pod
+	p := source.Pod()
+	p.Labels[CatalogSourceLabelKey] = ""
```
Denton24646: When we remove the labels here, to prevent the service routing to this pod, we also make the pod deletion in `ensurePod` not pick up this new pod - it looks at the labels to figure out which pod to delete. So we would need an additional label or mechanism to ensure these pods get deleted properly

ecordell:
@Denton24646 @dinhxuanvu Where can I see the manual test scripts / images you were using?

Denton24646:
> @Denton24646 @dinhxuanvu Where can I see the manual test scripts / images you were using?

@ecordell https://github.com/dinhxuanvu/catsrc-test

Denton24646:
I'm going to close this in favor of #1109 since that is the same implementation just further ahead and QE is asking about this.

I think we may want to revisit the overall design of this feature and have it be goroutine based, with a `timer.Tick` set by the interval and a new channel hanging off the top level operator struct. But that can be discussed at a later point.

PR #1104, opened 2019-10-31
Alerts for failing OLM operators

anik120:

- Add prometheus rule for firing an alert when csv_abnormal metric
  is emitted with phase=failed
- Alert message contains name and version of operator.

As a OpenShift administrator, I want to be able to receive a notification when an operator is consistently reaching a failed state so that I can immediately address known issues.

[Enhancement proposal](https://github.com/openshift/enhancements/pull/12)

**Reviewer Checklist**

### DIFF
--- a/deploy/chart/templates/0000_90_olm_01-prometheus-rule.yaml
+++ b/deploy/chart/templates/0000_90_olm_01-prometheus-rule.yaml
@@ -1,0 +1,20 @@
+{{ if and .Values.installType (eq .Values.installType "ocp") }}
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: olm-alert-rules
+  namespace: {{ .Values.namespace }}
+  labels:
+    prometheus: alert-rules
+    role: alert-rules
+spec:
+  groups:
+    - name: olm.failing_operators.rules
+      rules:
+        - alert: FailingOperator
+          annotations:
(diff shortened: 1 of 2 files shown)


deploy/chart/templates/0000_90_olm_01-prometheus-rule.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: olm-alert-rules
+  namespace: openshift-operator-lifecycle-manager
```
awgreene: Shouldn't the namespace be set as a variable (example: `{{ .Values.namespace }}`), like [other files](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/deploy/chart/templates/0000_90_olm_00-service-monitor.yaml#L6)) do in the templates dir?

anik120: @awgreene updated the PR. PTAL.

manifests/0000_90_olm_01-prometheus-rule.yaml
```diff
@@ -0,0 +1,19 @@
+
```
awgreene: Can we remove this newline?

anik120: @awgreene done.

PR #1105, opened 2019-11-01
Use the grpc response as the source of truth for update graph data

ecordell:
This trusts the values coming from the registry pod to determine what an operator replaces.

It also removes the remaining places that csvs needed to be parsed during resolution.

Depends on https://github.com/operator-framework/operator-registry/pull/113 and will need to be updated when it merges.

This allows us to change how the graph is generated in the future. 

**Reviewer Checklist**


pkg/controller/registry/resolver/operators.go
```diff
@@ -238,18 +238,14 @@ type Operator struct {
 
 var _ OperatorSurface = &Operator{}
 
-func NewOperatorFromBundle(bundle *api.Bundle, replaces string, startingCSV string, sourceKey CatalogKey) (*Operator, error) {
+func NewOperatorFromBundle(bundle *api.Bundle, startingCSV string, sourceKey CatalogKey) (*Operator, error) {
```
gallettilance: Just asking because I don't know how this works very well:\
since the replaces field in the db is an int64 and what is used here is a string that is the equivalent of the replaces field in the csv - does the `o.Identifier()` extract that replaces csv name from the database/query?

ecordell: The values from `replaces` in the registry isn't seen by olm at all. 

OLM queries for a replacement with `FindReplacement`, and uses whatever the registry returns as the value for the "replaces" field that it puts in the CSV.

jpeeler:
/retest
Failing test passed for me locally (TestSubscriptionUpdatesMultipleIntermediates).

PR #1109, opened 2019-11-04
Feat/catalog source polling mechanism

dinhxuanvu:
A dup of #1103 with an extra commit that is still pending-merge on Daniel's repo.

This branch is functional and ready for demo. 

**Reviewer Checklist**


Denton24646:
> Looks really good to me!
> 
> I was playing around with this last night and the one issue I ran into was that the sync function for the catalog source doing `ensuring registry server` was being called asynchronously in a somewhat random interval, and therefore the polling interval that we are introducing to the catsrc spec is not being honored by the catalog operator. I remember @ecordell mentioned something about a 15 minute interval being the minimum? I saw something like that when testing locally, I would have a polling interval of 1 minute on the catsrc but the update pod would only come up after 10 minutes or so....

I just talked with evan and there is indeed a default  for syncing the catalog source of either 10 or 15 minutes. For v1 of this feature we can just enforce a default catscrc polling of at least 15 minutes, to honor the catalog source polling interval, via the new default feature in 1.16 of k8s

pkg/controller/registry/reconciler/grpc.go
```diff
@@ -136,22 +160,78 @@ func (c *GrpcRegistryReconciler) EnsureRegistryServer(catalogSource *v1alpha1.Ca
...
+			logrus.WithField("CatalogSource", source.GetName()).Info("detect image update for catalogsource pod")
+
+			updateFlag = true
+			updatePod.Labels[CatalogSourceLabelKey] = source.GetName()
+			updatePod.Labels[CatalogSourceUpdateKey] = ""
+
+			// Update the update pod to promote it to serving pod
+			_, err := c.OpClient.KubernetesInterface().CoreV1().Pods(source.GetNamespace()).Update(updatePod)
+			if err != nil {
+				return errors.Wrapf(err, "error creating new pod: %s", source.Pod().GetName())
```
exdx: `error updating new catalog source pod` is clearer

exdx:
To be continued in #1161

PR #1080, opened 2019-10-17
Bug 1762769: Prioritize APIs from same CatSrc

awgreene:
Problem: When OLM installs an operator that requires dependencies that
are provided via multiple operators in different CatalogSources, the API
is pulled from any of the CatalogSources that provide the API.

Solution: This commit introduces a change so that OLM will generate a
list of operators that depend on the API, randomly select one of their
sources, and prioritize checking in that CatalogSource for the API prior
to checking the remaining CatalogSource for the API.

**Reviewer Checklist**
- [x] Implementation matches the proposed design, or proposal is updated to match implementation
- [x] Sufficient unit test coverage 
- [x] Sufficient end-to-end test coverage
- [x] Docs updated or added to `/docs` 
- [x] Commit messages sensible and descriptive

Fixes #1076

PR #1110, opened 2019-11-05
Bug 1763750: Prioritize APIs from same CatSrc

awgreene:
Problem: When OLM installs an operator that requires dependencies that
are provided via multiple operators in different CatalogSources, the API
is pulled from any of the CatalogSources that provide the API.

Solution: This commit introduces a change so that OLM will generate a
list of operators that depend on the API, randomly select one of their
sources, and prioritize checking in that CatalogSource for the API prior
to checking the remaining CatalogSource for the API.

This commit is different than the 4.3 fix because of a method signature in the e2e test.

Reviewer Checklist

 Implementation matches the proposed design, or proposal is updated to match implementation
 Sufficient unit test coverage
 Sufficient end-to-end test coverage
 Docs updated or added to /docs
 Commit messages sensible and descriptive

### DIFF
--- a/pkg/controller/registry/resolver/evolver.go
+++ b/pkg/controller/registry/resolver/evolver.go
@@ -112,8 +112,15 @@ func (e *NamespaceGenerationEvolver) queryForRequiredAPIs() error {
 		}
 		e.gen.MarkAPIChecked(*api)
 
+		// identify the initialSource
+		initialSource := CatalogKey{}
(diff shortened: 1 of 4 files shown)


ecordell:
This is a cherry-pick of https://github.com/operator-framework/operator-lifecycle-manager/pull/1080

dinhxuanvu:
@operator-framework/patch-managers Ready for review. PTAL.

PR #1114, opened 2019-11-07
Bug 1767004: defer provided api update in operator groups

jpeeler:
This is the case of a bug where the CSV status wasn't updated due to an
invalid CSV. The cause for this was twofold:

1) The CSV and OperatorGroup reconcile loops were undoing each others
changes continously, giving no chance for the CSV to get synced beyond
provided api conflict detection.

2) Due to the way the provided APIs are produced differently for
operator groups (uses GVKSTringToProvidedAPISet) and OLM (uses
NewOperatorFromV1Alpha1CSV) when an invalid CSV is in use, the resolver
attempts to continue to instruct that an operator group annotation
update is needed beyond the point of that being true.

The changes are to ensure that OLM does not get stuck attempting to
update the operator group annotations when they are no longer needed
combined with ensuring that the operator group sync loop does not try to
prematurely dismiss a provided api before the CSV has a chance to check
the requirements and produce status.

In order to accomplish "premature detection", providedAPIsFromCSVs was
modified to not only return the APIs but also the CSV that provides a
given API.

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/olm/operator_test.go
+++ b/pkg/controller/operators/olm/operator_test.go
@@ -4264,7 +4264,7 @@ func TestSyncOperatorGroups(t *testing.T) {
 			require.NoError(t, err)
 
 			// Sync csvs enough to get them back to succeeded state
-			for i := 0; i < 8; i++ {
+			for i := 0; i < 16; i++ {
 				opGroupCSVs, err := op.client.OperatorsV1alpha1().ClusterServiceVersions(operatorNamespace).List(metav1.ListOptions{})
 				require.NoError(t, err)
 
@@ -4292,7 +4292,7 @@ func TestSyncOperatorGroups(t *testing.T) {
 						require.NoError(t, err)
 					}
 
-					if i == 8 {
+					if i == 16 {
 						err = wait.PollImmediate(1*time.Millisecond, 10*time.Second, func() (bool, error) {
 							for namespace, objects := range tt.final.objects {
 								if err := RequireObjectsInCache(t, op.lister, namespace, objects, true); err != nil {
--- a/pkg/controller/operators/olm/operatorgroup.go
+++ b/pkg/controller/operators/olm/operatorgroup.go
@@ -19,6 +19,7 @@ import (
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/install"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/registry/resolver"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/ownerutil"
+	opregistry "github.com/operator-framework/operator-registry/pkg/registry"
 )
 
 const (
@@ -126,7 +127,13 @@ func (a *Operator) syncOperatorGroups(obj interface{}) error {
 	groupSurface := resolver.NewOperatorGroup(op)
 	groupProvidedAPIs := groupSurface.ProvidedAPIs()
 	providedAPIsForCSVs := a.providedAPIsFromCSVs(op, logger)
-	providedAPIsForGroup := providedAPIsForCSVs.Union(groupProvidedAPIs)
+	providedAPIsForGroup := make(resolver.APISet)
+	for api := range providedAPIsForCSVs {
+		providedAPIsForGroup[api] = struct{}{}
+	}
+	for api := range groupProvidedAPIs {
+		providedAPIsForGroup[api] = struct{}{}
+	}
 
 	csvs, err := a.findCSVsThatProvideAnyOf(providedAPIsForGroup)
 	if err != nil {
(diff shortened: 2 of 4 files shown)


jpeeler:
/retest
I think this PR is good to go, but CI seems to be broken right now.

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -230,20 +237,40 @@ func (a *Operator) providedAPIsFromCSVs(group *v1.OperatorGroup, logger *logrus.
...
+				logger.Debugf("aborting operator group provided API update due to CSV %v in phase %v", csv.GetName(), csv.Status.Phase)
+				return
+			}
+		}
+	}
+
 	// Prune providedAPIs annotation if the cluster has fewer providedAPIs (handles CSV deletion)
-	if intersection := groupProvidedAPIs.Intersection(providedAPIsFromCSVs); len(intersection) < len(groupProvidedAPIs) {
+	//if intersection := groupProvidedAPIs.Intersection(providedAPIsFromCSVs); len(intersection) < len(groupProvidedAPIs) {
+	if len(intersection) < len(groupProvidedAPIs) {
```
njhale: This should only happen when the `OperatorGroup` has provided APIs that no longer exist in the namespace. Could you explain how this condition is met otherwise?

pkg/controller/operators/olm/operator_test.go
```diff
@@ -4264,7 +4264,7 @@ func TestSyncOperatorGroups(t *testing.T) {
 			require.NoError(t, err)
 
 			// Sync csvs enough to get them back to succeeded state
-			for i := 0; i < 8; i++ {
+			for i := 0; i < 16; i++ {
```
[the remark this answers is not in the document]
jpeeler: I'm pretty sure doubling wasn't necessary, but due to the CSVs not getting updated quite as quickly some more loops were needed.

test/e2e/csv_e2e_test.go
```diff
@@ -3212,3 +3212,117 @@ func TestCreateCSVRequirementsEvents(t *testing.T) {
...
+	}
+	csvCheckPhaseAndRequirementStatus := func(csv *v1alpha1.ClusterServiceVersion) bool {
+		if csv.Status.Phase == v1alpha1.CSVPhasePending {
+			for _, status := range csv.Status.RequirementStatus {
+				if status.Message == notServedStatus.Message {
+					return true
+				}
+			}
+		}
+		return false
```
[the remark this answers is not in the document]
jpeeler: I wanted to do something similar to that, but it seems using assert also fails immediately if the condition isn't met.

pkg/controller/operators/olm/operatorgroup.go
jpeeler: This was changed mainly because proviedAPIsFromCSVs no longer is an APISet, so calculating the intersection isn't possible. But with an invalid api version specified in a CSV, it is possible that the operator group has apis that were never on the cluster (which is part of the problem here). I can remove the commented out line I left in there if you want.

jpeeler:
In general, it seems that having any loops updating resources other than the one the loop was started for leads to bad contention. But operator groups and CSVs are coupled in a way that makes the problem extremely bad.

test/e2e/csv_e2e_test.go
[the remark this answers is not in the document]
njhale: Good catch. I thought I've used a matcher package where the assert just returned a boolean and didn't fail. Sorry for the confusion.

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -230,20 +237,40 @@ func (a *Operator) providedAPIsFromCSVs(group *v1.OperatorGroup, logger *logrus.
 			logger.WithError(err).Warn("could not create OperatorSurface from csv")
 			continue
 		}
-		providedAPIsFromCSVs = providedAPIsFromCSVs.Union(operatorSurface.ProvidedAPIs().StripPlural())
+		for providedAPI := range operatorSurface.ProvidedAPIs().StripPlural() {
+			providedAPIsFromCSVs[providedAPI] = csv
```
njhale: Why lug the CSV around when you can filter out the ones you don't want [here](https://github.com/operator-framework/operator-lifecycle-manager/blob/1debd997de2f70ccfc7f4e9c82726aba7c36aa60/pkg/controller/operators/olm/operatorgroup.go#L225)?

pkg/controller/operators/olm/operatorgroup.go
njhale: > But with an invalid api version specified in a CSV, it is possible that the operator group has apis that were never on the cluster (which is part of the problem here)

What does "never on the cluster" mean? IIRC, the point here is to remove APIs that no CSV in the OperatorGroup's namespace claims to provide. I would rather have false positives and cleanup more often 

> I can remove the commented out line I left in there if you want.

I think we want to avoid commented code piling up, but we don't need to cause another round of CI over it.

njhale: also see: https://github.com/operator-framework/operator-lifecycle-manager/pull/1114#discussion_r346130664

njhale: also see: https://github.com/operator-framework/operator-lifecycle-manager/pull/1114#discussion_r346130664

jpeeler: Remember that this bug is dealing with an invalid api due to botched syntax. Due to the botched syntax as shown in the bug, the CRD version could never have possibly existed.

The problem with cleaning up more often is there are two sync loops that end up fighting each other with a constant stream of updates, which prevents validation from ever occurring and therefore not getting reported back to the user.

pkg/controller/operators/olm/operatorgroup.go
jpeeler: I don't think I want to filter out any CSVs. The CSVs are returned so that in pruneProvidedAPIs while iterating over the provided apis, the CSV phase for a given api can be checked to ensure it has progressed to the point of validation (past phase pending).

njhale: hmm, I think they are roughly isomorphic, but this isn't a big deal.

PR #1116, opened 2019-11-08
Bug 1774720: catalogsource context cancel function leak

Denton24646:
Start handling the returned cancel func from `context.WithTimeout` on so as not to leak cancel functions. This was done via a defer which required moving the default cause into an anonymous function and calling it immediately to have introduce function scope. 

This was brought up originally by @hdonnay

Improve memory utilization of the catalog operator 

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/registry/grpc/source.go
+++ b/pkg/controller/registry/grpc/source.go
@@ -142,27 +142,30 @@ func (s *SourceStore) watch(ctx context.Context, key resolver.CatalogKey, source
 		case <-ctx.Done():
 			return
 		default:
-			timer, _ := context.WithTimeout(ctx, s.stateTimeout(state))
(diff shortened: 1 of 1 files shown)


kevinrizza:
> Thanks for fixing this! I just want to make sure we don't need a BZ open before we merge it.
> 
> /hold

If we're not sure, why not just create a BZ so we can merge this? @njhale @Denton24646

Denton24646:
> > Thanks for fixing this! I just want to make sure we don't need a BZ open before we merge it.
> > /hold
> 
> If we're not sure, why not just create a BZ so we can merge this? @njhale @Denton24646

Sounds good, I'll create a BZ 👍

Denton24646:
bz opened @njhale @kevinrizza

PR #1118, opened 2019-11-08
Bug 1749031: test for OLM takes a while to detect internal CatalogSource changes

Denton24646:
This PR introduces an e2e test for updating a CSV via updating the configmap backing the catalogsource. The ultimate goal is to provide some meaningful timeouts for the different steps in the test and ensure in our CI that the configmap-based update time is satisfactory. For go tests there is a 5 minute default timeout so that is the timeout set initially. 

Bugzilla report 

**Reviewer Checklist**

### DIFF
--- a/test/e2e/subscription_e2e_test.go
+++ b/test/e2e/subscription_e2e_test.go
@@ -3,6 +3,7 @@ package e2e
 import (
 	"encoding/json"
 	"fmt"
+	"github.com/stretchr/testify/assert"
 	"strings"
 	"sync"
 	"testing"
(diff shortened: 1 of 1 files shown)


Denton24646:
The referenced bug has been resolved - the behavior in question was expected due to the fact that the catalog source/config map that was being updated was in a different namespace than the subscription object. In this case the default sync period of 15 minutes is in effect, so the fact that the sync took 5 minutes is actually normal. This test would time out after 5 minutes and is not really part of this bug story any more. 

@ecordell it might we worth revisiting the default 15 minute sync interval

PR #1123, opened 2019-11-12
Bug 1769030: Replacing operator creates duplicate secrets

Bowenislandsong:
Cause:
OLM catalog ensurer EnsureServiceAccount makes sure the service account
is updated when a new version of an operator is present. This
happens during ExecutePlan applying InstallPlan to a namespace.
If it is an update, fields of service account are updated but the
references to older secrets are dropped.

Consequence:
This process of dereferencing secret fails to clean up the older
secrets and result in the secrets pilling up as the operator upgrades.
Eventually, there will be too many old secrets laying around and only
getting cleaned up when the operator is uninstalled.

Fix:
We carry over older secrets through updating the service account.

Result:
Older secretes are again referred in the updated SA.

### DIFF
--- a/pkg/controller/operators/catalog/step_ensurer.go
+++ b/pkg/controller/operators/catalog/step_ensurer.go
@@ -3,15 +3,17 @@ package catalog
 import (
 	"fmt"
 
+	errorwrap "github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/clientset/versioned"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/operatorclient"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/ownerutil"
-	errorwrap "github.com/pkg/errors"
-	corev1 "k8s.io/api/core/v1"
-	rbacv1 "k8s.io/api/rbac/v1"
-	k8serrors "k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 func newStepEnsurer(kubeClient operatorclient.ClientInterface, crClient versioned.Interface) *StepEnsurer {
@@ -116,10 +118,23 @@ func (o *StepEnsurer) EnsureServiceAccount(namespace string, sa *corev1.ServiceA
 		return
 	}
 
-	sa.SetNamespace(namespace)
-	if _, updateErr := o.kubeClient.UpdateServiceAccount(sa); updateErr != nil {
-		err = errorwrap.Wrapf(updateErr, "error updating service account: %s", sa.GetName())
+	// Carrying secrets through the service account update.
+	preSa, getErr := o.kubeClient.KubernetesInterface().CoreV1().ServiceAccounts(namespace).Get(sa.Name,
+		metav1.GetOptions{})
+	if getErr != nil {
+		err = errorwrap.Wrapf(getErr, "error getting older version of service account: %s", sa.GetName())
 		return
+	}
+	sa.Secrets = preSa.Secrets
+
+	sa.SetNamespace(namespace)
+
+	// Use DeepDerivative to check if new SA is the same as the old SA. If no field is changed, we skip the update call.
+	if !apiequality.Semantic.DeepDerivative(sa, preSa) {
+		if _, updateErr := o.kubeClient.UpdateServiceAccount(sa); updateErr != nil {
+			err = errorwrap.Wrapf(updateErr, "error updating service account: %s", sa.GetName())
+			return
+		}
 	}
 
 	status = v1alpha1.StepStatusPresent
--- a/pkg/controller/operators/catalog/operator_test.go
+++ b/pkg/controller/operators/catalog/operator_test.go
@@ -382,6 +382,96 @@ func TestExecutePlan(t *testing.T) {
 			want: []runtime.Object{service("service", namespace), csv("csv", namespace, nil, nil)},
 			err:  nil,
 		},
+		{
+			testName: "CreateServiceAccount",
+			in: withSteps(installPlan("p", namespace, v1alpha1.InstallPlanPhaseInstalling, "csv"),
+				[]*v1alpha1.Step{
+					{
+						Resource: v1alpha1.StepResource{
+							CatalogSource:          "catalog",
+							CatalogSourceNamespace: namespace,
+							Group:                  "",
+							Version:                "v1",
+							Kind:                   "ServiceAccount",
+							Name:                   "sa",
+							Manifest: toManifest(serviceAccount("sa", namespace, "",
+								objectReference("init secret"))),
+						},
+						Status: v1alpha1.StepStatusUnknown,
+					},
+				},
+			),
+			want: []runtime.Object{serviceAccount("sa", namespace, "", objectReference("init secret"))},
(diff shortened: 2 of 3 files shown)


Bowenislandsong:
e2e test inprog

test/e2e/csv_e2e_test.go
```diff
@@ -1531,7 +1532,8 @@ func TestUpdateCSVWithOwnedAPIService(t *testing.T) {
```
JHrefiD: hello world.

h777xx:
https://github.com/operator-framework/operator-lifecycle-manager/pull/1123#discussion_r346065906

JHrefiD:
#1123

JHrefiD:
> [#1123 (comment)](https://github.com/operator-framework/operator-lifecycle-manager/pull/1123#discussion_r346065906)

Thank you.

pkg/controller/operators/catalog/step_ensurer.go
```diff
@@ -116,12 +116,35 @@ func (o *StepEnsurer) EnsureServiceAccount(namespace string, sa *corev1.ServiceA
...
+	}
+
 	sa.SetNamespace(namespace)
 	if _, updateErr := o.kubeClient.UpdateServiceAccount(sa); updateErr != nil {
 		err = errorwrap.Wrapf(updateErr, "error updating service account: %s", sa.GetName())
 		return
 	}
 
+	for _, secret := range preSa.Secrets {
+		foregroundDelete := metav1.DeletePropagationForeground
```
jpeeler: I don't think doing a foreground type of deletion is necessary if there's nothing left referencing these secrets as dependencies, right?

Bowenislandsong: You are right. This is unnecessary but I thought it would not hurt. Do you think we should just leave it blank then? Thanks

pkg/controller/operators/catalog/step_ensurer.go
```diff
@@ -116,12 +116,35 @@ func (o *StepEnsurer) EnsureServiceAccount(namespace string, sa *corev1.ServiceA
 		return
 	}
 
+	// Before UpdateServiceAccount and dereference secrets of the older SA,
```
ecordell: Doesn't the original bug report have this issue with the same serviceaccount name? 

From line 114, this section should only be hit if the serviceaccount doesn't exist at all, so I'm not convinced this will fix the bug.

I think the issue may be that `Create` on `ServiceAccount` always creates a secret even if the serviceaccount exists? Maybe we should start with a `Get` before the initial `Create`?

Do you agree, or am I missing how this fixes the issue?

Bowenislandsong: You might have missed `!k8serrors.IsAlreadyExists` which reports any SA creation problem other than SA exists. Therefore, the only scenario where *my addition* hits is if the SA **does** exists. 

The logic of the original code got me at first as well.

Bowenislandsong:
/hold 
E2e test includes pulling quay image. This should be removed before merging.

Bowenislandsong:
It seems like the reason that the Kubernetes Token Controller is not cleaning up the dereferenced secrets is that the design does not cover the scenario. The Kubernetes design doc https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/#token-controller states that `TokenController observes secret deletion and removes a reference from the corresponding ServiceAccount if needed.` There is no scenario for deleting secret when it is dereferenced by the ServiceAccount.

pkg/controller/operators/catalog/step_ensurer.go
jpeeler: I believe you should have nothing passed into the DeleteOptions below.

pkg/controller/operators/catalog/step_ensurer.go
ecordell: This is what I get for reviewing too quickly 😄  Ignore my previous comment.

However - I really don't think we should be in the business of deleting the serviceaccount secrets.

I did some testing, and the issue is that when we `Update`, our SA definition doesn't have the secrets set, which removes the configured secret and causes the SA controller to generate a new one:

```sh
$ k -n default create serviceaccount test
serviceaccount/test created

$ k -n default get secrets
NAME                  TYPE                                  DATA   AGE
default-token-67bw8   kubernetes.io/service-account-token   3      3m23s
test-token-9lx4p      kubernetes.io/service-account-token   3      3s

# an update that doesn't cause the bug
$ k -n default label serviceaccounts test with=test
serviceaccount/test labeled

# no duplicate secrets
$ k -n default get secrets
NAME                  TYPE                                  DATA   AGE
default-token-67bw8   kubernetes.io/service-account-token   3      6m40s
test-token-9lx4p      kubernetes.io/service-account-token   3      3m20s

# an update that causes the bug - remove the secrets block
$ k -n default edit serviceaccounts test
serviceaccount/test edited

# now there are duplicates
$ k -n default get secrets
NAME                  TYPE                                  DATA   AGE
default-token-67bw8   kubernetes.io/service-account-token   3      8m3s
test-token-9lx4p      kubernetes.io/service-account-token   3      4m43s
test-token-xwmpg      kubernetes.io/service-account-token   3      4s
```

The solution here shouldn't be to delete the extra secrets, it should be to not create the extra secrets in the first place.

A quick option for now would be to:

- Get the current ServiceAccount (if it exists)
- Add the current secret list to the object from the step
- Then call update with that object - it will have the secret list so a new one won't be generated

But it would be even better to use `DeepDerivative` or server-side apply.

Bowenislandsong: That was my other thought on the fix too. I eventually chose this to make sure we always have the right token in the secrets. I worry that secrets may change in the install plan rendering old secrets to be invalid. However I might be worrying for nothing. I’ll make a commit to use that approach.

Bowenislandsong: @ecordell I liked the server-side approach but it being in beta gives me goosebumps. I am not entirely sure about your suggestion about `DeepDerivative`. If you are talking about https://github.com/kubernetes/kubernetes/blob/6be19d85f206d64044ae940540a437c65675a526/third_party/forked/golang/reflect/deep_equal.go#L378  then I do not see how we can use it. If you are talking about the concept. Not every field can be copied (ie: timestamp, etc). If we copy every unset field in the new SA from the existing SA, we need to know exactly what is not copiable. I do not think my understanding of your `DeepDerivative` is correct. Therefore, please let me know what you are explicitly suggesting. Thank you!

Bowenislandsong:
@ecordell The easy approach is commited.

pkg/controller/operators/catalog/step_ensurer.go
```diff
@@ -116,7 +116,23 @@ func (o *StepEnsurer) EnsureServiceAccount(namespace string, sa *corev1.ServiceA
...
+	if getErr != nil {
+		err = errorwrap.Wrapf(getErr, "error getting older version of service account: %s", sa.GetName())
+		return
+	}
+	// Returns error if no secret is found for the old service account. According to Kubernetes,
+	// the controller loop ensures a secret with an API token exists for each service account.
+	if len(preSa.Secrets) == 0 {
+		err = fmt.Errorf("no secretes fetched from the older version of service account: %s", sa.GetName())
+		return
+	}
```
awgreene: Could you explain why you're adding this code? If Kubernetes ensures something it seems odd that we're adding a check.

awgreene: Also - if a there are no secrets, you would be adding an empty list of secrets which should cause k8s to generate the secret (the cause of the bug you're fixing).

pkg/controller/operators/catalog/step_ensurer.go
```diff
@@ -116,7 +116,23 @@ func (o *StepEnsurer) EnsureServiceAccount(namespace string, sa *corev1.ServiceA
...
+	sa.Secrets = preSa.Secrets
+
 	sa.SetNamespace(namespace)
+
```
awgreene: Please remove newline.

pkg/controller/operators/catalog/step_ensurer.go
Bowenislandsong: you are right, we do not need this anymore.

awgreene:
@Bowenislandsong you may want to use deepderivative to compare the objects.

PR #1125, opened 2019-11-14
Bug 1779313: Enable multiple namespaces sync if catsrc is updated in global ns

dinhxuanvu:
Currently, if CatalogSource is updated and connection is Ready,
only the namespace where CatalogSource resides gets resynced. Now, a list
of namespaces that contain Subscriptions that use updated CatalogSource
will get resynced instead.

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -348,11 +348,11 @@ func (o *Operator) syncSourceState(state grpc.SourceState) {
 	switch state.State {
 	case connectivity.Ready:
 		if o.namespace == state.Key.Namespace {
-			subs, err := index.CatalogSubscriberNamespaces(o.catalogSubscriberIndexer,
+			namespaces, err := index.CatalogSubscriberNamespaces(o.catalogSubscriberIndexer,
 				state.Key.Name, state.Key.Namespace)
 
 			if err == nil {
-				for _, ns := range subs {
+				for ns := range namespaces {
 					o.nsResolveQueue.Add(ns)
 				}
 			}
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -66,24 +66,25 @@ const (
 type Operator struct {
 	queueinformer.Operator
 
-	logger                 *logrus.Logger
-	clock                  utilclock.Clock
-	opClient               operatorclient.ClientInterface
-	client                 versioned.Interface
-	dynamicClient          dynamic.Interface
-	lister                 operatorlister.OperatorLister
-	catsrcQueueSet         *queueinformer.ResourceQueueSet
-	subQueueSet            *queueinformer.ResourceQueueSet
-	ipQueueSet             *queueinformer.ResourceQueueSet
-	nsResolveQueue         workqueue.RateLimitingInterface
-	namespace              string
-	sources                *grpc.SourceStore
-	sourcesLastUpdate      sharedtime.SharedTime
-	resolver               resolver.Resolver
-	reconciler             reconciler.RegistryReconcilerFactory
-	csvProvidedAPIsIndexer map[string]cache.Indexer
-	clientAttenuator       *scoped.ClientAttenuator
-	serviceAccountQuerier  *scoped.UserDefinedServiceAccountQuerier
+	logger                   *logrus.Logger
+	clock                    utilclock.Clock
+	opClient                 operatorclient.ClientInterface
+	client                   versioned.Interface
+	dynamicClient            dynamic.Interface
+	lister                   operatorlister.OperatorLister
+	catsrcQueueSet           *queueinformer.ResourceQueueSet
+	subQueueSet              *queueinformer.ResourceQueueSet
+	ipQueueSet               *queueinformer.ResourceQueueSet
+	nsResolveQueue           workqueue.RateLimitingInterface
+	namespace                string
+	sources                  *grpc.SourceStore
+	sourcesLastUpdate        sharedtime.SharedTime
+	resolver                 resolver.Resolver
+	reconciler               reconciler.RegistryReconcilerFactory
+	csvProvidedAPIsIndexer   map[string]cache.Indexer
+	catalogSubscriberIndexer map[string]cache.Indexer
+	clientAttenuator         *scoped.ClientAttenuator
+	serviceAccountQuerier    *scoped.UserDefinedServiceAccountQuerier
 }
 
 type CatalogSourceSyncFunc func(logger *logrus.Entry, in *v1alpha1.CatalogSource) (out *v1alpha1.CatalogSource, continueSync bool, syncError error)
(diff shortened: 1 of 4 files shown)


awgreene:
@dinhxuanvu 
`error: endpoint "https://registry.svc.ci.openshift.org" does not support v2 API`

njhale:
@dinhxuanvu could you open a BZ for this?

dinhxuanvu:
@njhale This PR is to address the issue on the existing bug that we were talking about during standup early on this week. I haven't tagged the bug on the title yet.

pkg/lib/index/catalog.go
```diff
@@ -0,0 +1,56 @@
...
+	// the index func with an indexer.
+	PresentCatalogIndexFuncKey string = "presentcatalogindexfunc"
+)
+
+// PresentCatalogIndexFunc returns index from CatalogSource/CatalogSourceNamespace
+// of the given object (Subscription)
+func PresentCatalogIndexFunc(obj interface{}) ([]string, error) {
+	sub, ok := obj.(*v1alpha1.Subscription)
+	if !ok {
+		return []string{""}, fmt.Errorf("invalid object of type: %T", obj)
```
Denton24646: nit: this can return nil, error

pkg/controller/operators/catalog/operator.go
```diff
@@ -202,6 +204,12 @@ func NewOperator(ctx context.Context, kubeconfigPath string, clock utilclock.Clo
 		// Wire Subscriptions
 		subInformer := crInformerFactory.Operators().V1alpha1().Subscriptions()
 		op.lister.OperatorsV1alpha1().RegisterSubscriptionLister(namespace, subInformer.Lister())
+		if err := subInformer.Informer().AddIndexers(cache.Indexers{index.PresentCatalogIndexFuncKey: index.PresentCatalogIndexFunc}); err != nil {
```
ecordell: An indexer takes an object and returns a set of keys that can be used to find that same object again in the future.

For this case, don't we want to add an indexer to namespaces that let's us find the namespaces to requeue? 

I.e. we want: `in: CatalogSource`, `out: Namespaces`, and then we requeue each of those namespaces.

Currently, this indexer is configured for:

`in: CatalogSource, out: Subscriptions` and we have to iterate over the subscriptions to get the list of namespaces.

it would be nice if our processing was just: 

```go
case connectivity.Ready:
  namespaces  := op.catalogSubscriberIndexer[state.key.Namespace].ByIndex(PresentCatalogIndexFuncKey, fmt.Sprintf("%s/%s", state.key.Name, state.key.Namespace))

  for _, ns := range namespaces {
     o.nsResolveQueue.Add(ns) 
  }
case ...:
```

dinhxuanvu: Hey Evan. I did, in fact, do `in: CatalogSource, out: Namespaces` in previous iteration of this PR but changed to current `in: CatalogSource, out: Subscriptions` implementation. The reason for it is I think it is more valuable to find out the list of `Subscriptions` that subscribes to a specific `CatalogSource` than the list of `namespaces`. Also, with this current implementation, the `namespaces` are in values of the returned map. In the way, we still have direct access to the list of namespaces without any extra works. There is a complication of duplicated namespaces but the `workerqueue` is using a map behind the scene and it automatically rejects duplicated item by default.

dinhxuanvu: Either way, if you feel there is no value in get `Subscriptions` list then I can simplify it to just get a `namespaces` list.

ecordell:
This looks good but the test is failing - holding off for now

jpeeler:
/hold
I think a 4.4 bug needs to be made for this.

pkg/lib/index/catalog.go
```diff
@@ -0,0 +1,53 @@
+package indexer
```
njhale: nit: although there are some exceptions, this probably shouldn't be in the `lib` folder since it contains OLM specific logic.

pkg/lib/index/catalog.go
```diff
@@ -0,0 +1,53 @@
...
+// PresentCatalogIndexFunc returns index from CatalogSource/CatalogSourceNamespace
+// of the given object (Subscription)
+func PresentCatalogIndexFunc(obj interface{}) ([]string, error) {
+	sub, ok := obj.(*v1alpha1.Subscription)
+	if !ok {
+		return []string{""}, fmt.Errorf("invalid object of type: %T", obj)
+	}
+
+	if sub.Spec.CatalogSource != "" && sub.Spec.CatalogSourceNamespace != "" {
+		return []string{sub.Spec.CatalogSource + "/" + sub.Spec.CatalogSourceNamespace}, nil
```
njhale: nit: the common convention is `namespace/name`.

pkg/lib/index/catalog.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	return []string{""}, nil
+}
+
+// CatalogSubscriberNamespaces returns the list of namespace (as a map with namespace as key)
+// which has Suscriptions(s) that subscribe(s) to a given CatalogSource (name/namespace)
+func CatalogSubscriberNamespaces(indexers map[string]cache.Indexer, name, namespace string) (map[string]struct{}, error) {
```
njhale: nit: the result map isn't being used as a map outside of this function -- seems kinda leaky

test/e2e/catalog_e2e_test.go
```diff
@@ -81,6 +81,115 @@ func TestCatalogLoadingBetweenRestarts(t *testing.T) {
...
+
+	globalNS := operatorNamespace
+	c := newKubeClient(t)
+	crc := newCRClient(t)
+
+	// Determine which namespace is global. Should be `openshift-marketplace` for OCP 4.2+.
+	// Locally it is `olm`
+	namespaces, _ := c.KubernetesInterface().CoreV1().Namespaces().List(metav1.ListOptions{})
+	for _, ns := range namespaces.Items {
+		if ns.GetName() == "openshift-marketplace" {
```
njhale: nit: it would be nice to take this as a test argument so we can avoid making configuration assumptions in our test logic.

test/e2e/catalog_e2e_test.go
```diff
@@ -81,6 +81,115 @@ func TestCatalogLoadingBetweenRestarts(t *testing.T) {
...
+	// Attempt to get the catalog source before creating install plan
+	_, err := fetchCatalogSource(t, crc, mainCatalogName, globalNS, catalogSourceRegistryPodSynced)
+	require.NoError(t, err)
+
+	subscriptionSpec := &v1alpha1.SubscriptionSpec{
+		CatalogSource:          mainCatalogName,
+		CatalogSourceNamespace: globalNS,
+		Package:                mainPackageName,
+		Channel:                stableChannel,
+		StartingCSV:            mainCSV.GetName(),
```
njhale: nit: no need to specify a `StartingCSV` or a manual approval strategy

test/e2e/catalog_e2e_test.go
```diff
@@ -81,6 +81,115 @@ func TestCatalogLoadingBetweenRestarts(t *testing.T) {
...
+	require.NotNil(t, subscription)
+
+	installPlanName := subscription.Status.Install.Name
+	requiresApprovalChecker := buildInstallPlanPhaseCheckFunc(v1alpha1.InstallPlanPhaseRequiresApproval)
+	fetchedInstallPlan, err := fetchInstallPlan(t, crc, installPlanName, requiresApprovalChecker)
+	require.NoError(t, err)
+
+	fetchedInstallPlan.Spec.Approved = true
+	_, err = crc.OperatorsV1alpha1().InstallPlans(testNamespace).Update(fetchedInstallPlan)
+	require.NoError(t, err)
```
njhale: Since you only put the manifests for the first CSV into the catalog, there's no need to step through updates using a manual approval strategy.

PR #1142, opened 2019-11-19
Bug 1760608: add resource limits to all OLM pods and the 0.13.0 release for OCP

Denton24646:
Add resource limits to all OLM pods. Dupe of #1075 with limits found from doing some tests on a live OCP CI cluster (see https://github.com/operator-framework/operator-lifecycle-manager/pull/1075#issuecomment-554465931 for rationale and additional details). 

Note: limits apply to OCP releases only and are not applied to upstream deployments.  

Bugs reporting memory overflows that resulted from not having resource limits set for OLM pods. 

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/registry/reconciler/reconciler.go
+++ b/pkg/controller/registry/reconciler/reconciler.go
@@ -121,6 +121,10 @@ func Pod(source *v1alpha1.CatalogSource, name string, image string, labels map[s
 						InitialDelaySeconds: livenessDelay,
 					},
 					Resources: v1.ResourceRequirements{
+						Limits: v1.ResourceList{
+							v1.ResourceCPU:    resource.MustParse("100m"),
+							v1.ResourceMemory: resource.MustParse("100Mi"),
+						},
 						Requests: v1.ResourceList{
 							v1.ResourceCPU:    resource.MustParse("10m"),
 							v1.ResourceMemory: resource.MustParse("50Mi"),
--- a/deploy/ocp/values.yaml
+++ b/deploy/ocp/values.yaml
@@ -91,8 +91,8 @@ package:
     tolerationSeconds: 120
   resources:
     limits:
-      cpu: 100m
-      memory: 100Mi
+      cpu: 200m
+      memory: 200Mi
     requests:
       cpu: 10m
       memory: 50Mi
--- a/deploy/ocp/values.yaml
+++ b/deploy/ocp/values.yaml
@@ -91,8 +91,8 @@ package:
     tolerationSeconds: 120
   resources:
     limits:
-      cpu: 200m
-      memory: 200Mi
+      cpu: 400m
+      memory: 400Mi
     requests:
       cpu: 10m
       memory: 50Mi
--- a/deploy/ocp/values.yaml
+++ b/deploy/ocp/values.yaml
@@ -32,6 +32,9 @@ olm:
   tlsCertPath: /var/run/secrets/serving-cert/tls.crt
   tlsKeyPath: /var/run/secrets/serving-cert/tls.key
   resources:
+    limits:
+      cpu: 400m
+      memory: 400Mi
     requests:
       cpu: 10m
       memory: 160Mi
@@ -58,6 +61,9 @@ catalog:
     effect: NoExecute
     tolerationSeconds: 120
   resources:
+    limits:
+      cpu: 200m
+      memory: 300Mi
     requests:
       cpu: 10m
       memory: 80Mi
@@ -84,6 +90,9 @@ package:
     effect: NoExecute
     tolerationSeconds: 120
   resources:
+    limits:
+      cpu: 100m
+      memory: 100Mi
     requests:
       cpu: 10m
       memory: 50Mi
--- a/manifests/0000_50_olm_07-olm-operator.deployment.yaml
+++ b/manifests/0000_50_olm_07-olm-operator.deployment.yaml
@@ -62,6 +62,9 @@ spec:
           - name: OPERATOR_NAME
             value: olm-operator
           resources:
+            limits:
+              cpu: 400m
+              memory: 400Mi
             requests:
               cpu: 10m
               memory: 160Mi
--- a/manifests/0000_50_olm_08-catalog-operator.deployment.yaml
+++ b/manifests/0000_50_olm_08-catalog-operator.deployment.yaml
@@ -55,6 +55,9 @@ spec:
             value: "0.0.1-snapshot"
           
           resources:
+            limits:
+              cpu: 200m
+              memory: 300Mi
             requests:
               cpu: 10m
               memory: 80Mi
--- a/manifests/0000_50_olm_15-packageserver.clusterserviceversion.yaml
+++ b/manifests/0000_50_olm_15-packageserver.clusterserviceversion.yaml
@@ -123,8 +123,8 @@ spec:
                 terminationMessagePolicy: FallbackToLogsOnError
                 resources:
                   limits:
-                    cpu: 100m
-                    memory: 100Mi
+                    cpu: 200m
+                    memory: 200Mi
                   requests:
                     cpu: 10m
                     memory: 50Mi
--- a/manifests/0000_50_olm_15-packageserver.clusterserviceversion.yaml
+++ b/manifests/0000_50_olm_15-packageserver.clusterserviceversion.yaml
@@ -123,8 +123,8 @@ spec:
                 terminationMessagePolicy: FallbackToLogsOnError
                 resources:
                   limits:
-                    cpu: 200m
-                    memory: 200Mi
+                    cpu: 400m
+                    memory: 400Mi
                   requests:
                     cpu: 10m
                     memory: 50Mi
--- a/manifests/0000_50_olm_15-packageserver.clusterserviceversion.yaml
+++ b/manifests/0000_50_olm_15-packageserver.clusterserviceversion.yaml
@@ -122,6 +122,9 @@ spec:
                     port: 5443
                 terminationMessagePolicy: FallbackToLogsOnError
                 resources:
+                  limits:
+                    cpu: 100m
+                    memory: 100Mi
                   requests:
                     cpu: 10m
                     memory: 50Mi


deploy/ocp/manifests/0.13.0/0000_50_olm_07-olm-operator.deployment.yaml
```diff
@@ -67,8 +67,11 @@ spec:
             requests:
               cpu: 10m
               memory: 160Mi
-            
-          
+            limits:
+              cpu: 400m
+              memory: 400Mi
+
+
```
awgreene: Was this generated by an automated process? If not, please remove empty newlines.

deploy/chart/values.yaml
```diff
@@ -19,6 +19,9 @@ olm:
   nodeSelector:
     beta.kubernetes.io/os: linux
   resources:
+    limits:
+     cpu: 400m
+     memory: 400Mi
```
awgreene: I see from the rest of your PR that you did not apply these limits to the upstream distribution. I believe that the next `make release` will apply them to upstream.

awgreene: [Here](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/deploy/chart/templates/0000_90_olm_01-prometheus-rule.yaml#L1) is an example where we don't add annotations to the olm namespace in upstream:

pkg/controller/registry/reconciler/reconciler.go
```diff
@@ -121,6 +121,10 @@ func Pod(source *v1alpha1.CatalogSource, name string, image string, labels map[s
 						InitialDelaySeconds: livenessDelay,
 					},
 					Resources: v1.ResourceRequirements{
+						Limits: v1.ResourceList{
+							v1.ResourceCPU:    resource.MustParse("100m"),
+							v1.ResourceMemory: resource.MustParse("100Mi"),
+						},
```
awgreene: This is the only part of the PR that I'm weary about given that Catalogs should continuously grow as more operators are created.

kevinrizza: @awgreene @Denton24646 do we have an idea of how these resources scale on an average AWS cluster as catalogs get bigger? could be that we need to set some reasonable limits anyway, and we probably can't just say "Catalogs can be of any size" with the current memory footprint as it scales. or maybe it handles scale well given normal pod resource limits? either way we should probably have some idea what those numbers look like before we merge this.

Denton24646:
I took a look at profiling the different catalog source pods based on two different methods. 

The first was simply load-testing on a stock CI cluster similar to how I got limits for the other OLM pods: send requests via `kubectl get packagemanifests` every two seconds and observe the CPU/memory profiles for the different catalog source pods. I added profiles for the community operator and red hat operator catalog sources to the slide deck (see slides 13-16) 

https://docs.google.com/presentation/d/1qSB50bkZJaRbPAKv2aIm-26mRyivMYcaUvOxZ3z_sZo/edit?usp=sharing

As we can see both CPU and memory grow as requests build up and then flatline. The resource utilization is within the 100m/100Mi limits we have setup, although the community catalog CPU usage does get up to 100m (which is ok since the limit would just throttle it). Since we anticipate more community operators in the future we can expect the memory footprint to rise. However, unless we want to provide resource limits per catalog source, we need limits that are reasonable for all catalogs. 

The second method of profiling is more to the point that's being described above - how do the catalog resources scale in the face of bigger catalogs? I discussed with @njhale and we decided some back of the envelope math could be helpful.

First I went to operator-registry, and found the average package size across all community operators via `find . -depth 1 -type d -print | while read dir; do du -s "$dir"; done | awk '{x+=$1*512; r++} END {print(x/r, "bytes")}'` which came out to 188301 bytes or roughly 0.2MB. So on average a package will take up 0.2MB on disk, with 100 operators it will be roughly 20MB on disk. In the pod itself we load all the packages into sqlite and presumably the pod holds the database in memory, which consists of all the manifests and the overhead of sqlite. So we can expect at least that memory usage, but in practice we see around double. 

Honestly based on the graphs I actually think these catalog limits are pretty good - they constrain resources but don't interfere with the catalog pods responding to the packageserver - which is what we want the limits for. So tl;dr leave those particular limits as they are from what I'm seeing

kevinrizza:
@Denton24646 this is awesome! Can we publish or save these somewhere? It's definitely really valuable. Especially if we document how we profile this and automate it we could get some really interesting stats over time.

/lgtm

deploy/ocp/manifests/0.13.0/0000_50_olm_07-olm-operator.deployment.yaml
```diff
@@ -64,6 +64,9 @@ spec:
           - name: OPERATOR_NAME
             value: olm-operator
           resources:
+            limits:
```
njhale: Since the `0.13.0` release was already cut, we don't want to change its manifests. These changes will be included when we cut the next release.

Please exclude all changes to the `deploy/ocp/manifests/0.13.0` directory.

Denton24646:
> @Denton24646 this is awesome! Can we publish or save these somewhere? It's definitely really valuable. Especially if we document how we profile this and automate it we could get some really interesting stats over time.
> 
> /lgtm

Yes, absolutely! I'm going to flesh out the slides with more details and then move them to our team's shared google drive so everyone will have access to it. If we then decide to make the load-testing a formal process we can have the slides as a reference point.

deploy/ocp/manifests/0.13.0/0000_50_olm_07-olm-operator.deployment.yaml
Denton24646: Ok good call @njhale I was wondering about that

pkg/controller/registry/reconciler/reconciler.go
```diff
@@ -121,6 +121,10 @@ func Pod(source *v1alpha1.CatalogSource, name string, image string, labels map[s
 						InitialDelaySeconds: livenessDelay,
 					},
 					Resources: v1.ResourceRequirements{
+						Limits: v1.ResourceList{
+							v1.ResourceCPU:    resource.MustParse("100m"),
```
njhale: I think this is fair.

FYI - this does not control the creation of any of the pods for the default catalogs, those pods are generated by marketplace.

kevinrizza: Yep, that is true. It's actually why I thought doing this exercise was a good idea in the first place, since we had to put arbitrary resource limits there before:

https://github.com/operator-framework/operator-marketplace/blob/master/pkg/registry/registry.go#L410

I was hoping to move away from that. I was thinking that once this merges we should maybe update that to match the resource limits described here. We could possibly even consider it part of the same bugzilla?

Denton24646: That works for me, I'm going to be in the marketplace codebase today adding a `NodeSelector` to prevent catalogs from being scheduld onto Windows nodes, so I can add the limits here as well.

Denton24646:
The reason that we see the console test failing is because of the first expensive call that the console makes to the package server to get package information, specifically the operator icons which are encoded in the CSV details of the package. These requests have been taking a long time to complete, and adding resource limits to the package server has only made them take longer. Looking over the logs from the CI cluster for the package server (https://storage.googleapis.com/origin-ci-test/pr-logs/pull/operator-framework_operator-lifecycle-manager/1142/pull-ci-operator-framework-operator-lifecycle-manager-master-e2e-aws-console-olm/1485/artifacts/e2e-aws-console-olm/pods/openshift-operator-lifecycle-manager_packageserver-b8f886865-wxb56_packageserver.log) we see extremely long load times for icons:
```
I1121 19:49:53.930970       1 httplog.go:90] GET /apis/packages.operators.coreos.com/v1/namespaces/openshift-marketplace/packagemanifests/fuse-apicurito/icon?resourceVersion=fuse-apicurito.alpha.apicuritooperator.v7.5.0: (20.496333276s) 200 [Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) HeadlessChrome/76.0.3809.0 Safari/537.36 10.129.0.1:44998]
```
For example this icon request from the front end took 20 seconds to complete. This is a known issue for the console folks. 

I discussed  with @njhale and he has several ideas around improving the performance on the package server side via parallelism and caching, which would improve performance significantly. We can work on this next sprint. In the context of the current release, I'm going to bump up the limits for the package server until our CI passes to get the new resource limits into the release.

njhale:
I am a little concerned w/ how large the limits for `packageserver` are, but since we're looking into optimizing it, and that this is only for OCP, I think we're good to put them in place for now.

/lgtm

spadgett:
I'm still a little worried about this change. Have we done any manual testing to see what kind of performance impact this has on the OperatorHub page in the console? The tests timeout is 60s, which is an eternity for a user to wait. If it's now 30s or 45s, it will pass tests, but that's still a really bad experience. I've also been advised to set only resource requests for cluster components and not limits.

How difficult would it be to rollback the icon changes from 4.2? I believe that's the real issue. It makes the initial request for CSVs smaller, but actually makes performance worse because we fire off so many icon requests. The browser only allows so many requests at a time, and it will block loading of subsequent pages until all the icons load. We switched to lazy loading in https://github.com/openshift/console/pull/3293, but it's still a significant problem.

I am really glad that we're running console tests in this repo or we would have missed this problem entirely, so thank you 👍

Denton24646:
> I'm still a little worried about this change. Have we done any manual testing to see what kind of performance impact this has on the OperatorHub page in the console? The tests timeout is 60s, which is an eternity for a user to wait. If it's now 30s or 45s, it will pass tests, but that's still a really bad experience. I've also been advised to set only resource requests for cluster components and not limits.
> 
> How difficult would it be to rollback the icon changes from 4.2? I believe that's the real issue. It makes the initial request for CSVs smaller, but actually makes performance worse because we fire off so many icon requests. The browser only allows so many requests at a time, and it will block loading of subsequent pages until all the icons load. We switched to lazy loading in [openshift/console#3293](https://github.com/openshift/console/pull/3293), but it's still a significant problem.
> 
> I am really glad that we're running console tests in this repo or we would have missed this problem entirely, so thank you 👍

We are looking into this as a team, thanks for bringing this up - I do think its worthy of some attention now that this PR landed. I will make sure we coordinate with you regarding our decisions. Having the console tests run as part of CI was clutch, absolutely.

njhale:
@spadgett Played around w/ optimizing packageserver. If you get the chance, please LMK if the gains mitigate issues you see on the frontend.

PR #1144, opened 2019-11-20
fix(grpc): tiny bug fixes and logging improvement

jpeeler:
This just adds a return that's added for readability, adds some logging to include the GRPC connection state text, and calls the cancel function for a context to not leak resources.


Denton24646:
I think the cancel func still leaks because the `defer` doesn't get called after each iteration, the execution context stays in the `for` loop. I have an open PR that addresses this leak in #1116 just waiting on a BZ

jpeeler:
Oh good point. Well I'm closing then then as there's not much value here otherwise.

PR #1149, opened 2019-11-22
Bug 1760608: remove resource limits from packageserver

exdx:
After merging #1142 there were some concerns about the response time of the packageserver when the console loads all the operators (specifically regarding loading the icons on the page). See 
https://github.com/operator-framework/operator-lifecycle-manager/pull/1142#issuecomment-557660043 for more info. This PR backs out the newly added limits for the packageserver only.
Note: the load time can be improved by making optimizations to the packageserver and potentially the console. 

Discussion with the team and the console team regarding the proposed limits for the package server. 

**Reviewer Checklist**

### DIFF
--- a/deploy/ocp/values.yaml
+++ b/deploy/ocp/values.yaml
@@ -90,9 +90,6 @@ package:
     effect: NoExecute
     tolerationSeconds: 120
   resources:
-    limits:
-      cpu: 400m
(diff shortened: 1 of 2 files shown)


Denton24646:
Holding to see how #1150 performs, maybe not have to remove the limits after all...

PR #1150, opened 2019-11-23
Bug 1779990: fix(packageserver): cache packagemanifests

njhale:

Generate and maintain a running cache of `PackageManifests`. Serve get/list requests directly from this cache.

Packageserver queries catalog gRPC APIs and constructs `PackageManifest` resources, on the fly, per request. When dealing with large catalogs containing icon images, this can cause requests to be very slow. Poor response times from packageserver can impact the OperatorHub UI's load time and cause its tests to timeout.

**Reviewer Checklist**


njhale:
Playing around with this on minikube w/ community-operators (87 packages total) reveals some pretty promising results running `kubectl get packagemanifests --all-namespaces`

*Before:*
 GET /apis/packages.operators.coreos.com/v1/packagemanifests?limit=500: (3.84524649s) 200 [kubectl/v1.16.3 (darwin/amd64) kubernetes/b3cbbae 192.168.64.13:49110]

After:
GET /apis/packages.operators.coreos.com/v1/packagemanifests?limit=500: (3.679124ms) 200 [kubectl/v1.16.3 (darwin/amd64) kubernetes/b3cbbae 192.168.64.13:44272]

~1000x  faster 😅 

The OperatorHub UI loads almost instantly too.

pkg/package-server/provider/registry.go
```diff
@@ -165,9 +200,107 @@ func (p *RegistryProvider) syncCatalogSource(obj interface{}) (syncError error)
 	p.setClient(newRegistryClient(source, conn), key)
 	logger.Info("new grpc connection added")
 
+	syncError = p.refreshCache(context.Background(), key)
```
ecordell: This is my only real design comment - we don't need to rebuild the cache on every sync. We should only need to do it when the catalogsource changes, which should be indicated by the lastupdatetime.

pkg/package-server/provider/registry.go
```diff
@@ -63,8 +86,13 @@ func NewRegistryProvider(ctx context.Context, crClient versioned.Interface, oper
 		Operator: operator,
 
 		globalNamespace: globalNamespace,
-		clients:         make(map[sourceKey]registryClient),
+		clients:         make(map[types.NamespacedName]registryClient),
+		cache: cache.NewIndexer(cache.MetaNamespaceKeyFunc, cache.Indexers{
```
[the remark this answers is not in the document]
njhale: I ran both `gofmt` and `goimports` and there were no changes to this file, so I think this is correct.

cmd/package-server/main.go
```diff
@@ -3,58 +3,25 @@ package main
...
-
-	options.SecureServing.AddFlags(flags)
-	options.Authentication.AddFlags(flags)
-	options.Authorization.AddFlags(flags)
-	options.Features.AddFlags(flags)
-
-	flags.AddGoFlagSet(flag.CommandLine)
-	flags.Parse(flag.Args())
-}
-
```
njhale: Decided to clean all this up while I'm here.

go.mod
```diff
@@ -43,6 +43,7 @@ require (
 	k8s.io/client-go v8.0.0+incompatible
 	k8s.io/code-generator v0.0.0
 	k8s.io/component-base v0.0.0
+	k8s.io/klog v0.4.0
```
njhale: OLM codepaths using `glog` were brought into the packageserver dependency tree transitively. This was causing flag registration conflicts with `klog`, so I swapped  all references to `glog` for `klog`.

pkg/package-server/provider/registry.go
```diff
@@ -10,108 +10,135 @@ import (
...
+
 // RegistryProvider aggregates several `CatalogSources` and establishes gRPC connections to their registry servers.
 type RegistryProvider struct {
 	queueinformer.Operator
+	runOnce sync.Once
 
-	mu              sync.RWMutex
 	globalNamespace string
-	clients         map[sourceKey]registryClient
+	sources         *registrygrpc.SourceStore
```
njhale: I replaced the local gRPC connection handling with @ecordell's `SourceStore`. It has a cleaner interface, and now with its use here, consolidates gRPC connection management logic across OLM.

pkg/package-server/provider/registry.go
```diff
@@ -10,108 +10,135 @@ import (
...
 
 		globalNamespace: globalNamespace,
-		clients:         make(map[sourceKey]registryClient),
+		cache: cache.NewIndexer(cache.MetaNamespaceKeyFunc, cache.Indexers{
+			cache.NamespaceIndex: cache.MetaNamespaceIndexFunc,
+			catalogIndex:         catalogIndexFunc,
+		}),
 	}
-
-	for _, namespace := range watchedNamespaces {
```
njhale: I got rid of multi-namespace configuration, since other OLM controllers already assume cluster-wide privileges. I also did this to avoid needing to register/access different `CatalogSource` listers for each namespace.

pkg/package-server/provider/registry.go
```diff
@@ -131,43 +158,176 @@ func (p *RegistryProvider) syncCatalogSource(obj interface{}) (syncError error)
...
+	catsrc, err = p.catsrcLister.CatalogSources(key.Namespace).Get(key.Name)
+	if err != nil {
+		return
+	}
 
+	client = newRegistryClient(catsrc, conn)
 	return
 }
 
+func (p *RegistryProvider) refreshCache(ctx context.Context, client *registryClient) error {
```
njhale: Decided to pass this a `*registryClient` instead of a key to make testing easier, since we have pre-existing tests that use a fake of this type to directly inject registry responses.

pkg/package-server/provider/registry_test.go
```diff
@@ -55,7 +56,7 @@ func server() {
 		logrus.Fatal(err)
 	}
 
-	loader := sqlite.NewSQLLoaderForDirectory(load, "manifests")
+	loader := sqlite.NewSQLLoaderForDirectory(load, filepath.Join("testdata", "manifests"))
```
njhale: We've been trying to use the `testdata` convention throughout operator-framework, so I updated it while I was here. I can remove it from the PR, but I don't think it's too risky to leave it.

pkg/package-server/provider/registry.go
```diff
@@ -233,66 +387,52 @@ func (p *RegistryProvider) Get(namespace, name string) (*operators.PackageManife
...
 		}
+		pkgList.Items = append(pkgList.Items, *out)
 	}
 
-	return &operators.PackageManifestList{Items: pkgs}, nil
+	return pkgList, nil
 }
 
-func toPackageManifest(logger *logrus.Entry, pkg *api.Package, client registryClient) (*operators.PackageManifest, error) {
+func newPackageManifest(ctx context.Context, logger *logrus.Entry, pkg *api.Package, client *registryClient) (*operators.PackageManifest, error) {
```
njhale: Plumbed down the context so that cache building can enforce a timeout on packagemanifest generation.

test/e2e/setup_test.go
```diff
@@ -39,9 +39,6 @@ var (
 )
 
 func TestMain(m *testing.M) {
-	if err := flag.Set("logtostderr", "true"); err != nil {
```
njhale: In `klog` this is [true by default](https://github.com/kubernetes/klog/pull/39).

njhale:
/retest

Flakes look related to OAuth.

jpeeler:
/retest

One of the flakes this time was this marketplace one https://bugzilla.redhat.com/show_bug.cgi?id=1780143.

pkg/package-server/provider/registry.go
```diff
@@ -131,43 +158,176 @@ func (p *RegistryProvider) syncCatalogSource(obj interface{}) (syncError error)
...
+
+	var err error
+	switch state.State {
+	case connectivity.Ready:
+		var client *registryClient
+		client, err = p.registryClient(key)
+		if err == nil {
+			err = p.refreshCache(timeout, client)
 		}
+	case connectivity.TransientFailure, connectivity.Shutdown:
```
jpeeler: I wonder if having transient failure here actually adds additional work. Did you find that gc-ing a source and then re-adding was quicker than letting a source transition back to connecting and (hopefully) ready?

pkg/package-server/provider/registry.go
```diff
@@ -131,43 +158,176 @@ func (p *RegistryProvider) syncCatalogSource(obj interface{}) (syncError error)
...
+		pkgName, err := stream.Recv()
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			logger.WithField("err", err.Error()).Warnf("error getting data")
+			break
+		}
+
+		g.Go(func() (pkgErr error) {
```
jpeeler: Nice :)

pkg/package-server/provider/registry.go
```diff
@@ -131,43 +158,176 @@ func (p *RegistryProvider) syncCatalogSource(obj interface{}) (syncError error)
...
+			pkg, err := client.GetPackage(ctx, &api.GetPackageRequest{Name: pkgName.GetName()})
+			if err != nil {
+				logger.WithField("err", err.Error()).Warnf("eliding package: error getting package")
+				return
```
exdx: In the anonymous function you defined a named-error return type `pkgErr` but then don't set `pkgErr` equal to the returned error before returning - does this function return the intended error or nil? Might be simpler to just return the error?

pkg/package-server/provider/registry.go
jpeeler: We discussed offline that having something in the cache in this state is probably not the best, so we'll just leave it as is.

pkg/package-server/provider/registry.go
njhale: Ah, I found I didn't want to return an error anywhere since this should be best effort, and didn't want to write `return nil` 4 times. l like the way errorgroups manage goroutines, but it may not be the best abstraction to use here. Any suggestions, using a waitgroup leaves something to be desired 🙁

exdx: Oh I see what you mean...You're saying based off the `*Group.Go()` signature you need the anonymous function to to return an error (https://godoc.org/golang.org/x/sync/errgroup#Group.Go) but you don't actually want `*Group.Wait()` to return a non-nil error if one does come up. Since `the derived Context is canceled the first time a function passed to Go returns a non-nil error` and this is best effort we don't want that. 

Maybe just have `_ := g.Wait()` instead and comment that the error is being dropped intentionally as this is best effort to make it more clear for a reader? I see what you mean about the abstraction though

jpeeler:
/retest
/lgtm

I did see one comment that might be worth addressing in a follow up. But let's try and get this in today.

pkg/package-server/provider/registry.go
njhale: @exdx So I took a look back at error group and found that some assumptions I made around how it handles context cancellation were wrong. I'll switch this to using `WaitGroup` instead and then we won't be returning an error from the anonymous function anyway. Should have this change up soon.

pkg/package-server/provider/registry.go
```diff
@@ -131,43 +157,176 @@ func (p *RegistryProvider) syncCatalogSource(obj interface{}) (syncError error)
...
+	address := source.Address()
+	logger = logger.WithField("address", address)
+
+	key := resolver.CatalogKey{
+		Namespace: source.GetNamespace(),
+		Name:      source.GetName(),
+	}
+	if sourceMeta := p.sources.GetMeta(key); sourceMeta != nil && sourceMeta.Address == address {
+		// If the address hasn't changed, don't bother creating a new source
+		logger.Debug("catalog address unchanged, skipping source creation")
```
ecordell: Can’t the pod be updated with new content without the address changing?

pkg/package-server/provider/registry.go
```diff
@@ -131,43 +157,176 @@ func (p *RegistryProvider) syncCatalogSource(obj interface{}) (syncError error)
...
+		"state":  state.State,
+	})
+	logger.Debug("source state changed")
+
+	timeout, cancel := context.WithTimeout(context.Background(), cacheTimeout)
+	defer cancel()
+
+	var err error
+	switch state.State {
+	case connectivity.Ready:
```
ecordell: It’s unlikely, but there are cases where you might not receive an event for this (for example, if the connection is already in the ready state before it’s events are watched, or if the states change too rapidly). In practice this doesn’t seem to be an issue for grpc connections, but I wanted to mention it. 

(This is due to relying on the grpc wait for state change, which requires you specify the state you’re waiting to change away from)

It’s not an issue in catalog-operator because the state change is used as an optimization and requeues a catalogsource for processing - if an event is missed like that, a periodic resync will fix it.

pkg/package-server/provider/registry.go
njhale: I'm assuming that, at minimum, the gRPC connection changes state when catalog content is updated. That's probably a bad assumption in general, but I believe it's the behavior of all current registry implementations.

pkg/package-server/provider/registry.go
njhale: Fair -- We can make caching more aggressive in the sync function if you want.

PR #1152, opened 2019-11-25
Add general design goals doc

njhale:

- Add a doc to express some general design goals for new features
- Rename `doc` directory to `docs`

### DIFF
--- a/doc/design/design-goals.md
+++ b/doc/design/design-goals.md
@@ -1,0 +1,23 @@
+# General Design Goals
+
+Whether designing new features or redesigning old ones, OLM should maintain these general design goals:
+
+## Build (reasonably) small, sharp features
+
+All designs should be kept as simple and tightly scoped as is reasonably possible. Where possible, attempt to follow the [unix philosophy](https://en.wikipedia.org/wiki/Unix_philosophy) for writing small, sharp features that compose well. Additionally, when writing a new proposal, always remember to [keep it simple ...student!](https://en.wikipedia.org/wiki/KISS_principle)
+
+## Upstream first
+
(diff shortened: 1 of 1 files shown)


docs/design/design-goals.md
```diff
@@ -0,0 +1,23 @@
+# General Design Goals
+
+Whether designing new features or redesigning old ones, OLM should maintain these general design goals:
+
+## Build (reasonably) small, sharp features
+
+All designs should be kept as simple and tightly scoped as is reasonably possible. Where possible, attempt to follow the [unix philosophy](https://en.wikipedia.org/wiki/Unix_philosophy) for writing small, sharp features that compose well. Additionally, when writing a new proposal, always remember to [keep it simple ...student!](https://en.wikipedia.org/wiki/KISS_principle)
```
kevinrizza: Just thinking out loud: Should we have a more formal process wrt design proposals here? What about https://github.com/operator-framework/enhancements ?

ecordell:
> Rename doc directory to docs

this will require some link updates in places we don't own. @adellape do you think you know where these might be?

adellape:
> this will require some link updates in places we don't own. @adellape do you think you know where these might be?

@ecordell There's a couple in the `openshift-docs` repo that I can update. Other than that, looked in the `openshift/console` repo and only found https://github.com/openshift/console/blob/master/frontend/packages/operator-lifecycle-manager/src/components/clusterserviceversion.tsx#L487. Were there other places that came to mind?

docs/design/design-goals.md
njhale: I agree -- but I want to keep this doc more abstract. Could we add it in a separate `dev-process` doc?

njhale:
@ecordell I removed the doc rename for now. PTAL.

njhale:
/retest

retesting docs 😅

PR #1154, opened 2019-11-27
fix: make run_console_local.sh working on WSLv1

wilmardo:

* Updates the if statement which check for WSL.
* Quoting of variables to mitigate [SC2086](https://github.com/koalaman/shellcheck/wiki/SC2086)
* Make sh compatible
* Removes -it from docker run, no need to start interactive when redirecting output

The script wasn't working for me in WSlv1 because `[[ "$(< /proc/version)" == *"@(Microsoft|WSL)"* ]]` never evaluated to true. 
Simplified the statement by using `uname -r` (no need to grep for Microsoft or WSL) and by using `grep` 

The second commit is for making the script sh compatible to support more environments. I couldn't check it on MacOS but according to [the internet](https://kinoshita.eti.br/2016/11/05/checking-the-operating-system-type-in-shell-script.html) OSTYPE should be defined.
The for loop is needed since the script is much faster in sh and the if statement to check for running is sometimes run before the container has started.

**Reviewer Checklist**

### DIFF
--- a/scripts/run_console_local.sh
+++ b/scripts/run_console_local.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 
 # Colors definition
 readonly RED=$(tput setaf 1)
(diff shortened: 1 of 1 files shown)


ecordell:
/approve

I'd like to double check that this still works on other systems as well, since we don't have CI for this script.

ecordell:
/retest
/lgtm

verified that it still works on macos!

PR #1171, opened 2019-12-04
Bug 1777144: packageserver update fails to adopt APIService

exdx:
This PR does 2 things to address the issue where updating the packageserver fails with a `unable to adopt APIService` message. One is add APIService to the cleanup owner references logic that OLM performs on startup to ensure owner references on APIService objects are correct. ~The other is adding verification on the labels of the APIService object to make sure that the label `olm.owner:packageserver` is set.~ 

OLM already checks several types for owner reference mismatches: mainly rbac types like roles and role bindings. Adding the check for APIService owner references will hopefully resolve the issue during cluster upgrade where the packageserver blocks upgrade because it is unable to set the ownership for its api service. The packageserver CSV fails to resolve with `message: unable to adopt APIService` and this causes cluster upgrade to fail. By checking APIService references on startup this could potentially resolve the issue. ~Owner references are also established by labels on OLM objects, so adding a check for `olm.owner:packageserver` should also help.~ 

**Reviewer Checklist**

### DIFF
--- a/cmd/olm/cleanup.go
+++ b/cmd/olm/cleanup.go
@@ -3,17 +3,23 @@ package main
 import (
 	"time"
 
+	"k8s.io/kube-aggregator/pkg/client/clientset_generated/clientset"
+
 	"github.com/sirupsen/logrus"
 	rbacv1 "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/util/retry"
+	apiregv1 "k8s.io/kube-aggregator/pkg/apis/apiregistration/v1"
 
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/clientset/versioned"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/operators/olm"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/operatorclient"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/ownerutil"
 )
 
 const (
@@ -52,6 +58,10 @@ func cleanup(logger *logrus.Logger, c operatorclient.ClientInterface, crc versio
 
 	if err := cleanupOwnerReferences(c, crc); err != nil {
 		logger.WithError(err).Fatal("couldn't cleanup cross-namespace ownerreferences")
+	}
+
+	if err := ensureAPIServiceLabels(c.ApiregistrationV1Interface()); err != nil {
+		logger.WithError(err).Fatal("couldn't ensure APIService labels")
 	}
 }
 
@@ -171,6 +181,11 @@ func cleanupOwnerReferences(c operatorclient.ClientInterface, crc versioned.Inte
 		objs = append(objs, &obj)
 	}
 
+	apiServices, _ := c.ApiregistrationV1Interface().ApiregistrationV1().APIServices().List(listOpts)
+	for _, obj := range apiServices.Items {
+		objs = append(objs, &obj)
+	}
+
 	for _, obj := range objs {
 		if !removeBadRefs(obj) {
 			continue
@@ -206,6 +221,11 @@ func cleanupOwnerReferences(c operatorclient.ClientInterface, crc versioned.Inte
 				_, err = c.KubernetesInterface().RbacV1().RoleBindings(v.GetNamespace()).Update(v)
 				return err
 			}
+		case *apiregv1.APIService:
+			update = func() error {
+				_, err = c.ApiregistrationV1Interface().ApiregistrationV1().APIServices().Update(v)
+				return err
+			}
 		}
 
 		if err := retry.RetryOnConflict(retry.DefaultBackoff, update); err != nil {
@@ -238,4 +258,36 @@ func crossNamespaceOwnerReferenceRemoval(kind string, uidNamespaces map[types.UI
 
 		return
 	}
+}
+
+// ensureAPIServiceLabels checks the labels of existing APIService objects to ensure ownership is set correctly.
+// If the APIService label does not correspond to the expected packageserver owner the APIService labels are updated
+func ensureAPIServiceLabels(c clientset.Interface) error {
+	apiService, err := c.ApiregistrationV1().APIServices().Get(olm.PackageserverName, metav1.GetOptions{})
+	if err != nil && !k8serrors.IsNotFound(err) {
+		logrus.WithField("apiService", "packageserver").Debugf("get error: %s", err)
+		return err
+	}
+	if apiService == nil || k8serrors.IsNotFound(err) || apiService.Name == "" {
+		logrus.WithField("apiService", "packageserver").Debugf("not found")
+		return nil
+	}
+
+	if l := apiService.GetLabels(); l == nil {
+		apiService.Labels = make(map[string]string)
+	}
+
+	if apiService.Labels[ownerutil.OwnerKey] != ownerutil.OwnerPackageServer {
+		apiService.Labels[ownerutil.OwnerKey] = ownerutil.OwnerPackageServer
+		update := func() error {
+			_, err := c.ApiregistrationV1().APIServices().Update(apiService)
+			return err
+		}
+		if err := retry.RetryOnConflict(retry.DefaultBackoff, update); err != nil {
+			logrus.WithField("apiService", apiService.Name).Warnf("update error: %s", err)
+			return err
+		}
+	}
+
+	return nil
 }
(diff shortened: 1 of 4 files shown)


cmd/olm/cleanup.go
```diff
@@ -171,6 +172,11 @@ func cleanupOwnerReferences(c operatorclient.ClientInterface, crc versioned.Inte
 		objs = append(objs, &obj)
 	}
 
+	apiServices, _ := c.ApiregistrationV1Interface().ApiregistrationV1().APIServices().List(listOpts)
+	for _, obj := range apiServices.Items {
+		objs = append(objs, &obj)
+	}
+
 	for _, obj := range objs {
 		if !removeBadRefs(obj) {
```
njhale: This is only cleaning up actual `OwnerReferences`, and not the label owner refs that cause the issue.

cmd/olm/cleanup_test.go
```diff
@@ -486,10 +496,59 @@ func TestCleanupOwnerReferences(t *testing.T) {
...
+				},
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.description, func(t *testing.T) {
+			listOpts := metav1.ListOptions{}
+			k8sClient := k8sfake.NewSimpleClientset()
+			c := operatorclient.NewClient(k8sClient, apiextensionsfake.NewSimpleClientset(), apiregistrationfake.NewSimpleClientset(tt.fields.apiServices...))
```
exdx: use `apiregistration.Interface` instead to not depend on unneeded interface methods

awgreene:
Nice work on this @exdx , could you add a rewrite the commit message to include:

- A short description of the bug you are fixing.
- How your solution addresses the issue.

awgreene:
@exdx could you also merge your commits into a single commit.

awgreene:
@exdx I saw that the PR description included:

> By checking api service references on startup this could potentially resolve the issue.

Were you able to test this and prove that the proposed solution will fix the bug?

cmd/olm/cleanup.go
```diff
@@ -53,6 +56,10 @@ func cleanup(logger *logrus.Logger, c operatorclient.ClientInterface, crc versio
 	if err := cleanupOwnerReferences(c, crc); err != nil {
 		logger.WithError(err).Fatal("couldn't cleanup cross-namespace ownerreferences")
 	}
+
+	if err := checkAPIServiceLabels(c.ApiregistrationV1Interface()); err != nil {
```
awgreene: Rather than `checkAPIServiceLabels` could we name this `ensureAPIServiceLabels`? To me, `Check` suggests a that a `boolean` would be returned.

cmd/olm/cleanup.go
```diff
@@ -239,3 +256,31 @@ func crossNamespaceOwnerReferenceRemoval(kind string, uidNamespaces map[types.UI
 		return
 	}
 }
+
+// checkAPIServiceLabels checks the labels of existing APIService objects to ensure ownership
+// solves a potential contention issues for packageservers (which own the same APIService) during upgrades
```
awgreene: If you wish to include a note highlighting the problems fixed by the function, could it be more robust?

I would personally suggest just writing what the function does.

cmd/olm/cleanup_test.go
```diff
@@ -540,3 +599,12 @@ func newRoleBinding(namespace, name string, uid types.UID, ownerRefs ...metav1.O
 	roleBinding.SetOwnerReferences(ownerRefs)
 	return roleBinding
 }
+
+func newAPIService(name string, uid types.UID, labels map[string]string, ownerRefs ...metav1.OwnerReference) *apiregv1.APIService {
```
awgreene: Could you explain how the `labels` and `ownerReferences` come into play in this bug in the commit?

exdx:
Thanks for the review @awgreene! Will update to incorporate your great feedback.

cmd/olm/cleanup.go
```diff
@@ -239,3 +258,30 @@ func crossNamespaceOwnerReferenceRemoval(kind string, uidNamespaces map[types.UI
 		return
 	}
 }
+
+// ensureAPIServiceLabels checks the labels of existing APIService objects to ensure ownership is set correctly
+// If APIService  label name does not correspond to the expected packageserver owner the APIService labels are updated
+func ensureAPIServiceLabels(c clientset.Interface) error {
+	fieldSelector := fmt.Sprintf("metadata.name=%s", packageserverName)
+	listOpts := metav1.ListOptions{FieldSelector: fieldSelector}
+	apiServices, err := c.ApiregistrationV1().APIServices().List(listOpts)
```
ecordell: you can just do `APIServices().Get(packageserverName)`

cmd/olm/cleanup.go
```diff
@@ -239,3 +259,33 @@ func crossNamespaceOwnerReferenceRemoval(kind string, uidNamespaces map[types.UI
...
 }
+
+// ensureAPIServiceLabels checks the labels of existing APIService objects to ensure ownership is set correctly
+// If the APIService label does not correspond to the expected packageserver owner the APIService labels are updated
+func ensureAPIServiceLabels(c clientset.Interface) error {
+	apiService, err := c.ApiregistrationV1().APIServices().Get(packageserverName, metav1.GetOptions{})
+	if err != nil && !k8serrors.IsNotFound(err) {
+		return err
+	}
+	if apiService == nil {
```
exdx: My thought process here is to assume this is running on both OCP and vanilla kube. Since running the packageserver and APIService is not a requirement, if there is no APIService it should not be an error. The user could only be using parts of OLM. Does that sound right?

cmd/olm/cleanup.go
```diff
@@ -1,24 +1,30 @@
 package main
 
 import (
+	"fmt"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/ownerutil"
```
ecordell: nit: import ordering (goimports should fix)

cmd/olm/cleanup.go
ecordell: Yep, that seems correct.

exdx:
Something is up with this PR on cluster initialization the e2e fails with `level=fatal msg="failed to initialize the cluster: Cluster operator operator-lifecycle-manager has not yet reported success"`

njhale:
Let's figure out what's going on before we get a false positive and this merges

/hold

exdx:
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/operator-framework_operator-lifecycle-manager/1171/pull-ci-operator-framework-operator-lifecycle-manager-master-e2e-aws-olm/3609/artifacts/e2e-aws-olm/pods/openshift-operator-lifecycle-manager_olm-operator-67d747856d-77wdw_olm-operator.log

pkg/controller/operators/olm/apiservices.go
```diff
@@ -729,6 +731,19 @@ func (a *Operator) installAPIServiceRequirements(desc v1alpha1.APIServiceDescrip
...
+	// set apiservice owner to packageserver for proper cleanup on cluster upgrade
+	// only for packageserver api service
+	if apiService.Name == PackageserverName {
+		if l := apiService.GetLabels(); l != nil {
+			if apiService.Labels[ownerutil.OwnerKey] != ownerutil.OwnerPackageServer {
+				apiService.Labels[ownerutil.OwnerKey] = ownerutil.OwnerPackageServer
+			}
+		} else {
+			apiService.Labels = make(map[string]string)
+			apiService.Labels[ownerutil.OwnerKey] = ownerutil.OwnerPackageServer
```
njhale: nit: you could just assign a map literal

pkg/controller/operators/olm/apiservices.go
```diff
@@ -729,6 +731,19 @@ func (a *Operator) installAPIServiceRequirements(desc v1alpha1.APIServiceDescrip
...
+		}
```
njhale: nit: doesn't `ownerutil.AddOwnerLabels(...)` handle this?

cmd/olm/cleanup.go
```diff
@@ -1,24 +1,30 @@
 package main
 
 import (
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/operators/olm"
```
njhale: nit: import ordering is messed up again

pkg/controller/operators/olm/apiservices.go
njhale: isn't the `OwnerNamespaceKey` also important?

pkg/controller/operators/olm/apiservices.go
exdx: Yes you're right, and we do so on L721. So I'm wondering if we need this code at all....

pkg/controller/operators/olm/apiservices.go
exdx: Yes it is, and is also handled by `ownerutil.AddOwnerLabels(...)` via `OwnerLabel()`
```
// OwnerLabel returns a label added to generated objects for later querying
func OwnerLabel(owner Owner, kind string) map[string]string {
	return map[string]string{
		OwnerKey:          owner.GetName(),
		OwnerNamespaceKey: owner.GetNamespace(),
		OwnerKind:         kind,
	}
}
```

exdx:
It looks like the packageserver owner labels to the api service should already be set when creating the APIService (we were not sure originally if they were being set correctly). That being said, the addition of the cleanup code should still be helpful in resolving the original bug on cluster upgrade?

Also the bug was targeting a 4.2.z release for a 4.1 -> 4.2 upgrade, need to check that this all still applies.

PR #1175, opened 2019-12-05
BUG: Fix CI Unit tests.

awgreene:
Testing a unit test failure

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,12 @@
 #  OLM - Build and Test  #
 ##########################
 
+# Undefine GOFLAGS environment variable.
+ifdef GOFLAGS
+$(warning Undefining GOFLAGS set in CI)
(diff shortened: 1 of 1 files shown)


jpeeler:
I don't think this is the correct fix, at least in all cases. I attempted to run the unit test target locally in a container with networking disabled and it did not work. I suspect that somewhere in CI the mod setting is getting set, which means it'll work in CI, but won't work correctly for local development (but might appear to at first due to modules being downloaded on demand).

jpeeler:
This is what changed: https://github.com/openshift/release/pull/6226. They are discouraging plumbing the command flag through all the commands... but we've already done all of that.

jpeeler:
/retest
Assuming CI failure and that this change didn't break the world.

PR #1185, opened 2019-12-11
Add note for version field

clyang82:
Signed-off-by: Chun Lin Yang <clyang@cn.ibm.com>

This change is to add a note if you have `version` field defined, then it must match the first item of `versions`. `operator-sdk olm-catalog gen-csv --csv-version 0.10.0 --update-crds` generates a `version` field. it is error-prone without a note here.

**Reviewer Checklist**

### DIFF
--- a/doc/design/dependency-resolution.md
+++ b/doc/design/dependency-resolution.md
@@ -35,6 +35,19 @@ versions:
     storage: true
 ```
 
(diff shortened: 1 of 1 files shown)


ecordell:
/lgtm

doc-only, merging without tests

PR #1186, opened 2019-12-11
 reduce GRPC transient failures

jpeeler:
This adds an additional function in the syncCatalogSource sync chain to ensure the GRPC readiness probes are reporting as ready before attempting any GPRC operations.

The idea is to reduce transient errors as reported on the catalog source status. Unfortunately, I've only been able to reliably improve the catalog sources that are of type grpc and not the ones that are backed by configmaps (which is what the reported bug is specifically about).

There's also some work done to make debugging easier in the future.

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -143,7 +143,7 @@ func NewOperator(ctx context.Context, kubeconfigPath string, clock utilclock.Clo
 		bundleUnpackerImage:      configmapRegistryImage, // Assume the configmapRegistryImage contains the unpacker for now.
 	}
 	op.sources = grpc.NewSourceStore(logger, 10*time.Second, 10*time.Minute, op.syncSourceState)
-	op.reconciler = reconciler.NewRegistryReconcilerFactory(lister, opClient, configmapRegistryImage, op.now)
+	op.reconciler = reconciler.NewRegistryReconcilerFactory(lister, opClient, configmapRegistryImage, op.now, op.logger)
 
 	// Wire OLM CR sharedIndexInformers
 	crInformerFactory := externalversions.NewSharedInformerFactoryWithOptions(op.client, resyncPeriod)
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -19,6 +19,7 @@ import (
 	extinf "k8s.io/apiextensions-apiserver/pkg/client/informers/externalversions"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	utilclock "k8s.io/apimachinery/pkg/util/clock"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
@@ -363,8 +364,7 @@ func (o *Operator) syncSourceState(state grpc.SourceState) {
 
 	o.logger.Infof("state.Key.Namespace=%s state.Key.Name=%s state.State=%s", state.Key.Namespace, state.Key.Name, state.State.String())
 
-	switch state.State {
-	case connectivity.Ready:
+	if state.State == connectivity.Ready {
 		if o.namespace == state.Key.Namespace {
 			namespaces, err := index.CatalogSubscriberNamespaces(o.catalogSubscriberIndexer,
 				state.Key.Name, state.Key.Namespace)
@@ -377,11 +377,11 @@ func (o *Operator) syncSourceState(state grpc.SourceState) {
 		}
 
 		o.nsResolveQueue.Add(state.Key.Namespace)
-	default:
-		if err := o.catsrcQueueSet.Requeue(state.Key.Namespace, state.Key.Name); err != nil {
-			o.logger.WithError(err).Info("couldn't requeue catalogsource from catalog status change")
-		}
 	}
+	if err := o.catsrcQueueSet.Requeue(state.Key.Namespace, state.Key.Name); err != nil {
+		o.logger.WithError(err).Info("couldn't requeue catalogsource from catalog status change")
+	}
+	return
 }
 
 func (o *Operator) requeueOwners(obj metav1.Object) {
@@ -631,6 +631,7 @@ func (o *Operator) syncConnection(logger *logrus.Entry, in *v1alpha1.CatalogSour
 
 		// Set connection status and return.
 		updateConnectionStateFunc(out, source)
+		return
 	}
 
 	// connection is already good, but we need to update the sync time
@@ -640,6 +641,47 @@ func (o *Operator) syncConnection(logger *logrus.Entry, in *v1alpha1.CatalogSour
 		out.Status.GRPCConnectionState.LastObservedState = source.ConnectionState.String()
 	}
 
+	return
+}
+
+func (o *Operator) checkBackingPodStatus(logger *logrus.Entry, in *v1alpha1.CatalogSource) (out *v1alpha1.CatalogSource, continueSync bool, syncError error) {
+	if in.Spec.Address != "" {
+		// if this is an imageless source type, there's no backing pod to check
+		return in, true, nil
+	}
+
+	out = in.DeepCopy()
+
+	selector := reconciler.CatalogSourceLabelForPod(in.GetName())
+	pods, err := o.lister.CoreV1().PodLister().Pods(in.GetNamespace()).List(labels.SelectorFromSet(selector))
+	if err != nil {
+		continueSync = false
+		syncError = fmt.Errorf("error while examining catalog source pod: %v", err)
+		return
+	}
+
+	if len(pods) > 0 {
+		// this assumes a 1:1 mapping for catalog sources to pods
+		pod := pods[0]
+		for _, cond := range pod.Status.Conditions {
+			if cond.Type != corev1.ContainersReady {
+				continue
+			}
+
+			if cond.Status == corev1.ConditionTrue {
+				continueSync = true
+			} else {
+				logger.Infof("backing pod '%s' not yet ready", pod.GetName())
+			}
+			break
+		}
+	} else {
+		continueSync = false
+		syncError = fmt.Errorf("did not find expected backing pod")
+		return
+	}
+
+	o.catsrcQueueSet.RequeueAfter(in.GetNamespace(), in.GetName(), 2*time.Second)
 	return
 }
 
@@ -701,6 +743,7 @@ func (o *Operator) syncCatalogSources(obj interface{}) (syncError error) {
 	chain := []CatalogSourceSyncFunc{
 		o.syncConfigMap,
 		o.syncRegistryServer,
+		o.checkBackingPodStatus,
 		o.syncConnection,
 	}
 
--- a/pkg/controller/registry/reconciler/reconciler.go
+++ b/pkg/controller/registry/reconciler/reconciler.go
@@ -5,6 +5,7 @@ import (
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/operatorclient"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/operatorlister"
+	"github.com/sirupsen/logrus"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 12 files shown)


pkg/controller/operators/catalog/operator.go
```diff
@@ -622,6 +628,35 @@ func (o *Operator) syncConnection(logger *logrus.Entry, in *v1alpha1.CatalogSour
...
+	selector := map[string]string{reconciler.CatalogSourceLabelKey: in.GetName()}
+	pods, err := o.lister.CoreV1().PodLister().Pods(in.GetNamespace()).List(labels.SelectorFromSet(selector))
+	if err != nil {
+		continueSync = false
+		syncError = fmt.Errorf("error while examining catalog source pod: %v", err)
+		return
+	}
+
+	if len(pods) > 0 {
+		for _, cond := range pods[0].Status.Conditions {
```
exdx: Here we are getting an array of pods but only checking the condition of the pod with index 0 so the first one. Can we can always expect to get one pod in the `pods` list, or if the first one is ready then the others must be as well? If we only look at the one, can we use a `Get` instead of a `List` when making the request on L635?

pkg/controller/registry/reconciler/reconciler.go
```diff
@@ -142,5 +146,18 @@ func Pod(source *v1alpha1.CatalogSource, name string, image string, labels map[s
 			},
 		},
 	}
+
+	if debugLogging == true {
```
exdx: nice

pkg/controller/registry/reconciler/reconciler.go
```diff
@@ -47,25 +47,28 @@ type registryReconcilerFactory struct {
 	Lister               operatorlister.OperatorLister
 	OpClient             operatorclient.ClientInterface
 	ConfigMapServerImage string
+	debugLogging         bool
 }
 
 // ReconcilerForSource returns a RegistryReconciler based on the configuration of the given CatalogSource.
-func (r *registryReconcilerFactory) ReconcilerForSource(source *v1alpha1.CatalogSource) RegistryReconciler {
+func (r *registryReconcilerFactory) ReconcilerForSource(source *v1alpha1.CatalogSource, debugLogging bool) RegistryReconciler {
```
njhale: Since `debugLogging` is already a member of `registryReconcilerFactory`, I don't think we need this parameter.

pkg/controller/operators/catalog/operator.go
jpeeler: Hrm, good point. I'm pretty sure the mapping is always 1:1...

pkg/controller/operators/catalog/operator.go
```diff
@@ -622,6 +628,35 @@ func (o *Operator) syncConnection(logger *logrus.Entry, in *v1alpha1.CatalogSour
...
+				return
+			}
+		}
+	} else {
+		continueSync = false
+		syncError = fmt.Errorf("did not find expected backing pod")
+		return
+	}
+
+	logger.Infof("backing pod not yet ready")
```
exdx: maybe we could add `"pod status %s", pods[0].Status.Message` to this log message?

test/e2e/catalog_e2e_test.go
```diff
@@ -761,3 +767,98 @@ func replicateCatalogPod(t *testing.T, c operatorclient.ClientInterface, crc ver
...
+	// Delete the registry pod (part of cleaning up)
+	defer cleaner.NotifyTestComplete(t, true)
+	sourceName := genName("catalog-")
+
+	// Create gRPC CatalogSource using an external registry image (community-operators)
+	source := &v1alpha1.CatalogSource{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       v1alpha1.CatalogSourceKind,
+			APIVersion: v1alpha1.CatalogSourceCRDAPIVersion,
+		},
```
njhale: Are there issues if the `TypeMeta` isn't added up front?

test/e2e/catalog_e2e_test.go
```diff
@@ -761,3 +767,98 @@ func replicateCatalogPod(t *testing.T, c operatorclient.ClientInterface, crc ver
...
+		Spec: v1alpha1.CatalogSourceSpec{
+			SourceType: v1alpha1.SourceTypeGrpc,
+			Image:      communityOperatorsImage,
+		},
+	}
+
+	crc := newCRClient(t)
+	selector := labels.SelectorFromSet(map[string]string{reconciler.CatalogSourceLabelKey: sourceName})
+	catSrcWatcher, err := crc.OperatorsV1alpha1().CatalogSources(testNamespace).Watch(metav1.ListOptions{LabelSelector: selector.String()})
```
njhale: Nice! You can also use a field selector on the name instead of using a label.

test/e2e/catalog_e2e_test.go
```diff
@@ -761,3 +767,98 @@ func replicateCatalogPod(t *testing.T, c operatorclient.ClientInterface, crc ver
...
+	require.NoError(t, err)
+
+	// Monitor GRPC connection status updates
+	ctx, cancel := context.WithCancel(context.Background())
+	// cancel called in defer func below
+	done := make(chan struct{})
+	errExit := make(chan error)
```
njhale: nit: defer closing these channels

exdx:
Interesting how this only works for image based catalog sources but not those based off config maps. The only real difference for the configmap based ones is that we build have to seed the DB manually from the configmap before serving gRPC connections. I wonder if they gRPC healthprobe is trying to query the pod before that action is complete, and after a series of failures it starts to back off and act funny. I know we were looking at something like this in the BZ. Maybe we can test the configmap case with a delay in the readiness probe (until after the gRPC connection is ready) and see if that improves behavior?

test/e2e/catalog_e2e_test.go
```diff
@@ -761,3 +767,98 @@ func replicateCatalogPod(t *testing.T, c operatorclient.ClientInterface, crc ver
...
+					if catSrc.Status.GRPCConnectionState != nil {
+						if catSrc.Status.GRPCConnectionState.LastObservedState == connectivity.Ready.String() {
+							done <- struct{}{}
+						}
+						require.NotEqual(t, connectivity.TransientFailure.String(), catSrc.Status.GRPCConnectionState.LastObservedState)
+					}
+				}
+
+			}
+		}
```
njhale: Since the watch was started earler, I think this can be merged with the `for {...}` at the bottom of this test.

pkg/controller/operators/catalog/operator.go
```diff
@@ -142,7 +143,7 @@ func NewOperator(ctx context.Context, kubeconfigPath string, clock utilclock.Clo
 		clientAttenuator:         scoped.NewClientAttenuator(logger, config, opClient, crClient),
 	}
 	op.sources = grpc.NewSourceStore(logger, 10*time.Second, 10*time.Minute, op.syncSourceState)
-	op.reconciler = reconciler.NewRegistryReconcilerFactory(lister, opClient, configmapRegistryImage, op.now)
+	op.reconciler = reconciler.NewRegistryReconcilerFactory(lister, opClient, configmapRegistryImage, op.now, op.logger.IsLevelEnabled(logrus.DebugLevel))
```
ecordell: I would lean towards just passing in a logrus.Entry instance instead of a debug boolean

pkg/controller/operators/catalog/operator.go
```diff
@@ -622,6 +628,35 @@ func (o *Operator) syncConnection(logger *logrus.Entry, in *v1alpha1.CatalogSour
 	return
 }
 
+func (o *Operator) waitForBackingPod(logger *logrus.Entry, in *v1alpha1.CatalogSource) (out *v1alpha1.CatalogSource, continueSync bool, syncError error) {
```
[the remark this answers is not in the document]
jpeeler: I changed it to checkBackingPodStatus.

pkg/controller/operators/catalog/operator.go
```diff
@@ -680,6 +715,7 @@ func (o *Operator) syncCatalogSources(obj interface{}) (syncError error) {
 	chain := []CatalogSourceSyncFunc{
 		o.syncConfigMap,
 		o.syncRegistryServer,
+		o.waitForBackingPod,
```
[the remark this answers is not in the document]
jpeeler: I honestly totally forgot about catalog sources of that type. I ended up adding a simple early return for that scenario.

pkg/controller/operators/catalog/operator.go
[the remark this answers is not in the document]
jpeeler: It could be, but I'm not sure it should be. With the way the code is structured and higher level operations (which I consider pod checking to be) present in the main operator code, it seems best to me where it is. There is a comment to add memoization for ReconcileForSource, which if done later on would make the requested refactoring a bit less invasive.

pkg/controller/operators/catalog/operator.go
```diff
@@ -622,6 +628,35 @@ func (o *Operator) syncConnection(logger *logrus.Entry, in *v1alpha1.CatalogSour
...
+		return
+	}
+
+	if len(pods) > 0 {
+		for _, cond := range pods[0].Status.Conditions {
+			if cond.Type == corev1.ContainersReady && cond.Status == corev1.ConditionTrue {
+				continueSync = true
+				return
+			}
+		}
```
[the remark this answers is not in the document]
jpeeler: To me, this might be more appropriate if the shared code (in this case pod status checking) was not shared by the majority of catalog source types (configmap, grpc). The recently added if statement to ignore grpc address types isn't too ugly, do you agree?

pkg/controller/registry/reconciler/reconciler.go
jpeeler: Oops, I double plumbed.

test/e2e/catalog_e2e_test.go
jpeeler: No, I just copied it from another test.

test/e2e/catalog_e2e_test.go
jpeeler: Since awaitPods also uses a label selector, will stick with consistency.

test/e2e/catalog_e2e_test.go
jpeeler: I don't want to miss any state changes, so I really want to start monitoring the watch before the catalog source is created. Since the result chan is unbuffered, I think that's the only safe way to do it, right?

pkg/controller/operators/catalog/operator.go
jpeeler: Turns out that Get can only be queried by name, so I'm sticking with List here.

pkg/controller/operators/catalog/operator.go
jpeeler: Maybe a good compromise is to indicate the pod name here? The message seems to be blank in my testing and printing the whole status block is going to be too much.

pkg/controller/operators/catalog/operator.go
jpeeler: Is the reason why so you can be able to have access to the operator logger in places that aren't currently possible?

jpeeler: I've done so now - at least to the level of generating the pod spec.

pkg/controller/operators/catalog/operator.go
```diff
@@ -363,6 +364,10 @@ func (o *Operator) syncSourceState(state grpc.SourceState) {
 		if err := o.catsrcQueueSet.Requeue(state.Key.Namespace, state.Key.Name); err != nil {
 			o.logger.WithError(err).Info("couldn't requeue catalogsource from catalog status change")
 		}
+		return
+	}
+	if err := o.catsrcQueueSet.Requeue(state.Key.Namespace, state.Key.Name); err != nil {
```
ecordell: isn't this just requeuing after ever state change? is that desired?

pkg/controller/operators/catalog/operator.go
```diff
@@ -622,6 +628,43 @@ func (o *Operator) syncConnection(logger *logrus.Entry, in *v1alpha1.CatalogSour
...
 
+func (o *Operator) checkBackingPodStatus(logger *logrus.Entry, in *v1alpha1.CatalogSource) (out *v1alpha1.CatalogSource, continueSync bool, syncError error) {
+	if in.Spec.Address != "" {
+		// if this is an imageless source type, there's no backing pod to check
+		return in, true, nil
+	}
+
+	out = in.DeepCopy()
+
+	selector := map[string]string{reconciler.CatalogSourceLabelKey: in.GetName()}
```
ecordell: shouldn't this use the pod decorator to get this selector?

pkg/controller/operators/catalog/operator.go
jpeeler: I actually do so that the status stays in sync with what the grpc connection is reporting. This only adds additional requeues when the connection is reported as ready as requeues were already being done otherwise. I could refactor since a lot of the above is similar if you agree.

pkg/controller/operators/catalog/operator.go
jpeeler: I didn't see a way to have access to the decorator at this level, so I refactored to put some commonality into the reconciler package and use that instead.

pkg/controller/operators/catalog/operator.go
jpeeler: I went ahead and refactored it.

jpeeler:
/retest
If this doesn't pass in a week or so, I'll probably just close it. But if does pass, I think it should be merged!

PR #1194, opened 2019-12-12
Bug 1781366: feat(resolver): fallback to csv parsing if grcp api does not contain info

ecordell:
Falls back to parsing the CSV in the grpc api response if expected fields are not present.

Recent changes to the grpc api pulls out information from the CSV on load time via OPM and shares it. OLM is looking in those new fields to find it. If an older registry is used against a newer OLM, OLM will see packages that have no required/provided apis and will fail to resolve dependencies.

No official content was ever shipped with this combination, but users building their own custom catalogs will hit this. 

Note: this change is also included in #1105 

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/registry/resolver/operators.go
+++ b/pkg/controller/registry/resolver/operators.go
@@ -257,7 +257,6 @@ func NewOperatorFromBundle(bundle *api.Bundle, replaces string, startingCSV stri
 	if err != nil {
 		v = nil
 	}
-
 	provided := APISet{}
 	for _, gvk := range bundle.ProvidedApis {
 		provided[registry.APIKey{Plural: gvk.Plural, Group: gvk.Group, Kind: gvk.Kind, Version: gvk.Version}] = struct{}{}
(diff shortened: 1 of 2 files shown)


pkg/controller/registry/resolver/operators.go
```diff
@@ -266,6 +265,34 @@ func NewOperatorFromBundle(bundle *api.Bundle, replaces string, startingCSV stri
...
 	}
+	sourceInfo := &OperatorSourceInfo{
+		Package:     bundle.PackageName,
+		Channel:     bundle.ChannelName,
+		StartingCSV: startingCSV,
+		Catalog:     sourceKey,
+	}
+
+	// legacy support - if the grpc api doesn't contain the information we need, fallback to csv parsing
+	if len(required) == 0 && len(provided) == 0 {
```
njhale: Just a thought -- but if JSONPath can let us query the provided and required APIs w/o unmarshaling the entire CSV, should we use it to more accurately verify the content of the bundle fields? 

(this question shouldn't block the PR)

ecordell: I was thinking that if the grpc has any provide/required apis, we just trust that (gives us more freedom on the side of the registry going forward).

njhale:
@spadgett Seeing some more console e2e failures here.

PR #1197, opened 2019-12-13
[release-4.3] Bug 1780378: Implement Structural Schema for CRDs

awgreene:
oc explain recently broke when OLM was rebased on top of version 1.16 of Kubernetes. This is due to the introduction of [CRDs Structural Schemas]([1] https://kubernetes.io/blog/2019/06/20/crd-structural-schema/). This PR updates OLMs CustomResourceDefinitions to include a structural schema.


ecordell:
@awgreene IIRC, we just need to include the v1alpha2 OperatorGroup version and this is good to go

awgreene:
@ecordell Regarding [your comment](https://github.com/operator-framework/operator-lifecycle-manager/pull/1197#issuecomment-568093283):

> @awgreene IIRC, we just need to include the v1alpha2 OperatorGroup version and this is good to go

This is a backport of a solution, should I create a new Bugzilla to add support for v1alpha2 to both 4.4 and 4.3? If so - could this PR be merged and a followup for the comment in the same release cycle.

awgreene:
@knobunc could you please consider applying the cherry-pick approved label to this PR?

Without this PR:
- OLM's CRDs are structures blobs, which introduces security concerns. 
- CSVs created on a 4.3 cluster may fail validation once this PR is merged. It would be best to introduce this change as quickly as possible to affect as few operators as possible.

PR #1207, opened 2019-12-20
fix(deploy): remove resource limits

ecordell:
Removes resource limits on OLM pods

OLM resource usages scales with the size of a cluster. Having limits means that OLM can be killed by the scheduler on large clusters even though it is behaving normally.

**Reviewer Checklist**


deploy/chart/values.yaml
```diff
@@ -1,4 +1,3 @@
-installType: upstream
```
awgreene: Why is this removed?

ecordell: This was duplicated - the same key/value is down below

PR #1209, opened 2019-12-26
Remove `optionalProperties` from the CSV definition.

mcanevet:
Remove `optionalProperties` from the CSV definition.

This key is silently ignored and the generated object in the API does not contains it, which make subsequent apply of this yaml file not idempotent, and also if you want to manage OLM using ArgoCD, this object will never be in sync.

**Reviewer Checklist**


ecordell:
Thanks for finding this issue @mcanevet -

Would you mind making the change in `deploy/chart/template`, running `make release`, and committing only the files in `deploy/chart/template` and `manifests`?

The current changes will only apply to an existing release of OLM, the above steps will ensure it makes it into all future releases.

mcanevet:
@ecordell it looks like it has already been fixed by 62be3ee141ca5124d03d4fb15049312023f62d28

PR #1215, opened 2020-01-06
Unpack bundles

njhale:

A refactor and continuation of #1081

Simplification of InstallPlan updates and unpacking job management.

**Reviewer Checklist**


pkg/controller/bundle/bundle_unpacker.go
```diff
@@ -0,0 +1,388 @@
...
+	}
+	role.Rules = append(role.Rules, rule)
+
+	role, err = c.client.RbacV1().Roles(role.GetNamespace()).Update(role)
+
+	return
+}
+
+func (c *configmapUnpacker) ensureRoleBinding(roleNamespace string) (roleBinding *rbacv1.RoleBinding, err error) {
+	fresh := &rbacv1.RoleBinding{
```
ecordell: shouldn't these ensured resources have some reasonable ownerrefs?

pkg/controller/bundle/bundle_unpacker.go
```diff
@@ -0,0 +1,388 @@
...
+		},
+		ResourceNames: []string{
+			configmapName,
+		},
+	}
+	fresh := &rbacv1.Role{
+		Rules: []rbacv1.PolicyRule{rule},
+	}
+	fresh.SetNamespace(configmapNamespace)
+	fresh.SetName(c.rbacName)
```
ecordell: since this uses the `rbacName` set for the bundle unpacker, won't this cause an issue for multiple bundle unpacks in the same namespace at the same time? Shouldn't the role name be based on the `configmapName` in some way (or generated with labels/ownerrefs?)

njhale: Instead of generating a role per bundle, we can aggregate rules on a single role that gets updated. This should shake out to be eventually consistent in the event of contention between concurrent bundle. It's simple enough to switch it over -- I'll do that.

pkg/controller/bundle/bundle_unpacker.go
njhale: Okay. I'll add ownerrefs in the direction of Role/RoleBindings/Jobs -> ConfigMap. 

The ConfigMap itself is a bit more tricky since there can exist InstallPlans that effectively reference it across several namespaces. The sensible thing to me is to add an ownerref in the direction of ConfigMap -> CatalogSource, and then propagate an error condition on the BundleLookup if it's not present when ensuring the ConfigMap.

pkg/controller/bundle/bundle_unpacker.go
ecordell: Isn't this the rolebinding here?

pkg/controller/bundle/bundle_unpacker.go
```diff
@@ -0,0 +1,388 @@
...
+
+	var job *batchv1.Job
+	job, err = c.ensureJob(cm.GetNamespace(), cm.GetName(), result.Path)
+	if err != nil {
+		return
+	}
+
+	cond := result.GetCondition(operatorsv1alpha1.BundleLookupPending)
+	now := c.now()
+	if job == nil || len(job.Status.Conditions) == 0 || len(job.Status.Conditions) > 0 && job.Status.Conditions[0].Type != batchv1.JobComplete {
```
ecordell: non-blocking: there may be some nice ways to clean this up a bit, e.g. `job.Status.ConditionMatches(operatorsv1alpha1.BundleLookupPending, func(Condition) bool { } )`

pkg/controller/bundle/bundle_unpacker.go
ecordell: > there can exist InstallPlans that effectively reference it across several namespaces

Can you describe the case where this happens? Won't everything be scoped to one namespace?

Also - ownerrefs may not be needed at all if OLM will perform the cleanup (which might be nice for the rbac; it's not really needed after the job runs)

pkg/controller/bundle/bundle_unpacker.go
njhale: It should end up being a separate `PolicyRule` in the same `Role`.

pkg/controller/registry/resolver/resolver.go
```diff
@@ -93,26 +100,20 @@ func (r *OperatorsV1alpha1Resolver) ResolveSteps(namespace string, sourceQuerier
 		}
 
 		// add steps for any new bundle
-		if op.Bundle() != nil {
-			bundleSteps, err := NewStepResourceFromBundle(op.Bundle(), namespace, op.Replaces(), op.SourceInfo().Catalog.Name, op.SourceInfo().Catalog.Namespace)
+		if op.Bundle().BundlePath == "" {
```
ecordell: double check - can `op.Bundle()` be nil?

pkg/controller/registry/resolver/resolver.go
```diff
@@ -122,14 +123,25 @@ func (r *OperatorsV1alpha1Resolver) ResolveSteps(namespace string, sourceQuerier
 			}
 		}
 
+		if op.Bundle().BundlePath != "" {
```
ecordell: same - double check that op.Bundle() can't return nil

pkg/lib/queueinformer/resourcequeue.go
```diff
@@ -68,6 +69,29 @@ func (r *ResourceQueueSet) Requeue(namespace, name string) error {
 	return fmt.Errorf("couldn't find queue for resource")
 }
 
+// TODO: this may not actually be required if the requeue is done on the namespace rather than the installplan
+// RequeueAfter requeues the resource in the set with the given name and namespace (just like Requeue), but only does so after duration has passed
+func (r *ResourceQueueSet) RequeueAfter(namespace, name string, duration time.Duration) error {
```
ecordell: I don't see this used anywhere?

pkg/controller/registry/resolver/resolver_test.go
```diff
@@ -449,14 +452,15 @@ func TestNamespaceResolverRBAC(t *testing.T) {
...
 			clientFake, informerFactory, _ := StartResolverInformers(namespace, stopc, tt.clusterState...)
 			lister := operatorlister.NewLister()
 			lister.OperatorsV1alpha1().RegisterSubscriptionLister(namespace, informerFactory.Operators().V1alpha1().Subscriptions().Lister())
 			lister.OperatorsV1alpha1().RegisterClusterServiceVersionLister(namespace, informerFactory.Operators().V1alpha1().ClusterServiceVersions().Lister())
 
-			resolver := NewOperatorsV1alpha1Resolver(lister, clientFake)
+			resolver := NewOperatorsV1alpha1Resolver(lister, clientFake, kClientFake)
 			querier := NewFakeSourceQuerier(map[CatalogKey][]*api.Bundle{catalog: tt.bundlesInCatalog})
-			steps, subs, err := resolver.ResolveSteps(namespace, querier)
+			steps, _, subs, err := resolver.ResolveSteps(namespace, querier)
```
ecordell: should have some tests that return bundle lookups, right?

pkg/lib/queueinformer/resourcequeue.go
njhale: This is being used in the [installplan sync](https://github.com/njhale/operator-lifecycle-manager/blob/unpack-bundles/pkg/controller/operators/catalog/operator.go#L1139).

pkg/controller/registry/resolver/resolver_test.go
njhale: yup! good catch.

pkg/controller/registry/resolver/operators.go
```diff
@@ -263,7 +264,11 @@ func NewOperatorFromBundle(bundle *api.Bundle, startingCSV string, sourceKey Cat
 	if len(required) == 0 && len(provided) == 0 {
 		// fallback to csv parsing
 		if bundle.CsvJson == "" {
-			return nil, fmt.Errorf("couldn't parse bundle")
+			if bundle.GetBundlePath() != "" {
```
ecordell: this doesn't seem like the right place - if we're falling back to CSV parsing, we don't care if there's a bundlepath at all.

njhale: I'll admit it seems odd, but I think we still want to throw a specific error when a bundlepath isn't accompanied by the expected metadata. WDYT?

PR #1224, opened 2020-01-12
Add alm-examples-metadata

Jamstah:
This is a suggested solution to be able to add extended metadata to the `alm-examples`. It is intended to start a discussion about how this data could be stored, so feel free to say if there's a better way that exists/could exist that I don't know because I'm fairly new to operators.

I'm also not sure that this is the official spec for CSVs, but I couldn't find a better alternative. That could be my poor google skills :)

> Add extended metadata for `alm-examples` to be able to extend the capabilities of UI creating CRs. Can be used to identify, for example, different backend technologies (eg storage), different HA options, different deployment sizes (small, medium, large, or example workload based sizings)

Add new field for `alm-examples-metadata`.

We are writing a UI to create PRs based on CRDs and associated CSV data. We want to be able to desribe the example CRs in more detail to the end user, and feel that the right place to store the data is in the CSV alongside the examples

**Reviewer Checklist**

### DIFF
--- a/doc/design/building-your-csv.md
+++ b/doc/design/building-your-csv.md
@@ -169,16 +169,17 @@ The Lifecycle Manager will check against the available CRDs and Operators in the
       description: Represents a cluster of etcd nodes.
 ```
 ## CRD Templates
-Users of your Operator will need to be aware of which options are required vs optional. You can provide templates for each of your CRDs with a minimum set of configuration as an annotation named `alm-examples`. Compatible UIs will pre-enter this template for users to further customize.
+Users of your Operator will need to be aware of which options are required vs optional. You can provide templates for each of your CRDs with a minimum set of configuration as an annotation named `alm-examples`. Metadata for each template, for exmaple an expanded description, can be included in an annotation named `alm-examples-metadata`, which should match the ordering of the `alm-examples` list. Compatible UIs will pre-enter the `alm-examples` template for users to further customize, and use the `alm-examples-metadata` to help users decide which template to select.
 
 The annotation consists of a list of the `kind`, eg. the CRD name, and the corresponding `metadata` and `spec` of the Kubernetes object. Here’s a full example that provides templates for `EtcdCluster`, `EtcdBackup` and `EtcdRestore`:
 
 ```yaml
 metadata:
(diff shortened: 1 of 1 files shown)


ecordell:
Thanks for the PR and idea!

We have been considering something along the lines of this ConsoleYamlSample CRD instead of annotations: https://github.com/openshift/api/blob/master/console/v1/0000_10_consoleyamlsample.crd.yaml

Would something like that address your UI needs as well? The annotation is fine, but the existing `alm-examples` is a bit unwieldy, and this seems to continue in that direction.

Jamstah:
I think that would work for us, but it raises questions:
- How are the samples created, are they installed with the operator?
- What loads the sample CRD, is that with the openshift console operator?
- This seems to move data away from the CSV and into an openshift console specific thing. I suggested the annotation here to keep it together, and support non-openshift clusters. Although I 100% agree this method is unwieldy as is!

chrisbygrave:
We agree that the existing `alm-examples` gets unwieldy at scale, and we're interested to work together on something better.

In the meantime, we'd still like to pursue the peer annotation `alm-examples-metadata` as a stop-gap. We envision this taking the format:

`alm-examples`
```
[{
   metadata: {
      name: 'alm-example-1'
   }
}, {
   metadata: {
      name: 'alm-example-2'
   }
}]
```

`alm-examples-metadata`
```
{
   'example-name-1': {
      description: 'something about example-name-1',
      foo: 'some other arbitrary metadata, such as sizing content'
   },
   'example-name-2': {
      description: 'something about example-name-2',
      foo: 'some other arbitrary metadata, such as sizing content'
   }
}
```

So, the metadata annotation is a hash, keyed of the example name. This requires each example requiring unique metadata to have a unique name. The reason for this being a hash is that the metadata annotation has no need to know the ordering of the examples header (which is not something SDK users have direct control of)

ecordell:
This seems reasonable as a stopgap, and is only documentation. 

/lgtm

Jamstah:
@ecordell Help, bot fight!

njhale:
/ok-to-test

Wow, that's some hot-loop.

PR #1161, opened 2019-12-02
feat: catalog source polling for updated registry image

exdx:
**Description of the change:**
Branch consolidates implementation of catalog polling feature that continues from #1109. 

**Motivation for the change:**
New feature request. Part of https://github.com/openshift/enhancements/blob/master/enhancements/olm/operator-registry.md. 

**Reviewer Checklist**
- [x] Implementation matches the proposed design, or proposal is updated to match implementation
- [x] Sufficient unit test coverage 
- [x] Sufficient end-to-end test coverage
- [x] Docs updated or added to `/docs` 
- [x] Commit messages sensible and descriptive

PR #1226, opened 2020-01-13
fix: refactor catalog polling to use updateStrategy field

exdx:
Updates the catalog source spec to change the polling field. Continuation of #1161. Example of new spec:
```
spec:
  displayName: CatalogSource Test
  sourceType: grpc
  image: quay.io/olmtest/catsrc-update-test:test
  updateStrategy:
    registryPoll:
      interval: 2m
```

`updateStrategy` is more generic and will be useful when adding additional update sources in the future. There was an attempt to look at `union` types and try to implement the update strategy as a union struct but we did not find that the latest controller-tools supported `union` types at this time. 

**Reviewer Checklist**


ecordell:
Shouldn't we update this doc too? https://github.com/operator-framework/operator-lifecycle-manager/blob/master/doc/design/catalog-polling.md

PR #1227, opened 2020-01-13
Use dynamic client to manage prometheus resources

awgreene:
This PR adds OLM support for the PrometheusRule and ServiceMonitor objects in the operator bundle.


pkg/api/client/dynamic/supported.go
```diff
@@ -0,0 +1,50 @@
...
+	supportedResources[ServiceMonitorKind] = supportedResource{kindToResource(ServiceMonitorKind), true}
+}
+
+// TODO: Resolve resource based on GroupVersion, could be done with a partial group version https://github.com/kubernetes/apimachinery/blob/3253b0a30d67e7e362b8615e463156bac729c82f/pkg/api/meta/interfaces.go#L124
+func kindToResource(s string) string {
+	return strings.ToLower(s) + "s"
+}
+
+// IsSupported returns the a SupportedResource pointer if the kind provided is supported and nil otherwise
+func IsSupported(kind string) SupportedResource {
```
Bowenislandsong: is it better to return err if the kind is not supported?

pkg/controller/operators/catalog/operator.go
```diff
@@ -1654,11 +1659,54 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
...
+
+				// Marshal the manifest into an unstructured object
+				dec := yaml.NewYAMLOrJSONDecoder(strings.NewReader(step.Resource.Manifest), 10)
+				unstructuredObject := &unstructured.Unstructured{}
+				if err := dec.Decode(unstructuredObject); err != nil {
+					return errorwrap.Wrapf(err, "error decoding %s object to an unstructured object: %s", step.Resource.Name)
+				}
 
+				var resourceInterface dynamic.ResourceInterface
+				if resource.IsNamespaced() == true {
```
Bowenislandsong: uhmmm. can this just be `if resource.IsNamespaced() {`?

pkg/controller/operators/catalog/operator.go
```diff
@@ -1654,11 +1659,54 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
...
+				// Ensure Object
+				status, err := ensurer.EnsureUnstructuredObject(resourceInterface, unstructuredObject)
+				if err != nil {
+					return err
+				}
+
+				logrus.Info("Updating plan status")
+				plan.Status.Plan[i].Status = status
+			}
 		default:
```
Bowenislandsong: is this an extra default?

awgreene: There are two nested switch cases, this default is for the initial switch.

pkg/api/client/dynamic/supported.go
```diff
@@ -0,0 +1,50 @@
...
+}
+
+// supportedResource contains information about Dynamic Resource that OLM supports.
+type supportedResource struct {
+	name       string
+	namespaced bool
+}
+
+// IsNamespaced returns whether or not the supportedResource is a namespaced object.
+func (sr supportedResource) IsNamespaced() bool {
```
exdx: just a small nit: I think in Go its more iditomatic to drop off the `Is` especially when the function is returning a boolean. Simply `Namespaced()` shows the intent and using it like `if sr.Namespaced()` its clear to understand what the function is intended to show.

pkg/api/client/dynamic/supported.go
awgreene: I do not see any value add in returning an error (or boolean).

By returning a pointer - we can convey if the object was present or not in the map.

pkg/api/apis/operators/installplan_types.go
```diff
@@ -248,6 +251,29 @@ func (b *BundleLookup) SetCondition(cond BundleLookupCondition) BundleLookupCond
 	return cond
 }
 
+func OrderSteps(steps []*Step) []*Step {
```
ecordell: I think a nice future feature would be to allow users to indicate sort order - maybe this is something like a runlevel annotation on resources. 

(this looks good, just looking forward)

pkg/api/apis/operators/installplan_types.go
```diff
@@ -248,6 +251,29 @@ func (b *BundleLookup) SetCondition(cond BundleLookupCondition) BundleLookupCond
...
+		switch step.Resource.Kind {
+		case crdKind:
+			crdList = append(crdList, step)
+		case ClusterServiceVersionKind:
+			csvList = append(csvList, step)
+		default:
+			remainingItems = append(remainingItems, step)
+		}
+	}
+	result := append(crdList, csvList...)
```
ecordell: Since the capacity is known up front, it would be better to allocate it and just assign the values in the right order, instead of `append`s (which will copy and expand the slice)

pkg/controller/operators/catalog/operator.go
```diff
@@ -1460,6 +1464,15 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
 				} else if err != nil {
 					return err
 				} else {
+					// Poll for the CRD to appear
+					logrus.Infof("TEST - About to poll")
```
ecordell: left over comment

pkg/controller/operators/catalog/operator.go
```diff
@@ -1654,11 +1667,62 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
...
+				}
+				unstructuredObject.SetOwnerReferences(updated)
+
+				// Ensure Unstructured Object
+				status, err := ensurer.EnsureUnstructuredObject(resourceInterface, unstructuredObject)
+				if err != nil {
+					return err
+				}
 
+				logrus.Info("Updating plan status")
```
ecordell: this log looks left over as well

pkg/controller/operators/catalog/operator.go
```diff
@@ -1460,6 +1464,15 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
 				} else if err != nil {
 					return err
 				} else {
+					// Poll for the CRD to appear
+					logrus.Infof("TEST - About to poll")
+					err = o.pollForCRD(time.Minute, crd.GetName())
```
ecordell: polling here will block the whole control loop - I think we need to exit, requeue the installplan (perhaps with a delay) and then check if the CRD is ready yet.

We do something similar for CSVs here: https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/controller/operators/olm/operator.go#L1312-L1317

pkg/controller/operators/catalog/operator.go
```diff
@@ -1654,11 +1667,62 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
 				plan.Status.Plan[i].Status = status
 
 			default:
-				return v1alpha1.ErrInvalidInstallPlan
-			}
+				if !isSupported(step.Resource.Kind) {
```
ecordell: This section looks good, but I think we should do some refactoring now that we have this.

Plan execution can be much simpler, now, roughly:

```
 if kind == CSV { create, continue }
 if kind == CRD  && CRD doesn't exist { create, requeue }
 if kind == CRD  && CRD exists && CRD not ready { requeue }
 if isSupported(kind) { treat as unstructured, create, continue }
```

in other words, we need to special case CRD/CSV handling for ordering reasons, but everything else that we support can be treated the same (for the most part; I think serviceaccounts might need some special handling as well).

pkg/controller/operators/catalog/operator.go
```diff
@@ -1770,3 +1834,54 @@ func getCSVNameSet(plan *v1alpha1.InstallPlan) map[string]struct{} {
...
+}
+
+const (
+	PrometheusRuleKind = "PrometheusRule"
+	ServiceMonitorKind = "ServiceMonitor"
+)
+
+// isSupported returns true if OLM supports this type of CustomResource.
+func isSupported(kind string) bool {
+	switch kind {
```
ecordell: see other note, we could treat our existing objects (role, clusterrole, service, etc) here as well.

pkg/api/apis/operators/installplan_types.go
awgreene: Sounds good - will do!

pkg/controller/operators/catalog/operator.go
exdx: would be nice if we could tie this with the registry validation work done recently, which basically does the same thing https://github.com/operator-framework/operator-registry/commit/d8e2ec52c19ad2b2a11a91f1ecccb1887bbae787#diff-e2dca81e4297cb851471d15b17acf7f4R23
I know we don't want to depend on the registry in OLM but maybe if we pull these types out and put them in the api long-term it could be nicer. that way there's one place where all our defined types are stored.

pkg/controller/operators/catalog/operator.go
awgreene: Great point, addressed in the latest version of the PR.

pkg/controller/operators/catalog/operator.go
awgreene: @ecordell and I discussed this offline and I will follow up this PR with a PR that refactors this code.

pkg/controller/operators/catalog/operator.go
awgreene: That's a great suggestion @exdx - we had discussed fleshing out an "operatorframework/apis" repo but did not get terribly far. Let's discuss tomorrow.

I assume that this too ca be closed due to the discussion that @ecordell and I had offline [1]

[1] https://github.com/operator-framework/operator-lifecycle-manager/pull/1227#discussion_r369834013

pkg/api/apis/operators/installplan_types.go
```diff
@@ -248,6 +253,57 @@ func (b *BundleLookup) SetCondition(cond BundleLookupCondition) BundleLookupCond
...
+		switch step.Resource.Kind {
+		case crdKind:
+			crdList = append(crdList, step)
+		case ClusterServiceVersionKind:
+			csvList = append(csvList, step)
+		default:
+			remainingResources = append(remainingResources, step)
+		}
+	}
+	length := len(crdList) + len(csvList) + len(remainingResources)
```
ecordell: nit: isn't this just `len(steps)`?

pkg/api/apis/operators/v1alpha1/installplan_types.go
```diff
@@ -65,10 +66,12 @@ const (
...
 const (
-	StepStatusUnknown    StepStatus = "Unknown"
-	StepStatusNotPresent StepStatus = "NotPresent"
-	StepStatusPresent    StepStatus = "Present"
-	StepStatusCreated    StepStatus = "Created"
+	StepStatusUnknown             StepStatus = "Unknown"
+	StepStatusNotPresent          StepStatus = "NotPresent"
+	StepStatusPresent             StepStatus = "Present"
+	StepStatusCreated             StepStatus = "Created"
+	StepStatusCreatedCRD          StepStatus = "CreatedCRD"
```
ecordell: this seems odd, since `CRD` is already in the step kind. What about something like "Pending" or "WaitingForApi" or "Registering" (or something better that you think of)?

pkg/controller/operators/catalog/operator.go
dinhxuanvu: This issue has been discussed on my validation PR on registry as well. See the thread (https://github.com/operator-framework/operator-registry/pull/146#discussion_r366883298). There is a tech debt on consolidating the supported types into some sort of API or importable code that can be imported between different repos. I would say it will need to start with OLM who is the one deciding which resources are supported/consumable. This is something we can discuss and address later as this PR is certainly not a right place to address that issue.

pkg/controller/operators/catalog/operator_test.go
```diff
@@ -1119,6 +1167,15 @@ func objectReference(name string) *corev1.ObjectReference {
...
 
+func yamlFromFilePath(fileName string) string {
+	yaml, err := ioutil.ReadFile(fileName)
+	if err != nil {
+		fmt.Printf("Error reading YAML file: %s\n", err)
+		return ""
+	}
+
+	return string(yaml)
+}
```
ecordell: nit: add a space

pkg/controller/operators/catalog/operator.go
```diff
@@ -1172,11 +1176,13 @@ func (o *Operator) syncInstallPlans(obj interface{}) (syncError error) {
 		}
 	}
 
+	logger.Infof("Phase coming in: %s", plan.Status.Phase)
```
ecordell: this looks like it should be debug level or trace

pkg/controller/operators/catalog/operator.go
```diff
@@ -1172,11 +1176,13 @@ func (o *Operator) syncInstallPlans(obj interface{}) (syncError error) {
...
 	outInstallPlan, syncError := transitionInstallPlanState(logger.Logger, o, *plan, o.now())
 
 	if syncError != nil {
 		logger = logger.WithField("syncError", syncError)
 	}
+	logger.Infof("Phase coming out: %s", outInstallPlan.Status.Phase)
```
ecordell: ditto - debug logs shouldn't be info

also for this type of thing it's a little nicer to use our structured logs

```go
logger.WithField("phase", outInstallPlan.Status.Phase).Debug("install plan updated")
```

pkg/controller/operators/catalog/operator.go
```diff
@@ -1240,8 +1246,17 @@ func transitionInstallPlanState(log *logrus.Logger, transitioner installPlanTran
...
+		// Loop over one final time to check and see if everything is good.
+		if !out.Status.NeedsRequeue() {
+			log.Info("Status does not need a requeue!")
+			out.Status.SetCondition(v1alpha1.ConditionMet(v1alpha1.InstallPlanInstalled, &now))
+			out.Status.Phase = v1alpha1.InstallPlanPhaseComplete
+		} else {
+			out.Status.Phase = v1alpha1.InstallPlanPhaseRequeue
+		}
+		return out, nil
+	case v1alpha1.InstallPlanPhaseRequeue:
```
ecordell: If a CRD takes multiple loops to register, won't this cycle Installing -> Requeue -> Installing?

I think we can make it just using `Installing` and remove the `Requeue` phase if we also remove these lines:

```go
	// no changes in status, don't update
	if outInstallPlan.Status.Phase == plan.Status.Phase {
		return
	}
```

pkg/controller/operators/catalog/operator_test.go
```diff
@@ -518,6 +520,50 @@ func TestExecutePlan(t *testing.T) {
 	}
 }
 
+func TestExecutePlanUnstructuredObjects(t *testing.T) {
```
ecordell: Shouldn't this also that prometheus objects are valid steps?

pkg/controller/operators/catalog/operator.go
awgreene: Sorry - left over debugging code!

pkg/controller/operators/catalog/operator.go
```diff
@@ -1654,18 +1686,70 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
...
 			default:
-				return v1alpha1.ErrInvalidInstallPlan
-			}
+				if !isSupported(step.Resource.Kind) {
+					// Not a supported resource
+					plan.Status.Plan[i].Status = v1alpha1.StepStatusUnsupportedResource
+					return v1alpha1.ErrInvalidInstallPlan
+				}
+
+				// Entering unstructured territory... To infinity and beyond!
```
dinhxuanvu: This line is unnecessary and should be removed unless you can add something more informative such "Attempt to install other supported resources". You don't really install unstructured resources, more like installing supported resources by decoding yaml into unstructured object first.

pkg/controller/operators/catalog/operator_test.go
```diff
@@ -1119,6 +1167,15 @@ func objectReference(name string) *corev1.ObjectReference {
 	return &corev1.ObjectReference{Name: name}
 }
 
+func yamlFromFilePath(fileName string) string {
```
dinhxuanvu: I understand what you are doing here but I wonder if that file doesn't exist and the func returns empty string, would the test would pass or fail or what would be expected result if that is the case. Maybe I'm bit over cautious here but if there is an error, we should return and check for it, simply avoid some false positive in case the test file being moved.

pkg/controller/operators/catalog/operator.go
```diff
@@ -1240,8 +1246,17 @@ func transitionInstallPlanState(log *logrus.Logger, transitioner installPlanTran
 			out.Status.Phase = v1alpha1.InstallPlanPhaseFailed
 			return out, err
 		}
-		out.Status.SetCondition(v1alpha1.ConditionMet(v1alpha1.InstallPlanInstalled, &now))
-		out.Status.Phase = v1alpha1.InstallPlanPhaseComplete
+		// Loop over one final time to check and see if everything is good.
+		if !out.Status.NeedsRequeue() {
+			log.Info("Status does not need a requeue!")
```
dinhxuanvu: I'm not sure if this is leftover from your debugging or you intend to keep it here. I would change it to `log.Debug` because the log message doesn't provide any informative material. Frankly, this is a bit misleading because you requere the InstallPlan, not Status.

pkg/controller/operators/catalog/operator_test.go
```diff
@@ -1119,6 +1167,15 @@ func objectReference(name string) *corev1.ObjectReference {
...
+	yaml, err := ioutil.ReadFile(fileName)
+	if err != nil {
```
dinhxuanvu: I understand what you are doing here but I wonder what is expected result of the test case if the error occurs here and an empty string is returned? I think you should return the error and handle it properly in the test case to avoid some false positive in case the test data is moved/removed accidentally.

pkg/controller/operators/catalog/operator.go
dinhxuanvu: This comment is unnecessary and probably should be removed unless you add some information here such as "Install other supported resources". You don't really install unstructured resources. You simply decode yaml into unstructured object in order to install it.

pkg/controller/operators/catalog/operator.go
```diff
@@ -1383,19 +1398,35 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
...
+		case v1alpha1.StepStatusCreatedCRD:
+			switch step.Resource.Kind {
+			case crdKind:
+				// Marshal the manifest into a CRD instance.
+				var crd v1beta1ext.CustomResourceDefinition
+				err := json.Unmarshal([]byte(step.Resource.Manifest), &crd)
+				if err != nil {
+					return errorwrap.Wrapf(err, "error parsing step manifest: %s", step.Resource.Name)
+				}
+				_, err = o.opClient.ApiextensionsV1beta1Interface().ApiextensionsV1beta1().CustomResourceDefinitions().Create(&crd)
```
dinhxuanvu: I find the InstallPlan status flow is a bit convoluted as we switch from `Installing` to `Requeue`/`Pending` and back to `Installing` solely basing on the an error from CRD creation step which only tells us if the CRD already exists or not. From what I understand is when the InstallPlan creates the CRD, it is there. The issue you are facing is the CRD is not yet ready for you to create the CR that is associated with it. Is that correct? If that is the case, maybe you should explicitly check the status to CRD to determine if the CRD is ready/accepted, before moving on. Otherwise, bring it back to `Requeue/Pending` state. Here is what we do on the OLM operator side (https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/controller/operators/olm/requirements.go#L122). What do you think?

pkg/controller/operators/catalog/operator.go
awgreene: Leftover code - will remove.

pkg/controller/operators/catalog/operator_test.go
awgreene: Seems to be a copy of https://github.com/operator-framework/operator-lifecycle-manager/pull/1227#discussion_r369936379

pkg/controller/operators/catalog/operator.go
awgreene: Seems to be a copy of https://github.com/operator-framework/operator-lifecycle-manager/pull/1227#discussion_r369935743

awgreene: Will remove

pkg/controller/operators/catalog/operator_test.go
awgreene: Adding an error check.

pkg/controller/operators/catalog/operator.go
awgreene: Evan and I had a similar discussion offline and we came to a similar conclusion - we should remove the `Requeue` phase. Regarding the CRD API availability, I had used API Discovery to check for GV, but I like the code you linked and will update the current implementation to use it.

pkg/api/apis/operators/v1alpha1/installplan_test.go
```diff
@@ -0,0 +1,61 @@
...
+
+func isOrdered(steps []*Step) bool {
+	var csvSeen, otherResourceSeen bool
+	for _, step := range steps {
+		switch step.Resource.Kind {
+		case crdKind:
+			if csvSeen || otherResourceSeen {
+				return false
+			}
+		case ClusterServiceVersionKind:
```
exdx: shouldn't we set `csvSeen` to true in this case? right now this function always returns true

pkg/api/apis/operators/v1alpha1/installplan_test.go
```diff
@@ -0,0 +1,61 @@
...
+			if otherResourceSeen {
```
exdx: set `otherResourceSeen` to true?

pkg/controller/operators/catalog/operator.go
```diff
@@ -1654,11 +1693,59 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
...
 
+				// Ensure Unstructured Object
+				status, err := ensurer.EnsureUnstructuredObject(resourceInterface, unstructuredObject)
+				if err != nil {
+					return err
+				}
+
+				plan.Status.Plan[i].Status = status
+			}
 		default:
```
dinhxuanvu: Now that you no longer return an error here, this default statement is useless. Maybe you should remove it altogether?

On the other hand, I wonder what we should do if the Step doesn't have a status. Maybe we should change to `StepStatusUnknown` so it can be processed later. Would that be an appropriate behavior @ecordell?

awgreene: After reviewing the switch block - we should probably continue to return the error.

The only case we should hit this issue is when the installPlan includes an Unsupported Resource, where we'd wan't it to fail anyways.

dinhxuanvu:
lgtm
Waiting for the CI test.

pkg/controller/operators/catalog/operator.go
```diff
@@ -1383,19 +1390,50 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
...
 	for i, step := range plan.Status.Plan {
 		switch step.Status {
 		case v1alpha1.StepStatusPresent, v1alpha1.StepStatusCreated:
 			continue
+		case v1alpha1.StepStatusWaitingForAPI:
+			switch step.Resource.Kind {
+			case crdKind:
+				crd, err := o.opClient.ApiextensionsV1beta1Interface().ApiextensionsV1beta1().CustomResourceDefinitions().Get(step.Resource.Name, metav1.GetOptions{})
+				if err != nil {
+					if k8serrors.IsNotFound(err) {
```
njhale: nit:
```go
if !k8serrors.IsNotFound(err) {
    return errorwrap.Wrapf(err, "error finding the %s CRD", crd.Name)
}
plan.Status.Plan[i].Status = v1alpha1.StepStatusNotPresent
continue
```

pkg/controller/operators/catalog/operator.go
```diff
@@ -1383,19 +1390,50 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
...
+					continue
+				}
 
+				established, namesAccepted := false, false
+				for _, cdt := range crd.Status.Conditions {
+					switch cdt.Type {
+					case v1beta1.Established:
+						if cdt.Status == v1beta1.ConditionTrue {
+							established = true
+						}
```
njhale: nit:

```go
established = cdt.Status == v1beta1.ConditionTrue
```

pkg/controller/operators/catalog/operator.go
```diff
@@ -1458,10 +1496,11 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
 					plan.Status.Plan[i].Status = v1alpha1.StepStatusPresent
 					continue
 				} else if err != nil {
+					// Unexpected error creating the CRD.
 					return err
 				} else {
-					// If no error occured, mark the step as Created.
-					plan.Status.Plan[i].Status = v1alpha1.StepStatusCreated
+					// If no error occured, make sure to wait for the API to become available.
```
njhale: nit: I don't think this needs an `else` statement.

pkg/controller/operators/catalog/operator.go
```diff
@@ -1383,19 +1390,50 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
...
+				if err != nil {
+					if k8serrors.IsNotFound(err) {
+						plan.Status.Plan[i].Status = v1alpha1.StepStatusNotPresent
+					} else {
+						return errorwrap.Wrapf(err, "error finding the %s CRD", crd.Name)
+					}
+					continue
+				}
 
+				established, namesAccepted := false, false
```
njhale: This can wait for a follow up, but I think the CRD readiness can be wrapped up into a helper function to clean this up some.

pkg/controller/operators/catalog/operator.go
```diff
@@ -1654,9 +1693,58 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
...
-				return v1alpha1.ErrInvalidInstallPlan
-			}
+				if !isSupported(step.Resource.Kind) {
+					// Not a supported resource
+					plan.Status.Plan[i].Status = v1alpha1.StepStatusUnsupportedResource
+					return v1alpha1.ErrInvalidInstallPlan
+				}
+
+				// Marshal the manifest into an unstructured object
+				dec := yaml.NewYAMLOrJSONDecoder(strings.NewReader(step.Resource.Manifest), 10)
```
njhale: nit: we do this enough that a `StepCodec` would probably be a welcome abstraction.

pkg/controller/operators/catalog/operator.go
```diff
@@ -1654,9 +1693,58 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
...
+
+				// Set up the dynamic client ResourceInterface
+				var resourceInterface dynamic.ResourceInterface
+				if r.Namespaced {
+					ownerutil.AddOwner(unstructuredObject, plan, false, false)
+					unstructuredObject.SetNamespace(namespace)
+					resourceInterface = o.dynamicClient.Resource(gvr).Namespace(namespace)
+				} else {
+					resourceInterface = o.dynamicClient.Resource(gvr)
+				}
```
njhale: This seems like a lot of ceremony just to create an unstructured resource. I think controller-runtime's client handles this for you.

test/e2e/dynamic_resource_e2e_test.go
```diff
@@ -0,0 +1,104 @@
...
+		require.NoError(t, c.KubernetesInterface().CoreV1().Namespaces().Delete(ns.GetName(), deleteOpts))
+	}()
+
+	catsrc := &v1alpha1.CatalogSource{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      genName("dynamic-catalog-"),
+			Namespace: ns.GetName(),
+		},
+		Spec: v1alpha1.CatalogSourceSpec{
+			Image:      "quay.io/olmtest/catsrc_dynamic_resources:e2e-test",
```
njhale: Should we lock this down to a known digest to avoid breaking tests with changes?

test/e2e/dynamic_resource_e2e_test.go
```diff
@@ -0,0 +1,104 @@
...
+	ipName := sub.Status.InstallPlanRef.Name
+	ip, err := waitForInstallPlan(t, crc, ipName, sub.GetNamespace(), buildInstallPlanPhaseCheckFunc(v1alpha1.InstallPlanPhaseFailed, v1alpha1.InstallPlanPhaseComplete))
+	require.NoError(t, err)
+	// Ensure the InstallPlan contains the steps resolved from the bundle image
+	expectedSteps := map[registry.ResourceKey]struct{}{
+		registry.ResourceKey{Name: "my-prometheusrule", Kind: "PrometheusRule"}: {},
+		registry.ResourceKey{Name: "my-servicemonitor", Kind: "ServiceMonitor"}: {},
+	}
+
+	for _, step := range ip.Status.Plan {
```
njhale: Is `Plan` not stable?

njhale:
/hold
/retest

Since the e2e test for this is failing, we should probably hold off on merging it until we verify that the cause is unrelated.

I also think something is up with the approval bot RN.

njhale:
/hold cancel

Removing hold since these tests appear to succeed against minikube.

pkg/controller/operators/catalog/operator.go
awgreene: Controller-runtime's client does not require the GVR - something we could switch to but it might make more sense to rip the dynamic-client out entirely given that the dynamic client is used elsewhere in OLM.

pkg/api/apis/operators/installplan_types.go
```diff
@@ -248,6 +252,57 @@ func (b *BundleLookup) SetCondition(cond BundleLookupCondition) BundleLookupCond
 	return cond
 }
 
+func OrderSteps(steps []*Step) []*Step {
+	// CRDs must be applied first
+	crdList := []*Step{}
```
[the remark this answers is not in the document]
ecordell: The issue is that the installplan can contain CRs for a CRD in the same installplan - so the CRD and the CRs will get created, but the CRs can fail if the CRD doesn't get accepted fast enough by the apiserver.

pkg/api/apis/operators/installplan_types.go
```diff
@@ -248,6 +252,57 @@ func (b *BundleLookup) SetCondition(cond BundleLookupCondition) BundleLookupCond
...
+
+	for j := range remainingResources {
+		result[i] = remainingResources[j]
+		i++
+	}
+
+	return result
+}
+
+func (s InstallPlanStatus) NeedsRequeue() bool {
```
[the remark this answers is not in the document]
awgreene: That's a fair point - something I'll try to address before pushing this PR through!

pkg/api/apis/operators/installplan_types.go
```diff
@@ -248,6 +252,57 @@ func (b *BundleLookup) SetCondition(cond BundleLookupCondition) BundleLookupCond
 	return cond
 }
 
+func OrderSteps(steps []*Step) []*Step {
```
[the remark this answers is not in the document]
awgreene: I thought we copied all methods to the unversioned types, is that not required? I assume this is due to the fact that we don't use the unversioned types?

awgreene:
/retest

```
Installing from release registry.svc.ci.openshift.org/ci-op-kg361l3m/release@sha256:1aecda4ccf607754b71696b2ad17a717a9dc28f93ba1a529a43ffa177a5f6558
AWS region: us-east-1 (zones: us-east-1b us-east-1c)
level=info msg="Credentials loaded from the \"default\" profile in file \"/etc/openshift-installer/.awscred\""
level=warning msg="Found override for release image. Please be warned, this is not advised"
level=info msg="Consuming Install Config from target directory"
level=info msg="Creating infrastructure resources..."
level=error
level=error msg="Error: [ERR]: Error building changeset: timeout while waiting for state to become 'accepted' (timeout: 5m0s)"
level=error
level=error msg="  on ../tmp/openshift-install-520768184/route53/base.tf line 33, in resource \"aws_route53_record\" \"api_external\":"
level=error msg="  33: resource \"aws_route53_record\" \"api_external\" {"
level=error
level=error
level=fatal msg="failed to fetch Cluster: failed to generate asset \"Cluster\": failed to create cluster: failed to apply using Terraform"
```

awgreene:
Closed due to a CI issue. Opened https://github.com/operator-framework/operator-lifecycle-manager/pull/1253

PR #1228, opened 2020-01-14
Update release script to require version 3 of helm.

benluddy:
The output of "helm template" is not stable from version 2 to version
3, which resulted in the reordering of certain documents in
multi-document yaml files. The package_release script has been updated
to abort with an error message when the helm major version isn't
exactly 3, in order to reduce ambiguity and prevent meaningless
changes to manifest files.


ecordell:
/lgtm 

double checked that this works on macos, just to be sure.

PR #1235, opened 2020-01-15
Update internal objects proposal to use annotation on CSV instead of CRD

robszumski:
Updates the internal objects proposal.

Many users are not able to read the full list of CRDs on a cluster, so this proposal now stores a list of internal CRD names as an annotation on the CSV.


ecordell:
/lgtm

doc-only, merging without tests

PR #1240, opened 2020-01-20
Bug 1787597: fix(packageserver): cache packagemanifests

njhale:

Backport packagemanifest cache to 4.2.


crawford:
This backport unnecessarily flattens the commit history of the [original PR](https://github.com/operator-framework/operator-lifecycle-manager/pull/1181) and appears to have picked up an extra commit.

njhale:
@crawford, rather than flattening, I only cherry-picked the first commit to avoid having to deal with vendor directory conflicts (the first commit contains the feature work). If it's a requirement to keep the entire commit history, then I can attempt to do so. 

The extra commit contains the license year removal so we could avoid firing up simultaneous tests . I can understand removing it, but it is a separate commit and can be reverted independently as is.

Thanks for taking the time to review these btw.

ashcrow:
> If it's a requirement to keep the entire commit history, then I can attempt to do so.

I'm not aware of a requirement, but it does make it easier to verify. In this case, as long as it doesn't break future cherry picks I think it's OK.

PR #1247, opened 2020-01-24
Vendor helm, since it is used for templating manifests.

benluddy:
Pinned to a recent commit including a change that resolves the
arbitrary reordering of multi-document manifests. See:
https://github.com/helm/helm/commit/4d8160eedf43db8ed137906a8af0c5d67e03e4fe#diff-1b5cea5b40418f5f5929d3b62034886dR107

This is related to the earlier PR https://github.com/operator-framework/operator-lifecycle-manager/pull/1228 -- I didn't notice at the time that the document reordering was arbitrary.


ecordell:
/lgtm

/retest

doesn't have to be this PR, but we should update the requirements in the release doc to indicate that helm is no longer a prereq 😄

benluddy:
> /lgtm
> 
> /retest
> 
> doesn't have to be this PR, but we should update the requirements in the release doc to indicate that helm is no longer a prereq smile

I'm ten steps ahead of you -- the doc never mentioned that helm was required. :1st_place_medal:

PR #1250, opened 2020-01-24
Fix overly broad .gitignore pattern that was excluding a vendored pac…

benluddy:


exdx:
This enables `make run-local` which is currently hiccuping on a missing helm pkg incorrectly ignored via the .gitignore directive

PR #1251, opened 2020-01-24
Install a C compiler in the builder image used by the upstream Dockerfile.

benluddy:
Importing the package
github.com/operator-framework/operator-registry/pkg/configmap took on
a transitive dependency on github.com/mattn/go-sqlite3, which needs to
build with CGO. This commit installs gcc and libc-dev on the builder
image used for upstream image builds.

### DIFF
--- a/upstream.Dockerfile
+++ b/upstream.Dockerfile
@@ -2,7 +2,7 @@ FROM golang:1.12-alpine as builder
 LABEL stage=builder
 WORKDIR /build
 
-RUN apk update && apk add bash make git mercurial jq && apk upgrade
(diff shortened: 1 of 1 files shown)


benluddy:
/hold
Removed the dependency on go-sqlite from the configmap package here: https://github.com/operator-framework/operator-registry/pull/161. I'll update this PR soon.

ecordell:
I believe this is already taken care of now - reopen if I've misunderstood

benluddy:
Yes, this PR was obsolete. `github.com/operator-framework/operator-registry/pkg/configmap` was not supposed to have a dependency on `github.com/mattn/go-sqlite3`.

PR #1252, opened 2020-01-25
Bug 1798645: fix(bundles): execute opm from tooling container

njhale:

Execute the `opm` command in a container of the image it was packaged with.

To prevent dynamic linking issues caused by running in a `FROM scratch` bundle container.

### DIFF
--- a/pkg/controller/bundle/bundle_unpacker.go
+++ b/pkg/controller/bundle/bundle_unpacker.go
@@ -54,9 +54,9 @@ func (c *ConfigMapUnpacker) job(cmRef *corev1.ObjectReference, bundlePath string
 					RestartPolicy: corev1.RestartPolicyOnFailure,
 					Containers: []corev1.Container{
 						{
-							Name:    cmRef.Name,
-							Image:   bundlePath,
-							Command: []string{"/injected/opm", "alpha", "bundle", "extract", "-n", cmRef.Namespace, "-c", cmRef.Name},
+							Name:    "extract",
+							Image:   c.opmImage,
+							Command: []string{"opm", "alpha", "bundle", "extract", "-m", "/bundle/", "-n", cmRef.Namespace, "-c", cmRef.Name},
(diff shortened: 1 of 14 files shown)


test/e2e/installplan_e2e_test.go
```diff
@@ -2574,7 +2574,6 @@ func TestInstallPlanCRDValidation(t *testing.T) {
 }
 
 func TestInstallPlanUnpacksBundleImage(t *testing.T) {
-	defer cleaner.NotifyTestComplete(t, true)
```
benluddy: Is this intentional?

njhale: yes -- it was attempting to deleted resources that no longer existed, and I didn't want to dig into the cleanup stuff for this PR. It's not needed here anyway since the namespace cleanup is explicitly deferred below.

pkg/controller/bundle/bundle_unpacker.go
```diff
@@ -65,28 +65,49 @@ func (c *ConfigMapUnpacker) job(cmRef *corev1.ObjectReference, bundlePath string
...
-									MountPath: "/copy-dest",
+									Name:      "tools",
+									MountPath: "/tools",
+								},
+							},
+						},
+						{
+							Name:    "pull",
+							Image:   bundlePath,
+							Command: []string{"/tools/cp", "-Rv", "/manifests", "/metadata", "/bundle"}, // Copy bundle content to its mount
```
njhale: Hard coding `/manifests` for now. We'll follow up with a script that uses the manifest directories label to determine what to copy.

njhale:
/hold

Need to decide what we're doing w/ the busybox image dependency.

njhale:
/hold cancel

Instead of using busybox, we decided to create a tiny command utility that copies bundle content between directories. The statically compiled binary is shipped within OLM's image for easy access.

PR #1257, opened 2020-01-28
Bug 1800625: fix: add internal registry for catalog polling e2e test

exdx:
Updates e2e test to rely on an internal docker registry by creating/deleting one during the process of the test, or using the existing openshift registry. This removes the need for robot credentials to the quay registry that the test currently relies on. 

Security. 

**Reviewer Checklist**

### DIFF
--- a/test/e2e/registry.go
+++ b/test/e2e/registry.go
@@ -1,0 +1,85 @@
+package e2e
+
+import (
+	"fmt"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/operatorclient"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"os/exec"
+)
+
+// This module contains helper functions for copying images and creating image registries
+// Use for tests that require manipulating images or use of custom container registries
+
+const (
+	registryImage = "registry:2.7.1"
+	registryName  = "registry"
+	localFQDN     = "localhost:5000"
+)
+
+func createDockerRegistry(client operatorclient.ClientInterface, namespace string) (string, error) {
+	registry := &corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      registryName,
+			Namespace: namespace,
+			Labels:    map[string]string{"name": registryName},
+		},
+		Spec: corev1.PodSpec{
+			Containers: []corev1.Container{
+				{
+					Name:  registryName,
+					Image: registryImage,
+					Ports: []corev1.ContainerPort{
+						{
+							HostPort:      int32(5000),
+							ContainerPort: int32(5000),
+						},
+					},
+				},
+			},
+		},
+	}
+
+	svc := &corev1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      registryName,
+			Namespace: namespace,
+		},
+		Spec: corev1.ServiceSpec{
+			Selector: map[string]string{"name": registryName},
+			Ports: []corev1.ServicePort{
+				{
+					Port: int32(5000),
+				},
+			},
+		},
+	}
+
+	_, err := client.KubernetesInterface().CoreV1().Pods(namespace).Create(registry)
+	if err != nil {
+		return "", fmt.Errorf("creating test registry: %s", err)
+	}
+
+	_, err = client.KubernetesInterface().CoreV1().Services(namespace).Create(svc)
+	if err != nil {
+		return "", fmt.Errorf("creating test registry: %s", err)
+	}
+
+	return localFQDN, nil
+}
+
+func deleteDockerRegistry(client operatorclient.ClientInterface, namespace string) {
+	_ = client.KubernetesInterface().CoreV1().Pods(namespace).Delete(registryName, &metav1.DeleteOptions{})
+	_ = client.KubernetesInterface().CoreV1().Services(namespace).Delete(registryName, &metav1.DeleteOptions{})
+}
+
+// port-forward registry pod port 5000 for local test
+// port-forwarding ends when registry pod is deleted: no need for explicit port-forward stop
+func registryPortForward(namespace string) error {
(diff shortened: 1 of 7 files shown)


exdx:
This now works locally as expected, but not on OpenShift CI

exdx:
Looks like the pod DNS resolution isn't working as expected.
```
DEBU[0006] Ping https://registry.test.openshift-operators.svc.cluster.local:5000/v2/ err Get https://registry.test.openshift-operators.svc.cluster.local:5000/v2/: dial tcp: lookup registry.test.openshift-operators.svc.cluster.local: no such host (&url.Error{Op:"Get", URL:"https://registry.test.openshift-operators.svc.cluster.local:5000/v2/", Err:(*net.OpError)(0xc0000e4dc0)})
DEBU[0006] GET http://registry.test.openshift-operators.svc.cluster.local:5000/v2/
DEBU[0011] Ping http://registry.test.openshift-operators.svc.cluster.local:5000/v2/ err Get http://registry.test.openshift-operators.svc.cluster.local:5000/v2/: dial tcp: lookup registry.test.openshift-operators.svc.cluster.local: no such host (&url.Error{Op:"Get", URL:"http://registry.test.openshift-operators.svc.cluster.local:5000/v2/", Err:(*net.OpError)(0xc000476050)})
DEBU[0011] GET https://registry.test.openshift-operators.svc.cluster.local:5000/v1/_ping
DEBU[0016] Ping https://registry.test.openshift-operators.svc.cluster.local:5000/v1/_ping err Get https://registry.test.openshift-operators.svc.cluster.local:5000/v1/_ping: dial tcp: lookup registry.test.openshift-operators.svc.cluster.local: no such host (&url.Error{Op:"Get", URL:"https://registry.test.openshift-operators.svc.cluster.local:5000/v1/_ping", Err:(*net.OpError)(0xc000476280)})
DEBU[0016] GET http://registry.test.openshift-operators.svc.cluster.local:5000/v1/_ping
DEBU[0021] Ping http://registry.test.openshift-operators.svc.cluster.local:5000/v1/_ping err Get http://registry.test.openshift-operators.svc.cluster.local:5000/v1/_ping: dial tcp: lookup registry.test.openshift-operators.svc.cluster.local: no such host (&url.Error{Op:"Get", URL:"http://registry.test.openshift-operators.svc.cluster.local:5000/v1/_ping", Err:(*net.OpError)(0xc0005ea000)})
FATA[0021] Error trying to reuse blob sha256:92fa4933bde0c606cb64e1fc6f88bd449ebb8595adcf08f9963a3db9a1a0bea6 at destination: error pinging docker registry registry.test.openshift-operators.svc.cluster.local:5000: Get http://registry.test.openshift-operators.svc.cluster.local:5000/v2/: dial tcp: lookup registry.test.openshift-operators.svc.cluster.local: no such host
```

exdx:
When running locally - generate and use a local docker registry. When running on OpenShift, expose the existing openshift registry and use that registry instead.

exdx:
The test was rewritten to launch skopeo as a pod with the copy arguments as args to the entrypoint instead of `exec`ing out to it. This means skopeo was able to access the openshift registry over the internal service since its running in the test cluster. Also removes the need for vendoring in more dependencies just for the purposes of the test. 

That being said locally I don't believe the local docker registry implementation will work. To allow insecure registry pulling you have to change node configuration under `/etc...` and i'm not sure how much value there is to testing this locally. Right now the test just skips it. It was working locally when we `exec`ed out to skopeo and used portforwarding, so I could add that code back, but its just messier since we have 2 different ways of dealing with skopeo. 

Running this locally is not great since it involves both skopeo and not having anything on port 5000. Ok with skipping the local test for now and just ensuring it works on openshift CI and the credentials are gone.

test/e2e/util_test.go
```diff
@@ -600,3 +600,28 @@ func buildCRCleanupFunc(c operatorclient.ClientInterface, apiGroup, version, nam
 		})
 	}
 }
+
+// Local determines whether test is running locally or in a container on openshift-CI.
+// Queries for a clusterversion object specific to OpenShift.
+func Local(client operatorclient.ClientInterface) (bool, error) {
```
benluddy: Did you consider separating the logic that copies images and provides the registry address into two implementations using build tags?

test/e2e/registry.go
```diff
@@ -0,0 +1,98 @@
...
+		return "", fmt.Errorf("creating test registry: %s", err)
+	}
+
+	_, err = client.KubernetesInterface().CoreV1().Services(namespace).Create(svc)
+	if err != nil {
+		return "", fmt.Errorf("creating test registry: %s", err)
+	}
+
+	// <name>.<namespace>.svc.cluster.local
+	// return strings.Join([]string{registryName, namespace, "svc.cluster.local:5000"}, "."), nil
```
benluddy: ^

test/e2e/registry.go
```diff
@@ -0,0 +1,98 @@
...
+
+	// open port 5000 to the registry for skopeo copy to work
+	// ensure pod is ready before attempting port-forwarding
+	time.Sleep(5 * time.Second)
```
benluddy: It might be worthwhile to replace the sleep with some kind of retry strategy, to cut down on flakes caused by registry pod slowness.

exdx: I agree, but I haven't seen this test flake out on CI once yet. Maybe we can revisit if we do see a flake occur?

test/e2e/util_test.go
exdx: That makes sense now that the two paths diverge pretty heavily, but originally the test looked the same for both. We could do that now but I'm not sure the additional effort is warranted given its passing as-is, the code is still pretty readable, and its better to focus on other tasks now.

test/e2e/registry.go
exdx: Ok I can add an `awaitPod` here, just so the local version is a bit cleaner.

test/e2e/util_test.go
benluddy: That's fine. Maybe it's something to consider the next time we introduce a test that has different local and non-local modes.

PR #1258, opened 2020-01-29
Bug 1796092: Update operator-registry dependency to v1.5.8.

benluddy:
This eliminates a transitive dependency on a package that needs to be
built with CGO enabled, and fixes the upstream image build, which does
not otherwise require the presence of a C compiler in its builder image.


njhale:
Thank you!

/retest
/lgtm

PR #1263, opened 2020-01-30
Bug 1783284: fix(perf): remove periodic resync

ecordell:
Removes periodic resync to avoid spikes in CPU every few minutes.

CatalogSource still requires a resync to ensure polled updates are found, but that can be addressed later (and involves few resources).

/hold 

holding to try this out in test clusters and make sure things look good.

**Reviewer Checklist**


ecordell:
This was replaced by a PR adding jitter.

PR #1265, opened 2020-01-30
Generate manifests for version 0.14.1.

benluddy:


ecordell:
this PR is upstream-only, merging without a BZ

PR #1273, opened 2020-02-03
Bug 1797700: [release-4.1] fix(packageserver): cache packagemanifests

njhale:

Backport packagemanifest cache to 4.1.


njhale:
/hold

This won't pass until the license year fix is backported. There's no use in spamming CI.

bparees:
At this point we are only accepting critical bug fixes for 4.1. This does not seem to meet that target, do you want to close it (and the BZ) or make a case for why it is critical?

ecordell:
@bparees This is a performance fix for the packageserver api. If we find that this becomes critical for some reason in the future, we can re-open.

sdodson:
/lgtm cancel
(to get if off patch manager queries)

PR #1277, opened 2020-02-05
Bug 1798135: Fix cases where an operator's ready channel may never close.

benluddy:
In at least three error/cancellation scenarios, an operator would
finish running without closing its ready channel. Users (i.e. olm,
catalog, and package-server) block first on <-o.Ready() and then on
<-o.Done(), so if the operator terminates without closing its ready
channel, they will block indefinitely, which prevents recovery via
automatic restarts.

**Reviewer Checklist**


njhale:
This looks great, thanks!

/lgtm

PR #1284, opened 2020-02-07
Bug 1786471: fix(packagemanifests): Fix index key func to avoid key collision

dinhxuanvu:
Currently, packagemanifests cache uses packagemanifest name and ns
as a key which is not unique as different catalogsources can have
the same package in the same namespace. As a result, key collision
happens in cache (map) which in turn leads to missing packagemanifests
when users query for available packageminifests in the namespace.

This PR adds new key function to use the combination of catalogsource
name, namespace and packagemanifest name to ensure the key uniqueness
to prevent collision.

**Reviewer Checklist**

### DIFF
--- a/pkg/package-server/provider/registry.go
+++ b/pkg/package-server/provider/registry.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


dinhxuanvu:
@benluddy PTAL. Thanks :)

PR #1289, opened 2020-02-13
Bug 1798051: Reinstate v1alpha2 operatorgroup

kevinrizza:
We did not properly deprecate v1alpha2 version of operatorgroup crd in
the 4.4 OpenShift release. Adding back that version to the crd spec

In addition, this pr cuts a new upstream release to keep the manifests
in lockstep (along with fixing the 0.14.1 upstream release)

https://bugzilla.redhat.com/show_bug.cgi?id=1798051

**Reviewer Checklist**


awgreene:
This breaks the use of controller-gen.

kevinrizza:
@awgreene 

I took a look at this. It seems like we have a few things we need to do:

- Update `make release` to include the controller gen target
- Update the release documentation to reference this change

Additionally, I did make a first pass attempt at this, but it seems like controller-gen doesn't support this directly out of the box with the project layout we have right now. Some problems I see:

There isn't a clear way to define crd versions as storage only.

Even if there was, controller-gen is run against individual go packages, and right now we have multiple crd definitions in each package. So if we were to do this it would break in the event that we migrated any of our other packages onto the `v1alpha2` version if there was a conflict about what was a storage version and what was a storage and serving version.

Give this, I think integrating `make manifests` into the core release process at the moment for anything other than generating the openapi validation spec is going to be some additional work and is outside the scope of this bugfix by itself. I think we should merge this PR and track that work elsewhere.

njhale:
/hold

Seems like we're having some issues w/ an [unrelated test](https://search.svc.ci.openshift.org/?search=FAIL%3A+TestInstallPlanUnpacksBundleImage&maxAge=168h&context=-1&type=all). Let's get that cleaned up first so this doesn't spam CI.

awgreene:
/retest

`TestInstallPlanUnpacksBundleImage` works just fine locally.

awgreene:
Note, `TestInstallPlanUnpacksBundleImage` works on an OpenShift cluster as well...

awgreene:
Closing this in favor of #1300

PR #1292, opened 2020-02-14
[release-4.3] Bug 1803135: fix(packagemanifests): Fix index key func to avoid key collision

dinhxuanvu:

**Reviewer Checklist**


njhale:
/retest

Let's keep testing, but from the failures it's starting to look like this patch may be triggering some issues w/ InstallPlans.

dinhxuanvu:
I'm running couple things in a separate cluster to verify if this is a flake or something else more problematic.

dinhxuanvu:
@benluddy @njhale This PR is good to do so if either of you can give it a lgtm/approve, that would be great. Thanks.

benluddy:
> I'm running couple things in a separate cluster to verify if this is a flake or something else more problematic.

What did you find? It's a little concerning that this test failed three times on this PR only: https://search.svc.ci.openshift.org/?search=FAIL%3A+TestSubscriptionInstallPlanStatus&maxAge=336h&context=2&type=all, though I don't see the connection.

dinhxuanvu:
This PR failed on several different test cases before reaching that 3 runs of that specific test case and then it failed in a different test afterwards. It all seemed like flakes. That test case doesn't have anything to do with this fix so I'm convinced it is flake. From the error log, it seemed to have some sort of race-condition which caused the InstallPlan got updated after the test got a local copy of it to some changes on its own. It does happen from time to time with all reconciliation that we do behind the scene.

PR #1301, opened 2020-02-19
Bug 1804812: fix(deployment): deployment spec hash

ecordell:
Use deployment spec hash for determining if change is needed on cluster

DeepDerivative will miss cases where values are removed, because an
empty value is treated as unchanged.

This adds some duplication to the apiservice handling, but I avoided doing any serious refactoring to ensure this cherry picks cleanly.

We had been using deep derivative to compare the operator's view of what the deployment should be to the cluster's. But that has a lot of edge cases (zero values for ints, unsetting previously set values), so instead we just store a hash of the deployment spec on the resulting deployment and recalculate it when we need to compare.

**Reviewer Checklist**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,12 +40,12 @@ all: test build
 test: clean cover.out
 
 unit:
-	go test $(MOD_FLAGS) $(SPECIFIC_UNIT_TEST) -v -race -tags=json1 -count=1 ./pkg/...
+	go test $(MOD_FLAGS) $(SPECIFIC_UNIT_TEST) -v -race -count=1 ./pkg/...
 
 schema-check:
 
 cover.out: schema-check
-	go test $(MOD_FLAGS) -v -race -tags=json1 -coverprofile=cover.out -covermode=atomic \
+	go test $(MOD_FLAGS) -v -race -coverprofile=cover.out -covermode=atomic \
 		-coverpkg ./pkg/controller/... ./pkg/...
 
 coverage: cover.out
--- a/pkg/controller/install/deployment.go
+++ b/pkg/controller/install/deployment.go
@@ -2,17 +2,21 @@ package install
 
 import (
 	"fmt"
+	"hash/fnv"
+	"k8s.io/apimachinery/pkg/util/rand"
 
 	log "github.com/sirupsen/logrus"
 	appsv1 "k8s.io/api/apps/v1"
-	"k8s.io/apimachinery/pkg/api/equality"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/apimachinery/pkg/util/diff"
+	hashutil "k8s.io/kubernetes/pkg/util/hash"
 
+	v1 "github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/wrappers"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/ownerutil"
 )
+
+const DeploymentSpecHashLabelKey = "olm.deployment-spec-hash"
 
 type StrategyDeploymentInstaller struct {
 	strategyClient      wrappers.InstallStrategyDeploymentInterface
@@ -68,7 +72,7 @@ func NewStrategyDeploymentInstaller(strategyClient wrappers.InstallStrategyDeplo
 
 func (i *StrategyDeploymentInstaller) installDeployments(deps []v1alpha1.StrategyDeploymentSpec) error {
 	for _, d := range deps {
-		deployment, err := i.deploymentForSpec(d.Name, d.Spec)
+		deployment, _, err := i.deploymentForSpec(d.Name, d.Spec)
 		if err != nil {
 			return err
 		}
@@ -81,7 +85,7 @@ func (i *StrategyDeploymentInstaller) installDeployments(deps []v1alpha1.Strateg
 	return nil
 }
 
-func (i *StrategyDeploymentInstaller) deploymentForSpec(name string, spec appsv1.DeploymentSpec) (deployment *appsv1.Deployment, err error) {
+func (i *StrategyDeploymentInstaller) deploymentForSpec(name string, spec appsv1.DeploymentSpec) (deployment *appsv1.Deployment, hash string, err error) {
 	dep := &appsv1.Deployment{Spec: spec}
 	dep.SetName(name)
 	dep.SetNamespace(i.owner.GetNamespace())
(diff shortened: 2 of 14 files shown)


pkg/controller/install/deployment.go
```diff
@@ -104,6 +107,11 @@ func (i *StrategyDeploymentInstaller) deploymentForSpec(name string, spec appsv1
 		return
 	}
 
+	// Calculate hash without template annotations
```
exdx: Interesting - why do we calculate the deployment hash without the annotations?

pkg/controller/install/deployment.go
```diff
@@ -280,3 +269,10 @@ func (i *StrategyDeploymentInstaller) cleanupOrphanedDeployments(deploymentSpecs
 
 	return nil
 }
+
+func HashDeploymentSpec(spec *appsv1.DeploymentSpec) string {
+	spec.Template.SetAnnotations(nil)
```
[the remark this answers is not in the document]
ecordell: good catch!

test/e2e/operator_groups_e2e_test.go
```diff
@@ -2099,11 +2100,14 @@ func TestCleanupCsvsWithBadOwnerOperatorGroups(t *testing.T) {
...
-	_, err = crc.OperatorsV1alpha1().ClusterServiceVersions(otherNamespaceName).Update(fetchedCSV)
-	require.NoError(t, err)
+	updateCSV := func() error {
+		fetchedCSV, err := crc.OperatorsV1alpha1().ClusterServiceVersions(otherNamespaceName).Get(csvName, metav1.GetOptions{})
+		require.NoError(t, err)
+		fetchedCSV.Annotations[v1.OperatorGroupNamespaceAnnotationKey] = fetchedCSV.GetNamespace()
+		_, err = crc.OperatorsV1alpha1().ClusterServiceVersions(otherNamespaceName).Update(fetchedCSV)
+		return err
+	}
+	require.NoError(t, retry.RetryOnConflict(retry.DefaultBackoff, updateCSV))
```
njhale: cool -- common source of flakes?

pkg/controller/operators/olm/apiservices.go
```diff
@@ -339,6 +343,119 @@ func (a *Operator) installOwnedAPIServiceRequirements(csv *v1alpha1.ClusterServi
...
+		mount := corev1.VolumeMount{
+			Name:      volume.Name,
+			MountPath: "/apiserver.local.config/certificates",
+		}
+		for i, container := range depSpec.Template.Spec.Containers {
+			found := false
+			for j, m := range container.VolumeMounts {
+				if m.Name == mount.Name {
+					found = true
+					break
```
awgreene: Is there a reason we don't need to check the path? It seems like we're hoping that `apiservice-cert` is unique to olm and not set by someone else.

pkg/controller/operators/olm/apiservices.go
```diff
@@ -339,6 +343,119 @@ func (a *Operator) installOwnedAPIServiceRequirements(csv *v1alpha1.ClusterServi
...
+				}
+
+				// Replace if mounting to the same location.
+				if m.MountPath == mount.MountPath {
+					container.VolumeMounts[j] = mount
+					found = true
```
awgreene: Couldn't the container have multiple VolumeMounts that mount to the same location?

pkg/controller/operators/olm/apiservices.go
```diff
@@ -339,6 +343,119 @@ func (a *Operator) installOwnedAPIServiceRequirements(csv *v1alpha1.ClusterServi
...
+					break
+				}
+			}
+			if !found {
+				container.VolumeMounts = append(container.VolumeMounts, mount)
+			}
+
+			depSpec.Template.Spec.Containers[i] = container
+		}
+		depSpec.Template.ObjectMeta.SetAnnotations(map[string]string{OLMCAHashAnnotationKey: caHash})
```
awgreene: So basically, this entire function is meant to update the sdd to have any Volumes and VolumeMounts that may have been introduced by an API Service so the correct hash is generated?

pkg/controller/operators/olm/apiservices.go
```diff
@@ -526,7 +643,7 @@ func (a *Operator) installAPIServiceRequirements(desc v1alpha1.APIServiceDescrip
 		ownerutil.AddNonBlockingOwner(secretRoleBinding, csv)
 		_, err = a.opClient.CreateRoleBinding(secretRoleBinding)
 		if err != nil {
-			log.Warnf("could not create secret rolebinding with dep spec: %+v", depSpec)
+			log.Warnf("could not create secret rolebinding with dep spec: %#v", depSpec)
```
awgreene: This change just adds the package to the go struct right?

Makefile
```diff
@@ -40,12 +40,12 @@ all: test build
 test: clean cover.out
 
 unit:
-	go test $(MOD_FLAGS) $(SPECIFIC_UNIT_TEST) -v -race -tags=json1 -count=1 ./pkg/...
+	go test $(MOD_FLAGS) $(SPECIFIC_UNIT_TEST) -v -race -count=1 ./pkg/...
```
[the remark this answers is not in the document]
ecordell: it was needed (for tests that used sqlite) when we were using the json sqlite extensions. we don't use them anymore though

pkg/controller/install/deployment.go
```diff
@@ -204,45 +209,26 @@ func (i *StrategyDeploymentInstaller) checkForDeployments(deploymentSpecs []v1al
...
 		}
 
-		// check equality
-		calculated, err := i.deploymentForSpec(spec.Name, spec.Spec)
-		if err != nil {
-			return err
+		// check that the deployment spec hasn't changed since it was created
+		labels := dep.GetLabels()
+		if len(labels) == 0 {
+			return StrategyError{Reason: StrategyErrDeploymentUpdated, Message: fmt.Sprintf("deployment doesn't have a spec hash, update it")}
```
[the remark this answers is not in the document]
ecordell: true, but I think this will be clearer to us when we're reading logs about it

pkg/controller/operators/olm/apiservices.go
ecordell: this is all replicating the current apiservice generation logic - definitely there is room for improvement

pkg/controller/operators/olm/apiservices.go
ecordell: yep!

pkg/controller/operators/olm/apiservices.go
ecordell: off the top of my head, I don't think so. you can mount into the same directory, but I think you need different mount paths. I would need to test to be sure though.

awgreene:
/lgtm

Thanks for answering my questions.

PR #1312, opened 2020-02-21
[release-4.1] Bug 1805576: Remove run-level 1 from olm and openshift-operators namespaces

njhale:
OLM and the operators that it deploys shouldn't have run-level 1
and should use SCC as "anyuid" or "restricted".


spadgett:
Console test fix in https://github.com/openshift/console/pull/5058

spadgett:
Console test fix in openshift/console#5058

PR #1313, opened 2020-02-21
add link to olm-book definition of subscription

madorn:
direct traffic to olm-book's definition of subscription


ecordell:
Not sure why this didn't auto merge.

PR #1316, opened 2020-02-21
Bug 1784024: Use generations to prevent duplicate InstallPlans

njhale:
Supersedes #1203


pkg/controller/operators/catalog/operator.go
```diff
@@ -753,6 +753,7 @@ func (o *Operator) syncResolvingNamespace(obj interface{}) error {
 	}
 
 	// TODO: parallel
+	maxGeneration := 0
```
dinhxuanvu: Is this variable name a bit confusing? It implies max but it is more like current. Am I wrong here?

njhale: This is intended to be the max generation among existing InstallPlans.

dinhxuanvu: I see. Thanks.

njhale:
/hold

[this failure](https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/operator-framework_operator-lifecycle-manager/1316/pull-ci-operator-framework-operator-lifecycle-manager-master-e2e-aws-olm/4138) is concerning. Investigating...

pkg/controller/operators/catalog/operator.go
```diff
@@ -965,44 +971,16 @@ func (o *Operator) ensureInstallPlan(logger *logrus.Entry, namespace string, sub
...
-			res, err := o.client.OperatorsV1alpha1().InstallPlans(namespace).UpdateStatus(out)
-			if err != nil {
-				return nil, err
-			}
-			return reference.GetReference(res)
+		if installPlan.Spec.Generation == gen {
+			return reference.GetReference(installPlan)
 		}
 	}
 	logger.Warn("no installplan found with matching manifests, creating new one")
```
ecordell: nit: not really an accurate log statement anymore

pkg/controller/operators/catalog/operator.go
```diff
@@ -813,12 +818,12 @@ func (o *Operator) syncResolvingNamespace(obj interface{}) error {
 			}
 		}
 
-		installPlanReference, err := o.ensureInstallPlan(logger, namespace, subs, installPlanApproval, steps, bundleLookups)
+		installPlanReference, err := o.ensureInstallPlan(logger, namespace, maxGeneration+1, subs, installPlanApproval, steps, bundleLookups)
```
ecordell: can this use `updatedSubs` instead of `subs`?

njhale: It's probably best to use `subs` since we want to apply `OwnerReferences` to the generated `InstallPlan` referencing all `Subscriptions` in the namespace. I think we also want to use `subs` a few lines below for the `Subscription` update so they point to the latest `InstallPlan`.

I'm probably missing some edge case though; e.g. bad `Subscriptions` should reference the latest `InstallPlan`.

pkg/controller/operators/catalog/operator.go
```diff
@@ -936,24 +942,52 @@ func (o *Operator) ensureSubscriptionCSVState(logger *logrus.Entry, sub *v1alpha
...
-			err = subErr
-		}
+
+		wg.Add(1)
+		go func(s v1alpha1.Subscription) {
+			defer wg.Done()
+
+			update := func() error {
+				// Update the status of the latest revision
+				latest, err := o.client.OperatorsV1alpha1().Subscriptions(s.GetNamespace()).Get(s.GetName(), getOpts)
```
ecordell: Are you seeing flakes in the e2e tests here?

We may want to use the ‘retry’ package here. Can’t the subscription status update loop conflict with this? 

Another option would be to determine what the current install plan generation is during the subscription update loop.

PR #1319, opened 2020-02-24
Add stale bot

Bowenislandsong:
stale bot automatically marks an inactive issue after 60 days before closing. This helps clean up stale issues and remind issues on the backburner.

**Reviewer Checklist**

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,17 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 60
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 7
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - pinned
+  - security
+# Label to use when marking an issue as stale
+staleLabel: wontfix
+# Comment to post when marking an issue as stale. Set to `false` to disable
+markComment: >
+  This issue has been automatically marked as stale because it has not had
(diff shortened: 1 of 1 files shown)


.github/stale.yml
```diff
@@ -0,0 +1,17 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 60
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 7
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - pinned
+  - security
```
awgreene: Does OpenShift use these labels?

Bowenislandsong: I wonder if we do want them as exceptions? If pinned/security, it does not close? Most of the other Openshift repos do not use github bot. They use openshift bot. For upstream purposes, we would prefer github-bot and maybe we want these exception labels? we can remove them too. Thanks.

awgreene: >I wonder if we do want them as exceptions? If pinned/security, it does not close? 
I certainly believe that we want a way to prevent the bot from marking an issue as stale. If you believe these 2 labels handles reasons OLM would want to prevent the bot from closing the issue, we can mark this as resolved.

> For upstream purposes, we would prefer github-bot and maybe we want these exception labels? we can remove them too. Thanks.
SGTM

PR #1322, opened 2020-02-25
Bug 1799463: Duplicate packages in packageserver APIService response 

exdx:
Bug report of duplicate packages being returned by the default packageserver APIService shipped with OLM. See BZ for details.  

It looks like OLM ignores the packagemanifest label selector when listing in all namespaces:

`$ oc get packagemanifests -l '!opsrc-owner-name' -n openshift-marketplace
No resources found.`

`$ oc get packagemanifests -l '!opsrc-owner-name' --all-namespaces | grep openshift-marketplace | wc -l
     198`

This is causing console to have duplicate items returned.

**Reviewer Checklist**

### DIFF
--- a/pkg/package-server/provider/registry.go
+++ b/pkg/package-server/provider/registry.go
@@ -412,7 +412,7 @@ func (p *RegistryProvider) Get(namespace, name string) (*operators.PackageManife
 func (p *RegistryProvider) List(namespace string, selector labels.Selector) (*operators.PackageManifestList, error) {
 	var pkgs []*operators.PackageManifest
 	if namespace == metav1.NamespaceAll {
-		all, err := p.pkgLister.List(labels.Everything())
+		all, err := p.pkgLister.List(selector)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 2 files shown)


exdx:
Test looks ok but need to validate via the UI that this is working as expected as well. 
/hold

pkg/package-server/provider/registry.go
```diff
@@ -412,9 +412,18 @@ func (p *RegistryProvider) Get(namespace, name string) (*operators.PackageManife
 func (p *RegistryProvider) List(namespace string, selector labels.Selector) (*operators.PackageManifestList, error) {
 	var pkgs []*operators.PackageManifest
 	if namespace == metav1.NamespaceAll {
-		all, err := p.pkgLister.List(labels.Everything())
-		if err != nil {
-			return nil, err
+		var all []*operators.PackageManifest
+		var err error
+		if selector.Empty() {
+			all, err = p.pkgLister.List(labels.Everything())
```
benluddy: It looks like a selector for which `Empty()` returns true is equivalent to a selector returns by `labels.Everything()`. Could the same change be made by substituting `labels.Everything()` with `selector` in the original L415?

exdx: I see what you mean - it would certainly simplify this logic. I will try it and add a test case to check 👍

exdx:
Validated via UI that this is working as expected
/hold cancel

PR #1323, opened 2020-02-25
Bug 1807128: Don't block on ctx.Done() if startup fails.

benluddy:
Operator startup errors aren't recoverable without restarting the
process, so it's critical that the operator process is allowed to
terminate in these cases.

### DIFF
--- a/pkg/lib/queueinformer/queueinformer_operator_test.go
+++ b/pkg/lib/queueinformer/queueinformer_operator_test.go
@@ -18,7 +18,7 @@ func (f versionFunc) ServerVersion() (*version.Info, error) {
 	return (func() (*version.Info, error))(f)()
 }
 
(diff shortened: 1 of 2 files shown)


pkg/lib/queueinformer/queueinformer_operator_test.go
```diff
@@ -78,10 +78,16 @@ func TestOperatorRunReadyChannelClosed(t *testing.T) {
 
 			o.Run(ctx)
 
-			select {
-			case <-o.Ready():
-			case <-time.After(time.Second):
-				t.Error("timed out before ready channel closed")
+			timeout := time.After(time.Second)
+			for n, ch := range map[string]<-chan struct{}{
```
njhale: Clever!

Issue #1298, opened 2020-02-18
Unable to override resources and tolerations for an operator deployment


secat:
I am currently unable to override the `resources` and `tolerations` field for an operator deployment. According to the `subscriptions.operators.coreos.com` CRD, there is a field named `subscription.spec.config` to do so.

I have deployed a `subscription` CR using the following manifest:
```yaml
apiVersion: operators.coreos.com/v1alpha1
kind: Subscription
metadata:
  name: tracing-operator-subscription
  namespace: fleet-system
spec:
  channel: preview
  name: tracing-operator
  source: tracing-operator-catalog
  sourceNamespace: fleet-system
  config:
    resources:
      limits:
        memory: 128Mi
      requests:
        cpu: 100m
        memory: 128Mi
    tolerations:
    - key: workload-isolation.infra.fleet
      operator: Equal
      value: default
      effect: NoSchedule
```

Here is the status:
```yaml
status:
  catalogHealth:
  - catalogSourceRef:
      apiVersion: operators.coreos.com/v1alpha1
      kind: CatalogSource
      name: tracing-operator-catalog
      namespace: fleet-system
      resourceVersion: "14510590"
      uid: 1ade9b61-2822-4986-887c-17867048b4e5
    healthy: true
    lastUpdated: "2020-02-18T16:13:29Z"
  conditions:
  - lastTransitionTime: "2020-02-18T16:13:29Z"
    message: all available catalogsources are healthy
    reason: AllCatalogSourcesHealthy
    status: "False"
    type: CatalogSourcesUnhealthy
  currentCSV: tracing-operator.v0.3.0
  installPlanRef:
    apiVersion: operators.coreos.com/v1alpha1
    kind: InstallPlan
    name: install-9wdlk
    namespace: fleet-system
    resourceVersion: "14519056"
    uid: 532649d6-5c9f-4060-be9f-3e83d7b84775
  installedCSV: tracing-operator.v0.3.0
  installplan:
    apiVersion: operators.coreos.com/v1alpha1
    kind: InstallPlan
    name: install-9wdlk
    uuid: 532649d6-5c9f-4060-be9f-3e83d7b84775
  lastUpdated: "2020-02-18T16:13:32Z"
  state: AtLatestKnown
```

These overrides are missing in the automatic generated `installplan` CRD.

Here is the generated `deployment` manifest:
```yaml
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  annotations:
[...]

PR #1334, opened 2020-02-28
Tolerations & Resources Override using .Subscription.Config

umangachapagain:
(feat) inject 'Tolerations' & 'Resources' from config

Updated deployment initializer function to inject tolerations
and resources specified in pod configuration of subscription
into deployment object.

- Toleration will be appended if it does not exist
- Resources will be overwritten

(test) update e2e test

#1298 

**Reviewer Checklist**

### DIFF
--- a/doc/design/subscription-config.md
+++ b/doc/design/subscription-config.md
@@ -84,3 +84,52 @@ spec:
     - mountPath: /config
       name: config-volume
 ```
+
+### Tolerations
+
+The `tolerations` field defines a list of [Tolerations](https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/) for the Pod created by OLM.
+
+> Note: Tolerations defined here will be appended to existing Tolerations, if not already present.
+
+#### Example
+
+Inject toleration to tolerate all taints.
+
+```
+kind: Subscription
+metadata:
+  name: my-operator
+spec:
+  package: etcd
+  channel: alpha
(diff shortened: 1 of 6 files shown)


hochdorf:
@umangachapagain was this feature intentionally planned to delete the resources section from the operator deployment even though the related config section is not defined in the subscription?

Based on the code changes and based on my tests if the resources config is not defined in the subscription but it was given in the CSV for the operator then it will be removed from the deployment during installation.

We define the resource needs of an operator on the CSV level because in this way you can define it differently for the separate versions.
After this feature, it's not possible.

Is this behaviour conceptual? Or just a bug?

umangachapagain:
> @umangachapagain was this feature intentionally planned to delete the resources section from the operator deployment even though the related config section is not defined in the subscription?
> 
> Based on the code changes and based on my tests if the resources config is not defined in the subscription but it was given in the CSV for the operator then it will be removed from the deployment during installation.
> 
> We define the resource needs of an operator on the CSV level because in this way you can define it differently for the separate versions.
> After this feature, it's not possible.
> 
> Is this behaviour conceptual? Or just a bug?
>
It's a bug. The intention was to override resources spec, but not blindly.

Please report a bug with your findings.

hochdorf:
> umangachapagain

Thanks! 
https://github.com/operator-framework/operator-lifecycle-manager/issues/1912

PR #1339, opened 2020-03-02
Bug 1805412: Support multiple APIServices on a single deployment

awgreene:


pkg/controller/operators/olm/apiservices.go
```diff
@@ -339,31 +358,52 @@ func (a *Operator) installOwnedAPIServiceRequirements(csv *v1alpha1.ClusterServi
...
 	if desc.ContainerPort > 0 {
 		containerPort = int(desc.ContainerPort)
 	}
+	return corev1.ServicePort{
+		Port:       int32(containerPort),
+		TargetPort: intstr.FromInt(containerPort),
+	}
+}
+
+func contains(list []corev1.ServicePort, element corev1.ServicePort) bool {
```
kevinrizza: This func could probably use a better name. `containsServicePort()`?

Bowenislandsong:
lgtm except for the format and the conflicts. Thanks!

pkg/controller/operators/olm/apiservices.go
```diff
@@ -339,31 +359,53 @@ func (a *Operator) installOwnedAPIServiceRequirements(csv *v1alpha1.ClusterServi
...
+		if !contains(result, getPort(desc)) {
+			result = append(result, getPort(desc))
+		}
+	}
 
-	// Create a service for the deployment
+	return result
+}
+
+func getPort(desc v1alpha1.APIServiceDescription) corev1.ServicePort {
```
harishsurf: nit: getServicePort( ) could be a better name?

pkg/controller/operators/olm/apiservices.go
```diff
@@ -277,6 +277,15 @@ func (a *Operator) areAPIServicesAvailable(csv *v1alpha1.ClusterServiceVersion)
 	return true, nil
 }
 
+func apiServiceDescriptionsForSSD(deploymentName string, descs []v1alpha1.APIServiceDescription) []v1alpha1.APIServiceDescription {
```
exdx: What's SSD referring to? The function is straightforward but could benefit from some comments. Is it referring to strategyDetailsDeployment (SDD?)

pkg/controller/operators/olm/apiservices.go
```diff
@@ -98,7 +98,7 @@ func (a *Operator) checkAPIServiceResources(csv *v1alpha1.ClusterServiceVersion,
 			return utilerrors.NewAggregate(errs)
 		}
 
-		serviceName := APIServiceNameToServiceName(apiServiceName)
+		serviceName := desc.DeploymentName
```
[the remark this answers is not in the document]
awgreene: I don't believe this is an issue - `oc get packagemanifests` worked after deploying OLM.

pkg/controller/operators/olm/apiservices.go
awgreene: That is correct - I will make sure to improve function names and comments in the next update!

pkg/controller/operators/olm/apiservices.go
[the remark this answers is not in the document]
njhale: Is the deployment name for packageserver not a valid DNS label?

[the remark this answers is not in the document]
awgreene: Ah - reviewed the link you had sent earlier - the packageserver name is indeed a valid dns label.

pkg/controller/operators/olm/apiservices.go
```diff
@@ -303,24 +312,35 @@ func (a *Operator) installOwnedAPIServiceRequirements(csv *v1alpha1.ClusterServi
 	}
 	rotateAt := expiration.Add(-1 * DefaultCertMinFresh)
 
+	apiDescs := csv.GetOwnedAPIServiceDescriptions()
 	depSpecs := make(map[string]appsv1.DeploymentSpec)
 	for _, sddSpec := range strategyDetailsDeployment.DeploymentSpecs {
+		descs := apiServiceDescriptionsForSSD(sddSpec.Name, apiDescs)
 		depSpecs[sddSpec.Name] = sddSpec.Spec
```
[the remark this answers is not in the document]
awgreene: It is primarily used to overwrite the existing strategyDetailsDeployment spec on Line [L358](https://github.com/operator-framework/operator-lifecycle-manager/pull/1339/files#diff-64a1aa5b985cd270ad2d1816018de5c2R358)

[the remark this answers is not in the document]
awgreene: Spoke offline - depSpecs is not required and will be removed in the next update.

awgreene:
/retest

```
Installing from release registry.svc.ci.openshift.org/ci-op-lrlc5yz5/release@sha256:cc4bf2980b6d1c4ec60580dae595f0417c6f484fb41d8ea3b1b619a31b1b2341
AWS region: us-west-2 (zones: us-west-2a us-west-2b)
level=info msg="Credentials loaded from the \"default\" profile in file \"/etc/openshift-installer/.awscred\""
level=warning msg="Found override for release image. Please be warned, this is not advised"
level=info msg="Consuming Install Config from target directory"
level=info msg="Creating infrastructure resources..."
level=error
level=error msg="Error: error adding tags: error tagging resource (ami-08c6dc33d1e7103a4): InvalidAMIID.NotFound: The image id '[ami-08c6dc33d1e7103a4]' does not exist"
level=error msg="\tstatus code: 400, request id: 1c184e9c-489c-4d49-a50c-bb367cf9004b"
level=error
level=error msg="  on ../tmp/openshift-install-166717711/main.tf line 104, in resource \"aws_ami_copy\" \"main\":"
level=error msg=" 104: resource \"aws_ami_copy\" \"main\" {"
level=error
level=error
level=fatal msg="failed to fetch Cluster: failed to generate asset \"Cluster\": failed to create cluster: failed to apply using Terraform"
```

pkg/controller/operators/olm/apiservices.go
```diff
@@ -921,3 +978,53 @@ func (a *Operator) isAPIServiceAdoptable(target *v1alpha1.ClusterServiceVersion,
...
 }
+
+// deleteLegacyAPIServiceResources deletes resources that were created by OLM for an APIService that used the old naming convention.
+func (a *Operator) deleteLegacyAPIServiceResources(namespace string, desc v1alpha1.APIServiceDescription) error {
+	apiServiceName := fmt.Sprintf("%s.%s", desc.Version, desc.Group)
+
+	// Attempt to delete the legacy service
+	err := a.opClient.DeleteService(namespace, legacyAPIServiceNameToServiceName(apiServiceName), &metav1.DeleteOptions{})
+	if err != nil && !k8serrors.IsNotFound(err) {
+		return err
```
anik120: Could we not try to go on deleting the other resources, and return an aggregated error, instead of returning an error immediately and letting the other resources hang around?

awgreene: We could - but these errors are only returned if OLM encounters an error other than IsNotFound - likely a connection error in which case all other requests would fail as well.

Additionally, the CSV should continue to be reconciled, so the install should eventually succeed.

pkg/controller/operators/olm/apiservices.go
```diff
@@ -921,3 +978,53 @@ func (a *Operator) isAPIServiceAdoptable(target *v1alpha1.ClusterServiceVersion,
 	adoptable = ownerutil.AdoptableLabels(apiService.GetLabels(), true, owners...)
 	return
 }
+
+// deleteLegacyAPIServiceResources deletes resources that were created by OLM for an APIService that used the old naming convention.
+func (a *Operator) deleteLegacyAPIServiceResources(namespace string, desc v1alpha1.APIServiceDescription) error {
```
ecordell: everything here should probably do a `get` from the lister to check that these resources are owned by OLM before delete.

awgreene: What is the correct way to ensure that the resources are owned by OLM?

awgreene: A check for a CSV OwnerReference or OwnerLabel?

ecordell: Yep exactly. I think there's already some code in this package for checking if a resource is adoptable.

pkg/controller/operators/olm/apiservices.go
```diff
@@ -830,19 +888,25 @@ func (a *Operator) installAPIServiceRequirements(desc v1alpha1.APIServiceDescrip
 		}
 
 		if !adoptable {
-			return nil, fmt.Errorf("pre-existing APIService %s is not adoptable", apiServiceName)
+			return fmt.Errorf("pre-existing APIService %s is not adoptable", fmt.Sprintf("%s.%s", desc.Version, desc.Group))
```
ecordell:
```suggestion
			return fmt.Errorf("pre-existing APIService %s.%s is not adoptable", desc.Version, desc.Group)
```

pkg/controller/operators/olm/apiservices.go
```diff
@@ -921,3 +978,119 @@ func (a *Operator) isAPIServiceAdoptable(target *v1alpha1.ClusterServiceVersion,
...
+	logger := log.WithFields(log.Fields{
+		"ownerName":      owner.GetName(),
+		"ownerNamespace": owner.GetNamespace(),
+		"ownerKind":      owner.GetObjectKind().GroupVersionKind().GroupKind().Kind,
+	})
+	namespace := owner.GetNamespace()
+	apiServiceName := fmt.Sprintf("%s.%s", desc.Version, desc.Group)
+
+	// Attempt to delete the legacy Service.
+	existingService, err := a.opClient.GetService(namespace, legacyAPIServiceNameToServiceName(apiServiceName))
```
ecordell: Double checking an edge case.

Given:

```go
func serviceName(deploymentName string) string {
	return deploymentName + "-service"
}
```
```go
func legacyAPIServiceNameToServiceName(apiServiceName string) string {
	return strings.Replace(apiServiceName, ".", "-", -1)
}
```

`deploymentName + "-service"` can never equal `strings.Replace(apiServiceName, ".", "-", -1)` because `apisServiceName` must be `<group>.<version>` and `<version>` cannot be `service`?

(I'm just double checking that there is no case where OLM creates "new" resources that could be detected as "legacy" resources)

ecordell: I have the same question for all of these. I wonder if it might not make sense to increment the ownership label?

awgreene: I don't believe that this is an issue. This code is:
- Called after creating / updating the APIService and related resources with the new naming convention.
- the `legacyAPIServiceNameToServiceName` and similar calls are used to get the legacy APIService resources that were created based on the old naming convention
- the `serviceName` function is used hereon for the new naming convention.

awgreene: I spoke with Evan offline to learn more about his concern and came to the conclusion that it was valid. The following code should capture the edgecase he pointed out and handle it appropriately:
```	
// Handle an edgecase where the legacy resources names matches the new names.
// This only occurs when the Group of the description matches the name of the deployment
// and the version is equal to "service".
if legacyAPIServiceNameToServiceName(apiServiceName) == serviceName(desc.DeploymentName) {
	return nil
}
```

pkg/controller/operators/olm/apiservices.go
awgreene: Accepted in the latest version of the PR.

awgreene:
/retest
```
error: unable to read image registry.svc.ci.openshift.org/ci-op-0mngk4xd/stable-initial@sha256:595142acf537b1b17b8e340c2de6b73364c79034401d58f5b09dfbd6b742e5c8: received unexpected HTTP status: 500 Internal Server Error
```

Trying one more time...

PR #1341, opened 2020-03-03
Bug 1808113: Fix OwnerReferences for bundled CRs

njhale:

Fix InstallPlan execution to generate OwnerReferences on CRs applied to the respective CSV rather than the InstallPlan itself.

Currently, bundled CRs are given OwnerReferences to the InstallPlan they were applied from, which is incorrect.

### DIFF
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -1726,22 +1726,38 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
 					Resource: r.Name,
 				}
 
-				// Set up the dynamic client ResourceInterface
+				if step.Resolving != "" {
+					owner := &v1alpha1.ClusterServiceVersion{}
+					owner.SetNamespace(plan.GetNamespace())
+					owner.SetName(step.Resolving)
+
+					if r.Namespaced {
+						// Set OwnerReferences for namespace-scoped resource
+						ownerutil.AddNonBlockingOwner(unstructuredObject, owner)
+
+						// Update UIDs on all CSV OwnerReferences
+						updated, err := o.getUpdatedOwnerReferences(unstructuredObject.GetOwnerReferences(), plan.Namespace)
+						if err != nil {
+							return errorwrap.Wrapf(err, "error generating ownerrefs for unstructured object: %s", unstructuredObject.GetName())
+						}
+
+						unstructuredObject.SetOwnerReferences(updated)
+					} else {
+						// Add owner labels to cluster-scoped resource
+						if err := ownerutil.AddOwnerLabels(unstructuredObject, owner); err != nil {
+							return err
+						}
+					}
+				}
+
+				// Set up the dynamic client ResourceInterface and set ownerrefs
 				var resourceInterface dynamic.ResourceInterface
 				if r.Namespaced {
-					ownerutil.AddOwner(unstructuredObject, plan, false, false)
 					unstructuredObject.SetNamespace(namespace)
 					resourceInterface = o.dynamicClient.Resource(gvr).Namespace(namespace)
 				} else {
 					resourceInterface = o.dynamicClient.Resource(gvr)
 				}
-
-				// Update UIDs on all CSV OwnerReferences
-				updated, err := o.getUpdatedOwnerReferences(unstructuredObject.GetOwnerReferences(), plan.Namespace)
-				if err != nil {
-					return errorwrap.Wrapf(err, "error generating ownerrefs for unstructured object: %s", unstructuredObject.GetName())
-				}
-				unstructuredObject.SetOwnerReferences(updated)
 
 				// Ensure Unstructured Object
 				status, err := ensurer.EnsureUnstructuredObject(resourceInterface, unstructuredObject)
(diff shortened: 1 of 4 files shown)


pkg/controller/operators/catalog/testdata/prometheusrule.cr.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: rule 
+  # namespace: placeholder
```
awgreene: Can we remove this?

pkg/controller/operators/catalog/operator.go
```diff
@@ -1726,22 +1726,34 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
...
+					unstructuredObject.SetOwnerReferences(updated)
 				} else {
 					resourceInterface = o.dynamicClient.Resource(gvr)
-				}
 
-				// Update UIDs on all CSV OwnerReferences
-				updated, err := o.getUpdatedOwnerReferences(unstructuredObject.GetOwnerReferences(), plan.Namespace)
-				if err != nil {
-					return errorwrap.Wrapf(err, "error generating ownerrefs for unstructured object: %s", unstructuredObject.GetName())
+					// Add owner labels to cluster-scoped resource
```
awgreene: The CSV is namespaced - should we add it as an owner to a cluster scopped resource? This doesn't seem to be a good call based on the note in [1] - are OwnerLabels an OLM construct that we use to search for Cluster Scoped resources when OLM deletes a namespaced resource?

[1] https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/#owners-and-dependents

pkg/controller/operators/catalog/operator.go
```diff
@@ -1726,22 +1726,34 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
 					Resource: r.Name,
 				}
 
-				// Set up the dynamic client ResourceInterface
+				owner := &v1alpha1.ClusterServiceVersion{}
+				owner.SetNamespace(plan.GetNamespace())
+				owner.SetName(step.Resolving)
```
awgreene: Does `step.Resolving` always point to the CSV name?

pkg/controller/operators/catalog/operator_test.go
```diff
@@ -1017,8 +1087,13 @@ func NewFakeOperator(ctx context.Context, namespace string, namespaces []string,
...
 	// Create client fakes
 	clientFake := fake.NewReactionForwardingClientsetDecorator(config.clientObjs, config.clientOptions...)
-	opClientFake := operatorclient.NewClient(k8sfake.NewSimpleClientset(config.k8sObjs...), apiextensionsfake.NewSimpleClientset(config.extObjs...), apiregistrationfake.NewSimpleClientset(config.regObjs...))
+	// TODO: Using the ReactionForwardingClientsetDecorator for k8s objects causes issues with adding Resources for discovery.
+	// For now, directly use a SimpleClientset instead.
+	k8sClientFake := k8sfake.NewSimpleClientset(config.k8sObjs...)
+	k8sClientFake.Resources = apiResourcesForObjects(append(config.extObjs, config.regObjs...))
+	opClientFake := operatorclient.NewClient(k8sClientFake, apiextensionsfake.NewSimpleClientset(config.extObjs...), apiregistrationfake.NewSimpleClientset(config.regObjs...))
 	dynamicClientFake := fakedynamic.NewSimpleDynamicClient(runtime.NewScheme())
+	// dynamicClientFake.Resources = k8sClientFake.Resources
```
awgreene: Can we remove this comment?

pkg/controller/operators/catalog/operator_test.go
```diff
@@ -1205,17 +1280,17 @@ func objectReference(name string) *corev1.ObjectReference {
 	return &corev1.ObjectReference{Name: name}
 }
 
-func yamlFromFilePath(fileName string) (string, error) {
+func yamlFromFilePath(t *testing.T, fileName string) string {
```
awgreene: Non-blocking nit: A few of these functions could probably be moved to an `e2e-util.go` file.

pkg/controller/operators/catalog/operator.go
njhale: These are labels, not OwnerReferences, so I'd say yes.

pkg/controller/operators/catalog/testdata/prometheusrule.cr.yaml
njhale: Just the commented namespace field, right?

pkg/controller/operators/catalog/operator.go
njhale: It should for all InstallPlans generated by OLM -- but you're right, it could be more robust. I'll go ahead and add some bumpers.

pkg/controller/operators/catalog/operator_test.go
njhale: Good idea! I think a follow-up would involve making a helper package that can be used across all of OLM's tests.

pkg/controller/operators/catalog/testdata/prometheusrule.cr.yaml
awgreene: correct.

exdx:
/lgtm

this only applies to the ownership of the new prometheus-related CRs right? No other effect on previous releases

PR #1345, opened 2020-03-03
add example test for metric test

Bowenislandsong:
This commit includes an example setup for metric tests so that the metering team can set up their tests quickly.

/cc @awgreene

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -117,7 +117,8 @@ setup-bare: clean e2e.namespace
 
 # e2e test exculding the rh-operators directory which tests rh-operators and their metric cardinality.
 e2e:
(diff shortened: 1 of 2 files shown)


Bowenislandsong:
/retest
This does not touch any existing CI tests, failure are flaks.

test/rh-operators/metric_test.go
```diff
@@ -6,13 +6,15 @@ import (
 	"testing"
 	"time"
 
+	_ "github.com/openshift/origin/test/extended/prometheus"
```
Bowenislandsong: import metric test

test/rh-operators/metric_test.go
```diff
@@ -66,6 +68,7 @@ func TestRHOperatorMetricCardinality(t *testing.T) {
 		}
 
 		// Add metric cardinality tests here.
+		ginkgo.RunSpecs(t, "Prometheus")
```
Bowenislandsong: run ginkgo metric test.

PR #1348, opened 2020-03-04
Bug 1797747: Make provided API ClusterRoles be owned by the corresponding API.

benluddy:
The existing CSV owner labels lead to a neverending battle for
ownership in clusters with more than one CSV providing the same
API. The amount of additional work generated by the owner label
conflict manifested (not exclusively) as excessive operator
installation latency in affected clusters.

This conflict is resolved by abandoning CSV owner labels in favor of
OwnerReferences to the providing APIService or
CustomResourceDefinition.

Aggregation of API ClusterRoles to the providing OperatorGroup
ClusterRoles is also changed. Instead of labeling API ClusterRoles
based on the names of OperatorGroups providing the associated API,
each API ClusterRole is labeled based on the name of the API
itself. OperatorGroup ClusterRoles accumulate one label selector per
provided API.

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/olm/operator.go
+++ b/pkg/controller/operators/olm/operator.go
@@ -1093,7 +1093,7 @@ func (a *Operator) operatorGroupFromAnnotations(logger *logrus.Entry, csv *v1alp
 	}
 
 	// Target namespaces don't match
-	if targets != strings.Join(operatorGroup.Status.Namespaces, ",") {
+	if targets != operatorGroup.BuildTargetNamespaces() {
 		logger.Info("olm.targetNamespaces annotation doesn't match operatorgroup status")
 		return nil
 	}
@@ -1105,21 +1105,21 @@ func (a *Operator) operatorGroupForCSV(csv *v1alpha1.ClusterServiceVersion, logg
 	now := a.now()
 
 	// Attempt to associate an OperatorGroup with the CSV.
-	operatorGroups, err := a.client.OperatorsV1().OperatorGroups(csv.GetNamespace()).List(metav1.ListOptions{})
+	operatorGroups, err := a.lister.OperatorsV1().OperatorGroupLister().OperatorGroups(csv.GetNamespace()).List(labels.Everything())
 	if err != nil {
 		logger.Errorf("error occurred while attempting to associate csv with operatorgroup")
 		return nil, err
 	}
 	var operatorGroup *v1.OperatorGroup
 
-	switch len(operatorGroups.Items) {
+	switch len(operatorGroups) {
 	case 0:
 		err = fmt.Errorf("csv in namespace with no operatorgroups")
 		logger.Warn(err)
 		csv.SetPhaseWithEvent(v1alpha1.CSVPhaseFailed, v1alpha1.CSVReasonNoOperatorGroup, err.Error(), now, a.recorder)
 		return nil, err
 	case 1:
-		operatorGroup = &operatorGroups.Items[0]
+		operatorGroup = operatorGroups[0]
 		logger = logger.WithField("opgroup", operatorGroup.GetName())
 		if a.operatorGroupAnnotationsDiffer(&csv.ObjectMeta, operatorGroup) {
 			a.setOperatorGroupAnnotations(&csv.ObjectMeta, operatorGroup, true)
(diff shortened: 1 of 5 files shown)


pkg/controller/operators/olm/operator.go
```diff
@@ -1223,13 +1223,16 @@ func (a *Operator) transitionCSVState(in v1alpha1.ClusterServiceVersion) (out *v
 	}
 
 	// Check for intersecting provided APIs in intersecting OperatorGroups
-	options := metav1.ListOptions{
-		FieldSelector: fmt.Sprintf("metadata.name!=%s,metadata.namespace!=%s", operatorGroup.GetName(), operatorGroup.GetNamespace()),
+	allGroups, err := a.lister.OperatorsV1().OperatorGroupLister().List(labels.Everything())
+	otherGroups := make([]v1.OperatorGroup, 0, len(allGroups))
+	for _, g := range allGroups {
+		if g.GetName() != operatorGroup.GetName() || g.GetNamespace() != operatorGroup.GetNamespace() {
```
benluddy: I _think_ the previous FieldSelector was subtly wrong, and would have excluded OperatorGroups with the same name in different namespaces, right?

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -295,32 +295,39 @@ func (a *Operator) pruneProvidedAPIs(group *v1.OperatorGroup, groupProvidedAPIs
...
 		}
-	} else if err != nil {
-		a.logger.WithError(err).Errorf("Create cluster role failed: %v", clusterRole)
+	}
+
+	if reflect.DeepEqual(existingCR.Rules, clusterRole.Rules) && ownerutil.EnsureOwner(clusterRole, api) && labels.Equals(existingCR.Labels, clusterRole.Labels) {
+		return nil
+	}
+
+	clusterRole.Labels = labels.Merge(existingCR.Labels, clusterRole.Labels)
```
benluddy: These labels are going to accumulate. Maybe this needs to be refactored to run per-API instead of per-CSV so that the update logic will know the complete set of relevant OperatorGroups?

pkg/controller/operators/olm/operator.go
ecordell: Agreed - it should be an "or", which afaict you can't express in a field selector.

pkg/controller/operators/olm/operatorgroup.go
ecordell: That would make the most sense - 

`ensureProvidedAPIClusterRole` should just requeue the CRD or APIServices owned by the csv, and syncing on those resources should make sure these CRs exist with the right labels/owners

that would also be the place to GC these ClusterRoles (if we stuck with labels)

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -862,7 +878,7 @@ func (a *Operator) ensureOpGroupClusterRole(op *v1.OperatorGroup, suffix string)
 			ClusterRoleSelectors: []metav1.LabelSelector{
 				{
 					MatchLabels: map[string]string{
-						operatorGroupAggregrationKeyPrefix + suffix: op.GetName(),
+						operatorGroupAggregrationKeyPrefix + op.GetName() + "-" + suffix: "true",
```
ecordell: nit: should we have a `suffix` const that includes the `-`?

njhale:
> We have an e2e test for testing kube GC behavior - I would be a little more comfortable with this if we added one that tested ClusterScope -> ClusterScope Owners / GC

From the [k8s GC docs](https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/):

> 2) Cluster-scoped dependents can only specify cluster-scoped owners, but not namespace-scoped owners.

This note implies that ClusterScope -> ClusterScope works. Given that it's explicitly called out in the docs, I would classify the GC test as a nice-to-have.

pkg/controller/operators/olm/operatorgroup.go
njhale: or [`strings.Join(...)`](https://golang.org/pkg/strings/#Join)

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -872,6 +888,11 @@ func (a *Operator) ensureOpGroupClusterRole(op *v1.OperatorGroup, suffix string)
 	if err != nil {
 		return err
 	}
+
+	if _, err := a.lister.RbacV1().ClusterRoleLister().Get(clusterRole.Name); err == nil {
```
njhale: nit: may want to ensure the error is a 404 before continuing

pkg/lib/ownerutil/util.go
```diff
@@ -365,6 +367,18 @@ func InferGroupVersionKind(obj runtime.Object) error {
 			Version: v1.GroupVersion,
 			Kind:    "OperatorGroup",
 		})
+	case *apiregistrationv1.APIService:
+		objectKind.SetGroupVersionKind(schema.GroupVersionKind{
+			Group:   apiregistrationv1.GroupName,
+			Version: apiregistrationv1.SchemeGroupVersion.Version,
+			Kind:    "APIService",
+		})
+	case *apiextensionsv1beta1.CustomResourceDefinition:
```
njhale: Can we get `v1` now too?

benluddy:
/test unit

Appears to be a known flake, with fix WIP in https://github.com/operator-framework/operator-lifecycle-manager/compare/master...ecordell:og-flake#diff-0a19f44e9192894885902071db8d28cdR4266-R4270.

PR #1349, opened 2020-03-04
[release-4.4] Bug 1808113: Fix OwnerReferences for bundled CRs

njhale:

Fix InstallPlan execution to generate OwnerReferences on CRs applied to the respective CSV rather than the InstallPlan itself.

Currently, bundled CRs are given OwnerReferences to the InstallPlan they were applied from, which is incorrect.


njhale:
Superseded by #1351

PR #1350, opened 2020-03-04
Clarify roles and adds info about Package Server

### LINKED ISSUE #1346: Docs: Clarification opportunities in the Architecture document
Now that I've been working with OLM for a bit and have wrapped my head around the basics, I came away from reading the [architecture document](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/doc/design/architecture.md) with a few points of confusion and even some places where it seems like the document is contradicting itself. It would be great if the doc could be updated to clarify these and I'd just generally like to know better.

I hope this format is useful. I'd be glad to attempt a reword if I can get an appropriate understanding.

**What about PackageManifests?**
The first and second table in the Architecture document do not include the PackageManifest resources. Why? 

**What about the Package Server stuff?** 
The role of the 'Package Server' pods isn't discussed. How is it involved? Perhaps the 'Catalog (Registry) Design' section is related to this? 

**How does an InstallPlan actually get implemented?**
After reading through, I'm unclear who is responsible for what when resolving the InstallPlan. Is the Catalog Operator going to create resources in some instances and if that doesn't happen, then the OLM Operator will create them in other instances? 

I read this quote which leads me to believe that Catalog Operator creates all the resources for an operator to function.
"Once approved, _the Catalog Operator will create all of the resources in an InstallPlan_; this should then independently satisfy the OLM Operator, which will proceed to install the ClusterServiceVersions." 

However, the table at the top of the document says the OLM operator creates the Roles, RoleBindings ServiceAccounts and Deployments. Also, earlier in the document it says "OLM is responsible for installing applications." I'm assuming applications means operators specifically and the parts required for them to run. Additionally, the ClusterServiceVersion is identified at the top as being created by the Catalog Operator, not OLM Operator.

flickerfly:
Clarifies roles of the two Operators and adds info about Package Server's place in all of this, while making clear it is not a required component. Based on information that @ecordell provided in issue #1346 

I found the current documentation to be confusing around these areas and lacking some details.

**Reviewer Checklist**

Closes #1346

### DIFF
--- a/doc/design/architecture.md
+++ b/doc/design/architecture.md
@@ -11,11 +11,12 @@ Each of these Operators is responsible for managing the CRDs that are the basis 
 | CatalogSource         | catsrc         | Catalog | a repository of CSVs, CRDs, and packages that define an application                        |
 | Subscription          | sub        | Catalog | used to keep CSVs up to date by tracking a channel in a package                            |
 | OperatorGroup         | og     | OLM     | used to group multiple namespaces and prepare for use by an operator                     |
-| PackageManifest       |        | PackageServer | provides the user with an api presentation of the data a CatalogSource provides |
+
+In addition, a PackageManifest resource provides the user with an API presentation of the data a CatalogSource provides. This is not an actual CRD, but a simpler synthetic aggregated API resource. As a result, it cannot be managed in the same manner as a CRD.
 
 Each of these Operators are also responsible for creating resources:
 
-| Component | Creatable Resources        |
+| Component | Managed Resources          |
 |-----------|----------------------------|
 | OLM       | Deployment                 |
 | Catalog   | Service Account            |
--- a/doc/design/architecture.md
+++ b/doc/design/architecture.md
@@ -133,7 +133,7 @@ The Package Server provides a veneer API over the CatalogSource resources and th
 
 It is not neccessary to run this for OLM to function, but enhances the user experience by providing the information needed for Subscription generation to the user. In a Production scenario where no discovery work is needed and all Subscriptions are predefined, this may be removed.
 
-OpenShift/OKD GUIs will use this API extention to present operator package information such as CSVs, packages and CRDs found in the CatalogSource to users. Additionally, `kubectl` and `oc` users are able to issue commands like `kubectl get PackageManifests` to list out available Operators and then run `kubectl describe` on those resources to find the appropriate data to build a Subscription to it.
+Client GUI interfaces will use this API extention to present operator package information such as CSVs, packages and CRDs found in the CatalogSource to users. Additionally, CLI tools are able to examine the list of PackageManifests to list out available Operators and then examine the specific PackageManifests resources to find the appropriate data to build a Subscription to it.
 
 ## Catalog (Registry) Design
 
--- a/doc/design/architecture.md
+++ b/doc/design/architecture.md
@@ -1,6 +1,6 @@
 # Architecture
 
-OLM is composed of two Operators: the OLM Operator and the Catalog Operator. There is also the Package Server, an optional veneer API, that is deployed by default. 
+OLM is composed of two Operators: the OLM Operator and the Catalog Operator.
 
 Each of these Operators is responsible for managing the CRDs that are the basis for the OLM framework:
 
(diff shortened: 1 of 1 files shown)


doc/design/architecture.md
```diff
@@ -120,6 +127,14 @@ None --> UpgradeAvailable --> UpgradePending --> AtLatestKnown -+
 | UpgradePending   | `InstallPlan` has been created (referenced in `status.installplan`) to install a new CSV                   |
 | AtLatestKnown    | `status.installedCSV` matches the latest available CSV in catalog                                             |
 
+## Package Server
+
+The Package Server provides a veneer API over the CatalogSource resources and the information they provide to ease user interaction. 
+
+It is not neccessary to run this for OLM to function, but enhances the user experience by providing the information needed for Subscription generation to the user. In a Production scenario where no discovery work is needed and all Subscriptions are predefined, this may be removed.
+
+OpenShift/OKD GUIs will use this API extention to present operator package information such as CSVs, packages and CRDs found in the CatalogSource to users. Additionally, `kubectl` and `oc` users are able to issue commands like `kubectl get PackageManifests` to list out available Operators and then run `kubectl describe` on those resources to find the appropriate data to build a Subscription to it.
```
dinhxuanvu: This is upstream doc so I don't think we should reference to a specific platform like OpenShift/OKD here. I would prefer to remove all OpenShift/OKD and `oc` from this doc.

flickerfly: Good point, I rewrote that paragraph.

doc/design/architecture.md
```diff
@@ -11,17 +11,19 @@ Each of these Operators is responsible for managing the CRDs that are the basis
 | CatalogSource         | catsrc         | Catalog | a repository of CSVs, CRDs, and packages that define an application                        |
 | Subscription          | sub        | Catalog | used to keep CSVs up to date by tracking a channel in a package                            |
 | OperatorGroup         | og     | OLM     | used to group multiple namespaces and prepare for use by an operator                     |
+| PackageManifest       |        | PackageServer | provides the user with an api presentation of the data a CatalogSource provides |
```
kevinrizza: nit: PackageManifest isn't actually a CRD, it's just a synthetic aggregated api resource. You can't, for example, create a PackageManifest CR and have it stored in etcd. We should probably not put it in the list of CRDs.

doc/design/architecture.md
```diff
@@ -11,17 +11,19 @@ Each of these Operators is responsible for managing the CRDs that are the basis
...
-| Catalog  | ClusterServiceVersion      |
+| Component | Creatable Resources        |
+|-----------|----------------------------|
+| OLM       | Deployment                 |
+| Catalog   | Service Account            |
+| Catalog   | (Cluster)Roles             |
+| Catalog   | (Cluster)RoleBindings      |
+| Catalog   | Custom Resource Definition |
+| Catalog   | ClusterServiceVersion      |
+| Package Server | PackageManifest       |
```
kevinrizza: Same comment here, you can't actually create a packagemanifest resource. It needs to be derived from a catalog.

doc/design/architecture.md
```diff
@@ -82,10 +86,13 @@ Replacing --> Deleting
...
-A user that wishes to track a package in a channel creates a Subscription resource configuring the desired package, channel, and the catalog source from which to pull updates. When updates are found, an appropriate InstallPlan is written into the namespace on behalf of the user.
-Users can also create an InstallPlan resource directly, containing the names of the desired ClusterServiceVersions and an approval strategy and the Catalog Operator will create an execution plan for the creation of all of the required resources.
-Once approved, the Catalog Operator will create all of the resources in an InstallPlan; this should then independently satisfy the OLM Operator, which will proceed to install the ClusterServiceVersions.
+The Catalog Operator is responsible for resolving ClusterServiceVersions and the required resources they specify. It is also responsible for watching CatalogSources. It updates to packages in channels, and upgrading them (optionally automatically) to the latest available versions. It also starts and maintains pods required to provide the grpc endpoints from container images those CatalogSources identify.
```
kevinrizza:
```
It updates to packages in channels, and upgrading them (optionally automatically) to the latest available versions.
```

I think you're missing a verb here? Maybe:

`It aggregates updates to packages in channels`?

Or something similar?

doc/design/architecture.md
dinhxuanvu: I would recommend to change the word "creatable" to "Managed"

doc/design/architecture.md
dinhxuanvu: I think splitting up the original sentence into 2 sentences may cost some confusion here. Perhaps, just keep the original sentence as it is.

doc/design/architecture.md
flickerfly: I dealt with the first table by removing it and adding a paragraph below on the PackageManifest so it is close, but clear that it isn't the same. I also added @dinhxuanvu's word change idea.

doc/design/architecture.md
flickerfly: This is where I originally found confusion. I'll try again when I get a moment as my primary objective was to make it clear. :-) Thanks!

dinhxuanvu: I would suggest this:
`The Catalog Operator is responsible for resolving ClusterServiceVersions and the required resources they specify. It is also responsible for watching catalog sources for packages' updates in channels, and then upgrading those packages (optionally automatically) to the latest available versions.`
Does that sound better?

flickerfly: When we're inside the cluster and have read the catalog source. What actually is changing when we "upgrade a package"?

I know an InstallPlan is triggered and the Subscription's status is updated. Of course the InstallPlan triggers some operand changes as required. The PackageManifest isn't related to the Catalog Operator. I think of the package as something that exists on the other side of the CatalogSource and not changed by OLM as a whole. If I want to change any part of the package, I replace the catalog, not a component of OLM.

dinhxuanvu: The package in this context is operator bundle AKA manifests yaml that make up the operator such as CSV, CRD and etc. Those manifests are pulled down inside the registry pod that is running.

flickerfly:
Thanks to @ecordell for helping me understand what is going on better [over in slack](https://kubernetes.slack.com/archives/CAW0GV7A5/p1584117612231400). I'm going to give this another rewrite next week.

flickerfly:
The failure isn't actually caused by this commit, best I can tell. It's just documentation changes. Am I wrong?

dinhxuanvu:
@flickerfly There are some issues with e2e-gcp test so please ignore the failures.

flickerfly:
Are we good to merge then or are there some other issues?

ecordell:
/lgtm

Thanks @flickerfly! This should merge automatically soon.

PR #1352, opened 2020-03-04
[wip] Moving packages to api

Bowenislandsong:
This commit is corresponding to the olm part of https://github.com/operator-framework/api/pull/21, moving olm/pkg/lib/version -> api/pkg/lib/version and olm/pkg/api/apis/operators -> api/pkg/operators. It also update openshift/api version from the invalide 3.9.


go.mod
```diff
@@ -42,8 +42,8 @@ require (
 
 replace (
 	github.com/docker/docker => github.com/moby/moby v0.7.3-0.20190826074503-38ab9da00309 // Required by Helm
-	github.com/openshift/api => github.com/openshift/api v3.9.1-0.20190924102528-32369d4db2ad+incompatible
 	github.com/openshift/client-go => github.com/openshift/client-go v0.0.0-20190923180330-3b6373338c9b
+	github.com/operator-framework/api => ../api
```
Bowenislandsong: This will be updated once https://github.com/operator-framework/api/pull/21 is merged.

Bowenislandsong:
OLM is still dependent on registry for the `client`, `server`, `configmap` and `lib/bundle`.

njhale:
Moving the API types is probably going to cause trouble as we move towards kubebuilder for new controllers. IIRC the norm is that they exist in the same repo.

ecordell:
Since this is out of date and we need a bit more discussion, let's close for now. We can make a new PR when we have the plan worked out.

PR #1353, opened 2020-03-05
Bug 1810698: feat(jitter): add jitter to controllers to smooth out our spiky resource

ecordell:
usage.

**Reviewer Checklist**


ecordell:
@kevinrizza I have a cluster spinning up that I plan to check - I don't think this completely solves the problem but I expect to have more, smaller spikes.

PR #1362, opened 2020-03-10
chore(deps): bump kube to 1.17

anik120:
Updates kube dependencies to 1.17

**Reviewer Checklist**


go.mod
```diff
@@ -35,8 +35,8 @@ require (
 	k8s.io/component-base v0.0.0
 	k8s.io/klog v1.0.0
 	k8s.io/kube-aggregator v0.0.0
-	k8s.io/kube-openapi v0.0.0-20190918143330-0270cf2f1c1d
-	k8s.io/kubernetes v1.16.0
```
LouisPlisso: can you please move all k8s modules to native versions: no more `v0.0.0`

this would be in line with https://github.com/operator-framework/operator-registry/issues/201 and https://github.com/operator-framework/api/pull/17

Also please note that k8s/kubernetes should not be a requirement and is used for locking the versions of the other libraries

Can you please try to remove the import of `k8s.io/kubernetes` and move all other imports as `k8s.io/api v0.17.3` to have them as k8s 1.17.3

Could you please remove the k8s repos from the replace directive down in the go.mod, otherwise any change in the main will be overriden by the replace

anik120: @LouisPlisso thank you for your review. It was a WIP PR, apologies for not editing the title with WIP. PTAL now. 
Unfortunately though, we still have to keep the `replace` directive, since `openshift/oc` [requires](https://github.com/openshift/oc/blob/master/go.mod#L62) `k8s.io/kubernetes`, which in turn [requires `v0.0.0` of all `k8s.io` modules](https://github.com/kubernetes/kubernetes/blob/master/go.mod#L156-L164). 

Unless we go and update `openshift/oc` to not require `k8s.io/kubernetes` first and that way we could probably get rid of the k8s repos from `replace` directive altogether. Thoughts @operator-framework/team-olm ?

anik120: I've attempted to remove dependency on `k8s.io/kubernetes` and import all other `k8s.io/kubernetes` as `v0.17.3` by manually vendoring in `k8s.io/kubernetes` in #1370. I'll close this PR in favour of #1370 if that idea has sufficient votes.

njhale:
@anik120 has this been superseded by #1370?

anik120:
@njhale  yes. 

Closing in favor of #1370

PR #1369, opened 2020-03-13
[wip] adding metric cardinality tests

Bowenislandsong:
This commits adds a script to run metric cardinality tests from origin/test/extended/Prometheus test package. The test package will remain effective with operators uninstalled from the cluster because alerts and messages fired to Prometheus on the cluster are not erased.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -117,7 +117,7 @@ setup-bare: clean e2e.namespace
 
 # e2e test exculding the rh-operators directory which tests rh-operators and their metric cardinality.
 e2e:
-	./test/rh-operators/e2e-origin-prometheus.sh
+	. ./test/rh-operators/e2e-origin-prometheus.sh
 
 e2e-local: build-linux build-wait build-util-linux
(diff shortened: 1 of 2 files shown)


benluddy:
Can this live in the job config instead of in the OLM repo?

Bowenislandsong:
> Can this live in the job config instead of in the OLM repo?

thought it's cleaner here .... I know some projects have it in the config but i think its messy

benluddy:
I'll let others share their opinions, but my preference is to put this in `openshift/release` because it is exclusively targeting OpenShift.

Bowenislandsong:
> I'll let others share their opinions, but my preference is to put this in `openshift/release` because it is exclusively targeting OpenShift.

Ahh, if you put it that way, then you are right, it makes more sense. I'll move it after i'm done with PR testing.

Bowenislandsong:
superseded by https://github.com/openshift/release/pull/7731
/close

PR #1370, opened 2020-03-13
chore(deps): bump to kube 1.17.3

anik120:

* Updates kube dependencies to 1.17.3.
* Removes dependency on k8s.io/kubernetes my manually
vendoring 1.17 version of it inside pkg/lib/ as a stop gap
measure till other alternatives to rbac authorizer and any
other parts of k8s.io/kubernetes used in olm is found.
* Regenerate API types in accordance with openapi3 
specification

**Reviewer Checklist**


anik120:
/hold-cancel
@Bowenislandsong please refer https://github.com/operator-framework/operator-registry/issues/201
Specifically, 

> Since k8s repos now offer a "native" way to pin versions k8s 1.16.7 maps to a go.mod version of v0.16.7, we can achieve the pinning to a particular kubernetes version without using the replace directive.

>

go.mod
```diff
@@ -24,52 +25,29 @@ require (
...
+	k8s.io/kube-openapi v0.0.0-20191107075043-30be4d16710a
 	sigs.k8s.io/controller-tools v0.2.4
 )
 
+// NOTE: k8s.io/Kubernetes v1.17 is currently vendored by hand in pkg/lib and should be
+// bumped along with other kube modules unitll dependency on k8s.io/Kubernetes is removed
+
 replace (
 	github.com/docker/docker => github.com/moby/moby v0.7.3-0.20190826074503-38ab9da00309 // Required by Helm
 	github.com/openshift/api => github.com/openshift/api v3.9.1-0.20190924102528-32369d4db2ad+incompatible
```
ecordell: Can this dependency be fixed as well?

go.mod
```diff
@@ -24,52 +25,29 @@ require (
...
-	k8s.io/kube-scheduler => k8s.io/kube-scheduler v0.0.0-20190918162820-3b5c1246eb18
-	k8s.io/kubectl => k8s.io/kubectl v0.0.0-20190918164019-21692a0861df
-	k8s.io/kubelet => k8s.io/kubelet v0.0.0-20190918162654-250a1838aa2c
-	k8s.io/legacy-cloud-providers => k8s.io/legacy-cloud-providers v0.0.0-20190918163543-cfa506e53441
-	k8s.io/metrics => k8s.io/metrics v0.0.0-20190918162108-227c654b2546
-	k8s.io/node-api => k8s.io/node-api v0.0.0-20190918163711-2299658ad911
-	k8s.io/sample-apiserver => k8s.io/sample-apiserver v0.0.0-20190918161442-d4c9c65c82af
-	k8s.io/sample-cli-plugin => k8s.io/sample-cli-plugin v0.0.0-20190918162410-e45c26d066f2
-	k8s.io/sample-controller => k8s.io/sample-controller v0.0.0-20190918161628-92eb3cb7496c
+	sigs.k8s.io/structured-merge-diff => sigs.k8s.io/structured-merge-diff v1.0.1-0.20191108220359-b1b620dd3f06
```
ecordell: this doesn't work if we just `require 1.0.1`?

go.mod
```diff
@@ -14,8 +15,8 @@ require (
 	github.com/mitchellh/hashstructure v1.0.0
 	github.com/openshift/api v3.9.1-0.20190924102528-32369d4db2ad+incompatible
```
Bowenislandsong: you might wan to fix this!

anik120: @Bowenislandsong thanks for noticing it! I've changed it to the tag being used by `openshift/oc` currently.

go.mod
anik120: @ecordell https://github.com/operator-framework/operator-lifecycle-manager/pull/1370#discussion_r393301639

go.mod
anik120: @ecordell unfortunately, this is the version that `k8s.io/apiserver` needs, but there is another module that is changing the version of `sigs.k8s.io/structured-merge-diff` to a version that is breaking for v0.17.3 of `k8s.io/apiserver`. So had to pin this version to appease all modules.

anik120:
Updated PR to include a commit that updates API types in accordance with updated openapi specifications. 

References:
[What are the annotations for in the commit? ](https://github.com/kubernetes-sigs/kubebuilder/issues/1081)
[How can I verify if the annotations are correct?](https://github.com/kubernetes/kubernetes/pull/77354/files#diff-ca188b89d2fc7104865fe95be2114e81) 
[Why are the variable names being changed in the commit? ](https://github.com/kubernetes/kube-openapi/blob/master/pkg/generators/rules/names_match.go#L54-L76)
[Why is the verification still failing with `API rule violation` errors?](https://github.com/kubernetes/kubernetes/blob/36981002246682ed7dc4de54ccc2a96c1a0cbbdb/api/api-rules/violation_exceptions.list)

LouisPlisso:
Thanks for doing this: hopefully soon the replace usage will be rare and we'll be able to import without problems :)

go.mod
```diff
@@ -49,30 +50,5 @@ replace (
 	github.com/openshift/client-go => github.com/openshift/client-go v0.0.0-20190923180330-3b6373338c9b
 	github.com/prometheus/client_golang => github.com/prometheus/client_golang v0.9.3-0.20190127221311-3c4408c8b829
 	google.golang.org/grpc => google.golang.org/grpc v1.26.0 // https://github.com/etcd-io/etcd/issues/11563
-
-	// Pin to kube 1.16
```
awgreene: Could you add this comment back and keep the pinned requirements separate?

pkg/api/apis/operators/v1alpha1/clusterserviceversion_types.go
```diff
@@ -77,62 +78,73 @@ func (d *StrategyDetailsDeployment) GetStrategyName() string {
...
+	DisplayName string `json:"displayName,omitempty"`
+	Description string `json:"description,omitempty"`
+	// +listType=set
+	Resources []APIResourceReference `json:"resources,omitempty"`
+	// +listType=set
+	StatusDescriptors []StatusDescriptor `json:"statusDescriptors,omitempty"`
+	// +listType=set
+	SpecDescriptors []SpecDescriptor `json:"specDescriptors,omitempty"`
+	// +listType=set
+	ActionDescriptors []ActionDescriptor `json:"actionDescriptors,omitempty"`
```
awgreene: Although this is correct grammatically, this is a breaking API Change.

The change to the json tag will result in an updated CRD that does not include the `ActionDescriptor` field, in turn breaking any CRs on cluster that use the field.

pkg/api/apis/operators/v1alpha1/clusterserviceversion_types.go
```diff
@@ -77,62 +78,73 @@ func (d *StrategyDetailsDeployment) GetStrategyName() string {
...
+	DisplayName    string `json:"displayName,omitempty"`
+	Description    string `json:"description,omitempty"`
+	// +listType=set
+	Resources []APIResourceReference `json:"resources,omitempty"`
+	// +listType=set
+	StatusDescriptors []StatusDescriptor `json:"statusDescriptors,omitempty"`
+	// +listType=set
+	SpecDescriptors []SpecDescriptor `json:"specDescriptors,omitempty"`
+	// +listType=set
+	ActionDescriptors []ActionDescriptor `json:"actionDescriptors,omitempty"`
```
awgreene: The breaking API Change comment applies here as well.

go.mod
anik120: @awgreene we're getting rid of pinned kube requirements. Please refer to [this](https://github.com/operator-framework/operator-lifecycle-manager/pull/1370#issuecomment-599712543) comment.

pkg/api/apis/operators/v1alpha1/clusterserviceversion_types.go
anik120: @awgreene I just changed the internal struct and not the json tag. Please see the changes to the CRDs in this [commit](https://github.com/operator-framework/operator-lifecycle-manager/pull/1370/commits/57b580e57e88f197617c7557d974057f6d9695a5). None of the fields are changed, only comments are updated. Note that `actionDescriptors` in the CRD remains `actionDescriptors`.

go.mod
awgreene: I should have double checked the previous comment, my bad.

pkg/api/apis/operators/v1alpha1/clusterserviceversion_types.go
awgreene: Ah - I see - the json tag was always `actionDescriptors`. Got it.

ecordell:
/approve

my one nit can be addressed in a follow up, and afaict Alex's comment is already addressed. Thanks for working through the details on this one!

Issue #1347, opened 2020-03-03
Operator upgrades using the same channel failing


huizengaJoe:
I have a ocp 4.4 nightly from 3/2/2020, I am trying to test out using olm to automatically update my operator to a new z level ie going from 0.0.3 to 0.0.4 to 0.0.5.  In my example I have not changed the CRD between versions, i just created a new CSV that now `replaces` my old version.  I am using operator-sdk 0.15.2 with opm.  I built my 3 operator bundles

https://quay.io/repository/huizenga/app-operator-bundle:0.0.3,0.0.4 and 0.0.5

I used opm to built 3 separate index images
```
opm index add -c docker --bundles quay.io/huizenga/app-operator-bundle:0.0.3  --tag quay.io/huizenga/app-operator-index:0.0.3
```
```
opm index add -c docker --bundles quay.io/huizenga/app-operator-bundle:0.0.3,quay.io/huizenga/app-operator-bundle:0.0.4  --tag quay.io/huizenga/app-operator-index:0.0.4
```

```
opm index add -c docker --bundles quay.io/huizenga/app-operator-bundle:0.0.3,quay.io/huizenga/app-operator-bundle:0.0.4,quay.io/huizenga/app-operator-bundle:0.0.5  --tag quay.io/huizenga/app-operator-index:0.0.5
```
For my last attempt, I created my catalogsource pointing to  quay.io/huizenga/app-operator-index:0.0.4
I then created a subscription which deployed eveything as I would have expected (ie version 0.0.4 of my app-operator got deployed)

I then updated my catalogsource to point to quay.io/huizenga/app-operator-index:0.0.5, the new registry pod started up, olm detected that there was a new version of my operator (0.0.5) in the alpha channel and proceeded to try to upgrade my operator.  The update fails, wonder what I am doing wrong.

Here is the installplan output for the update:
```
apiVersion: operators.coreos.com/v1alpha1
kind: InstallPlan
metadata:
  generateName: install-
  selfLink: >-
    /apis/operators.coreos.com/v1alpha1/namespaces/joeolm/installplans/install-szj4r
  resourceVersion: '173620'
  name: install-szj4r
  uid: e24a9a21-a2da-474c-aa8d-753021b4138c
  creationTimestamp: '2020-03-03T21:47:04Z'
  generation: 1
  namespace: joeolm
  ownerReferences:
    - apiVersion: operators.coreos.com/v1alpha1
      blockOwnerDeletion: false
[...]

PR #1386, opened 2020-03-18
Adds csv names to the spec of installplan after unpacking bundles.

fantapsody:

Adds the CSV name to `InstallPlanSpec.ClusterServiceVersionNames` after unpacking a bundle.

Addressing https://github.com/operator-framework/operator-lifecycle-manager/issues/1347, especially the case I mentioned in the comment, that packages from a CatalogSource defined by an index image cannot be upgraded and reports error like `pre-existing CRD owners found for owned CRD(s) of dependent CSV`. The `ClusterServiceVersionNames` field in the Spec of InstallPlan is empty, although steps of kind `CustomResourceDefinition` exists in the Status.

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -1100,6 +1100,18 @@ func (o *Operator) unpackBundles(plan *v1alpha1.InstallPlan) (bool, *v1alpha1.In
 		// Add steps and remove resolved bundle lookup
 		out.Status.Plan = append(out.Status.Plan, steps...)
 		out.Status.BundleLookups = append(out.Status.BundleLookups[:i], out.Status.BundleLookups[i+1:]...)
+		// add csv name to the list
+		found := false
+		// the list should be short enough to iterate over
+		for _, name := range out.Spec.ClusterServiceVersionNames {
+			if name == res.Bundle().CsvName {
+				found = true
+				break
+			}
+		}
+		if !found {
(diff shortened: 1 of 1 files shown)


pkg/controller/operators/catalog/operator.go
```diff
@@ -1100,6 +1100,18 @@ func (o *Operator) unpackBundles(plan *v1alpha1.InstallPlan) (bool, *v1alpha1.In
 		// Add steps and remove resolved bundle lookup
 		out.Status.Plan = append(out.Status.Plan, steps...)
 		out.Status.BundleLookups = append(out.Status.BundleLookups[:i], out.Status.BundleLookups[i+1:]...)
+		// add csv name to the list
```
ecordell: We should know these names before we unpack the bundles - they're returned as a part of the resolved set of operators from the resolver

pkg/controller/operators/catalog/operator.go
```diff
@@ -1140,6 +1152,14 @@ func (o *Operator) syncInstallPlans(obj interface{}) (syncError error) {
 			return
 		}
 
+		if !reflect.DeepEqual(plan.Spec.ClusterServiceVersionNames, out.Spec.ClusterServiceVersionNames) {
```
ecordell: Updating the spec here, followed by updating the status right below, will be a source of errors/flakes (i.e. we have to be sure to handle the case where updating the spec errors before updating the status).

fantapsody:
Thanks for the fix, I will close this PR. @ecordell

PR #1387, opened 2020-03-18
Add more CSV information to the PackageManifest

dinhxuanvu:
Transfer `spec.relatedImages`, `spec.minKubeVersion`, and `spec.nativeAPIs`
information from the CSV onto the packagemanifest response.

CSVs are transformed from the in-bundle representation to the packagemanifest
by selecting fields to display. Those additional fields will support the UI.

**Reviewer Checklist**


pkg/package-server/apis/operators/packagemanifest.go
```diff
@@ -69,3 +72,21 @@ func descriptionsForAPIServices(apis []operatorsv1alpha1.APIServiceDescription)
 	}
 	return descriptions
 }
+
+// GetRelatedImages returns a list of images listed in Deployments
+func GetRelatedImages(csv *operatorsv1alpha1.ClusterServiceVersion) []string {
```
ecordell: You also want to pick up the images explicitly listed in the CSV, see: https://github.com/operator-framework/operator-registry/blob/7418392774193f088a5f9c4be733da3584de1384/pkg/registry/csv.go#L210

test/e2e/packagemanifest_e2e_test.go
```diff
@@ -85,6 +85,9 @@ var _ = Describe("Package Manifest", func() {
 		require.NotNil(GinkgoT(), pm)
 		require.Equal(GinkgoT(), packageName, pm.GetName())
 		require.Equal(GinkgoT(), expectedStatus, pm.Status)
+		require.Equal(GinkgoT(), "0.0.0", pm.Status.Channels[0].CurrentCSVDesc.MinKubeVersion)
+		require.Equal(GinkgoT(), *dummyImage, pm.Status.Channels[0].CurrentCSVDesc.RelatedImages[0])
+		require.Equal(GinkgoT(), csv.Spec.NativeAPIs, pm.Status.Channels[0].CurrentCSVDesc.NativeAPIs)
 		require.Equal(GinkgoT(), "label", pm.GetLabels()["projected"])
```
ecordell: It would be great if we could add a test that makes sure that if a bundle has a csv with `spec.relatedImages` field that those are also included

dinhxuanvu:
@ecordell Ready for review

dinhxuanvu:
I expect the `RelatedImages` info will be a part of registry API response when you query bundle info in the future. Then, we no longer need to parse the raw bundle JSON in the future.

AndrienkoAleksandr:
Sorry, I see  https://github.com/operator-framework/operator-lifecycle-manager/pull/1387/files#diff-f23331df68a5aa78273da61eca6bac31R830 to the CSVDescription was applied RelatedImages. I have a question: **RelatedImages section will be exposed or this information is encapsulated inside registry**? I mean when I have installed operator and I do command: `kubectl get csv some-operator -n somenamespace` Should response  contains field with relatedImages?

dinhxuanvu:
`RelatedImages` is captured in the registry, not in the CSV object. You can get `RelatedImages` if you query `PackageManifest`.

AndrienkoAleksandr:
> You can get RelatedImages if you query PackageManifest

I see. But, I guess it returns RelatedImages only for latest bundle. Am I correct?

dinhxuanvu:
Correct, latest bundle for each channel.

PR #1388, opened 2020-03-18
Add e2e test for multiple APIServices on a pod

awgreene:
This commit introduces a test that ensures that a single pod can support
multiple APIServices.

### DIFF
--- a/test/e2e/csv_e2e_test.go
+++ b/test/e2e/csv_e2e_test.go
@@ -3521,6 +3521,129 @@ func deleteLegacyAPIResources(t *testing.T, desc v1alpha1.APIServiceDescription)
 	require.NoError(t, err)
 }
 
+func TestMultipleAPIServicesOnSinglePod(t *testing.T) {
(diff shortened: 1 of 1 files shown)


test/e2e/csv_e2e_test.go
```diff
@@ -3521,6 +3521,106 @@ func deleteLegacyAPIResources(t *testing.T, desc v1alpha1.APIServiceDescription)
...
+	}
+	csv.SetName("csv-hat-1")
+	csv.SetNamespace(testNamespace)
+
+	// Create the APIService CSV
+	cleanupCSV, err := createCSV(t, c, crc, csv, testNamespace, false, false)
+	require.NoError(t, err)
+	defer cleanupCSV()
+
+	_, err = fetchCSV(t, crc, csv.Name, testNamespace, csvSucceededChecker)
```
njhale: I think we also want to check that `caBundle` is the same for both `APIServices`.

awgreene: Great idea - added in the latest version of the PR.

PR #1398, opened 2020-03-19
Add OG label to namespaces in OperatorGroup

awgreene:
This commit introduces a change so OLM will append a label that can be
used to select all namespaces that are associated with a specific
OperatorGroup. The label appended to namespaces is equivalent to:
"olm.operatorgroup/<OPERATORGROUP_NAMESPACE>.<OPERATORGROUP_NAME>"

### DIFF
--- a/pkg/controller/operators/olm/operatorgroup.go
+++ b/pkg/controller/operators/olm/operatorgroup.go
@@ -27,6 +27,8 @@ const (
 	AdminSuffix = "admin"
 	EditSuffix  = "edit"
 	ViewSuffix  = "view"
+
+	operatorGroupLabelTemplate = "olm.operatorgroup/%s.%s"
 )
 
 var (
@@ -98,6 +100,10 @@ func (a *Operator) syncOperatorGroups(obj interface{}) error {
 	logger.WithField("targetNamespaces", targetNamespaces).Debug("updated target namespaces")
 
 	if namespacesChanged(targetNamespaces, op.Status.Namespaces) {
+		// Labels are only applied to namespaces that are not empty strings.
+		prunedStatusNamespaces := pruneEmptyStrings(op.Status.Namespaces)
+		prunedTargetNamespaces := pruneEmptyStrings(targetNamespaces)
+
 		// Update operatorgroup target namespace selection
 		logger.WithField("targets", targetNamespaces).Debug("namespace change detected")
 		op.Status = v1.OperatorGroupStatus{
@@ -108,6 +114,25 @@ func (a *Operator) syncOperatorGroups(obj interface{}) error {
 		if _, err = a.client.OperatorsV1().OperatorGroups(op.GetNamespace()).UpdateStatus(op); err != nil && !k8serrors.IsNotFound(err) {
 			logger.WithError(err).Warn("operatorgroup update failed")
 			return err
+		}
+
+		// Remove labels from targetNamespaces that were dropped.
+		ogLabel := getOperatorGroupLabel(*op)
+		for _, namespaceName := range setDifference(prunedStatusNamespaces, prunedTargetNamespaces) {
+			err = a.removeNamespaceLabel(namespaceName, ogLabel)
+			if err != nil {
+				logger.WithError(err).Warnf("failed to remove operatorgroup label from namespace %s", namespaceName)
+				return err
+			}
+		}
+
+		// Add labels to targetNamespaces that were added.
+		for _, namespaceName := range setDifference(prunedTargetNamespaces, prunedStatusNamespaces) {
+			err = a.addNamespaceLabel(namespaceName, ogLabel, "")
+			if err != nil {
+				logger.WithError(err).Warnf("failed to add operatorgroup to from namespace %s", namespaceName)
+				return err
+			}
 		}
 		logger.Debug("namespace change detected and operatorgroup status updated")
 		// CSV requeue is handled by the succeeding sync in `annotateCSVs`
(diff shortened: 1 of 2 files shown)


pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -98,6 +100,28 @@ func (a *Operator) syncOperatorGroups(obj interface{}) error {
 	logger.WithField("targetNamespaces", targetNamespaces).Debug("updated target namespaces")
 
 	if namespacesChanged(targetNamespaces, op.Status.Namespaces) {
+		ogLabel := getOperatorGroupLabel(*op)
```
awgreene: Open Question: Should we update the labels after or before updating the operatorgroup?

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -160,6 +184,37 @@ func (a *Operator) syncOperatorGroups(obj interface{}) error {
 	return nil
 }
 
+func unique(logger *logrus.Entry, a, b []string) []string {
```
awgreene: Open Question: Anyone have a better name?

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -98,6 +100,28 @@ func (a *Operator) syncOperatorGroups(obj interface{}) error {
...
+		for _, namespaceName := range unique(logger, op.Status.Namespaces, targetNamespaces) {
+			err = a.removeNamespaceLabel(namespaceName, ogLabel)
+			if err != nil {
+				logger.WithError(err).Warnf("failed to remove operatorgroup label from namespace %s", namespaceName)
+				return err
+			}
+		}
+
+		// Add labels to targetNamespaces that were added.
+		unique := unique(logger, targetNamespaces, op.Status.Namespaces)
```
anik120: super nit: could we use a better name for `unique` that won't require the reader to scroll to and read the func?

pkg/controller/operators/olm/operatorgroup.go
benluddy: [Set difference](https://en.wikipedia.org/wiki/Complement_(set_theory))?

awgreene: I support this setDifference.

pkg/controller/operators/olm/operatorgroup.go
awgreene: Going with Ben's suggestion: https://github.com/operator-framework/operator-lifecycle-manager/pull/1398#discussion_r395353177

pkg/controller/operators/olm/operatorgroup.go
dinhxuanvu: I would say after OG is updated.

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -160,6 +184,37 @@ func (a *Operator) syncOperatorGroups(obj interface{}) error {
...
 }
 
+func unique(logger *logrus.Entry, a, b []string) []string {
+	presentInANotInB := []string{}
+	// Check if either list is is empty to save time.
+	// If the list is of length 1, we much check that the namespace is not an
+	// empty string because the status object defaults to the following:
+	// status:
+	//   namespaces:
+	//   - ""
```
dinhxuanvu: If the namespaces list is set `""`, it means `AllNamespaces`. So does that mean you need to add the OG label to all namespaces that OLM is watching?

pkg/controller/operators/olm/operatorgroup.go
dinhxuanvu: Would be nice if we can have a few lines of comments above to describe what this func does.

pkg/controller/operators/olm/operatorgroup.go
awgreene: In this case I would not apply a label.

pkg/controller/operators/olm/operatorgroup.go
awgreene: Addressed in the latest version of the PR.

pkg/controller/operators/olm/operatorgroup.go
awgreene: Resolved.

test/e2e/operator_groups_e2e_test.go
```diff
@@ -2115,3 +2118,143 @@ func createProjectAdmin(t GinkgoTInterface, c operatorclient.ClientInterface, na
 		_ = c.DeleteRoleBinding(rb.GetNamespace(), rb.GetName(), metav1.NewDeleteOptions(0))
 	}
 }
+
+func TestOperatorGroupLabels(t *testing.T) {
```
benluddy: Could you turn this into a Ginkgo spec?

awgreene: Done!

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -98,6 +100,10 @@ func (a *Operator) syncOperatorGroups(obj interface{}) error {
 	logger.WithField("targetNamespaces", targetNamespaces).Debug("updated target namespaces")
 
 	if namespacesChanged(targetNamespaces, op.Status.Namespaces) {
+		// Labels are only applied to namespaces that are not empty strings.
+		prunedStatusNamespaces := pruneEmptyStrings(op.Status.Namespaces)
```
harishsurf: nit: since these strings are `namespaces`, `pruneEmptyNamespaces` is more contextual/ meaningful?

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -160,6 +185,43 @@ func (a *Operator) syncOperatorGroups(obj interface{}) error {
...
+
+	// Find the Set Difference.
+	m := make(map[string]bool)
+
+	for _, item := range b {
+		m[item] = true
+	}
+
+	for _, item := range a {
+		if _, ok := m[item]; !ok {
```
harishsurf: nit: `itemPresent`  or `itemExist` instead of `ok` ?

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -160,6 +185,43 @@ func (a *Operator) syncOperatorGroups(obj interface{}) error {
...
+			diff = append(diff, item)
+		}
+	}
+	return diff
+}
+
 func (a *Operator) operatorGroupDeleted(obj interface{}) {
```
harishsurf: nit: not part of your change, but `a` could have a better name?
Edit: I see `a` being used as a variable name for `operator` across multiple helper functions. If that's the norm followed, then you could keep it as is.

test/e2e/operator_groups_e2e_test.go
```diff
@@ -2043,6 +2045,169 @@ var _ = Describe("Operator Group", func() {
...
+		}
+
+		// Create the namespaces
+		for _, namespace := range testNamespaces {
+			_, err := c.KubernetesInterface().CoreV1().Namespaces().Create(&corev1.Namespace{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: namespace,
+				},
+			})
+			require.NoError(GinkgoT(), err)
```
harishsurf: FYI: The basic ginkgo conversion just added a wrapper of `Describe/ It` container around existing tests. The plan is to have a full-fledge BDD ginkgo tests which includes - writing assertions using [gomega](https://onsi.github.io/gomega/) assertions. I'll probably have a document on how to do this conversion so that it is incorporated by the team as they write more e2e tests

PR #1402, opened 2020-03-21
Apply OG Label to namespaces before updating OG

awgreene:
This commit fixes a potential bug where an OperatorGroup Status is
updated, but the OG Label fails to be added or removed from affected
namespaces.

Currently, OLM updates the list of namespaces in an OperatorGroup's
status prior to adding/removing the OG from namespaces in that
OperatorGroup. This can lead to a bug where the OperatorGroup is
updated, but the label is not applied correctly. When the OperatorGroup
is resynced because of a label update failure, OLM does not notice any
changes to the list of namespaces in the OperatorGroup and will not
update labels correctly.

### DIFF
--- a/pkg/controller/operators/olm/operatorgroup.go
+++ b/pkg/controller/operators/olm/operatorgroup.go
@@ -100,9 +100,33 @@ func (a *Operator) syncOperatorGroups(obj interface{}) error {
 	logger.WithField("targetNamespaces", targetNamespaces).Debug("updated target namespaces")
 
 	if namespacesChanged(targetNamespaces, op.Status.Namespaces) {
+		logger.Debug("OperatorGroup namespaces change detected")
 		// Labels are only applied to namespaces that are not empty strings.
 		prunedStatusNamespaces := pruneEmptyStrings(op.Status.Namespaces)
 		prunedTargetNamespaces := pruneEmptyStrings(targetNamespaces)
+
+		// Remove labels from targetNamespaces that were dropped.
+		logger.Debug("Removing OperatorGroup Labels from dropped namespaces")
+		ogLabel := getOperatorGroupLabel(*op)
+		for _, namespaceName := range setDifference(prunedStatusNamespaces, prunedTargetNamespaces) {
+			err = a.removeNamespaceLabel(namespaceName, ogLabel)
+			if err != nil {
+				logger.WithError(err).Warnf("failed to remove operatorgroup label from namespace %s", namespaceName)
+				return err
+			}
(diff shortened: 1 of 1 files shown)


awgreene:
@dinhxuanvu given our discussion on [PR 1398](https://github.com/operator-framework/operator-lifecycle-manager/pull/1398#discussion_r395657525) you may be interested in weighing in on this PR.

awgreene:
CC @benluddy since you were very helpful on the last review.

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -257,16 +263,20 @@ func (a *Operator) operatorGroupDeleted(obj interface{}) {
...
 
-	namespaceUpdateFunc(namespace)
+	out := in.DeepCopy()
+	namespaceUpdateFunc(out)
 
-	_, err = a.opClient.KubernetesInterface().CoreV1().Namespaces().Update(namespace)
-	if err != nil {
-		return err
+	// Avoid unnecessary updates.
+	if in != out {
```
ecordell: Won't this always be true? `in` and `out` are different pointers.

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -100,41 +100,47 @@ func (a *Operator) syncOperatorGroups(obj interface{}) error {
...
 			if err != nil {
 				logger.WithError(err).Warnf("failed to remove operatorgroup label from namespace %s", namespaceName)
 				return err
 			}
+			logger.Debugf("Removed OperatorGroup label from namespace %s", namespaceName)
 		}
 
 		// Add labels to targetNamespaces that were added.
+		logger.Debug("Adding OperatorGroup Labels to new namespaces")
 		for _, namespaceName := range setDifference(prunedTargetNamespaces, prunedStatusNamespaces) {
```
ecordell: This didn't change in this PR, but there could be a lot of namespaces in a cluster (we scale test with 2000+) which will make syncing an operatorgroup fairly slow (2000*kube api latency)

It would be better to just update the target namespaces when syncing the OperatorGroup, and simply requeue all of the namespaces. 

Syncing a namespace can start by checking if its labels are up to date. If not, we can update them, then exit (requeue again to continue processing the namespace).

awgreene: Will do!

awgreene:
Closing this in favor of https://github.com/operator-framework/operator-lifecycle-manager/pull/1408

PR #1404, opened 2020-03-23
Bug 1815957: increase grpc_health_probe timeout

wangzheng422:
**Description of the change:**
increase grpc_health_probe timeout by adding -rpc-timeout=10s

**Motivation for the change:**
the community operator catalog source pod will continue to restart because the healthcheck failed, due to the timeout is too small.

**Reviewer Checklist**
- [ ] Implementation matches the proposed design, or proposal is updated to match implementation
- [ ] Sufficient unit test coverage 
- [ ] Sufficient end-to-end test coverage
- [ ] Docs updated or added to `/docs` 
- [ ] Commit messages sensible and descriptive

PR #1407, opened 2020-03-23
Bug 1815957: increase grpc_health_probe timeout

wangzheng422:
increase grpc_health_probe timeout by adding -rpc-timeout=10s

the community operator catalog source pod will continue to restart because the healthcheck failed, due to the timeout is too small.

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/registry/reconciler/reconciler.go
+++ b/pkg/controller/registry/reconciler/reconciler.go
@@ -129,6 +129,7 @@ func Pod(source *v1alpha1.CatalogSource, name string, image string, labels map[s
 							},
 						},
 						InitialDelaySeconds: livenessDelay,
+						TimeoutSeconds:      10,
 					},
 					Resources: v1.ResourceRequirements{
 						Requests: v1.ResourceList{
--- a/pkg/controller/registry/reconciler/reconciler.go
+++ b/pkg/controller/registry/reconciler/reconciler.go
@@ -114,7 +114,8 @@ func Pod(source *v1alpha1.CatalogSource, name string, image string, labels map[s
 					ReadinessProbe: &v1.Probe{
 						Handler: v1.Handler{
 							Exec: &v1.ExecAction{
-								Command: []string{"grpc_health_probe", "-addr=localhost:50051"},
+								Command: []string{"grpc_health_probe", "-addr=localhost:50051",
+								"-rpc-timeout=10s"},
 							},
 						},
 						InitialDelaySeconds: readinessDelay,
@@ -123,7 +124,8 @@ func Pod(source *v1alpha1.CatalogSource, name string, image string, labels map[s
 					LivenessProbe: &v1.Probe{
 						Handler: v1.Handler{
 							Exec: &v1.ExecAction{
-								Command: []string{"grpc_health_probe", "-addr=localhost:50051"},
+								Command: []string{"grpc_health_probe", "-addr=localhost:50051",
+								"-rpc-timeout=10s"},
 							},
 						},
 						InitialDelaySeconds: livenessDelay,
--- a/pkg/controller/registry/reconciler/reconciler.go
+++ b/pkg/controller/registry/reconciler/reconciler.go
@@ -119,7 +119,7 @@ func Pod(source *v1alpha1.CatalogSource, name string, image string, labels map[s
 							},
 						},
 						InitialDelaySeconds: readinessDelay,
-						TimeoutSeconds:      5,
+						TimeoutSeconds:      10,
 					},
 					LivenessProbe: &v1.Probe{
 						Handler: v1.Handler{


wangzheng422:
link from: https://github.com/operator-framework/operator-lifecycle-manager/pull/1404

pkg/controller/registry/reconciler/reconciler.go
```diff
@@ -114,7 +114,8 @@ func Pod(source *v1alpha1.CatalogSource, name string, image string, labels map[s
 					ReadinessProbe: &v1.Probe{
 						Handler: v1.Handler{
 							Exec: &v1.ExecAction{
-								Command: []string{"grpc_health_probe", "-addr=localhost:50051"},
+								Command: []string{"grpc_health_probe", "-addr=localhost:50051",
+								"-rpc-timeout=10s"},
```
njhale: `-rpc-timeout` should probably match the `TimeoutSeconds` field.

pkg/controller/registry/reconciler/reconciler.go
```diff
@@ -123,7 +124,8 @@ func Pod(source *v1alpha1.CatalogSource, name string, image string, labels map[s
 					LivenessProbe: &v1.Probe{
 						Handler: v1.Handler{
 							Exec: &v1.ExecAction{
-								Command: []string{"grpc_health_probe", "-addr=localhost:50051"},
+								Command: []string{"grpc_health_probe", "-addr=localhost:50051",
```
njhale: The default value for liveness probe timeout is [1 second](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes), so we probably want to set the `TimeoutSeconds` field to 10 seconds as well.

wangzheng422: yes, I agree.

njhale:
@wangzheng422 

Robert Heinzmann ran into the same issue: https://bugzilla.redhat.com/show_bug.cgi?id=1815957

He had some really nice suggestions for how to decrease the amount of time the health probe requires: https://access.redhat.com/solutions/4927461

The one that makes the most sense to me is to swap "localhost" for 127.0.0.1, to avoid DNS resolution altogether.

What do you think?

wangzheng422:
> @wangzheng422
> 
> Robert Heinzmann ran into the same issue: https://bugzilla.redhat.com/show_bug.cgi?id=1815957
> 
> He had some really nice suggestions for how to decrease the amount of time the health probe requires: https://access.redhat.com/solutions/4927461
> 
> The one that makes the most sense to me is to swap "localhost" for 127.0.0.1, to avoid DNS resolution altogether.
> 
> What do you think?

I don't this localhost -> 127.0.0.1 is the root cause in my ocp 4.3 cluster. Because I run the command 'grpc_health_probe -addr=localhost:50051' in operator-catalog pod directly, sometime is will return within 1s, but most of time, it will timeout. The error message is:
```
timeout: health rpc did not complete within 1s
command terminated with exit code 3
```
So, the health check command itself needs rpc timeout parameter, because the default value for rpc timeout is 1s.

njhale:
> I don't this localhost -> 127.0.0.1 is the root cause in my ocp 4.3 cluster. Because I run the command 'grpc_health_probe -addr=localhost:50051' in operator-catalog pod directly, sometime is will return within 1s, but most of time, it will timeout. 

@wangzheng422 I think you can test this by swapping `localhost` with `127.0.0.1` when you're running the command in the catalog pod directly.

That being said, I agree a longer timeout will help too.

ecordell:
Thank you for working on this @wangzheng422 

I'm concerned that this is even necessary in the first place. The default timeout is 1s - so we're saying it takes >1s to resolve `localhost` within a single container (consistently, since the health probe is tried multiple times). I'd like to look for other options before we increase the timeout here.

ecordell:
It looks like we may be able to fix this with a line in the `operator-registry` Dockerfile:

https://github.com/operator-framework/operator-registry/pull/227

I would prefer to see if this will fix the issue before we increase the timeout here.

wangzheng422:
> It looks like we may be able to fix this with a line in the `operator-registry` Dockerfile:
> 
> [operator-framework/operator-registry#227](https://github.com/operator-framework/operator-registry/pull/227)
> 
> I would prefer to see if this will fix the issue before we increase the timeout here.

yes, please test it again. In my ocp 4.3 cluster, I tried and get this result:
```
[root@helper ocp4]# oc exec community-operator-catalog-7q69g -- grpc_health_probe -addr=127.0.0.1:50051
status: SERVING
[root@helper ocp4]# oc exec community-operator-catalog-7q69g -- grpc_health_probe -addr=127.0.0.1:50051
timeout: health rpc did not complete within 1s
command terminated with exit code 3
```

ecordell:
@wangzheng422 This should be fixed with the latest operator-registry images. Please let us know if it is not working for you!

PR #1408, opened 2020-03-23
Reconcile Namespace OG Labels in Namespace Syncer

awgreene:
This commit introduces a change so OLM will not attempt to update the
labels on all namespaces in an OperatorGroup while reconciling an
OperatorGroup.

Instead, OLM will trigger a sync on all namespaces in the OperatorGroup.
The sync will cause OLM to update the labels on each namespace in the
OperatorGroup separately.

This PR was born out of [this conversation](https://github.com/operator-framework/operator-lifecycle-manager/pull/1402#discussion_r396490845) .

### DIFF
--- a/pkg/controller/operators/olm/operator.go
+++ b/pkg/controller/operators/olm/operator.go
@@ -68,6 +68,7 @@ type Operator struct {
 	csvCopyQueueSet       *queueinformer.ResourceQueueSet
 	csvGCQueueSet         *queueinformer.ResourceQueueSet
 	objGCQueueSet         *queueinformer.ResourceQueueSet
+	nsQueueSet            workqueue.RateLimitingInterface
 	apiServiceQueue       workqueue.RateLimitingInterface
 	csvIndexers           map[string]cache.Indexer
 	recorder              record.EventRecorder
@@ -394,6 +395,7 @@ func newOperatorWithConfig(ctx context.Context, config *operatorConfig) (*Operat
 	// register namespace queueinformer
 	namespaceInformer := k8sInformerFactory.Core().V1().Namespaces()
 	op.lister.CoreV1().RegisterNamespaceLister(namespaceInformer.Lister())
+	op.nsQueueSet = workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "resolver")
 	namespaceInformer.Informer().AddEventHandler(
 		&cache.ResourceEventHandlerFuncs{
 			DeleteFunc: op.namespaceAddedOrRemoved,
@@ -403,8 +405,9 @@ func newOperatorWithConfig(ctx context.Context, config *operatorConfig) (*Operat
 	namespaceQueueInformer, err := queueinformer.NewQueueInformer(
 		ctx,
 		queueinformer.WithLogger(op.logger),
+		queueinformer.WithQueue(op.nsQueueSet),
 		queueinformer.WithInformer(namespaceInformer.Informer()),
-		queueinformer.WithSyncer(queueinformer.LegacySyncHandler(op.syncObject).ToSyncer()),
+		queueinformer.WithSyncer(queueinformer.LegacySyncHandler(op.syncNamespace).ToSyncer()),
 	)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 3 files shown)


dinhxuanvu:
So this PR supersedes the #1402 or this is a separate PR? Is there a reason why we don't combine both of them into 1 PR?

awgreene:
@dinhxuanvu I closed #1402

exdx:
So if I understand correctly - before we were updating all relevant namespaces at once in a batch operation (but this was potentially very costly) and now we are incrementally resyncing the labels on the namespace on a rolling basis. Makes a lot of sense, wondering if there is any change of behavior on the multitenancy side we should be concerned with if things don't sync up correctly

awgreene:
> Makes a lot of sense, wondering if there is any change of behavior on the multitenancy side we should be concerned with if things don't sync up correctly

@exdx what implications should we consider?

awgreene:
@dinhxuanvu I believe that the e2e test implemented in https://github.com/operator-framework/operator-lifecycle-manager/pull/1398 should be enough.

pkg/controller/operators/olm/operator.go
```diff
@@ -1893,3 +1917,66 @@ func (a *Operator) ensureLabels(in *v1alpha1.ClusterServiceVersion, labelSets ..
...
+		if list[i] == item {
+			return true
+		}
+	}
+	return false
+}
+
+// getOperatorGroupLabel returns a label that is applied to Namespaces to signify that the
+// namespace is a part of the OperatorGroup using selectors.
+func getOperatorGroupLabel(o *v1.OperatorGroup) string {
```
njhale: nit: would this be prettier as a method on an OperatorGroup wrapper?

pkg/controller/operators/olm/operator.go
```diff
@@ -396,8 +420,8 @@ func newOperatorWithConfig(ctx context.Context, config *operatorConfig) (*Operat
 	op.lister.CoreV1().RegisterNamespaceLister(namespaceInformer.Lister())
 	namespaceInformer.Informer().AddEventHandler(
 		&cache.ResourceEventHandlerFuncs{
-			DeleteFunc: op.namespaceAddedOrRemoved,
-			AddFunc:    op.namespaceAddedOrRemoved,
+			DeleteFunc: op.namespacesAddedOrRemoved,
```
njhale: Do we need this to happen in parallel to `op.syncNamespace`?

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -1075,3 +985,34 @@ func (a *Operator) findCSVsThatProvideAnyOf(provide resolver.APISet) ([]*v1alpha
 
 	return providers, nil
 }
+
+// namespacesAddedOrRemoved returns the union of:
+// - the set of elements in A but not in B
+// - the set of elements in B but not in A
+func namespacesAddedOrRemoved(a, b []string) []string {
+	check := make(map[string]bool)
```
njhale: `make(map[string]struct{})`

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -1075,3 +985,34 @@ func (a *Operator) findCSVsThatProvideAnyOf(provide resolver.APISet) ([]*v1alpha
...
+	delete(check, "")
+
+	keys := make([]string, len(check))
+	i := 0
+	for key := range check {
+		keys[i] = key
+		i++
+	}
+
+	return keys
```
njhale:
```go
var keys []string
for key := range check {
    keys = append(keys, key)
}

return keys
```

^ retains useful zero-value for `keys`

pkg/controller/operators/olm/operator.go
```diff
@@ -327,6 +333,24 @@ func newOperatorWithConfig(ctx context.Context, config *operatorConfig) (*Operat
 			return nil, err
 		}
 
+		// Namespace sync for resolving subscriptions
+		namespaceInformer := informers.NewSharedInformerFactory(op.opClient.KubernetesInterface(), config.resyncPeriod()).Core().V1().Namespaces()
```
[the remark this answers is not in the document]
awgreene: This is a very good point - I was unaware that there was an existing namespace informer. Merged the logic with the existing sync function in the latest version of the PR.

pkg/controller/operators/olm/operator.go
```diff
@@ -1893,3 +1917,66 @@ func (a *Operator) ensureLabels(in *v1alpha1.ClusterServiceVersion, labelSets ..
...
+			ns.Labels[getOperatorGroupLabel(og)] = ""
+		}
+	}
+
+	// Update the Namespace
+	_, err = o.opClient.KubernetesInterface().CoreV1().Namespaces().Update(ns)
+	return err
+}
+
+func containsString(list []string, item string) bool {
```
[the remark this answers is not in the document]
awgreene: Agreed - I'll work on creating this in a separate PR

pkg/controller/operators/olm/operator.go
awgreene: Updated so the OperatorGroup struct has a GetLabel() func.

pkg/controller/operators/olm/operator.go
awgreene: That's a great idea.

Issue #829, opened 2019-04-29
Add helm chart to the release  


eyusupov:
OLM is already using

ekarlso:
Anyone from the devs care to weigh in on this?

ecordell:
We don't intend to continue to use helm as our templating layer forever, so this has not been a high priority for us.

RichiCoder1:
+1 for this, if only because it makes it super easy to use a tool like Flux to manage the OLM (which then manages everything else :D)

kfox1111:
If folks step up to support this, would the project accept the pr's? It seems like a useful thing to support.

RichiCoder1:
Bump

mkilchhofer:
Bump again. I am also interested in having a helm chart for managing OLM.

timothymathison:
I'm also interested in this capability, so that we can manage the installation with the terraform helm provider.

ikorzh:
This is good idea.
I am also interested in having a helm chart for managing OLM

yelhouti:
interested in this two please

sathieu:
I've published the chart here: [https://gitlab.com/kubitus-project/external-helm-charts](https://gitlab.com/kubitus-project/external-helm-charts/-/blob/main/README.md#operator-lifecycle-manager).

It'll be updated automatically when new olm versions appears.

chrisdewar:
Above gitlab.com link 404's for me... any update on this?

sturman:
@sathieu @chrisdewar Looks like the git repo has been moved to https://gitlab.com/kubitus-project/external-helm-charts/operator-lifecycle-manager.  Helm repository address has not been changed and it still is https://gitlab.com/api/v4/projects/29255838/packages/helm/stable as mentioned in [README](https://gitlab.com/kubitus-project/external-helm-charts/operator-lifecycle-manager#operator-lifecycle-manager). Also you can check available Helm repository packages in [GitLab packages](https://gitlab.com/kubitus-project/external-helm-charts/operator-lifecycle-manager/-/packages)

sathieu:
@chrisdewar Sorry, I've just updated the URL above. As @sturman said, the repo has moved.

sturman:
@sathieu the chart version 0.19.1 is not found in the repository. I can see version 0.19.0 only
```shell
user@think-t460p:~/projects$ helm search repo lifecycle
NAME                          	CHART VERSION	APP VERSION	DESCRIPTION                
[...]

PR #1410, opened 2020-03-24
End-to-end test binary can set up test clusters

benluddy:
This change gives the end-to-end test binary the ability, gated by
build tags, to provision, deprovision, and perform test operation
installation on test clusters.

Without any tags, the default "provisioner" points to an existing
cluster specified by either $KUBECONFIG, $HOME/.kube/config, or
in-cluster service discovery. An alternate Kubernetes-in-Docker (kind)
provisioner is available. The default installer is a no-op, but this
commit also introduces an optional Helm installer, which installs the
OLM chart (now built into the test binary).

Provisioning, installation, and deprovisioning are run as part of the
Ginkgo BeforeSuite/AfterSuite closures.


deploy/chart/Chart.yaml
```diff
@@ -1,3 +1,4 @@
 apiVersion: v2
 description: A Helm chart for Kubernetes
```
awgreene: Is this description verbose enough?

test/e2e/e2e_test.go
```diff
@@ -37,32 +42,35 @@ var (
...
 )
 
-func TestConvert(t *testing.T) {
+func TestEndToEnd(t *testing.T) {
 	RegisterFailHandler(Fail)
-	RunSpecs(t, "Convert Suite")
+	RunSpecs(t, "End-to-end")
 }
 
+var deprovision func() = func() {}
```
awgreene: Is there a reason `provision` is not capitalized?

benluddy: As in "deProvision"? I've only ever seen "deprovision" used a single word stemming from "provision" with the [de-](https://www.dictionary.com/browse/de-) prefix indicating reversal/negation, although "deprovision" doesn't appear in my dictionary.

deploy/chart/Chart.yaml
benluddy: This is the original description from 2017. If we decide to [include the chart as part of the release](https://github.com/operator-framework/operator-lifecycle-manager/issues/829) (as opposed to only using it for templating internally) then it should definitely be improved, but I don't think it needs to be touched as part of this PR.

test/e2e/e2e_test.go
harishsurf: Trying to understanding this line, what does it actually mean in go?

test/e2e/installer/helm.go
```diff
@@ -0,0 +1,121 @@
...
+
+func (a clientAdapter) Namespace() (string, bool, error) {
+	return "default", false, nil
+}
+
+func (a clientAdapter) ConfigAccess() clientcmd.ConfigAccess {
+	return clientcmd.NewDefaultClientConfigLoadingRules()
+}
+
+func Install(ctx *ctx.TestContext) error {
```
harishsurf: nit: function name could be descriptive?

test/e2e/installer/helm.go
```diff
@@ -0,0 +1,121 @@
...
+
+func (a clientAdapter) ToRawKubeConfigLoader() clientcmd.ClientConfig {
+	return clientcmd.ClientConfig(a)
+}
+
+func (a clientAdapter) RawConfig() (clientcmdapi.Config, error) {
+	return clientcmdapi.Config{}, fmt.Errorf("not supported")
+}
+
+func (a clientAdapter) ClientConfig() (*rest.Config, error) {
```
harishsurf: ultra minor nit: function name could be a verb instead of a noun?

test/e2e/installer/helm.go
```diff
@@ -0,0 +1,121 @@
...
+	return a.ToRESTConfig()
+}
+
+func (a clientAdapter) Namespace() (string, bool, error) {
```
harishsurf: nit: same comment as above - func name as a verb

x.mk
```diff
@@ -0,0 +1,36 @@
...
+
+bin/e2e-local.image.tar: export GOOS=linux
+bin/e2e-local.image.tar: export GOARCH=386
+bin/e2e-local.image.tar: e2e.Dockerfile bin/wait bin/cpb $(CMDS)
+	docker build -t quay.io/operator-framework/olm:local -f $< bin
+	docker save -o $@ quay.io/operator-framework/olm:local
+
+.PHONY: e2e-local
+e2e-local: bin/e2e-local.test bin/e2e-local.image.tar
+	$(GINKGO) -p -randomizeAllSpecs $(if $(TEST),-focus "$(TEST)") -v -timeout 70m $< -- -namespace=operators -olmNamespace=operator-lifecycle-manager -dummyImage=bitnami/nginx:latest -kind.images=e2e-local.image.tar
```
harishsurf: +1 for leveraging ginkgo's capabilities via these flags :)

test/e2e/e2e_test.go
ecordell: `deprovision` is a function set to be an empty function here. It gets assigned a different function based on the provisioner you choose.

It's defined here so that it's in scope for both `BeforeSuite` (where it gets defined) and `AfterSuite` (where it gets called).

It's an empty function instead of nil so that we don't have to bother checking nil at call sites.

test/e2e/installer/helm.go
benluddy: This is an adapter to use a RESTConfig as [genericclioptions.RESTClientGetter](https://github.com/kubernetes/cli-runtime/blob/master/pkg/genericclioptions/config_flags.go#L71-L83), which Helm requires for initialization (check out where it's used below). I'll add a comment near the clientAdapter type definition to make this clear.

njhale:
/hold

This may seem like a nit, but can we amend the commit name before we merge this?

njhale:
/hold cancel

... missed the joke

scripts/package_release.sh
```diff
@@ -12,15 +12,15 @@ version=$1
...
 
-charttmpdir=`mktemp -d 2>/dev/null || mktemp -d -t 'charttmpdir'`
+charttmpdir=$(mktemp -d 2>/dev/null || mktemp -d -t charttmpdir)
 
 charttmpdir=${charttmpdir}/chart
 
-cp -R deploy/chart/ ${charttmpdir}
-echo "Version: $1" >> ${charttmpdir}/Chart.yaml
+cp -R deploy/chart/ "${charttmpdir}"
+sed -i "s/^[Vv]ersion:.*\$/version: ${version}/" "${charttmpdir}/Chart.yaml"
```
njhale: We may have compatibility issues for macos users that don't have gsed. Historically, this has plagued us as a team.

scripts/package_release.sh
```diff
@@ -12,15 +12,15 @@ version=$1
...
 
-mkdir -p ${chartdir}
+mkdir -p "${chartdir}"
 
-go run -mod=vendor helm.sh/helm/v3/cmd/helm template -n olm -f ${values} --include-crds --output-dir ${charttmpdir} ${charttmpdir}
+go run -mod=vendor helm.sh/helm/v3/cmd/helm template -n olm -f "${values}" --include-crds --output-dir "${charttmpdir}" "${charttmpdir}"
```
njhale: Nice! I'd like to follow the [google style guide](https://google.github.io/styleguide/shellguide.html) for all our shell scripts if we can.

test/e2e/ctx/ctx.go
```diff
@@ -0,0 +1,97 @@
...
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/clientset/versioned"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/operatorclient"
+	pversioned "github.com/operator-framework/operator-lifecycle-manager/pkg/package-server/client/clientset/versioned"
+)
+
+var ctx TestContext
+
+// TestContext represents the environment of an executing test. It can
+// be considered roughly analogous to a kubeconfig context.
+type TestContext struct {
```
njhale: Most of this (context, provisioner, installer) is useful outside of OLM. I suggest that at some point we remove the OLM-specific imports here, move the context to its own repo or `pkg/lib`, and create an OLM-specific types that embed/decorate them when necessary.

test/e2e/e2e_test.go
```diff
@@ -38,34 +39,37 @@ var (
 	communityOperatorsImage = ""
 )
 
-func TestConvert(t *testing.T) {
+func TestEndToEnd(t *testing.T) {
 	RegisterFailHandler(Fail)
 	SetDefaultEventuallyTimeout(1 * time.Minute)
 	SetDefaultEventuallyPollingInterval(1 * time.Second)
-	RunSpecs(t, "Convert Suite")
+	RunSpecs(t, "End-to-end")
```
njhale: 👍

test/e2e/scoped_client_test.go
```diff
@@ -34,14 +34,11 @@ var _ = Describe("Scoped Client", func() {
...
-		require.NotEmpty(GinkgoT(), kubeConfigPath)
-
-		config, err = clientcmd.BuildConfigFromFlags("", *kubeConfigPath)
-		require.NoError(GinkgoT(), err)
+		config = ctx.Ctx().RESTConfig()
 
 		kubeclient = newKubeClient(GinkgoT())
 		crclient = newCRClient(GinkgoT())
-		dynamicclient = newDynamicClient(GinkgoT(), config)
+		dynamicclient = ctx.Ctx().DynamicClient()
```
njhale: nit: reminder to fix variable casing later

scripts/package_release.sh
benluddy: [Shellcheck](https://github.com/koalaman/shellcheck/) is great, too.

test/e2e/ctx/ctx.go
benluddy: :+1: Agreed! I'm treating this context implementation as a rough draft (just enough to be able to point different test nodes to different clusters without making the test aware of it). Thanks for taking the time to discuss this yesterday.

scripts/package_release.sh
benluddy: Yes, I know I always ended up putting GNU utils at the front of my PATH when using a Mac in the past. In this case, I hope that we can sidestep the problem by automating release generation to run in a container/VM.

PR #1411, opened 2020-03-24
Add tests for GC assumptions.

benluddy:
The ClusterRoles that OLM creates per provided API now have owner
references to their respective CRD or APIService instead of owner
labels to a providing OperatorGroup. These tests verify that the
Kubernetes garbage collector behaves as expected.

### DIFF
--- a/test/e2e/gc_e2e_test.go
+++ b/test/e2e/gc_e2e_test.go
@@ -1,17 +1,115 @@
 package e2e
 
 import (
+	"fmt"
+
 	"github.com/blang/semver"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
 	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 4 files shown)


test/e2e/gc_e2e_test.go
```diff
@@ -1,17 +1,106 @@
...
+			},
+			Spec: apiregistrationv1.APIServiceSpec{
+				Group:                group,
+				Version:              "v1",
+				GroupPriorityMinimum: 1,
+				VersionPriority:      1,
+			},
+		})
+		Expect(err).NotTo(HaveOccurred())
+		defer c.DeleteAPIService(as.GetName(), &metav1.DeleteOptions{})
```
ecordell: is there an `IgnoreError()` wrapper we can use for defers like this?

benluddy: Not exactly, but it sounds worthwhile so I've added `IgnoreError` to a new `dsl` package.

PR #1417, opened 2020-04-01
Wait for successful deployment before proceeding

awgreene:
This commit introduces a change that updates OLM so it waits for a
deployment to reach a ready state prior to proceeding to the next
deployment when installing a CSV.

### DIFF
--- a/pkg/controller/install/deployment.go
+++ b/pkg/controller/install/deployment.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/operator-framework/api/pkg/operators/v1alpha1"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/wrappers"
+	depstatus "github.com/operator-framework/operator-lifecycle-manager/pkg/controller/install/status"
 	hashutil "github.com/operator-framework/operator-lifecycle-manager/pkg/lib/kubernetes/pkg/util/hash"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/ownerutil"
 )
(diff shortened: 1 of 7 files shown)


Bowenislandsong:
sgtm

pkg/controller/install/deployment.go
```diff
@@ -84,7 +85,22 @@ func NewStrategyDeploymentInstaller(strategyClient wrappers.InstallStrategyDeplo
...
+			ready, err := i.strategyClient.IsDeploymentReady(i.owner.GetNamespace(), prev.Name)
+			if err != nil {
+				log.Infof("Unable to retrieve deployment: %v", err)
+				return err
+			}
+
+			if !ready {
+				msg := fmt.Sprintf("Waiting on deployment %s/%s to be ready", prev.Name, i.owner.GetNamespace())
+				log.Infof(msg)
+				return fmt.Errorf(msg)
```
Bowenislandsong, line 101: 1. what happens if the previous deployment isn't ready? I thought we are expected to wait rather than error out? 
2. Is this expect every loop to instantly succeed? I did not see any wait function in the loop. 
Correct me if I'm wrong. Thanks!

awgreene: > what happens if the previous deployment isn't ready?

Then the next deployment is not installed until it is ready. The returned error simply requeues the CSV in the installing phase.

Bowenislandsong: ahh, requeue. Thanks.

awgreene:
Closing this so we can redesign the ordered rollout feature.

PR #1418, opened 2020-04-01
Create APIService after API server deployment

awgreene:
This commit introduces a change so the APISerivce is created after the
API Server is successfully deployed on the cluster.

### DIFF
--- a/pkg/controller/operators/olm/apiservices.go
+++ b/pkg/controller/operators/olm/apiservices.go
@@ -298,7 +298,7 @@ func apiServiceDescriptionsForDeployment(descs []v1alpha1.APIServiceDescription,
 	return result
 }
 
-func (a *Operator) installOwnedAPIServiceRequirements(csv *v1alpha1.ClusterServiceVersion, strategy install.Strategy) (install.Strategy, error) {
+func (a *Operator) installOwnedAPIServiceRequirements(csv *v1alpha1.ClusterServiceVersion, strategy install.Strategy) (install.Strategy, map[string][]byte, error) {
 	logger := log.WithFields(log.Fields{
 		"csv":       csv.GetName(),
 		"namespace": csv.GetNamespace(),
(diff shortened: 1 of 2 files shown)


exdx:
/lgtm

Looks a little funny returning the map everywhere, but no way around it outside of a named return (which I personally don't like) and I don't think work in cases where only one return argument is named and others are not. Pointers to maps also don't work. Putting the map on the `Operator` object as an attribute also doesn't seem right. So lets leave it as is.

pkg/controller/operators/olm/apiservices.go
```diff
@@ -298,29 +298,30 @@ func apiServiceDescriptionsForDeployment(descs []v1alpha1.APIServiceDescription,
...
 		"csv":       csv.GetName(),
 		"namespace": csv.GetNamespace(),
 	})
 
 	// Assume the strategy is for a deployment
+	deploymentNameToCAPEM := map[string][]byte{}
 	strategyDetailsDeployment, ok := strategy.(*v1alpha1.StrategyDetailsDeployment)
 	if !ok {
-		return nil, fmt.Errorf("unsupported InstallStrategy type")
+		return nil, deploymentNameToCAPEM, fmt.Errorf("unsupported InstallStrategy type")
```
exdx, line 311: wait could we return a `nil` map here instead?

awgreene: Actually, we can just return nil because maps are pointers, as shown [here](https://dave.cheney.net/2017/04/30/if-a-map-isnt-a-reference-variable-what-is-it).

pkg/controller/operators/olm/apiservices.go
```diff
@@ -307,24 +307,25 @@ func (a *Operator) installOwnedAPIServiceRequirements(csv *v1alpha1.ClusterServi
...
 	ca, err := certs.GenerateCA(expiration, Organization)
 	if err != nil {
 		logger.Debug("failed to generate CA")
-		return nil, err
+		return nil, nil, err
 	}
 	rotateAt := expiration.Add(-1 * DefaultCertMinFresh)
 
 	apiDescs := csv.GetOwnedAPIServiceDescriptions()
+	deploymentNameToCAPEM := map[string][]byte{}
```
exdx, line 328: Best to use the `make` built-in when initializing maps ie `deploymentNameToCAPEM := make(map[...])`

awgreene: Updated, but based on [this](https://stackoverflow.com/questions/31064688/which-is-the-nicer-way-to-initialize-a-map-in-golang) conversation from StackOverflow there is no performance gain.

awgreene:
`[Fail] Install Plan with CRD schema change Test [It] existing version is present in new CRD (deprecated field) `

Ran just fine locally.

/retest

awgreene:
/retest
`delete internal registry pod triggers recreation `

PR #1424, opened 2020-04-03
Modify dynamic_resource_e2e_test to ginkgo

harishsurf:
This change adds more structure to the tests by leveraging ginkgo's container/It blocks. The changes also cleans up some of the functions in util_test.go. I have posted some questions in the Files changed section for which I need some inputs. Thanks!

**Reviewer Checklist**

### DIFF
--- a/test/e2e/catalog_e2e_test.go
+++ b/test/e2e/catalog_e2e_test.go
@@ -31,7 +31,7 @@ import (
 var _ = Describe("Catalog", func() {
 	It("loading between restarts", func() {
 
-		defer cleaner.NotifyTestComplete(GinkgoT(), true)
+		defer cleaner.NotifyTestComplete(true)
 
 		// create a simple catalogsource
 		packageName := genName("nginx")
@@ -52,15 +52,15 @@ var _ = Describe("Catalog", func() {
 		namedStrategy := newNginxInstallStrategy(genName("dep-"), nil, nil)
 		csv := newCSV(packageStable, testNamespace, "", semver.MustParse("0.1.0"), []apiextensions.CustomResourceDefinition{crd}, nil, namedStrategy)
 
-		c := newKubeClient(GinkgoT())
-		crc := newCRClient(GinkgoT())
+		c := newKubeClient()
+		crc := newCRClient()
 
 		catalogSourceName := genName("mock-ocs-")
 		_, cleanupSource := createInternalCatalogSource(GinkgoT(), c, crc, catalogSourceName, operatorNamespace, manifests, []apiextensions.CustomResourceDefinition{crd}, []v1alpha1.ClusterServiceVersion{csv})
 		defer cleanupSource()
 
 		// ensure the mock catalog exists and has been synced by the catalog operator
-		catalogSource, err := fetchCatalogSource(GinkgoT(), crc, catalogSourceName, operatorNamespace, catalogSourceRegistryPodSynced)
+		catalogSource, err := fetchCatalogSourceOnStatus(crc, catalogSourceName, operatorNamespace, catalogSourceRegistryPodSynced)
 		require.NoError(GinkgoT(), err)
 
 		// get catalog operator deployment
@@ -74,7 +74,7 @@ var _ = Describe("Catalog", func() {
 		GinkgoT().Log("Catalog operator rescaled")
 		GinkgoT( // check for last synced update to catalogsource
 		).Log("Checking for catalogsource lastSync updates")
-		_, err = fetchCatalogSource(GinkgoT(), crc, catalogSourceName, operatorNamespace, func(cs *v1alpha1.CatalogSource) bool {
+		_, err = fetchCatalogSourceOnStatus(crc, catalogSourceName, operatorNamespace, func(cs *v1alpha1.CatalogSource) bool {
 			before := catalogSource.Status.GRPCConnectionState
 			after := cs.Status.GRPCConnectionState
 			if after != nil && after.LastConnectTime.After(before.LastConnectTime.Time) {
@@ -88,11 +88,11 @@ var _ = Describe("Catalog", func() {
 	})
 	It("global update triggers subscription sync", func() {
 
-		defer cleaner.NotifyTestComplete(GinkgoT(), true)
+		defer cleaner.NotifyTestComplete(true)
 
 		globalNS := operatorNamespace
-		c := newKubeClient(GinkgoT())
-		crc := newCRClient(GinkgoT())
+		c := newKubeClient()
+		crc := newCRClient()
 
 		// Determine which namespace is global. Should be `openshift-marketplace` for OCP 4.2+.
 		// Locally it is `olm`
@@ -135,7 +135,7 @@ var _ = Describe("Catalog", func() {
 		createInternalCatalogSource(GinkgoT(), c, crc, mainCatalogName, globalNS, mainManifests, []apiextensions.CustomResourceDefinition{mainCRD}, []v1alpha1.ClusterServiceVersion{mainCSV})
 
 		// Attempt to get the catalog source before creating install plan
-		_, err := fetchCatalogSource(GinkgoT(), crc, mainCatalogName, globalNS, catalogSourceRegistryPodSynced)
+		_, err := fetchCatalogSourceOnStatus(crc, mainCatalogName, globalNS, catalogSourceRegistryPodSynced)
 		require.NoError(GinkgoT(), err)
 
 		subscriptionSpec := &v1alpha1.SubscriptionSpec{
@@ -151,7 +151,7 @@ var _ = Describe("Catalog", func() {
 		subscriptionName := genName("sub-")
 		createSubscriptionForCatalogWithSpec(GinkgoT(), crc, testNamespace, subscriptionName, subscriptionSpec)
 
-		subscription, err := fetchSubscription(GinkgoT(), crc, testNamespace, subscriptionName, subscriptionHasInstallPlanChecker)
+		subscription, err := fetchSubscription(crc, testNamespace, subscriptionName, subscriptionHasInstallPlanChecker)
 		require.NoError(GinkgoT(), err)
 		require.NotNil(GinkgoT(), subscription)
 
@@ -182,10 +182,10 @@ var _ = Describe("Catalog", func() {
 		updateInternalCatalog(GinkgoT(), c, crc, mainCatalogName, globalNS, []apiextensions.CustomResourceDefinition{mainCRD}, []v1alpha1.ClusterServiceVersion{mainCSV, replacementCSV}, mainManifests)
 
 		// Get updated catalogsource
-		fetchedUpdatedCatalog, err := fetchCatalogSource(GinkgoT(), crc, mainCatalogName, globalNS, catalogSourceRegistryPodSynced)
+		fetchedUpdatedCatalog, err := fetchCatalogSourceOnStatus(crc, mainCatalogName, globalNS, catalogSourceRegistryPodSynced)
 		require.NoError(GinkgoT(), err)
 
-		subscription, err = fetchSubscription(GinkgoT(), crc, testNamespace, subscriptionName, subscriptionStateUpgradePendingChecker)
+		subscription, err = fetchSubscription(crc, testNamespace, subscriptionName, subscriptionStateUpgradePendingChecker)
 		require.NoError(GinkgoT(), err)
 		require.NotNil(GinkgoT(), subscription)
 
@@ -197,7 +197,7 @@ var _ = Describe("Catalog", func() {
 	})
 	It("config map update triggers registry pod rollout", func() {
 
-		defer cleaner.NotifyTestComplete(GinkgoT(), true)
+		defer cleaner.NotifyTestComplete(true)
 
 		mainPackageName := genName("nginx-")
 		dependentPackageName := genName("nginxdep-")
@@ -216,8 +216,8 @@ var _ = Describe("Catalog", func() {
 		mainCSV := newCSV(mainPackageStable, testNamespace, "", semver.MustParse("0.1.0"), nil, []apiextensions.CustomResourceDefinition{dependentCRD}, mainNamedStrategy)
 		dependentCSV := newCSV(dependentPackageStable, testNamespace, "", semver.MustParse("0.1.0"), []apiextensions.CustomResourceDefinition{dependentCRD}, nil, dependentNamedStrategy)
 
-		c := newKubeClient(GinkgoT())
-		crc := newCRClient(GinkgoT())
+		c := newKubeClient()
+		crc := newCRClient()
 
 		mainCatalogName := genName("mock-ocs-main-")
 
@@ -246,7 +246,7 @@ var _ = Describe("Catalog", func() {
 		createInternalCatalogSource(GinkgoT(), c, crc, mainCatalogName, testNamespace, mainManifests, nil, []v1alpha1.ClusterServiceVersion{mainCSV})
 
 		// Attempt to get the catalog source before creating install plan
-		fetchedInitialCatalog, err := fetchCatalogSource(GinkgoT(), crc, mainCatalogName, testNamespace, catalogSourceRegistryPodSynced)
+		fetchedInitialCatalog, err := fetchCatalogSourceOnStatus(crc, mainCatalogName, testNamespace, catalogSourceRegistryPodSynced)
 		require.NoError(GinkgoT(), err)
 
 		// Get initial configmap
@@ -265,7 +265,7 @@ var _ = Describe("Catalog", func() {
 		updatedConfigMap, err := c.KubernetesInterface().CoreV1().ConfigMaps(testNamespace).Get(context.TODO(), fetchedInitialCatalog.Spec.ConfigMap, metav1.GetOptions{})
 		require.NoError(GinkgoT(), err)
 
-		fetchedUpdatedCatalog, err := fetchCatalogSource(GinkgoT(), crc, mainCatalogName, testNamespace, func(catalog *v1alpha1.CatalogSource) bool {
+		fetchedUpdatedCatalog, err := fetchCatalogSourceOnStatus(crc, mainCatalogName, testNamespace, func(catalog *v1alpha1.CatalogSource) bool {
 			before := fetchedInitialCatalog.Status.ConfigMapResource
 			after := catalog.Status.ConfigMapResource
 			if after != nil && before.LastUpdateTime.Before(&after.LastUpdateTime) &&
@@ -297,9 +297,9 @@ var _ = Describe("Catalog", func() {
 
 		// Create Subscription
 		subscriptionName := genName("sub-")
-		createSubscriptionForCatalog(GinkgoT(), crc, testNamespace, subscriptionName, fetchedUpdatedCatalog.GetName(), mainPackageName, stableChannel, "", v1alpha1.ApprovalAutomatic)
+		createSubscriptionForCatalog(crc, testNamespace, subscriptionName, fetchedUpdatedCatalog.GetName(), mainPackageName, stableChannel, "", v1alpha1.ApprovalAutomatic)
 
-		subscription, err := fetchSubscription(GinkgoT(), crc, testNamespace, subscriptionName, subscriptionStateAtLatestChecker)
+		subscription, err := fetchSubscription(crc, testNamespace, subscriptionName, subscriptionStateAtLatestChecker)
 		require.NoError(GinkgoT(), err)
 		require.NotNil(GinkgoT(), subscription)
 		_, err = fetchCSV(GinkgoT(), crc, subscription.Status.CurrentCSV, testNamespace, buildCSVConditionChecker(v1alpha1.CSVPhaseSucceeded))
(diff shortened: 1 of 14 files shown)


test/e2e/util_test.go
```diff
@@ -73,25 +74,25 @@ func newNamespaceCleaner(namespace string) *namespaceCleaner {
...
 		return
 	}
 
-	cleanupOLM(t, c.namespace)
+	cleanupOLM(c.namespace)
 }
 
 // newKubeClient configures a client to talk to the cluster defined by KUBECONFIG
-func newKubeClient(t GinkgoTInterface) operatorclient.ClientInterface {
+func newKubeClient() operatorclient.ClientInterface {
```
harishsurf, line 91: I was thinking of removing this wrapper method that returns a client and directly get the client from the `ctx` object? Thoughts?

test/e2e/dynamic_resource_e2e_test.go
```diff
@@ -1,105 +1,163 @@
...
-				}
+	BeforeEach(func() {
+		c = newKubeClient()
+		crc = newCRClient()
+		dynamicClient = ctx.Ctx().DynamicClient()
+	})
+
+	Context("Create namespace and CatalogSource ", func() {
+		var (
+			ns         *corev1.Namespace
```
harishsurf, line 39: The `ns` variable's type is supposed to be `*v1.Namespace` as per the Create() method's return type.
`Create(*v1.Namespace) (*v1.Namespace, error)`
 But the compiler kept complaining, but for some reason, it worked when I changed the type to `*corev1.Namespace`
Not sure if it is correct?

test/e2e/dynamic_resource_e2e_test.go
```diff
@@ -1,105 +1,163 @@
...
+							Kind: step.Resource.Kind,
+						}
+						for expected := range expectedSteps {
+							if strings.HasPrefix(key.Name, expected.Name) && key.Kind == expected.Kind {
+								delete(expectedSteps, expected)
+							}
+						}
+					}
+					//Expect(expectedSteps).To(BeZero(), "Resource steps do not match expected: %#v", expectedSteps)
+					require.Lenf(GinkgoT(), expectedSteps, 0, "Resource steps do not match expected: %#v", expectedSteps)
```
harishsurf, line 118: This line needs to be replaced with the commented line above

test/e2e/dynamic_resource_e2e_test.go
```diff
@@ -1,105 +1,163 @@
...
 	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/registry"
 	"github.com/operator-framework/operator-lifecycle-manager/test/e2e/ctx"
 )
 
-var _ = Describe("Dynamic Resource", func() {
-	It("resolve prometheus API", func() {
+var _ = FDescribe("Dynamic Resource", func() {
```
benluddy, line 21: This will need to be unfocused before it can be merged. Ginkgo's runner will set a nonzero exit status when any of the test nodes are focused, as a safeguard against accidentally merging them, so the E2E check should be failing.

test/e2e/dynamic_resource_e2e_test.go
```diff
@@ -1,105 +1,163 @@
...
 
-		defer cleaner.NotifyTestComplete(GinkgoT(), true)
+	defer GinkgoRecover()
```
benluddy, line 23: `GinkgoRecover` is only necessary when a test spawns its own goroutine, so it's not needed here.

test/e2e/util_test.go
```diff
@@ -73,25 +74,25 @@ func newNamespaceCleaner(namespace string) *namespaceCleaner {
 }
 
 // notifyOnFailure checks if a test has failed or cleanup is true before cleaning a namespace
-func (c *namespaceCleaner) NotifyTestComplete(t GinkgoTInterface, cleanup bool) {
-	if t.Failed() {
+func (c *namespaceCleaner) NotifyTestComplete(t GinkgoTestDescription, cleanup bool) {
```
benluddy, line 77: Does this method need to accept `GinkgoTestDescription` as a parameter? `CurrentGinkgoTestDescription` can be called from within this method to get the same value.

test/e2e/util_test.go
benluddy: Let's leave the `new*Client` functions in place for now. I don't think the `ctx` API has settled yet, so increasing the number of places that depend on it may end up making it harder to change (it's completely fine to be using `ctx.Ctx()` as you have done elsewhere in this PR).

test/e2e/assets/chart/zz_chart.go
```diff
@@ -676,9 +676,9 @@ var _bindata = map[string]func() (*asset, error){
 	"crds/0000_50_olm_10-operatorgroup.crd.yaml":                        crds0000_50_olm_10OperatorgroupCrdYaml,
 	"crds/0000_50_olm_03-clusterserviceversion.crd.yaml":                crds0000_50_olm_03ClusterserviceversionCrdYaml,
 	"crds/0000_50_olm_04-installplan.crd.yaml":                          crds0000_50_olm_04InstallplanCrdYaml,
-	"values.yaml":                                                       valuesYaml,
-	"Chart.yaml":                                                        chartYaml,
-	".helmignore":                                                       Helmignore,
+	"values.yaml": valuesYaml,
```
benluddy, line 679: This file is automatically generated -- it looks like your editor may have decided to reformat this and then the change was accidentally committed.

pkg/api/apis/operators/install/install.go
```diff
@@ -15,4 +15,4 @@ func Install(scheme *runtime.Scheme) {
 	utilruntime.Must(v1alpha1.AddToScheme(scheme))
 	utilruntime.Must(v1.AddToScheme(scheme))
 	utilruntime.Must(scheme.SetVersionPriority(v1.SchemeGroupVersion, v1alpha1.SchemeGroupVersion))
-}
\ No newline at end of file
+}
```
benluddy, line 18: I'm all for newlines at the end of files, but it's good to avoid bundling trivial changes like this into commits that don't otherwise affect a particular file. It makes the commit log a little more useful and helps PR reviewers to focus on the important stuff.

harishsurf: Running `gofmt` on the entire project caused this. I did exclude files outside of `test` dir, but this one slipped through. I have configured GoLand to apply `gofmt` only to the files I change. Hopefully, this will not occur in future commits. My apologies!

test/e2e/dynamic_resource_e2e_test.go
harishsurf: The Ginkgo's doc says, 
if your test launches a goroutine that calls Fail (or, equivalently, invokes a **failing Gomega assertion),** there’s no way for Ginkgo to rescue the panic that Fail invokes. 
So, my understanding was every Gomega assertion that fails causes a panic and hence `GinkgoRecover()` is needed

test/e2e/dynamic_resource_e2e_test.go
ecordell: That's correct, check out line 10:

```go
corev1 "k8s.io/api/core/v1"
```
`Create`'s signature is written in terms of the package it comes from, so the docs for it just say `v1`, but when that package is imported into this file, we rename it to `corev1`. The compiler will check that they refer to the same type

test/e2e/dynamic_resource_e2e_test.go
benluddy: That should be read as "if your test launches a gorouting that calls Fail" OR "if your test launches a goroutine that invokes a failing Gomega assertion". They are both essentially the same thing, because [Gomega is configured to call Fail when an assertion fails](https://github.com/operator-framework/operator-lifecycle-manager/blob/bc06046b43f3f113a7c9e69b89cc3acc72a718bd/test/e2e/e2e_test.go#L43).

test/e2e/dynamic_resource_e2e_test.go
harishsurf: Oh, my bad. I forgot to notice that the package was using an alias.

test/e2e/assets/chart/zz_chart.go
harishsurf: I'll exclude this and other files that are not part of the change

ecordell:
This is on the right track @harishsurf, but we need to work on the test nodes a bit more before merging. This is what it looks like currently:

```
 Dynamic Resource
  Create namespace and CatalogSource 
    Generate a Subscription
      Resolve prometheus API
        Wait for the expected InstallPlan's execution to either fail or succeed
```

Some of this is subjective, but in my opinion the nodes we define have two goals:

 - Be as readable as possible to a wide audience
 - Define contexts that we can (eventually) re-use between tests

```
[Describe] Subscriptions create required objects from Catalogs
  [Context] Given a Namespace with an OwnNamespace OperatorGroup
      [Context] when a CatalogSource is created with a bundle that contains prometheus objects
         [It] should install the operator successfully
             [By] creating a subscription to the CatalogSource
         [It] should have created the expected prometheus objects
```

in the future we may re-use the catalog context by adding more sub-contexts or `it` blocks - the pattern of creating a catalog source with some content, creating a subscription, and verifying the installed objects is one of the most common patterns we have in the e2e tests.

test/e2e/dynamic_resource_e2e_test.go
```diff
@@ -3,103 +3,164 @@ package e2e
...
-			}
-			for expected := range expectedSteps {
-				if strings.HasPrefix(key.Name, expected.Name) && key.Kind == expected.Kind {
-					delete(expectedSteps, expected)
+
+		AfterEach(func() {
+
+			deleteOpts = &metav1.DeleteOptions{}
+			// Delete Namespace
+			err := c.KubernetesInterface().CoreV1().Namespaces().Delete(ns.GetName(), deleteOpts)
```
Bowenislandsong, line 60: IIRC, there are efforts to clean the namespace before deleting it. I wonder if you want to add them here before deleting the ns. I understand the current work would delete all resources that are in the ns, but may cause deletion to hang if some resources decide to stick around. 

On the worst case, I wonder if adding delediate deletion/gracePeriod=0 would help.

test/e2e/dynamic_resource_e2e_test.go
```diff
@@ -3,103 +3,164 @@ package e2e
...
-
-		gvr.Resource = "servicemonitors"
-		err = waitForGVR(dynamicClient, gvr, "my-servicemonitor", ns.GetName())
-		require.NoError(GinkgoT(), err)
+
+				catsrc, err = crc.OperatorsV1alpha1().CatalogSources(catsrc.GetNamespace()).Create(catsrc)
+				Expect(err).NotTo(HaveOccurred())
+
+				// Wait for the CatalogSource to be ready
+				catsrc, err = fetchCatalogSource(crc, catsrc.GetName(), catsrc.GetNamespace(), catalogSourceRegistryPodSynced)
```
Bowenislandsong, line 87: You can change the `fetchCatalogSource` function name if you feel need to explain what it does.

test/e2e/dynamic_resource_e2e_test.go
```diff
@@ -3,103 +3,164 @@ package e2e
...
-				Name: step.Resource.Name,
-				Kind: step.Resource.Kind,
-			}
-			for expected := range expectedSteps {
-				if strings.HasPrefix(key.Name, expected.Name) && key.Kind == expected.Kind {
-					delete(expectedSteps, expected)
+
+		AfterEach(func() {
+
+			deleteOpts = &metav1.DeleteOptions{}
```
Bowenislandsong, line 58: either make this global or declare it every time you use it.

Bowenislandsong: Ginkgo randomize executions. This deleteOpts may miss its ordering.

test/e2e/dynamic_resource_e2e_test.go
harishsurf: @Bowenislandsong , what resources should be cleaned up before deleting a namespace? I was under the impression that [`cleaner.NotifyTestComplete(true)` ](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/test/e2e/dynamic_resource_e2e_test.go#L20)which internally calls `cleanupOLM` deletes OLM CR's.

test/e2e/dynamic_resource_e2e_test.go
harishsurf: `fetchCatalogSourceOnReady`  or `fetchCatalogSourceWhenAvailable` would be a better name?

Bowenislandsong: `fetchCatalogSourceOnStatus` ? Since you define the status of it to wait on.

test/e2e/dynamic_resource_e2e_test.go
Bowenislandsong: ahh. I see. Then this is good. Thanks for the explanation!

test/e2e/dynamic_resource_e2e_test.go
```diff
@@ -3,103 +3,164 @@ package e2e
...
-		crc := newCRClient(GinkgoT())
-		dynamicClient := ctx.Ctx().DynamicClient()
+	BeforeEach(func() {
+		c = newKubeClient()
+		crc = newCRClient()
+		dynamicClient = ctx.Ctx().DynamicClient()
+	})
+
+	AfterEach(func() {
+		deleteOpts = &metav1.DeleteOptions{}
```
benluddy, line 36: Why is it important to assign `deleteOpts` here? Should this be happening in `BeforeEach`?

test/e2e/dynamic_resource_e2e_test.go
```diff
@@ -3,103 +3,164 @@ package e2e
...
+	})
 
-		ns, err := c.KubernetesInterface().CoreV1().Namespaces().Create(&corev1.Namespace{
-			ObjectMeta: metav1.ObjectMeta{
-				Name: genName("ns-"),
-			},
+	Context("Given a Namespace with an OwnNamespace OperatorGroup", func() {
+		var (
+			ns  *corev1.Namespace
+			err error
```
benluddy, line 44: Does `err` need to be shared across this Context, or just ns?

test/e2e/dynamic_resource_e2e_test.go
```diff
@@ -3,103 +3,164 @@ package e2e
...
+
+		AfterEach(func() {
+
+			// Delete Namespace
+			err := c.KubernetesInterface().CoreV1().Namespaces().Delete(ns.GetName(), deleteOpts)
+			Expect(err).NotTo(HaveOccurred())
+
+		})
+
+		Context("when a CatalogSource is created with a bundle that contains prometheus objects", func() {
```
benluddy, line 65: `When` is functionally equivalent to `Context`, so you could change this to `When("a CatalogSource is created..."`.

test/e2e/dynamic_resource_e2e_test.go
```diff
@@ -3,103 +3,164 @@ package e2e
...
+
+			Context("should install the operator successfully", func() {
+				var (
+					subName    string
+					sub        *v1alpha1.Subscription
+					cleanupSub cleanupFunc
+				)
+
+				AfterEach(func() {
+					cleanupSub()
```
benluddy, line 107: This is fragile because it will panic if `cleanupSub` is not assigned a non-nil value during every spec that is added to this context. Maybe all three of the context-scoped variables belong inside the spec, since there is no common setup happening here.

test/e2e/dynamic_resource_e2e_test.go
harishsurf: It was because `deleteOpts` are only used for deleting resources (ns, catalog source) and which were only used in `AfterEach`

test/e2e/dynamic_resource_e2e_test.go
harishsurf: `err` variable is used in almost all of the context. There could be 2 ways of declaring it: 
1. globally declare it (in the outermost Describe block) or
2. redeclare it in every `Context`/`Describe` wherever it is used. 

Because each `It` spec calls BeforeEach (outermost to inner block), the variable gets redeclared and so there won't be any test pollution. In the above case, I have declared only once, starting in the `Context` where it gets used.

benluddy: IMO, there isn't any value in reducing the number of error identifiers that are declared. The real motivation behind declaring variables at the Context scope is so that values can be assigned to them in a shared Before node and later accessed in specs. Maybe I missed it, but I don't see that happening in this situation -- the _value_ assigned to `err` doesn't need to be shared throughout this test scope.

Another factor to consider is: how obvious will it be to a person who later modifies this test so that it doesn't always assign a value to `err` in `BeforeEach`?

test/e2e/dynamic_resource_e2e_test.go
benluddy: Keep in mind that `AfterEach` closures run from innermost to outermost in situations where you have nested test scopes, like this one. So this assignment is happening AFTER it is being used elsewhere. I suspect that the client's Delete method allows nil for its DeleteOptions argument and it coincidentally working, but I don't think this is doing what you expect. Since it never changes, why not just assign it a value at the time it is declared?

test/e2e/dynamic_resource_e2e_test.go
```diff
@@ -3,103 +3,156 @@ package e2e
...
+
+						err = waitForGVR(dynamicClient, gvr, "my-prometheusrule", ns.GetName())
+						Expect(err).NotTo(HaveOccurred())
+
+						gvr.Resource = "servicemonitors"
+						err = waitForGVR(dynamicClient, gvr, "my-servicemonitor", ns.GetName())
+						Expect(err).NotTo(HaveOccurred())
+					})
+
+					defer cleanupSub()
```
benluddy, line 150: Should this statement appear on the line after `cleanupSub` is assigned? If `fetchSubscription` returns an error, this will never be executed.

test/e2e/dynamic_resource_e2e_test.go
harishsurf: yep, you are right. Moved the initialization to `BeforeEach` block

test/e2e/dynamic_resource_e2e_test.go
harishsurf: Makes sense. I have redeclared the variable every time it gets used in a `Context`

test/e2e/dynamic_resource_e2e_test.go
```diff
@@ -3,103 +3,160 @@ package e2e
...
+
+				// Wait for the CatalogSource to be ready
+				catsrc, err = fetchCatalogSourceOnStatus(crc, catsrc.GetName(), catsrc.GetNamespace(), catalogSourceRegistryPodSynced)
+				Expect(err).NotTo(HaveOccurred())
+
+			})
+
+			AfterEach(func() {
+				// Delete CatalogSource
+				err := crc.OperatorsV1alpha1().CatalogSources(catsrc.GetNamespace()).Delete(catsrc.GetName(), deleteOpts)
```
benluddy, line 97: Will this panic when it tries to call `catsrc.GetNamespace()` if `fetchCatalogSourceOnStatus` returned nil on L90?

harishsurf: I have added `catsrc` being non-nil check

benluddy:
/hold

Stopping unnecessary CI jobs while feature freeze is in effect.

test/e2e/csv_e2e_test.go
```diff
@@ -36,10 +36,10 @@ import (
 var _ = Describe("CSV", func() {
 	It("create with unmet requirements mini kube version", func() {
 
-		defer cleaner.NotifyTestComplete(GinkgoT(), true)
+		defer cleaner.NotifyTestComplete(true)
```
awgreene, line 39: Shouldn't we use AfterEach?

test/e2e/catalog_e2e_test.go
```diff
@@ -31,7 +31,7 @@ import (
 var _ = Describe("Catalog", func() {
 	It("loading between restarts", func() {
 
-		defer cleaner.NotifyTestComplete(GinkgoT(), true)
+		defer cleaner.NotifyTestComplete(true)
```
awgreene, line 34: Shouldn't this use the `AfterEach` Ginkgo feature? Applies in all places

test/e2e/csv_e2e_test.go
harishsurf: @awgreene , these will be changed when I pick up csv_e2e_test.go for ginkgo conversion. For now, this PR is only concerned with dynamic_e2e_test and some of the changes in helper methods

test/e2e/catalog_e2e_test.go
harishsurf: Same explanation as above

awgreene: Oops

shawn-hurley:
Overriding valid bug because this is fixing tests

PR #1426, opened 2020-04-05
WIP: Move APIService management to the Installer

awgreene:


awgreene:
This was completed in a separate PR.

PR #1429, opened 2020-04-06
Convert gc_e2e_test.go to ginkgo

harishsurf:
This change converts gc_e2e_test.go to leverage ginkgo's BDD approach

**Reviewer Checklist**

### DIFF
--- a/test/e2e/gc_e2e_test.go
+++ b/test/e2e/gc_e2e_test.go
@@ -3,11 +3,11 @@ package e2e
 import (
 	"context"
 	"fmt"
+	"github.com/operator-framework/api/pkg/operators/v1alpha1"
 
 	"github.com/blang/semver"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
-	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
@@ -23,7 +23,7 @@ import (
 	. "github.com/operator-framework/operator-lifecycle-manager/test/e2e/dsl"
 )
 
-var _ = Describe("Garbage collector", func() {
+var _ = Describe("Garbage collection for dependent resources", func() {
 	var (
 		kubeClient     operatorclient.ClientInterface
 		operatorClient versioned.Interface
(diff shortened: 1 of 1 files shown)


test/e2e/gc_e2e_test.go
```diff
@@ -33,159 +34,243 @@ var _ = Describe("Garbage collector", func() {
...
+
+						// wait for deletion of ownerB
+						waitForDelete(func() error {
+							_, err := operatorClient.OperatorsV1alpha1().ClusterServiceVersions(testNamespace).Get(ownerB.GetName(), metav1.GetOptions{})
+							return err
+						})
+
+						//By("check for dependent (should be deleted since last blocking owner was deleted", func() {
+						// check for dependent (should be deleted since last blocking owner was deleted)
+						cf, err := kubeClient.KubernetesInterface().CoreV1().ConfigMaps(testNamespace).Get(dependent.GetName(), metav1.GetOptions{})
```
harishsurf, line 262: The test ensures that the dependent `configMap` is not deleted if one of its owner reference is deleted. 

Problem: The assertion at line 266 fails because err is `nil`

What I found: 
When ownerB CSV is deleted, an empty `configMap` struct is returned in the older version. (line 262)
See below output: 
Current version:
```
###B &ConfigMap{ObjectMeta:{dependent  openshift-operators /api/v1/namespaces/openshift-operators/configmaps/dependent 412d8073-d3ce-4d11-9a77-7876f0e8e062 161740 0 2020-04-06 18:25:30 -0400 EDT <nil> <nil> map[] map[] [
```
Older version: 
```
###B &ConfigMap{ObjectMeta:{      0 0001-01-01 00:00:00 +0000 UTC <nil> <nil> map[] map[] [] []  []},Data:map[string]string{},BinaryData:map[string][]byte{},}cleaning up any remaining non persistent resources...
```
The `cf` returned here has value whereas in the older version of the file, `cf` returns an empty ConfigMap struct as seen above

Because of this, at line 266
Current version:
```
 Error is <nil> and value is <nil>
```
vs 
Older version:
```
 Error type *errors.StatusError and value is configmaps "dependent" not found
```

Bowenislandsong: I wonder if that has anything todo with your forgrand deletion option. Since the object in the configmap returnd is the dependent.

harishsurf: Figured out the issue. It was related to the placement of the deletion code inside of `It` block. Thanks for looking into this!

test/e2e/gc_e2e_test.go
```diff
@@ -33,159 +32,281 @@ var _ = Describe("Garbage collector", func() {
...
+					)
+
+					BeforeEach(func() {
+						propagation = "Foreground"
+						options = metav1.DeleteOptions{PropagationPolicy: &propagation}
+					})
+
+					When("removing one of the owner", func() {
+
+						It("deleting ownerA CSV in the foreground", func() {
```
ecordell, line 228: nit: 
```suggestion
						It("deletes ownerA CSV in the foreground", func() {
```

test/e2e/gc_e2e_test.go
```diff
@@ -33,159 +32,281 @@ var _ = Describe("Garbage collector", func() {
...
+					// add owners
+					ownerutil.AddOwner(dependent, fetchedA, true, false)
+					ownerutil.AddOwner(dependent, fetchedB, true, false)
+
+					// create ConfigMap dependent
+					_, err = kubeClient.KubernetesInterface().CoreV1().ConfigMaps(testNamespace).Create(dependent)
+					Expect(err).NotTo(HaveOccurred(), "dependent could not be created")
+				})
+
+				Context("owner reference behavior can be verified", func() {
```
ecordell, line 215: shouldn't this describe the context more? something about the foreground propogation policy?

test/e2e/gc_e2e_test.go
```diff
@@ -33,159 +32,281 @@ var _ = Describe("Garbage collector", func() {
...
+	})
+
+	// TestOwnerReferenceGCBehavior runs a simple check on OwnerReference behavior to ensure
+	// a resource with multiple OwnerReferences will not be garbage collected when one of its
+	// owners has been deleted.
+	// Test Case:
+	//				CSV-A     CSV-B                        CSV-B
+	//				   \      /      --Delete CSV-A-->       |
+	//				   ConfigMap						 ConfigMap
+	Describe("Multiple OwnerReferences will not be garbage collected", func() {
```
ecordell, line 172:
```suggestion
	Describe("Resources with Multiple OwnerReferences", func() {
```

test/e2e/gc_e2e_test.go
harishsurf: I have added it :)

test/e2e/gc_e2e_test.go
```diff
@@ -3,27 +3,26 @@ package e2e
...
-
-	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/clientset/versioned"
-	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/operatorclient"
-	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/ownerutil"
-	"github.com/operator-framework/operator-lifecycle-manager/test/e2e/ctx"
-	. "github.com/operator-framework/operator-lifecycle-manager/test/e2e/dsl"
 )
 
-var _ = Describe("Garbage collector", func() {
+var _ = Describe("Dependant objects are deleted when its owner is deleted", func() {
```
benluddy, line 25: Nit: I think we want top-level container nodes to name either a class of behaviors or an object whose behaviors are being tested. To me, this comes across as maybe more specific than necessary. I'm imagining something like "Dependent garbage collection". It would be great if we could agree as a team on a good set of guidelines for this sort of convention and write it down somewhere.

test/e2e/gc_e2e_test.go
```diff
@@ -34,159 +33,268 @@ var _ = Describe("Garbage collector", func() {
...
-						Name:    "v1",
-						Served:  true,
-						Storage: true,
-						Schema: &apiextensionsv1.CustomResourceValidation{
-							OpenAPIV3Schema: &apiextensionsv1.JSONSchemaProps{Type: "object"},
+	AfterEach(func() {
+		cleaner.NotifyTestComplete(GinkgoT(), true)
+	})
+
+	Context("Given a ClusterRole owned by a CustomResourceDefinition", func() {
```
benluddy, line 40: The context description should line up with what's actually happening on the test cluster. In this case, the referenced ClusterRole and CRD only exist during a nested spec. How about creating them in a BeforeEach node at this level?

test/e2e/gc_e2e_test.go
```diff
@@ -34,159 +33,268 @@ var _ = Describe("Garbage collector", func() {
...
+
+		When("CustomResourceDefinition is deleted", func() {
```
benluddy, line 42: I have the same feedback for this node. Specs under this context should be able to assume that a ClusterRole exists, and that the ClusterRole was at one point owned by a CRD, and that the CRD has since been deleted. Then the below spec can simply assert that the ClusterRole will eventually not exist.

test/e2e/gc_e2e_test.go
```diff
@@ -34,159 +33,268 @@ var _ = Describe("Garbage collector", func() {
...
-		defer func() {
-			IgnoreError(kubeClient.DeleteAPIService(as.GetName(), &metav1.DeleteOptions{}))
-		}()
-
-		cr, err := kubeClient.CreateClusterRole(&rbacv1.ClusterRole{
-			ObjectMeta: metav1.ObjectMeta{
-				GenerateName:    "clusterrole-",
-				OwnerReferences: []metav1.OwnerReference{ownerutil.NonBlockingOwner(as)},
-			},
+	Context("Given a ClusterRole owned by a APIService", func() {
```
benluddy, line 103: Since the APIService spec is essentially the same as the CRD one, my other comments apply equally here.

test/e2e/gc_e2e_test.go
```diff
@@ -34,159 +33,268 @@ var _ = Describe("Garbage collector", func() {
...
+	// TestOwnerReferenceGCBehavior runs a simple check on OwnerReference behavior to ensure
+	// a resource with multiple OwnerReferences will not be garbage collected when one of its
+	// owners has been deleted.
+	// Test Case:
+	//				CSV-A     CSV-B                        CSV-B
+	//				   \      /      --Delete CSV-A-->       |
+	//				   ConfigMap						 ConfigMap
+	Describe("Resources with multiple OwnerReferences", func() {
+
+		Context("Given a dependent resource associated with multiple owners", func() {
```
benluddy, line 164: These contexts seem redundant to me. Can they be condensed into a single container node? I also have the same feedback about the context description matching the state of the cluster -- the resources should be created here if the context description says they are given.

test/e2e/gc_e2e_test.go
```diff
@@ -34,159 +33,268 @@ var _ = Describe("Garbage collector", func() {
...
+						Expect(err).NotTo(HaveOccurred())
+
+						// wait for deletion of ownerB
+						err = waitForDelete(func() error {
+							_, err := operatorClient.OperatorsV1alpha1().ClusterServiceVersions(testNamespace).Get(context.TODO(), ownerB.GetName(), metav1.GetOptions{})
+							return err
+						})
+						Expect(err).NotTo(HaveOccurred())
+
+						By("should have deleted the dependent since both the owners were deleted", func() {
```
benluddy, line 283: This right here is the spec: "It should delete the dependent".

I'd also suggest that we don't pass the optional closure to `ginkgo.By`. It gives the impression that a new container or spec node is being created, but it doesn't actually work that way.

test/e2e/gc_e2e_test.go
```diff
@@ -34,159 +33,268 @@ var _ = Describe("Garbage collector", func() {
...
+							return err
+						})
+						Expect(err).NotTo(HaveOccurred())
+
+						// delete ownerB in the foreground (to ensure any "blocking" dependents are deleted before ownerB)
+						err = operatorClient.OperatorsV1alpha1().ClusterServiceVersions(testNamespace).Delete(context.TODO(), fetchedB.GetName(), options)
+						Expect(err).NotTo(HaveOccurred())
+
+						// wait for deletion of ownerB
+						err = waitForDelete(func() error {
```
benluddy, line 277: Can we use `Eventually` for our polling assertions?

test/e2e/gc_e2e_test.go
```diff
@@ -34,159 +33,268 @@ var _ = Describe("Garbage collector", func() {
...
+	// owners has been deleted.
+	// Test Case:
+	//				CSV-A     CSV-B                        CSV-B
+	//				   \      /      --Delete CSV-A-->       |
+	//				   ConfigMap						 ConfigMap
+	Describe("Resources with multiple OwnerReferences", func() {
+
+		Context("Given a dependent resource associated with multiple owners", func() {
+
+			Context("using 'Foreground' deletion policy, owner reference behavior can be verified ", func() {
```
benluddy, line 166: The "owner reference behavior can be verified" part may be overspecified. The context description only needs to explain the assumed state, and individual specs can describe the behavior they verify.

test/e2e/gc_e2e_test.go
```diff
@@ -3,27 +3,26 @@ package e2e
...
 
 	"github.com/blang/semver"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
-	"github.com/stretchr/testify/require"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/clientset/versioned"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/operatorclient"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/ownerutil"
+	"github.com/operator-framework/operator-lifecycle-manager/test/e2e/ctx"
+	. "github.com/operator-framework/operator-lifecycle-manager/test/e2e/dsl"
```
awgreene, line 15: OLM typically leaves it's own imports in it's own block.

test/e2e/gc_e2e_test.go
```diff
@@ -3,27 +3,26 @@ package e2e
...
-
-	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/clientset/versioned"
-	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/operatorclient"
-	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/ownerutil"
-	"github.com/operator-framework/operator-lifecycle-manager/test/e2e/ctx"
-	. "github.com/operator-framework/operator-lifecycle-manager/test/e2e/dsl"
 )
 
-var _ = Describe("Garbage collector", func() {
+var _ = Describe("Dependant garbage collection", func() {
```
awgreene, line 25: What do you mean by Dependant?

test/e2e/gc_e2e_test.go
```diff
@@ -34,159 +33,257 @@ var _ = Describe("Garbage collector", func() {
...
 
-		Expect(kubeClient.DeleteAPIService(as.GetName(), &metav1.DeleteOptions{})).To(Succeed())
-		Eventually(func() bool {
-			_, err := kubeClient.GetClusterRole(cr.GetName())
-			return k8serrors.IsNotFound(err)
-		}).Should(BeTrue(), "get cluster role should eventually return \"not found\"")
+		When("APIService is deleted", func() {
+
+			BeforeEach(func() {
+
```
awgreene, line 162: Don't need this newline.

test/e2e/gc_e2e_test.go
```diff
@@ -34,159 +33,257 @@ var _ = Describe("Garbage collector", func() {
...
+				// Delete API service
+				Expect(kubeClient.DeleteAPIService(as.GetName(), &metav1.DeleteOptions{})).To(Succeed())
+			})
+
+			It("should delete the associated ClusterRole", func() {
+
```
awgreene, line 168: Newline comment applies here as well.

test/e2e/gc_e2e_test.go
```diff
@@ -34,159 +33,257 @@ var _ = Describe("Garbage collector", func() {
...
+			// create ConfigMap dependent
+			_, err = kubeClient.KubernetesInterface().CoreV1().ConfigMaps(testNamespace).Create(context.TODO(), dependent, metav1.CreateOptions{})
+			Expect(err).NotTo(HaveOccurred(), "dependent could not be created")
+
+			propagation = metav1.DeletePropagationForeground
+			options = metav1.DeleteOptions{PropagationPolicy: &propagation}
+		})
+
+		When("removing one of the owner using 'Foreground' deletion policy", func() {
```
awgreene, line 230: Might be useful to include a link that describes the `Foreground` deletion policy.

test/e2e/gc_e2e_test.go
harishsurf: Changed it to `Garbage collection for dependant resources`

test/e2e/gc_e2e_test.go
harishsurf: Is it okay to include links to external docs in source code?

test/e2e/gc_e2e_test.go
awgreene: Thanks!

test/e2e/gc_e2e_test.go
```diff
@@ -3,11 +3,11 @@ package e2e
 import (
 	"context"
 	"fmt"
+	"github.com/operator-framework/api/pkg/operators/v1alpha1"
```
ecordell, line 6: nit: should be in the last block with other local imports

non-blocking.

harishsurf: gotcha!

shawn-hurley:
Adding a valid bug because making CI tests more reliable.

shawn-hurley:
/hold

I am worried that we are adding flakes to CI

the test failures appear to be related to this change can someone look?

harishsurf:
> /hold
> 
> I am worried that we are adding flakes to CI
> 
> the test failures appear to be related to this change can someone look?

@shawn-hurley, I ran some analysis(with Ben's help) on our ci-failures using prow's history. Could you take a look at [this gist](https://gist.github.com/harishsurf/3fcd743ec08f00ae4d2396473aa3880b) and see if it has enough evidence to have this and other ginkgo PRs merged? 
cc @benluddy @ecordell

PR #1431, opened 2020-04-07
Enable Operator Discovery

njhale:
__Description of the change:__

Add the operators.coreos.com/v2alpha1 API group and Operator resource with component selection.


cmd/olm/manager.go
```diff
@@ -0,0 +1,58 @@
+package main
+
+import (
+	"fmt"
+	"path/filepath"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/envtest"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
```
exdx, line 10: interesting that controller runtime has a zap log pkg - is it related to the uber zap logger?

exdx:
The operator discovering other operators. The meta has arrived

cmd/olm/manager.go
exdx: Looks like it is

cmd/olm/main.go
```diff
@@ -14,12 +14,13 @@ import (
 	configv1client "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/queueinformer"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
```
ecordell, line 16: nit: import order

cmd/olm/main.go
```diff
@@ -227,5 +234,10 @@ func main() {
 		go monitor.Run(op.Done())
 	}
 
+	// Start the controller manager
+	if err := mgr.Start(ctx.Done()); err != nil {
```
ecordell, line 238: this will bypass our clusteroperator reporting, won't it?

cmd/olm/manager.go
```diff
@@ -0,0 +1,58 @@
...
+
+	// Setup a Manager
+	setupLog.Info("configuring manager")
+	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{MetricsBindAddress: "0"}) // TODO(njhale): Enable metrics on non-conflicting port (not 8080)
+	if err != nil {
+		return nil, err
+	}
+
+	// Setup a new controller to reconcile Operators
+	setupLog.Info("configuring controller")
```
ecordell, line 30: I would remove these info lines or change to debug

deploy/chart/crds/0000_50_olm_00-installplans.yaml
```diff
@@ -78,8 +77,6 @@ spec:
               type: array
               items:
                 type: string
-            generation:
```
ecordell, line 81: we lost these fields?

deploy/chart/crds/0000_50_olm_00-subscriptions.yaml
```diff
@@ -1702,10 +1726,6 @@ spec:
             currentCSV:
               description: CurrentCSV is the CSV the Subscription is progressing to.
               type: string
-            installPlanGeneration:
```
ecordell, line 1705: we lost this field too

pkg/controller/operators/operator_controller.go
```diff
@@ -0,0 +1,259 @@
...
+		Watches(&source.Kind{Type: &corev1.Secret{}}, enqueueOperator).
+		Watches(&source.Kind{Type: &rbacv1.Role{}}, enqueueOperator).
+		Watches(&source.Kind{Type: &rbacv1.RoleBinding{}}, enqueueOperator).
+		Watches(&source.Kind{Type: &rbacv1.ClusterRole{}}, enqueueOperator).
+		Watches(&source.Kind{Type: &rbacv1.ClusterRoleBinding{}}, enqueueOperator).
+		Watches(&source.Kind{Type: &apiextensionsv1.CustomResourceDefinition{}}, enqueueOperator).
+		Watches(&source.Kind{Type: &apiregistrationv1.APIService{}}, enqueueOperator).
+		Watches(&source.Kind{Type: &operatorsv1alpha1.Subscription{}}, enqueueOperator).
+		Watches(&source.Kind{Type: &operatorsv1alpha1.InstallPlan{}}, enqueueOperator).
+		Watches(&source.Kind{Type: &operatorsv1alpha1.ClusterServiceVersion{}}, enqueueOperator).
```
ecordell, line 99: This list needs to be kept in sync with the supported bundle types - i.e. configmap and secret support will merge soon.

pkg/feature/feature.go
```diff
@@ -0,0 +1,40 @@
...
+const (
+	// MyFeature featuregate.Feature = "MyFeature"
+	// owner: @username
+	// alpha: v1.X
+	// (see https://github.com/kubernetes/kubernetes/blob/master/pkg/features/kube_features.go)
+
+	// OperatorLifecycleManagerV2 enables OLM's v2 APIs.
+	// owner: @njhale
+	// alpha: v0.15.0
+	OperatorLifecycleManagerV2 featuregate.Feature = "OperatorLifecycleManagerV2"
```
ecordell, line 18: We probably want to be more granular than this. i.e. `OperatorV1`

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM openshift/origin-release:golang-1.12 as builder
+FROM openshift/origin-release:golang-1.13 as builder
```
ecordell, line 1: Alas, this same change is pending in the 1.18 rebase PR.

Makefile
```diff
@@ -39,8 +42,16 @@ all: test build
 
 test: clean cover.out
 
-unit:
-	go test $(MOD_FLAGS) $(SPECIFIC_UNIT_TEST) -v -race -count=1 ./pkg/...
+unit: kubebuilder
+	$(KUBEBUILDER_ENV) go test $(MOD_FLAGS) $(SPECIFIC_UNIT_TEST) -v -race -count=1 ./pkg/...
```
ecordell, line 46: is there a way we can tie envtest tests to e2e instead of unit? there's a substantial startup cost, right?

Makefile
```diff
@@ -39,8 +42,16 @@ all: test build
...
+
+# Install kubebuilder if not found
+KUBEBUILDER_DIR := /tmp/kubebuilder_2.3.1_$(OS)_$(ARCH)/bin
```
ecordell, line 49: Can do something like`TMPDIR := $(shell mktemp -d)` instead?

Makefile
```diff
@@ -39,8 +42,16 @@ all: test build
...
 
-unit:
-	go test $(MOD_FLAGS) $(SPECIFIC_UNIT_TEST) -v -race -count=1 ./pkg/...
+unit: kubebuilder
+	$(KUBEBUILDER_ENV) go test $(MOD_FLAGS) $(SPECIFIC_UNIT_TEST) -v -race -count=1 ./pkg/...
+
+# Install kubebuilder if not found
+KUBEBUILDER_DIR := /tmp/kubebuilder_2.3.1_$(OS)_$(ARCH)/bin 
+kubebuilder:
+ifeq (, $(shell which kubebuilder))
```
ecordell, line 51: personally I manage kubebuilder with `brew`. is it better ux to just log a message asking you to install it?

Makefile
```diff
@@ -182,11 +189,10 @@ verify-mockgen: mockgen diff
 verify-manifests: manifests diff
 verify: verify-codegen verify-mockgen verify-manifests
 
-
 # before running release, bump the version in OLM_VERSION and push to master,
 # then tag those builds in quay with the version in OLM_VERSION
 release: ver=$(shell cat OLM_VERSION)
-release:
+release: manifests
```
ecordell, line 195: this seems like a major regression if the manifest generation went missing. thanks for catching this.

cmd/olm/main.go
```diff
@@ -37,53 +37,53 @@ const (
 
 // config flags defined globally so that they appear on the test binary as well
 var (
-	kubeConfigPath = flag.String(
```
ecordell, line 40: where did kubeconfig go?

pkg/controller/operators/decorators/operator.go
```diff
@@ -0,0 +1,277 @@
...
+
+		names = append(names, types.NamespacedName{
+			Name: strings.TrimPrefix(key, ComponentLabelKeyPrefix),
+		})
+	}
+
+	return
+}
+
+type OperatorFactory interface {
```
ecordell, line 47: nit: naming here is getting a bit confusing for the callsites, right? `operator := factory.NewOperator(operator)`?

pkg/controller/operators/operator_controller.go
```diff
@@ -0,0 +1,259 @@
...
+		return nil, err
+	}
+
+	factory, err := decorators.NewSchemedOperatorFactory(scheme)
+	if err != nil {
+		return nil, err
+	}
+
+	return &OperatorReconciler{
+		Client: cli,
```
ecordell, line 118: nit: gofmt

pkg/controller/operators/suite_test.go
```diff
@@ -0,0 +1,147 @@
...
+
+func TestAPIs(t *testing.T) {
+	RegisterFailHandler(Fail)
+
+	RunSpecsWithDefaultAndCustomReporters(t,
+		"Controller Suite",
+		[]Reporter{printer.NewlineReporter{}})
+}
+
+var _ = BeforeSuite(func() {
```
ecordell, line 60: does this conflict with the test runner that Ben worked on in (see: `x.mk`)?

pkg/feature/feature.go
```diff
@@ -0,0 +1,40 @@
...
+	utilruntime.Must(mutableGate.Add(featureGates))
+}
+
+// AddFlag adds the feature gates defined in this package to the to the given FlagSet.
+func AddFlag(fs *pflag.FlagSet) {
+	mutableGate.AddFlag(fs)
+}
+
+var featureGates = map[featuregate.Feature]featuregate.FeatureSpec{
+	// OperatorLifecycleManagerV2: {Default: false, PreRelease: featuregate.Alpha},
```
ecordell, line 38: mistake?

pkg/lib/codec/mapstructure.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"reflect"
+	"time"
+
+	"github.com/mitchellh/mapstructure"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func MetaTimeHookFunc() mapstructure.DecodeHookFunc {
```
ecordell, line 11: I expected a small unit test for this

Dockerfile
njhale: Are there blockers on that? Can we push it through first? 

Without this change, the controller-runtime client fails because it requires go 1.13.

Makefile
njhale: I would like to reuse this locally so we don't need to download each time.

Makefile
njhale: the tests execute in just under 4 seconds -- locally, they seem to start pretty quickly.

Makefile
njhale: probably, but we want this to run automatically and it's not clear to me that openshift-ci runs unit tests in an image that we can control the build of (the `src` image). If we do, then I'm all for that route instead.

cmd/olm/main.go
njhale: it's being surfaced as an option by the controller manager from controller-runtime -- they conflict if this is left here -- so we get the value from that package now.

cmd/olm/main.go
njhale: What part in particular? No reporting that currently exists has been removed, is there anything of the Operator API/controller we want/need to add to reporting at this point?

ecordell: Ah, sorry - I just meant that the healthiness of this control loop does not affect the clusteroperator's health. That is not necessarily something we need to figure out right now

Dockerfile
ecordell: My plan was to rebase it after friday. There's nothing blocking it, but pretty much all PRs will need to be rebased (and they're non-trivial since APIs changed)

Makefile
ecordell: We control the base image there - see `base.Dockerfile`. It would be great to not download this every time and be dependent on the network

cmd/olm/manager.go
njhale: I bumped the log level to 4, so we shouldn't see these unless `-v=4` at minimum.

pkg/controller/operators/operator_controller.go
njhale: true -- not sure how we want to handle it. This is a stopgap for generic "watch all" behavior.

pkg/feature/feature.go
njhale: My thinking here is that we're probably going to enable all of the v2 APIs at once when we jump to v2. Do you see a situation where we want to enable APIs independently? i.e. `InstallV2` w/o `OperatorV2`?

Makefile
njhale: sure, I can add it there.

njhale: done

pkg/controller/operators/suite_test.go
njhale: I don't think so. This is scoped to the suite defined in this package, not to mention that the test sets options for the e2e tests and not unit tests.

njhale:
/hold

until API fix is tagged and module version is bumped.

pkg/lib/codec/mapstructure.go
njhale: added

njhale:
Superseded by #1453

PR #1432, opened 2020-04-07
[release-4.3] Bug 1821318: feat(packagemanifest): add default os/arch labels if not present

shawn-hurley:
Manual cherry-pick of: #1220

**Reviewer Checklist**

### DIFF
--- a/pkg/package-server/storage/subresources_test.go
+++ b/pkg/package-server/storage/subresources_test.go
@@ -2,10 +2,11 @@ package storage
 
 import (
 	"context"
-	"k8s.io/apimachinery/pkg/labels"
 	"net/http"
 	"net/http/httptest"
 	"testing"
+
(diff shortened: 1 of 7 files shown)


pkg/package-server/storage/subresources_test.go
```diff
@@ -7,6 +7,8 @@ import (
 	"net/http/httptest"
 	"testing"
 
+	"k8s.io/apimachinery/pkg/labels"
```
ecordell, line 10: this is a surprising diff - I would expect the build to fail

ecordell:
the unit test failure looks legit and related to these changes

sdodson:
@shawn-hurley Can you cherry-pick-approved this PR while I track down why I was removed from the operator-framework org?

PR #1434, opened 2020-04-08
feat: support configmap creation from the bundle

exdx:
Follow up on https://github.com/operator-framework/operator-registry/pull/256.

Now that configmap objects are supported in bundles as first-class citizens, OLM needs to be able to create config maps from the bundle and lifecycle them along with the CSV. This PR enables the catalog operator to work with configmap types. 

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -61,6 +61,7 @@ const (
 	secretKind             = "Secret"
 	clusterRoleKind        = "ClusterRole"
 	clusterRoleBindingKind = "ClusterRoleBinding"
+	configMapKind          = "ConfigMap"
 	serviceAccountKind     = "ServiceAccount"
 	serviceKind            = "Service"
 	roleKind               = "Role"
@@ -1725,6 +1726,28 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
 				s.SetNamespace(namespace)
 
 				status, err := ensurer.EnsureService(namespace, &s)
+				if err != nil {
+					return err
+				}
+
+				plan.Status.Plan[i].Status = status
+
+			case configMapKind:
+				var cfg corev1.ConfigMap
+				err := json.Unmarshal([]byte(step.Resource.Manifest), &cfg)
+				if err != nil {
+					return errorwrap.Wrapf(err, "error parsing step manifest: %s", step.Resource.Name)
+				}
+
+				// Update UIDs on all CSV OwnerReferences
+				updated, err := o.getUpdatedOwnerReferences(cfg.OwnerReferences, plan.Namespace)
+				if err != nil {
+					return errorwrap.Wrapf(err, "error generating ownerrefs for configmap: %s", cfg.GetName())
+				}
+				cfg.SetOwnerReferences(updated)
+				cfg.SetNamespace(namespace)
+
+				status, err := ensurer.EnsureConfigMap(plan.Namespace, &cfg)
 				if err != nil {
 					return err
 				}
--- a/pkg/controller/operators/catalog/step_ensurer.go
+++ b/pkg/controller/operators/catalog/step_ensurer.go
@@ -300,3 +300,26 @@ func (o *StepEnsurer) EnsureUnstructuredObject(client dynamic.ResourceInterface,
 	status = v1alpha1.StepStatusPresent
 	return
 }
+
+// EnsureConfigMap writes the specified ConfigMap object to the cluster.
+func (o *StepEnsurer) EnsureConfigMap(namespace string, configmap *corev1.ConfigMap) (status v1alpha1.StepStatus, err error) {
+	_, createErr := o.kubeClient.KubernetesInterface().CoreV1().ConfigMaps(namespace).Create(configmap)
+	if createErr == nil {
+		status = v1alpha1.StepStatusCreated
+		return
+	}
+
+	if !k8serrors.IsAlreadyExists(createErr) {
+		err = errorwrap.Wrapf(createErr, "error updating configmap: %s", configmap.GetName())
+		return
+	}
+
+	configmap.SetNamespace(namespace)
+	if _, updateErr := o.kubeClient.UpdateConfigMap(configmap); updateErr != nil {
+		err = errorwrap.Wrapf(updateErr, "error updating configmap: %s", configmap.GetName())
+		return
+	}
+
+	status = v1alpha1.StepStatusPresent
+	return
+}
--- a/pkg/lib/operatorclient/configmap.go
+++ b/pkg/lib/operatorclient/configmap.go
@@ -1,0 +1,39 @@
+package operatorclient
+
+import (
+	"fmt"
+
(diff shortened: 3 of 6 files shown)


exdx:
@ecordell would prefer to keep this PR configmap only and follow up with an additional PR for secret types

pkg/controller/operators/catalog/operator.go
```diff
@@ -1731,6 +1732,28 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
 
 				plan.Status.Plan[i].Status = status
 
+			case configMapKind:
+				var cfg corev1.ConfigMap
+				err := json.Unmarshal([]byte(step.Resource.Manifest), &cfg)
+				if err != nil {
+					return errorwrap.Wrapf(err, "error parsing step manifest: %s", step.Resource.Name)
+				}
```
njhale, line 1740: might be nice for `StepResource` to have an unmarshal method -- just something to think about for the future

pkg/controller/operators/catalog/operator.go
```diff
@@ -1731,6 +1732,28 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
...
+
+				// Update UIDs on all CSV OwnerReferences
+				updated, err := o.getUpdatedOwnerReferences(cfg.OwnerReferences, plan.Namespace)
+				if err != nil {
+					return errorwrap.Wrapf(err, "error generating ownerrefs for service: %s", cfg.GetName())
+				}
+				cfg.SetOwnerReferences(updated)
+				cfg.SetNamespace(namespace)
```
njhale, line 1748: seems like we're repeating this a lot too.

pkg/controller/operators/catalog/step_ensurer.go
```diff
@@ -300,3 +300,26 @@ func (o *StepEnsurer) EnsureUnstructuredObject(client dynamic.ResourceInterface,
...
+	}
+
+	if !k8serrors.IsAlreadyExists(createErr) {
+		err = errorwrap.Wrapf(createErr, "error updating configmap: %s", configmap.GetName())
+		return
+	}
+
+	configmap.SetNamespace(namespace)
+	if _, updateErr := o.kubeClient.UpdateConfigMap(configmap); updateErr != nil {
+		err = errorwrap.Wrapf(updateErr, "error updating service: %s", configmap.GetName())
```
njhale, line 319: nit: "service" -> "configmap"

pkg/controller/operators/catalog/step_ensurer.go
```diff
@@ -300,3 +300,26 @@ func (o *StepEnsurer) EnsureUnstructuredObject(client dynamic.ResourceInterface,
 	status = v1alpha1.StepStatusPresent
 	return
 }
+
+// EnsureConfigMap writes the specified ConfigMap object to the cluster.
+func (o *StepEnsurer) EnsureConfigMap(namespace string, configmap *corev1.ConfigMap) (status v1alpha1.StepStatus, err error) {
```
njhale, line 305: hmm, couldn't we use the dynamic client for all resources going forward?

pkg/lib/operatorclient/configmap.go
```diff
@@ -0,0 +1,39 @@
...
+	klog.V(4).Infof("[UPDATE ConfigMap]: %s", configmap.GetName())
+	oldSa, err := c.GetConfigMap(configmap.GetNamespace(), configmap.GetName())
+	if err != nil {
+		return nil, err
+	}
+	patchBytes, err := createPatch(oldSa, configmap)
+	if err != nil {
+		return nil, fmt.Errorf("error creating patch for ConfigMap: %v", err)
+	}
+	return c.CoreV1().ConfigMaps(configmap.GetNamespace()).Patch(configmap.GetName(), types.StrategicMergePatchType, patchBytes)
```
njhale, line 38: Do we want to perform a merge patch for ConfigMap, or do a normal update? IIRC we wanted replace on operator upgrade, @ecordell thoughts?

pkg/controller/operators/catalog/operator.go
exdx: Agreed, they all use the same logic here

pkg/controller/operators/catalog/step_ensurer.go
exdx: We should probably, yes. Right now we have a `Ensure()` method for every type that OLM supports in the `StepEnsurer` but we can probably pack everything into the dynamic client logic instead. Would result in one very large function versus several smaller ones though

pkg/lib/operatorclient/configmap.go
exdx: We can do that - I was following the existing logic for secret types, which is what I figured configmaps most closely resemble 
https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/lib/operatorclient/secret.go#L38

exdx:
@njhale thanks for the review. In the interest of CI time would you be ok LGTMing and I can update the text error fields ("service" -> "configmap") in my follow up PR for additional OLM types (secrets, PSP, etc)? As well as the patch strategy if we want to change that. 

Your other points regarding code duplication and the dynamic client make a lot of sense but I'm not sure I have the time to address those this week

pkg/controller/operators/catalog/operator.go
```diff
@@ -1731,6 +1732,28 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
 
 				plan.Status.Plan[i].Status = status
 
+			case configMapKind:
```
njhale, line 1735: Won't this be caught by the `default` case now?

exdx: the default case as it stands now calls `isSupported` which checks for supported custom resources only
```
// isSupported returns true if OLM supports this type of CustomResource.
func isSupported(kind string) bool {
	switch kind {
	case PrometheusRuleKind, ServiceMonitorKind:
		return true
	default:
		return false
	}
}
```
for configmaps this would return false and the install plan would fail. I think for core types having the explicit handling we have here is useful. If we use the dynamic client in the default case we lose the merge strategy control we have in the typed client

njhale: I'm not sure this is too big a deal for the majority of cases, but it can be changed later should we decide to do so.

exdx: so we should just change this PR to add ConfigMapKind to `isSupported` and call it a day?

exdx: I still feel like there's value treating `ConfigMap` types similarly to the existing core types like secrets, service accounts, and roles in the code. leaving the default option for CR resources only

njhale:
Let's move forward with this as is -- we can follow up on some of these comments later.

/lgtm

pkg/controller/operators/catalog/operator.go
njhale: I can't immediately see why we are treating those types differently today. Since it's the precedent,  I don't think there's harm in addressing this later.

PR #1435, opened 2020-04-08
Removing delineation of APIService and CRD requirements

Bowenislandsong:
OLM treats APIService and CRD the same in the backend. The differeciation between the two concepts forces Users to understand the difference and increase deficulties implmeneting easy UX. This commit gets rid of OLM differentiating between APIService and CRD. The requirement of them are say tobe met if the GVK of those resource exists.

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/olm/operator_test.go
+++ b/pkg/controller/operators/olm/operator_test.go
@@ -761,10 +761,10 @@ func apiService(group, version, serviceName, serviceNamespace, deploymentName st
 	return apiService
 }
 
-func crd(name, version, group string) *apiextensionsv1.CustomResourceDefinition {
+func crd(kind, version, group string) *apiextensionsv1.CustomResourceDefinition {
 	return &apiextensionsv1.CustomResourceDefinition{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: name + "." + group,
+			Name: kind + "." + group,
 		},
 		Spec: apiextensionsv1.CustomResourceDefinitionSpec{
 			Group: group,
@@ -776,7 +776,7 @@ func crd(name, version, group string) *apiextensionsv1.CustomResourceDefinition 
 				},
 			},
 			Names: apiextensionsv1.CustomResourceDefinitionNames{
-				Kind: name,
+				Kind: kind,
 			},
 		},
 		Status: apiextensionsv1.CustomResourceDefinitionStatus{
@@ -3383,9 +3383,9 @@ func TestSyncOperatorGroups(t *testing.T) {
 			Message: "CSV minKubeVersion (0.0.0) less than server version (" + serverVersion + ")",
 		},
 		{
-			Group:   "apiextensions.k8s.io",
+			Group:   "fake.api.group",
 			Version: "v1",
-			Kind:    "CustomResourceDefinition",
+			Kind:    "c1",
 			Name:    crd.GetName(),
 			Status:  v1alpha1.RequirementStatusReasonPresent,
 			Message: "CRD is present and Established condition is true",
--- a/pkg/controller/operators/olm/requirements.go
+++ b/pkg/controller/operators/olm/requirements.go
@@ -5,15 +5,14 @@ import (
 	"fmt"
 	"strings"
 
-	"github.com/sirupsen/logrus"
-
 	"github.com/coreos/go-semver/semver"
+	"github.com/sirupsen/logrus"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
 	"github.com/operator-framework/api/pkg/operators/v1alpha1"
 	olmErrors "github.com/operator-framework/operator-lifecycle-manager/pkg/controller/errors"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/install"
-	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
-
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 func (a *Operator) minKubeVersionStatus(name string, minKubeVersion string) (met bool, statuses []v1alpha1.RequirementStatus) {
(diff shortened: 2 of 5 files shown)


ecordell:
The changes look good, but there are a lot of unit and e2e tests that need to be updated

Bowenislandsong:
> The changes look good, but there are a lot of unit and e2e tests that need to be updated

Yes. Doing that right now. A lot of the tests does not apply anymore. Might be just deleted.

pkg/controller/operators/olm/operator_test.go
```diff
@@ -3353,7 +3347,7 @@ func TestSyncOperatorGroups(t *testing.T) {
 		},
 	}
 	deploymentName := "csv1-dep1"
-	crd := crd("c1", "v1", "fake.api.group")
+	crd := crd("CustomResourceDefinition", "v1beta1", "apiextensions.k8s.io")
```
ecordell, line 3350: this is confusing - I assume this change is to make the test cases that check for CRDs match, but what this is actually doing is defining a CSV that owns a CRD that defines an API called CRD. Can you keep these values the same, and update all of the expected statuses to the right values instead?

pkg/controller/operators/olm/requirements.go
```diff
@@ -70,148 +69,57 @@ func (a *Operator) minKubeVersionStatus(name string, minKubeVersion string) (met
...
+			Group:   group,
+			Version: r.Version,
+			Kind:    r.Kind,
 			Name:    r.Name,
 		}
 
-		// check if CRD exists - this verifies group, version, and kind, so no need for GVK check via discovery
-		crd, err := a.lister.APIExtensionsV1beta1().CustomResourceDefinitionLister().Get(r.Name)
-		if err != nil {
+		if err := a.isGVKRegistered(group, r.Version, r.Kind); err != nil {
```
ecordell, line 94: Is there a reason we're not just building one big combined list of GVKs and checking them all at once (on line 131)?

pkg/controller/operators/olm/requirements.go
```diff
@@ -70,165 +69,65 @@ func (a *Operator) minKubeVersionStatus(name string, minKubeVersion string) (met
...
-			Name:    name,
-		}
-
-		if err := a.isGVKRegistered(r.Group, r.Version, r.Kind); err != nil {
+	for _, status := range resourceGVKNs {
+		if err := a.isGVKRegistered(status.Group, status.Version, status.Kind); err != nil {
 			status.Status = v1alpha1.RequirementStatusReasonNotPresent
-			status.Message = "Native API does not exist"
+			status.Message = "resource does not exist"
 			met = false
```
Bowenislandsong, line 125: @njhale I wonder if my understanding is wrong on the count that NativeAPIs are considered optional? It apparently still affects the requirements if the GVK of a nativeAPI is not found. Did I understand something wrong? Thanks.

pkg/controller/operators/olm/operator_test.go
```diff
@@ -760,10 +760,10 @@ func apiService(group, version, serviceName, serviceNamespace, deploymentName st
 	return apiService
```
awgreene, line 760: Your commit contains a few misspelt workds: `differeciation, deficulties, implmeneting,tobe`

pkg/controller/operators/olm/requirements.go
```diff
@@ -72,118 +71,9 @@ func (a *Operator) minKubeVersionStatus(name string, minKubeVersion string) (met
...
-		if !a.isAPIServiceAvailable(apiService) {
-			status.Status = "NotPresent"
-			met = false
-		} else {
-			status.Status = "Present"
-			status.UUID = string(apiService.GetUID())
-		}
-		statuses = append(statuses, status)
-	}
-
```
awgreene, line 186: Before - a check for the APIService was added in addition to a GVK check - could you explain why this is not necessary anymore?

pkg/controller/operators/olm/requirements_test.go
```diff
@@ -410,14 +410,14 @@ func TestRequirementAndPermissionStatus(t *testing.T) {
 						},
 					},
 				},
-				{"apiextensions.k8s.io", "v1beta1", "CustomResourceDefinition", "c1.g1"}: {
```
awgreene, line 413: Why did you need to change these values in the tests?

pkg/controller/operators/olm/requirements_test.go
```diff
@@ -485,87 +485,7 @@ func TestRequirementAndPermissionStatus(t *testing.T) {
...
-					newCRD := crd("c1", "v2", "g1")
-					// condition order: established, name accepted
-					newCRD.Status.Conditions[0].Status = v1beta1.ConditionFalse
-					newCRD.Status.Conditions[1].Status = v1beta1.ConditionTrue
-					return newCRD
-				}(),
-			},
-			met: false,
-			expectedRequirementStatuses: map[gvkn]v1alpha1.RequirementStatus{
-				{"apiextensions.k8s.io", "v1beta1", "CustomResourceDefinition", "c1.g1"}: {
```
awgreene, line 568: Could you explain why removing these tests are the right move?

pkg/controller/operators/olm/requirements_test.go
```diff
@@ -485,87 +485,7 @@ func TestRequirementAndPermissionStatus(t *testing.T) {
...
-					Version: "v1alpha1",
-					Kind:    "ClusterServiceVersion",
-					Name:    "csv1",
-					Status:  v1alpha1.RequirementStatusReasonPresent,
-				},
-			},
-			expectedError: nil,
-		},
-		{
-			description: "RequirementNotMet/NamesConflictedCRD",
```
awgreene, line 506: It seems like we'd want a test to make sure that name conflicts don't occur. Is there another test that protects against this bug?

pkg/controller/operators/olm/requirements.go
```diff
@@ -72,118 +71,9 @@ func (a *Operator) minKubeVersionStatus(name string, minKubeVersion string) (met
...
-		if err != nil {
-			status.Status = v1alpha1.RequirementStatusReasonNotPresent
-			status.Message = "CRD is not present"
-			a.logger.Debugf("Setting 'met' to false, %v with status %v, with err: %v", r.Name, status, err)
-			met = false
-			statuses = append(statuses, status)
-			continue
-		}
-
-		if crd.Spec.Version != r.Version {
```
awgreene, line 98: It looks like the version check is removed in this PR.

pkg/controller/operators/olm/requirements.go
```diff
@@ -72,118 +71,9 @@ func (a *Operator) minKubeVersionStatus(name string, minKubeVersion string) (met
...
-			served := false
-			for _, version := range crd.Spec.Versions {
-				if version.Name == r.Version {
-					if version.Served {
```
awgreene, line 102: Why was this removed?

Bowenislandsong: This tries to see if the crdDesc has the version that the CRD's api has. If we are already checking the gvk from the CRDDesc, then this is not necessary anymore, right?

pkg/controller/operators/olm/requirements.go
Bowenislandsong: in addition to checking the GVK of an APIService, this used to check is a APIService resource exist and if it is available. If the GVK of that APIService exist, shouldn't that imply that the APIService is available? If it is available, why do we care if the APISewrvice resource is there or not, since, for required APIService, we just care if the APIService is available to serve us. Right?

pkg/controller/operators/olm/requirements_test.go
Bowenislandsong: you are right, after some thought, i realize the CRD's gvk should be on the display here not the GVK of the api from that CRD.

Bowenislandsong: I was just about to change this back, but then i realized that we are trying to blur the difference between CRD and APIService. In CSVless bundle, we are interpreting all required CRD/APIService without description as a CRD. Therefore, we can no longer, at this stage, tell if that is a CRD or APIService.

pkg/controller/operators/olm/requirements.go
awgreene: I do not believe that this is true. It is possible that a single CRD contains multiple versions that are `served`. For example, imaging that you are upgrading your CRD from v1 to v2. Only one of these versions of the CRD are stored in etcd/object store. However, there are cases where I should still be able to create/update/delete v1 of the CRD.

Check out bullet point 1 in [this doc](https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning/#overview)

pkg/controller/operators/olm/requirements_test.go
awgreene: Got it - so we now defined the GVK of the resource defined in the CRD/APIService. SGTM.

pkg/controller/operators/olm/requirements.go
awgreene: > In addition to checking the GVK of an APIService, this used to check is a APIService resource exist and if it is available. I. If the GVK of that APIService exist, shouldn't that imply that the APIService is available?

I do not believe that this handles upgrades. Say that the deployment that implements the APIService is taken down and redeployed - although the GVK exists, the APIService would be unavailable during the duration of the upgrade. If the operator expects the APIService to be available, this would introduce an issue.

vendor/github.com/onsi/gomega/go.mod
```diff
@@ -1,7 +1,5 @@
 module github.com/onsi/gomega
 
-go 1.14
-
```
awgreene, line 4: This looks like noise.

pkg/controller/operators/olm/requirements.go
Bowenislandsong: Yes, but these are CRD Descriptions which are singletons. Under the same CRD, different versions have different descriptions. At the very least the descriptions would have different versions.

Bowenislandsong:
I used online rebase and somehow squashing locally won't be so easy now. I will just squash this online once it is approved and lgtmed.

Please ignore squashing commits, which will be done at merge time.

Bowenislandsong:
/hold 
Wating on https://github.com/operator-framework/operator-lifecycle-manager/pull/1472/files to merge first.

Bowenislandsong:
/hold cancel
No longer waiting on that PR. My mistake
/retest

Bowenislandsong:
We can not merge this in 4.5 at this time and therefore made https://github.com/operator-framework/operator-registry/pull/252/commits/57fc774e83bbb239a70de58c6a31dddb56004df3 to avoid this problem.
/close
This can be used in the future if needed.

PR #1439, opened 2020-04-13
Convert package manifest e2e to ginkgo test

harishsurf:
This change converts packagemanifest_e2e_test.go to leverage ginkgo's BDD approach.
The change also involves adding `GinkgoRecover` wherever go routines are called

**Reviewer Checklist**

### DIFF
--- a/test/e2e/catalog_e2e_test.go
+++ b/test/e2e/catalog_e2e_test.go
@@ -56,7 +56,7 @@ var _ = Describe("Catalog", func() {
 		crc := newCRClient()
 
 		catalogSourceName := genName("mock-ocs-")
-		_, cleanupSource := createInternalCatalogSource(GinkgoT(), c, crc, catalogSourceName, operatorNamespace, manifests, []apiextensions.CustomResourceDefinition{crd}, []v1alpha1.ClusterServiceVersion{csv})
+		_, cleanupSource := createInternalCatalogSource(c, crc, catalogSourceName, operatorNamespace, manifests, []apiextensions.CustomResourceDefinition{crd}, []v1alpha1.ClusterServiceVersion{csv})
 		defer cleanupSource()
 
 		// ensure the mock catalog exists and has been synced by the catalog operator
@@ -132,7 +132,7 @@ var _ = Describe("Catalog", func() {
 		}
 
 		// Create the initial catalog source
-		createInternalCatalogSource(GinkgoT(), c, crc, mainCatalogName, globalNS, mainManifests, []apiextensions.CustomResourceDefinition{mainCRD}, []v1alpha1.ClusterServiceVersion{mainCSV})
+		createInternalCatalogSource(c, crc, mainCatalogName, globalNS, mainManifests, []apiextensions.CustomResourceDefinition{mainCRD}, []v1alpha1.ClusterServiceVersion{mainCSV})
 
 		// Attempt to get the catalog source before creating install plan
 		_, err := fetchCatalogSourceOnStatus(crc, mainCatalogName, globalNS, catalogSourceRegistryPodSynced)
@@ -243,7 +243,7 @@ var _ = Describe("Catalog", func() {
 		}
 
 		// Create the initial catalogsource
-		createInternalCatalogSource(GinkgoT(), c, crc, mainCatalogName, testNamespace, mainManifests, nil, []v1alpha1.ClusterServiceVersion{mainCSV})
+		createInternalCatalogSource(c, crc, mainCatalogName, testNamespace, mainManifests, nil, []v1alpha1.ClusterServiceVersion{mainCSV})
 
 		// Attempt to get the catalog source before creating install plan
 		fetchedInitialCatalog, err := fetchCatalogSourceOnStatus(crc, mainCatalogName, testNamespace, catalogSourceRegistryPodSynced)
@@ -363,7 +363,7 @@ var _ = Describe("Catalog", func() {
 		}
 
 		// Create the initial catalogsource
-		_, cleanupSource := createInternalCatalogSource(GinkgoT(), c, crc, mainCatalogName, testNamespace, mainManifests, nil, []v1alpha1.ClusterServiceVersion{mainCSV})
+		_, cleanupSource := createInternalCatalogSource(c, crc, mainCatalogName, testNamespace, mainManifests, nil, []v1alpha1.ClusterServiceVersion{mainCSV})
 
 		// Attempt to get the catalog source before creating install plan
 		fetchedInitialCatalog, err := fetchCatalogSourceOnStatus(crc, mainCatalogName, testNamespace, catalogSourceRegistryPodSynced)
@@ -381,7 +381,7 @@ var _ = Describe("Catalog", func() {
 		cleanupSource()
 
 		// create a catalog with the same name
-		createInternalCatalogSource(GinkgoT(), c, crc, mainCatalogName, testNamespace, append(mainManifests, dependentManifests...), []apiextensions.CustomResourceDefinition{dependentCRD}, []v1alpha1.ClusterServiceVersion{mainCSV, dependentCSV})
+		createInternalCatalogSource(c, crc, mainCatalogName, testNamespace, append(mainManifests, dependentManifests...), []apiextensions.CustomResourceDefinition{dependentCRD}, []v1alpha1.ClusterServiceVersion{mainCSV, dependentCSV})
 
 		// Create Subscription
 		subscriptionName := genName("sub-")
@@ -465,8 +465,8 @@ var _ = Describe("Catalog", func() {
 		}
 
 		// Create ConfigMap CatalogSources
-		createInternalCatalogSource(GinkgoT(), c, crc, mainSourceName, testNamespace, append(mainManifests, dependentManifests...), []apiextensions.CustomResourceDefinition{dependentCRD}, []v1alpha1.ClusterServiceVersion{mainCSV, dependentCSV})
-		createInternalCatalogSource(GinkgoT(), c, crc, replacementSourceName, testNamespace, append(replacementManifests, dependentManifests...), []apiextensions.CustomResourceDefinition{dependentCRD}, []v1alpha1.ClusterServiceVersion{replacementCSV, mainCSV, dependentCSV})
+		createInternalCatalogSource(c, crc, mainSourceName, testNamespace, append(mainManifests, dependentManifests...), []apiextensions.CustomResourceDefinition{dependentCRD}, []v1alpha1.ClusterServiceVersion{mainCSV, dependentCSV})
+		createInternalCatalogSource(c, crc, replacementSourceName, testNamespace, append(replacementManifests, dependentManifests...), []apiextensions.CustomResourceDefinition{dependentCRD}, []v1alpha1.ClusterServiceVersion{replacementCSV, mainCSV, dependentCSV})
 
 		// Wait for ConfigMap CatalogSources to be ready
 		mainSource, err := fetchCatalogSourceOnStatus(crc, mainSourceName, testNamespace, catalogSourceRegistryPodSynced)
@@ -565,7 +565,7 @@ var _ = Describe("Catalog", func() {
 
 		c := newKubeClient()
 		crc := newCRClient()
-		_, cleanupSource := createInternalCatalogSource(GinkgoT(), c, crc, sourceName, testNamespace, manifests, []apiextensions.CustomResourceDefinition{crd}, []v1alpha1.ClusterServiceVersion{csv})
+		_, cleanupSource := createInternalCatalogSource(c, crc, sourceName, testNamespace, manifests, []apiextensions.CustomResourceDefinition{crd}, []v1alpha1.ClusterServiceVersion{csv})
 		defer cleanupSource()
 
 		// Wait for a new registry pod to be created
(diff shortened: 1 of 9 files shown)


test/e2e/catalog_e2e_test.go
```diff
@@ -242,10 +242,10 @@ var _ = Describe("Catalog", func() {
 		}
 
 		// Create the initial catalogsource
-		createInternalCatalogSource(GinkgoT(), c, crc, mainCatalogName, testNamespace, mainManifests, nil, []v1alpha1.ClusterServiceVersion{mainCSV})
+		createInternalCatalogSource(c, crc, mainCatalogName, testNamespace, mainManifests, nil, []v1alpha1.ClusterServiceVersion{mainCSV})
```
harishsurf, line 245: This method uses deprecated `internal` catalog source type. I was thinking of rewriting this method to use `configMap` catalog source type as a separate PR. Thoughts?

test/e2e/packagemanifest_e2e_test.go
```diff
@@ -2,170 +2,196 @@ package e2e
...
-
-		relatedImages := pm.Status.Channels[0].CurrentCSVDesc.RelatedImages
-		require.Equal(GinkgoT(), len(expectedRelatedImages), len(relatedImages))
+	})
 
-		for _, v := range relatedImages {
-			_, ok := expectedRelatedImages[v]
-			require.True(GinkgoT(), ok, "Expect this image %s to exist in the related images list\n", v)
-		}
+	Context("Given an operator-registry image", func() {
```
harishsurf, line 127: Not a 100% sure, if the `Context` wording is accurate. I felt a second look at it would be good.

test/e2e/packagemanifest_e2e_test.go
```diff
@@ -3,170 +3,196 @@ package e2e
...
+
+					pm, err := fetchPackageManifest(pmc, testNamespace, packageName, packageManifestHasStatus)
+					Expect(err).NotTo(HaveOccurred(), "error getting package manifest")
+					Expect(pm).ShouldNot(BeNil())
+					Expect(pm.GetName()).Should(Equal(packageName))
+
+					relatedImages := pm.Status.Channels[0].CurrentCSVDesc.RelatedImages
+					Expect(len(relatedImages)).To(Equal(len(expectedRelatedImages)))
+
+					for _, v := range relatedImages {
```
benluddy, line 189: This could be:
```
Expect(relatedImages).To(ConsistOf([]string{
  "quay.io/coreos/etcd@sha256:3816b6daf9b66d6ced6f0f966314e2d4f894982c6b1493061502f8c2bf86ac84",
  "quay.io/coreos/etcd@sha256:49d3d4a81e0d030d3f689e7167f23e120abf955f7d08dbedf3ea246485acee9f",
  "quay.io/coreos/etcd-operator@sha256:c0301e4686c3ed4206e370b42de5a3bd2229b9fb4906cf85f3f30650424abec2"
}))
```

which would cover this for loop as well as the length assertion on L187.

test/e2e/packagemanifest_e2e_test.go
```diff
@@ -3,170 +3,196 @@ package e2e
...
-		err = wait.Poll(pollInterval, 1*time.Minute, func() (bool, error) {
-			GinkgoT().Logf("Polling package-server...")
-			_, err := pmc.OperatorsV1().PackageManifests(testNamespace).List(context.TODO(), metav1.ListOptions{})
-			if err == nil {
-				return true, nil
-			}
-			return false, nil
+	Context("Given a ConfigMap containing package manifest, CSV and CRD", func() {
+		When("A CatalogSource is created using the ConfigMap as catalog source type", func() {
+			It("Package manifest loading can be verified", func() {
```
benluddy, line 43: Spec descriptions need to state the behavior that is being verified!

test/e2e/packagemanifest_e2e_test.go
```diff
@@ -3,170 +3,196 @@ package e2e
...
+				crd := newCRD(crdPlural)
+				catalogSourceName := genName("mock-ocs")
+				namedStrategy := newNginxInstallStrategy(genName("dep-"), nil, nil)
+				csv := newCSV(packageStable, testNamespace, "", semver.MustParse("0.1.0"), []apiextensions.CustomResourceDefinition{crd}, nil, namedStrategy)
+				csv.SetLabels(map[string]string{"projected": "label"})
+				csv.Spec.NativeAPIs = []metav1.GroupVersionKind{{Group: "kubenative.io", Version: "v1", Kind: "Native"}}
+				csvJSON, _ := json.Marshal(csv)
+
+				// Wait for package-server to be ready
+				err := wait.Poll(pollInterval, 1*time.Minute, func() (bool, error) {
```
benluddy, line 71: Can this use `Eventually`?

test/e2e/packagemanifest_e2e_test.go
```diff
@@ -3,170 +3,196 @@ package e2e
...
+				}
+
+				crdPlural := genName("ins")
+				crd := newCRD(crdPlural)
+				catalogSourceName := genName("mock-ocs")
+				namedStrategy := newNginxInstallStrategy(genName("dep-"), nil, nil)
+				csv := newCSV(packageStable, testNamespace, "", semver.MustParse("0.1.0"), []apiextensions.CustomResourceDefinition{crd}, nil, namedStrategy)
+				csv.SetLabels(map[string]string{"projected": "label"})
+				csv.Spec.NativeAPIs = []metav1.GroupVersionKind{{Group: "kubenative.io", Version: "v1", Kind: "Native"}}
+				csvJSON, _ := json.Marshal(csv)
```
benluddy, line 68: I'm sure this isn't new, but the test shouldn't be ignoring the error value returned from Marshal. A nice thing to add (in a separate PR) would be automation for running static analysis tools like errcheck on every change.

test/e2e/packagemanifest_e2e_test.go
```diff
@@ -3,170 +3,196 @@ package e2e
...
-
-		// Wait for package-server to be ready
-		err = wait.Poll(pollInterval, 1*time.Minute, func() (bool, error) {
-			GinkgoT().Logf("Polling package-server...")
-			_, err := pmc.OperatorsV1().PackageManifests(testNamespace).List(context.TODO(), metav1.ListOptions{})
-			if err == nil {
-				return true, nil
-			}
-			return false, nil
+	Context("Given a ConfigMap containing package manifest, CSV and CRD", func() {
```
benluddy, line 41: The ConfigMap mentioned in the container description should be created as part of this container (in a BeforeEach).

Edit: I have the same comment with a few other container nodes, please take a look at those.

test/e2e/packagemanifest_e2e_test.go
```diff
@@ -3,170 +3,196 @@ package e2e
...
+				Expect(err).ToNot(HaveOccurred(), "package-server not available")
+
+				_, cleanupCatalogSource := createInternalCatalogSource(c, crc, catalogSourceName, testNamespace, manifests, []apiextensions.CustomResourceDefinition{crd}, []v1alpha1.ClusterServiceVersion{csv})
+				Expect(err).ToNot(HaveOccurred())
+				defer cleanupCatalogSource()
+
+				_, err = fetchCatalogSource(crc, catalogSourceName, testNamespace, catalogSourceRegistryPodSynced)
+				Expect(err).ToNot(HaveOccurred())
+
+				By("querying the package manifest using the package name", func() {
```
benluddy, line 88: IMO it's clearer not to pass the option function to `By`, to reduce nesting and avoid giving the impression of a new container node. The only use case I can imagine for the optional function argument would be calling a named function that was used in multiple specs.

test/e2e/packagemanifest_e2e_test.go
```diff
@@ -3,170 +3,196 @@ package e2e
...
-		require.Equal(GinkgoT(), len(expectedRelatedImages), len(relatedImages))
+	})
 
-		for _, v := range relatedImages {
-			_, ok := expectedRelatedImages[v]
-			require.True(GinkgoT(), ok, "Expect this image %s to exist in the related images list\n", v)
-		}
+	Context("Given an operator-registry image", func() {
+		Context("When a CatalogSource is created using gRPC catalog source type", func() {
+			When("Package manifest is available", func() {
```
benluddy, line 130: I wonder if it would be clearer to change the test from:

1. poll until list package manifests returns without error
2. require that get catalog source returns without error
3. poll until get package manifest returns without error

to:

1. poll until get package manifest returns without error

It seems to me that we're checking for multiple related conditions, when a failure in any one of them would imply a failure of the primary test condition.

test/e2e/packagemanifest_e2e_test.go
benluddy: Until we update this test to build its own registry image during the test run, it makes more sense to me to merge the 
"CatalogSource created..." context into this and leave the image reference string as an implementation detail.

test/e2e/packagemanifest_e2e_test.go
```diff
@@ -3,170 +3,181 @@ package e2e
...
-				Image:      image,
-			},
-		}
 
-		expectedRelatedImages := map[string]string{
-			"quay.io/coreos/etcd@sha256:3816b6daf9b66d6ced6f0f966314e2d4f894982c6b1493061502f8c2bf86ac84":          "",
-			"quay.io/coreos/etcd@sha256:49d3d4a81e0d030d3f689e7167f23e120abf955f7d08dbedf3ea246485acee9f":          "",
-			"quay.io/coreos/etcd-operator@sha256:c0301e4686c3ed4206e370b42de5a3bd2229b9fb4906cf85f3f30650424abec2": "",
-		}
+		It("lists the CatalogSource contents using the PackageManifest API", func() {
```
benluddy, line 91: Is this two specs in one ("packagemanifest can be retrieved by name" and "packagemanifest appears in list")? If yes, I think the second one is underspecified -- it would pass if `Items` is an empty slice or if it contained a packagemanifest other than the expected one.

test/e2e/packagemanifest_e2e_test.go
```diff
@@ -3,170 +3,181 @@ package e2e
...
+		AfterEach(func() {
+			err := crc.OperatorsV1alpha1().CatalogSources(catalogSource.GetNamespace()).Delete(context.TODO(), catalogSource.GetName(), metav1.DeleteOptions{})
+			Expect(err).NotTo(HaveOccurred())
+		})
 
-		pm, err := fetchPackageManifest(GinkgoT(), pmc, testNamespace, packageName, packageManifestHasStatus)
-		require.NoError(GinkgoT(), err, "error getting package manifest")
-		require.NotNil(GinkgoT(), pm)
-		require.Equal(GinkgoT(), packageName, pm.GetName())
+		It("lists the CatalogSource contents using the PackageManifest API", func() {
```
benluddy, line 165: Is this precisely stating the behavior the test is intended to verify? It looks like a `relatedImages` test.

test/e2e/packagemanifest_e2e_test.go
harishsurf: I split them into two `It` specs

test/e2e/packagemanifest_e2e_test.go
```diff
@@ -178,18 +188,28 @@ func packageManifestHasStatus(pm *packagev1.PackageManifest) bool {
...
-	})
+	}).Should(BeTrue())
 
 	return fetched, err
 }
+
+func containsPackageManifest(pmList []packagev1.PackageManifest, pkgName string) bool {
+	for _, pm := range pmList {
+		if pm.GetName() == pkgName {
+			fmt.Printf("$$$ %v\n", pm.GetName())
```
benluddy, line 210: Was this left in by mistake?

test/e2e/packagemanifest_e2e_test.go
```diff
@@ -178,18 +188,28 @@ func packageManifestHasStatus(pm *packagev1.PackageManifest) bool {
...
 			return true, err
 		}
 		return check(fetched), nil
-	})
+	}).Should(BeTrue())
 
 	return fetched, err
 }
+
+func containsPackageManifest(pmList []packagev1.PackageManifest, pkgName string) bool {
```
benluddy, line 207: This could be expressed by composing the `ContainsElement` and `WithTransform` matchers. I suspect this is a common pattern in the tests, so maybe that's something to come back to in the future when considering reusable matchers/fixtures.

test/e2e/packagemanifest_e2e_test.go
```diff
@@ -178,18 +188,28 @@ func packageManifestHasStatus(pm *packagev1.PackageManifest) bool {
...
-		t.Logf("Polling...")
+	Eventually(func() (bool, error) {
+		ctx.Ctx().Logf("Polling...")
 		fetched, err = pmc.OperatorsV1().PackageManifests(namespace).Get(context.TODO(), name, metav1.GetOptions{})
 		if err != nil && !errors.IsNotFound(err) {
 			return true, err
 		}
 		return check(fetched), nil
-	})
+	}).Should(BeTrue())
```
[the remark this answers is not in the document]
harishsurf: Noted. I'll look into [this](https://onsi.github.io/gomega/#adding-your-own-matchers) and follow it up with another PR

Bowenislandsong:
@shawn-hurley could you help merge this? Thank you!

shawn-hurley:
Adding valid bug because of testing improvements only

Bowenislandsong:
Are we going to push this in soon?

harishsurf:
@shawn-hurley, I have rebased this PR and is ready for merge. Could you please take a look?

PR #1440, opened 2020-04-13
Add initial SAT solver package.

benluddy:
The sat package provides interfaces for expressing a problem in terms
of installables and constraints, and implements Solve by integrating
with an existing SAT solver library.

We'll eventually want to revisit how weighting works and improve performance on larger inputs, but it's more important to get end-to-end dependency resolution working first.


pkg/controller/registry/resolver/sat/dict.go
```diff
@@ -0,0 +1,268 @@
...
+		if f(x) {
+			return
+		}
+	}
+	f(-1) // omit cardinality constraint
+}
+
+func binarySearch(min, max int, f func(int) bool) {
+	for {
+		x := (min + max) / 2
```
harishsurf, line 254: could add a possible check for [overflow](https://stackoverflow.com/questions/6735259/calculating-mid-in-binary-search)?

benluddy: Sure, although if the input size is anywhere near to overflowing min+max, we'll have other problems. Good catch.

dinhxuanvu:
Waiting for a final lgtm from Evan.

pkg/controller/registry/resolver/sat/solve.go
```diff
@@ -0,0 +1,95 @@
...
+	for i, a := range e {
+		s[i] = a.String()
+	}
+	return fmt.Sprintf("%s: %s", msg, strings.Join(s, ", "))
+}
+
+const (
+	satisfiable   = 1
+	unsatisfiable = -1
+	timeout       = 0
```
ecordell, line 49: nit: is this used?

benluddy: Not anymore. OK with you to mop that up in a future PR?

pkg/controller/registry/resolver/sat/solve.go
```diff
@@ -0,0 +1,94 @@
...
+	unsatisfiable = -1
+)
+
+// Solve takes a slice containing all Installables and returns a slice
+// containing only those Installables that were selected for
+// installation. The solution will attempt to minimize the total
+// weight of all selected Installables. If no solution is possible, an
+// error is returned.
+func Solve(installables []Installable) ([]Installable, error) {
+	return SolveWithContext(context.Background(), installables)
```
exdx, line 57: nit: could this take a `context` as an argument instead? and if none use a `context.TODO()`? 

From what I recall `context.Background()` should be called only once in the program, at the beginning as the parent context

benluddy: Sure, TODO may be more appropriate for this, although I think the distinction only exists to clarify intent when reading the source (https://github.com/golang/go/blob/master/src/context/context.go#L199-L202). In production I assume SolveWithContext is always going to be used, but it felt convenient to include this contextless signature for things like tests or utilities.

benluddy:
Manually merging this to unblock dependent changes. This change should be inert, as it introduces a new package without also introducing any references to it.

PR #1443, opened 2020-04-14
Convert metrics_e2e_test to ginkgo

harishsurf:
This change converts metrics_e2e_test.go to leverage ginkgo's BDD approach

**Reviewer Checklist**

### DIFF
--- a/test/e2e/csv_e2e_test.go
+++ b/test/e2e/csv_e2e_test.go
@@ -74,11 +74,11 @@ var _ = Describe("CSV", func() {
 			},
 		}
 
-		cleanupCSV, err := createCSV(GinkgoT(), c, crc, csv, testNamespace, false, false)
+		cleanupCSV, err := createCSV(c, crc, csv, testNamespace, false, false)
 		require.NoError(GinkgoT(), err)
 		defer cleanupCSV()
 
-		_, err = fetchCSV(GinkgoT(), crc, csv.Name, testNamespace, csvPendingChecker)
+		_, err = fetchCSV(crc, csv.Name, testNamespace, csvPendingChecker)
 		require.NoError(GinkgoT(), err)
 
 		// Shouldn't create deployment
@@ -137,11 +137,11 @@ var _ = Describe("CSV", func() {
 			},
 		}
 
-		cleanupCSV, err := createCSV(GinkgoT(), c, crc, csv, testNamespace, false, false)
+		cleanupCSV, err := createCSV(c, crc, csv, testNamespace, false, false)
 		require.NoError(GinkgoT(), err)
 		defer cleanupCSV()
 
-		_, err = fetchCSV(GinkgoT(), crc, csv.Name, testNamespace, csvPendingChecker)
+		_, err = fetchCSV(crc, csv.Name, testNamespace, csvPendingChecker)
 		require.NoError(GinkgoT(), err)
 
 		// Shouldn't create deployment
@@ -247,11 +247,11 @@ var _ = Describe("CSV", func() {
 		require.NoError(GinkgoT(), err)
 		defer cleanupCRD()
 
-		cleanupCSV, err := createCSV(GinkgoT(), c, crc, csv, testNamespace, true, false)
+		cleanupCSV, err := createCSV(c, crc, csv, testNamespace, true, false)
 		require.NoError(GinkgoT(), err)
 		defer cleanupCSV()
 
-		_, err = fetchCSV(GinkgoT(), crc, csv.Name, testNamespace, csvPendingChecker)
+		_, err = fetchCSV(crc, csv.Name, testNamespace, csvPendingChecker)
 		require.NoError(GinkgoT(), err)
 
 		// Shouldn't create deployment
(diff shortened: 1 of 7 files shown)


test/e2e/ctx/ctx.go
```diff
@@ -40,6 +40,14 @@ func (ctx TestContext) Logf(f string, v ...interface{}) {
 	fmt.Fprintf(GinkgoWriter, f, v...)
 }
 
+func (ctx TestContext) Fatalf(f string, v ...interface{}) {
```
benluddy, line 43: The failure message can be generated with a format string using `fmt.Sprintf` before passing it to Fail, if necessary. Is the shorthand for generating failure messages worth making this API bigger?

Maybe you could add `Failf(string, ...interface{})` to https://github.com/operator-framework/operator-lifecycle-manager/blob/master/test/e2e/dsl/dsl.go -- or `Fail(string, ...interface{})` if we wanted to stop using dot-imports for the ginkgo DSL -- but I haven't convinced myself that it's a good suggestion.

test/e2e/metrics_e2e_test.go
```diff
@@ -4,88 +4,116 @@ package e2e
...
+			AfterEach(func() {
+				if cleanupCSV != nil {
+					cleanupCSV()
+				}
+			})
+			It("CSV will transition to a failure state with reason UnsupportedOperatorGroup", func() {
+
+				// TestMetrics tests the metrics endpoint of the OLM pod.
+				rawOutput, err := getMetricsFromPod(c, getOLMPodName(c), operatorNamespace, "8081")
+				if err != nil {
```
benluddy, line 74: `Expect(err).NotTo(HaveOccurred())` should accomplish the same thing as the old `*T.Fatalf` behavior.

test/e2e/metrics_e2e_test.go
```diff
@@ -4,88 +4,116 @@ package e2e
...
+
+				// Verify metrics have been emitted for packageserver csv
+				Expect(rawOutput).To(ContainSubstring("csv_abnormal"))
+				Expect(rawOutput).To(ContainSubstring("name=\"" + failingCSV.Name + "\""))
+				Expect(rawOutput).To(ContainSubstring("phase=\"Failed\""))
+				Expect(rawOutput).To(ContainSubstring("reason=\"UnsupportedOperatorGroup\""))
+				Expect(rawOutput).To(ContainSubstring("version=\"0.0.0\""))
+				Expect(rawOutput).To(ContainSubstring("csv_succeeded"))
+
+				ctx.Ctx().Logf(rawOutput)
```
benluddy, line 86: Is this left over from debugging? I suspect this will be logged as part of the failure message if any of the assertions fail.

test/e2e/metrics_e2e_test.go
```diff
@@ -4,88 +4,116 @@ package e2e
...
+func getOLMPodName(client operatorclient.ClientInterface) string {
 	listOptions := metav1.ListOptions{LabelSelector: "app=olm-operator"}
 	podList, err := client.KubernetesInterface().CoreV1().Pods(operatorNamespace).List(context.TODO(), listOptions)
 	if err != nil {
-		log.Infof("Error %v\n", err)
-		t.Fatalf("Listing pods failed: %v\n", err)
+		ctx.Ctx().Logf("Error %v\n", err)
+		ctx.Ctx().Fatalf("Listing pods failed: %v\n", err)
 	}
 	if len(podList.Items) != 1 {
```
benluddy, line 100: These can become gomega expectations instead of `if ... { Fatalf }`.

test/e2e/csv_e2e_test.go
```diff
@@ -4187,19 +4189,20 @@ func buildCSVCleanupFunc(t GinkgoTInterface, c operatorclient.ClientInterface, c
 			}
 		}
 
-		require.NoError(t, waitForDelete(func() error {
+		err = waitForDelete(func() error {
```
benluddy, line 4192: Can we use `Eventually` here?

test/e2e/ctx/ctx.go
harishsurf: The test was modified to exclude usage of `Fatalf` for logging altogether. I haven't added this for now but will refer to it if I come across a scenario where gomega assertions won't be of much help.

test/e2e/metrics_e2e_test.go
```diff
@@ -5,59 +5,83 @@ package e2e
...
+				_, err = fetchCSV(crc, failingCSV.Name, testNamespace, csvFailedChecker)
+				Expect(err).ToNot(HaveOccurred())
+			})
+
+			AfterEach(func() {
+				if cleanupCSV != nil {
+					cleanupCSV()
+				}
+			})
+			It("CSV will transition to a failure state with reason UnsupportedOperatorGroup", func() {
```
benluddy, line 72: Does this precisely state the desired behavior being tested? This test is verifying that the `csv_abnormal` metric is reported by OLM.

test/e2e/metrics_e2e_test.go
```diff
@@ -5,59 +5,83 @@ package e2e
...
-		// Verify metrics have been emitted for packageserver csv
-		Expect(getMetricsFromPod(c, getOLMPod(c), "8081")).To(And(
-			ContainSubstring("csv_abnormal"),
-			ContainSubstring(fmt.Sprintf("name=\"%s\"", failingCSV.Name)),
-			ContainSubstring("phase=\"Failed\""),
-			ContainSubstring("reason=\"UnsupportedOperatorGroup\""),
-			ContainSubstring("version=\"0.0.0\""),
-			ContainSubstring("csv_succeeded"),
-		))
+	Context("Given an operator group that supports all namespaces", func() {
```
benluddy, line 33: Nit: It would be nice to call out the dependency (maybe with By) on the pre-existing `testNamespace` and on the default OperatorGroup potentially set up in BeforeSuite. It's Monday morning, so it took me a moment to realize why there was no test fixture setup in this context.

PR #1444, opened 2020-04-14
fix: use busybox csv for catalog polling e2e test

exdx:
This test fixes the catalog polling test that seems to flake very badly in kube 1.18. 

The operator being deployed in the test is now a simple busybox csv instead of the whole etcd 0.9.0 operator. Tags were updated on quay and this PR reflects the code changes required. Test now passes locally consistently. 

**Reviewer Checklist**

### DIFF
--- a/test/e2e/catalog_e2e_test.go
+++ b/test/e2e/catalog_e2e_test.go
@@ -774,8 +774,8 @@ var _ = Describe("Catalog", func() {
 		defer cleaner.NotifyTestComplete(GinkgoT(), true)
 
 		sourceName := genName("catalog-")
-		packageName := "etcd"
-		channelName := "clusterwide-alpha"
+		packageName := "busybox"
+		channelName := "alpha"
 
 		// Create gRPC CatalogSource using an external registry image and poll interval
 		var image string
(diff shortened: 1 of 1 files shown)


test/e2e/catalog_e2e_test.go
```diff
@@ -912,7 +912,7 @@ var _ = Describe("Catalog", func() {
 		require.NoError(GinkgoT(), err, "error awaiting registry pod")
 
 		subChecker := func(sub *v1alpha1.Subscription) bool {
-			return sub.Status.InstalledCSV == "etcdoperator.v0.9.2-clusterwide"
+			return sub.Status.State == v1alpha1.SubscriptionStateAtLatest
```
ecordell, line 915: I think this still needs to check installed CSV. AtLatest just means that we're at the latest version we know about, it doesn't verify that OLM has seen the newest version in the catalog

exdx: we do the check further down at the end of the test still right?

exdx: https://github.com/operator-framework/operator-lifecycle-manager/pull/1444/files#diff-c8a24dc59e4efc3ed8ec9b48db6c5559R933

exdx: we could change that to "nginx.v2.0.0" though if i think that's what you're suggesting

ecordell: Yes, but there's no polling on that check. That means this could flake and fail (because OLM hasn't seen the updated catalog yet, but the subscription is at latest, and we check the csv version and fail because it's old)

exdx: ok gotcha, will update 👍

PR #1445, opened 2020-04-15
chore(gh-actions): install conntrack before minikube

njhale:
Get minikube action working by installing its missing conntrack dependency.

### DIFF
--- a/.github/workflows/test-scripts.yml
+++ b/.github/workflows/test-scripts.yml
@@ -4,6 +4,7 @@ on:
     paths:
     - 'scripts/**'
     - 'Makefile'
+    - '.github/workflows/test-scripts.yml'
   schedule:
   - cron: '0 0 * * *' # daily to pick up releases
 jobs:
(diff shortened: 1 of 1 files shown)


awgreene:
`level=error msg="Error: error updating S3 Bucket (terraform-20200415132127375000000001) tags: error listing resource tags (terraform-20200415132127375000000001): NoSuchBucket: The specified bucket does not exist"`
/retest

.github/workflows/test-scripts.yml
```diff
@@ -12,6 +12,7 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - run: |
+        sudo apt-get install conntrack
```
benluddy, line 15: Does this get stuck on interactive prompts without `-y`?

njhale: An example I saw didn't, but to be safe I updated the script to fire when it's changed so we can know for sure.

benluddy: Might be covered by the runner VM: https://github.com/actions/virtual-environments/commit/b1026148e82a01ca6f1a9e3157ad806ec825df46

njhale:
Superseded by #1452

PR #1446, opened 2020-04-15
chore(docker): add kubebuilder to base image

njhale:
Add kubebuilder to base image to make it available for tests in openshift-ci.

Blocks #1431


ecordell:
This change is ci-infra only, force merging since there are issues with e2e-gcp right now.

PR #1447, opened 2020-04-15
feat: support secrets from bundle

exdx:
PR adds support for OLM generating secrets from a user-bundle. 

**Reviewer Checklist**


pkg/controller/registry/resolver/steps.go
```diff
@@ -71,6 +75,13 @@ func NewStepResourceFromObject(obj runtime.Object, catalogSourceName, catalogSou
 		CatalogSourceNamespace: catalogSourceNamespace,
 	}
 
+	// treat secret objects with a special case
```
awgreene, line 78: Super nit: capitalize `treat`

pkg/controller/operators/catalog/step_ensurer.go
```diff
@@ -109,6 +109,29 @@ func (o *StepEnsurer) EnsureSecret(operatorNamespace, planNamespace, name string
...
+		return
+	}
+
+	if !k8serrors.IsAlreadyExists(createErr) {
+		err = errorwrap.Wrapf(createErr, "error updating secret: %s", secret.GetName())
+		return
+	}
+
+	secret.SetNamespace(namespace)
+	if _, updateErr := o.kubeClient.UpdateSecret(secret); updateErr != nil {
```
awgreene, line 126: This will cause any annotations/changes applied to the secret to be lost - unsure if that matter but it's worth calling out.

pkg/api/apis/operators/installplan_types.go
```diff
@@ -354,6 +354,9 @@ type StepResource struct {
 	Kind                   string
 	Name                   string
 	Manifest               string
+	// BundleSecret is a one-off flag for handling secrets from a user bundle versus from the catalog source.
+	// Longer term StepResources will be refactored.
+	BundleSecret           bool
```
awgreene, line 359: Can we call out that this field is handled internally by OLM and should not be exposed by the API?

pkg/controller/operators/catalog/step_ensurer.go
awgreene: Also - does this work without the `ResourceID`?

exdx: That's a good point -- this is existing code that hasn't been changed in this PR though. I can make a note however.

pkg/api/apis/operators/installplan_types.go
exdx: Sure!

pkg/controller/operators/catalog/step_ensurer.go
awgreene: Could we make a unit test instead of a note?

exdx:
PR is ready to go in but with automatic rebasing turned off will rebase manually.

ecordell:
Passed upstream e2e

PR #1448, opened 2020-04-15
feat(resolver): Add sat resolver for version dependencies

kevinrizza:

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/registry/resolver/operators.go
+++ b/pkg/controller/registry/resolver/operators.go
@@ -26,6 +26,14 @@ func (k *CatalogKey) String() string {
 	return fmt.Sprintf("%s/%s", k.Name, k.Namespace)
 }
 
+func (k *CatalogKey) IsEmpty() bool {
+	return k.Name == "" && k.Namespace == ""
+}
+
+func (k *CatalogKey) IsEqual(compare CatalogKey) bool {
+	return k.Name == compare.Name && k.Namespace == compare.Namespace
+}
+
 type APISet map[opregistry.APIKey]struct{}
 
 func EmptyAPISet() APISet {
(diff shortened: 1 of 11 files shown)


pkg/controller/registry/resolver/operators.go
```diff
@@ -225,16 +233,25 @@ type OperatorSurface interface {
...
+	replaces            string
+	providedAPIs        APISet
+	requiredAPIs        APISet
+	version             *semver.Version
+	bundle              *api.Bundle
+	sourceInfo          *OperatorSourceInfo
+	versionDependencies []VersionDependency
+}
+
+type VersionDependency struct {
```
dinhxuanvu, line 250: nit: Would be nice to name it `PackageDependency` so it will match the same type in operator-registry.

pkg/controller/registry/resolver/operators.go
```diff
@@ -225,16 +233,25 @@ type OperatorSurface interface {
...
-	bundle       *api.Bundle
-	sourceInfo   *OperatorSourceInfo
+	name                string
+	replaces            string
+	providedAPIs        APISet
+	requiredAPIs        APISet
+	version             *semver.Version
+	bundle              *api.Bundle
+	sourceInfo          *OperatorSourceInfo
+	versionDependencies []VersionDependency
```
dinhxuanvu, line 247: Does this field get populated anywhere? It seems you can only the list of bundles from cache. Then you will need to extract the dependencies information from the bundle (`api.Bundle`).

pkg/controller/registry/resolver/operators.go
```diff
@@ -225,16 +233,25 @@ type OperatorSurface interface {
...
+}
+
+type VersionDependency struct {
+	Package string
+	Version semver.Version
```
dinhxuanvu, line 252: Are we supposed to handle the `semver.Range` as well given users can specify version of a package as a semver range.

pkg/controller/registry/resolver/satresolver.go
```diff
@@ -0,0 +1,327 @@
...
+		bundleInstallable := NewBundleInstallable(csvName, bundle.bundle.ChannelName, bundleSource.Catalog)
+
+		bundleWeight := weight
+		if bundleSource.Catalog != preferredCatalog {
+			bundleWeight++
+		}
+		bundleInstallable.AddWeight(bundleWeight)
+
+		for _, depVersion := range bundle.VersionDependencies() {
+			depCandidates, err := namespacedCache.GetPackageVersionFromAllCatalogs(depVersion.Package, depVersion.Version)
```
dinhxuanvu, line 209: This only works with a specific version but won't work with a range.

dinhxuanvu:
Closed. See #1466

Issue #1380, opened 2020-03-16
TestCatalogImageUpdate e2e test does not work locally on kind clusters


exdx:
## Bug Report

**What did you do?**
A clear and concise description of the steps you took (or insert a code snippet).
The `TestCatalogImageUpdate` test does not work on KIND clusters locally (it does work on minikube however). This seems due to the port-forwarding that is required to talk to the local docker registry that is deployed during the course of this test. 

[Test Here](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/test/e2e/catalog_e2e_test.go#L706)

Fixing this quickly would require a check to see if we are in a kind cluster and then a `t.Skip()` if so. Fixing this properly would require some thought and more diagnosis. 

**What did you expect to see?**
A clear and concise description of what you expected to happen (or insert a code snippet).
Test passes on KIND clusters 

**What did you see instead? Under which circumstances?**
A clear and concise description of what you expected to happen (or insert a code snippet).
Test fails on KIND clusters

**Environment**
* operator-lifecycle-manager version:

* Kubernetes version information:

* Kubernetes cluster kind: 

**Possible Solution**

**Additional context**

ecordell:
The test is still skipped - not sure we should close this until it can be enabled

timflannagan:
I did a quick github/local search and wasn't immediately able to find the test name in question - it looks like it was renamed to `It("delete gRPC registry pod triggers recreation")` in https://github.com/operator-framework/operator-lifecycle-manager/commit/06e84d94d38a63b770b43930d6daf4343e8f9b61 if I'm following everything correctly. Am I missing anything, or is this a problem on another already released, yet still supported branch?

PR #1451, opened 2020-04-16
Added Github workflow for running e2e tests

harishsurf:
Added a [workflow](https://help.github.com/en/actions/configuring-and-managing-workflows/configuring-a-workflow) yml file that triggers a Github Action which runs our e2e-tests using [kind](https://kind.sigs.k8s.io/docs/user/quick-start/) cluster for every PR created against this repo.  
This PR also adds ginkgo's `Skip()` as a workaround for a failing spec in catalog_e2e. See #1380 for more details

**Reviewer Checklist**


.github/workflows/e2e-tests.yml
```diff
@@ -0,0 +1,12 @@
+# This workflow runs e2e tests as a Github Action for every pull request created against this repo
+name: e2e-tests
+on:
+  pull_request:
+  schedule:
```
benluddy, line 5: I don't see a benefit to running these nightly. Barring flakes, the results should be the same. It might be good to trigger this on pushes to master as well as on PR.

.github/workflows/e2e-tests.yml
```diff
@@ -0,0 +1,12 @@
...
+    - cron: '0 0 * * *' # daily to pick up releases
+jobs:
+  run-e2e-kind:
```
benluddy, line 8: Nit, but my preference is not to put "kind" in the job name. The fact that this makefile target happens to be using kind clusters is an implementation detail.

.github/workflows/e2e-tests.yml
```diff
@@ -0,0 +1,12 @@
+# This workflow runs e2e tests as a Github Action for every pull request created against this repo
```
benluddy, line 1: Is this comment providing any information that isn't clear from a quick scan of this file? It's technically already out of sync with the workflow config, since it runs both on PR and nightly.

harishsurf: you are right. I have removed the comment

ecordell:
>   error copying old registry file: failed to exec []string{"skopeo", "--debug", "--insecure-policy=true", "copy", "--dest-tls-verify=false", "--format=v2s2", "--dest-no-creds=true", "docker://quay.io/olmtest/catsrc-update-test:old", "docker://localhost:5000/catsrc-update:xt2t79"}: exec: "skopeo": executable file not found in $PATH

looks like we need to install skopeo in the test vm

benluddy:
> looks like we need to install skopeo in the test vm

IIRC there is a bit more work required than that. We're tracking it here: https://github.com/operator-framework/operator-lifecycle-manager/issues/1380 -- I've asked Harish to temporarily mark it as skipped when it runs as part of this job so that we can get some utility from this right away.

ecordell:
/lgtm
/approve

infra-only, upstream-only, merging without full checks

PR #1454, opened 2020-04-17
Regenerate the chart embedded in the test installer as needed.

benluddy:
For now, this should clear up any problems that can arise from testing
with a stale chart.


benluddy:
This change only affects upstream tests at the moment, so I'm fast-tracking it to unblock several other PRs.

PR #1455, opened 2020-04-17
fix: add empty go file to packge to fix make verify

exdx:

**Reviewer Checklist**


test/e2e/assets/chart/empty.go
```diff
@@ -0,0 +1 @@
+package chart
```
njhale, line 1: nit: a `doc.go` file with comments that explain the purpose of the package might be nicer.

PR #1458, opened 2020-04-17
Add an operator cache for use by the resolver.

benluddy:
Depends on https://github.com/operator-framework/operator-lifecycle-manager/pull/1462 for the Kube 1.18 and operator-registry dependency bumps.

### DIFF
--- a/pkg/controller/registry/resolver/cache.go
+++ b/pkg/controller/registry/resolver/cache.go
@@ -1,0 +1,308 @@
+package resolver
+
+import (
+	"context"
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/blang/semver"
+	"github.com/sirupsen/logrus"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/clientset/versioned"
+	"github.com/operator-framework/operator-registry/pkg/client"
+)
+
+type ClientProvider interface {
+	Get() (client.Interface, error)
+}
+
+type LazyClient struct {
+	address string
+	c       client.Interface
+	m       sync.Mutex
+}
+
+func (lc *LazyClient) Get() (client.Interface, error) {
+	lc.m.Lock()
+	defer lc.m.Unlock()
+	if lc.c != nil {
+		return lc.c, nil
+	}
+	client, err := client.NewClient(lc.address)
+	if err != nil {
+		return nil, err
+	}
+	lc.c = client
+	return lc.c, nil
+}
+
+type RegistryClientProvider interface {
+	ClientsForNamespaces(namespaces ...string) map[CatalogKey]ClientProvider
+}
+
+type DefaultRegistryClientProvider struct {
+	logger logrus.FieldLogger
+	c      versioned.Interface
+}
+
+func NewDefaultRegistryClientProvider(c versioned.Interface) *DefaultRegistryClientProvider {
+	return &DefaultRegistryClientProvider{
+		logger: logrus.New(),
+		c:      c,
+	}
+}
+
+func (rcp *DefaultRegistryClientProvider) ClientsForNamespaces(namespaces ...string) map[CatalogKey]ClientProvider {
+	result := make(map[CatalogKey]ClientProvider)
+	for _, namespace := range namespaces {
+		list, err := rcp.c.OperatorsV1alpha1().CatalogSources(namespace).List(context.TODO(), metav1.ListOptions{})
+		if err != nil {
+			rcp.logger.Errorf("failed to list catalogsources in %s: %s", namespace, err.Error())
+			continue
+		}
+		for _, source := range list.Items {
+			if source.Status.RegistryServiceStatus == nil {
+				continue
+			}
+			key := CatalogKey{
+				Namespace: source.Namespace,
+				Name:      source.Name,
+			}
+			result[key] = &LazyClient{address: source.Address()}
+		}
+	}
+	return result
+}
+
+type CatalogDependencyCache interface {
+	GetCSVNameFromCatalog(csvName string, catalog CatalogKey) (Operator, error)
+	GetCSVNameFromAllCatalogs(csvName string) ([]Operator, error)
+	GetPackageFromAllCatalogs(pkg string) ([]Operator, error)
+	GetPackageVersionFromAllCatalogs(pkg string, version semver.Version) ([]Operator, error)
+}
+
+type OperatorCache struct {
+	logger    logrus.FieldLogger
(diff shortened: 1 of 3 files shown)


pkg/controller/registry/resolver/cache.go
```diff
@@ -0,0 +1,287 @@
...
+
+	"github.com/blang/semver"
+	"github.com/sirupsen/logrus"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/clientset/versioned"
+	"github.com/operator-framework/operator-registry/pkg/client"
+)
+
+type ClientProvider interface {
```
ecordell, line 17: This looks good, but I'm curious why we couldn't use the sourcemanager for this? The long-lived grpc connections to the catalogs are maintained there.

pkg/controller/registry/resolver/cache.go
```diff
@@ -0,0 +1,287 @@
...
+func (c *OperatorCache) Namespaced(namespaces ...string) *NamespacedOperatorCache {
+	now := time.Now()
+	clients := c.rcp.ClientsForNamespaces(namespaces...)
+
+	result := NamespacedOperatorCache{
+		snapshots: make(map[CatalogKey]*CatalogSnapshot),
+	}
+
+	var misses []CatalogKey
+	c.m.RLock()
```
ecordell, line 120: nit: I personally like using a nested function context for this. makes it harder to mess up in refactoring.

```go
func() {
  lock
  defer unlock
  doStuff 
}()
```

pkg/controller/registry/resolver/cache.go
```diff
@@ -0,0 +1,287 @@
...
+	for _, miss := range misses {
+		s := CatalogSnapshot{
+			logger: c.logger.WithField("catalog", miss),
+			key:    miss,
+			expiry: now.Add(c.ttl),
+		}
+		s.m.Lock()
+		c.snapshots[miss] = &s
+		result.snapshots[miss] = &s
+		go c.populate(&s, clients[miss])
```
ecordell, line 168: self: check that we're tracking in-flight populates

pkg/controller/registry/resolver/cache.go
```diff
@@ -0,0 +1,287 @@
...
+	expiry    time.Time
+	operators []Operator
+	m         sync.RWMutex
+}
+
+func (s *CatalogSnapshot) Expired(at time.Time) bool {
+	return !at.Before(s.expiry)
+}
+
+func (s *CatalogSnapshot) Find(p func(*Operator) bool) []Operator {
```
ecordell, line 220: naming suggestion: `Find(p PredicateFunc)`

pkg/controller/registry/resolver/cache.go
```diff
@@ -0,0 +1,287 @@
...
+	}
+
+	snapshot.operators = operators
+}
+
+type CatalogSnapshot struct {
+	logger    logrus.FieldLogger
+	key       CatalogKey
+	expiry    time.Time
+	operators []Operator
```
ecordell, line 212: I'm assuming this is just for simplicity now - we likely don't want to store the entire definition in the cache, just the subset required for resolution.

pkg/controller/registry/resolver/cache_test.go
```diff
@@ -0,0 +1,289 @@
...
+	return nil
+}
+
+type RegistryClientProviderStub map[CatalogKey]ClientProvider
+
+func (s RegistryClientProviderStub) ClientsForNamespaces(namespaces ...string) map[CatalogKey]ClientProvider {
+	return map[CatalogKey]ClientProvider(s)
+}
+
+func TestOperatorCacheConcurrency(t *testing.T) {
```
ecordell, line 74: Have you considered marking this a Benchmark test?

pkg/controller/registry/resolver/cache.go
```diff
@@ -0,0 +1,287 @@
...
+	m         sync.RWMutex
+}
+
+func NewOperatorCache(rcp RegistryClientProvider) *OperatorCache {
+	return &OperatorCache{
+		logger:    logrus.New(),
+		rcp:       rcp,
+		snapshots: make(map[CatalogKey]*CatalogSnapshot),
+		ttl:       5 * time.Minute,
+		sem:       make(chan struct{}, 4),
```
ecordell, line 101: nit: a constant for MaxConcurrentSnapshotUpdates would be nice here

pkg/controller/registry/resolver/cache.go
benluddy: Do you mean `SourceStore`? I originally wrote the interface `RegistryClientProvider` to accept that, but I added this implementation (and renamed `AsClients` to `ClientsForNamespaces`) because I wasn't sure we wanted the extra persistent connections and CatalogSource informer in the OLM operator. Happy to go either way on that, it doesn't make much difference to the cache itself.

pkg/controller/registry/resolver/cache_test.go
benluddy: No, why? I'd be worried that it would never be run (as far as I know no one regularly runs tests with `-bench` in this repo) and the actual timings aren't important. I primarily wanted to see a reasonable number of concurrent cache reads and writes working properly with the race detector enabled.

pkg/controller/registry/resolver/cache.go
benluddy: Yes, I'll include that on the list of follow-up items to address once the resolver is integrated end-to-end.

pkg/controller/registry/resolver/cache.go
benluddy: :+1:. I'd like to stuff this into a subpackage a bit later and provide nicer knobs for configuring the cache.

pkg/controller/registry/resolver/cache.go
benluddy: Thanks, I had forgotten to add a timeout to the RPC in populate and cancellation on snapshot eviction. Updating that now.

pkg/controller/registry/resolver/cache.go
ecordell: Maybe I’m confused about where this will live - won’t this cache be in the catalog operator to be close to the resolver?

It can wait for a follow up but I think the source store would make sense - I’m wary of having two abstractions for the same thing.

benluddy: Oh, yes, you're right. I got crossed up somehow on Friday and added this implementation unnecessarily. We can absolutely change the interface back to match the SourceStore method in a follow-up.

PR #1460, opened 2020-04-17
Fix failing E2E bundle unpacking test.

benluddy:
The local E2E image didn't contain /bin/cp, which bundle unpacking
requires.


e2e.Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM scratch
+FROM busybox # ./pkg/controller/bundle/bundle_unpacker.go requires "/bin/cp"
```
dinhxuanvu, line 1: Look like the comment is not playing nice. Maybe move to to before FROM line

PR #1466, opened 2020-04-20
[WIP] feat(resolver): Add new resolver for dependency resolution

dinhxuanvu:
Add a commit to extract dependencies info from bundle and add to Operator object.

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/registry/resolver/cache.go
+++ b/pkg/controller/registry/resolver/cache.go
@@ -81,7 +81,7 @@ type CatalogDependencyCache interface {
 	GetCSVNameFromCatalog(csvName string, catalog CatalogKey) (Operator, error)
 	GetCSVNameFromAllCatalogs(csvName string) ([]Operator, error)
 	GetPackageFromAllCatalogs(pkg string) ([]Operator, error)
-	GetPackageVersionFromAllCatalogs(pkg string, version semver.Version) ([]Operator, error)
+	GetPackageVersionFromAllCatalogs(pkg string, inRange semver.Range) ([]Operator, error)
 	GetPackageChannelFromCatalog(pkg, channel string, catalog CatalogKey) ([]Operator, error)
 	GetRequiredAPIFromAllCatalogs(requiredAPI registry.APIKey) ([]Operator, error)
 	GetChannelCSVNameFromCatalog(csvName, channel string, catalog CatalogKey) (Operator, error)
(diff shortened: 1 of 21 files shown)


dinhxuanvu:
Add another commit on top of this PR #1448

pkg/controller/registry/resolver/cache.go
```diff
@@ -306,3 +337,60 @@ func (n *NamespacedOperatorCache) GetPackageVersionFromAllCatalogs(pkg string, v
...
+		return o.Bundle().GetChannelName() == channel && o.Package() == pkg
+	})
+	if len(result) == 0 {
+		return nil, fmt.Errorf("operator %s not found in channel %s in catalog %s", pkg, channel, catalog)
+	}
+
+	return result, nil
+}
+
+func (n *NamespacedOperatorCache) GetCacheCatalogSize() int {
```
benluddy, line 394: The name of this method is a little ambiguous. It sounds like it could be referring to the number of entries within a catalog, but it's returning the number of catalogs. Maybe `GetCatalogCount`?

pkg/controller/registry/resolver/operators.go
```diff
@@ -331,6 +342,10 @@ func NewOperatorFromBundle(bundle *api.Bundle, startingCSV string, sourceKey Cat
 		}
 	}
 
+	if len(depList) < 1 {
```
benluddy, line 345: You can append to a nil slice, so instead of doing this it could be:
```
var depList []VersionDependency
...
depList = append(depList, vd)
```
and depList will be `nil` if `append` is never called.

pkg/controller/registry/resolver/operators.go
```diff
@@ -285,13 +302,58 @@ func NewOperatorFromBundle(bundle *api.Bundle, startingCSV string, sourceKey Cat
...
+		case registry.GVKType:
+			gvkDep := registry.GVKDependency{}
+			err := json.Unmarshal([]byte(v.GetValue()), &gvkDep)
+			// Ignore if can't parse JSON string
+			if err != nil {
+				continue
+			}
+			// Add to requiredAPIs list
+			// Set Plural to be empty string
+			required[registry.APIKey{Group: gvkDep.Group, Kind: gvkDep.Kind, Version: gvkDep.Version}] = struct{}{}
```
benluddy, line 318: This may be a breaking change, plural is not unused elsewhere.

pkg/controller/registry/resolver/cache.go
dinhxuanvu: I think this is something we can address at a later PR.

pkg/controller/registry/resolver/operators.go
dinhxuanvu: I don't think so given there shouldn't be different APIs with same GVK but different plural forms. We can enable the new resolver and see if any of the existing e2e test cases fail due to this change.

dinhxuanvu:
This PR will be on hold until the next release (4.6).

pkg/controller/registry/resolver/satresolver.go
```diff
@@ -0,0 +1,311 @@
...
+	for _, sub := range subs {
+		pkg := sub.Spec.Package
+		filter := installableFilter{
+			channel: sub.Spec.Channel,
+			catalog: CatalogKey{
+				Name:      sub.Spec.CatalogSource,
+				Namespace: sub.Spec.CatalogSourceNamespace,
+			},
+			startingCSV: sub.Spec.StartingCSV,
+			currentCSV:  sub.Status.CurrentCSV,
```
ecordell, line 50: Note to not forget - the thing we probably want here is `installedCSV`, not `currentCSV`

PR #1470, opened 2020-04-21
Bug 1825330: support creating v1beta CRDs to avoid data loss

exdx:
Supporting v1 CRDs originally intended to create and update all CRDs at v1 - converting those provided at v1beta1. However, due to some issues around the validation, and the fact that CRDs are not fully backward-compatible, OLM will have to create v1beta1 CRDs with the v1beta1 client. This PR adds back support for creating v1beta1 CRDs. 

**Reviewer Checklist**


njhale:
It would be nice to have a test that reproduces the issue so we can be certain of the cause and catch it in the future.

njhale:
Putting this here for record keeping purposes: https://github.com/kubernetes/kubernetes/issues/87231

sttts:
/hold

Here is so much wrong about conversion.

pkg/lib/crd/serialize.go
```diff
@@ -51,18 +51,64 @@ func Serialize(manifest string) (*apiextensionsv1.CustomResourceDefinition, erro
...
+// Compatible with v1beta1 or v1 CRD manifests.
+func SerializeV1Beta1(manifest string) (*apiextensionsv1beta1.CustomResourceDefinition, error) {
+	u := &unstructured.Unstructured{}
+	reader := bytes.NewReader([]byte(manifest))
+	decoder := yaml.NewYAMLOrJSONDecoder(reader, 30)
+	if err := decoder.Decode(u); err != nil {
+		return nil, fmt.Errorf("crd unmarshaling failed: %s", err)
+	}
+
+	unversioned := &apiextensions.CustomResourceDefinition{}
```
sttts, line 78: never ever use internal apiserver types.

pkg/lib/crd/serialize.go
```diff
@@ -51,18 +51,64 @@ func Serialize(manifest string) (*apiextensionsv1.CustomResourceDefinition, erro
...
+	if err := scheme.Convert(u, unversioned, nil); err != nil {
+		return nil, fmt.Errorf("failed to convert crd from unstructured to internal: %s\nto v1: %s", u, err)
+	}
+
+	crd := &apiextensionsv1beta1.CustomResourceDefinition{}
+	if err := scheme.Convert(unversioned, crd, nil); err != nil {
```
sttts, line 84: Why don't you decode directly into v1beta1 types? Why conversion?

pkg/lib/crd/serialize.go
```diff
@@ -51,18 +51,64 @@ func Serialize(manifest string) (*apiextensionsv1.CustomResourceDefinition, erro
...
-
+	crd.TypeMeta.APIVersion = Group + V1Version
 
 	// for each version in the CRD, check and make sure there is a schema
 	// if not a schema, give a default schema of props
 	for i := range crd.Spec.Versions {
 		if crd.Spec.Versions[i].Schema == nil {
 			schema := &apiextensionsv1.JSONSchemaProps{Type: "object"}
-			crd.Spec.Versions[i].Schema = &apiextensionsv1.CustomResourceValidation{OpenAPIV3Schema:schema}
+			crd.Spec.Versions[i].Schema = &apiextensionsv1.CustomResourceValidation{OpenAPIV3Schema: schema}
```
sttts, line 61: the schema will prune everything. Is that intentional?

sttts: Instead refuse these CRDs. They are wrong. Don't try to be clever to fix them up.

exdx:
Closing this in favor of adding a v1beta1 codepath (something we had originally done in prior versions of the v1 support PR). There will be no client-side conversion between v1beta1 and v1 CRD types in OLM.

test/e2e/installplan_e2e_test.go
```diff
@@ -2315,8 +2295,8 @@ var _ = Describe("Install Plan", func() {
 
 		crdPlural := genName("ins")
 		crdName := crdPlural + ".cluster.com"
-		var min float64 = 2.1
```
ecordell, line 2318: Doesn't this change remove the important part of the test? Verifying that we don't hit the float -> int conversion issue?

test/e2e/crd_e2e_test.go
```diff
@@ -0,0 +1,194 @@
...
+		}()
+
+		mainPackageName := genName("nginx-update2-")
+		mainPackageStable := fmt.Sprintf("%s-stable", mainPackageName)
+		stableChannel := "stable"
+		mainNamedStrategy := newNginxInstallStrategy(genName("dep-"), nil, nil)
+
+		crdPlural := genName("ins-v1beta1-")
+		crdName := crdPlural + ".cluster.com"
+		mainCRD := apiextensions.CustomResourceDefinition{
```
ecordell, line 111: shouldn't this be a v1beta1 instead of internal?

test/e2e/crd_e2e_test.go
```diff
@@ -0,0 +1,194 @@
...
+		// Attempt to get the catalog source before creating install plan
+		_, err = fetchCatalogSource(GinkgoT(), crc, mainCatalogName, testNamespace, catalogSourceRegistryPodSynced)
+		require.NoError(GinkgoT(), err)
+
+		subscriptionName := genName("sub-nginx-update2-")
+		subscriptionCleanup := createSubscriptionForCatalog(GinkgoT(), crc, testNamespace, subscriptionName, mainCatalogName, mainPackageName, stableChannel, "", operatorsv1alpha1.ApprovalAutomatic)
+		defer subscriptionCleanup()
+
+		subscription, err := fetchSubscription(GinkgoT(), crc, testNamespace, subscriptionName, subscriptionHasInstallPlanChecker)
+		require.NoError(GinkgoT(), err)
```
ecordell, line 170: if this is a new test, shouldn't we write it with gomega assertions?

test/e2e/crd_e2e_test.go
```diff
@@ -0,0 +1,194 @@
...
+		require.NotNil(GinkgoT(), subscription)
+		require.NotNil(GinkgoT(), subscription.Status.InstallPlanRef)
+		require.Equal(GinkgoT(), mainCSV.GetName(), subscription.Status.CurrentCSV)
+
+		installPlanName := subscription.Status.InstallPlanRef.Name
+
+		// Wait for InstallPlan to be status: Complete before checking resource presence
+		fetchedInstallPlan, err := fetchInstallPlan(GinkgoT(), crc, installPlanName, buildInstallPlanPhaseCheckFunc(operatorsv1alpha1.InstallPlanPhaseComplete))
+		require.NoError(GinkgoT(), err)
+		GinkgoT().Logf("Install plan %s fetched with status %s", fetchedInstallPlan.GetName(), fetchedInstallPlan.Status.Phase)
```
ecordell, line 180: You can write directly with `fmt.Fprintf(GinkgoWriter, "log: %v\n", val)`

test/e2e/crd_e2e_test.go
```diff
@@ -0,0 +1,194 @@
...
+		require.Equal(GinkgoT(), operatorsv1alpha1.InstallPlanPhaseComplete, fetchedInstallPlan.Status.Phase)
+	})
```
ecordell, line 182: I was expecting an assertion that the v1beta1 CRD on the cluster was updated?

test/e2e/crd_e2e_test.go
```diff
@@ -0,0 +1,194 @@
...
+					Singular: crdName,
+					Plural:   crdNamePlural,
+				},
+			},
+		}
+
+		err = catalog.EnsureV1CRDVersions(oldCRDConvertedToV1, newCRD)
+		Expect(err).ToNot(HaveOccurred())
+	})
+	It("creates v1beta1 crds with a v1beta1 schema successfully", func() {
```
ecordell, line 94: Shouldn't there be an equivalent test for v1 CRDs?

test/e2e/crd_e2e_test.go
benluddy: There's also `(*ctx.TestContext) Logf()`, which does what you suggest.

pkg/lib/crd/version.go
```diff
@@ -23,24 +25,33 @@ var supportedCRDVersions = map[string]struct{}{
...
+	}
+
 	dec := yaml.NewYAMLOrJSONDecoder(strings.NewReader(*manifest), 10)
 	unst := &unstructured.Unstructured{}
 	if err := dec.Decode(unst); err != nil {
 		return "", err
 	}
 
 	v := unst.GetObjectKind().GroupVersionKind().Version
+	// some e2e test fixtures do not provide an version in their typemeta
```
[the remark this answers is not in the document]
exdx: I agree. This was done as a hack to get the e2e to pass. I don't know where all of the fixtures are being generated but if someone wants to help me update them we can do that instead.

pkg/controller/operators/catalog/step.go
```diff
@@ -98,61 +111,161 @@ func (b *builder) NewCRDStep(manifest string, client clientset.Interface, status
...
+				return v1alpha1.StepStatusUnknown, createError
+			}
+			// If no error occured, make sure to wait for the API to become available.
+			return v1alpha1.StepStatusWaitingForAPI, nil
+		}
+		return v1alpha1.StepStatusUnknown, nil
+	}
+}
+
+func (b *builder) NewCRDV1Beta1Step(client apiextensionsv1beta1client.ApiextensionsV1beta1Interface, step *v1alpha1.Step) StepperFunc {
```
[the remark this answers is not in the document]
exdx: I think in principle that's a good idea - not sure if we need to do that in this PR though

test/e2e/installplan_e2e_test.go
exdx: When we had 2 as the value we hit the float -> int conversion issue when we were upgrading all CRDs to v1. Changing to 2.1 removed us hitting this issue with v1 CRDs.

Now we are back to handling v1beta1 CRDs I moved the value back to where it was previously. Should we have another test, with real floats, for this case?

test/e2e/crd_e2e_test.go
exdx: all of our e2e tests currently create an internal crd which gets converted to a v1beta1 later on when creating the CSV. I could create the v1beta1 directly but would then need to write my own `createCSV` and other helper functions

test/e2e/crd_e2e_test.go
exdx: yes, I copied it from an existing test

test/e2e/crd_e2e_test.go
exdx: Yes

test/e2e/crd_e2e_test.go
exdx: This test is just ensuring we can install v1beta1 CRDs with an invalid v1 schema (proving we handle v1beta1 CRDs with the v1beta1 client). Not sure what you mean by update

pkg/lib/crd/unmarshal.go
```diff
@@ -0,0 +1,60 @@
...
+func UnmarshalV1Beta1(manifest string) (*apiextensionsv1beta1.CustomResourceDefinition, error) {
+	crd := &apiextensionsv1beta1.CustomResourceDefinition{}
+	reader := bytes.NewReader([]byte(manifest))
+	decoder := yaml.NewYAMLOrJSONDecoder(reader, 30)
+	if err := decoder.Decode(crd); err != nil {
+		return nil, fmt.Errorf("v1beta1 crd unmarshaling failed: %s", err)
+	}
+
+	// set CRD type meta
+	// for purposes of fake client for unit tests to pass
```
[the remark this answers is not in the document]
exdx: this is for the purposes of the fake client

pkg/controller/operators/catalog/operator.go
```diff
@@ -1410,6 +1478,28 @@ func removeDeprecatedV1StoredVersions(oldCRD *apiextensionsv1.CustomResourceDefi
...
+	newStoredVersions := []string{}
+	newCRDVersions := GetCRDV1Beta1VersionsMap(newCRD)
+	for _, v := range oldCRD.Status.StoredVersions {
+		_, ok := newCRDVersions[v]
+		if ok {
+			newStoredVersions = append(newStoredVersions, v)
+		}
+	}
+
+	if len(newStoredVersions) < 1 {
```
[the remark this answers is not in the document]
exdx: Good call

pkg/controller/operators/catalog/step.go
```diff
@@ -98,61 +111,161 @@ func (b *builder) NewCRDStep(manifest string, client clientset.Interface, status
...
+		case v1alpha1.StepStatusUnknown, v1alpha1.StepStatusNotPresent:
+			crd, err := crdlib.UnmarshalV1Beta1(step.Resource.Manifest)
+			if err != nil {
 				return v1alpha1.StepStatusUnknown, err
 			}
+			b.logger.Debugf("creating v1beta1 CRD %#v", crd)
+
+			_, createError := client.CustomResourceDefinitions().Create(context.TODO(), crd, metav1.CreateOptions{})
+			if k8serrors.IsAlreadyExists(createError) {
+				currentCRD, _ := client.CustomResourceDefinitions().Get(context.TODO(), crd.GetName(), metav1.GetOptions{})
```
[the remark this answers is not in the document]
exdx: Seems likely, although I'm not entirely sure. I was trying to keep the existing convention as much as possible when adding support for v1 CRDs

pkg/lib/crd/unmarshal.go
```diff
@@ -0,0 +1,60 @@
...
+
+func init() {
+	// Add conversions between CRD versions
+	install.Install(scheme)
+}
+
+// UnmarshalV1 takes in a CRD manifest and returns a v1 versioned CRD object.
+func UnmarshalV1(manifest string) (*apiextensionsv1.CustomResourceDefinition, error) {
+	crd := &apiextensionsv1.CustomResourceDefinition{}
+	reader := bytes.NewReader([]byte(manifest))
```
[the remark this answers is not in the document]
exdx: Not really, just took this from existing code

exdx:
@sttts we addressed the primary concerns you had regarding our CRD conversions - no more using internal types. That being said, OLM has a lot of responsibility around handling and upgrading CRDs and it would extremely valuable to us if we could have a minute of your time to discuss these things further on a call. 

/hold cancel

pkg/controller/operators/catalog/operator.go
```diff
@@ -1329,6 +1344,27 @@ func EnsureCRDVersions(oldCRD *apiextensionsv1.CustomResourceDefinition, newCRD
 	return nil
 }
 
+// Ensure all existing served versions are present in new CRD
+func EnsureV1Beta1CRDVersions(oldCRD *apiextensionsv1beta1.CustomResourceDefinition, newCRD *apiextensionsv1beta1.CustomResourceDefinition) error {
```
sttts, line 1348: why is this an issue? Versions can go away, like old alpha versions.

pkg/controller/operators/catalog/operator.go
```diff
@@ -1364,6 +1400,38 @@ func validateV1CRDCompatibility(dynamicClient dynamic.Interface, oldCRD *apiexte
 	return nil
 }
 
+// Validate all existing served versions against new CRD's validation (if changed)
+func validateV1Beta1CRDCompatibility(dynamicClient dynamic.Interface, oldCRD *apiextensionsv1beta1.CustomResourceDefinition, newCRD *apiextensionsv1beta1.CustomResourceDefinition) error {
+	logrus.Debugf("Comparing %#v to %#v", oldCRD.Spec.Validation, newCRD.Spec.Validation)
+	// If validation schema is unchanged, return right away
+	if reflect.DeepEqual(oldCRD.Spec.Validation, newCRD.Spec.Validation) {
```
sttts, line 1407: there are also validation schemas per version. You need both.

pkg/controller/operators/catalog/operator.go
```diff
@@ -1364,6 +1400,38 @@ func validateV1CRDCompatibility(dynamicClient dynamic.Interface, oldCRD *apiexte
...
+		return nil
+	}
+	convertedCRD := &apiextensions.CustomResourceDefinition{}
+	if err := apiextensionsv1beta1.Convert_v1beta1_CustomResourceDefinition_To_apiextensions_CustomResourceDefinition(newCRD, convertedCRD, nil); err != nil {
```
sttts, line 1411: here is still a conversion.

pkg/controller/operators/catalog/step.go
```diff
@@ -98,61 +111,161 @@ func (b *builder) NewCRDStep(manifest string, client clientset.Interface, status
...
+						if err != nil {
+							return v1alpha1.StepStatusUnknown, errorwrap.Wrapf(err, "error missing existing CRD version(s) in new CRD: %s", step.Resource.Name)
+						}
+
+						if err = validateV1Beta1CRDCompatibility(b.dynamicClient, currentCRD, crd); err != nil {
+							return v1alpha1.StepStatusUnknown, errorwrap.Wrapf(err, "error validating existing CRs agains new CRD's schema: %s", step.Resource.Name)
+						}
+					}
+					// Remove deprecated version in CRD storedVersions
+					storeVersions := removeDeprecatedV1Beta1StoredVersions(currentCRD, crd)
```
sttts, line 247: this whole block is dangerous and might cause serious data loss. Just because a version not served anymore, does not mean that it is not stored in etcd anymore. If I follow you here, you are forcefully removing the remove version from the `StoredVersions` field. I.e. you are claiming that you have migrated every CR. Have you? I don't think.

pkg/lib/crd/version.go
```diff
@@ -23,24 +25,28 @@ var supportedCRDVersions = map[string]struct{}{
...
 	return v, nil
 }
 
-// NotEqual determines whether two CRDs are equal based on the versions and validations of both.
-// NotEqual looks at the range of the old CRD versions to ensure index out of bounds errors do not occur.
+// V1NotEqual determines whether two v1 CRDs are equal based on the versions and validations of both.
+// V1NotEqual looks at the range of the old CRD versions to ensure index out of bounds errors do not occur.
 // If true, then we know we need to update the CRD on cluster.
-func NotEqual(currentCRD *apiextensionsv1.CustomResourceDefinition, oldCRD *apiextensionsv1.CustomResourceDefinition) bool {
+func V1NotEqual(currentCRD *apiextensionsv1.CustomResourceDefinition, oldCRD *apiextensionsv1.CustomResourceDefinition) bool {
```
sttts, line 49: does this func mean that CRDs from an upgrade are not updated in the cluster if the versions are equal? What about the other 47 fields of a CRD?

test/e2e/util_test.go
```diff
@@ -551,6 +632,21 @@ func serializeCRD(t GinkgoTInterface, crd apiextensions.CustomResourceDefinition
 	return manifest.String()
 }
 
+func serializeV1CRD(t GinkgoTInterface, crd *apiextensionsv1.CustomResourceDefinition) string {
+	scheme := runtime.NewScheme()
+	require.NoError(t, extScheme.AddToScheme(scheme))
+	require.NoError(t, k8sscheme.AddToScheme(scheme))
```
sttts, line 638: nit: why do you need extensions and k8s types?

test/e2e/crd_e2e_test.go
exdx: TODO: we should update all our e2e to create v1beta1 CRDs and update the createCSV to accept the versioned CRD type

pkg/controller/operators/catalog/operator.go
exdx: This a good point and we have addressed it by removing this function entirely.

pkg/controller/operators/catalog/operator.go
exdx: This function verifies existing CRs against the new CRD validation and the comparison of the deep equals is just to avoid this check which is potentially expensive

pkg/controller/operators/catalog/operator.go
exdx: Here we convert to the internal version to do static schema validation. Specifically, the NewSchemaValidator which creates an openapi schema validator for the given CRD validation, expects an internal CRD type to be provided to it, so OLM does the conversion here when validating the new CRD.

The internal CRD is never written to the cluster again. We can make a follow-up BZ to use dry-run APIs instead of using the internal validation package, if necessary.

sttts: Yes, would prefer dry-run (follow-up). The schema validator is also a component considered private.

pkg/controller/operators/catalog/step.go
exdx: You are correct and we have discovered this code block to be faulty - we have removed it entirely from the code path and will address in a followup

pkg/lib/crd/version.go
exdx: That's a good point. We have elected to remove this check entirely since `InstallPlan` resources are not continually reconciled, and so we will attempt to update the CRD on the cluster each time.

pkg/controller/operators/catalog/operator.go
ecordell: on second look - this should be removed entirely. Validation is an optional field, so if both are nil we may skip cr compatibility checks

pkg/controller/operators/catalog/operator.go
```diff
@@ -1364,6 +1341,40 @@ func validateV1CRDCompatibility(dynamicClient dynamic.Interface, oldCRD *apiexte
 	return nil
 }
 
+// Validate all existing served versions against new CRD's validation (if changed)
+func validateV1Beta1CRDCompatibility(dynamicClient dynamic.Interface, oldCRD *apiextensionsv1beta1.CustomResourceDefinition, newCRD *apiextensionsv1beta1.CustomResourceDefinition) error {
+	logrus.Debugf("Comparing %#v to %#v", oldCRD.Spec.Validation, newCRD.Spec.Validation)
+	// If validation schema is unchanged, return right away
+	if reflect.DeepEqual(oldCRD.Spec.Validation, newCRD.Spec.Validation) {
+		return nil
+	}
```
ecordell, line 1350:
```suggestion
```

pkg/controller/operators/catalog/operator.go
ecordell: How can we dry-run against an API that has not been applied to the cluster yet?

The goal is to catch the mistake before the CRD has been created.

There seems to be little practical issue with using the internal types, correct? We could just as easily read the json schema from the CRD spec and use an off-the-shelf schema validator to accomplish the same goals.

pkg/controller/operators/catalog/operator.go
exdx: Removed

pkg/lib/crd/unmarshal.go
```diff
@@ -0,0 +1,50 @@
...
+	decoder := yaml.NewYAMLOrJSONDecoder(reader, 30)
+	if err := decoder.Decode(crd); err != nil {
+		return nil, fmt.Errorf("v1 crd unmarshaling failed: %s", err)
+	}
+
+	return crd, nil
+}
+
+// UnmarshalV1 takes in a CRD manifest and returns a v1beta1 versioned CRD object.
```
ecordell, line 40:
```suggestion
// UnmarshalV1Beta1 takes in a CRD manifest and returns a v1beta1 versioned CRD object.
```

PR #1472, opened 2020-04-21
Bug 1826443: Pod Config Deployment Hash Error

awgreene:
The Pod Config e2e test have been failing because the deployment is not reinstalled when the actual deployment hash does not match the calculated deployment hash. This commit updates OLM to reinstall a deployment when the hash doesn't match the calculated hash.

### DIFF
--- a/pkg/controller/install/errors.go
+++ b/pkg/controller/install/errors.go
@@ -38,11 +38,11 @@ func IsErrorUnrecoverable(err error) bool {
 	if err == nil {
 		return false
 	}
-	_, ok := unrecoverableErrors[reasonForError(err)]
+	_, ok := unrecoverableErrors[ReasonForError(err)]
 	return ok
 }
 
-func reasonForError(err error) string {
+func ReasonForError(err error) string {
 	switch t := err.(type) {
 	case StrategyError:
 		return t.Reason
--- a/pkg/controller/operators/olm/operator.go
+++ b/pkg/controller/operators/olm/operator.go
@@ -1685,7 +1685,11 @@ func (a *Operator) updateInstallStatus(csv *v1alpha1.ClusterServiceVersion, inst
 	}
 
 	if strategyErr != nil {
-		csv.SetPhaseWithEventIfChanged(requeuePhase, requeueConditionReason, fmt.Sprintf("installing: %s", strategyErr), now, a.recorder)
+		if install.ReasonForError(strategyErr) == install.StrategyErrDeploymentUpdated {
+			csv.SetPhaseWithEventIfChanged(v1alpha1.CSVPhaseInstallReady, requeueConditionReason, fmt.Sprintf("installing: %s", strategyErr), now, a.recorder)
+		} else {
+			csv.SetPhaseWithEventIfChanged(requeuePhase, requeueConditionReason, fmt.Sprintf("installing: %s", strategyErr), now, a.recorder)
+		}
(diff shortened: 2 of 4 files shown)


pkg/controller/install/errors.go
```diff
@@ -42,6 +42,15 @@ func IsErrorUnrecoverable(err error) bool {
 	return ok
 }
 
+// IsErrorDeploymentUpdated reports if a given strategy error is due to an update to the deployment.
```
Bowenislandsong, line 45: what's the point of the function if all input is going to have error? Could we just move the reasonForError statement to where it is used?

awgreene: Sure - that sounds good to me.

awgreene:
e2e-gcp-upgrade failed to bootstrap
/retest

pkg/controller/install/errors.go
```diff
@@ -38,11 +38,11 @@ func IsErrorUnrecoverable(err error) bool {
 	if err == nil {
 		return false
 	}
-	_, ok := unrecoverableErrors[reasonForError(err)]
+	_, ok := unrecoverableErrors[ReasonForError(err)]
 	return ok
 }
 
-func reasonForError(err error) string {
+func ReasonForError(err error) string {
```
benluddy, line 45: Just thinking "out loud" here: it might be nice to (later!) use some of [the newer error wrapping features from the standard library](https://blog.golang.org/go1.13-errors) in place of the existing {Reason, Message} that exists today.

test/e2e/subscription_e2e_test.go
```diff
@@ -1212,10 +1212,6 @@ var _ = Describe("Subscription", func() {
 		require.NotNil(GinkgoT(), subscription)
 
 		csv, err := fetchCSV(GinkgoT(), crClient, subscription.Status.CurrentCSV, testNamespace, buildCSVConditionChecker(v1alpha1.CSVPhaseSucceeded))
-		if err != nil {
```
benluddy, line 1215: :tada:

pkg/controller/operators/olm/operator.go
```diff
@@ -1685,7 +1685,11 @@ func (a *Operator) updateInstallStatus(csv *v1alpha1.ClusterServiceVersion, inst
 	}
 
 	if strategyErr != nil {
-		csv.SetPhaseWithEventIfChanged(requeuePhase, requeueConditionReason, fmt.Sprintf("installing: %s", strategyErr), now, a.recorder)
+		if install.ReasonForError(strategyErr) == install.StrategyErrDeploymentUpdated {
+			csv.SetPhaseWithEventIfChanged(v1alpha1.CSVPhaseInstallReady, requeueConditionReason, fmt.Sprintf("installing: %s", strategyErr), now, a.recorder)
+		} else {
+			csv.SetPhaseWithEventIfChanged(requeuePhase, requeueConditionReason, fmt.Sprintf("installing: %s", strategyErr), now, a.recorder)
```
benluddy, line 1691: Is the reason `InstallWaiting` when this branch is hit, and is that the most appropriate reason in this case?

ecordell: This is somewhat weird given the function signature - you would expect the requeue phase to be used.

ecordell: Discussed offline - this whole state management should probably be refactored to make this behavior clearer, but doing so here would make it difficult to backport.

PR #1473, opened 2020-04-21
BUG 1825349: Generation required bug

awgreene:
In the InstallPlan, the Generation field should not be required and may
be omitted.

Bugzilla Ref: https://bugzilla.redhat.com/show_bug.cgi?id=1825349

### DIFF
--- a/deploy/chart/templates/0000_50_olm_04-installplan.crd.yaml
+++ b/deploy/chart/templates/0000_50_olm_04-installplan.crd.yaml
@@ -67,7 +67,6 @@ spec:
           - approval
           - approved
           - clusterServiceVersionNames
-          - generation
           properties:
             approval:
               description: Approval is the user approval policy for an InstallPlan.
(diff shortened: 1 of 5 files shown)


awgreene:
/retest
```TestDeleteInternalRegistryPodTriggersRecreation``` passes locally

awgreene:
/test e2e-gcp

`ERROR: (gcloud.compute.instance-groups.list-instances) could not parse resource []`

awgreene:
/test e2e-gcp
`ERROR: (gcloud.compute.instance-groups.list-instances) could not parse resource []`

russellb:
@awgreene There are 3 commits in this PR, but only the first one seems related to the referenced bug.  Can you add some more clarity on the other changes?

It would be really helpful to have a lot more detail in each commit message.  For example, the other 2 commits only have a 1 line commit message and they don't do much to explain why the change was being made, so I'm having a hard time determining whether allowing those changes to be included is appropriate.

Thanks!

awgreene:
Good question @russellb

>@awgreene There are 3 commits in this PR, but only the first one seems related to the referenced bug. Can you add some more clarity on the other changes?

Sure - let's review the commits:
1. As you said, the first commit is the fix for the bug mentioned in the BZ.
2. The second commit updates an e2e test that would always fail because an image that the test relied on was updated.
3. The third updates an e2e test that was experiencing an extremely high failure rate - preventing bug fixes from getting into 4.4.

PR #1476, opened 2020-04-22
Run upstream E2E workflow in parallel across multiple runners.

benluddy:
Forgive my AWK. Github allows up to twenty concurrent jobs, so I've
configured the end-to-end test workflow to split the specs across
multiple jobs.


Bowenislandsong:
lgtm

benluddy:
This was fun, but we can get comparable speedups without hackery if we provide our own runners or improve test isolation.

PR #1481, opened 2020-04-24
test(resolver): add a test case that explicitly verifies ownership transfer

ecordell:
This adds a test case to the resolver that explicitly checks that an api can transfer ownership from one operator to another during an upgrade.

**Reviewer Checklist**


awgreene:
Merging the changes introduced here into https://github.com/operator-framework/operator-lifecycle-manager/pull/1483

PR #1482, opened 2020-04-24
Bug 1823714: Update PkgManifest upon catsrc update

Bowenislandsong:
The catsrc update sync triggred by the informer does not update pkgmanifest. This commit addes update mechanism.

**Reviewer Checklist**

### DIFF
--- a/pkg/package-server/provider/registry.go
+++ b/pkg/package-server/provider/registry.go
@@ -9,8 +9,6 @@ import (
 	"sync"
 	"time"
 
-	utillabels "github.com/operator-framework/operator-lifecycle-manager/pkg/lib/kubernetes/pkg/util/labels"
-	"github.com/operator-framework/operator-registry/pkg/api"
 	"github.com/sirupsen/logrus"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/connectivity"
@@ -26,9 +24,11 @@ import (
 	operatorslisters "github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/listers/operators/v1alpha1"
 	registrygrpc "github.com/operator-framework/operator-lifecycle-manager/pkg/controller/registry/grpc"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/registry/resolver"
+	utillabels "github.com/operator-framework/operator-lifecycle-manager/pkg/lib/kubernetes/pkg/util/labels"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/queueinformer"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/package-server/apis/operators"
 	pkglisters "github.com/operator-framework/operator-lifecycle-manager/pkg/package-server/client/listers/operators/internalversion"
+	"github.com/operator-framework/operator-registry/pkg/api"
 )
 
 const (
@@ -189,9 +189,17 @@ func (p *RegistryProvider) syncCatalogSource(obj interface{}) (syncError error) 
 		Namespace: source.GetNamespace(),
 		Name:      source.GetName(),
 	}
+
 	if sourceMeta := p.sources.GetMeta(key); sourceMeta != nil && sourceMeta.Address == address {
-		// If the address hasn't changed, don't bother creating a new source
-		logger.Debug("catalog address unchanged, skipping source creation")
+		logger.Infof("updating PackageManifest based on CatalogSource changes: %v", key)
+		timeout, cancel := context.WithTimeout(context.Background(), cacheTimeout)
+		defer cancel()
+		var client *registryClient
+		client, syncError = p.registryClient(key)
+		if syncError != nil {
+			return
+		}
+		syncError = p.refreshCache(timeout, client)
 		return
 	}
 
--- a/test/e2e/packagemanifest_e2e_test.go
+++ b/test/e2e/packagemanifest_e2e_test.go
@@ -3,6 +3,7 @@ package e2e
 import (
 	"context"
 	"encoding/json"
+
 	"github.com/blang/semver"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
@@ -145,12 +146,13 @@ var _ = Describe("Package Manifest API lists available Operators from Catalog So
 
 	Context("Given a CatalogSource created using gRPC catalog source type", func() {
 		var (
-			packageName   string
-			catalogSource *v1alpha1.CatalogSource
+			packageName, displayName string
+			catalogSource            *v1alpha1.CatalogSource
 		)
 		BeforeEach(func() {
 			sourceName := genName("catalog-")
 			packageName = "etcd-test"
+			displayName = "etcd test catalog"
 			image := "quay.io/olmtest/catsrc-update-test:related"
 
 			catalogSource = &v1alpha1.CatalogSource{
@@ -164,8 +166,9 @@ var _ = Describe("Package Manifest API lists available Operators from Catalog So
 					Labels:    map[string]string{"olm.catalogSource": sourceName},
 				},
 				Spec: v1alpha1.CatalogSourceSpec{
-					SourceType: v1alpha1.SourceTypeGrpc,
-					Image:      image,
+					SourceType:  v1alpha1.SourceTypeGrpc,
+					Image:       image,
+					DisplayName: displayName,
 				},
 			}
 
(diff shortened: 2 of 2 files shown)


pkg/package-server/provider/registry.go
```diff
@@ -189,6 +189,15 @@ func (p *RegistryProvider) syncCatalogSource(obj interface{}) (syncError error)
 		Namespace: source.GetNamespace(),
```
awgreene, line 189: You commit message has misspelled words: `triggred` and`addes`. You should consider[ enabling spellcheck in VIM](https://www.ostechnix.com/use-spell-check-feature-vim-text-editor/) or whichever editor you use to create your commit messages.

pkg/package-server/provider/registry.go
```diff
@@ -189,6 +189,15 @@ func (p *RegistryProvider) syncCatalogSource(obj interface{}) (syncError error)
...
 		Name:      source.GetName(),
 	}
+
+	logger.Infof("updating PackageManifest based on CatalogSource changes: %v", key)
+	timeout, cancel := context.WithTimeout(context.Background(), cacheTimeout)
+	defer cancel()
+	client, err := p.registryClient(key)
+	if err == nil {
+		err = p.refreshCache(timeout, client)
+	}
```
awgreene, line 199: I see that your adding this code to check if the list of packages has changed. The code looks correct in that it will update the list of packagemanifests but I have two questions:
1. Could you make sure that the error is captured when the PackageServer is unable to refresh the cache? Either as a log or returning the error?
2. It looks like this could be moved into the [if statement below](https://github.com/operator-framework/operator-lifecycle-manager/blob/006c12bdbeda179ed0b24193e639596dfd4569ef/pkg/package-server/provider/registry.go#L201) to prevent it from being called when sources don't already exist. Could you investigate?

test/e2e/packagemanifest_e2e_test.go
```diff
@@ -168,6 +169,74 @@ var _ = Describe("Package Manifest", func() {
 			require.True(GinkgoT(), ok, "Expect this image %s to exist in the related images list\n", v)
 		}
 	})
+	It("updating CatalogSource", func() {
```
harishsurf, line 172: Could the test be more structured to use ginkgo's container blocks and replacing assertions with gomega matcher library? You could refer to #1424 as a reference

harishsurf: you could refer to this [cheatsheet](https://docs.google.com/document/d/1SwDGGrOTpCfE32gPE8GyFOwyZ058BUMhcI_KkSzU7Bg/edit#)

pkg/package-server/provider/registry.go
Bowenislandsong: neat! did not know that's an option

Bowenislandsong: funny, i went and looked. turns out i had it on. Somehow i dont see notifiations. I'll check what was wrong with that.

pkg/package-server/provider/registry.go
Bowenislandsong: 1. wow that was a very nice catch, I assumed the return was err instead of syncError. Kinda a weird way to express dislike for this kind of error return method.
2. good call, that would reduce the traffic. I somehow thought about it backward initially.

test/e2e/packagemanifest_e2e_test.go
Bowenislandsong: Right, that would be much better, but may I suggest that we do that in a separate PR and have this test arranged exactly as the other tests in this file? IMO refactoring should be its own PR rather than smuggles in bug fixes.

Bowenislandsong: I updated the test to at least not to use assert and conform to the other tests.

benluddy: This is a new test -- why create more future work for ourselves? If this bug were on fire, I can see the argument for addressing the test feedback in a follow-up, but it doesn't seem to be urgent enough to do that.

Bowenislandsong: Lets wait for https://github.com/operator-framework/operator-lifecycle-manager/pull/1439/files to merge first.

awgreene: I agree with @benluddy

pkg/package-server/provider/registry.go
```diff
@@ -189,7 +189,18 @@ func (p *RegistryProvider) syncCatalogSource(obj interface{}) (syncError error)
...
+
 	if sourceMeta := p.sources.GetMeta(key); sourceMeta != nil && sourceMeta.Address == address {
+		logger.Infof("updating PackageManifest based on CatalogSource changes: %v", key)
+		timeout, cancel := context.WithTimeout(context.Background(), cacheTimeout)
+		defer cancel()
+		client, err := p.registryClient(key)
+		if err == nil {
+			syncError = p.refreshCache(timeout, client)
+		} else {
+			syncError = err
```
awgreene, line 201: Should we add a return here?

pkg/package-server/provider/registry.go
```diff
@@ -189,7 +189,18 @@ func (p *RegistryProvider) syncCatalogSource(obj interface{}) (syncError error)
...
+		}
```
awgreene, line 202:
```suggestion
		if err!= nil {
			syncError = err
			return
		}
	syncError = p.refreshCache(timeout, client)
```

awgreene: @Bowenislandsong you should exit earlier rather than using an `if/else`

Bowenislandsong: You know they take exactly the same number of steps! lol

But i see your point of less indent from the if statement.

test/e2e/packagemanifest_e2e_test.go
Bowenislandsong: Yes, I agree, but I am saying waiting for https://github.com/operator-framework/operator-lifecycle-manager/pull/1439/files will resolve all of our concerns. That PR is very close to merging.

/hold

pkg/package-server/provider/registry.go
```diff
@@ -189,7 +189,18 @@ func (p *RegistryProvider) syncCatalogSource(obj interface{}) (syncError error)
...
 	}
+
 	if sourceMeta := p.sources.GetMeta(key); sourceMeta != nil && sourceMeta.Address == address {
+		logger.Infof("updating PackageManifest based on CatalogSource changes: %v", key)
+		timeout, cancel := context.WithTimeout(context.Background(), cacheTimeout)
+		defer cancel()
+		client, err := p.registryClient(key)
+		if err != nil {
+			syncError = err
+			return
```
awgreene, line 200:
```suggestion
		client, syncError := p.registryClient(key)
		if syncErr != nil {
			return
		}
```

pkg/package-server/provider/registry.go
```diff
@@ -189,7 +189,18 @@ func (p *RegistryProvider) syncCatalogSource(obj interface{}) (syncError error)
...
+		}
+		syncError = p.refreshCache(timeout, client)
+
 		// If the address hasn't changed, don't bother creating a new source
```
awgreene, line 204: This comment is not longer correct

pkg/package-server/provider/registry.go
Bowenislandsong: Thanks! I will update once the dependent PR is merged!

Bowenislandsong: Just realized, i can't do this, and was delieberatly avoiding it. This reinitialized `syncError`.

test/e2e/packagemanifest_e2e_test.go
```diff
@@ -195,6 +198,28 @@ var _ = Describe("Package Manifest API lists available Operators from Catalog So
 				"quay.io/coreos/etcd-operator@sha256:c0301e4686c3ed4206e370b42de5a3bd2229b9fb4906cf85f3f30650424abec2",
 			}), "Expected images to exist in the related images list\n")
 		})
+
+		It("updating CatalogSource", func() {
```
harishsurf, line 202: This could be written as : 
```
When ("the display name for catalog source is updated", func() {
       BeforeEach(func() {
            // code to update field
       }) 
       It("should successfully update the CatalogSource field", func () {
          // assertion for catsrc update validation
       })
})
```
Note: The reason for this change is going back to what our `It` spec should be. It should only have the assertions which forms the crux of the functionality being tested. Everything else lives outside.
In this case, assertions for fetching `packageManifest` & `catsrc` are part of pre-setup and hence they can live in parent container blocks (so that they get reused for future `It` specs if they use the same pre-setup)

pkg/package-server/provider/registry.go
```diff
@@ -189,7 +189,18 @@ func (p *RegistryProvider) syncCatalogSource(obj interface{}) (syncError error)
...
 		logger.Debug("catalog address unchanged, skipping source creation")
```
dinhxuanvu, line 205: I feel this debug line should be moved above the sync process. We should sync after we determine that catalog is unchanged right?

pkg/package-server/provider/registry.go
dinhxuanvu: Can we just `return err` on line 199?

pkg/package-server/provider/registry.go
Bowenislandsong: you are right to pick on this, but I also thing this bugbug is nolonger needed since we are syncing and not skipping the updatae entirely.

Bowenislandsong:
Can we skip the err nit, that is really a nit and breaks the way how this function is structured.

pkg/package-server/provider/registry.go
```diff
@@ -189,9 +189,17 @@ func (p *RegistryProvider) syncCatalogSource(obj interface{}) (syncError error)
 		Namespace: source.GetNamespace(),
 		Name:      source.GetName(),
 	}
+
 	if sourceMeta := p.sources.GetMeta(key); sourceMeta != nil && sourceMeta.Address == address {
-		// If the address hasn't changed, don't bother creating a new source
-		logger.Debug("catalog address unchanged, skipping source creation")
```
dinhxuanvu, line 194: We still want to this debug line right? I thought my comment said it should be moved to the top.

dinhxuanvu: Saw your comment below. I'm fine with skipping this.

pkg/package-server/provider/registry.go
awgreene: @Bowenislandsong just change `:=` to `=` and decalre the client variable before the function call

test/e2e/packagemanifest_e2e_test.go
```diff
@@ -195,6 +198,30 @@ var _ = Describe("Package Manifest API lists available Operators from Catalog So
...
+				pm, err := fetchPackageManifest(pmc, testNamespace, packageName, packageManifestHasStatus)
+				Expect(err).NotTo(HaveOccurred(), "error getting package manifest")
+				Expect(pm).ShouldNot(BeNil())
+				Expect(pm.GetName()).Should(Equal(packageName))
+				Expect(pm.Status.CatalogSourceDisplayName).Should(Equal(displayName))
+
+				catalogSource, err = crc.OperatorsV1alpha1().CatalogSources(testNamespace).Get(context.TODO(), catalogSource.GetName(), metav1.GetOptions{})
+				Expect(err).NotTo(HaveOccurred(), "error getting catalogSource")
+
+				displayName := "updated Name"
```
awgreene, line 213: shouldn't this be `=` not `:=`

PR #1488, opened 2020-04-28
Set up a bit of static analysis.

benluddy:
It's configured to pass even when it reports issues, and most of the
linters are disabled, but this can at least get the conversation
started.


.github/workflows/lint.yml
```diff
@@ -0,0 +1,12 @@
...
+  push:
+    branches:
+      - master
+  pull_request:
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - run: make lint
```
harishsurf, line 12: Any reason for not downloading and using the release binary from golangcli-lint? Trying to understand what the standard practice should be

benluddy: I lean toward a container image that can be shared with the rest of the org, but I wanted to open a self-contained PR to demonstrate utility first. Good question -- curious what opinions others have.

PR #1489, opened 2020-04-28
Bug 1826446: (fix) Admission Webhook names must be unique

awgreene:
Problem:
OLM currently creates all ValidatingWebhookConfigurations and
MutatingWebhookConfigurations using the name provided in the
WebhookDescription. This causes an issue when an operator that defines
an admission webhook in the CSV is installed in multiple namespaces,
causes the two operators to fight over the admission webhook.

Solution:
Generate a unique name for each webhook using the GenerateName field in
the Object Metadata.


test/e2e/webhook_e2e_test.go
```diff
@@ -0,0 +1,955 @@
...
+
+		c := newKubeClient(GinkgoT())
+		crc := newCRClient(GinkgoT())
+
+		namespace, nsCleanupFunc := newNamespace(GinkgoT(), c, genName("csc-test-"))
+		defer nsCleanupFunc()
+
+		og := newOperatorGroup(namespace.Name, genName("test-og-"), nil, nil, []string{}, false)
+		og, err := crc.OperatorsV1().OperatorGroups(namespace.Name).Create(context.TODO(), og, metav1.CreateOptions{})
+		require.NoError(GinkgoT(), err)
```
harishsurf, line 31: could be replaced with Gomega's matcher: `Expect(err).ToNot(HaveOccurred())`
As we are transitioning towards ginkgo, the idea is to stray away from `/testify/require` package and use gomega assertions instead

test/e2e/webhook_e2e_test.go
```diff
@@ -0,0 +1,955 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/operator-framework/api/pkg/operators/v1alpha1"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/install"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/operatorclient"
+)
+
+var _ = Describe("Webhook", func() {
+	It("creates global webhooks", func() {
```
harishsurf, line 20: Instead of having the entire test in a single `It` spec, you could use ginkgo's container blocks to restructure the test. 
For example: 
```
Describe("Given a CSV containing a webhook in its spec", func() { 
      BeforeEach(func(){
          // creation of csv and webhook
      })
      AfterEach(func(){
         // optional cleanups if needed
      })
      It("creates the webhook", func() {
           // use gomega's matcher library to verify webhook creation
      })
})
```
This way, if you choose to reuse the same CSV for a different test, you can add that test as an `It` spec inside of this Describe block. That way, the `BeforeEach` does the setup required for you. 
This avoids code duplication and also gives better error descriptions(using the descriptions in container blocks) in case of a test failure

test/e2e/webhook_e2e_test.go
```diff
@@ -0,0 +1,955 @@
...
+			ContainerPort:           443,
+			AdmissionReviewVersions: []string{"v1beta1", "v1"},
+			SideEffects:             &sideEffect,
+		}
+
+		csv.Spec.WebhookDefinitions = []v1alpha1.WebhookDescription{
+			webhook,
+		}
+		cleanupCSV, err := createCSV(GinkgoT(), c, crc, csv, namespace.Name, false, false)
+		require.NoError(GinkgoT(), err)
```
harishsurf, line 81: nit: same comment as above

test/e2e/webhook_e2e_test.go
```diff
@@ -0,0 +1,955 @@
...
+
+		actualWebhook, err := getWebhookWithGenName(c, webhook.Name)
+		require.NoError(GinkgoT(), err)
+
+		expected := &metav1.LabelSelector{
+			MatchLabels:      map[string]string(nil),
+			MatchExpressions: []metav1.LabelSelectorRequirement(nil),
+		}
+
+		require.Equal(GinkgoT(), expected, actualWebhook.Webhooks[0].NamespaceSelector)
```
harishsurf, line 95: `Expect(actualWebhook.Webhooks[0].NamespaceSelector).Should(Equal(expected))`

pkg/controller/operators/olm/apiservices.go
```diff
@@ -260,6 +263,8 @@ func (a *Operator) areAPIServicesAvailable(csv *v1alpha1.ClusterServiceVersion)
 // but duplicates a lot of installAPIServiceRequirements and should be refactored.
 func (a *Operator) getCaBundle(csv *v1alpha1.ClusterServiceVersion) ([]byte, error) {
 	for _, desc := range csv.GetOwnedAPIServiceDescriptions() {
+
+		log.Info("It is a APIService")
```
dinhxuanvu, line 267: nit: Leftover debug comment I assume?

awgreene:
> Looks good. I had some review comments for the webhooks test. For ginkgo conversion, you could refer to these slides. There is also a cheatsheet
@harishsurf I knew this was coming! Already working on it!

pkg/controller/install/webhook.go
```diff
@@ -69,80 +69,105 @@ func (i *StrategyDeploymentInstaller) createOrUpdateWebhook(caPEM []byte, desc v
 }
 
 func (i *StrategyDeploymentInstaller) createOrUpdateMutatingWebhook(ogNamespacelabelSelector *metav1.LabelSelector, caPEM []byte, desc v1alpha1.WebhookDescription) error {
-	webhooks := []admissionregistrationv1.MutatingWebhook{
-		desc.GetMutatingWebhook(i.owner.GetNamespace(), ogNamespacelabelSelector, caPEM),
+	webhookSelector := labels.SelectorFromSet(ownerutil.OwnerLabel(i.owner, "ClusterServiceVersion")).String()
```
dinhxuanvu, line 72: nit: This string "ClusterServiceVersion" is used quite a few times. So maybe it will be cleaner to create a const for it (CSVKind).

pkg/controller/operators/olm/operator.go
```diff
@@ -920,6 +959,29 @@ func (a *Operator) handleClusterServiceVersionDeletion(obj interface{}) {
 		syncError := a.objGCQueueSet.RequeueEvent("", kubestate.NewResourceEvent(kubestate.ResourceUpdated, cr))
 		logger.Debugf("handleCSVdeletion - requeued update event for %v, res=%v", cr, syncError)
 	}
+
+	webhookSelector := labels.SelectorFromSet(ownerutil.OwnerLabel(clusterServiceVersion, "ClusterServiceVersion")).String()
+	mWebhooks, err := a.opClient.KubernetesInterface().AdmissionregistrationV1().MutatingWebhookConfigurations().List(context.TODO(), metav1.ListOptions{LabelSelector: webhookSelector})
+	if err != nil {
+		logger.WithError(err).Warn("cannot list MutatingWebhookConfigurations")
+	}
+	logger.Infof("Alex: Deleteing %d mutatingWebhooks", len(mWebhooks.Items))
```
dinhxuanvu, line 968: nit: Not sure why your name is on the log? Debugging?

pkg/controller/operators/olm/operator.go
```diff
@@ -920,6 +959,29 @@ func (a *Operator) handleClusterServiceVersionDeletion(obj interface{}) {
...
+	for _, webhook := range mWebhooks.Items {
+		logger.Infof("Alex: Syncing mutatingWebhook %v", webhook)
```
dinhxuanvu, line 970: nit: same as above.

pkg/controller/operators/olm/operator.go
```diff
@@ -920,6 +959,29 @@ func (a *Operator) handleClusterServiceVersionDeletion(obj interface{}) {
...
+		syncError := a.objGCQueueSet.RequeueEvent("", kubestate.NewResourceEvent(kubestate.ResourceUpdated, &webhook))
+		logger.Debugf("handleCSVdeletion - requeued update event for %v, res=%v", webhook, syncError)
+	}
+
+	vWebhooks, err := a.opClient.KubernetesInterface().AdmissionregistrationV1().ValidatingWebhookConfigurations().List(context.TODO(), metav1.ListOptions{LabelSelector: webhookSelector})
+	if err != nil {
+		logger.WithError(err).Warn("cannot list MutatingWebhookConfigurations")
+	}
+	logger.Infof("Alex: Deleteing %d validatingWebhooks", len(vWebhooks.Items))
```
dinhxuanvu, line 979: nit: More Alex ;)

pkg/controller/operators/olm/operator.go
```diff
@@ -920,6 +959,29 @@ func (a *Operator) handleClusterServiceVersionDeletion(obj interface{}) {
...
+	for _, webhook := range vWebhooks.Items {
+		logger.Infof("Alex: Syncing mutatingWebhook %v", webhook)
```
dinhxuanvu, line 981: nit: Alex!

pkg/controller/operators/olm/operator.go
Bowenislandsong: leftovers! lol

pkg/controller/operators/olm/operator.go
awgreene: Hahaha oops!

awgreene:
@dinhxuanvu I've addressed your comment, could you please review?

awgreene:
/retest
> could not wait for build: the build root failed after 2m11s with reason FetchSourceFailed: Failed to fetch the input source.

awgreene:
/retest

> release "release-latest" failed: the pod ci-op-7482wfws/release-latest failed after 18s (failed containers: release): ContainerFailed one or more containers exited
Container release exited with code 1, reason Error

pkg/controller/install/webhook.go
```diff
@@ -69,80 +69,105 @@ func (i *StrategyDeploymentInstaller) createOrUpdateWebhook(caPEM []byte, desc v
 }
 
 func (i *StrategyDeploymentInstaller) createOrUpdateMutatingWebhook(ogNamespacelabelSelector *metav1.LabelSelector, caPEM []byte, desc v1alpha1.WebhookDescription) error {
-	webhooks := []admissionregistrationv1.MutatingWebhook{
-		desc.GetMutatingWebhook(i.owner.GetNamespace(), ogNamespacelabelSelector, caPEM),
+	webhookSelector := labels.SelectorFromSet(ownerutil.OwnerLabel(i.owner, i.owner.GetObjectKind().GroupVersionKind().Kind)).String()
```
exdx, line 72: nit: you can call `GroupVersionKind()` directly on the object without `GetObjectKind()` first

awgreene:
/hold
Working on Ginkgo e2e tests.

test/e2e/webhook_e2e_test.go
```diff
@@ -0,0 +1,529 @@
...
+	})
+	When("Installed in an OperatorGroup that defines a selector", func() {
+		var nsLabels = map[string]string{
+			"foo": "bar",
+		}
+		var cleanupCSV cleanupFunc
+		var ogSelector = &metav1.LabelSelector{
+			MatchLabels: nsLabels,
+		}
+		var err error
```
harishsurf, line 49: We only want to declare variables at `Container` level only if we want to share them between multiple specs `It`. `err` isn't something we would want to share across multiple specs. So it is advisable to declare it right before it is used. 
Refer to [this example](https://github.com/operator-framework/operator-lifecycle-manager/blob/db290a7d8503dcfa8f49cb976bcc3da1b054392b/test/e2e/metrics_e2e_test.go#L60)

test/e2e/webhook_e2e_test.go
```diff
@@ -0,0 +1,529 @@
...
+				return true, nil
+			})
+			Expect(err).Should(BeNil())
+
+			og := newOperatorGroup(namespace.Name, genName("selector-og-"), nil, ogSelector, nil, false)
+			_, err = crc.OperatorsV1().OperatorGroups(namespace.Name).Create(context.TODO(), og, metav1.CreateOptions{})
+			Expect(err).Should(BeNil())
+		})
+		AfterEach(func() {
+			cleanupCSV()
```
harishsurf, line 71: nit: add a `nil` check

test/e2e/webhook_e2e_test.go
```diff
@@ -0,0 +1,529 @@
...
+			"foo": "bar",
+		}
+		var cleanupCSV cleanupFunc
+		var ogSelector = &metav1.LabelSelector{
+			MatchLabels: nsLabels,
+		}
+		var err error
+		BeforeEach(func() {
+			// Add a label to the namespace
+			err = wait.PollImmediate(time.Second*5, time.Minute*5, func() (bool, error) {
```
harishsurf, line 52: Gomega offers [Eventually](https://onsi.github.io/gomega/#eventually) which could be used instead of `wait.Poll`
See [example](https://github.com/operator-framework/operator-lifecycle-manager/blob/051772d2e0816359e7c45122c52917ab4cf04e57/test/e2e/scoped_client_test.go#L138)
Note when using `Eventually`:
1. It takes polling interval and timeout as an optional argument. By default, all our ginkgo tests use this [default setting](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/test/e2e/e2e_test.go#L45)
2. Also, be sure to read how `Eventually` behaves for [multiple return types](https://github.com/onsi/gomega/blob/master/gomega_dsl.go#L193)

test/e2e/webhook_e2e_test.go
```diff
@@ -0,0 +1,529 @@
...
+			actualWebhook, err := getWebhookWithGenName(c, webhookName)
+			Expect(err).Should(BeNil())
+
+			Expect(actualWebhook.Webhooks[0].NamespaceSelector).Should(Equal(ogSelector))
+		})
+	})
+	When("Installed in a SingleNamespace OperatorGroup", func() {
+		var cleanupCSV cleanupFunc
+		var og *v1.OperatorGroup
+		var err error
```
harishsurf, line 100: same comment as above. We don't want `err` to be shared across multiple `It`

test/e2e/webhook_e2e_test.go
```diff
@@ -0,0 +1,529 @@
...
+			Expect(err).Should(BeNil())
+
+			_, err = getWebhookWithGenName(c, webhook.Name)
+			Expect(err).Should(BeNil())
+
+			// Make sure old resources are cleaned up.
+			err = waitForCSVToDelete(GinkgoT(), crc, csv.Spec.Replaces)
+			Expect(err).Should(BeNil())
+
+			err = waitForNotFound(GinkgoT(), func() error {
```
harishsurf, line 314: Similar comment as above. `Eventually` could be used

pkg/controller/install/webhook.go
awgreene: This method does not exist on the owner object.

pkg/controller/install/webhook.go
```diff
@@ -69,80 +69,105 @@ func (i *StrategyDeploymentInstaller) createOrUpdateWebhook(caPEM []byte, desc v
...
+	existingWebhooks, err := i.strategyClient.GetOpClient().KubernetesInterface().AdmissionregistrationV1().MutatingWebhookConfigurations().List(context.TODO(), metav1.ListOptions{LabelSelector: webhookSelector})
+	if err != nil {
+		return err
+	}
+
+	var webhook *admissionregistrationv1.MutatingWebhookConfiguration
+	for _, item := range existingWebhooks.Items {
+		if strings.HasPrefix(item.Name, desc.Name) {
```
benluddy, line 80: What happens if one prefix is a prefix of another, e.g. `prefix1` and `prefix12`? Won't `prefix12` match both descriptions?

awgreene: This has been resolved in the latest commit by:

- Adding label to Webhooks that are based on the webhook description name
- Making sure that all Webhooks have unique names.

awgreene:
/unhold
Tests are now BDD styled tests.

awgreene:
/test e2e-gcp
> ERROR: (gcloud.compute.instance-groups.list-instances) could not parse resource []

pkg/controller/install/webhook.go
```diff
@@ -69,80 +68,108 @@ func (i *StrategyDeploymentInstaller) createOrUpdateWebhook(caPEM []byte, desc v
...
-		return err
-	}
+		for _, webhook := range existingWebhooks.Items {
+			// Update the list of webhooks
+			webhook.Webhooks = []admissionregistrationv1.MutatingWebhook{
+				desc.GetMutatingWebhook(i.owner.GetNamespace(), ogNamespacelabelSelector, caPEM),
+			}
 
+			// Attempt an update
+			if _, err := i.strategyClient.GetOpClient().KubernetesInterface().AdmissionregistrationV1().MutatingWebhookConfigurations().Update(context.TODO(), &webhook, metav1.UpdateOptions{}); err != nil {
```
ecordell, line 106: This will always attempt an update, even if there are no changes needed?

ecordell: Discussed offline, will be addressed in a follow up bugfix.

pkg/controller/install/webhook.go
```diff
@@ -69,80 +68,108 @@ func (i *StrategyDeploymentInstaller) createOrUpdateWebhook(caPEM []byte, desc v
...
-		}
-	} else if k8serrors.IsNotFound(err) {
-		hook := admissionregistrationv1.MutatingWebhookConfiguration{
-			ObjectMeta: metav1.ObjectMeta{Name: desc.Name,
-				Namespace: i.owner.GetNamespace(),
+	if len(existingWebhooks.Items) == 0 {
+		// Create a MutatingWebhookConfiguration
+		webhook := admissionregistrationv1.MutatingWebhookConfiguration{
+			ObjectMeta: metav1.ObjectMeta{
+				GenerateName: desc.Name + "-",
```
njhale, line 84: I'm pretty concerned that `WebhookConfig.Name` will be confusing for a user. I would opt to add a new `WebhookConfig.GenerateName` field which follows the expected semantics. Desirable or not, as a corollary, users would still be able ensure a singleton webhooks with `WebhookConfig.Name`.

awgreene: Do we need to leave the old name? OLM hasn't cut a release since the feature was implemented.

njhale: If that's the case then I don't seen an issue replacing `Name` w/ `GenerateName`. 

Before we go ahead, can we get some consensus among the other reviewers? If this is crazy then I don't want us to be stuck with it once we cut the next release.

benluddy: SGTM

njhale:
/hold

Awaiting https://github.com/operator-framework/api/pull/34 and a new api version tag.

pkg/controller/install/webhook.go
awgreene: Completed in the latest version of the PR

njhale:
I think this looks good. 

Note: I think the e2e tests need some style tweaks, but we can follow up later.

/hold cancel
/approve

PR #1496, opened 2020-05-01
Bug 1830031: Update OLM to use UID for OG Labels

awgreene:
Problem:
OLM is using an older version of the operator-framework/api repository
with which OLM applies a label with a blank value whose key is equal to
"olm.operatorgroup/<OperatorGroup Namespace>.<OperatorGroup Name>".
Kubernetes limits the lengths of label values and keys to 63 characters.
This limit can easily be overcome when the OperatorGroup has a long name
or when it is deployed in a namespace with a long name.

Solution:
Update OLM to use import the latest version of the operator-framework/api
repository that uses "olm.operatorgroup.uid/<OperatorGroup UID>" as the key
for OperatorGroup labels. The length of this label will always be 58
characters as UIDs are 36 characters long.


go.mod
```diff
@@ -58,7 +57,7 @@ replace (
 	github.com/openshift/api => github.com/openshift/api v0.0.0-20200331152225-585af27e34fd // release-4.5
 	github.com/openshift/client-go => github.com/openshift/client-go v0.0.0-20200326155132-2a6cd50aedd0 // release-4.5
 
-	github.com/operator-framework/api => github.com/operator-framework/api v0.3.1
+	github.com/operator-framework/api => github.com/operator-framework/api v0.3.2
```
benluddy, line 60: Why does `github.com/operator-framework/api` need a replaces directive instead of bumping the required version to v0.3.2 directly?

awgreene: Sounds good - addressed in the latest commit.

benluddy:
It looks like a libcontainer package using CGO is being transitively pulled in via the latest operator-registry: https://github.com/operator-framework/operator-lifecycle-manager/pull/1496/checks?check_run_id=637003810, which breaks the upstream image build. It has +nocgo implementations, though.

benluddy:
Looks like there's a legitimate E2E failure -- OpenShift Prow was broken over most of the weekend and incorrectly marking failing checks green.

/test e2e-aws-olm

test/e2e/operator_groups_e2e_test.go
```diff
@@ -2096,7 +2096,7 @@ var _ = Describe("Operator Group", func() {
 		}()
 
 		// Create the OperatorGroup Label
-		ogLabel := fmt.Sprintf("olm.operatorgroup/%s.%s", testNamespaceA, operatorGroup.GetName())
```
njhale, line 2099: Note: can we make this a method of `*OperatorGroup` or a wrapper?

upstream.Dockerfile
```diff
@@ -17,8 +17,8 @@ COPY go.sum go.sum
 COPY cmd cmd
 COPY util util
 COPY test test
-RUN make build
-RUN make build-util
+RUN CGO_ENABLED=0 make build
+RUN CGO_ENABLED=0 make build-util
```
njhale, line 21: `make build-util` explicitly disables CGO (see the `bin/cpb` rule)

upstream.Dockerfile
```diff
@@ -17,8 +17,8 @@ COPY go.sum go.sum
 COPY cmd cmd
 COPY util util
 COPY test test
-RUN make build
-RUN make build-util
+RUN CGO_ENABLED=0 make build
```
njhale, line 20: Is there a reason we need to disable CGO here? Does an updated dependency require it?

Additionally, the downstream `Dockerfile` explicitly enables CGO.

upstream.Dockerfile
awgreene: Removed `CGO_ENABLED=0` from make build-util

upstream.Dockerfile
awgreene: The upstream dockerfile image will not build otherwise:
docker build -t quay.io/agreene/olm:webhook-new -f upstream.Dockerfile .
Sending build context to Docker daemon  100.9MB
Step 1/28 : FROM golang:1.13-alpine as builder
 ---> 0e5426771795
Step 2/28 : LABEL stage=builder
 ---> Using cache
 ---> 3531687a249c
Step 3/28 : WORKDIR /build
 ---> Using cache
 ---> b7b246aa5d08
Step 4/28 : RUN apk update && apk add bash make git mercurial jq && apk upgrade
 ---> Using cache
 ---> 0d62b0495a72
Step 5/28 : COPY .git/HEAD .git/HEAD
 ---> Using cache
 ---> ab84e548599c
Step 6/28 : COPY .git/refs/heads/. .git/refs/heads
 ---> 5696581301a4
Step 7/28 : RUN mkdir -p .git/objects
 ---> Running in 9546e89f28c6
Removing intermediate container 9546e89f28c6
 ---> 55607d071b85
Step 8/28 : COPY Makefile Makefile
 ---> d33c08a8c47f
Step 9/28 : COPY OLM_VERSION OLM_VERSION
 ---> a934ee7e6ba2
Step 10/28 : COPY pkg pkg
 ---> 94b52e4fc3ef
Step 11/28 : COPY vendor vendor
 ---> 7e489b00274a
Step 12/28 : COPY go.mod go.mod
 ---> 52a0b3207e73
Step 13/28 : COPY go.sum go.sum
 ---> a5c109d58075
Step 14/28 : COPY cmd cmd
 ---> 33bdad10f285
Step 15/28 : COPY util util
 ---> 118c343f87c3
Step 16/28 : COPY test test
 ---> 7bda269d5a01
Step 17/28 : RUN make build
 ---> Running in d336e22af4f6
go build -mod=vendor -ldflags "-X github.com/operator-framework/operator-lifecycle-manager/pkg/version.GitCommit=4e093603eb621856a26ed6c507aec54733de93d4 -X github.com/operator-framework/operator-lifecycle-manager/pkg/version.OLMVersion=`cat OLM_VERSION`" -o bin/catalog github.com/operator-framework/operator-lifecycle-manager/cmd/catalog
# github.com/opencontainers/runc/libcontainer/system
exec: "gcc": executable file not found in $PATH
make: *** [Makefile:104: github.com/operator-framework/operator-lifecycle-manager/cmd/catalog] Error 2
The command '/bin/sh -c make build' returned a non-zero code: 2
```

File: test/e2e/csv_e2e_test.go [hunk 5]
```diff
@@ -3337,7 +3337,7 @@ var _ = Describe("CSV", func() {
 		require.NoError(GinkgoT(), err)
 
 		expected := &metav1.LabelSelector{
-			MatchLabels:      map[string]string{"olm.operatorgroup/" + namespace.Name + "." + og.GetName(): ""},
+			MatchLabels:      map[string]string{fmt.Sprintf("olm.operatorgroup.uid/%s", og.GetUID()): ""},
```
benluddy, line 3340: Sorry if this suggestion is overly cautious, but I'd like to see this label construction put into a method that can return an error when UID is empty. I don't think that condition can occur as this is currently written, but it's the sort of non-obvious bug that could be introduced a few months down the line. At the least, it'll reduce code duplication slightly.

awgreene: This is an e2e test - the values we expect should be hard coded should the not?

awgreene: I think making it easier to debug unexpected test failures is a valid request. I have updated the PR in response.

PR #1500, opened 2020-05-04
Convert scoped_client_test to ginkgo

harishsurf:
The change converts scoped_e2e_test.go to ginkgo

**Reviewer Checklist**

### DIFF
--- a/test/e2e/scoped_client_test.go
+++ b/test/e2e/scoped_client_test.go
@@ -2,17 +2,14 @@ package e2e
 
 import (
 	"context"
-	"time"
-
 	. "github.com/onsi/ginkgo"
 	"github.com/onsi/ginkgo/extensions/table"
+	. "github.com/onsi/gomega"
 	"github.com/sirupsen/logrus"
-	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/rest"
 
@@ -22,17 +19,18 @@ import (
 	"github.com/operator-framework/operator-lifecycle-manager/test/e2e/ctx"
 )
 
-var _ = Describe("Scoped Client", func() {
+var _ = Describe("Scoped Client bound to a service account can be used to make API calls", func() {
 	// TestScopedClient ensures that we can create a scoped client bound to a
 	// service account and then we can use the scoped client to make API calls.
+	var (
+		config *rest.Config
 
-	var config *rest.Config
+		kubeclient    operatorclient.ClientInterface
+		crclient      versioned.Interface
+		dynamicclient dynamic.Interface
 
-	var kubeclient operatorclient.ClientInterface
-	var crclient versioned.Interface
-	var dynamicclient dynamic.Interface
-
-	var logger *logrus.Logger
+		logger *logrus.Logger
+	)
 
 	BeforeEach(func() {
 		config = ctx.Ctx().RESTConfig()
@@ -42,6 +40,7 @@ var _ = Describe("Scoped Client", func() {
 		dynamicclient = ctx.Ctx().DynamicClient()
 
 		logger = logrus.New()
+		logger.SetOutput(GinkgoWriter)
 	})
 
 	type testParameter struct {
@@ -55,30 +54,28 @@ var _ = Describe("Scoped Client", func() {
 		// scoped client has enough permission, we expect a NotFound error code.
 		// Otherwise, we expect a 'Forbidden' error code due to lack of permission.
 
-		table.Entry("ServiceAccountDoesNotHaveAnyPermission", testParameter{
+		table.Entry("returns error on API calls as ServiceAccount does not have any permission", testParameter{
 			// The service account does not have any permission granted to it.
 			// We expect the get api call to return 'Forbidden' error due to
 			// lack of permission.
-			name: "ServiceAccountDoesNotHaveAnyPermission",
 			assertFunc: func(errGot error) {
-				require.True(GinkgoT(), k8serrors.IsForbidden(errGot))
+				Expect(k8serrors.IsForbidden(errGot)).To(BeTrue())
 			},
 		}),
-		table.Entry("ServiceAccountHasPermission", testParameter{
+		table.Entry("successfully allows API calls to be made when ServiceAccount has permission", testParameter{
 			// The service account does have permission granted to it.
 			// We expect the get api call to return 'NotFound' error.
-			name: "ServiceAccountHasPermission",
 			grant: func(namespace, name string) (cleanup cleanupFunc) {
 				cleanup = grantPermission(GinkgoT(), kubeclient, namespace, name)
 				return
 			},
 			assertFunc: func(errGot error) {
-				require.True(GinkgoT(), k8serrors.IsNotFound(errGot))
+				Expect(k8serrors.IsNotFound(errGot)).To(BeTrue())
 			},
 		}),
 	}
 
-	table.DescribeTable("Test", func(tt testParameter) {
+	table.DescribeTable("API call using scoped client", func(tc testParameter) {
 		// Steps:
 		// 1. Create a new namespace
 		// 2. Create a service account.
@@ -91,10 +88,16 @@ var _ = Describe("Scoped Client", func() {
 		defer cleanupNS()
 
 		saName := genName("user-defined-")
-		sa, cleanupSA := newServiceAccount(GinkgoT(), kubeclient, namespace, saName)
+		sa, cleanupSA := newServiceAccount(kubeclient, namespace, saName)
 		defer cleanupSA()
 
-		waitForServiceAccountSecretAvailable(GinkgoT(), kubeclient, sa.GetNamespace(), sa.GetName())
+		By("Wait for ServiceAccount secret to be available")
+		Eventually(func() (*corev1.ServiceAccount, error) {
+			sa, err := kubeclient.KubernetesInterface().CoreV1().ServiceAccounts(sa.GetNamespace()).Get(context.TODO(), sa.GetName(), metav1.GetOptions{})
+			return sa, err
+		}).ShouldNot(WithTransform(func(v *corev1.ServiceAccount) []corev1.ObjectReference {
+			return v.Secrets
+		}, BeEmpty()))
 
 		strategy := scoped.NewClientAttenuator(logger, config, kubeclient, crclient, dynamicclient)
 		getter := func() (reference *corev1.ObjectReference, err error) {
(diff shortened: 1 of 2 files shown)


test/e2e/scoped_client_test.go
```diff
@@ -87,14 +87,14 @@ var _ = Describe("Scoped Client", func() {
...
+		_, cleanupNS := newNamespace(kubeclient, namespace)
 		defer cleanupNS()
 
 		saName := genName("user-defined-")
-		sa, cleanupSA := newServiceAccount(GinkgoT(), kubeclient, namespace, saName)
+		sa, cleanupSA := newServiceAccount(kubeclient, namespace, saName)
 		defer cleanupSA()
 
-		waitForServiceAccountSecretAvailable(GinkgoT(), kubeclient, sa.GetNamespace(), sa.GetName())
+		waitForServiceAccountSecretAvailable(kubeclient, sa.GetNamespace(), sa.GetName())
```
exdx, line 97: should we wrap this in an `Eventually`?

test/e2e/scoped_client_test.go
```diff
@@ -106,35 +106,36 @@ var _ = Describe("Scoped Client", func() {
...
+		testParam.assertFunc(errGot)
 	}, tableEntries...)
 })
 
-func waitForServiceAccountSecretAvailable(t GinkgoTInterface, client operatorclient.ClientInterface, namespace, name string) *corev1.ServiceAccount {
+func waitForServiceAccountSecretAvailable(client operatorclient.ClientInterface, namespace, name string) *corev1.ServiceAccount {
 	var sa *corev1.ServiceAccount
-	err := wait.Poll(5*time.Second, time.Minute, func() (bool, error) {
+
+	Eventually(func() (bool, error) {
```
benluddy, line 138: If this fails, the failure stack trace will start from here instead of the place that is calling `waitForServiceAccountSecretAvailable`. Gomega provides [EventuallyWithOffset](https://pkg.go.dev/github.com/onsi/gomega?tab=doc#EventuallyWithOffset) to remove helper functions from the error stack trace.

test/e2e/scoped_client_test.go
```diff
@@ -61,7 +61,7 @@ var _ = Describe("Scoped Client", func() {
 			// lack of permission.
 			name: "ServiceAccountDoesNotHaveAnyPermission",
 			assertFunc: func(errGot error) {
-				require.True(GinkgoT(), k8serrors.IsForbidden(errGot))
+				Expect(k8serrors.IsForbidden(errGot)).To(BeTrue())
```
benluddy, line 64: Side note: The api-machinery error classification utils are used throughout the end-to-end tests. It might be nice to have a matcher that accepts a `func(error) bool`, and tries to use reflection to grab the function's name (e.g. `IsForbidden`) to generate meaningful failure messages. It's not a necessity, but it would be nice to have better messages than "expected true but was false" without having to write them all manually.

test/e2e/scoped_client_test.go
```diff
@@ -106,35 +106,36 @@ var _ = Describe("Scoped Client", func() {
 			return
 		}
 
-		if tt.grant != nil {
```
benluddy, line 109: Nit: In my experience, `tt` or `tc` (for test case) are common identifiers for rows in table-driven tests. I'm not sure this test is made more readable by changing it.

test/e2e/scoped_client_test.go
```diff
@@ -61,7 +61,7 @@ var _ = Describe("Scoped Client", func() {
 			// lack of permission.
 			name: "ServiceAccountDoesNotHaveAnyPermission",
```
benluddy, line 62: The purpose of the `name` field on `testParameter` is served by the first argument to each call to `table.Entry`, so it can be removed. And a nit: those entry descriptions could better describe the behavior under test. They're analogous to `It` descriptions -- Ginkgo generates an `It` for each table entry.

test/e2e/scoped_client_test.go
```diff
@@ -106,35 +106,36 @@ var _ = Describe("Scoped Client", func() {
...
 		}
 
 		// We expect to get scoped client instance(s).
 		kubeclientGot, crclientGot, dynamicClientGot, errGot := strategy.AttenuateClient(getter)
-		require.NoError(GinkgoT(), errGot)
-		require.NotNil(GinkgoT(), kubeclientGot)
-		require.NotNil(GinkgoT(), crclientGot)
+		Expect(errGot).ToNot(HaveOccurred())
+		Expect(kubeclientGot).ToNot(BeNil())
+		Expect(crclientGot).ToNot(BeNil())
```
benluddy, line 118: There's no nil check for `dynamicClientGot` here.

test/e2e/scoped_client_test.go
```diff
@@ -73,12 +73,12 @@ var _ = Describe("Scoped Client", func() {
...
 			},
 			assertFunc: func(errGot error) {
-				require.True(GinkgoT(), k8serrors.IsNotFound(errGot))
+				Expect(k8serrors.IsNotFound(errGot)).To(BeTrue())
 			},
 		}),
 	}
 
-	table.DescribeTable("Test", func(tt testParameter) {
+	table.DescribeTable("API call using scoped client", func(testParam testParameter) {
```
benluddy, line 81: Would it be simpler and clearer if there were two separate specs instead of being a table with only two entries?

test/e2e/scoped_client_test.go
```diff
@@ -106,35 +106,36 @@ var _ = Describe("Scoped Client", func() {
...
 		_, errGot = dynamicClientGot.Resource(gvr).Namespace(namespace).Get(context.TODO(), genName("does-not-exist-"), metav1.GetOptions{})
-		require.Error(GinkgoT(), errGot)
-		tt.assertFunc(errGot)
+		Expect(errGot).To(HaveOccurred())
+		testParam.assertFunc(errGot)
 	}, tableEntries...)
 })
 
-func waitForServiceAccountSecretAvailable(t GinkgoTInterface, client operatorclient.ClientInterface, namespace, name string) *corev1.ServiceAccount {
+func waitForServiceAccountSecretAvailable(client operatorclient.ClientInterface, namespace, name string) *corev1.ServiceAccount {
```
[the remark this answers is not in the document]
exdx: I had the same suggestion

test/e2e/scoped_client_test.go
harishsurf: kept it as `tc`

test/e2e/scoped_client_test.go
harishsurf: Agreed, but I have used `By`s and addressed some of the above review comments which I feel has improved the readability of the test. I am inclined to retain this structure.

test/e2e/scoped_client_test.go
harishsurf: Are you referring to writing a [Custom Matcher](https://onsi.github.io/gomega/#adding-your-own-matchers). If so, could this be done as part of a dedicated PR for this specific change? I'll make a note of this and work on it on the side.

test/e2e/scoped_client_test.go
```diff
@@ -145,9 +146,7 @@ func waitForServiceAccountSecretAvailable(t GinkgoTInterface, client operatorcli
 		}
 
 		return false, nil
+	}, time.Minute, 5*time.Second).Should(BeTrue())
```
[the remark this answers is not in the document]
harishsurf: removed it

test/e2e/scoped_client_test.go
```diff
@@ -87,14 +87,14 @@ var _ = Describe("Scoped Client", func() {
 		// 5. Invoke Get API call on non existent object(s) to check if
 		//    the call can be made successfully.
 		namespace := genName("a")
-		_, cleanupNS := newNamespace(GinkgoT(), kubeclient, namespace)
+		_, cleanupNS := newNamespace(kubeclient, namespace)
 		defer cleanupNS()
```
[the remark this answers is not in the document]
harishsurf: This ties back to your comment on restructuring the test which I have commented [here](https://github.com/operator-framework/operator-lifecycle-manager/pull/1500#discussion_r435501710).

test/e2e/scoped_client_test.go
```diff
@@ -87,14 +87,14 @@ var _ = Describe("Scoped Client", func() {
...
 
 		saName := genName("user-defined-")
-		sa, cleanupSA := newServiceAccount(GinkgoT(), kubeclient, namespace, saName)
+		sa, cleanupSA := newServiceAccount(kubeclient, namespace, saName)
 		defer cleanupSA()
 
-		waitForServiceAccountSecretAvailable(GinkgoT(), kubeclient, sa.GetNamespace(), sa.GetName())
+		waitForServiceAccountSecretAvailable(kubeclient, sa.GetNamespace(), sa.GetName())
 
 		strategy := scoped.NewClientAttenuator(logger, config, kubeclient, crclient, dynamicclient)
```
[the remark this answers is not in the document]
harishsurf: I see it defines functions explicitly as opposed to `logrus.Logger` which uses `level` to categorize logs. Do I go ahead and create a PR to replace the logger for `ClientAttenuator` struct [here](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/lib/scoped/attenuator.go#L55)

test/e2e/scoped_client_test.go
```diff
@@ -87,67 +83,55 @@ var _ = Describe("Scoped Client", func() {
...
-		sa, cleanupSA := newServiceAccount(GinkgoT(), kubeclient, namespace, saName)
+		sa, cleanupSA := newServiceAccount(kubeclient, namespace, saName)
 		defer cleanupSA()
 
-		waitForServiceAccountSecretAvailable(GinkgoT(), kubeclient, sa.GetNamespace(), sa.GetName())
+		By("Wait for ServiceAccount secret to be available")
+		Eventually(func() (bool, error) {
+			sa, err := kubeclient.KubernetesInterface().CoreV1().ServiceAccounts(sa.GetNamespace()).Get(context.TODO(), sa.GetName(), metav1.GetOptions{})
+			Expect(err).ShouldNot(HaveOccurred())
+			if len(sa.Secrets) > 0 {
```
Bowenislandsong, line 97: Could put these 2 in the same if statement. there is also no error thrown (other than timeout) if there's no secret in the SA.

harishsurf: Clarified it offline. The current implementation seems reasonable

test/e2e/scoped_client_test.go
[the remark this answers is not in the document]
benluddy: I don't insist on it, but I think it's a good example of a small and useful refactoring. This is already a fairly big PR, though, so how about [setting the output](https://pkg.go.dev/github.com/sirupsen/logrus?tab=doc#Logger.SetOutput) of `logger` to `GinkgoWriter`?

test/e2e/scoped_client_test.go
harishsurf: Being done as part of #1730

go.sum
```diff
@@ -1055,6 +1055,7 @@ k8s.io/kube-openapi v0.0.0-20200410145947-61e04a5be9a6 h1:Oh3Mzx5pJ+yIumsAD0MOEC
 k8s.io/kube-openapi v0.0.0-20200410145947-61e04a5be9a6/go.mod h1:GRQhZsXIAJ1xR0C9bd8UpWHZ5plfAS9fzPjJuQ6JL3E=
 k8s.io/kubectl v0.18.0 h1:hu52Ndq/d099YW+3sS3VARxFz61Wheiq8K9S7oa82Dk=
 k8s.io/kubectl v0.18.0/go.mod h1:LOkWx9Z5DXMEg5KtOjHhRiC1fqJPLyCr3KtQgEolCkU=
+k8s.io/kubernetes v1.13.0 h1:qTfB+u5M92k2fCCCVP2iuhgwwSOv1EkAkvQY1tQODD8=
```
exdx, line 1058: should this be changing?

harishsurf: removed this change

PR #1504, opened 2020-05-06
Bug 1828550: safely deprecate stored versions of CRDs

exdx:
This PR adds back the ability for OLM to deprecate storage versions of CRDs during the CRD upgrade process. This was removed when adding support for v1 CRDs because the existing storage deprecation code could cause data loss issues - CRs of removed storage versions would remain in etcd but would not be `GET`able from the client, causing data loss. 

Implement a safe way to deprecate old storage versions of CRDs. This can be done in several ways, but [the k8s docs](https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning/#upgrade-existing-objects-to-a-new-stored-version) suggest writing each CR to the new storage version first before removing the old stored version. 

**Reviewer Checklist**


exdx:
Closing in favor of #1529

PR #1507, opened 2020-05-07
Replace assertions for csv-e2e to use Gomega's matcher library

harishsurf:
This PR replaces testify/require assertions with Gomega's assertions. This PR will be followed up with another PR that uses ginkgo's Container blocks to ensure BDD style structuring

**Reviewer Checklist**

### DIFF
--- a/test/e2e/csv_e2e_test.go
+++ b/test/e2e/csv_e2e_test.go
@@ -9,7 +9,6 @@ import (
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
-	"github.com/stretchr/testify/require"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
@@ -34,15 +33,21 @@ import (
 	"github.com/operator-framework/operator-lifecycle-manager/test/e2e/ctx"
 )
 
-var _ = Describe("CSV", func() {
+var _ = Describe("ClusterServiceVersion", func() {
+	var (
+		c   operatorclient.ClientInterface
+		crc versioned.Interface
+	)
+	BeforeEach(func() {
+
+		c = newKubeClient()
+		crc = newCRClient()
+	})
 	AfterEach(func() {
 		TearDown(testNamespace)
 	})
 
-	It("create with unmet requirements mini kube version", func() {
-
-		c := newKubeClient()
-		crc := newCRClient()
+	It("create with unmet requirements min kube version", func() {
 
 		depName := genName("dep-")
 		csv := v1alpha1.ClusterServiceVersion{
@@ -78,21 +83,20 @@ var _ = Describe("CSV", func() {
 		}
 
 		cleanupCSV, err := createCSV(c, crc, csv, testNamespace, false, false)
-		require.NoError(GinkgoT(), err)
+		Expect(err).ShouldNot(HaveOccurred())
 		defer cleanupCSV()
 
 		_, err = fetchCSV(crc, csv.Name, testNamespace, csvPendingChecker)
-		require.NoError(GinkgoT(), err)
+		Expect(err).ShouldNot(HaveOccurred())
 
 		// Shouldn't create deployment
-		_, err = c.GetDeployment(testNamespace, depName)
-		require.Error(GinkgoT(), err)
+		Consistently(func() bool {
+			_, err := c.GetDeployment(testNamespace, depName)
+			return k8serrors.IsNotFound(err)
+		}).Should(BeTrue())
 	})
 	// TODO: same test but missing serviceaccount instead
 	It("create with unmet requirements CRD", func() {
-
-		c := newKubeClient()
-		crc := newCRClient()
 
 		depName := genName("dep-")
 		csv := v1alpha1.ClusterServiceVersion{
@@ -139,20 +143,20 @@ var _ = Describe("CSV", func() {
 		}
 
 		cleanupCSV, err := createCSV(c, crc, csv, testNamespace, false, false)
-		require.NoError(GinkgoT(), err)
+		Expect(err).ShouldNot(HaveOccurred())
 		defer cleanupCSV()
 
 		_, err = fetchCSV(crc, csv.Name, testNamespace, csvPendingChecker)
-		require.NoError(GinkgoT(), err)
+		Expect(err).ShouldNot(HaveOccurred())
 
 		// Shouldn't create deployment
-		_, err = c.GetDeployment(testNamespace, depName)
-		require.Error(GinkgoT(), err)
+		Consistently(func() bool {
+			_, err := c.GetDeployment(testNamespace, depName)
+			return k8serrors.IsNotFound(err)
+		}).Should(BeTrue())
 	})
-	It("create with unmet permissions CRD", func() {
 
-		c := newKubeClient()
-		crc := newCRClient()
+	It("create with unmet permissions CRD", func() {
 
 		saName := genName("dep-")
 		permissions := []v1alpha1.StrategyDeploymentPermissions{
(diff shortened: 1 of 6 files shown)


test/e2e/e2e_test.go
```diff
@@ -42,7 +42,7 @@ var (
 
 func TestEndToEnd(t *testing.T) {
 	RegisterFailHandler(Fail)
-	SetDefaultEventuallyTimeout(1 * time.Minute)
+	SetDefaultEventuallyTimeout(5 * time.Minute)
```
benluddy, line 45: I mentioned this in one of the other PRs, but my preference is to keep the default timeout at 1 minute and make exceptions if there's a good reason to do it on a case-by-case basis.

test/e2e/csv_e2e_test.go
```diff
@@ -33,13 +32,22 @@ import (
...
+	})
 
-		defer cleaner.NotifyTestComplete(true)
+	AfterEach(func() {
+		cleaner.NotifyTestComplete(true)
+	})
 
-		c := newKubeClient()
-		crc := newCRClient()
+	It("create with unmet requirements mini kube version", func() {
```
benluddy, line 50: Took me a minute to understand why there would be a test specific to minikube...

```suggestion
	It("create with unmet requirements MinKubeVersion", func() {
```

At a later date, these spec descriptions could use some revision -- it's not clear from reading most of them what behavior of OLM is being tested.

test/e2e/csv_e2e_test.go
```diff
@@ -244,27 +242,22 @@ var _ = Describe("CSV", func() {
...
 		defer cleanupCSV()
 
 		_, err = fetchCSV(GinkgoT(), crc, csv.Name, testNamespace, csvPendingChecker)
-		require.NoError(GinkgoT(), err)
+		Expect(err).ShouldNot(HaveOccurred())
 
 		// Shouldn't create deployment
 		_, err = c.GetDeployment(testNamespace, depName)
-		require.Error(GinkgoT(), err)
+		Expect(err).Should(HaveOccurred())
```
benluddy, line 257: This assertion is not specific enough... it will pass if (for example) the API server is temporarily unreachable, even if the deployment was created. It'll also pass if a deployment ends up being created after some time. That's what the `Consistently` Gomega assertion is for.

test/e2e/csv_e2e_test.go
```diff
@@ -1317,28 +1274,27 @@ var _ = Describe("CSV", func() {
...
 			}
 			if len(fetched.Status.Namespaces) > 0 {
-				require.ElementsMatch(GinkgoT(), expectedOperatorGroupStatus.Namespaces, fetched.Status.Namespaces)
+				Expect(fetched.Status.Namespaces).Should(ConsistOf(expectedOperatorGroupStatus.Namespaces))
 				return true, nil
 			}
 			return false, nil
-		})
-		require.NoError(GinkgoT(), err)
+		}).Should(BeTrue())
```
benluddy, line 1297:
```
Eventually(func() ([]string, error) {
  og, err := crc.OperatorsV1().OperatorGroups(secondNamespaceName).Get(context.TODO(), operatorGroup.Name, metav1.GetOptions{})
  if err != nil {
    return nil, err
  }
  return og.Status.Namespaces, nil
}).Should(ConsistOf(expectedOperatorGroupStatus.Namespaces))
```

test/e2e/csv_e2e_test.go
```diff
@@ -33,13 +32,22 @@ import (
 	"github.com/operator-framework/operator-lifecycle-manager/test/e2e/ctx"
 )
 
-var _ = Describe("CSV", func() {
-	It("create with unmet requirements mini kube version", func() {
+var _ = Describe("CSV represents a specific version of an Operator running on a cluster", func() {
```
benluddy, line 35: Nit: I don't know that a container node description should necessarily describe the purpose of the object under test. It's sort of the subject of all its inner specs, so, IMO, it should either name a "thing" or a category of related behavior.

test/e2e/csv_e2e_test.go
```diff
@@ -3465,44 +3376,47 @@ func waitForDeployment(c operatorclient.ClientInterface, name string) error {
...
 			return false, err
 		}
 		return false, nil
-	})
+	}).Should(BeTrue())
+	return err
 }
 
-func waitForCSVToDelete(t GinkgoTInterface, c versioned.Interface, name string) error {
+func waitForCSVToDelete(c versioned.Interface, name string) error {
```
benluddy, line 3398: With this change, won't the return value always be nil? `err` is shadowed in the closure, so the `err` declared on L3399 is never assigned. Besides, the test will end immediately if the Eventually fails, so no return value ever be read.

I think the idiomatic way to adapt this is to replace calls to `waitForCSVToDelete` with Eventually assertions. If this is used all over the place, and the duplication is unbearable, then it could make sense to define a function `csvExists` and write `Eventually(csvExists).Should(BeFalse())`.

test/e2e/csv_e2e_test.go
```diff
@@ -3465,44 +3376,47 @@ func waitForDeployment(c operatorclient.ClientInterface, name string) error {
...
 		}
 		return false, nil
-	})
+	}).Should(BeTrue())
 
 	return err
 }
 
-func waitForNotFound(t GinkgoTInterface, getFunction func() error) error {
+func waitForNotFound(getFunction func() error) error {
```
benluddy, line 3416: The same basic comment from `waitForCSVToDelete` applies here. They're nearly the same function, but this one is apparently only called from a single place. Can we delete this function altogether and use Eventually directly at its one callsite?

test/e2e/csv_e2e_test.go
```diff
@@ -33,13 +32,22 @@ import (
...
+		crc = newCRClient()
+	})
+
+	AfterEach(func() {
+		cleaner.NotifyTestComplete(true)
+	})
 
-		c := newKubeClient()
-		crc := newCRClient()
+	It("create with unmet requirements MiniKubeVersion", func() {
```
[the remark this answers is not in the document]
Bowenislandsong: Ha, I just want to come over and say i got burned more than twice on this too.

test/e2e/csv_e2e_test.go
```diff
@@ -988,61 +967,61 @@ var _ = Describe("CSV", func() {
...
+		Expect(err).ShouldNot(HaveOccurred(), "error getting expected extension-apiserver-authentication-reader RoleBinding")
 
 		// Store the ca sha annotation
 		oldCAAnnotation, ok := dep.Spec.Template.GetAnnotations()[install.OLMCAHashAnnotationKey]
-		require.True(GinkgoT(), ok, "expected olm sha annotation not present on existing pod template")
+		Expect(ok).Should(BeTrue(), "expected olm sha annotation not present on existing pod template")
 
 		// Induce a cert rotation
 		now := metav1.Now()
 		fetchedCSV.Status.CertsLastUpdated = &now
```
Bowenislandsong, line 1012: a rather frequent flake happens here:    	
```
Error Trace:	csv_e2e_test.go:1036
  	            				runner.go:113
  	            				runner.go:64
  	            				it_node.go:26
  	            				spec.go:215
  	            				spec.go:138
  	            				spec_runner.go:200
  	            				spec_runner.go:170
  	            				spec_runner.go:66
  	            				suite.go:62
  	            				ginkgo_dsl.go:226
  	            				ginkgo_dsl.go:214
  	            				e2e_test.go:54
  	            	Operation cannot be fulfilled on clusterserviceversions.operators.coreos.com "hat-server45hc6": the object has been modified; please apply your changes to the latest version and try again 
```
Maybe fetch CSV right before updating it.

https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gs/origin-ci-test/pr-logs/pull/operator-framework_operator-lifecycle-manager/1592/pull-ci-operator-framework-operator-lifecycle-manager-master-e2e-aws-olm/1275777878046281728

benluddy: https://github.com/operator-framework/operator-lifecycle-manager/pull/1604

harishsurf: this is addressed in #1604

test/e2e/csv_e2e_test.go
harishsurf: Added `func csvExists`

test/e2e/csv_e2e_test.go
```diff
@@ -1051,40 +1030,35 @@ var _ = Describe("CSV", func() {
...
-		require.NoError(GinkgoT(), err)
+		Expect(err).ShouldNot(HaveOccurred())
 
 		// Wait for CSV success
 		fetchedCSV, err = fetchCSV(crc, csv.GetName(), testNamespace, func(csv *v1alpha1.ClusterServiceVersion) bool {
 			// Should create an APIService
 			apiService, err := c.GetAPIService(apiServiceName)
 			if err != nil {
-				require.True(GinkgoT(), k8serrors.IsNotFound(err))
+				Expect(k8serrors.IsNotFound(err)).Should(BeTrue())
```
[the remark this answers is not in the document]
harishsurf: The inner assertions were removed as part of a different PR fixing flaky tests

PR #1428, opened 2020-04-06
[release-4.4] Bug 1818850: Add spec validation and improved status for CatalogSources

ecordell:
**Description of the change:**
Cherry pick of #1414

**Motivation for the change:**

**Reviewer Checklist**
- [ ] Implementation matches the proposed design, or proposal is updated to match implementation
- [ ] Sufficient unit test coverage 
- [ ] Sufficient end-to-end test coverage
- [ ] Docs updated or added to `/docs` 
- [ ] Commit messages sensible and descriptive

PR #1508, opened 2020-05-07
Bug 1818851: feat(catalogs): add spec validation for sourcetypes

benluddy:
Backport of error messaging changes from https://github.com/operator-framework/operator-lifecycle-manager/pull/1428.


njhale:
/approve

At a glance, I think the error on the console test could be due to an overspecified expectation: https://github.com/openshift/console/blob/9d9397269afa57fd460aa87c73cc5e946f7e88d2/frontend/packages/operator-lifecycle-manager/integration-tests/scenarios/single-installmode.scenario.ts#L23

sdodson:
/retitle Bug 1818851: feat(catalogs): add spec validation for sourcetypes

PR #1509, opened 2020-05-08
Bug 1833419: Detect WebhookDescription changes in CSVs

awgreene:
Problem:
Once a CSV has reached the Succeeded phase, updating
WebhookDescriptions in the CSV on cluster does not cause OLM to update
existing ValidatingWebhookConfigurations or
MutatingWebhookConfigurations on cluster.

Solution:
Update OLM to calculate if a change to the WebhookDescription has
occurred when a CSV is updated and if so update the on cluster
resources.

### DIFF
--- a/pkg/controller/install/webhook.go
+++ b/pkg/controller/install/webhook.go
@@ -3,14 +3,17 @@ package install
 import (
 	"context"
 	"fmt"
+	"hash/fnv"
 
 	"github.com/operator-framework/api/pkg/operators/v1alpha1"
+	hashutil "github.com/operator-framework/operator-lifecycle-manager/pkg/lib/kubernetes/pkg/util/hash"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/ownerutil"
 
 	log "github.com/sirupsen/logrus"
 	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/rand"
 )
 
 func ValidWebhookRules(rules []admissionregistrationv1.RuleWithOperations) error {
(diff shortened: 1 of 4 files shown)


pkg/controller/install/webhook.go
```diff
@@ -163,6 +168,7 @@ func (i *StrategyDeploymentInstaller) createOrUpdateValidatingWebhook(ogNamespac
 }
 
 const WebhookDescKey = "webhookDescriptionGenerateName"
+const WebhookHashKey = "webhookDescriptionSHA"
```
ecordell, line 171: We should probably align with the naming convention for `pod-template-hash` (see the dep hash label too)

pkg/controller/operators/olm/apiservices.go
```diff
@@ -486,3 +486,34 @@ func (a *Operator) updateDeploymentSpecsWithApiServiceData(csv *v1alpha1.Cluster
 	}
 	return strategyDetailsDeployment, nil
 }
+
+func (a *Operator) areWebhooksAvailable(csv *v1alpha1.ClusterServiceVersion) (bool, error) {
```
ecordell, line 490: Won't this be an issue if there's an update to a webhook that changes the name of the webhook? I think we'll be left with an "old" webhook, right?

pkg/controller/install/webhook.go
awgreene: Understood, I'll make the following update:
```
const WebhookHashKey = "olm.webhook-description-hash"
```

pkg/controller/operators/olm/apiservices.go
awgreene: Yep - I'll update the PR to handle this case and make sure that the added e2e test checks for this.

pkg/controller/install/webhook.go
awgreene: Updated in the latest version of the PR.

pkg/controller/operators/olm/apiservices.go
awgreene: Updated in the latest version of the PR.

pkg/controller/operators/olm/operator.go
```diff
@@ -1721,13 +1721,14 @@ func (a *Operator) checkReplacementsAndUpdateStatus(csv *v1alpha1.ClusterService
 func (a *Operator) updateInstallStatus(csv *v1alpha1.ClusterServiceVersion, installer install.StrategyInstaller, strategy install.Strategy, requeuePhase v1alpha1.ClusterServiceVersionPhase, requeueConditionReason v1alpha1.ConditionReason) error {
 	apiServicesInstalled, apiServiceErr := a.areAPIServicesAvailable(csv)
 	strategyInstalled, strategyErr := installer.CheckInstalled(strategy)
+	webhooksInstalled, webhookErr := a.areWebhooksAvailable(csv)
```
ecordell, line 1724: should this happen after checking `strategyErr`?

test/e2e/webhook_e2e_test.go
```diff
@@ -26,7 +26,7 @@ const (
 	webhookName = "webhook.test.com"
 )
 
-var _ = Describe("CSVs with a Webhook", func() {
+var _ = FDescribe("CSVs with a Webhook", func() {
```
ecordell, line 29: need to remove focus to merge

PR #1513, opened 2020-05-12
Bug 1834136: fix(queues): use a single gc queue

njhale:

- Use a single GC queue for notifying CSVs of changes to resources that affect them
- Requeue events all-namespace queues if registered

### DIFF
--- a/pkg/controller/operators/olm/operator.go
+++ b/pkg/controller/operators/olm/operator.go
@@ -702,13 +702,14 @@ func (a *Operator) syncObject(obj interface{}) (syncError error) {
 			a.requeueOwnerCSVs(metaObj)
 		} else {
 			switch metaObj.(type) {
-			case *rbacv1.ClusterRole, *rbacv1.ClusterRoleBinding:
+			case *rbacv1.ClusterRole, *rbacv1.ClusterRoleBinding, *admissionregistrationv1.MutatingWebhookConfiguration, *admissionregistrationv1.ValidatingWebhookConfiguration:
 				resourceEvent := kubestate.NewResourceEvent(
 					kubestate.ResourceUpdated,
 					metaObj,
 				)
-				syncError = a.objGCQueueSet.RequeueEvent(ns, resourceEvent)
-				logger.Debugf("syncObject - requeued update event for %v, res=%v", resourceEvent, syncError)
+				if syncError = a.objGCQueueSet.RequeueEvent("", resourceEvent); syncError != nil {
+					logger.WithError(syncError).Warnf("failed to requeue gc event: %v", resourceEvent)
+				}
 				return
 			}
 		}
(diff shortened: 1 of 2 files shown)


pkg/controller/operators/olm/operator.go
```diff
@@ -702,13 +702,14 @@ func (a *Operator) syncObject(obj interface{}) (syncError error) {
...
 			switch metaObj.(type) {
-			case *rbacv1.ClusterRole, *rbacv1.ClusterRoleBinding:
+			case *rbacv1.ClusterRole, *rbacv1.ClusterRoleBinding, *admissionregistrationv1.MutatingWebhookConfiguration, *admissionregistrationv1.ValidatingWebhookConfiguration:
 				resourceEvent := kubestate.NewResourceEvent(
 					kubestate.ResourceUpdated,
 					metaObj,
 				)
-				syncError = a.objGCQueueSet.RequeueEvent(ns, resourceEvent)
-				logger.Debugf("syncObject - requeued update event for %v, res=%v", resourceEvent, syncError)
+				if syncError = a.objGCQueueSet.RequeueEvent("", resourceEvent); syncError != nil {
```
njhale, line 710: Sanity check: Does this requeue indefinitely?

pkg/controller/operators/olm/operator.go
```diff
@@ -950,36 +951,42 @@ func (a *Operator) handleClusterServiceVersionDeletion(obj interface{}) {
...
+		}
 	}
 
 	webhookSelector := labels.SelectorFromSet(ownerutil.OwnerLabel(clusterServiceVersion, v1alpha1.ClusterServiceVersionKind)).String()
 	mWebhooks, err := a.opClient.KubernetesInterface().AdmissionregistrationV1().MutatingWebhookConfigurations().List(context.TODO(), metav1.ListOptions{LabelSelector: webhookSelector})
 	if err != nil {
 		logger.WithError(err).Warn("cannot list MutatingWebhookConfigurations")
 	}
-	for _, webhook := range mWebhooks.Items {
-		syncError := a.objGCQueueSet.RequeueEvent("", kubestate.NewResourceEvent(kubestate.ResourceUpdated, &webhook))
```
njhale, line 972: If I'm not mistaken, the value of `webhook` was changing underneath the pointer since it's a free variable in the loop's body.

awgreene:
/hold
Waiting for @dinhxuanvu or @ecordell to get a second set of eyes on this.

pkg/controller/operators/olm/operator.go
```diff
@@ -950,36 +951,42 @@ func (a *Operator) handleClusterServiceVersionDeletion(obj interface{}) {
...
-		logger.Debugf("handleCSVdeletion - requeued update event for %v, res=%v", webhook, syncError)
+	for i := 0; i < len(mWebhooks.Items); i++ {
```
ecordell, line 974: you can still keep the `range`, just make sure to do `w := webhook` in the loop body

pkg/lib/queueinformer/resourcequeue.go
```diff
@@ -39,6 +39,11 @@ func (r *ResourceQueueSet) RequeueEvent(namespace string, resourceEvent kubestat
 	r.mutex.RLock()
 	defer r.mutex.RUnlock()
 
+	if queue, ok := r.queueSet[metav1.NamespaceAll]; len(r.queueSet) == 1 && ok {
```
ecordell, line 42: Why is this necessary? if `""` is passed as the namespace, the following block will requeue in the "allnamespace" queue

pkg/controller/operators/olm/operator.go
awgreene: Great catch!

pkg/lib/queueinformer/resourcequeue.go
njhale: because we also want all of the requeues that specifically mention a namespace to go to the all-namespace queue as well.

PR #1515, opened 2020-05-12
Bug 1833426: Update Webhook E2E Namespace Labels

awgreene:
Problem:
An E2E test that ensures that Webhooks are scoped properly when the
OperatorGroup uses a namespaceSelector fails because no namespaces
are identified as a part of the OperatorGroup.

Solution:
Rather than update the Namespace to include the label, simply create
the Namespace with the Label from the start.

### DIFF
--- a/test/e2e/webhook_e2e_test.go
+++ b/test/e2e/webhook_e2e_test.go
@@ -31,10 +31,29 @@ var _ = Describe("CSVs with a Webhook", func() {
 	var crc versioned.Interface
 	var namespace *corev1.Namespace
 	var nsCleanupFunc cleanupFunc
+	var nsLabels map[string]string
 	BeforeEach(func() {
 		c = newKubeClient(GinkgoT())
 		crc = newCRClient(GinkgoT())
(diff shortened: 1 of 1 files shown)


test/e2e/webhook_e2e_test.go
```diff
@@ -31,43 +31,42 @@ var _ = Describe("CSVs with a Webhook", func() {
...
+		Expect(err).Should(BeNil())
+		Expect(namespace).ShouldNot(BeNil())
+
+		nsCleanupFunc = func() {
+			err := c.KubernetesInterface().CoreV1().Namespaces().Delete(context.TODO(), namespace.GetName(), metav1.DeleteOptions{})
+			Expect(err).Should(BeNil())
+		}
 	})
 	AfterEach(func() {
 		if nsCleanupFunc != nil {
```
ecordell, line 59: should this ever be nil?

awgreene: It should not - but it prevents a potential nil pointer problem. It might be worth changing the if statement to `Expect(nsCleanupFunc).ShouldNot(BeNil())`

PR #1516, opened 2020-05-12
Bug 1833195: delete bundle objects after CSV gets deleted

exdx:
OLM now supports secrets and configmaps in the bundle. These should have ownerrefs associated with the CSV in the bundle. When the CSV is deleted, these objects should be deleted as well. Since they are namespaced scoped, the kube gc will delete them automatically. 

**Reviewer Checklist**

### DIFF
--- a/test/e2e/gc_e2e_test.go
+++ b/test/e2e/gc_e2e_test.go
@@ -3,11 +3,11 @@ package e2e
 import (
 	"context"
 	"fmt"
-	"github.com/operator-framework/api/pkg/operators/v1alpha1"
 
 	"github.com/blang/semver"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	. "github.com/operator-framework/operator-lifecycle-manager/test/e2e/dsl"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
@@ -16,11 +16,11 @@ import (
 	"k8s.io/apimachinery/pkg/util/rand"
 	apiregistrationv1 "k8s.io/kube-aggregator/pkg/apis/apiregistration/v1"
 
+	"github.com/operator-framework/api/pkg/operators/v1alpha1"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/clientset/versioned"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/operatorclient"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/ownerutil"
 	"github.com/operator-framework/operator-lifecycle-manager/test/e2e/ctx"
-	. "github.com/operator-framework/operator-lifecycle-manager/test/e2e/dsl"
 )
 
 var _ = Describe("Garbage collection for dependent resources", func() {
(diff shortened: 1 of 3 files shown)


test/e2e/gc_e2e_test.go
```diff
@@ -3,24 +3,22 @@ package e2e
...
-
 	"github.com/blang/semver"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	"github.com/operator-framework/api/pkg/operators/v1alpha1"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/clientset/versioned"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/operatorclient"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/ownerutil"
+	"github.com/operator-framework/operator-lifecycle-manager/test/e2e/ctx"
+	. "github.com/operator-framework/operator-lifecycle-manager/test/e2e/dsl"
```
awgreene, line 14: OLM resources are typically imported in a third import block

test/e2e/gc_e2e_test.go
```diff
@@ -284,4 +282,91 @@ var _ = Describe("Garbage collection for dependent resources", func() {
...
+			owner = newCSV("owner", testNamespace, "", semver.MustParse("0.0.0"), nil, nil, newNginxInstallStrategy("test", nil, nil))
+			// create all owners
+			fetchedOwner, err = operatorClient.OperatorsV1alpha1().ClusterServiceVersions(testNamespace).Create(context.TODO(), &owner, metav1.CreateOptions{})
+			Expect(err).NotTo(HaveOccurred())
+
+			// wait for csv install to complete
+			Eventually(func() bool {
+				csv, _ := operatorClient.OperatorsV1alpha1().ClusterServiceVersions(testNamespace).Get(context.TODO(), fetchedOwner.GetName(), metav1.GetOptions{})
+				return csv.Status.Phase == v1alpha1.CSVPhaseSucceeded
+			}).Should(BeTrue())
```
awgreene, line 311: Alternatively, there is a [fetchCSV](https://github.com/operator-framework/operator-lifecycle-manager/blob/d23b7ab12e3925a16bdb961cc9f8b8ddbad01cce/test/e2e/csv_e2e_test.go#L3409) function which waits for a CSV to reach the desired phase.

test/e2e/gc_e2e_test.go
```diff
@@ -284,4 +282,91 @@ var _ = Describe("Garbage collection for dependent resources", func() {
...
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      secretName,
+					Namespace: testNamespace,
+				},
+				Data: map[string][]byte{"name": []byte("Alice")},
+			}
+
+			// set ownerrefs to objects
+			ownerutil.AddOwner(cfg, fetchedOwner, true, false)
+			ownerutil.AddOwner(s, fetchedOwner, true, false)
```
awgreene, line 331: Nit: This tests ensures that secrets that are owned by the CSV are GC'd when the CSV is deleted, but it does not ensure that the Secrets have the ownerrefs. Could come back to bite us.

test/e2e/gc_e2e_test.go
```diff
@@ -284,4 +282,91 @@ var _ = Describe("Garbage collection for dependent resources", func() {
...
+			// Create a secret
+			secret, err = kubeClient.CreateSecret(s)
+			Expect(err).NotTo(HaveOccurred())
+
+			propagation = metav1.DeletePropagationForeground
+			options = metav1.DeleteOptions{PropagationPolicy: &propagation}
+		})
+
+		When("CSV is deleted", func() {
+
```
awgreene, line 346: Nit: Please remove the newline.

test/e2e/gc_e2e_test.go
```diff
@@ -284,4 +282,91 @@ var _ = Describe("Garbage collection for dependent resources", func() {
...
+			It("should delete the associated configmap and secret", func() {
+				_, err := kubeClient.GetConfigMap(testNamespace, configMap.GetName())
+				Expect(err).To(HaveOccurred())
+				Expect(k8serrors.IsNotFound(err)).To(BeTrue())
+				ctx.Ctx().Logf("configmap successfully garbage collected after csv deleted")
+
+				_, err = kubeClient.GetSecret(testNamespace, secret.GetName())
+				Expect(err).To(HaveOccurred())
+				Expect(k8serrors.IsNotFound(err)).To(BeTrue())
+				ctx.Ctx().Logf("secret successfully garbage collected after csv deleted")
```
awgreene, line 368: We may want to put this in an `Eventually` block.

test/e2e/gc_e2e_test.go
exdx: This was the result of `go fmt` but I can adjust it

test/e2e/gc_e2e_test.go
exdx: That is a good point, I'm wondering if there is a need for an additional follow-up test, I'll look into our existing tests and see if something like this exists. I imagine it would be a csv e2e test.

test/e2e/gc_e2e_test.go
exdx: Originally I did have these in `Eventually` blocks, but I realized checking it this way is actually more precise. Since we are using the foreground deletion option when deleting the CSV, that ensures that dependent objects (the secret and configmap) are GC'd before the CSV itself. Therefore once the CSV is gone we should be 100% sure that the dependent objects are gone as well. Checking it this way confirms that to be the case

awgreene: Ah great point!

test/e2e/gc_e2e_test.go
```diff
@@ -284,4 +285,104 @@ var _ = Describe("Garbage collection for dependent resources", func() {
...
+				},
+			}
+
+			source, err = operatorClient.OperatorsV1alpha1().CatalogSources(source.GetNamespace()).Create(context.TODO(), source, metav1.CreateOptions{})
+			Expect(err).ToNot(HaveOccurred(), "could not create catalog source")
+
+			// wait for catalog pods to come up
+			selector := labels.SelectorFromSet(map[string]string{"olm.catalogSource": sourceName})
+			_, err = awaitPods(GinkgoT(), kubeClient, source.GetNamespace(), selector.String(), podCount(1))
+			Expect(err).ToNot(HaveOccurred(), "catalog source pod not found")
```
awgreene, line 329: Do we need to wait for the pod to come up? Won't the subscription eventually succeed?

awgreene:
Unit tests are failing.

exdx:
> Unit tests are failing.

didn't check things again after rebasing off master 🤦

test/e2e/gc_e2e_test.go
exdx: That's an interesting thought - we could create them in parallel. I'm a little concerned about that though because I believe `Eventually` has a default 60 second timeout - its possible on a slower cluster it could flake on CI. Let me see if I can set the timeout.

exdx: Ok so I addressed this by relying on the default subscription checker polling which is 5 minutes. It did make the code cleaner this way, so nice catch.

test/e2e/gc_e2e_test.go
```diff
@@ -284,4 +284,101 @@ var _ = Describe("Garbage collection for dependent resources", func() {
...
+			Eventually(func() error {
+				_, err = fetchSubscription(operatorClient, testNamespace, subName, subscriptionStateAtLatestChecker)
+				return err
+			}).Should(BeNil())
+
+			// confirm extra bundle objects (secret and configmap) are installed
+			Eventually(func() bool {
+				_, err := kubeClient.GetSecret(testNamespace, secretName)
+				return k8serrors.IsNotFound(err)
+			}).Should(BeFalse())
```
benluddy, line 338: This assertion would pass if err was some non-nil value other than a not found error. Also, the message in the test report on failures would be something like "expected False but was True". Something like this:

```suggestion
			Eventually(func() error {
				_, err := kubeClient.GetSecret(testNamespace, secretName)
				return err
			}).Should(Succeed())
```
should give a more helpful failure message like "expected no error but was 'some error'".

test/e2e/gc_e2e_test.go
```diff
@@ -284,4 +284,101 @@ var _ = Describe("Garbage collection for dependent resources", func() {
...
+			// Create a Subscription for package
+			_ = createSubscriptionForCatalog(operatorClient, source.GetNamespace(), subName, source.GetName(), packageName, channelName, "", v1alpha1.ApprovalAutomatic)
+
+			// Wait for the Subscription to succeed
+			Eventually(func() error {
+				_, err = fetchSubscription(operatorClient, testNamespace, subName, subscriptionStateAtLatestChecker)
+				return err
+			}).Should(BeNil())
+
+			// confirm extra bundle objects (secret and configmap) are installed
```
harishsurf, line 334: Since you are verifying that your `CatalogSource` and `Subscription`  which is part of your _pre-setup_ created `secret` & `configMap`, it could be part of its own `It` spec.  For example, your test could look something like: 
```
When("a catalog source is created using grpc and a subscription", func() {
         BeforeEach(func(){
         // create catsrc and subscription  
         })
         It("should have created a secret and a configmap", func() {
                 // validate these objects
         })
```

test/e2e/gc_e2e_test.go
```diff
@@ -284,4 +284,101 @@ var _ = Describe("Garbage collection for dependent resources", func() {
...
 	})
 
+	Context("When installing a bundle with configmap and secret objects these objects should be deleted when the CSV is deleted", func() {
+		const (
+			sourceName    = "test-catalog"
+			imageName     = "quay.io/olmtest/single-bundle-index:objects"
+			packageName   = "busybox"
+			channelName   = "alpha"
+			subName       = "test-subscription"
+			csvName       = "busybox.v2.0.0"
```
harishsurf, line 294: It is better to declare variables in the Container blocks where they are used. For example, `csvName` could be declared inside of `When` block at line 346.

test/e2e/gc_e2e_test.go
```diff
@@ -284,4 +284,101 @@ var _ = Describe("Garbage collection for dependent resources", func() {
...
+			secretName    = "mysecret"
+			configmapName = "special-config"
+		)
+
+		var (
+			source *v1alpha1.CatalogSource
+			err    error
```
harishsurf, line 301: We only want to declare variables at Container level only if we want to share them between multiple `It` specs. err isn't something we would want to share across multiple specs. So it is advisable to declare it right before it is used. [See here](https://github.com/operator-framework/operator-lifecycle-manager/blob/db290a7d8503dcfa8f49cb976bcc3da1b054392b/test/e2e/metrics_e2e_test.go#L60)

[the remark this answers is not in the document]
exdx: Agreed - I hate declaring the error at the top scope (very unidiomatic) but I think I got bit by having it at a local scope in another test and changed it. Will change it back. Still getting the hang of the framework

test/e2e/gc_e2e_test.go
```diff
@@ -284,4 +284,101 @@ var _ = Describe("Garbage collection for dependent resources", func() {
 
 	})
 
+	Context("When installing a bundle with configmap and secret objects these objects should be deleted when the CSV is deleted", func() {
```
harishsurf, line 287: the `should be deleted when CSV is deleted` could just be part of the `It` spec description. 
The way the Container block descriptions are written is :
`Describe` blocks to describe the individual behaviors of your code and `Context` blocks exercise those behaviors under different circumstances. You could additionally use `When` for better readability. 
For example:
```
[Describe] Subscriptions create required objects from Catalogs
  [Context] Given a Namespace with an OwnNamespace OperatorGroup
      [When] a CatalogSource is created with a bundle that contains prometheus objects
         [It] should install the operator successfully
             [By] creating a subscription to the CatalogSource
         [It] should have created the expected prometheus objects
```
Also, note that: `Describe/ Context / When` are functionally identical and only differ in their names.

exdx:
Added two additional e2e tests for the 2 upgrade scenarios @ecordell mentioned.

Also added the ownership code to service objects in the bundle as well for consistency across all bundle objects OLM resolves.

test/e2e/gc_e2e_test.go
```diff
@@ -284,4 +284,280 @@ var _ = Describe("Garbage collection for dependent resources", func() {
...
+				// Ensure the new csv is installed
+				Eventually(func() error {
+					_, err := operatorClient.OperatorsV1alpha1().ClusterServiceVersions(testNamespace).Get(context.TODO(), newCSVname, metav1.GetOptions{})
+					return err
+				}).Should(BeNil())
+			})
+
+			It("OLM should have put new configmap in place", func() {
+				Eventually(func() error {
+					_, err := kubeClient.GetConfigMap(testNamespace, upgradedConfigMapName)
```
ecordell, line 556: should we check that the old one was deleted?

exdx: we can check that, sure

exdx:
Tests should pass once Quay is back in a healthy state

test/e2e/gc_e2e_test.go
```diff
@@ -284,4 +284,285 @@ var _ = Describe("Garbage collection for dependent resources", func() {
...
+				Eventually(func() bool {
+					cfg, err := kubeClient.GetConfigMap(testNamespace, configmapName)
+					if err != nil {
+						return false
+					}
+					// check data in configmap to ensure it is the new data (configmap was updated in the newer bundle)
+					// new value in the configmap is "updated-very-much"
+					data := cfg.Data["special.how"]
+					return data == "updated-very-much"
+				}).Should(BeTrue())
```
benluddy, line 470: I think both Expect and Eventually/Consistently can handle multiple return values and implicitly fail if anything after the first return value is non-nil, which would let you write this:

```suggestion
				Eventually(func() (string, error) {
					cfg, err := kubeClient.GetConfigMap(testNamespace, configmapName)
					if err != nil {
						return "", err
					}
					return cfg.Data["special.how"], nil
				}).Should(Equal("updated-very-much"))
```

exdx: nice

PR #1519, opened 2020-05-13
Bug 1822396: Delete subscription metric when an operator is uninstalled

anik120:
When an operator was subscribed to using a Subscription Object, the subscription_sync_total
metric was emitted whenever the Subscription Object was created/updated/deleted. This PR
updates that behaviour to emit the metric only when the Subscription object is created/updated,
and deletes the time series for that particular subscription when the subscription object is
deleted.

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/catalog/subscription/syncer.go
+++ b/pkg/controller/operators/catalog/subscription/syncer.go
@@ -50,7 +50,7 @@ func (s *subscriptionSyncer) Sync(ctx context.Context, event kubestate.ResourceE
 		return err
 	}
 
-	s.recordMetrics(res)
+	metrics.EmitSubMetric(res)
 
 	logger := s.logger.WithFields(logrus.Fields{
 		"reconciling": fmt.Sprintf("%T", res),
@@ -68,8 +68,10 @@ func (s *subscriptionSyncer) Sync(ctx context.Context, event kubestate.ResourceE
 		initial = initial.Add()
 	case kubestate.ResourceUpdated:
 		initial = initial.Update()
+		metrics.UpdateSubsSyncCounterStorage(res)
 	case kubestate.ResourceDeleted:
 		initial = initial.Delete()
+		metrics.DeleteSubsMetric(res)
 	}
 
 	reconciled, err := s.reconcilers.Reconcile(ctx, initial)
@@ -83,15 +85,6 @@ func (s *subscriptionSyncer) Sync(ctx context.Context, event kubestate.ResourceE
 	}).Debug("reconciliation successful")
 
 	return nil
-}
-
-func (s *subscriptionSyncer) recordMetrics(sub *v1alpha1.Subscription) {
-	// sub.Spec is not a required field.
-	if sub.Spec == nil {
-		return
-	}
-
-	metrics.CounterForSubscription(sub.GetName(), sub.Status.InstalledCSV, sub.Spec.Channel, sub.Spec.Package).Inc()
 }
 
 func (s *subscriptionSyncer) Notify(event kubestate.ResourceEvent) {
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -2,6 +2,7 @@ package metrics
 
 import (
 	"context"
+
 	"github.com/prometheus/client_golang/prometheus"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
@@ -168,7 +169,18 @@ var (
 		},
 		[]string{NAMESPACE_LABEL, NAME_LABEL, VERSION_LABEL, PHASE_LABEL, REASON_LABEL},
 	)
+
+	// subscriptionSyncCounters keeps a record of the promethues counters emitted by
+	// Subscription objects. The key of a record is the Subscription name, while the value
+	//  is struct containing label values used in the counter
+	subscriptionSyncCounters = make(map[string]subscriptionSyncLabelValues)
 )
+
+type subscriptionSyncLabelValues struct {
+	installedCSV string
+	pkg          string
+	channel      string
+}
 
 func RegisterOLM() {
 	prometheus.MustRegister(csvCount)
@@ -215,5 +227,45 @@ func EmitCSVMetric(oldCSV *olmv1alpha1.ClusterServiceVersion, newCSV *olmv1alpha
 	} else {
 		csvSucceededGauge.Set(0)
 		csvAbnormal.WithLabelValues(newCSV.Namespace, newCSV.Name, newCSV.Spec.Version.String(), string(newCSV.Status.Phase), string(newCSV.Status.Reason)).Set(1)
+	}
+}
+
+func EmitSubMetric(sub *olmv1alpha1.Subscription) {
+	if sub.Spec == nil {
+		return
+	}
+	SubscriptionSyncCount.WithLabelValues(sub.GetName(), sub.Status.InstalledCSV, sub.Spec.Channel, sub.Spec.Package).Inc()
+	if _, present := subscriptionSyncCounters[sub.GetName()]; !present {
+		subscriptionSyncCounters[sub.GetName()] = subscriptionSyncLabelValues{
+			installedCSV: sub.Status.InstalledCSV,
+			pkg:          sub.Spec.Package,
+			channel:      sub.Spec.Channel,
+		}
+	}
+}
+
+func DeleteSubsMetric(sub *olmv1alpha1.Subscription) {
+	if sub.Spec == nil {
+		return
+	}
+	SubscriptionSyncCount.DeleteLabelValues(sub.GetName(), sub.Status.InstalledCSV, sub.Spec.Channel, sub.Spec.Package)
+}
+
+func UpdateSubsSyncCounterStorage(sub *olmv1alpha1.Subscription) {
+	if sub.Spec == nil {
+		return
+	}
+	counterValues := subscriptionSyncCounters[sub.GetName()]
+
+	if sub.Spec.Channel != counterValues.channel ||
+		sub.Spec.Package != counterValues.pkg ||
+		sub.Status.InstalledCSV != counterValues.installedCSV {
+
+		// Delete metric will label values of old Subscription first
+		SubscriptionSyncCount.DeleteLabelValues(sub.GetName(), counterValues.installedCSV, counterValues.channel, counterValues.pkg)
+
+		counterValues.installedCSV = sub.Status.InstalledCSV
+		counterValues.pkg = sub.Spec.Package
+		counterValues.channel = sub.Spec.Channel
 	}
 }
--- a/test/e2e/metrics_e2e_test.go
+++ b/test/e2e/metrics_e2e_test.go
@@ -5,20 +5,23 @@ package e2e
 import (
 	"context"
 	"fmt"
+	"regexp"
+	"time"
+
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/net"
-	"regexp"
 
 	"github.com/operator-framework/api/pkg/operators/v1alpha1"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/clientset/versioned"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/operatorclient"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/metrics"
 	"github.com/operator-framework/operator-lifecycle-manager/test/e2e/ctx"
 )
 
-var _ = Describe("Metrics are generated for OLM pod", func() {
+var _ = Describe("Metrics are generated for OLM managed resources", func() {
 
 	var (
 		c   operatorclient.ClientInterface
@@ -69,7 +72,7 @@ var _ = Describe("Metrics are generated for OLM pod", func() {
 			It("generates csv_abnormal metric for OLM pod", func() {
 
 				// Verify metrics have been emitted for packageserver csv
-				Expect(getMetricsFromPod(c, getOLMPod(c), "8081")).To(And(
+				Expect(getMetricsFromPod(c, getPodWithLabel(c, "app=olm-operator"), "8081")).To(And(
 					ContainSubstring("csv_abnormal"),
 					ContainSubstring(fmt.Sprintf("name=\"%s\"", failingCSV.Name)),
 					ContainSubstring("phase=\"Failed\""),
@@ -91,7 +94,7 @@ var _ = Describe("Metrics are generated for OLM pod", func() {
 
 				It("deletes its associated CSV metrics", func() {
 					// Verify that when the csv has been deleted, it deletes the corresponding CSV metrics
-					Expect(getMetricsFromPod(c, getOLMPod(c), "8081")).ToNot(And(
+					Expect(getMetricsFromPod(c, getPodWithLabel(c, "app=olm-operator"), "8081")).ToNot(And(
 						ContainSubstring("csv_abnormal{name=\"%s\"", failingCSV.Name),
 						ContainSubstring("csv_succeeded{name=\"%s\"", failingCSV.Name),
 					))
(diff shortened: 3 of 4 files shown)


pkg/metrics/metrics.go
```diff
@@ -212,3 +208,16 @@ func EmitCSVMetric(oldCSV *olmv1alpha1.ClusterServiceVersion, newCSV *olmv1alpha
 		csvAbnormal.WithLabelValues(newCSV.Namespace, newCSV.Name, newCSV.Spec.Version.String(), string(newCSV.Status.Phase), string(newCSV.Status.Reason)).Set(1)
 	}
 }
+
+func EmitSubsMetrics(sub *olmv1alpha1.Subscription) {
+	// sub.Spec is not a required field.
+	if sub.Spec == nil {
+		return
+	}
```
awgreene, line 216: As far as I can tell, this is no longer true. I tried applying the following files:
```
$ cat empty.sub.cr.yaml 
kind: Subscription
metadata:
  name: my-operator-subsription
  namespace: default

$ oc apply -f empty.sub.cr.yaml 
The Subscription "my-operator-subsription" is invalid: spec: Required value
```

&

```
$ cat empty.sub.cr.yaml 
kind: Subscription
metadata:
  name: my-operator-subsription
  namespace: default
spec: {}

$ oc apply -f empty.sub.cr.yaml 
The Subscription "my-operator-subsription" is invalid: 
* spec.name: Required value
* spec.source: Required value
* spec.sourceNamespace: Required value
```

pkg/metrics/metrics.go
```diff
@@ -212,3 +208,16 @@ func EmitCSVMetric(oldCSV *olmv1alpha1.ClusterServiceVersion, newCSV *olmv1alpha
...
+
+	SubscriptionSyncCount.WithLabelValues(sub.GetName(), sub.Status.InstalledCSV, sub.Spec.Channel, sub.Spec.Package).Inc()
+}
+
+func DeleteSubsMetric(sub *olmv1alpha1.Subscription) {
+	SubscriptionSyncCount.DeleteLabelValues(sub.GetName(), sub.Status.InstalledCSV, sub.Spec.Channel, sub.Spec.Package)
+}
```
awgreene, line 223: Can we add a test to the [metrics-e2e test](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/test/e2e/metrics_e2e_test.go)? The test could:
- Create a sub
- check for the metric
- update the sub
- check for the metric
- delete the sub
- ensure the metrics does not exist

pkg/metrics/metrics.go
```diff
@@ -212,3 +208,16 @@ func EmitCSVMetric(oldCSV *olmv1alpha1.ClusterServiceVersion, newCSV *olmv1alpha
...
 	}
 }
+
+func EmitSubsMetrics(sub *olmv1alpha1.Subscription) {
+	// sub.Spec is not a required field.
+	if sub.Spec == nil {
+		return
+	}
+
+	SubscriptionSyncCount.WithLabelValues(sub.GetName(), sub.Status.InstalledCSV, sub.Spec.Channel, sub.Spec.Package).Inc()
```
awgreene, line 218: Note: sub.Spec.Channel and sub.Spec.Package are not required fields - these could be empty strings. Should we emit these metrics?

pkg/metrics/metrics.go
```diff
@@ -212,3 +208,16 @@ func EmitCSVMetric(oldCSV *olmv1alpha1.ClusterServiceVersion, newCSV *olmv1alpha
...
+}
+
+func DeleteSubsMetric(sub *olmv1alpha1.Subscription) {
+	SubscriptionSyncCount.DeleteLabelValues(sub.GetName(), sub.Status.InstalledCSV, sub.Spec.Channel, sub.Spec.Package)
```
awgreene, line 222: Note: sub.Spec.Channel and sub.Spec.Package are not required fields - these could be empty strings. Should we emit these metrics?

awgreene, line 222: sub.Spec.Channel could be an empty string, but we should be safe here because the CSV name is probably unique. Thoughts?

pkg/metrics/metrics.go
awgreene: @anik120 this is the only non-nit that I have on the PR, great work.

anik120:
@awgreene thanks for the review! I have addressed the nit comments, and I'm waiting for https://github.com/operator-framework/operator-lifecycle-manager/pull/1443/ to merge before I write the test you suggested.

pkg/metrics/metrics.go
anik120: @awgreene what is the issue if we emit empty string?

pkg/metrics/metrics.go
anik120: @awgreene added an e2e test case here: https://github.com/operator-framework/operator-lifecycle-manager/pull/1519/files#diff-41336190eb5305cb6fc38321ea468723R70-R84

anik120:
>@awgreene thanks for the review! I have addressed the nit comments, and I'm waiting for #1443 to merge before I write the test you suggested.

Since #1443 is not merged yet, adding to the existing portfolio that'll need to be incorporated into #1443

pkg/metrics/metrics.go
awgreene: Should we waste telemeter bandwith reporting a sub that doesn't have:

- A name
- A CSV that it installs
- Define a package or channel to pull from?

test/e2e/metrics_e2e_test.go
```diff
@@ -80,6 +97,18 @@ func getOLMPod(client operatorclient.ClientInterface) *corev1.Pod {
 	return &podList.Items[0]
 }
 
+func getCatalogPod(client operatorclient.ClientInterface) *corev1.Pod {
```
awgreene, line 100: This function seems to be extremely specific to a single test. It should probably be made more generic (takes a label) or just implemented inline.

anik120: @awgreene good catch. Updated the PR with a generic function.

pkg/metrics/metrics.go
anik120: @awgreene  isn't name, installedCSV required fields?

pkg/controller/operators/catalog/subscription/syncer.go
```diff
@@ -45,34 +45,37 @@ func (s *subscriptionSyncer) now() *metav1.Time {
 // Sync reconciles Subscription events by invoking a sequence of reconcilers, passing the result of each
 // successful reconciliation as an argument to its successor.
 func (s *subscriptionSyncer) Sync(ctx context.Context, event kubestate.ResourceEvent) error {
-	res := &v1alpha1.Subscription{}
-	if err := scheme.Convert(event.Resource(), res, nil); err != nil {
+	sub := &v1alpha1.Subscription{}
```
ecordell, line 48: these changes are unrelated to the bug (and pretty subjective), can you remove them so that we can focus on the fix when we are backporting?

anik120: @ecordell removed the changes (although I'd have liked to discuss the pros and cons of making changes like these if we had a bit more time)

pkg/metrics/metrics.go
awgreene: >If the Subscription is being created on cluster successfully, I'm not convinced we should choose tweak the metrics to not emit data point based on field values.
That would essentially mean we have subscriptions on cluster, but telemeter is unaware of them.

If we can use those incomplete metrics we omit to find identify something useful, then yes, we should emit them. Otherwise, we're just consuming our telemeter quota for no reason.

anik120: @awgreene so in summary: 

a) We could either not emit those metrics now, and if we run into a use case where we need those data points we can start emitting them too. But then they'd have to be backported into previous z stream releases (which we already know is an EXTREMELY delayed process, compared to how rapidly telemeter dashboards can be created)

b) We could just emit the metric since a subscription is being created (and note that it'll be deleted now once the subscription is deleted) which will use up some of the telemeter quota. Question here is: how often are we expecting these kind of subscriptions to be created (my guess is they won't be created in any prod clusters)? And as a result, how much quota are we expecting to use up for this edge case? 

I just want to make sure we don't put our cart before our horses and potentially expose ourselves to a bug back porting process.

pkg/controller/operators/catalog/subscription/syncer.go
```diff
@@ -69,7 +69,10 @@ func (s *subscriptionSyncer) Sync(ctx context.Context, event kubestate.ResourceE
 	case kubestate.ResourceUpdated:
 		initial = initial.Update()
 	case kubestate.ResourceDeleted:
-		initial = initial.Delete()
+		{
```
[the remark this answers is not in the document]
anik120: @awgreene yes. Keeping the statements for this `case` in a block of {}

pkg/metrics/metrics.go
awgreene: I am unconvinced that there will ever be a point that we will want to consume the metric when the installedCSV is blank.

However, given that the metrics are deleted when the subscription is deleted, I see no way that the metric could blow up. I also agree with your point about the ease of updating telemeter vs backporting changes. Given those points, it probably isn't worth delaying this bug fix.

pkg/controller/operators/catalog/subscription/syncer_test.go
```diff
@@ -6,11 +6,19 @@ import (
...
 	"github.com/operator-framework/api/pkg/operators/v1alpha1"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/kubestate"
 )
 
+const (
+	name             = "test-subscription"
+	packageName      = "test-package"
+	channel          = "test-channel"
+	installedCSVName = "test-csv"
+)
```
awgreene, line 20: non-blocking nit - these strings are only used in one place, could be inline.

pkg/controller/operators/catalog/subscription/syncer.go
```diff
@@ -45,34 +45,37 @@ func (s *subscriptionSyncer) now() *metav1.Time {
...
 	}
 
-	s.recordMetrics(res)
+	metrics.EmitSubsMetrics(subs)
 
 	logger := s.logger.WithFields(logrus.Fields{
-		"reconciling": fmt.Sprintf("%T", res),
-		"selflink":    res.GetSelfLink(),
+		"reconciling": fmt.Sprintf("%T", subs),
+		"selflink":    subs.GetSelfLink(),
```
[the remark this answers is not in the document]
ecordell: FYI, selflink is deprecated now.

pkg/controller/operators/catalog/subscription/syncer.go
```diff
@@ -70,6 +70,7 @@ func (s *subscriptionSyncer) Sync(ctx context.Context, event kubestate.ResourceE
 		initial = initial.Update()
 	case kubestate.ResourceDeleted:
 		initial = initial.Delete()
+		metrics.DeleteSubsMetric(res)
```
ecordell, line 73: This only deletes the time series when the subscription has been deleted, but from the bug report:

> Metrics for Subscriptions of CSVs that have been upgraded to new versions still hang around after new Subscription/CSV have succeeded. 

Because installedCSV is one of the labels/values, we would need to delete the metric (with any non-current installedCSV) on upgrade as well. Or have I misunderstood the bug?

anik120: @ecordell removing the metrics for the old subscription when a subscription was the original intention of the bug. 

Updated the PR.

test/e2e/metrics_e2e_test.go
```diff
@@ -61,15 +69,46 @@ var _ = Describe("Metrics", func() {
...
-		Expect(getMetricsFromPod(c, getOLMPod(c), "8081")).ToNot(And(
+		Expect(getMetricsFromPod(c, getPodWithLabel(c, "app=olm-operator"), "8081")).ToNot(And(
 			ContainSubstring("csv_abnormal{name=\"%s\"", failingCSV.Name),
 			ContainSubstring("csv_succeeded{name=\"%s\"", failingCSV.Name),
 		))
+
+		subscriptionCleanup, subscription := createSubscriptionWithInstalledCSV(GinkgoT(), crc, testNamespace, "metric-subscription", testPackageName, stableChannel, installedCSV, v1alpha1.ApprovalManual)
+
+		// Verify metrics have been emitted for subscription
+		Expect(getMetricsFromPod(c, getPodWithLabel(c, "app=catalog-operator"), "8081")).To(And(
```
njhale, line 80: Should these checks use the `Eventually` matcher instead? If I'm reading this correctly, if the metric isn't emitted on time, then the test will fail.

pkg/controller/operators/catalog/subscription/syncer.go
```diff
@@ -67,9 +69,21 @@ func (s *subscriptionSyncer) Sync(ctx context.Context, event kubestate.ResourceE
 	case kubestate.ResourceAdded:
 		initial = initial.Add()
 	case kubestate.ResourceUpdated:
+		oldSubscription, err := s.client.OperatorsV1alpha1().Subscriptions(res.GetNamespace()).Get(context.TODO(), res.GetName(), metav1.GetOptions{})
+		if err != nil {
+			logger.Errorf("Could not get old subscription %s in namespace %s. Skipping attempt to delete metrics", oldSubscription.GetName(), oldSubscription.GetNamespace())
+		} else {
+			if res.GetName() != oldSubscription.GetName() ||
```
njhale, line 76: `res.GetName() != oldSubscription.GetName()` should never happen, right?

Also, based on https://github.com/operator-framework/operator-lifecycle-manager/pull/1519/files#r431776633, I think we want to make sure `currentCSV != installedCSV` before deleting.

@ecordell does this line up with your comment?

pkg/controller/operators/catalog/subscription/syncer.go
```diff
@@ -66,10 +66,19 @@ func (s *subscriptionSyncer) Sync(ctx context.Context, event kubestate.ResourceE
 	switch event.Type() {
 	case kubestate.ResourceAdded:
 		initial = initial.Add()
+		metrics.EmitSubMetric(res)
 	case kubestate.ResourceUpdated:
+		oldSubscription, err := s.client.OperatorsV1alpha1().Subscriptions(res.GetNamespace()).Get(context.TODO(), res.GetName(), metav1.GetOptions{})
```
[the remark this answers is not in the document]
anik120: @njhale could you explain what you mean by "from the cache"? The only subscription I'm aware of is the one on cluster, which I decided to grab and delete the metrics for before updating the on cluster subscription.

pkg/metrics/metrics.go
```diff
@@ -217,3 +218,11 @@ func EmitCSVMetric(oldCSV *olmv1alpha1.ClusterServiceVersion, newCSV *olmv1alpha
 		csvAbnormal.WithLabelValues(newCSV.Namespace, newCSV.Name, newCSV.Spec.Version.String(), string(newCSV.Status.Phase), string(newCSV.Status.Reason)).Set(1)
 	}
 }
+
+func EmitSubMetric(sub *olmv1alpha1.Subscription) {
+	SubscriptionSyncCount.WithLabelValues(sub.GetName(), sub.Status.InstalledCSV, sub.Spec.Channel, sub.Spec.Package).Inc()
```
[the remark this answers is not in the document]
anik120: @njhale I see your point. If we have consensus that this PR will be a good place to include the namespace label [here](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/metrics/metrics.go#L153) I don't mind adding it.

test/e2e/metrics_e2e_test.go
```diff
@@ -61,15 +69,46 @@ var _ = Describe("Metrics", func() {
...
+			ContainSubstring("subscription_sync_total"),
+			ContainSubstring(fmt.Sprintf("%s=\"%s\"", metrics.NAME_LABEL, "metric-subscription")),
+			ContainSubstring(fmt.Sprintf("%s=\"%s\"", metrics.CHANNEL_LABEL, stableChannel)),
+			ContainSubstring(fmt.Sprintf("%s=\"%s\"", metrics.INSTALLED_LABEL, installedCSV)),
+			ContainSubstring(fmt.Sprintf("%s=\"%s\"", metrics.PACKAGE_LABEL, testPackageName)),
+		))
+
+		subscription.Status.InstalledCSV = installedCSVUpdated
```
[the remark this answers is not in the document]
anik120: @njhale I see what you mean https://github.com/operator-framework/operator-lifecycle-manager/pull/1519/checks?check_run_id=718197936#step:3:1582

Any ideas on how we could test the behavior on update in that case?

test/e2e/metrics_e2e_test.go
```diff
@@ -117,3 +156,35 @@ func getMetricsFromPod(client operatorclient.ClientInterface, pod *corev1.Pod, p
...
+		},
+	}
+
+	subscription, err := crc.OperatorsV1alpha1().Subscriptions(namespace).Create(context.TODO(), subscription, metav1.CreateOptions{})
+	require.NoError(t, err)
+	return buildSubscriptionCleanupFunc(crc, subscription), subscription
+}
+
+func updateSubscription(t GinkgoTInterface, crc versioned.Interface, subscription *v1alpha1.Subscription) {
+	_, err := crc.OperatorsV1alpha1().Subscriptions(subscription.GetNamespace()).Update(context.TODO(), subscription, metav1.UpdateOptions{})
```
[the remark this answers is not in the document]
anik120: It doesn't. TIL.

test/e2e/metrics_e2e_test.go
```diff
@@ -117,3 +156,35 @@ func getMetricsFromPod(client operatorclient.ClientInterface, pod *corev1.Pod, p
...
+			Package:                packageName,
+			Channel:                channel,
+			InstallPlanApproval:    approval,
+		},
+		Status: v1alpha1.SubscriptionStatus{
+			InstalledCSV: installed,
+		},
+	}
+
+	subscription, err := crc.OperatorsV1alpha1().Subscriptions(namespace).Create(context.TODO(), subscription, metav1.CreateOptions{})
```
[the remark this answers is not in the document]
anik120: Apparently it doesn't. Latest attempt was to update the spec.Package.

pkg/controller/operators/catalog/subscription/syncer.go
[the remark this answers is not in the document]
njhale: > could you explain what you mean by "from the cache"? 

I mean the one being passed from the informer to this `Sync` function (via `event.Resource()`.

> The only subscription I'm aware of is the one on cluster, which I decided to grab and delete the metrics for before updating the on cluster subscription.

So here you are being given `res.Resource()` which _hopefully_ represents the latest `Subscription` and you are comparing it against the `Subscription` returned from the client, which you _know_ is the latest. Since `res.Resource()` is coming from an informer -- a watch -- chances are pretty good that it's up to date, which means you'll only ever trigger the nested if-statement below in the rare case the `Subscription` is updated between when the event is picked up by the informer and when this sync is called.

pkg/metrics/metrics.go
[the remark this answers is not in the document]
njhale: We can potentially fix this in a separate bug.

test/e2e/metrics_e2e_test.go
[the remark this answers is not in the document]
njhale: could you actually install an operator? using busybox or hang:10?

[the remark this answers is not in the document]
anik120: @njhale do you mean create a subscription for an operator, and then upgrade the operator? 
That could be done, it just feels like an overkill to test metrics being omitted, which is why my original inclination was to keep the test limited to dealing with a standalone subscription object.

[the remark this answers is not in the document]
njhale: it does, but this is an e2e test and we are testing behaviors of the highest-scoped API. I'm all in for an alternative if we can think of one (that doesn't involve manually setting the status).

pkg/metrics/metrics.go
```diff
@@ -217,3 +229,34 @@ func EmitCSVMetric(oldCSV *olmv1alpha1.ClusterServiceVersion, newCSV *olmv1alpha
...
```
ecordell, line 234: `spec` is not a required field, so this would panic without 

```
	if sub.Spec == nil {
		return
	}
```

pkg/metrics/metrics.go
```diff
@@ -217,3 +229,34 @@ func EmitCSVMetric(oldCSV *olmv1alpha1.ClusterServiceVersion, newCSV *olmv1alpha
...
+		subscriptionSyncCounters[sub.GetName()] = subscriptionSyncLabelValues{
+			installedCSV: sub.Status.InstalledCSV,
+			pkg:          sub.Spec.Package,
+			channel:      sub.Spec.Channel,
+		}
+	}
+}
+
+func DeleteSubsMetric(sub *olmv1alpha1.Subscription) {
+	SubscriptionSyncCount.DeleteLabelValues(sub.GetName(), sub.Status.InstalledCSV, sub.Spec.Channel, sub.Spec.Package)
```
ecordell, line 245: The should check for nil on sub.status and sub.spec

pkg/metrics/metrics.go
```diff
@@ -217,3 +229,34 @@ func EmitCSVMetric(oldCSV *olmv1alpha1.ClusterServiceVersion, newCSV *olmv1alpha
...
+}
+
+func UpdateSubsSyncCounterStorage(sub *olmv1alpha1.Subscription) {
+	counterValues := subscriptionSyncCounters[sub.GetName()]
+
```
ecordell, line 250: sub.spec and sub.status can be nil

PR #1520, opened 2020-05-13
Bug 1812019: PackageManifests should include Keywords

awgreene:
Problem:
Although CSVs allow users to define keywords for their operators, these
keywords are not available in the PackageManifest created for the
operator.

Solution:
Copy the keywords from the CSV to the PackageManifest.


awgreene:
> this looks good, but we should probably go ahead and include the rest of the metadata: links, maintainers, maturity

Are these items that people would want to search on? I suppose Maturity may be, but it seems odd to allow people to search based on links or maintainers.

pkg/package-server/apis/operators/packagemanifest.go
```diff
@@ -49,6 +51,22 @@ func CreateCSVDescription(csv *operatorsv1alpha1.ClusterServiceVersion, csvJSON
 		desc.Icon = icons
 	}
 
+	desc.Links = make([]AppLink, len(csv.Spec.Links))
```
anik120, line 54: Can we not use [copy(dst, src []Type)](https://golang.org/pkg/builtin/#copy) here? Seems like both the slices are of the same type.

awgreene: They're similar but one of the types is provided via the packagemanfiest_types.go and the other is provide via `clusterserviceversion_types.go`.

anik120: Can we abstract this out to another function at least then? Seems like we're writing the same code in 4 places.

awgreene: The same code is written in two places on purpose. OLM stores different versions of the `PackageManifest` type (internal/v1alpha1).

PR #1521, opened 2020-05-14
Bug 1834936: fix(resolver): don't pick operator from same package when resolve requiredAPI

dinhxuanvu:
When resolver looks for an operator that provides requiredAPIs, it
doesn't exclude operators from the same package that the original
operator that requires those APIs. As a result, if there is an
existing version of that operator now provides those requiredAPIs
(due to API ownership transfer), resolver will select an older or
newer version of the same operator to fulfill that requiredAPIs
which leads to API ownership conflicts on providedAPIs.

The resolver will look up all latest channel entries that provides
the necessary requiredAPIs and filter out those from the same package
before go on with the operator selection.

Note to reviewer:
This PR is designed to be backport-friendly as it is planned to be backported all the way down to 4.2. As a result, there are significant amount of code which is here solely for backport reason. After the backport is done, there will be follow-up PR(s) to update OLM and registry and clean up the codebase in the master.

**Reviewer Checklist**


test/e2e/subscription_e2e_test.go
```diff
@@ -1265,6 +1265,145 @@ var _ = Describe("Subscription", func() {
...
+
+		catalogSourceName := genName("catsrc")
+		catsrc, cleanup := createInternalCatalogSource(GinkgoT(), kubeClient, crClient, catalogSourceName, testNamespace, manifests, []apiextensions.CustomResourceDefinition{crd, crd2}, []v1alpha1.ClusterServiceVersion{csvA, csvAB, csvB})
+		require.NotNil(GinkgoT(), catsrc)
+		require.NotNil(GinkgoT(), cleanup)
+		defer cleanup()
+
+		// Ensure that the catalog source is resolved before we create a subscription.
+		_, err := fetchCatalogSource(GinkgoT(), crClient, catsrc.GetName(), testNamespace, catalogSourceRegistryPodSynced)
+		require.NoError(GinkgoT(), err)
```
harishsurf, line 1366:
```suggestion
Expect(err).ShouldNot(HaveOccurred)
```

pkg/controller/registry/resolver/util_test.go
```diff
@@ -356,16 +359,107 @@ func NewFakeSourceQuerier(bundlesByCatalog map[CatalogKey][]*api.Bundle) *Namesp
...
 		sources[catKey] = source
 	}
-	return NewNamespaceSourceQuerier(sources)
+	return NewNamespaceSourceQuerier(sources, clients)
+}
+
+// SortBundleInPackageChannel will sort into map of package-channel key and list
+// sorted (oldest to latest version) of bundles as value
+func SortBundleInPackageChannel(bundles []*api.Bundle) map[string][]*api.Bundle {
+	sorted := map[string][]*api.Bundle{}
```
harishsurf, line 400: nit: `sortedBundle` for better readability

test/e2e/subscription_e2e_test.go
```diff
@@ -1265,6 +1265,145 @@ var _ = Describe("Subscription", func() {
...
+		catsrc2, cleanup2 := createInternalCatalogSource(GinkgoT(), kubeClient, crClient, catalogSourceName2, testNamespace, manifests2, []apiextensions.CustomResourceDefinition{crd2}, []v1alpha1.ClusterServiceVersion{csvC})
+		require.NotNil(GinkgoT(), catsrc)
+		require.NotNil(GinkgoT(), cleanup)
+		defer cleanup2()
+
+		// Ensure that the catalog source is resolved before we create a subscription.
+		_, err = fetchCatalogSource(GinkgoT(), crClient, catsrc2.GetName(), testNamespace, catalogSourceRegistryPodSynced)
+		require.NoError(GinkgoT(), err)
+
+		// Create a subscription that has a dependency
```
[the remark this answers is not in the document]
dinhxuanvu: Hey Harish. Thank you for the feedbacks on the e2e test case. This PR is designed to be able to backport all the way down to 4.2. As a result, I'm sticking to "vanilla" version of e2e test structure so that it can be backported easier. There will be a follow-up PR that will clean all of these up in master later. In fact, there are a big chuck of code here which are only here for backport reason and will be removed after the backport process done along side with the cleanup PR.

pkg/controller/registry/registry_client.go
```diff
@@ -0,0 +1,105 @@
...
+	registryapi "github.com/operator-framework/operator-registry/pkg/api"
+	"github.com/operator-framework/operator-registry/pkg/client"
+	opregistry "github.com/operator-framework/operator-registry/pkg/registry"
+)
+
+type ChannelEntryStream interface {
+	Recv() (*registryapi.ChannelEntry, error)
+}
+
+type RegistryClientInterface interface {
```
njhale, line 18: Let's compose the interface and types to avoid juggling two distinct clients (it looks like you were close already):

```go
import (
        "context"
	"fmt"
	"io"

        "google.golang.org/grpc"

	registryapi "github.com/operator-framework/operator-registry/pkg/api"
	"github.com/operator-framework/operator-registry/pkg/client"
	opregistry "github.com/operator-framework/operator-registry/pkg/registry"
)

type Interface interface { // Note: I hate this naming convention
    client.Interface // Include the method set of this other interface

    FindBundleThatProvides(ctx context.Context, group, version, kind, pkgName string) (*registryapi.Bundle, error)
    GetLatestChannelEntriesThatProvide(ctx context.Context, group, version, kind string) (*ChannelEntryIterator, error)
}

type Client struct {
    client.Interface // Embed an instance of the imported interface to satisfy its method set
}

func NewClientFromConn(conn *grpc.ClientConn) *Client {
    return &Client{
        Interface: client.NewClientFromConn(conn),
    }
}

var _ Interface = &Client{} // True
var _ client.Interface = &Client{} // Also true

// GetLatestChannelEntriesThatProvide uses registry client to get a list of
// latest channel entries that provide the requested API (via an iterator)
func (rc *Client) GetLatestChannelEntriesThatProvide(ctx context.Context, group, version, kind string) (*ChannelEntryIterator, error) {
	stream, err := rc.Registry.GetLatestChannelEntriesThatProvide(ctx, &registryapi.GetLatestProvidersRequest{Group: group, Version: version, Kind: kind})
	if err != nil {
		return nil, err
	}
	return NewChannelEntryIterator(stream), nil
}

// FindBundleThatProvides returns a bundle that provides the request API and
// doesn't belong to the provided package
func (rc *OLMRegistryClient) FindBundleThatProvides(ctx context.Context, group, version, kind, pkgName string) (*registryapi.Bundle, error) {
	it, err := rc.GetLatestChannelEntriesThatProvide(ctx, group, version, kind)
	if err != nil {
		return nil, err
	}
	entry := rc.filterChannelEntries(it, pkgName)
	if entry == nil {
		return nil, fmt.Errorf("Unable to find a channel entry which doesn't belong to package %s", pkgName)
	}
	bundle, err := rc.Registry.GetBundle(ctx, &registryapi.GetBundleRequest{PkgName: entry.PackageName, ChannelName: entry.ChannelName, CsvName: entry.BundleName})
	if err != nil {
		return nil, err
	}
	return bundle, nil
}
```

pkg/controller/registry/grpc/source.go
```diff
@@ -200,11 +202,12 @@ func (s *SourceStore) AsClients(namespaces ...string) map[resolver.CatalogKey]cl
 		}
 		for _, namespace := range namespaces {
 			if key.Namespace == namespace {
-				refs[key] = client.NewClientFromConn(source.Conn)
+				refsInterface[key] = client.NewClientFromConn(source.Conn)
+				refsClient[key] = registry.NewRegistryClient(client.NewClientFromConn(source.Conn))
```
njhale, line 206: If we [use composition](https://github.com/operator-framework/operator-lifecycle-manager/pull/1521/files#r428138508), we can avoid passing around two clients.

pkg/controller/registry/registry_client.go
```diff
@@ -0,0 +1,105 @@
...
+	var entry *opregistry.ChannelEntry
+	for e := it.Next(); e != nil; e = it.Next() {
+		if e.PackageName != pkgName {
+			entry = &opregistry.ChannelEntry{
+				PackageName: e.PackageName,
+				ChannelName: e.ChannelName,
+				BundleName:  e.BundleName,
+				Replaces:    e.Replaces,
+			}
+			break
```
benluddy, line 101: Does bailing out without reaching EOF on a gRPC stream introduce overhead on the server side? Presumably the server will clean up after a timeout, and I don't see anything on `ClientStream` to deal with cleanly closing from the client side.

pkg/controller/registry/resolver/util_test.go
```diff
@@ -356,16 +359,107 @@ func NewFakeSourceQuerier(bundlesByCatalog map[CatalogKey][]*api.Bundle) *Namesp
...
+		pkgChanKey := v.PackageName + "/" + v.ChannelName
+		b, ok := sorted[pkgChanKey]
+		if ok {
+			b = append(b, v)
+			sorted[pkgChanKey] = b
+		} else {
+			blist := []*api.Bundle{}
+			blist = append(blist, v)
+			sorted[pkgChanKey] = blist
+		}
```
benluddy, line 412:
```suggestion
		sorted[pkgChanKey] = append(sorted[pkgChanKey], v)
```

pkg/controller/registry/resolver/util_test.go
```diff
@@ -356,16 +359,107 @@ func NewFakeSourceQuerier(bundlesByCatalog map[CatalogKey][]*api.Bundle) *Namesp
 					return b, nil
 				}
 			}
+
 			return nil, fmt.Errorf("no bundle found")
 		}
+
+		client.FindBundleThatProvidesStub = func(ctx context.Context, groupOrName, version, kind, pkgName string) (*api.Bundle, error) {
```
benluddy, line 366: There's a lot to maintain here for a test fake. How about passing an `*api.Bundle` and `error` to `NewFakeSourceQuerier` and returning them here?

dinhxuanvu: This `NewFakeSourceQuerier` is used in many places. It is not worth it to change it and have to update the test cases everywhere. The code change is needlessly more. I also prefer we have some actual logic for stubs, not just return what is provided.

pkg/controller/registry/registry_client.go
```diff
@@ -0,0 +1,112 @@
...
+	registryapi "github.com/operator-framework/operator-registry/pkg/api"
+	"github.com/operator-framework/operator-registry/pkg/client"
+	opregistry "github.com/operator-framework/operator-registry/pkg/registry"
+)
+
+type ChannelEntryStream interface {
+	Recv() (*registryapi.ChannelEntry, error)
+}
+
+type RegistryClientInterface interface {
```
njhale, line 18: nit: exported name stutters `registry.RegistryClientInterface` (as opposed to `registry.ClientInterface`)

pkg/controller/registry/registry_client.go
```diff
@@ -0,0 +1,112 @@
...
+	return next
+}
+
+func (ceit *ChannelEntryIterator) Error() error {
+	return ceit.error
+}
+
+type RegistryClient struct {
+	client.Interface
+	Client *client.Client
```
njhale, line 53: you don't need both of these

pkg/controller/registry/registry_client.go
```diff
@@ -0,0 +1,112 @@
...
+	FindBundleThatProvides(ctx context.Context, group, version, kind, pkgName string) (*registryapi.Bundle, error)
+	GetLatestChannelEntriesThatProvide(ctx context.Context, group, version, kind string) (*ChannelEntryIterator, error)
+}
+
+type ChannelEntryIterator struct {
+	stream ChannelEntryStream
+	error  error
+}
+
+func NewChannelEntryIterator(stream ChannelEntryStream) *ChannelEntryIterator {
```
njhale, line 29: Let's add go-doc comments to all exported functions and methods.

pkg/controller/registry/registry_client.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+type RegistryClient struct {
+	client.Interface
+	Client *client.Client
+}
+
+func NewRegistryClient(client *client.Client) *RegistryClient {
+	return &RegistryClient{client, client}
+}
```
njhale, line 58: Why bother taking the client when you can just:

```go
func NewClientFromConn(conn *grpc.ClientConn) *Client {
    return &Client{
        Interface: client.NewClientFromConn(conn),
    }
}
```

Remember, since `RegistryClient` implements `client.Interface`, it can be used anywhere that accepts `client.Interface`; i.e. seems like we're still managing two separate clients.

pkg/controller/registry/registry_client.go
```diff
@@ -0,0 +1,112 @@
...
+
+var _ RegistryClientInterface = &RegistryClient{}
+var _ client.Interface = &RegistryClient{}
```
njhale, line 61: I added this to the example for show. The line above covers this check.

pkg/controller/registry/registry_client.go
dinhxuanvu: I have no way to access the client "underneath" the `client.Interface`. That's why I need both. The code simply won't compile.

pkg/controller/registry/registry_client.go
dinhxuanvu: ack

pkg/controller/registry/registry_client.go
dinhxuanvu: ack

pkg/controller/registry/registry_client.go
dinhxuanvu: ack

pkg/controller/registry/registry_client.go
njhale: Glancing at this a second time, I think we should just embed `*client.Client` since you need access to the underlying fields of the struct.

```go
type Client struct {
    *client.Client
}
```

pkg/controller/registry/registry_client.go
```diff
@@ -0,0 +1,112 @@
...
+
+// GetLatestChannelEntriesThatProvide uses registry client to get a list of
+// latest channel entries that provide the requested API (via an iterator)
+func (rc *RegistryClient) GetLatestChannelEntriesThatProvide(ctx context.Context, group, version, kind string) (*ChannelEntryIterator, error) {
+	stream, err := rc.Client.Registry.GetLatestChannelEntriesThatProvide(ctx, &registryapi.GetLatestProvidersRequest{Group: group, Version: version, Kind: kind})
```
njhale, line 66: If we embed `*client.Client`, we can access fields and methods directly:

```go
stream, err := rc.Client.Registry.GetLatestChannelEntriesThatProvide(ctx, &registryapi.GetLatestProvidersRequest{Group: group, Version: version, Kind: kind})
```

pkg/controller/registry/registry_client.go
```diff
@@ -0,0 +1,112 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return NewChannelEntryIterator(stream), nil
+}
+
+// FindBundleThatProvides returns a bundle that provides the request API and
+// doesn't belong to the provided package
+func (rc *RegistryClient) FindBundleThatProvides(ctx context.Context, group, version, kind, pkgName string) (*registryapi.Bundle, error) {
```
njhale, line 75: The method name and the `pkgName` variable are a somewhat deceptive.

pkg/controller/registry/resolver/querier.go
```diff
@@ -24,20 +28,20 @@ type SourceRef struct {
 }
 
 type SourceQuerier interface {
-	FindProvider(api opregistry.APIKey, initialSource CatalogKey) (*api.Bundle, *CatalogKey, error)
+	FindProvider(api opregistry.APIKey, initialSource CatalogKey, pkgName string) (*api.Bundle, *CatalogKey, error)
```
njhale, line 31: Again, `pkgName` feels deceptive. What do we think about something like `dependentPkgName`?

pkg/controller/registry/resolver/querier.go
```diff
@@ -1,4 +1,5 @@
-//go:generate go run github.com/maxbrunsfeld/counterfeiter/v6 -o fakes/fake_registry_client.go ../../../../vendor/github.com/operator-framework/operator-registry/pkg/client/client.go Interface
+//go:generate go run github.com/maxbrunsfeld/counterfeiter/v6 -o fakes/fake_registry_client.go ../../../../vendor/github.com/operator-framework/operator-registry/pkg/api.RegistryClient
+//go:generate go run github.com/maxbrunsfeld/counterfeiter/v6 -o fakes/fake_registry_interface.go ../../../../vendor/github.com/operator-framework/operator-registry/pkg/client/client.go Interface
```
njhale, line 2: Now I think we want a single fake that implements the new interface you added.

pkg/controller/registry/registry_client.go
dinhxuanvu: We will create a client anyway. Why do we need to import `grpc` in here just to create a client in the end?

pkg/controller/registry/resolver/querier.go
dinhxuanvu: I will go with excludingPkgName.

pkg/controller/registry/resolver/querier.go
dinhxuanvu: The `go:generate` line for that fake is included in `registry_client.go`.

test/e2e/subscription_e2e_test.go
```diff
@@ -1253,6 +1253,141 @@ var _ = Describe("Subscription", func() {
...
+					Plural:   crdPlural2,
+					Singular: crdPlural2,
+					Kind:     crdPlural2,
+					ListKind: "list" + crdPlural2,
+				},
+				Scope: "Namespaced",
+			},
+		}
+
+		// Create CSV
```
ecordell, line 1305: This test could use a few more comments and/or some better naming. I wrote this out to make sure I was following:

CatSrc:
 - Package A
    - Default Channel: Stable
    - Channel Stable:
       -  Operator A (Requires: CRD 1, CRD 2 )
    - Channel Alpha:
       -  Operator ABC (Provides: CRD 1, CRD 2 )
- Package B
    - Default Channel: Stable
    - Channel Stable: 
       - Operator B (Provides: CRD)

CatSrc2:
 - Package B
    - Default Channel: Stable
    - Channel Stable: 
       - Operator C (Provides: CRD 2)

Then create a subscription:

CatalogSource: CatSrc      
Package: Package A,
Channel:  Stable,
StartingCSV:   CSV A

Check installed:

CSV A, CSV B, CSV C

CSV A required B and C but didn't get them from Package A ✅

test/e2e/subscription_e2e_test.go
```diff
@@ -1253,6 +1253,141 @@ var _ = Describe("Subscription", func() {
...
+		// Check that a single catalog source was used to resolve the InstallPlan
+		_, err = fetchInstallPlan(GinkgoT(), crClient, subscription.Status.InstallPlanRef.Name, buildInstallPlanPhaseCheckFunc(v1alpha1.InstallPlanPhaseComplete))
+		require.NoError(GinkgoT(), err)
+		// Fetch CSVs
+		_, err = fetchCSV(GinkgoT(), crClient, csvB.Name, testNamespace, csvSucceededChecker)
+		require.NoError(GinkgoT(), err)
+		_, err = fetchCSV(GinkgoT(), crClient, csvC.Name, testNamespace, csvSucceededChecker)
+		require.NoError(GinkgoT(), err)
+		_, err = fetchCSV(GinkgoT(), crClient, csvA.Name, testNamespace, csvSucceededChecker)
+		require.NoError(GinkgoT(), err)
```
ecordell, line 1389: I realize this should fail resolution, but can you add one negative check to make sure ABC doesn't exist?

test/e2e/installplan_e2e_test.go
```diff
@@ -340,7 +340,6 @@ var _ = Describe("Install Plan", func() {
 			mainPackageStable := fmt.Sprintf("%s-stable", mainPackageName)
 			mainPackageBeta := fmt.Sprintf("%s-beta", mainPackageName)
 			dependentPackageStable := fmt.Sprintf("%s-stable", dependentPackageName)
-			dependentPackageBeta := fmt.Sprintf("%s-beta", dependentPackageName)
```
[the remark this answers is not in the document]
dinhxuanvu: This test is designed to test the CRD ownership for an operator that is replaced by a newer version. The dependent operator is not needed because we only test the main operator, not the dependent one. I wrote the test case and not sure why I added the dependent operator in the first place.
The older resolution would pick the older version of dependent operator but with the new resolution, it will pick the latest in the channel that provides the requiredAPIs. As a result, the current test will fail. I simply remove one version of dependent operator to address the issue.

PR #1529, opened 2020-05-18
[WIP] Bug 1828550: add storage migrator to OLM

exdx:
**Description of the change:**
Alternative to #1504 

**Motivation for the change:**

**Reviewer Checklist**
- [ ] Implementation matches the proposed design, or proposal is updated to match implementation
- [ ] Sufficient unit test coverage 
- [ ] Sufficient end-to-end test coverage
- [ ] Docs updated or added to `/docs` 
- [ ] Commit messages sensible and descriptive

PR #1535, opened 2020-05-20
Bug 1828550: add check for storage version changes when installing CRDs

exdx:
Minimal bugfix to avoid  potential dataloss when upgrading CRDs. Additional work in this area will be done as part of the CRD epic in the upcoming release. 

Related to #1529 #1504 

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/catalog/step.go
+++ b/pkg/controller/operators/catalog/step.go
@@ -139,7 +139,16 @@ func (b *builder) NewCRDV1Step(client apiextensionsv1client.ApiextensionsV1Inter
 					}
 				}
 
-				// TODO ensure stored version compatibility
+				// check to see if stored versions changed and whether the upgrade could cause potential data loss
+				safe, err := crdlib.SafeStorageVersionUpgrade(currentCRD, crd)
+				if !safe {
+					b.logger.Errorf("risk of data loss updating %s: %s", step.Resource.Name, err)
+					return v1alpha1.StepStatusUnknown, errorwrap.Wrapf(err, "risk of data loss updating %s", step.Resource.Name)
+				}
+				if err != nil {
+					return v1alpha1.StepStatusUnknown, errorwrap.Wrapf(err, "checking CRD for potential data loss updating %s", step.Resource.Name)
+				}
+
 				// Update CRD to new version
 				_, err = client.CustomResourceDefinitions().Update(context.TODO(), crd, metav1.UpdateOptions{})
 				if err != nil {
@@ -219,7 +228,16 @@ func (b *builder) NewCRDV1Beta1Step(client apiextensionsv1beta1client.Apiextensi
 						return v1alpha1.StepStatusUnknown, errorwrap.Wrapf(err, "error validating existing CRs agains new CRD's schema: %s", step.Resource.Name)
 					}
 				}
-				// TODO ensure stored version compatibility
+
+				// check to see if stored versions changed and whether the upgrade could cause potential data loss
+				safe, err := crdlib.SafeStorageVersionUpgrade(currentCRD, crd)
+				if !safe {
+					b.logger.Errorf("risk of data loss updating %s: %s", step.Resource.Name, err)
+					return v1alpha1.StepStatusUnknown, errorwrap.Wrapf(err, "risk of data loss updating %s", step.Resource.Name)
+				}
+				if err != nil {
+					return v1alpha1.StepStatusUnknown, errorwrap.Wrapf(err, "checking CRD for potential data loss updating %s", step.Resource.Name)
+				}
 
 				// Update CRD to new version
 				_, err = client.CustomResourceDefinitions().Update(context.TODO(), crd, metav1.UpdateOptions{})
--- a/pkg/lib/crd/storage.go
+++ b/pkg/lib/crd/storage.go
@@ -1,0 +1,68 @@
+package crd
+
+import (
+	"fmt"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+// SafeStorageVersionUpgrade determines whether the new CRD spec includes all the storage versions of the existing on-cluster CRD.
+// For each stored version in the status of the CRD on the cluster (there will be at least one) - each version must exist in the spec of the new CRD that is being installed.
+// See https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning/#upgrade-existing-objects-to-a-new-stored-version.
+func SafeStorageVersionUpgrade(existingCRD runtime.Object, newCRD runtime.Object) (bool, error) {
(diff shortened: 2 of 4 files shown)


pkg/controller/operators/catalog/step.go
```diff
@@ -139,7 +139,13 @@ func (b *builder) NewCRDV1Step(client apiextensionsv1client.ApiextensionsV1Inter
 					}
 				}
 
-				// TODO ensure stored version compatibility
+				// check to see if stored versions changed and whether the upgrade could cause potential data loss
+				err = crdlib.SafeStorageVersionUpgrade(currentCRD, crd)
+				if err != nil {
+					return v1alpha1.StepStatusUnknown, errorwrap.Wrapf(err, "potential for data loss updating: %s", step.Resource.Name)
```
ecordell, line 145: can we make this a little more descriptive? i.e. "potential for data loss: new CRD removes <version> that is listed a as a stored version on the existing CRD."

maybe even "consult docs for CRD storage version for how to safely remove." etc

exdx: so this wraps the error coming from the function which would add `"could not find existing stored version %s in the spec of the new CRD", name`. I can log the error out and take a look to make it as descriptive as possible 👍

pkg/lib/crd/storage.go
```diff
@@ -0,0 +1,66 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+// SafeStorageVersionUpgrade determines whether the new CRD spec includes all the storage versions of the existing on-cluster CRD.
+// For each stored version in the status of the CRD on the cluster (there will be at least one) - each version must exist in the spec of the new CRD that is being installed.
+// See https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning/#upgrade-existing-objects-to-a-new-stored-version.
+func SafeStorageVersionUpgrade(existingCRD runtime.Object, newCRD runtime.Object) error {
+	newSpecVersions, existingStatusVersions := getStoredVersions(existingCRD, newCRD)
+	if newSpecVersions == nil {
+		return fmt.Errorf("could not find a stored version in the spec of the new CRD: found %#v", newSpecVersions)
```
ecordell, line 16: the message could be better here - isn't it more like "couldn't find any versions in new CRD"?

pkg/lib/crd/storage.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+	if existingStatusVersions == nil {
```
ecordell, line 18: I think this should return nil - if there are no existing stored versions, we don't care what the new versions are, right?

test/e2e/crd_e2e_test.go
```diff
@@ -173,15 +170,153 @@ var _ = Describe("CRD Versions", func() {
...
+
+		// Check the error on the installplan - should be related to data loss and the CRD upgrade missing a stored version
+		Eventually(func() bool {
+			ip, err := crc.OperatorsV1alpha1().InstallPlans(testNamespace).Get(context.TODO(), s.Status.InstallPlanRef.Name, metav1.GetOptions{})
+			Expect(err).ToNot(HaveOccurred(), "could not get installplan")
+
+			Expect(ip.Status.Phase).To(Equal(operatorsv1alpha1.InstallPlanPhaseFailed))
+			Expect(ip.Status.Conditions[len(ip.Status.Conditions)-1].Message).To(ContainSubstring("potential for data loss"))
+			return true
+		}).Should(BeTrue())
```
ecordell, line 320: Can we test the case where the user has resolved the issue as well?

i.e. remove the version from old.status.storedVersions and ensure we can update?

pkg/lib/crd/storage.go
```diff
@@ -0,0 +1,66 @@
...
+			// potential data loss of CRs without a storage migration - throw error and block the CRD upgrade
+			return fmt.Errorf("new CRD removes version %s that is listed as a stored version on the existing CRD", name)
+		}
+	}
+
+	return nil
+}
+
+// getStoredVersions returns the storage versions listed in the status of the old on-cluster CRD
+// and the storage versions listed in the spec of the new CRD.
```
dinhxuanvu, line 34: Is this correct? The code return all versions in the spec of new CRD, not just storage version.

pkg/lib/crd/storage.go
```diff
@@ -0,0 +1,66 @@
...
+	case *apiextensionsv1beta1.CustomResourceDefinition:
+		for _, version := range crd.Status.StoredVersions {
+			existingStatusVersions[version] = struct{}{}
+		}
+	}
+
+	switch crd := newCRD.(type) {
+	case *apiextensionsv1.CustomResourceDefinition:
+		for _, version := range crd.Spec.Versions {
+			newSpecVersions[version.Name] = struct{}{}
```
dinhxuanvu, line 54: If you intend to add only storage version, then you need to check the `storage` field here.

pkg/lib/crd/storage.go
dinhxuanvu: In reality, this never happens given at least one version must be storage version. If user attempts to remove the only storage version in the status, I'm pretty sure it will error out on kubeclt.

exdx: yes, but is this an error on the kube side of things - every CRD on the cluster has at-least one storage version in its status right? I agree its not an OLM issue exactly. 

I guess returning early here is safe

test/e2e/crd_e2e_test.go
exdx: Sure -- I think I can add another test to verify this behavior.

pkg/lib/crd/storage.go
exdx: I changed this recently (when writing the unit tests) to pull all versions -- I think this is correct because there can only be one version set to the storage version on the spec of the CRD. So if we only look at storage version we only compare one version to the one or more versions on the status of the existing CRD.

I think we want to confirm all versions stored in the CRD on-cluster are present in the new CRD (many to many relationship).

pkg/lib/crd/storage.go
exdx: See above

pkg/lib/crd/storage.go
dinhxuanvu: Then please update the comment to reflect that all versions, not just storage versions.

pkg/controller/operators/catalog/step.go
```diff
@@ -139,7 +139,14 @@ func (b *builder) NewCRDV1Step(client apiextensionsv1client.ApiextensionsV1Inter
 					}
 				}
 
-				// TODO ensure stored version compatibility
+				// check to see if stored versions changed and whether the upgrade could cause potential data loss
+				err = crdlib.SafeStorageVersionUpgrade(currentCRD, crd)
```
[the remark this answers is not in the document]
exdx: That's an good point - originally I had the signature returning bool and error as you suggest. Can change it back.

pkg/lib/crd/storage.go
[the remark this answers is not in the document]
exdx: fixed

exdx:
I think this is close, just need to figure out how to get the second e2e test to pass.

exdx:
both e2e tests now pass locally -- once quay is back up e2e should pass, assuming no existing e2e tests do unsafe storage version upgrades with their CRDs.

PR #1542, opened 2020-05-22
Fix duplicated CRDs in quickstart manifests.

benluddy:
The script that generates the quickstart manifest files was dependent on an existing manifest file naming convention (CRD manifest files had the extension .crd.yaml). When the source of truth for CRD manifests was changed to point to the operator-framework/api package, CRD manifests were no longer adhering to that convention.

Since the per-release manifest directories are generated via `helm template --include-crds`, those CRD manifests are dumped into the same directory as non-CRD manifests.

Rather than attempting any YAML wizardry in `package_quickstart.sh` to separate CRD manifests from others, this PR simply restores the old naming convention for now.


.github/workflows/quickstart.yml
```diff
@@ -0,0 +1,23 @@
...
+        sudo mv kind /bin/
+        curl -LO "https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl"
+        chmod +x kubectl
+        sudo mv kubectl /bin/
+        kind create cluster
+        kind export kubeconfig
+        kubectl apply -f deploy/upstream/quickstart/crds.yaml
+        kubectl wait --timeout=5m --for=condition=Established crd $(kubectl get crd --output=jsonpath='{.items[*].metadata.name}')
+        kubectl apply -f deploy/upstream/quickstart/olm.yaml
+        kubectl wait --timeout=5m --for=condition=Available -n olm deploy olm-operator catalog-operator packageserver
```
kevinrizza, line 23: 👍

PR #1549, opened 2020-05-26
Bug 1838054: fix(catalog): no operatorgroups in a namespace should be an error when resolving

ecordell:
This is a 1-line fix plus tests, to avoid applying installplans when there is no operatorgroup in the namespace. 

There should be follow-up work to address the UX - this should probably fail earlier in the process with a message that users can understand (i.e. in status, on subscriptions, may an event)

InstallPlans were creating service accounts with RBAC in namespaces that don't have operatorgroups.

**Reviewer Checklist**

### DIFF
--- a/pkg/lib/scoped/querier_test.go
+++ b/pkg/lib/scoped/querier_test.go
@@ -1,0 +1,159 @@
+package scoped
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/sirupsen/logrus/hooks/test"
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	v1 "github.com/operator-framework/api/pkg/operators/v1"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/clientset/versioned"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/clientset/versioned/fake"
+)
+
+func TestUserDefinedServiceAccountQuerier(t *testing.T) {
+	tests := []struct {
+		name          string
+		crclient      versioned.Interface
+		namespace     string
+		wantReference *corev1.ObjectReference
+		wantErr       bool
+		err           error
+	}{
+		{
+			name:      "NoOperatorGroup",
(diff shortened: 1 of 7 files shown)


pkg/lib/scoped/querier_test.go
```diff
@@ -0,0 +1,159 @@
...
+			wantReference: &corev1.ObjectReference{
+				Kind:      "ServiceAccount",
+				Namespace: "ns",
+				Name:      "sa",
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			logger, _ := test.NewNullLogger()
```
Bowenislandsong, line 145: We probably want to start using test logger in every other test as well.

test/e2e/installplan_e2e_test.go
```diff
@@ -2621,7 +2621,75 @@ var _ = Describe("Install Plan", func() {
...
+		require.NotNil(GinkgoT(), subscription)
+
+		installPlanName := subscription.Status.InstallPlanRef.Name
+
+		fetchedInstallPlan, err := fetchInstallPlanWithNamespace(GinkgoT(), crc, installPlanName, ns.GetName(), buildInstallPlanPhaseCheckFunc(operatorsv1alpha1.InstallPlanPhaseInstalling))
+		require.NoError(GinkgoT(), err)
+		log(fmt.Sprintf("Install plan %s fetched with status %s", fetchedInstallPlan.GetName(), fetchedInstallPlan.Status.Phase))
+
+		fetchedInstallPlan, err = fetchInstallPlanWithNamespace(GinkgoT(), crc, installPlanName, ns.GetName(), buildInstallPlanPhaseCheckFunc(operatorsv1alpha1.InstallPlanPhaseInstalling))
+		require.NoError(GinkgoT(), err)
```
awgreene, line 2690: With the changes in this PR, does the InstallPlan remain in the `Installing` phase indefinitely?

ecordell: It will keep retrying with backoff, yes. 

The ideal thing here would be to transition to failed and only retry if we detect a change in operatorgroups. But currently the catalog operator doesn't watch operatorgroups at all, so that seemed like a bigger change than we should make for this bug.

PR #1558, opened 2020-05-29
Reduce install.sh flakiness.

benluddy:
Occasionally install.sh flakes out when a CRD takes too long to become
Established, or when the packageserver CSV takes longer than 30s to
reach the Succeeded phase. This commit adds a kubectl wait after
applying the CRDs and bumps the packageserver CSV timeout to 5
minutes, which should make the script more reliable.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -15,21 +15,25 @@ release=$1
 url=https://github.com/operator-framework/operator-lifecycle-manager/releases/download/${release}
 namespace=olm
 
-kubectl apply -f ${url}/crds.yaml
-kubectl apply -f ${url}/olm.yaml
+kubectl apply -f "${url}/crds.yaml"
+kubectl wait --for=condition=Established -f "${url}/crds.yaml"
+kubectl apply -f "${url}/olm.yaml"
 
 # wait for deployments to be ready
 kubectl rollout status -w deployment/olm-operator --namespace="${namespace}"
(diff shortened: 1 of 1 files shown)


scripts/install.sh
```diff
@@ -15,21 +15,25 @@ release=$1
...
+kubectl wait --for=condition=Established -f "${url}/crds.yaml"
+kubectl apply -f "${url}/olm.yaml"
 
 # wait for deployments to be ready
 kubectl rollout status -w deployment/olm-operator --namespace="${namespace}"
 kubectl rollout status -w deployment/catalog-operator --namespace="${namespace}"
 
-retries=50
-until [[ $retries == 0 || $new_csv_phase == "Succeeded" ]]; do
+retries=30
```
kevinrizza, line 26: This increases the length of this by a significant factor (from 50s to 300s). Do we have any metrics or anything that suggest doing anything more than say, multiplying by a factor of two is needed here?

benluddy: It's arbitrary, but depends on how often packageserver installation is expected to fail. Since this script is intended for installing from releases, failures should be the exception, so I would rather avoid confusing a potential new user over reporting failures a couple minutes earlier.

benluddy: I should add that in the linked issue, one person noted that they spent time poking around trying to understand why installation failed, and then within a few minutes it reached Succeeded on its own. That's the UX I would like to avoid here.

PR #1560, opened 2020-05-29
WIP: Fix generated sub name

awgreene:


awgreene:
Fixes https://github.com/operator-framework/operator-lifecycle-manager/issues/1559

pkg/controller/registry/resolver/steps.go
```diff
@@ -89,7 +89,7 @@ func NewSubscriptionStepResource(namespace string, info OperatorSourceInfo) (v1a
 	return NewStepResourceFromObject(&v1alpha1.Subscription{
 		ObjectMeta: metav1.ObjectMeta{
 			Namespace: namespace,
-			Name:      strings.Join([]string{info.Package, info.Channel, info.Catalog.Name, info.Catalog.Namespace}, "-"),
+			Name:      strings.ToLower(strings.Join([]string{info.Package, info.Channel, info.Catalog.Name, info.Catalog.Namespace}, "-")),
```
ecordell, line 92: While we're here - 

This should probably use a GenerateName instead, and should probably have protections around length (i.e. if I have too long of a package/channel/catalog name, this object could become invalid)

As it stands it would be possible (though unlikely) for two different packages to generate the same subscription.

njhale:
closing due to staleness

PR #1562, opened 2020-06-01
beta.kubernetes.io/os deprecated since v1.14, removed in v1.19

kramvan1:

Switch to supported os node selector, beta one is deprecated AND removed now.

https://github.com/kubernetes/kubernetes/pull/89460

**Reviewer Checklist**


kramvan1:
/assign @ecordell
@ecordell   Hi, this is now needed for kube 1.19, as the old beta value has now been removed from kube.

kramvan1:
@ecordell @Bowenislandsong   Could use a bit of help on why these tests are failing, I'm not as familiar with them.    Was hoping this was going to be a simple change, no logic was changed.

exdx:
> @ecordell @Bowenislandsong Could use a bit of help on why these tests are failing, I'm not as familiar with them. Was hoping this was going to be a simple change, no logic was changed.

Our tests flake with some regularity - sometimes it's instability on the cloud provider side. Your changes look to be fine and we appreciate the contribution

exdx:
Looks like there is an issue with the e2e tests if the upstream is failing as well. Potentially the nodes in the cluster don't carry this annotation and the pods fail to schedule?

kramvan1:
@exdx Hi, tried retest a couple times, not sure how many to do.  Any thoughts on this failure?

exdx:
> @exdx Hi, tried retest a couple times, not sure how many to do. Any thoughts on this failure?

Looks like the upstream e2e test is consistently failing so there must be an issue with our e2e tests. I'm going to pull this branch down and see if they pass.

exdx:
/retest

Looking at this more, I don't see any issues why the new node selectors would cause the tests to fail - the new selector is already present on all nodes in the cluster. The upstream e2e also passed. We can continue testing the openshift aws e2e and it will pass eventually I believe.

kramvan1:
@exdx  thx for the help here, we keep retesting for a while.

PR #1564, opened 2020-06-02
docs: add design docs for including additional objects in bundles

exdx:
Docs for including new objects in the bundle.

- PodDisruptionBudget (policy/v1beta1)
- PriorityClass (scheduling.k8s.io/v1alpha,v1beta1,v1)
- VerticalPodAutoScaler

Docs were written in a generic way to target both upstream and downstream OpenShift audiences. The underlying assumption is that we will allow users to include these objects in the bundle with no limitations on the actual content of these objects. 

**Reviewer Checklist**

### DIFF
--- a/doc/design/adding-pod-disruption-budgets.md
+++ b/doc/design/adding-pod-disruption-budgets.md
@@ -1,0 +1,54 @@
+# Adding Pod Disruption Budgets
+
+## Description
+
+OLM supports users including `PodDisruptionBudget` (PDB) objects in their bundle alongside their operator manifests. `PodDisruptionBudgets`
+are used to provide detailed information to the kube-scheduler about how many pods in a collection can be available or unavailable at given time. 
+For more info, see the docs at https://kubernetes.io/docs/tasks/run-application/configure-pdb/#protecting-an-application-with-a-poddisruptionbudget. 
+
+## Caveats
+
+PDBs are useful for configuring how many operator replicas or operands should run at any given time. However, it's important
+to set reasonable values for any PDBs included in the bundle and carefully consider how the PDB can affect the lifecycle of other resources
+in the cluster, such as nodes, to ensure cluster autoscaling and cluster upgrades are able to proceed if they are enabled. 
+
+PDBs are namespaced resources that only affect certain pods selected by the pod selector. However, 
+setting `maxUnavailable` to 0 or 0% (or `minAvailable` to 100%) on the PDB means zero voluntary evictions. 
+This can make a node impossible to drain and block important lifecycle actions like operator upgrades or even cluster upgrades. 
+
+Multiple PDBs can exist in one namespace- this can cause conflicts. For example, a PDB with the same name may already exist in the namespace.
+PDBs should target a unique collection of pods and not overlap with existing pods in the namespace. 
+Be sure to know of existing PDBs in the namespace in which your operator and operands will exist in the cluster. 
+
+PDBs for pods controlled by operators have additional restrictions around them. See https://kubernetes.io/docs/tasks/run-application/configure-pdb/#arbitrary-controllers-and-selectors
+for additional details - PDBs for operands managed by OLM-installed operators will fall into these restrictions. 
+
+## Technical Details
+
+PDB yaml manifests can be placed in the bundle alongside existing manifests in the `/manifests` directory. The PDB manifest will be stored 
+in the bundle image. 
+
+When OLM attempts to install the bundle, it will see the PDB and create it on-cluster. Since PDBs are namespace-scoped resources, 
+it will be created in the same namespace as the `InstallPlan` associated with the operator. The PDB will be visible in the `InstallPlan`
+and if the PDB fails to be installed OLM will provide a descriptive error in the `InstallPlan`. 
+
+OLM installs additional objects in the bundle after installing the CRDs and the CSV, to ensure proper owner references between the objects
+and the CSV. Therefore, there may be an initial period where additional objects are not available to the operator. 
+
+When the operator is removed, the PDB will be removed as well via the kubernetes garbage collector. The PDB will be updated when installing a newer version of the operator - 
+the existing PDB will be updated to the new PDB on-cluster. An upgrade to an operator bundle which does not include a PDB will remove the existing PDB from the cluster. 
+
+Prior versions of OLM (pre-0.16.0) do not support PDBs. If a PDB is present in a bundle attempting to be installed on-cluster, OLM will throw an invalid installplan error
+specifying that the resource is unsupported. 
+
+## Limitations on Pod Disruption Budgets
+
+No limitations are placed on the contents of a PDB at this time when installing on-cluster, but that may change as OLM develops
+an advanced strategy to ensure installed objects do not compromise the cluster. 
+
+However, the following are suggested guidelines to follow when including PDB objects in a bundle. 
+
+* `maxUnavailable` field cannot be set to 0 or 0%. 
+    * This can make a node impossible to drain and block important lifecycle actions like operator upgrades or even cluster upgrades.
+* `minAvailable` field cannot be set to 100%.
+    * This can make a node impossible to drain and block important lifecycle actions like operator upgrades or even cluster upgrades.
--- a/doc/design/adding-priority-classes.md
+++ b/doc/design/adding-priority-classes.md
@@ -1,0 +1,51 @@
+# Adding Priority Classes
+
+## Description
+
+OLM supports users including `PriorityClass` objects in their bundle alongside their operator manifests. `PriorityClass`
+is used to establish a priority, or weight, to a collection of pods in order to aid the kube-scheduler when assigning pods
+to nodes. For more info, see the docs at https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/#priorityclass. 
+
+## Caveats
+
+`PriorityClasses` are useful but also potentially far-reaching in nature. Be sure to understand the state of your cluster and
+your scheduling requirements before including one in your bundle alongside your operator. Best practice would be to 
+include a `PriorityClass` that only affects pods like your operator deployment and the respective operands. 
+
+`PriorityClass` objects are clusterwide in scope, meaning they can affect the scheduling of pods in all namespaces. Operators that specify a PriorityClass can affect other tenants on a multi-tenant cluster.
+All pods have a default priority of zero, and only those pods explicitly selected by the `PriorityClass` object will be given a priority when created.
+Existing pods running on the cluster are not affected by a new `PriorityClass`, but since clusters are dynamic and pods can be 
+rescheduled as nodes cycle in and out, a `PriorityClass` can have an impact on the long term behavior of the cluster. 
+
+Only one `PriorityClass` object in the cluster is allowed to have the `globalDefault` setting set to true. Attempting to install a `PriorityClass` with `globalDefault` set to true when one
+with `globalDefault` already exists on-cluster will result in a Forbidden error from the api-server. Setting `globalDefault` on a `PriorityClass` means that all pods in the cluster
+without an explicit priority class will use this default `PriorityClass`. 
+
+Pods with higher priorities can preempt pods with lower priorities when they are being scheduled onto nodes: preemption can result in lower-priority pods being evicted to make room for the higher priority pod. 
+If the `PriorityClass` of the pod is extremely high (higher than the priority of core components) scheduling the pod can potentially disrupt core components running in the cluster. 
+
+Once a `PriorityClass` is removed, no further pods can be created that reference the deleted `PriorityClass`. 
+
+## Technical Details
+
+`PriorityClass` yaml manifests can be placed in the bundle alongside existing manifests in the `/manifests` directory. The `PriorityClass` manifest will be present
+in the bundle image. 
+
+`PriorityClass` objects are clusterwide in scope, and will be applied by OLM directly to the cluster. The `PriorityClass` object will have
+a label referencing the operator that it is associated with. 
+
+OLM installs additional objects in the bundle after installing the CRDs and the CSV, to ensure proper owner references between the objects
+and the CSV. Therefore, there may be an initial period where additional objects are not available to the operator. 
+
+Prior versions of OLM (pre-0.16.0) do not support `PriorityClass` objects. If a `PriorityClass` is present in a bundle attempting to be installed on-cluster, OLM will throw an invalid installplan error
+specifying that the resource is unsupported. 
+
+## Limitations on Priority Classes 
+
+No limitations are placed on the contents of a `PriorityClass` manifest at this time when installing on-cluster, but that may change as OLM develops
+an advanced strategy to ensure installed objects do not compromise the cluster. 
+
+However, the following is a suggested guideline to follow when including `PriorityClass` objects in a bundle. 
+* `globalDefault` should always be `false` on a `PriorityClass` included in a bundle.
+    * Setting `globalDefault` on a `PriorityClass` means that all pods in the cluster without an explicit priority class will use this default `PriorityClass`. 
+    This can unintentionally affect other pods running in the cluster. 
--- a/doc/design/adding-vertical-pod-autoscaler.md
+++ b/doc/design/adding-vertical-pod-autoscaler.md
@@ -1,0 +1,45 @@
+# Adding Vertical Pod Autoscaler 
+
+## Description
+
+OLM supports users including `VerticalPodAutoscaler` (VPA) objects in their bundle alongside their operator manifests. `VerticalPodAutoscalers`
+objects are used to configure the VerticalPodAutoscaler controller to dynamically allocate resources to pods based on their usage of CPU, memory, 
+and other custom metrics. VPAs allow for more efficient use of cluster resources as pod resource needs are continually evaluated and adjusted by the VPA controller.
+For more info, see the docs at https://github.com/kubernetes/autoscaler/tree/master/vertical-pod-autoscaler. 
+
+## Caveats
+
+Adding a VPA object in your bundle can lead to more efficient use of resource in your cluster. Best practices include limiting
+the VPA to only the objects associated with your bundle. Consider your existing autoscaling setup in the cluster before adding
+VPA objects to a bundle and installing the bundle on the cluster. 
+
+`VerticalPodAutoscaler` objects watch a controller reference, such as deployment, to find a collection of pods to resize. Be sure to pass
+the appropriate reference to your operator or operands depending on which you would like the VPA to watch. 
+
+The VerticalPodAutoscaler controller must be enabled and active in the cluster for the VPA objects included in the bundle to have an effect. 
+Alternatively, the installing operator could also add the VPA as a required API to ensure the VPA operator is present in the cluster.
+
+The VPA will continually terminate pods and adjust the resource limits as needed - be sure your application is tolerant of restarts
+before including a VPA alongside it. 
+
+Note: at this time it is not recommended for the VPA to run alongside the HorizontalPodAutoscaler (HPA) on the same set of pods. 
+VPA can however be used with an HPA that is configured to use either external or custom metrics. 
+
+## Technical Details
+
+VPA yaml manifests can be placed in the bundle alongside existing manifests in the `/manifests` directory. The VPA manifest will be present
+in the bundle image. 
+
+VPA objects are clusterwide in scope, and will be applied by OLM directly to the cluster. The VPA object will have
+a label referencing the operator that it is associated with. 
+
+OLM installs additional objects in the bundle after installing the CRDs and the CSV, to ensure proper owner references between the objects
+and the CSV. Therefore, there may be an initial period where additional objects are not available to the operator. 
+
+Prior versions of OLM (pre-0.16.0) do not support VPA objects. If a VPA is present in a bundle attempting to be installed on-cluster, OLM will throw an invalid installplan error
+specifying that the resource is unsupported. 
+
+## Limitations on Vertical Pod Autoscalers 
+
+No limitations are placed on the contents of a VPA manifest at this time when installing on-cluster, but that may change as OLM develops
+an advanced strategy to ensure installed objects do not compromise the cluster. 


doc/design/adding-priority-classes.md
```diff
@@ -0,0 +1,47 @@
...
+to nodes. For more info, see the docs at https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/#priorityclass. 
+
+## Caveats
+
+`PriorityClass` objects are clusterwide in scope, meaning they can affect the scheduling of pods in all namespaces. All pods
+have a default priority of zero, and only those pods explicitly selected by the `PriorityClass` object will be given a priority when created.
+Existing pods running on the cluster are not affected by a new `PriorityClass`, but since clusters are dynamic and pods can be 
+rescheduled as nodes cycle in and out, a `PriorityClass` can have an impact on the long term behavior of the cluster. 
+
+Exactly one `PriorityClass` object is allowed to have the `globalDefault` setting set to true - this means that all pods on the cluster
```
awgreene, line 16: Could we explicitly call out what the effects of setting a Global Default may be? Should an operator author be allowed to configure the global default setting?

doc/design/adding-vertical-pod-autoscaler.md
```diff
@@ -0,0 +1,42 @@
...
+objects are used to configure the VerticalPodAutoscaler controller to dynamically allocate resources to pods based on their usage of CPU, memory, 
+and other custom metrics. VPAs allow for more efficient use of cluster resources as pod resource needs are continually evaluated and adjusted by the VPA controller.
+For more info, see the docs at https://github.com/kubernetes/autoscaler/tree/master/vertical-pod-autoscaler. 
+
+## Caveats
+
+`VerticalPodAutoscaler` objects watch a controller reference, such as deployment, to find a collection of pods to resize. Be sure to pass
+the appropriate reference to your operator or operands depending on which you would like the VPA to watch. 
+
+The VerticalPodAutoscaler controller must be enabled and active in the cluster for the VPA objects included in the bundle to have an effect.
```
awgreene, line 15: The operator could also add the VPA as a required API.

doc/design/adding-vertical-pod-autoscaler.md
```diff
@@ -0,0 +1,42 @@
...
+
+`VerticalPodAutoscaler` objects watch a controller reference, such as deployment, to find a collection of pods to resize. Be sure to pass
+the appropriate reference to your operator or operands depending on which you would like the VPA to watch. 
+
+The VerticalPodAutoscaler controller must be enabled and active in the cluster for the VPA objects included in the bundle to have an effect. 
+
+The VPA will continually terminate pods and adjust the resource limits as needed - be sure your application is tolerant of restarts
+before including a VPA alongside it. 
+
+Note: at this time it is not recommended for the VPA to run alongside the HorizontalPodAutoscaler.
```
awgreene, line 20: Is this a warning against using HPA and VPA on the same cluster or against targeting a single deployment with both?

doc/design/adding-pod-disruption-budgets.md
```diff
@@ -0,0 +1,45 @@
...
+in the built bundle image and will be placed into the index database when the bundle is added to an index. 
+
+When OLM attempts to install the bundle, it will see the PDB and create it on-cluster. Since PDBs are namespace-scoped resources, 
+it will be created in the same namespace as the `InstallPlan` associated with the operator. The PDB will be visible in the `InstallPlan`
+and if the PDB fails to be installed OLM will provide a descriptive error in the `InstallPlan`. 
+
+When the operator is removed, the PDB will be removed as well via the kubernetes garbage collector. The PDB will be replaced when installing a newer version of the operator - 
+the newer PDB will replace the existing PDB on-cluster. 
+
+Prior versions of OLM (pre-0.16.0) will not support creating PDBs from the bundle. If a PDB is present in the bundle, OLM will throw an invalid installplan error
```
awgreene, line 33: If OLM detects a PDB should it check the MinKubeVersion and make sure it is greater than 0.16.0? Perhaps not since it would be difficult to maintain a map as new GVKs emerge. Thoughts?

doc/design/adding-priority-classes.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Description
+
+OLM supports users including `PriorityClass` objects in their bundle alongside their operator manifests. `PriorityClass`
+is used to establish a priority, or weight, to a collection of pods in order to aid the kube-scheduler when assigning pods
+to nodes. For more info, see the docs at https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/#priorityclass. 
+
+## Caveats
+
+`PriorityClass` objects are clusterwide in scope, meaning they can affect the scheduling of pods in all namespaces. All pods
```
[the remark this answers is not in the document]
awgreene: I also find it confusing that a namespaced operator could affect all namespaces on the cluster.

doc/design/adding-priority-classes.md
```diff
@@ -0,0 +1,47 @@
...
+have a default priority of zero, and only those pods explicitly selected by the `PriorityClass` object will be given a priority when created.
+Existing pods running on the cluster are not affected by a new `PriorityClass`, but since clusters are dynamic and pods can be 
+rescheduled as nodes cycle in and out, a `PriorityClass` can have an impact on the long term behavior of the cluster. 
+
+Exactly one `PriorityClass` object is allowed to have the `globalDefault` setting set to true - this means that all pods on the cluster
+without an explicit priority class will use this default `PriorityClass`. Make sure you understand the effects of setting a global
+priority for pods running in the cluster before setting a `PriorityClass` as the `globalDefault`.
+
+Pods with higher priorities can preempt pods with lower priorities when they are being scheduled onto nodes: preemption can result in lower-priority pods being evicted to make room for the higher priority pod. 
+If the `PriorityClass` of the pod is extremely high (higher than the priority of core components) scheduling the pod can potentially disrupt core components running in the cluster.
```
awgreene, line 21: When the feature is added, could we make sure that a "How to set priorities" doc exists and describes priority ranges of core kube components?

doc/design/adding-priority-classes.md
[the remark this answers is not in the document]
exdx: I agree, it is one of the risks of introducing this object in the bundle

doc/design/adding-priority-classes.md
exdx: I believe this is one of the low hanging fruit that we can limit (no `globalDefault true` on a `PriorityClass`) IF we decide to limit these objects in some way.

doc/design/adding-vertical-pod-autoscaler.md
exdx: Good point

doc/design/adding-priority-classes.md
exdx: But from what I understand, at this moment, we are not going to do that

doc/design/adding-vertical-pod-autoscaler.md
exdx: Its the former - using both simultaneously can lead to a lot of churn on the cluster. Will clarify.

exdx: The latter. Updated for clarity.

doc/design/adding-pod-disruption-budgets.md
exdx: I'm not sure what you mean- this is referring to the OLM version not the kube version. I'm basically trying to state that older versions of OLM won't have this feature without saying OpenShift 4.6+ (to accommodate both upstream and downstream audiences).

doc/design/adding-priority-classes.md
exdx: Yes, I agree we need better docs on what objects are supported in bundles. Currently I'm not aware of any such docs and we now have multiple different sources for our docs so having an agreement on where these docs should live would be helpful.

exdx: That being said, if we deliver the "put arbitrary kube manifests in the bundle" feature in 4.6 then we won't really need said docs - we will simply say we support all objects. 

I also don't think we need to provide a "how to set priorities doc" since its not really in the scope of OLM. It's heavily dependent on the application needs.

doc/design/adding-priority-classes.md
```diff
@@ -0,0 +1,47 @@
...
+`PriorityClass` objects are clusterwide in scope, and will be applied by OLM directly to the cluster. The `PriorityClass` object will have
+a label referencing the operator that it is associated with, and will be manually garbage collected by OLM when the operator is removed. 
+
+Prior versions of OLM (pre-0.16.0) will not support creating `PriorityClass` from the bundle. If a `PriorityClass` is present in the bundle, OLM will throw an invalid installplan error
+specifying that the resource is unsupported. 
+
+## Limitations on Priority Classes 
+
+No limitations are placed on the contents of a `PriorityClass` manifest at this time, but that may change as OLM develops
+an advanced strategy to ensure installed objects do not compromise the cluster.
```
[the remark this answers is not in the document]
exdx: I see where you're going, but since `ResourceQuota` is not directly related to PriorityClasses (not mentioned in the k8s docs), and we don't support `ResourceQuota` objects in the bundle (as of this enhancement) I don't necessarily think it's a good idea to bring it up.

doc/design/adding-priority-classes.md
exdx: We should, yes. I'm not aware of such a document existing, but it should probably be featured prominently on our new doc site

exdx:
As this is doc-only we should be OK to merge without CI pending approval.

doc/design/adding-pod-disruption-budgets.md
```diff
@@ -0,0 +1,53 @@
...
+Multiple PDBs can exist in one namespace- this can cause conflicts. For example, a PDB with the same name may already exist in the namespace.
+PDBs should target a unique collection of pods and not overlap with existing pods in the namespace. 
+Be sure to know of existing PDBs in the namespace in which your operator and operands will exist in the cluster. 
+
+PDBs for pods controlled by operators have additional restrictions around them. See https://kubernetes.io/docs/tasks/run-application/configure-pdb/#arbitrary-controllers-and-selectors
+for additional details - PDBs for operands managed by OLM-installed operators will fall into these restrictions. 
+
+## Technical Details
+
+PDB yaml manifests can be placed in the bundle alongside existing manifests in the `/manifests` directory. The PDB manifest will be present
```
njhale, line 24: The PDB itself will be stored in the registry database? or just the bundle image?

doc/design/adding-pod-disruption-budgets.md
```diff
@@ -0,0 +1,53 @@
...
+in the built bundle image and will be placed into the index database when the bundle is added to an index. 
+
+When OLM attempts to install the bundle, it will see the PDB and create it on-cluster. Since PDBs are namespace-scoped resources,
```
njhale, line 27: Is there an ordering concern with PDBs and CSVs? i.e. is there a reason why I would want to apply a PDB _before_ a CSV it affects? Conversely, are there any reasons a CSV should not be created before a PDB it is selected by?

doc/design/adding-pod-disruption-budgets.md
```diff
@@ -0,0 +1,53 @@
...
+
+PDB yaml manifests can be placed in the bundle alongside existing manifests in the `/manifests` directory. The PDB manifest will be present
+in the built bundle image and will be placed into the index database when the bundle is added to an index. 
+
+When OLM attempts to install the bundle, it will see the PDB and create it on-cluster. Since PDBs are namespace-scoped resources, 
+it will be created in the same namespace as the `InstallPlan` associated with the operator. The PDB will be visible in the `InstallPlan`
+and if the PDB fails to be installed OLM will provide a descriptive error in the `InstallPlan`. 
+
+When the operator is removed, the PDB will be removed as well via the kubernetes garbage collector. The PDB will be replaced when installing a newer version of the operator - 
+the newer PDB will replace the existing PDB on-cluster. An upgrade to an operator bundle which does not include a PDB will remove the existing PDB from the cluster.
```
njhale, line 32: The garbage collector handles this too right? That means the PDB must have an OwnerReference to the CSV, which implies the CSV is created first.

njhale, line 32: > the newer PDB will replace the existing PDB on-cluster

update, patch, or delete+create?

doc/design/adding-pod-disruption-budgets.md
njhale: I understand the intent, but it is surprising to me that we're specifying the behavior of older versions in a design proposal for _newer_ versions. It makes me feel like we're changing something in older versions to get this behavior.

Might I suggest changing the wording to: "Prior versions of OLM **do not** support PDBs"

doc/design/adding-pod-disruption-budgets.md
```diff
@@ -0,0 +1,53 @@
...
+in the built bundle image and will be placed into the index database when the bundle is added to an index. 
+
+When OLM attempts to install the bundle, it will see the PDB and create it on-cluster. Since PDBs are namespace-scoped resources, 
+it will be created in the same namespace as the `InstallPlan` associated with the operator. The PDB will be visible in the `InstallPlan`
+and if the PDB fails to be installed OLM will provide a descriptive error in the `InstallPlan`. 
+
+When the operator is removed, the PDB will be removed as well via the kubernetes garbage collector. The PDB will be replaced when installing a newer version of the operator - 
+the newer PDB will replace the existing PDB on-cluster. An upgrade to an operator bundle which does not include a PDB will remove the existing PDB from the cluster. 
+
+Prior versions of OLM (pre-0.16.0) will not support creating PDBs from the bundle. If a PDB is present in the bundle, OLM will throw an invalid installplan error
```
njhale, line 34: I understand the intent, but it is surprising to me that we're specifying the behavior of older versions in a design proposal for newer versions. It makes me feel like we're changing something in older versions to get this behavior.

Might I suggest changing the wording to: "Prior versions of OLM do not support PDBs"

doc/design/adding-pod-disruption-budgets.md
```diff
@@ -0,0 +1,53 @@
...
+an advanced strategy to ensure installed objects do not compromise the cluster. 
+
+However, the following are suggested guidelines to follow when including PDB objects in a bundle. 
+
+* `maxUnavailable` field cannot be set to 0 or 0%. 
+    * This can make a node impossible to drain and block important lifecycle actions like operator upgrades or even cluster upgrades.
+* `minAvailable` field cannot be set to 100%.
+    * This can make a node impossible to drain and block important lifecycle actions like operator upgrades or even cluster upgrades.
+
+## Conclusions
```
njhale, line 49: nit: Not sure a design proposal should have a "conclusion". It's something we've already decided on doing, we're just hammering out the implementation details at this point. These are good points, so maybe they should live in the "caveats" section and drive design decisions on the restrictions PDBs installed by OLM should have, if any.

doc/design/adding-priority-classes.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Technical Details
+
+`PriorityClass` yaml manifests can be placed in the bundle alongside existing manifests in the `/manifests` directory. The `PriorityClass` manifest will be present
+in the built bundle image and will be placed into the index database when the bundle is added to an index. 
+
+`PriorityClass` objects are clusterwide in scope, and will be applied by OLM directly to the cluster. The `PriorityClass` object will have
+a label referencing the operator that it is associated with, and will be manually garbage collected by OLM when the operator is removed. 
+
+Prior versions of OLM (pre-0.16.0) will not support creating `PriorityClass` from the bundle. If a `PriorityClass` is present in the bundle, OLM will throw an invalid installplan error
```
njhale, line 32: see https://github.com/operator-framework/operator-lifecycle-manager/pull/1564/files#r447156885

doc/design/adding-priority-classes.md
```diff
@@ -0,0 +1,49 @@
...
+specifying that the resource is unsupported. 
+
+## Limitations on Priority Classes 
+
+No limitations are placed on the contents of a `PriorityClass` manifest at this time when installing on-cluster, but that may change as OLM develops
+an advanced strategy to ensure installed objects do not compromise the cluster. 
+
+However, the following is a suggested guideline to follow when including `PriorityClass` objects in a bundle.
```
njhale, line 40: So here's an edge case that I think we should cover: 
1. I install an operator with a `globalDefault=true` PDB
2. An update is pushed that changes the name of the PDB
3. _Boom_?

doc/design/adding-priority-classes.md
```diff
@@ -0,0 +1,49 @@
...
+to nodes. For more info, see the docs at https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/#priorityclass. 
+
+## Caveats
+
+`PriorityClass` objects are clusterwide in scope, meaning they can affect the scheduling of pods in all namespaces. Operators that specify a PriorityClass can affect other tenants on a multi-tenant cluster.
+All pods have a default priority of zero, and only those pods explicitly selected by the `PriorityClass` object will be given a priority when created.
+Existing pods running on the cluster are not affected by a new `PriorityClass`, but since clusters are dynamic and pods can be 
+rescheduled as nodes cycle in and out, a `PriorityClass` can have an impact on the long term behavior of the cluster. 
+
+Exactly one `PriorityClass` object is allowed to have the `globalDefault` setting set to true. Setting `globalDefault` on a `PriorityClass` means that all pods in the cluster
```
njhale, line 16: What happens if I try to apply a `PriorityClass` with `globalDefault=true` when one already exists. Will my request be rejected?

doc/design/adding-priority-classes.md
```diff
@@ -0,0 +1,49 @@
...
+
+No limitations are placed on the contents of a `PriorityClass` manifest at this time when installing on-cluster, but that may change as OLM develops
+an advanced strategy to ensure installed objects do not compromise the cluster. 
+
+However, the following is a suggested guideline to follow when including `PriorityClass` objects in a bundle. 
+* `globalDefault` should always be `false` on a `PriorityClass` included in a bundle.
+    * Setting `globalDefault` on a `PriorityClass` means that all pods in the cluster without an explicit priority class will use this default `PriorityClass`. 
+    This can unintentionally affect other pods running in the cluster. 
+
+## Conclusion
```
njhale, line 45: see [my other comment](https://github.com/operator-framework/operator-lifecycle-manager/pull/1564/files#r447158468)

doc/design/adding-vertical-pod-autoscaler.md
njhale: Do we mean as a native API?

doc/design/adding-vertical-pod-autoscaler.md
```diff
@@ -0,0 +1,44 @@
...
+
+The VPA will continually terminate pods and adjust the resource limits as needed - be sure your application is tolerant of restarts
+before including a VPA alongside it. 
+
+Note: at this time it is not recommended for the VPA to run alongside the HorizontalPodAutoscaler (HPA) on the same set of pods. 
+VPA can however be used with an HPA that is configured to use either external or custom metrics. 
+
+## Technical Details
+
+VPA yaml manifests can be placed in the bundle alongside existing manifests in the `/manifests` directory. The VPA manifest will be present
```
njhale, line 26: Can we get into a situation where we have contradictory VPAs installed? Are there really no sticky situations we can get into when dealing with VPAs during an upgrade; e.g. an old VPA prevents a new CSV's pods from scaling up?

Is the idea to pass the buck on these questions until we have a generic way to specify install behavior?

doc/design/adding-pod-disruption-budgets.md
exdx: I was under the impression that it would be both, but after we discussed the changes to the registry DB I believe it will be stored in just the bundle image. I will make a note of this.

doc/design/adding-pod-disruption-budgets.md
exdx: This is a really good point. Right now OLM creates the CRDs first, then the CSV, then all other objects (because of the ownerreferences that the extra objects require, which in turn require a UUID of the CSV). So there will potentially be a period where these objects are unavailable in the current implementation. I can make a note of this.

doc/design/adding-pod-disruption-budgets.md
exdx: CSV is created first.

exdx: Looks like a straight `Update()` according to https://github.com/operator-framework/operator-lifecycle-manager/blob/afd23480a756d8db5e7ce08869ae882dcdba0ed8/pkg/controller/operators/catalog/step_ensurer.go#L320

doc/design/adding-pod-disruption-budgets.md
exdx: I removed the `Conclusion` section and moved the content into `Caveats` - you're right, it reads better this way

doc/design/adding-priority-classes.md
exdx: If I understand this correctly, I think we should be ok in this case - per `If you delete a PriorityClass, existing Pods that use the name of the deleted PriorityClass remain unchanged, but you cannot create more Pods that use the name of the deleted PriorityClass.` so the rename would't affect existing pods. 

If the operator continues to use the name of the old PDB in their code then that's an issue, but on their end

doc/design/adding-priority-classes.md
exdx: Looks like it. `Only one PriorityClass with globalDefault set to true can exist in the system.`  per https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/#priorityclass

doc/design/adding-vertical-pod-autoscaler.md
exdx: I think so. The VPA spec is extremely straightforward - what object is being autoscaled, and the update strategy employed by the VPA for that object. I looked briefly into the update strategy as a way of mitigating risk but the reality is it's hard to picture the ways in which the VPA can affect upgrades. It may be fine, or it may blow up. 

I think we will get more clarity as we get users actually using them in the bundles

doc/design/adding-priority-classes.md
njhale: If the PDB is renamed in an update, the existing PDB sticks around until the CSV is deleted, which only happens once the new CSV is ready. I think we still would have a problem since there would be a time where you have two PDBs with the global field set true concurrently.

exdx: Yes, this is an edge-case that could get us during upgrades of operators with PDBs. I think as we think through the usecases we add e2e tests that can verify OLM behavior in this scenario.

exdx:
manually merging as this is doc-only.

PR #1566, opened 2020-06-03
Bug 1827000: Pod Config Deployment Hash Error

awgreene:
The Pod Config e2e test have been failing because the deployment is not reinstalled when the actual deployment hash does not match the calculated deployment hash. This commit updates OLM to reinstall a deployment when the hash doesn't match the calculated hash.

### DIFF
--- a/pkg/controller/install/errors.go
+++ b/pkg/controller/install/errors.go
@@ -38,11 +38,11 @@ func IsErrorUnrecoverable(err error) bool {
 	if err == nil {
 		return false
 	}
-	_, ok := unrecoverableErrors[reasonForError(err)]
+	_, ok := unrecoverableErrors[ReasonForError(err)]
 	return ok
 }
 
-func reasonForError(err error) string {
+func ReasonForError(err error) string {
 	switch t := err.(type) {
 	case StrategyError:
 		return t.Reason
--- a/pkg/controller/operators/olm/operator.go
+++ b/pkg/controller/operators/olm/operator.go
@@ -1617,7 +1617,11 @@ func (a *Operator) updateInstallStatus(csv *v1alpha1.ClusterServiceVersion, inst
 	}
 
 	if strategyErr != nil {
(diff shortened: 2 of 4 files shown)


pkg/controller/operators/olm/operator.go
```diff
@@ -1617,7 +1617,11 @@ func (a *Operator) updateInstallStatus(csv *v1alpha1.ClusterServiceVersion, inst
 	}
 
 	if strategyErr != nil {
-		csv.SetPhaseWithEventIfChanged(requeuePhase, requeueConditionReason, fmt.Sprintf("installing: %s", strategyErr), now, a.recorder)
+		if install.ReasonForError(strategyErr) == install.StrategyErrDeploymentUpdated {
+			csv.SetPhaseWithEventIfChanged(v1alpha1.CSVPhaseInstallReady, requeueConditionReason, fmt.Sprintf("installing: %s", strategyErr), now, a.recorder)
```
Bowenislandsong, line 1621: Cloud you please add a comment about the implication of changing phase to CSVinstallReady? Thank you!

pkg/controller/install/errors.go
```diff
@@ -38,11 +38,11 @@ func IsErrorUnrecoverable(err error) bool {
 	if err == nil {
 		return false
 	}
-	_, ok := unrecoverableErrors[reasonForError(err)]
+	_, ok := unrecoverableErrors[ReasonForError(err)]
 	return ok
 }
 
-func reasonForError(err error) string {
+func ReasonForError(err error) string {
```
exdx, line 45: Not related to the PR, but this function is confusing. Since we own the `StrategyError` we should be consistent in using either the error type or a pointer. Checking for both seems a little off.

pkg/controller/operators/olm/operator.go
exdx, line 1621: should we change the log message to be different - from `installing:...` in both cases to `redeploying:...` in the case where the deployment was updated?

PR #1577, opened 2020-06-08
feat: support priorityclass, poddisruptionbudget, and vpa objects

exdx:
Support objects types for aggregated api-servers:

- `PodDisruptionBudgets`
- `PriorityClasses`
- `VerticalPodAutoscaler`

These objects are supported with create/recreate features only as per the epic. 

See #1564 

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -1906,16 +1906,23 @@ func (o *Operator) apiresourceFromGVK(gvk schema.GroupVersionKind) (metav1.APIRe
 }
 
 const (
-	PrometheusRuleKind = "PrometheusRule"
-	ServiceMonitorKind = "ServiceMonitor"
+	PrometheusRuleKind        = "PrometheusRule"
+	ServiceMonitorKind        = "ServiceMonitor"
+	PodDisruptionBudgetKind   = "PodDisruptionBudget"
+	PriorityClassKind         = "PriorityClass"
+	VerticalPodAutoscalerKind = "VerticalPodAutoscaler"
 )
+
+var supportedKinds = map[string]struct{}{
+	PrometheusRuleKind:        {},
+	ServiceMonitorKind:        {},
+	PodDisruptionBudgetKind:   {},
+	PriorityClassKind:         {},
+	VerticalPodAutoscalerKind: {},
+}
 
 // isSupported returns true if OLM supports this type of CustomResource.
 func isSupported(kind string) bool {
-	switch kind {
-	case PrometheusRuleKind, ServiceMonitorKind:
-		return true
-	default:
-		return false
-	}
+	_, ok := supportedKinds[kind]
+	return ok
 }
--- a/test/e2e/bundle_e2e_test.go
+++ b/test/e2e/bundle_e2e_test.go
@@ -1,0 +1,111 @@
+package e2e
+
+import (
+	"context"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/operator-framework/api/pkg/operators/v1alpha1"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/clientset/versioned"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/operatorclient"
+	"github.com/operator-framework/operator-lifecycle-manager/test/e2e/ctx"
+)
+
+var _ = Describe("Installing bundles with new object types", func() {
+	var (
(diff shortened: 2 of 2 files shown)


pkg/controller/operators/catalog/operator.go
```diff
@@ -1906,16 +1906,23 @@ func (o *Operator) apiresourceFromGVK(gvk schema.GroupVersionKind) (metav1.APIRe
...
+	ServiceMonitorKind        = "ServiceMonitor"
+	PodDisruptionBudgetKind   = "PodDisruptionBudget"
+	PriorityClassKind         = "PriorityClass"
+	VerticalPodAutoscalerKind = "VerticalPodAutoscaler"
 )
 
+var supportedKinds = map[string]struct{}{
+	PrometheusRuleKind:        {},
+	ServiceMonitorKind:        {},
+	PodDisruptionBudgetKind:   {},
```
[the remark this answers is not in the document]
exdx: That's a good point - we could potentially add this check to the validation library. But it seems to be better off as a part of the https://github.com/operator-framework/enhancements/pull/28 story which is more comprehensive. 

Also we don't 100% know they are shipping the PDB for use with the operands - maybe they want the PDB to be used with the operator deployment itself.

exdx:
Holding this until the enhancement for managing potentially unsafe operator installations via alerts is approved. 

/hold

exdx:
We are OK going ahead with this merge as-is as we continue to refine the policy around bundle objects in the bad operator actions enhancement. 

/hold cancel

pkg/lib/operatorlister/lister.go
```diff
@@ -190,13 +190,13 @@ func newOperatorsV1Lister() *operatorsV1Lister {
...
 
 type lister struct {
-	appsV1Lister               *appsV1Lister
-	coreV1Lister               *coreV1Lister
-	rbacV1Lister               *rbacV1Lister
-	apiRegistrationV1Lister    *apiRegistrationV1Lister
-	apiExtensionsV1Lister      *apiExtensionsV1Lister
-	operatorsV1alpha1Lister    *operatorsV1alpha1Lister
-	operatorsV1Lister          *operatorsV1Lister
+	appsV1Lister            *appsV1Lister
```
kevinrizza, line 193: The edits to this file look like noise?

pkg/lib/ownerutil/util.go
```diff
@@ -398,6 +400,19 @@ func InferGroupVersionKind(obj runtime.Object) error {
...
+			Version: policyv1beta1.SchemeGroupVersion.Version,
+			Kind:    "PodDisruptionBudget",
+		})
+	case *schedulingv1.PriorityClass:
+		objectKind.SetGroupVersionKind(schema.GroupVersionKind{
+			Group:   schedulingv1.GroupName,
+			Version: schedulingv1.SchemeGroupVersion.Version,
+			Kind:    "PriorityClass",
+		})
+		// TODO add GVK for VPA objects
```
kevinrizza, line 415: Should we just include VPAs as well?

pkg/lib/operatorlister/lister.go
exdx: yes - just ran `go fmt` on it

pkg/lib/ownerutil/util.go
exdx: we could but I was wondering if it was worth importing the whole VPA library as a dependency just for this one additional typemeta check (which is somewhat optional). Just wanted some thoughts around it.

kevinrizza: If all we're ever going to do on this object at runtime is apply it indiscriminately, then I think ignoring it is fine It seems like that would be the right thing to do given our future move towards arbitrary kube manifests.

But if that is the case, do we need the type check on any of the things we're adding? IMO we should do either all or none. It feels weird to do it for some but specifically not for the one that's an external thing we happen to not vendor.

dinhxuanvu:
@exdx Both upstream and downstream e2e fail on your test case. Can you double-check to see if it passes locally for you?

ecordell:
@jianzhangbjz please review the test cases

jianzhangbjz:
@kuiwang02 Could you help have a review? Thanks!

test/e2e/bundle_e2e_test.go
```diff
@@ -0,0 +1,111 @@
...
+			)
+			// create catalog source
+			source := &v1alpha1.CatalogSource{
+				TypeMeta: metav1.TypeMeta{
+					Kind:       v1alpha1.CatalogSourceKind,
+					APIVersion: v1alpha1.CatalogSourceCRDAPIVersion,
+				},
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      sourceName,
+					Namespace: testNamespace,
```
jianzhangbjz, line 56: Sorry, what's the value of this `testNamespace` variable?

exdx: it depends on whether its run locally or on CI
* locally its `operators` 
* on openshift CI its `openshift-operators`

PR #1583, opened 2020-06-17
Bug 1847540: resolve only default channels

kevinrizza:
Non deterministic behavior was introduced where, in some cases, the
dependency resolver would select a channel entry from the registry
that was not part of the default channel and create a subscription
on that channel. This bugfix updates the registry client's filter
function to ignore non default channel entries.

### DIFF
--- a/pkg/controller/registry/registry_client.go
+++ b/pkg/controller/registry/registry_client.go
@@ -86,9 +86,9 @@ func (rc *Client) FindBundleThatProvides(ctx context.Context, group, version, ki
 	if err != nil {
 		return nil, err
 	}
-	entry := rc.filterChannelEntries(it, excludedPkgName)
+	entry := rc.filterChannelEntries(ctx, it, excludedPkgName)
 	if entry == nil {
-		return nil, fmt.Errorf("Unable to find a channel entry which doesn't belong to package %s", excludedPkgName)
+		return nil, fmt.Errorf("Unable to find a channel entry which doesn't belong to package %s's default channel", excludedPkgName)
 	}
 	bundle, err := rc.Client.Registry.GetBundle(ctx, &registryapi.GetBundleRequest{PkgName: entry.PackageName, ChannelName: entry.ChannelName, CsvName: entry.BundleName})
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/controller/registry/registry_client.go
```diff
@@ -97,25 +100,56 @@ func (rc *Client) FindBundleThatProvides(ctx context.Context, group, version, ki
...
-		return entries[0]
+		return entries[0], nil
+	}
+	return nil, nil
+}
+
+// GetDefaultPackageChannel uses registry client to get the default
+// channel name for a given package name
+func (rc *Client) getDefaultPackageChannel(ctx context.Context, pkgName string) (string, error) {
+	pkg, err := rc.Client.Registry.GetPackage(ctx, &registryapi.GetPackageRequest{Name: pkgName})
```
dinhxuanvu, line 143: We do have `GetDefaultPackage` that will return the default channel right away. So you don't need to extract it from `PackageManifest` later. I recommend using that func instead.

pkg/controller/registry/registry_client.go
```diff
@@ -97,25 +100,56 @@ func (rc *Client) FindBundleThatProvides(ctx context.Context, group, version, ki
...
-			entry := &opregistry.ChannelEntry{
-				PackageName: e.PackageName,
-				ChannelName: e.ChannelName,
-				BundleName:  e.BundleName,
-				Replaces:    e.Replaces,
+			// keep track of the default channel for each package
+			if _, ok := defChannels[e.PackageName]; !ok {
+				defChannel, err := rc.getDefaultPackageChannel(ctx, e.PackageName)
+				if err != nil {
+					return nil, err
```
dinhxuanvu, line 116: Is there a reason why we don't use a `continue` here? If there is a bad gRPC call or a bad package that causes an error here, it doesn't mean the other packages are bad. We can simply skip this iteration and continue to the next. There is a reason that I avoided returning in error before reaching EOF (as you can see that I could just pick the first item that meets the requirement and `break` but instead I add all items in the list and pick the first one on the list later). This is a gRPC connection stream, break the stream before it is done may potentially add overhead on the gRPC server to clean up the hanging/open stream connection. I would prefer if we don't do so.

pkg/controller/registry/registry_client.go
dinhxuanvu: Ignore this comment. The func for that is unfortunately not available in gRPC API even though there is a query for it in the registry. This is the only way to get the default channel info for now.

pkg/controller/registry/registry_client.go
```diff
@@ -97,25 +100,56 @@ func (rc *Client) FindBundleThatProvides(ctx context.Context, group, version, ki
...
 	for e := it.Next(); e != nil; e = it.Next() {
 		if e.PackageName != excludedPkgName {
-			entry := &opregistry.ChannelEntry{
-				PackageName: e.PackageName,
-				ChannelName: e.ChannelName,
-				BundleName:  e.BundleName,
-				Replaces:    e.Replaces,
+			// keep track of the default channel for each package
+			if _, ok := defChannels[e.PackageName]; !ok {
+				defChannel, err := rc.getDefaultPackageChannel(ctx, e.PackageName)
```
ecordell, line 114: this calls the grpc api for every channel entry, but we only need to get the default channel once up front

pkg/controller/registry/registry_client.go
```diff
@@ -97,25 +100,56 @@ func (rc *Client) FindBundleThatProvides(ctx context.Context, group, version, ki
...
+					BundleName:  e.BundleName,
+					Replaces:    e.Replaces,
+				}
+				entries = append(entries, entry)
 			}
-			entries = append(entries, entry)
 		}
 	}
 
 	if entries != nil {
```
benluddy, line 134: Nit: I will sleep better if this becomes a length-based condition, since entries can always be a non-nil but empty slice.

pkg/controller/registry/registry_client.go
```diff
@@ -97,25 +100,56 @@ func (rc *Client) FindBundleThatProvides(ctx context.Context, group, version, ki
...
+	}
+
+	var defChan string
+	if pkg != nil {
+		defChan = pkg.DefaultChannelName
+	} else {
+		return "", fmt.Errorf("package %s not found in registry", pkgName)
 	}
-	return nil
+	return defChan, nil
```
benluddy, line 154: This separate declaration and assignment looks vestigal. Can you change this to return the error if `pkg == nil`, drop the `else`, and `return pkg.DefaultChannelName, nil` to make the control flow a little clearer?
```

kevinrizza:
@benluddy @ecordell @dinhxuanvu addressed everyone's comments.

Also, at this point I think I give up. I haven't been able to reproduce any of the failures in the e2e tests, so my guess is they're just flaky. I updated the subscription e2e test to explicitly check that we're creating a sub with the default channel.

PR #1589, opened 2020-06-23
Bug 1849164: don't store full manifests in installplan status (for bundle images)

ecordell:
If a manifest comes from a bundle image, instead of storing the manifest itself in the installplan status, we now store a reference to the configmap that contains the manifest. When installing, OLM will look up the manifest and use the correct one.

One additional change was necessary - role/rolebinding name generation is now deterministic. This is because some resources for an installplan are generated by OLM, rather than read directly from the unpacked bundle, and we need identities for them so that we can track the the status of an individual step.

Steps that are references will look like this:

```yaml
    - resolving: ibm-cloud-databases-redis.v1.0.0
      resource:
        group: rbac.authorization.k8s.io
        kind: ClusterRoleBinding
        manifest: >-
         {"kind":"ConfigMap","name":"cbd102757181a13f9cee81c42815586023b4840d95543e67c1d68f1572e4667","namespace":"openshift-marketplace","catalogSourceName":"aspera","catalogSourceNamespace":"openshift-marketplace","replaces":""}
        name: ibm-cloud-databases-redis.v1.0.079b98b8bd4-i7bdf84b456
        sourceName: aspera
        sourceNamespace: openshift-marketplace
        version: v1
      status: Created
```
where the configmap references the unpacked bundle for that step.

It is possible to hit size limits in the installplan, if resolution results in lots of large manifests. Bundle images are already unpacked into configmaps on-cluster, so persisting the unpacked manifests in the installplan is not necessary.

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -1105,6 +1105,16 @@ func (o *Operator) createInstallPlan(namespace string, gen int, subs []*v1alpha1
 	return reference.GetReference(res)
 }
 
+type UnpackedBundleReference struct {
+	Kind                   string `json:"kind"`
+	Name                   string `json:"name"`
+	Namespace              string `json:"namespace"`
+	CatalogSourceName      string `json:"catalogSourceName"`
+	CatalogSourceNamespace string `json:"catalogSourceNamespace"`
+	Replaces               string `json:"replaces"`
+}
+
+// unpackBundles makes one walk through the bundlelookups and attempts to progress them
 func (o *Operator) unpackBundles(plan *v1alpha1.InstallPlan) (bool, *v1alpha1.InstallPlan, error) {
 	out := plan.DeepCopy()
 	unpacked := true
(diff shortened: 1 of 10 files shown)


pkg/controller/operators/catalog/operator.go
```diff
@@ -1418,9 +1453,79 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
...
+			if manifest == step.Resource.Manifest {
+				return "", errorwrap.Wrapf(err, "couldn't find unpacked step for %v", step)
+			}
+		}
+		return manifest, nil
+	}
+
 	for i, step := range plan.Status.Plan {
+		// TODO: should lazy load only when the manifest is actually needed
+		manifest, err := manifestForStep(step)
```
exdx, line 1522: See above -- I think linking this to step creation via the `builder` helps with the lazy loading.

exdx, line 1522: nit: WDYT about having a `type Manifest string` and using that in the code? helps improve readability, and we can hang methods off the type

vendor/github.com/operator-framework/api/pkg/operators/v1alpha1/installplan_types.go
```diff
@@ -297,7 +297,7 @@ func (b *BundleLookup) SetCondition(cond BundleLookupCondition) BundleLookupCond
 		if existing.Type != cond.Type {
 			continue
 		}
-		if existing.Status == cond.Status {
+		if existing.Status == cond.Status && existing.Reason == cond.Reason {
```
exdx, line 300: Should we open up a PR in `api` for this?

exdx:
I will lgtm since this is an urgent bugfix 
/lgtm

ecordell:
/hold

I'd like to test with the set of operators we discovered this with before merging

ecordell:
> Let's see how the tests perform but I'm a little concerned about the potential for transient read errors (failed to find configmap) impacting the actual step resolution. 

Resyncs should cover those cases, but I agree that this needs more stress testing to get numbers. 

> We are basically adding a roundtrip to the api-server for each object in the bundle.

The current implementation doesn't do any extra roundtripping - it reads the configmap data from the cache. It also only goes to the cache once per bundle (not once per object in the bundle).

I wanted to do a followup that lazy loads the data and only checks if the step needs some work done (so it would skip installed steps / bundles for a sync). The only reason I didn't do that here is because I would've needed to change the `stepper` interface a lot more than I did (the stepper would need access to the same cache/listers), and I was wary of backporting that.

exdx:
> The current implementation doesn't do any extra roundtripping - it reads the configmap data from the cache. It also only goes to the cache once per bundle (not once per object in the bundle).
> 

That makes sense - forgot we have our informer cache to work with 

> I wanted to do a followup that lazy loads the data and only checks if the step needs some work done (so it would skip installed steps / bundles for a sync). The only reason I didn't do that here is because I would've needed to change the `stepper` interface a lot more than I did (the stepper would need access to the same cache/listers), and I was wary of backporting that.

That makes sense

vendor/github.com/operator-framework/api/pkg/operators/v1alpha1/installplan_types.go
ecordell: Good call - I will do that.

pkg/controller/operators/catalog/operator.go
```diff
@@ -1418,9 +1453,79 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
 	ensurer := newStepEnsurer(kubeclient, crclient, dynamicClient)
 	b := newBuilder(kubeclient, dynamicClient, o.csvProvidedAPIsIndexer, o.logger)
 
+	refForStep := func(step *v1alpha1.Step) *UnpackedBundleReference {
```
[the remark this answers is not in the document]
ecordell: I chose not to do that originally because the builder does not exist prior to 4.5 and I knew we intend to backport this.

But this is a good point - instead of adding to the builder, I'll make a new interface / method for resolving manifests and handling the step cache. In 4.5, the builder can depend on that interface, and in < 4.4 it can be used directly in the catalog operator.

pkg/controller/operators/catalog/operator.go
ecordell: Since the type on the API is `string`, we'd have to handle manifests that are already unpacked differently from manifests that are in configmaps - instead I'll move all manifest handling into its own file, which should make it easier to keep track of what's what?

pkg/controller/operators/catalog/operator.go
```diff
@@ -1105,9 +1106,20 @@ func (o *Operator) createInstallPlan(namespace string, gen int, subs []*v1alpha1
...
+	CatalogSourceName      string `json:"catalogSourceName"`
+	CatalogSourceNamespace string `json:"catalogSourceNamespace"`
+	Replaces               string `json:"replaces"`
+}
+
+// unpackBundles makes one walk through the bundlelookups and attempts to progress them
 func (o *Operator) unpackBundles(plan *v1alpha1.InstallPlan) (bool, *v1alpha1.InstallPlan, error) {
 	out := plan.DeepCopy()
 	unpacked := true
+	unpackedCount := len(out.Status.BundleLookups)
```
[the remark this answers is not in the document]
ecordell: whoops - a planned refactor that I abandoned. will remove.

ecordell:
> IIRC, bundles are unpacked -- in their entirety -- into a single ConfigMap, which should run into the resource storage limit before the point that this fix will help.

There is a limit for how big a bundle can be right now, which is the amount that we can fit into a single configmap. This PR removes (well, extends greatly) the limit of how many bundles can be unpacked and installed via a single installplan. This limit was hit with 19 operators in a cluster.

ecordell:
@exdx @njhale Thanks for the thorough reviews - I've addressed all the feedback and I think it's ready to go

pkg/controller/registry/resolver/rbac.go
```diff
@@ -2,19 +2,38 @@ package resolver
...
+)
+
+// HashObject calculates a hash of the object for stable name generation
+func HashObject(o interface{}) string {
+	hasher := fnv.New32a()
+	hashutil.DeepHashObject(hasher, o)
+	return utilrand.SafeEncodeString(fmt.Sprint(hasher.Sum32()))
+}
+
+func generateName(base string, hash string) string {
```
benluddy, line 31: Is the contract of this function to return the concatenation of base+hash, truncating base such that the result is short enough to be an object name? I'm worried about slice range out-of-bound panics and overly-long return values. Couldn't base have length 127 in the invocation on L116?

ecordell: pushed tests for this - thanks for calling it out. see `rbac_test.go`

PR #1592, opened 2020-06-24
Remove failfast from operator e2e test

Bowenislandsong:
The operator e2e test would like to resume testing regardless of
some operators failed to install. This commit drops the failfast
cmd in the make file which is called by the ci periodic job.

**Reviewer Checklist**


Bowenislandsong:
/cc @ecordell @lilic 
I forgot I put the flag there to cause the failfast behavior.

PR #1595, opened 2020-06-24
[release-4.4] Bug 1850632: don't store full manifests in installplan status (for bundle images)

ecordell:
This is a manual cherry-pick of #1594


ecordell:
CI failures were due to differences from 4.4->4.5 in CRD handling that I didn't originally account for when backporting.

Here's a diff of the patches from this commit to the commit against master: https://gist.github.com/ecordell/c0055f0e6b9ca45ab44976c49463ccc5

kind of hard to read because it's a diff of patches, but it should make the cherry-pick changes somewhat clear:

- This version uses the k8s.io hashutil, because we hadn't manually vendored that yet in 4.4
- This version has a `CRD` handling inline in executeplan, which needs to use the "resolved" manifest instead of the inline manifest (this is the source of the CI failure, now fixed)
- This version doesn't have the "stepper" interface at all, so related changes are dropped

PR #1598, opened 2020-06-24
Disable parallel e2e-local runs by default.

benluddy:
Relying on Ginkgo's -p heuristic is confusing, so parallel runs are now opt-in via NODES.


benluddy:
@exdx @ecordell

exdx:
So this will make the e2e local tests run successfully on macOS without modifying the makefile (to remove `-p`), which is nice. 
/lgtm

benluddy:
Set NODES=2 on the workflow so that we don't sacrifice the speedup there.

ecordell:
/lgtm

🎉

PR #1599, opened 2020-06-24
Bump to kind 0.8.1.

benluddy:
This fixes an issue with log buffering in the test cluster provisioner.


benluddy:
I think there's a race to create the "kind" network here https://github.com/kubernetes-sigs/kind/blob/v0.8.1/pkg/cluster/internal/providers/docker/network.go#L44, which will be a problem for our tests in parallel mode.

/hold

PR #1603, opened 2020-06-25
Bug 1822396: Update metric when Subscription is updated

anik120:

In #1519, the subscription_sync_total metric was updated when the
related subscription object was updated. However, the PR had a bug,
in which the map used to store metric related information was not
being updated when the metric was udpated. As a result of which any
update after the first update was not removing the old metric. This
PR fixes that bug.

**Reviewer Checklist**

### DIFF
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -267,5 +267,7 @@ func UpdateSubsSyncCounterStorage(sub *olmv1alpha1.Subscription) {
 		counterValues.installedCSV = sub.Status.InstalledCSV
 		counterValues.pkg = sub.Spec.Package
 		counterValues.channel = sub.Spec.Channel
+
+		subscriptionSyncCounters[sub.GetName()] = counterValues
 	}
 }
--- a/test/e2e/metrics_e2e_test.go
+++ b/test/e2e/metrics_e2e_test.go
@@ -75,15 +75,15 @@ var _ = Describe("Metrics are generated for OLM managed resources", func() {
 				Expect(getMetricsFromPod(c, getPodWithLabel(c, "app=olm-operator"), "8081")).To(And(
 					ContainElement(LikeMetric(
 						WithFamily("csv_abnormal"),
-						WithLabel("name", failingCSV.Name),
-						WithLabel("phase", "Failed"),
-						WithLabel("reason", "UnsupportedOperatorGroup"),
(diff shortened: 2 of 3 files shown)


test/e2e/metrics_e2e_test.go
```diff
@@ -170,17 +170,53 @@ var _ = Describe("Metrics are generated for OLM managed resources", func() {
 				}).Should(And(
 					Not(ContainElement(LikeMetric(
 						WithFamily("subscription_sync_total"),
-						WithLabel("name", "metric-subscription-for-update"),
-						WithLabel("channel", stableChannel),
+						WithName("metric-subscription-for-update"),
+						WithChannel(stableChannel),
```
benluddy, line 174: Should this matcher (and the new ones added in this PR) also specify the package?

test/e2e/metrics_e2e_test.go
```diff
@@ -170,17 +170,56 @@ var _ = Describe("Metrics are generated for OLM managed resources", func() {
...
+							WithName("metric-subscription-for-update"),
+							WithChannel(betaChannel),
+							WithPackage(testPackageName),
+						))),
+						ContainElement(LikeMetric(
+							WithFamily("subscription_sync_total"),
+							WithName("metric-subscription-for-update"),
+							WithChannel(alphaChannel),
+							WithPackage(testPackageName),
+							WithPackage(testPackageName),
```
benluddy, line 219: It's harmless, but this line is duplicated.

pkg/metrics/metrics.go
```diff
@@ -267,5 +267,7 @@ func UpdateSubsSyncCounterStorage(sub *olmv1alpha1.Subscription) {
 		counterValues.installedCSV = sub.Status.InstalledCSV
 		counterValues.pkg = sub.Spec.Package
 		counterValues.channel = sub.Spec.Channel
+
+		subscriptionSyncCounters[sub.GetName()] = counterValues
```
benluddy, line 271: I don't think it needs to block this bugfix, but there should really be some kind of synchronization around access to `subscriptionSyncCounters`. Would you be interested in making a separate PR with that change?

anik120: @benluddy sure! I'm curious to see what you mean by synchronization, once we discuss that I'd be happy to open up a PR if needed.

PR #1605, opened 2020-06-25
Bug 1851095: Delete subscription metric when an operator is uninstalled

anik120:

When an operator was subscribed to using a Subscription Object,
the subscription_sync_total metric was emitted whenever the Subscription
Object was created/updated/deleted. This PR updates that behaviour to emit
the metric only when the Subscription object is created/updated, and deletes
the time series for that particular subscription when the subscription object
is deleted.

**Reviewer Checklist**


njhale:
closing due to staleness

PR #1606, opened 2020-06-25
[WIP] feat: fail installplan in case of operatorgroup dependency issues

exdx:
This PR would put the installplan in a failed state in case of operatorgroup dependency issues in the plan namespace. This provides clarity to users on why there installation failed - previously OLM would emit a log but the CSV would remain in a unknown state indefinitely.

See BZ

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -1204,7 +1204,26 @@ func (o *Operator) syncInstallPlans(obj interface{}) (syncError error) {
 	querier := o.serviceAccountQuerier.NamespaceQuerier(plan.GetNamespace())
 	ref, err := querier()
 	if err != nil {
+		// issue with operator group/service account in installplan namespace
+		// fail installplan (if not already failed)
 		syncError = fmt.Errorf("attenuated service account query failed - %v", err)
+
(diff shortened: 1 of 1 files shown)


exdx:
The thing I'm concerned with here is that by failing the installplan (if its not in a failed state) it contradicts another codepath where OLM can resync the installplan in case the serviceaccount is updated (to the correct permissions) and that allows the installplan to succeed. Not sure if that is the right context, but conceptually that's my concern.

Right now once an installplan is failed thats a terminal condition - I want to make sure that this change doesn't result in a installplan bouncing around between failed and another status.

exdx:
Per @ecordell 
```
we just need to verify that if an installplan goes into the failed state, and the reason then it failed has 
changed (via serviceaccount or operatorgroup changes), it will get retried (if it’s the current active installplan).
add some e2e testcases for coming back out of the failed state and then we’ll know for sure that works
```

exdx:
No longer associated with a BZ but made into an RFE instead. 
https://issues.redhat.com/browse/RFE-1073

exdx:
Closing as RFE priority was lowered -- this will be addressed in a potentially more robust way in a later release.

PR #1607, opened 2020-06-25
feat(operator): automatically adopt generated resources

njhale:
Automatically adopt resources generated by OLM during operator
installation. Any CSV installed via Subscription is adopted by an
operator named after the subscribed package and its install namespace; other resources with
OwnerReferences or OLM-style owner labels referencing component CSVs are
transitively adopted.


pkg/controller/operators/adoption_controller.go
```diff
@@ -0,0 +1,408 @@
...
+		Watches(&source.Kind{Type: &corev1.Namespace{}}, enqueueCSV).
+		Watches(&source.Kind{Type: &corev1.Service{}}, enqueueCSV).
+		Watches(&source.Kind{Type: &corev1.ServiceAccount{}}, enqueueCSV).
+		Watches(&source.Kind{Type: &corev1.Secret{}}, enqueueCSV).
+		Watches(&source.Kind{Type: &corev1.ConfigMap{}}, enqueueCSV).
+		Watches(&source.Kind{Type: &rbacv1.Role{}}, enqueueCSV).
+		Watches(&source.Kind{Type: &rbacv1.RoleBinding{}}, enqueueCSV).
+		Watches(&source.Kind{Type: &rbacv1.ClusterRole{}}, enqueueCSV).
+		Watches(&source.Kind{Type: &rbacv1.ClusterRoleBinding{}}, enqueueCSV).
+		Watches(&source.Kind{Type: &apiextensionsv1.CustomResourceDefinition{}}, enqueueCSV).
```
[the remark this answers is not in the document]
njhale: We'll get the same resources twice then.

pkg/controller/operators/adoption_controller.go
```diff
@@ -0,0 +1,408 @@
...
+		if adopted {
+			// Only update the CSV if freshly adopted
+			if err := r.Update(ctx, csv); err != nil {
+				log.Error(err, "Error adopting installed CSV")
+				return reconcile.Result{}, err
+			}
+		}
+	}
+
+	if name := in.Status.CurrentCSV; name != "" && name != in.Status.CurrentCSV {
```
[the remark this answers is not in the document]
njhale: Sorry, this should be `if name := in.Status.CurrentCSV; name != in.Status.InstalledCSV` -- I'll update it.

pkg/controller/operators/adoption_controller.go
```diff
@@ -0,0 +1,408 @@
...
+		if adopted {
+			// Only update the CSV if it's been freshly adopted
+			if err := r.Update(ctx, csv); err != nil {
+				log.Error(err, "Error adopting installed CSV")
+				return reconcile.Result{}, err
+			}
+		}
+	}
+
+	return ctrl.Result{}, nil
```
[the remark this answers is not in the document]
njhale: That's on purpose. `reconcile.Result{}` is an alternative requeue signal. I requeue with non-nil errors above, and have no need to use this elsewhere.

pkg/controller/operators/adoption_controller.go
```diff
@@ -0,0 +1,408 @@
...
+		&corev1.ServiceList{},
+		&corev1.NamespaceList{},
+		&corev1.ServiceAccountList{},
+		&corev1.SecretList{},
+		&corev1.ConfigMapList{},
+		&rbacv1.RoleList{},
+		&rbacv1.RoleBindingList{},
+		&rbacv1.ClusterRoleList{},
+		&rbacv1.ClusterRoleBindingList{},
+		&apiextensionsv1.CustomResourceDefinitionList{},
```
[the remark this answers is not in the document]
njhale: see https://github.com/operator-framework/operator-lifecycle-manager/pull/1607/files/796734003ed7ee993c1ca38d6971e7e46249624c#r445814464

pkg/controller/operators/suite_test.go
```diff
@@ -47,15 +53,30 @@ var (
 		GracePeriodSeconds: &gracePeriod,
 		PropagationPolicy:  &propagation,
 	}
-	genName = names.SimpleNameGenerator.GenerateName
+	genName  = names.SimpleNameGenerator.GenerateName
+	fixtures = testobj.NewFixtureFiller(
```
ecordell, line 57: 👍 

Aside - I've been meaning to write a library (or generator) that would let you perturb a fixture for certain tests.

```go
inputRoleBinding := fixtureRB.Set("metadata.name", "newName").Set("other.field", "value")
```

pkg/controller/operators/adoption_controller.go
```diff
@@ -0,0 +1,407 @@
...
+	}
+
+	// OLM generated Operators are named after their packages
+	if in.Spec == nil || in.Spec.Package == "" {
+		log.Info("subscription missing status, ignoring")
+		return reconcile.Result{}, nil
+	}
+
+	// Wrap with convenience decorator
+	operator, err := r.factory.NewPackageOperator(in.Spec.Package)
```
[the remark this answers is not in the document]
njhale: Good question. My thought when writing the enhancement is that instances of the same operator in different namespaces would be aggregated under a single `Operator` resource, but that may have been short-sighted since an `Operator` resource will need to eventually install _an_ operator.

pkg/controller/operators/adoption_controller.go
```diff
@@ -0,0 +1,407 @@
...
+		&corev1.ConfigMapList{},
+		&rbacv1.RoleList{},
+		&rbacv1.RoleBindingList{},
+		&rbacv1.ClusterRoleList{},
+		&rbacv1.ClusterRoleBindingList{},
+		&apiextensionsv1.CustomResourceDefinitionList{},
+		&apiregistrationv1.APIServiceList{},
+		&operatorsv1alpha1.SubscriptionList{},
+		&operatorsv1alpha1.InstallPlanList{},
+		&operatorsv1alpha1.ClusterServiceVersionList{},
```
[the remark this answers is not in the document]
exdx: Yes -- feels like we should move the map to a general place like `api`

pkg/controller/operators/adoption_controller.go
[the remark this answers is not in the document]
njhale: Added a namespace suffix.

pkg/controller/operators/adoption_controller.go
```diff
@@ -0,0 +1,409 @@
...
+		Watches(&source.Kind{Type: &corev1.Namespace{}}, enqueueCSV).
+		Watches(&source.Kind{Type: &corev1.Service{}}, enqueueCSV).
+		Watches(&source.Kind{Type: &corev1.ServiceAccount{}}, enqueueCSV).
+		Watches(&source.Kind{Type: &corev1.Secret{}}, enqueueCSV).
+		Watches(&source.Kind{Type: &corev1.ConfigMap{}}, enqueueCSV).
+		Watches(&source.Kind{Type: &rbacv1.Role{}}, enqueueCSV).
+		Watches(&source.Kind{Type: &rbacv1.RoleBinding{}}, enqueueCSV).
+		Watches(&source.Kind{Type: &rbacv1.ClusterRole{}}, enqueueCSV).
+		Watches(&source.Kind{Type: &rbacv1.ClusterRoleBinding{}}, enqueueCSV).
+		Watches(&source.Kind{Type: &apiextensionsv1.CustomResourceDefinition{}}, enqueueCSV).
```
exdx, line 92: should we put watch on `apiextensionsv1beta1.CustomResourceDefinition` types as well?

pkg/controller/operators/adoption_controller.go
```diff
@@ -0,0 +1,409 @@
...
+	// Wrap with convenience decorator
+	operator, err := r.factory.NewPackageOperator(in.Spec.Package, in.GetNamespace())
+	if err != nil {
+		log.Error(err, "Could not wrap Operator with convenience decorator")
+		return reconcile.Result{}, nil
+	}
+
+	// Find the Subscription's CSVs and apply the component label if necessary
+	csv := &operatorsv1alpha1.ClusterServiceVersion{}
+	if name := in.Status.InstalledCSV; name != "" {
```
exdx, line 159: Just double-checking, does `InstalledCSV` get populated if the Subscription fails to install the CSV? I think we want to surface the Operator object regardless of whether the Subscription succeeds or not right? Just want to make sure we aren't limiting this to only successful CSV installations -- if the install fails the Operator still surfaces useful information to the end user.

pkg/controller/operators/adoption_controller.go
```diff
@@ -0,0 +1,409 @@
...
+		Watches(&source.Kind{Type: &corev1.Service{}}, enqueueCSV).
+		Watches(&source.Kind{Type: &corev1.ServiceAccount{}}, enqueueCSV).
+		Watches(&source.Kind{Type: &corev1.Secret{}}, enqueueCSV).
+		Watches(&source.Kind{Type: &corev1.ConfigMap{}}, enqueueCSV).
+		Watches(&source.Kind{Type: &rbacv1.Role{}}, enqueueCSV).
+		Watches(&source.Kind{Type: &rbacv1.RoleBinding{}}, enqueueCSV).
+		Watches(&source.Kind{Type: &rbacv1.ClusterRole{}}, enqueueCSV).
+		Watches(&source.Kind{Type: &rbacv1.ClusterRoleBinding{}}, enqueueCSV).
+		Watches(&source.Kind{Type: &apiextensionsv1.CustomResourceDefinition{}}, enqueueCSV).
+		Watches(&source.Kind{Type: &apiregistrationv1.APIService{}}, enqueueCSV).
```
exdx, line 93: As @ecordell noted in https://github.com/operator-framework/operator-lifecycle-manager/pull/1607/files#r446177548 there are some additional types that we are adding support for. Since the `watched` objects are the same as the `adoptees` in theory, is there a way to share a common data structure between the two parts of the code?

pkg/controller/operators/decorators/operator.go
```diff
@@ -122,24 +182,51 @@ func (a *Operator) ComponentSelector() (labels.Selector, error) {
...
+	}
+
+	labels := m.GetLabels()
+	if labels == nil {
+		labels = map[string]string{}
+		m.SetLabels(labels)
+	}
+
+	if _, ok := labels[labelKey]; !ok {
+		labels[labelKey] = ""
```
exdx, line 218: nit: this is just a preference, but if we are adding a `""` to signify that a label is present, is it more intuitive to use an annotation instead?

pkg/controller/operators/adoption_controller.go
njhale: It's my intention to have this dynamically watch all resources in the near future, so for this use case it didn't seem as pressing. If we still want to have a "supported" struct to restrict these resources, we could add it in a follow up.

pkg/controller/operators/adoption_controller.go
njhale: you'd end up with duplicate resources since a list returns _all_ CRDs converted to the request version.

pkg/controller/operators/decorators/operator.go
njhale: annotations are also key value pairs, is there another reason why an annotation is more intuitive?

pkg/controller/install/certresources.go
```diff
@@ -215,21 +215,21 @@ func (i *StrategyDeploymentInstaller) installCertRequirementsForDeployment(deplo
 	}
 	service.SetName(ServiceName(deploymentName))
 	service.SetNamespace(i.owner.GetNamespace())
-	ownerutil.AddNonBlockingOwner(service, i.owner)
```
ecordell, line 218: Should we just make AddNonBlockingOwner idempotent so it's safe to call multiple times with the same object/owner pair?

pkg/controller/operators/adoption_controller.go
```diff
@@ -0,0 +1,438 @@
+/*
+Copyright 2020 The Kubernetes authors.
```
ecordell, line 2: 😲  this is not the right copyright - we also removed the year from all of our generated files (and don't typically have the notice on non-generated files).

We discussed using the short names everywhere but haven't done it yet.

pkg/controller/operators/adoption_controller.go
```diff
@@ -0,0 +1,438 @@
...
+		} else {
+			log.Error(err, "Error finding Subscription")
+		}
+
+		return reconcile.Result{}, nil
+	}
+
+	// OLM generated Operators are named after their packages and further qualified by the install namespace
+	if in.Spec == nil || in.Spec.Package == "" {
+		log.Info("subscription missing status, ignoring")
```
ecordell, line 149: nit: log message

pkg/controller/operators/adoption_controller.go
```diff
@@ -0,0 +1,438 @@
...
+	}
+
+	if name := in.Status.InstalledCSV; name != "" {
+		if err := adoptCSV(name); err != nil {
+			log.Error(err, "Error adopting installed CSV")
+			errs = append(errs, err)
+		}
+	}
+
+	if name := in.Status.CurrentCSV; name != "" && name != in.Status.InstalledCSV {
```
ecordell, line 209: do we need to adopt the current csv at all? it may not exist.

pkg/controller/operators/adoption_controller.go
```diff
@@ -0,0 +1,438 @@
...
+			// Skip on transient error or missing CRD
+			continue
+		}
+
+		if crd == nil || !selector.Matches(labels.Set(crd.GetLabels())) {
+			// Skip empty and labelled CRDs
+			continue
+		}
+
+		crd.SetManagedFields(nil) // Clear managedFields to support patch
```
ecordell, line 366: this is needed for a non-apply patch?

test/e2e/operator_test.go
```diff
@@ -108,71 +126,199 @@ var _ = Describe("Operator", func() {
...
+			Eventually(func() error {
+				return client.Create(clientCtx, ns)
+			}).Should(Succeed())
+
+			// Default to AllNamespaces
+			og := &operatorsv1.OperatorGroup{}
+			og.SetNamespace(ns.GetName())
+			og.SetName(genName("og-"))
+			Eventually(func() error {
+				return client.Create(clientCtx, og)
```
ecordell, line 232: this line is failing in e2e tests

```
        <*errors.StatusError | 0xc000b07e00>: {
            ErrStatus: {
                TypeMeta: {Kind: "Status", APIVersion: "v1"},
                ListMeta: {
                    SelfLink: "",
                    ResourceVersion: "",
                    Continue: "",
                    RemainingItemCount: nil,
                },
                Status: "Failure",
                Message: "OperatorGroup in version \"v1\" cannot be handled as a OperatorGroup: unmarshalerDecoder: Object 'Kind' is missing in '{\"metadata\":{\"name\":\"og-mjsp5\",\"namespace\":\"ns-cpnst\",\"creationTimestamp\":null},\"spec\":{},\"status\":{\"lastUpdated\":null}}', error found in #10 byte of ...|ed\":null}}\n|..., bigger context ...|mp\":null},\"spec\":{},\"status\":{\"lastUpdated\":null}}\n|...",
                Reason: "BadRequest",
                Details: nil,
                Code: 400,
            },
        }
        OperatorGroup in version "v1" cannot be handled as a OperatorGroup: unmarshalerDecoder: Object 'Kind' is missing in '{"metadata":{"name":"og-mjsp5","namespace":"ns-cpnst","creationTimestamp":null},"spec":{},"status":{"lastUpdated":null}}', error found in #10 byte of ...|ed":null}}
        |..., bigger context ...|mp":null},"spec":{},"status":{"lastUpdated":null}}
        |...
```

I'm guessing the client doesn't have the v1 api in it's scheme - or you can just add typemeta here in the test

pkg/controller/install/certresources.go
njhale: IIRC it is idempotent, the problem here was the fact that we added, then did an append below

pkg/controller/operators/adoption_controller.go
njhale: will remove -- this was me using kubebuilder defaults, sorry

pkg/controller/operators/adoption_controller.go
njhale: you mean to drop this?

pkg/controller/operators/adoption_controller.go
njhale: we can probably drop this. I had thought there was an edge case where we can have both, but maybe that's not possible.

pkg/controller/operators/adoption_controller.go
njhale: I don't think so, I'll drop

ecordell:
@jianzhangbjz Please take a look at the test cases

pkg/controller/operators/adoption_controller.go
ecordell: it says "subscription missing status" but the condition is "spec missing package"

jianzhangbjz:
@kuiwang02 Please have a review, thanks!

pkg/controller/operators/suite_test.go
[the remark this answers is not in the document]
njhale: namespaces are pretty simple to construct, so I omitted one from the filler.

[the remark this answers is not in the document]
kuiwang02: ok, got it. thanks

njhale:
/hold

Preventing turn while test failures are being investigated.

njhale:
/hold cancel

Failures look unrelated. Opening separate PRs to address. In the short term, I'm unholding this.

test/e2e/operator_test.go
njhale: Turns out the helm provisioner was setting some fields on the rest config that were causing the controller-runtime client to use a different -- failing -- codec

njhale:
/hold cancel

Some patches merged that address the blocking tests.

PR #1610, opened 2020-06-26
Add more precise matching to the metric end-to-end tests.

benluddy:
The metric tests scrape metrics in the Prometheus text format and have
been mainly asserting based on the presence or absence of substrings
within the entire response body.

Metrics are now parsed into a slice of structs, and there is a new
Gomega matcher that can reference metric family, labels, and
value (for untyped metrics, gauges, and counters).

### DIFF
--- a/test/e2e/metrics_e2e_test.go
+++ b/test/e2e/metrics_e2e_test.go
@@ -3,13 +3,15 @@
 package e2e
 
 import (
+	"bytes"
 	"context"
-	"fmt"
 	"regexp"
-	"time"
+	"strings"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	io_prometheus_client "github.com/prometheus/client_model/go"
+	"github.com/prometheus/common/expfmt"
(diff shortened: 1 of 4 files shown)


Bowenislandsong:
nice!
lgtm (wait for a second review)

test/e2e/metrics_e2e_test.go
```diff
@@ -3,21 +3,22 @@
...
+	"bytes"
 	"context"
-	"fmt"
 	"regexp"
-	"time"
+	"strings"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	io_prometheus_client "github.com/prometheus/client_model/go"
```
njhale, line 13: nit: Is there a reason for the snake-case? AFAIK it's not conventional for import aliases.

benluddy: This package is actually generated by protobuf, so `io_prometheus_client` is its real name. I can sneak in a human-friendly import alias in a separate PR.

PR #1611, opened 2020-06-26
fix: update README to point to the OLM book website 

exdx:

A new user coming to the OLM repo and reading the README would be interested in the OLM book. The link we provide should point to the OLM book itself, not the git repo backing the content. This makes for a better user experience for someone looking into OLM for the first time. 

**Reviewer Checklist**

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 
 ## Overview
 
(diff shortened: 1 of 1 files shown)


Bowenislandsong:
should we put https://olm.operatorframework.io/ on there?

exdx:
will manually merge as this is doc-only (only updates a link in the README).

PR #1617, opened 2020-06-30
test(e2e): patch installplans w/ server-side apply

njhale:
Use server-side apply to patch InstallPlans under test in order to prevent
update conflict flakes.

### DIFF
--- a/test/e2e/subscription_e2e_test.go
+++ b/test/e2e/subscription_e2e_test.go
@@ -22,6 +22,7 @@ import (
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/discovery"
 
@@ -47,7 +48,7 @@ var _ = Describe("Subscription", func() {
 		c := newKubeClient()
 		crc := newCRClient()
 		defer func() {
-			require.NoError(GinkgoT(), crc.OperatorsV1alpha1().Subscriptions(testNamespace).DeleteCollection(context.TODO(), metav1.DeleteOptions{}, metav1.ListOptions{}))
+			require.NoError(GinkgoT(), crc.OperatorsV1alpha1().Subscriptions(testNamespace).DeleteCollection(context.Background(), metav1.DeleteOptions{}, metav1.ListOptions{}))
 		}()
 		require.NoError(GinkgoT(), initCatalog(GinkgoT(), c, crc))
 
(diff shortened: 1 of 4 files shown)


njhale:
/hold

Looks like the patch is still running into issues with `resourceVersion`.

test/e2e/subscription_e2e_test.go
```diff
@@ -948,106 +944,106 @@ var _ = Describe("Subscription", func() {
...
-		require.NoError(GinkgoT(), err)
+		plan, err := crc.OperatorsV1alpha1().InstallPlans(ref.Namespace).Get(context.Background(), ref.Name, metav1.GetOptions{})
+		Expect(err).ToNot(HaveOccurred())
 
 		// Set the InstallPlan's approval mode to Manual
 		plan.Spec.Approval = v1alpha1.ApprovalManual
 		plan.Spec.Approved = false
-		plan, err = crc.OperatorsV1alpha1().InstallPlans(ref.Namespace).Update(context.TODO(), plan, metav1.UpdateOptions{})
-		require.NoError(GinkgoT(), err)
+		setTypeMeta := func(p *v1alpha1.InstallPlan) *v1alpha1.InstallPlan {
```
ecordell, line 962: if this is in an eventually, it should work?

exdx:
This is really interesting. Does it make any sense to consider turning this logic into a library such that more of the e2e tests can start to use server-side apply?

njhale:
> This is really interesting. Does it make any sense to consider turning this logic into a library such that more of the e2e tests can start to use server-side apply?

@exdx The function is included with our e2e tests here, so it's already available to use in any of them.

test/e2e/util_test.go
```diff
@@ -882,3 +884,104 @@ func toggleFeatureGates(deployment *appsv1.Deployment, toToggle ...featuregate.F
...
+		panic(fmt.Sprintf("unable to extract key from resource: %s", err))
+	}
+
+	return func() error {
+		cp := obj.DeepCopyObject().(Object)
+		if err := client.Get(bg, key, cp); err != nil {
+			return err
+		}
+		cp.GetObjectKind().SetGroupVersionKind(obj.GetObjectKind().GroupVersionKind())
+		cp.SetManagedFields(nil)
```
ecordell, line 946: is this needed? shouldn't the `ForceOwnership` make `test` become the owner of any changed fields?

njhale: this came out of some red/green/refactor-ing -- an error is returned without it.

test/e2e/util_test.go
```diff
@@ -882,3 +884,110 @@ func toggleFeatureGates(deployment *appsv1.Deployment, toToggle ...featuregate.F
...
+		out = change.Call([]reflect.Value{reflect.ValueOf(cp)})
+		if err := out[0].Interface(); err != nil {
+			return err.(error)
+		}
+
+		if err := client.Status().Patch(bg, cp, controllerclient.Apply, controllerclient.ForceOwnership, controllerclient.FieldOwner("test")); err != nil {
+			return err
+		}
+
+		reflect.ValueOf(obj).Elem().Set(cpv.Elem())
```
benluddy, line 989: Does the doc comment make it clear that the value addressed by`obj` will be updated?

PR #1619, opened 2020-06-30
Introduce preference order for solver.

benluddy:
Installables that appear earlier in a dependency constraint have
higher preference than those appearing later in the same constraint,
and the dependencies of Installables that appear earlier in the input
have higher preference than the dependencies of those appearing later
in the input.


pkg/controller/registry/resolver/solver/dict.go
```diff
@@ -191,36 +187,69 @@ func compileDict(installables []Installable) *dict {
...
-		if w >= 0 {
-			// Omitting the cardinality constraint on the
-			// last iteration prevents the cardinality
-			// constraint literal from showing up in the
-			// set of failed assumptions should there be
-			// no solution.
-			g.Assume(d.cards.Leq(w))
+func (d *dict) MinimizeCardinality(g *gini.Gini) error {
+	for w := 0; w <= d.cards.N(); w++ {
+		g.Assume(d.cards.Leq(w))
```
ecordell, line 208: If we're spending a lot of time here, can we make an argument for a minimum value that's greater than 0?

The cardinality of the solution should be at least the length of all mandatory identifiers, right?

pkg/controller/registry/resolver/solver/solve.go
```diff
@@ -0,0 +1,249 @@
...
+type searcher struct {
+	solver  *solver
+	g       *gini.Gini
+	assumed *orderedLitSet
+	buffer  []z.Lit
+}
+
+// ... Must always return either satisfiable or unsatisfiable!
+func (h *searcher) search(ctx context.Context, upto int) int {
+	if upto < len(h.assumed.Slice()) {
```
ecordell, line 148: nit: check if `>` and return early?

pkg/controller/registry/resolver/solver/tracer.go
```diff
@@ -0,0 +1,33 @@
...
+}
+
+type DefaultTracer struct{}
+
+func (DefaultTracer) Trace(_ SearchPosition) {
+}
+
+type PrintfTracer struct{}
+
+func (PrintfTracer) Trace(p SearchPosition) {
```
ecordell, line 23: `Tracer` could just embed `io.Writer` and then it would be easy to write tests for the Tracer / validate its output

non blocking, but it would be nice to see the trace output for unsatisfiable tests

pkg/controller/registry/resolver/solver/dict.go
benluddy: Most of the cycles were spent building the sorting network in `compileDict`. I'm pushing up a change that postpones building the sorting network until after a solution is found, and had really good results.

```
// before
BenchmarkSolve-8   	     216	  54731543 ns/op
// after
BenchmarkSolve-8   	    9351	   1234811 ns/op
```

I updated BenchmarkSolve as well so that it covers both `compileDict` and `Solve` together. I think that's more realistic.

pkg/controller/registry/resolver/solver/solve.go
benluddy: I went back and forth on it, because I felt that it was clearest to end with `Solve()`. PTAL, I tried something different in my next push.

pkg/controller/registry/resolver/solver/tracer.go
benluddy: > non blocking, but it would be nice to see the trace output for unsatisfiable tests

Done, good idea.

pkg/controller/registry/resolver/solver/dict.go
```diff
@@ -191,36 +184,77 @@ func compileDict(installables []Installable) *dict {
...
+// set of lits to fix, attempts to minimize the cardinality of the
+// solution without removing any of the fixed lits.
+func (d *dict) MinimizeCardinality(g inter.S, assumed litSet) error {
+	var extras []z.Lit
+	for m := range d.installables {
+		if g.Value(m) && !assumed.Contains(m) {
+			extras = append(extras, m)
+		}
+	}
+	cards := d.c.CardSort(extras)
```
benluddy, line 211: I think there needs to be a call to https://pkg.go.dev/github.com/irifrance/gini/logic?tab=doc#C.CnfSince to add any new clauses. Looking into it...

ecordell:
/lgtm

This is not yet user-facing, so not pinging for QE review

PR #1623, opened 2020-07-07
Bug 1851213: Set Resource to plural in CSV RelatedObjects field

ankitathomas:
The `Resource` property in `RelatedObjects` field of `ClusterServiceVersionStatus` is meant to be the lowercase plural value, but were being populated with the `Kind` instead. Made modifications in this PR to switch all `Kind` values in `Resource` to the expected format.

**Reviewer Checklist**

### DIFF
--- a/pkg/lib/operatorstatus/builder_test.go
+++ b/pkg/lib/operatorstatus/builder_test.go
@@ -336,7 +336,7 @@ func TestBuilder(t *testing.T) {
 		{
 			name: "WithRelatedObject/ReferenceNotPresentInStatus",
 			action: func(b *Builder) {
-				b.WithRelatedObject("group", "resource", "namespace", "name")
+				b.WithRelatedObject("group", "resources", "namespace", "name")
 			},
 			expected: &configv1.ClusterOperatorStatus{
 				Conditions: []configv1.ClusterOperatorStatusCondition{},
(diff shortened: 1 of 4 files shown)


pkg/lib/operatorstatus/builder_test.go
```diff
@@ -399,7 +399,7 @@ func TestBuilder(t *testing.T) {
 					configv1.ObjectReference{
 						Group:     "group",
 						Resource:  "resource",
```
ecordell, line 401: shouldn't this be `Resource:  "resources",`, not `Namespace: "namespaces",`?

pkg/lib/operatorstatus/csv_reporter.go
```diff
@@ -76,7 +76,7 @@ func (r *csvStatusReporter) GetNewStatus(existing *configv1.ClusterOperatorStatu
 		gvk := csv.GetObjectKind().GroupVersionKind()
 
 		builder.WithoutVersion(csv.GetName(), csv.Spec.Version.String()).
-			WithoutRelatedObject(gvk.Group, gvk.Kind, csv.GetNamespace(), csv.GetName())
+			WithoutRelatedObject(gvk.Group, "clusterserviceversions", csv.GetNamespace(), csv.GetName())
```
ecordell, line 79: It would be nice if we could thread the resource through here the same way we do the GVK from context.

For now, just making this string a constant (since the CSV-reported is pretty specific to CSVs anyway) would help.

ankitathomas: Added in a constant for this

PR #1624, opened 2020-07-08
Bug 1853601: Retry conflicting catalog source pod update after fetching latest pod spec

ankitathomas:
Updating the catalog source pod was intermittently failing due to resource version conflicts. This change retries the update after fetching the latest spec of the catalog source pod.

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"k8s.io/apimachinery/pkg/runtime"
 	"reflect"
 	"sort"
 	"strings"
@@ -49,6 +50,7 @@ import (
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/registry/grpc"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/registry/reconciler"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/registry/resolver"
+	controllerclient "github.com/operator-framework/operator-lifecycle-manager/pkg/lib/controller-runtime/client"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/event"
 	index "github.com/operator-framework/operator-lifecycle-manager/pkg/lib/index"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/operatorclient"
@@ -74,6 +76,7 @@ const (
 	generatedByKey         = "olm.generated-by"
 	maxInstallPlanCount    = 5
 	maxDeletesPerSweep     = 5
+	RegistryFieldManager   = "olm.registry"
 )
 
 // Operator represents a Kubernetes operator that executes InstallPlans by
@@ -107,7 +110,7 @@ type Operator struct {
 type CatalogSourceSyncFunc func(logger *logrus.Entry, in *v1alpha1.CatalogSource) (out *v1alpha1.CatalogSource, continueSync bool, syncError error)
 
 // NewOperator creates a new Catalog Operator.
-func NewOperator(ctx context.Context, kubeconfigPath string, clock utilclock.Clock, logger *logrus.Logger, resync time.Duration, configmapRegistryImage, utilImage string, operatorNamespace string) (*Operator, error) {
+func NewOperator(ctx context.Context, kubeconfigPath string, clock utilclock.Clock, logger *logrus.Logger, resync time.Duration, configmapRegistryImage, utilImage string, operatorNamespace string, scheme *runtime.Scheme) (*Operator, error) {
 	resyncPeriod := queueinformer.ResyncWithJitter(resync, 0.2)
 	config, err := clientcmd.BuildConfigFromFlags("", kubeconfigPath)
 	if err != nil {
@@ -142,6 +145,11 @@ func NewOperator(ctx context.Context, kubeconfigPath string, clock utilclock.Clo
 		return nil, err
 	}
 
+	ssaClient, err := controllerclient.NewForConfig(config, scheme, RegistryFieldManager)
+	if err != nil {
+		return nil, err
+	}
+
 	// Allocate the new instance of an Operator.
 	op := &Operator{
 		Operator:                 queueOperator,
@@ -162,9 +170,8 @@ func NewOperator(ctx context.Context, kubeconfigPath string, clock utilclock.Clo
 		clientAttenuator:         scoped.NewClientAttenuator(logger, config, opClient, crClient, dynamicClient),
 	}
 	op.sources = grpc.NewSourceStore(logger, 10*time.Second, 10*time.Minute, op.syncSourceState)
-	op.reconciler = reconciler.NewRegistryReconcilerFactory(lister, opClient, configmapRegistryImage, op.now)
+	op.reconciler = reconciler.NewRegistryReconcilerFactory(lister, opClient, configmapRegistryImage, op.now, ssaClient)
 	res := resolver.NewOperatorStepResolver(lister, crClient, opClient.KubernetesInterface(), operatorNamespace, op.sources, logger)
-	op.resolver = resolver.NewOperatorStepResolver(lister, crClient, opClient.KubernetesInterface(), operatorNamespace, op.sources, logger)
 	op.resolver = resolver.NewInstrumentedResolver(res, metrics.RegisterDependencyResolutionSuccess, metrics.RegisterDependencyResolutionFailure)
 
 	// Wire OLM CR sharedIndexInformers
--- a/pkg/controller/registry/reconciler/grpc.go
+++ b/pkg/controller/registry/reconciler/grpc.go
@@ -3,8 +3,8 @@ package reconciler
 import (
 	"context"
 	"fmt"
-
 	"github.com/operator-framework/api/pkg/operators/v1alpha1"
+	controllerclient "github.com/operator-framework/operator-lifecycle-manager/pkg/lib/controller-runtime/client"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/operatorclient"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/operatorlister"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/ownerutil"
@@ -70,9 +70,10 @@ func (s *grpcCatalogSourceDecorator) Pod() *v1.Pod {
 }
 
 type GrpcRegistryReconciler struct {
-	now      nowFunc
-	Lister   operatorlister.OperatorLister
-	OpClient operatorclient.ClientInterface
+	now       nowFunc
+	Lister    operatorlister.OperatorLister
+	OpClient  operatorclient.ClientInterface
+	SSAClient *controllerclient.ServerSideApplier
 }
 
 var _ RegistryReconciler = &GrpcRegistryReconciler{}
@@ -194,16 +195,17 @@ func (c *GrpcRegistryReconciler) ensureUpdatePod(source grpcCatalogSourceDecorat
 			logrus.WithField("CatalogSource", source.GetName()).Info("detect image update for catalogsource pod")
 
 			updateFlag = true
-			updatePod.Labels[CatalogSourceLabelKey] = source.GetName()
-			updatePod.Labels[CatalogSourceUpdateKey] = ""
 
 			// Update the update pod to promote it to serving pod
-			_, err := c.OpClient.KubernetesInterface().CoreV1().Pods(source.GetNamespace()).Update(context.TODO(), updatePod, metav1.UpdateOptions{})
-			if err != nil {
-				return errors.Wrapf(err, "error creating new pod: %s", source.Pod().GetName())
-			}
+			err := c.SSAClient.Apply(context.TODO(), updatePod, func(p *v1.Pod) error {
+				p.Labels[CatalogSourceLabelKey] = source.GetName()
+				p.Labels[CatalogSourceUpdateKey] = ""
+				return nil
+			})()
 
-			break
+			if err != nil {
+				return errors.Wrapf(err, "error updating catalog source pod: %s", source.Pod().GetName())
+			}
 		}
 	}
 
--- a/pkg/lib/controller-runtime/client/fake_ssa.go
+++ b/pkg/lib/controller-runtime/client/fake_ssa.go
@@ -1,0 +1,56 @@
+package client
+
+import (
+	"context"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	k8scontrollerclient "sigs.k8s.io/controller-runtime/pkg/client"
+	fakecontrollerclient "sigs.k8s.io/controller-runtime/pkg/client/fake"
+)
+
+// FakeApplier provides a wrapper around the fake k8s controller client to convert the unsupported apply-type patches to merge patches.
+func NewFakeApplier(scheme *runtime.Scheme, owner string, objs ...runtime.Object) *ServerSideApplier {
+	return &ServerSideApplier{
+		client: &fakeApplier{fakecontrollerclient.NewFakeClientWithScheme(scheme, objs...)},
+		Scheme: scheme,
+		Owner:  k8scontrollerclient.FieldOwner(owner),
+	}
+}
+
+type fakeApplier struct {
+	k8scontrollerclient.Client
+}
+
+func (c *fakeApplier) Patch(ctx context.Context, obj runtime.Object, patch k8scontrollerclient.Patch, opts ...k8scontrollerclient.PatchOption) error {
+	patch, opts = convertApplyToMergePatch(patch, opts...)
+	return c.Client.Patch(ctx, obj, patch, opts...)
+}
+
+func (c *fakeApplier) Status() k8scontrollerclient.StatusWriter {
+	return fakeStatusWriter{c.Client.Status()}
+}
+
+type fakeStatusWriter struct {
+	k8scontrollerclient.StatusWriter
+}
+
+func (c fakeStatusWriter) Patch(ctx context.Context, obj runtime.Object, patch k8scontrollerclient.Patch, opts ...k8scontrollerclient.PatchOption) error {
+	patch, opts = convertApplyToMergePatch(patch, opts...)
+	return c.StatusWriter.Patch(ctx, obj, patch, opts...)
+}
+
+func convertApplyToMergePatch(patch k8scontrollerclient.Patch, opts ...k8scontrollerclient.PatchOption) (k8scontrollerclient.Patch, []k8scontrollerclient.PatchOption) {
+	// Apply patch type is not supported on the fake controller
+	if patch.Type() == types.ApplyPatchType {
+		patch = k8scontrollerclient.Merge
+		patchOptions := make([]k8scontrollerclient.PatchOption, 0, len(opts))
+		for _, opt := range opts {
+			if opt == k8scontrollerclient.ForceOwnership {
+				continue
+			}
+			patchOptions = append(patchOptions, opt)
+		}
+		opts = patchOptions
+	}
+	return patch, opts
+}
--- a/pkg/lib/controller-runtime/client/ssa.go
+++ b/pkg/lib/controller-runtime/client/ssa.go
@@ -1,0 +1,196 @@
+package client
+
+import (
+	"context"
+	"fmt"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	kscheme "k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/rest"
+	apiregistrationv1 "k8s.io/kube-aggregator/pkg/apis/apiregistration/v1"
+	"reflect"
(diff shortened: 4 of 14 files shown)


pkg/controller/registry/reconciler/grpc.go
```diff
@@ -199,11 +201,24 @@ func (c *GrpcRegistryReconciler) ensureUpdatePod(source grpcCatalogSourceDecorat
 
 			// Update the update pod to promote it to serving pod
 			_, err := c.OpClient.KubernetesInterface().CoreV1().Pods(source.GetNamespace()).Update(context.TODO(), updatePod, metav1.UpdateOptions{})
-			if err != nil {
-				return errors.Wrapf(err, "error creating new pod: %s", source.Pod().GetName())
+			if err == nil {
+				break
 			}
+			if !(apierrors.IsConflict(err) && err.Error() == registry.OptimisticLockErrorMsg) {
+				return errors.Wrapf(err, "error updating catalog source pod labels: %s", source.Pod().GetName())
```
Bowenislandsong, line 208: Why not use fmt.errorf()?

ankitathomas: I'm guessing that was for the stacktrace, we can switch to errorf

pkg/controller/registry/reconciler/grpc.go
```diff
@@ -198,12 +199,18 @@ func (c *GrpcRegistryReconciler) ensureUpdatePod(source grpcCatalogSourceDecorat
...
+			forceApply := true
+			opts := metav1.PatchOptions{
+				Force:        &forceApply,
+				FieldManager: "operator-lifecycle-manager-registry-reconciler",
 			}
 
-			break
+			updatePodBytes, err := client.Apply.Data(updatePod)
+			if err != nil {
+				return errors.Wrapf(err, "error updating catalog source pod: %s", source.Pod().GetName())
```
exdx, line 210: can we make this error log more specific like "error serializing updated catalog source pod to bytes: ..."

The wrapped error will have some more info, but I wouldn't want this error to look too similar to the one below.

pkg/controller/registry/reconciler/grpc.go
```diff
@@ -198,12 +199,18 @@ func (c *GrpcRegistryReconciler) ensureUpdatePod(source grpcCatalogSourceDecorat
...
+			}
+			_, err = c.OpClient.KubernetesInterface().CoreV1().Pods(source.GetNamespace()).Patch(context.TODO(), updatePod.Name, types.ApplyPatchType, updatePodBytes, opts)
+			return errors.Wrapf(err, "error updating catalog source pod: %s", source.Pod().GetName())
```
exdx, line 213: I think this is missing an `if err != nil`?

exdx:
I think we may need to update the object more before patching it, see #1641 for an example of a test with a similar server-side apply that passes locally, but would not patch successfully without the additional metadata.

pkg/controller/registry/reconciler/grpc.go
```diff
@@ -198,12 +199,20 @@ func (c *GrpcRegistryReconciler) ensureUpdatePod(source grpcCatalogSourceDecorat
...
+				Force:        &forceApply,
+				FieldManager: "operator-lifecycle-manager-registry-reconciler",
 			}
 
-			break
+			updatePodBytes, err := client.Apply.Data(updatePod)
+			if err != nil {
+				return errors.Wrapf(err, "error serializing updated catalog source pod to bytes: %s", source.Pod().GetName())
+			}
+			_, err = c.OpClient.KubernetesInterface().CoreV1().Pods(source.GetNamespace()).Patch(context.TODO(), updatePod.Name, types.ApplyPatchType, updatePodBytes, opts)
```
njhale, line 212: I recently wrote a utility function that gets rid of the boilerplate for performing SSAs in our tests. See https://github.com/operator-framework/operator-lifecycle-manager/pull/1641#discussion_r455507033 for more details. PTAL.

exdx: This is the context of the reconciler code though, not the e2e tests. Should we move the library to `pkg` somewhere and have access to it in multiple places? Rather than import a test library in the live code

ankitathomas: I've moved the SSA code over to a pkg/lib/util. It needs a full kube client to work, and the registry reconciler won't be able to use it without adding that in.

exdx: Seems like if we want to use nick’s SSA library in the code we need to plumb the client in by hanging it off the Operator object at the top-level of the catalog operator

pkg/lib/util/util.go
```diff
@@ -0,0 +1,177 @@
+package util
```
exdx, line 1: nit: can we call this file `ssa.go` or something more informative

pkg/lib/util/util.go
```diff
@@ -0,0 +1,177 @@
...
+		panic(fmt.Sprintf("unexpected number of return values in change function signature: expected 1, present %d", n))
+	}
+	var err error
+	if rt := changeType.Out(0); !rt.Implements(reflect.TypeOf((*error)(nil)).Elem()) {
+		panic(fmt.Sprintf("unexpected return type in change function signature: expected %t, present %s", err, rt))
+	}
+
+	// Determine if we need to apply a status subresource
+	// FIXME: use discovery client instead of relying on storage struct fields.
+	_, applyStatus := objType.Elem().FieldByName("Status")
```
njhale, line 113: we'll want to use the discovery client instead of checking the field value for this to work with `unstructured.Unstructured`

pkg/lib/util/util.go
```diff
@@ -0,0 +1,177 @@
...
+
+			return cp, nil
+		}
+
+		cp, err := changed(obj)
+		if err != nil {
+			return err
+		}
+
+		if err := client.Patch(bg, cp, controllerclient.Apply, controllerclient.ForceOwnership, controllerclient.FieldOwner("test")); err != nil {
```
njhale, line 153: Let's plumb the field owner in, either as a parameter, or as a struct field.

pkg/lib/util/util.go
```diff
@@ -0,0 +1,177 @@
...
+)
+
+var (
+	localSchemeBuilder = runtime.NewSchemeBuilder(
+		kscheme.AddToScheme,
+		apiextensionsv1.AddToScheme,
+		apiregistrationv1.AddToScheme,
+		operatorsv1alpha1.AddToScheme,
+		operatorsv1.AddToScheme,
+		operatorsv2alpha1.AddToScheme,
```
njhale, line 25: Since we're passing in a scheme, we can leave out references to OLM's types.

pkg/lib/util/util.go
ankitathomas: Since all we need is to check the status field, I'm checking for a status key if the object can be mapped to `unstructured.Unstructured`

pkg/controller/registry/reconciler/grpc.go
ankitathomas: I've added an SSA client to the `TestContext` used in our e2e tests. It should be easy to add it into the operator object if we need it there

dinhxuanvu:
/hold
Put a hold on this PR temporarily to ensure nothing getting merged into master while the new resolver PR is running CI. Would like to have that PR getting in first.

pkg/controller/registry/reconciler/grpc.go
```diff
@@ -198,12 +200,28 @@ func (c *GrpcRegistryReconciler) ensureUpdatePod(source grpcCatalogSourceDecorat
 			updatePod.Labels[CatalogSourceUpdateKey] = ""
 
 			// Update the update pod to promote it to serving pod
-			_, err := c.OpClient.KubernetesInterface().CoreV1().Pods(source.GetNamespace()).Update(context.TODO(), updatePod, metav1.UpdateOptions{})
+			forceApply := true
+			opts := metav1.PatchOptions{
+				Force:        &forceApply,
+				FieldManager: "operator-lifecycle-manager-registry-reconciler",
```
njhale, line 206: nit: we can shorten this down `olm.registry`? and this should either be a member of the `grpcCatalogSourceDecorator` or be defined as a constant.

pkg/lib/util/ssa.go
```diff
@@ -0,0 +1,187 @@
...
+// Ex. Change the spec of an existing InstallPlan
+//
+// plan := &InstallPlan{}
+// plan.SetNamespace("ns")
+// plan.SetName("install-123def")
+// plan.GetObjectKind().SetGroupVersionKind(schema.GroupVersionKind{
+//		Group:   GroupName,
+//		Version: GroupVersion,
+//		Kind:    InstallPlanKind,
+// })
```
njhale, line 77: nit: this section of comment is no longer necessary.

pkg/lib/util/ssa.go
```diff
@@ -0,0 +1,187 @@
+package util
```
njhale, line 1: I do think this needs to go in a different package. `lib/util` is a little to generic, especially when we _could_ have `lib/controller-runtime/client` package or something similarly descriptive.

pkg/lib/util/ssa_test.go
```diff
@@ -0,0 +1,36 @@
...
+
+import (
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"testing"
+)
+
+func TestSharedTime(t *testing.T) {
```
njhale, line 11: `SharedTime`?

pkg/lib/util/ssa.go
```diff
@@ -0,0 +1,187 @@
...
+	_, applyStatus := objType.Elem().FieldByName("Status")
+
+	if unstructuredObj, ok := obj.(*unstructured.Unstructured); ok {
+		_, applyStatus = unstructuredObj.Object["status"]
+	} else {
+		_, applyStatus = objType.Elem().FieldByName("Status")
+	}
+
+	var (
+		bg = context.Background()
```
njhale, line 122: If we use this outside of tests, `ctx` should be a formal parameter.

test/e2e/csv_e2e_test.go
```diff
@@ -1015,7 +1015,7 @@ var _ = Describe("CSV", func() {
 		require.True(GinkgoT(), ok, "expected olm sha annotation not present on existing pod template")
 
 		// Induce a cert rotation
-		Eventually(Apply(fetchedCSV, func(csv *v1alpha1.ClusterServiceVersion) error {
+		Eventually(ctx.Ctx().ServerSideApplier().Apply(fetchedCSV, func(csv *v1alpha1.ClusterServiceVersion) error {
```
njhale, line 1018: Could we make this read nicer by having an `Apply` method on `ctx.Ctx()` or performing this access in a top-level `Apply` function?

Bowenislandsong:
This PR **failed tests for 1 times** with 2 individual failed tests and 4 skipped tests. A test is considered flaky if failed on multiple commits. 
<details>

 totaltestcount: 1
flaketestcount: 2
skippedtestcount: 4
flaketests:
  - classname: End-to-end
    name: '**Subscription creation if not installed**'
    counts: 3
    details:
      - count: 3
        error: |4-


            /home/runner/work/operator-lifecycle-manager/operator-lifecycle-manager/test/e2e/subscription_e2e_test.go:52

            	Error Trace:	subscription_e2e_test.go:59
            	            				runner.go:113
            	            				runner.go:64
            	            				it_node.go:26
            	            				spec.go:215
            	            				spec.go:138
            	            				spec_runner.go:200
            	            				spec_runner.go:170
            	            				spec_runner.go:66
            	            				suite.go:62
            	            				ginkgo_dsl.go:226
            	            				ginkgo_dsl.go:214
            	            				e2e_test.go:54

    meandurationsec: 0.03773066666666667
  - classname: End-to-end
    name: '**Operator Group cleanup csvs with bad owner operator groups**'
    counts: 1
    details:
      - count: 1
        error: |4-


            /home/runner/work/operator-lifecycle-manager/operator-lifecycle-manager/test/e2e/operator_groups_e2e_test.go:1800

            	Error Trace:	operator_groups_e2e_test.go:2040
            	            				runner.go:113
            	            				runner.go:64
            	            				it_node.go:26
            	            				spec.go:215
            	            				spec.go:138
            	            				spec_runner.go:200
            	            				spec_runner.go:170
            	            				spec_runner.go:66
            	            				suite.go:62
            	            				ginkgo_dsl.go:226
            	            				ginkgo_dsl.go:214
            	            				e2e_test.go:54

    meandurationsec: 614.088057
skippedtests:
  - classname: End-to-end
    counts: 1
    details: []
    meandurationsec: 0.027671
  - classname: End-to-end
    counts: 1
    details: []
    meandurationsec: 0.465978
  - classname: End-to-end
        prometheus objects**'
    counts: 1
    details: []
    meandurationsec: 15.307669
  - classname: End-to-end
        successfully**'
    counts: 1
    details: []
    meandurationsec: 9.294765

pkg/controller/registry/reconciler/grpc.go
```diff
@@ -12,12 +11,42 @@ import (
 	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/api/core/v1"
+	kscheme "k8s.io/client-go/kubernetes/scheme"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/client-go/rest"
+	controllerclient "github.com/operator-framework/operator-lifecycle-manager/pkg/lib/controller-runtime/client"
```
njhale, line 20: nit: import order

pkg/controller/registry/reconciler/grpc.go
```diff
@@ -12,12 +11,42 @@ import (
...
+	k8scontrollerclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	CatalogSourceUpdateKey = "catalogsource.operators.coreos.com/update"
+	RegistryFieldManager = "olm.registry"
 )
+var (
+	scheme = runtime.NewScheme()
+	ssaClient *controllerclient.ServerSideApplier
```
njhale, line 30: This should be dependency injected or constructed from a client that is dependency injected.

pkg/lib/controller-runtime/client/ssa.go
```diff
@@ -0,0 +1,183 @@
...
+		cp, err := changed(obj)
+		if err != nil {
+			return err
+		}
+
+		if len(client.Owner) == 0 {
+			client.Owner = defaultOwner
+		}
+
+		if err := client.Patch(ctx, cp, controllerclient.Apply, controllerclient.ForceOwnership, client.Owner); err != nil {
```
njhale, line 159: note: later on we could probably be a little bit smarter about recognizing which bits are being changed -- spec + metadata, or status -- and only patch the the subresource that matters.

ankitathomas: easiest way for that would probably be a parameter + wrappers for apply status vs apply spec/metadata. Though, we may also be able to pass a dummy version of the object through and inspect the changed object in the change func.

pkg/controller/operators/catalog/operator.go
```diff
@@ -142,6 +148,32 @@ func NewOperator(ctx context.Context, kubeconfigPath string, clock utilclock.Clo
 		return nil, err
 	}
 
+	var restConfig *rest.Config
```
exdx, line 151: nit: Can we move the below logic into a standalone function?

If you see above, we have things like the `opClient` and `dynamicClient` that are returned from `NewForConfig` functions. Using that same pattern here would help the readability IMO

pkg/controller/registry/reconciler/grpc.go
```diff
@@ -17,7 +17,9 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
-const CatalogSourceUpdateKey = "catalogsource.operators.coreos.com/update"
+const (
```
exdx, line 20: nit: why is this being changed?

pkg/lib/controller-runtime/client/ssa_test.go
```diff
@@ -0,0 +1,35 @@
...
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			SetDefaultGroupVersionKind(tt.obj, tt.s)
+			require.EqualValues(t, tt.obj.GetObjectKind().GroupVersionKind(), tt.want)
+		})
+	}
+}
```
exdx, line 35: Having a unit test here that uses server side apply to mimic the same functionality that we want in the hot path (editing a pod label) via the fake client would be great. Not sure if we can SSA via the fake client. Would be nicer than another e2e test, but either approach works. I think the existing image polling e2e test covers the base case

pkg/controller/registry/reconciler/grpc.go
ankitathomas: That was from moving around constants, reformatting that to the old form

pkg/lib/controller-runtime/client/fake_ssa.go
```diff
@@ -0,0 +1,56 @@
...
+func (c *fakeApplier) Status() k8scontrollerclient.StatusWriter {
+	return fakeStatusWriter{c.Client.Status()}
+}
+
+type fakeStatusWriter struct {
+	k8scontrollerclient.StatusWriter
+}
+
+func (c fakeStatusWriter) Patch(ctx context.Context, obj runtime.Object, patch k8scontrollerclient.Patch, opts ...k8scontrollerclient.PatchOption) error {
+	patch, opts = convertApplyToMergePatch(patch, opts...)
```
njhale, line 38: I'm not sure an apply patch can be directly converted to a merge patch (since I think you need both the old and new resource to generate the delta for a merge patch), but if this is working, I think it might be okay.

I'm a little concerned that we're swapping SSA out for another patch type in our unit tests, but at least it lets us test the rest of the logic in our unit tests.

test/e2e/ctx/ctx.go
```diff
@@ -1,7 +1,9 @@
 package ctx
 
 import (
+	"context"
 	"fmt"
+	controllerclient "github.com/operator-framework/operator-lifecycle-manager/pkg/lib/controller-runtime/client"
```
njhale, line 6: nit: import order

test/e2e/util_test.go
```diff
@@ -872,144 +871,33 @@ func deploymentReplicas(replicas int32) predicateFunc {
...
-				return nil, err.(error)
-			}
+const (
+	cvoNamespace      = "openshift-cluster-version"
+	cvoDeploymentName = "cluster-version-operator"
+)
 
-			return cp, nil
-		}
+func toggleCVO() {
```
njhale, line 879: this function should have been removed when the `Operators` API was released under the `v1` group.

njhale, line 879: This function has been removed in `master`.

test/e2e/util_test.go
```diff
@@ -872,144 +871,33 @@ func deploymentReplicas(replicas int32) predicateFunc {
...
-				panic(fmt.Sprintf("unexpected number of return values from apply mutation func: expected 1, returned %d", len(out)))
-			}
-
-			if err := out[0].Interface(); err != nil {
-				return nil, err.(error)
-			}
+const (
+	cvoNamespace      = "openshift-cluster-version"
+	cvoDeploymentName = "cluster-version-operator"
+)
```
njhale, line 877: also removed in `master`

test/e2e/crd_e2e_test.go
```diff
@@ -299,7 +299,7 @@ var _ = Describe("CRD Versions", func() {
 		// old CRD has been installed onto the cluster - now upgrade the subscription to point to the channel with the new CRD
 		// installing the new CSV should fail with a warning about data loss, since a storage version is missing in the new CRD
 		// use server-side apply to apply the update to the subscription point to the alpha channel
-		Eventually(Apply(subscription, func(s *operatorsv1alpha1.Subscription) error {
+		Eventually(ctx.Ctx().Apply(subscription, func(s *operatorsv1alpha1.Subscription) error {
```
njhale, line 302: This is unnecessarily verbose. I think we can write a wrapper function that gives us back:

```go
Eventually(Apply(subscription, func(s *operatorsv1alpha1.Subscription) error {
```

pkg/lib/controller-runtime/client/ssa.go
```diff
@@ -0,0 +1,202 @@
...
+		// TODO: The same object can be registered for multiple group versions
+		// (although in practise this doesn't seem to be used).
+		// In such case, the version set may not be correct.
+		gvk = gvks[0]
+	}
+	obj.GetObjectKind().SetGroupVersionKind(gvk)
+}
+
+type ServerSideApplier struct {
+	k8scontrollerclient.Client
```
njhale, line 76: Is there a reason for embedding this? Do we plan on using `ServerSideApplier` as a unified client everywhere?

ankitathomas: This was mainly so that the testcontext could use its k8s controller client as the `ServerSideApplier`. We can create a separate one just for doing the apply.

test/e2e/util_test.go
```diff
@@ -872,144 +872,6 @@ func deploymentReplicas(replicas int32) predicateFunc {
...
-		if err := client.Status().Patch(bg, cp, controllerclient.Apply, controllerclient.ForceOwnership, controllerclient.FieldOwner("test")); err != nil {
-			fmt.Printf("second patch error: %s\n", err)
-			return err
-		}
-
-		reflect.ValueOf(obj).Elem().Set(reflect.ValueOf(cp).Elem())
-
-		return nil
-	}
+func Apply(obj controllerclient.Object, changeFunc interface{}) func() error {
```
njhale, line 875: 👍

PR #1631, opened 2020-07-10
fix: check installplan status in bundle e2e test

exdx:

Reduce flakiness of bundle with additional objects e2e test. 

**Reviewer Checklist**

### DIFF
--- a/test/e2e/bundle_e2e_test.go
+++ b/test/e2e/bundle_e2e_test.go
@@ -2,12 +2,13 @@ package e2e
 
 import (
 	"context"
-	"k8s.io/apimachinery/pkg/runtime/schema"
-	"k8s.io/client-go/dynamic"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
 
 	"github.com/operator-framework/api/pkg/operators/v1alpha1"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/clientset/versioned"
@@ -45,6 +46,8 @@ var _ = Describe("Installing bundles with new object types", func() {
 				sourceName = "test-catalog"
 				imageName  = "quay.io/olmtest/single-bundle-index:pdb"
 			)
+
+			var installPlanRef string
 			// create catalog source
 			source := &v1alpha1.CatalogSource{
 				TypeMeta: metav1.TypeMeta{
@@ -69,10 +72,16 @@ var _ = Describe("Installing bundles with new object types", func() {
 			_ = createSubscriptionForCatalog(operatorClient, source.GetNamespace(), subName, source.GetName(), packageName, channelName, "", v1alpha1.ApprovalAutomatic)
 
 			// Wait for the Subscription to succeed
-			Eventually(func() error {
-				_, err = fetchSubscription(operatorClient, testNamespace, subName, subscriptionStateAtLatestChecker)
-				return err
-			}).Should(BeNil())
+			sub, err := fetchSubscription(operatorClient, testNamespace, subName, subscriptionStateAtLatestChecker)
+			Expect(err).ToNot(HaveOccurred(), "could not get subscription at latest status")
+
+			installPlanRef = sub.Status.InstallPlanRef.Name
+
+			// Wait for the installplan to complete (5 minute timeout)
+			_, err = fetchInstallPlan(GinkgoT(), operatorClient, installPlanRef, buildInstallPlanPhaseCheckFunc(v1alpha1.InstallPlanPhaseComplete))
+			Expect(err).ToNot(HaveOccurred(), "could not get installplan at complete phase")
+
+			ctx.Ctx().Logf("install plan %s completed", installPlanRef)
 		})
 
 		It("should create the additional bundle objects", func() {


test/e2e/bundle_e2e_test.go
```diff
@@ -70,9 +73,19 @@ var _ = Describe("Installing bundles with new object types", func() {
...
+				sub, err := fetchSubscription(operatorClient, testNamespace, subName, subscriptionStateAtLatestChecker)
+				if sub != nil && sub.Status.InstallPlanRef != nil {
+					installPlanRef = sub.Status.InstallPlanRef.Name
+				}
+				return err
+			}).Should(BeNil())
+
+			// Wait for the installplan to complete (5 minute timeout)
+			Eventually(func() error {
+				_, err := fetchInstallPlan(GinkgoT(), operatorClient, installPlanRef, buildInstallPlanPhaseCheckFunc(v1alpha1.InstallPlanPhaseComplete))
```
njhale, line 85: Can a subscription be at latest if the InstallPlan isn't complete?

test/e2e/bundle_e2e_test.go
```diff
@@ -70,9 +73,19 @@ var _ = Describe("Installing bundles with new object types", func() {
 
 			// Wait for the Subscription to succeed
 			Eventually(func() error {
-				_, err = fetchSubscription(operatorClient, testNamespace, subName, subscriptionStateAtLatestChecker)
+				sub, err := fetchSubscription(operatorClient, testNamespace, subName, subscriptionStateAtLatestChecker)
+				if sub != nil && sub.Status.InstallPlanRef != nil {
+					installPlanRef = sub.Status.InstallPlanRef.Name
+				}
+				return err
+			}).Should(BeNil())
```
njhale, line 81:
```suggestion
			}).Should(Succeed())
```

test/e2e/bundle_e2e_test.go
```diff
@@ -70,9 +73,19 @@ var _ = Describe("Installing bundles with new object types", func() {
 
 			// Wait for the Subscription to succeed
 			Eventually(func() error {
-				_, err = fetchSubscription(operatorClient, testNamespace, subName, subscriptionStateAtLatestChecker)
+				sub, err := fetchSubscription(operatorClient, testNamespace, subName, subscriptionStateAtLatestChecker)
```
njhale, line 76: hmm, fetchSubscription already polls, so I think the `Eventually` matcher is redundant here.

njhale:
```
2020-07-10T15:22:31.9409853Z     [91mTimed out after 60.000s.
2020-07-10T15:22:31.9410158Z     expected no error finding vpa object associated with csv
2020-07-10T15:22:31.9410383Z     Expected success, but got an error:
2020-07-10T15:22:31.9410835Z         <*errors.StatusError | 0xc000b85cc0>: {
2020-07-10T15:22:31.9410958Z             ErrStatus: {
2020-07-10T15:22:31.9411081Z                 TypeMeta: {Kind: "Status", APIVersion: "v1"},
2020-07-10T15:22:31.9411200Z                 ListMeta: {
2020-07-10T15:22:31.9411315Z                     SelfLink: "",
2020-07-10T15:22:31.9411432Z                     ResourceVersion: "",
2020-07-10T15:22:31.9411545Z                     Continue: "",
2020-07-10T15:22:31.9411636Z                     RemainingItemCount: nil,
2020-07-10T15:22:31.9411752Z                 },
2020-07-10T15:22:31.9411863Z                 Status: "Failure",
2020-07-10T15:22:31.9412203Z                 Message: "verticalpodautoscalers.autoscaling.k8s.io \"busybox-vpa\" not found",
2020-07-10T15:22:31.9412339Z                 Reason: "NotFound",
2020-07-10T15:22:31.9412654Z                 Details: {
2020-07-10T15:22:31.9412905Z                     Name: "busybox-vpa",
2020-07-10T15:22:31.9413172Z                     Group: "autoscaling.k8s.io",
2020-07-10T15:22:31.9413299Z                     Kind: "verticalpodautoscalers",
2020-07-10T15:22:31.9413416Z                     UID: "",
2020-07-10T15:22:31.9413531Z                     Causes: nil,
2020-07-10T15:22:31.9413647Z                     RetryAfterSeconds: 0,
2020-07-10T15:22:31.9413757Z                 },
2020-07-10T15:22:31.9413867Z                 Code: 404,
2020-07-10T15:22:31.9413976Z             },
2020-07-10T15:22:31.9414060Z         }
2020-07-10T15:22:31.9414530Z         verticalpodautoscalers.autoscaling.k8s.io "busybox-vpa" not found[0m
```

^ Sample failure here. If this is what we're seeing in general, I think there may be a deeper issue with conflicting test setup or even with OLM's reconciliation. The fact that it still doesn't exist after 60 seconds is suspicious to me.

test/e2e/bundle_e2e_test.go
exdx: I assumed so but let me double check - that's basically my concern with this fix

exdx:
I was troubleshooting this with @ecordell locally and it looks like the VPA CRD installplan stepresource gets stuck in `WaitingforAPI` even though the VPA CRD itself is established and accepted on-cluster. So this causes the e2e test to fail (on the first run) locally. However it sometimes passes, as we've seen when it gets through CI.

I think there was also a question of whether the VPA is shipped in OpenShift by default and the provided CRD is unnecessary - but I could not find a reference to the VPA CRD in a stock clusterbot cluster.

test/e2e/bundle_e2e_test.go
exdx: From what I can see the, the answer is yes. The `SubscriptionStateAtLatest` is concerned only with whether an update exists for that given CSV rather than the actual status of its InstallPlan. 

see https://github.com/operator-framework/operator-lifecycle-manager/blob/e72d31aee442945cd9757f47bd075ec15ec0ea1c/pkg/controller/operators/catalog/operator.go#L939

njhale:
> WaitingforAPI even though the VPA CRD itself is established and accepted on-cluster. So this causes the e2e test to fail (on the first run) locally. However it sometimes passes, as we've seen when it gets through CI.

Do we ensure CRDs are ordered before any CRs? If so, what happens if the CR is applied by the InstallPlan while the API is unavailable? Are we resyncing InstallPlans in this state when CRDs they've applied have changed?

exdx:
> > WaitingforAPI even though the VPA CRD itself is established and accepted on-cluster. So this causes the e2e test to fail (on the first run) locally. However it sometimes passes, as we've seen when it gets through CI.
> 
> Do we ensure CRDs are ordered before any CRs? If so, what happens if the CR is applied by the InstallPlan while the API is unavailable? Are we resyncing InstallPlans in this state when CRDs they've applied have changed?

We do ensure that CRDs are created first (they are the first thing that gets reconciled in `ExecutePlan`). We don't currently have a check in our dynamic object creation that the CR that is attempting to be created actually has the associated CRD in a ready and established state on the cluster (which I think you're getting at). So potentially we have a race here, where sometimes the VPA CRD is still in `WaitingforApi` before the VPA CR is created from the bundle. That would explain the flakiness. I wonder if we would have this issue with the prometheus objects @awgreene added as well though? 

Yes, we should be re-syncing the installplan with respect to the CRDs. The code certain looks that way. The issue here is that it looks like the sync is not happening fast enough?

exdx:
So with this change instead of creating VPA CRs before the VPA CRD is Complete, the InstallPlan check will fail and the test will fail there. 

A little confusing since a lot of the flakes we saw where around the PDB not so much the VPA. I think this fix helps with the PDB check (we were checking too early, before the installplan got a chance to create it) but uncovered another issue with the VPA object.

exdx:
Personally I think this PR addresses the original intended issues and makes things marginally better so we could merge and address the potential race we uncovered in a follow-up. Thoughts @njhale @ecordell

PR #1635, opened 2020-07-13
test(e2e): delete all subs, installplans, catalogs, and csvs in test teardown

njhale:
Replace the existing test cleaner, which preserved test resources on
failure, with a tear down function that always deletes
non-essential Subscriptions, InstallPlans, CatalogSources, and
ClusterServiceVersions in the given test namespace.

### DIFF
--- a/test/e2e/catalog_e2e_test.go
+++ b/test/e2e/catalog_e2e_test.go
@@ -29,10 +29,11 @@ import (
 )
 
 var _ = Describe("Catalog", func() {
+	AfterEach(func() {
+		TearDown(testNamespace)
+	})
+
 	It("loading between restarts", func() {
-
-		defer cleaner.NotifyTestComplete(true)
-
 		// create a simple catalogsource
 		packageName := genName("nginx")
 		stableChannel := "stable"
@@ -87,8 +88,6 @@ var _ = Describe("Catalog", func() {
 		GinkgoT().Logf("Catalog source sucessfully loaded after rescale")
 	})
 	It("global update triggers subscription sync", func() {
-
-		defer cleaner.NotifyTestComplete(true)
 
 		globalNS := operatorNamespace
 		c := newKubeClient()
@@ -196,8 +195,6 @@ var _ = Describe("Catalog", func() {
 		require.True(GinkgoT(), timeLapse < 60)
 	})
 	It("config map update triggers registry pod rollout", func() {
-
-		defer cleaner.NotifyTestComplete(true)
 
 		mainPackageName := genName("nginx-")
 		dependentPackageName := genName("nginxdep-")
@@ -316,8 +313,6 @@ var _ = Describe("Catalog", func() {
 	})
 	It("config map replace triggers registry pod rollout", func() {
 
-		defer cleaner.NotifyTestComplete(true)
-
 		mainPackageName := genName("nginx-")
 		dependentPackageName := genName("nginxdep-")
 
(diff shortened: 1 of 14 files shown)


test/e2e/util_test.go
```diff
@@ -369,76 +337,169 @@ func fetchCatalogSourceOnStatus(crc versioned.Interface, name, namespace string,
...
 
-	err = waitForEmptyList(func() (int, error) {
-		res, err := crc.OperatorsV1alpha1().CatalogSources(namespace).List(context.TODO(), metav1.ListOptions{FieldSelector: nonPersistentCatalogsFieldSelector})
-		ctx.Ctx().Logf("%d %s remaining", len(res.Items), "catalogs")
-		return len(res.Items), err
-	})
-	Expect(err).NotTo(HaveOccurred())
 }
 
+// func cleanupOLM(namespace string) {
```
ecordell, line 447: probably just delete this?

test/e2e/util_test.go
```diff
@@ -369,76 +337,169 @@ func fetchCatalogSourceOnStatus(crc versioned.Interface, name, namespace string,
...
+		Selector: createFieldNotEqualSelector("metadata.name", persistentConfigMapNames...),
+	}
+	persistentCSVNames        = []string{packageServerCSV}
+	ephemeralCSVFieldSelector = controllerclient.MatchingFieldsSelector{
+		Selector: createFieldNotEqualSelector("metadata.name", persistentCSVNames...),
+	}
+)
+
+// TearDown deletes all OLM resources in the corresponding namespace and at the cluster scope.
+func TearDown(namespace string) {
```
ecordell, line 379: it would be useful if there were a way to control this (i.e. via an env var) so that locally I can quickly switch cleanup off to look at the cluster state.

And I think really what we want is a sort of "must-gather" per test. Doesn't need to block this PR but I'd be interesting in something that say, read all of this content off the cluster and saved it to configmaps before deleting everything.

njhale: @benluddy and I were just talking about this over Slack.

He recommended implementing a [Reporter](https://pkg.go.dev/github.com/onsi/ginkgo/reporters?tab=doc#Reporter) to do this. We were thinking that for an initial pass we could export all resources in the `operators.coreos.com` API group to a tar file hosted in cloud storage at the end of each failed spec, then the reporter could post a link to each export in the suite summary.

ecordell: That sounds good, but it would be nice if we can make it work across github actions, openshift ci, and locally.

benluddy: > That sounds good, but it would be nice if we can make it work across github actions, openshift ci, and locally.

It may be good enough to accept a template string via environment variables to construct those URLs and set it appropriately in the corresponding config.

test/e2e/catalog_e2e_test.go
```diff
@@ -31,7 +31,7 @@ import (
 var _ = Describe("Catalog", func() {
 	It("loading between restarts", func() {
 
-		defer cleaner.NotifyTestComplete(true)
+		defer TearDown(testNamespace)
```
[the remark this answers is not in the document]
njhale: > Can these all turn into one AfterEach since you're touching them anyway?

Yeah, can do.

>  but if this cleanup behavior is supposed to be idempotent, that seems like an overcomplication.

I agree, but I think we need to consider cleaning these up carefully though. Not all tests use `testNamespace`, but instead generate one or more namespaces on the fly, which is what I think some of these `defer` statements service. I n general, I think we should standardize on one of the two above patterns across all of our tests. Suggestions?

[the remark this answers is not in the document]
benluddy: Can we use the TestContext to provide namespaces to tests? We'd be able to make cleanup, randomization, etc. transparent to individual tests.

PR #1637, opened 2020-07-14
fix: install VPA CRD directly on clusterfor bundle e2e test

exdx:
This fix installs the VPA CRD on-cluster directly (instead of providing it in the bundle) in the bundle e2e test. There is a check to ensure the VPA CRD is not already present. The index contents were updated alongside this PR, to remove the packaged CRD. 

**Reviewer Checklist**


test/e2e/bundle_e2e_test.go
```diff
@@ -34,7 +105,7 @@ var _ = Describe("Installing bundles with new object types", func() {
 	})
 
 	When("a bundle with a pdb, priorityclass, and VPA object is installed", func() {
-		By("including the VPA CRD in the CSV")
+		By("first installing the VPA CRD on cluster")
```
benluddy, line 108: This call to By should be inside the BeforeEach so that it happens at test runtime instead of suite load time.

test/e2e/bundle_e2e_test.go
```diff
@@ -47,6 +118,30 @@ var _ = Describe("Installing bundles with new object types", func() {
 				imageName  = "quay.io/olmtest/single-bundle-index:pdb"
 			)
 
+			// create VPA CRD on cluster
+			data, err := yaml.YAMLToJSON([]byte(vpaCRDManifest))
+			Expect(err).ToNot(HaveOccurred(), "could not convert vpa crd to json")
+
+			crd := &apiextensionsv1beta1.CustomResourceDefinition{}
+			err = json.Unmarshal(data, crd)
+			Expect(err).ToNot(HaveOccurred(), "could not convert vpa crd to unstructured")
```
njhale, line 127: You're unmarshaling it to a CRD, not an unstructured -- though it's probably better to use an unstructured to make bumping the fixture version easier.

test/e2e/bundle_e2e_test.go
```diff
@@ -47,6 +118,30 @@ var _ = Describe("Installing bundles with new object types", func() {
...
+			}
+
+			// ensure vpa crd is established and accepted on the cluster before continuing
+			Eventually(func () bool {
+				crd, err := kubeClient.ApiextensionsInterface().ApiextensionsV1beta1().CustomResourceDefinitions().Get(context.TODO(), crd.GetName(), metav1.GetOptions{})
+				if err != nil {
+					return false
+				}
+				return crdReady(&crd.Status)
+			}).Should(BeTrue())
```
benluddy, line 143: Returning the error as a second return value from the assertion will make any failure reports more useful, like this:
```suggestion
			Eventually(func () bool, error {
				crd, err := kubeClient.ApiextensionsInterface().ApiextensionsV1beta1().CustomResourceDefinitions().Get(context.TODO(), crd.GetName(), metav1.GetOptions{})
				if err != nil {
					return false, err
				}
				return crdReady(&crd.Status), nil
			}).Should(BeTrue())
```

test/e2e/bundle_e2e_test.go
```diff
@@ -16,6 +20,73 @@ import (
 	"github.com/operator-framework/operator-lifecycle-manager/test/e2e/ctx"
 )
 
+const vpaCRDManifest = `
```
njhale, line 23: can we store this fixture in a `testdata` directory?

test/e2e/bundle_e2e_test.go
```diff
@@ -47,6 +118,30 @@ var _ = Describe("Installing bundles with new object types", func() {
 				imageName  = "quay.io/olmtest/single-bundle-index:pdb"
 			)
 
+			// create VPA CRD on cluster
+			data, err := yaml.YAMLToJSON([]byte(vpaCRDManifest))
```
njhale, line 122: This should all probably go in a `BeforeEach`, and I think we'll also want to delete this specific CRD in `AfterEach` too.

test/e2e/bundle_e2e_test.go
```diff
@@ -47,6 +118,30 @@ var _ = Describe("Installing bundles with new object types", func() {
...
+			Expect(err).ToNot(HaveOccurred(), "could not convert vpa crd to json")
+
+			crd := &apiextensionsv1beta1.CustomResourceDefinition{}
+			err = json.Unmarshal(data, crd)
+			Expect(err).ToNot(HaveOccurred(), "could not convert vpa crd to unstructured")
+
+			_, err = kubeClient.ApiextensionsInterface().ApiextensionsV1beta1().CustomResourceDefinitions().Create(context.TODO(), crd, metav1.CreateOptions{})
```
njhale, line 129: Can we use the [controller-runtime client](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/test/e2e/ctx/ctx.go#L78) instead?

test/e2e/bundle_e2e_test.go
```diff
@@ -47,6 +118,30 @@ var _ = Describe("Installing bundles with new object types", func() {
...
+			if err != nil {
+				if !k8serrors.IsAlreadyExists(err) {
+					Expect(err).ToNot(HaveOccurred(), "could not create vpa crd to on cluster")
+				}
+			}
+
+			// ensure vpa crd is established and accepted on the cluster before continuing
```
njhale, line 136: Are we just masking an issue with `InstallPlan` reconciliation? e.g. `InstallPlans` should eventually install CRs once a CRD becomes available.

test/e2e/bundle_e2e_test.go
```diff
@@ -65,7 +160,7 @@ var _ = Describe("Installing bundles with new object types", func() {
 				},
 			}
 
-			source, err := operatorClient.OperatorsV1alpha1().CatalogSources(source.GetNamespace()).Create(context.TODO(), source, metav1.CreateOptions{})
+			source, err = operatorClient.OperatorsV1alpha1().CatalogSources(source.GetNamespace()).Create(context.TODO(), source, metav1.CreateOptions{})
```
njhale, line 163: Needs to be in an `Eventually` block.

exdx:
> Thanks for taking the time to look at this.
> 
> It does make me wonder how this would ever have passed in our upstream tests, was the VPA CRD pard of the bundle?

Yes, originally it was included in the bundle as a provided API on the busybox CSV

test/e2e/bundle_e2e_test.go
exdx: Ok, that was my original thought but I decided to stick it in the file because I was unsure if the file would be there in the e2e test execution context on prow. Will try that approach.

test/e2e/bundle_e2e_test.go
exdx: They should but I believe we currently have an issue where the user provides the VPA CRD in the same bundle as the VPA CR. Although I guess in practice this won't be the case? Maybe this is just an issue with the way the test was written originally. I'm not sure having this check hurts though.

exdx: It does result in using the `v1beta1` client to check the status, but I don't think its a big deal. When we move off `v1beta1` we will require a lot of updates to our e2e suite.

test/e2e/bundle_e2e_test.go
```diff
@@ -34,19 +41,49 @@ var _ = Describe("Installing bundles with new object types", func() {
 	})
 
 	When("a bundle with a pdb, priorityclass, and VPA object is installed", func() {
-		By("including the VPA CRD in the CSV")
 		const (
 			packageName = "busybox"
 			channelName = "alpha"
 			subName     = "test-subscription"
 		)
+		var u unstructured.Unstructured
```
njhale, line 49: `u` could be anything, I think it will be easier to read if this variable name is more descriptive

test/e2e/bundle_e2e_test.go
```diff
@@ -116,5 +155,33 @@ var _ = Describe("Installing bundles with new object types", func() {
 				return err
 			}).Should(Succeed(), "expected no error finding vpa object associated with csv")
 		})
+
+		AfterEach(func() {
+			By("Deleting the VPA CRD")
+			Eventually(func() error {
+				err := ctx.Ctx().Client().Delete(context.TODO(), &u)
+				return err
+			}).Should(Succeed())
```
njhale, line 164: We should probably filter out `NotFound` errors here; e.g.:

```go
Eventually(func() (err error) {
    if err = ctx.Ctx().Client().Delete(context.Background(), &u); apierrors.IsNotFound(err) {
        // We're good if we get a NotFound error
        err = nil
    }
    return
}).Should(Succeed())
```

test/e2e/data/vpa/crd.yaml
```diff
@@ -0,0 +1,64 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
njhale, line 1: `testdata` is the cannonical directory for fixture files in go

test/e2e/bundle_e2e_test.go
```diff
@@ -34,19 +41,49 @@ var _ = Describe("Installing bundles with new object types", func() {
...
 
 		BeforeEach(func() {
+			By("first installing the VPA CRD on cluster")
 			const (
 				sourceName = "test-catalog"
 				imageName  = "quay.io/olmtest/single-bundle-index:pdb"
 			)
 
+			// create VPA CRD on cluster
+			y, err := ioutil.ReadFile("data/vpa/crd.yaml")
```
njhale, line 59: this can't be found in the upstream tests. I suspect we'll need to pull it into the test binary with go `bindata`.

exdx:
Turns out the name of the priority class object changed, that's why we were seeing those additional 404 errors. Changed to the correct name in this PR as well.

test/e2e/testdata/vpa/crd.yaml
```diff
@@ -0,0 +1,64 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
njhale, line 1: Note: we'll want to add a make rule to `x.mk` to regenerate this.

PR #1641, opened 2020-07-15
fix: use server-side apply in crd e2e test 

exdx:
Use server-side apply in crd e2e test when changing the subscription to point to a new channel to avoid transient object modified errors (409 errors). 

This change can probably be ported to other tests that flake with this error.

**Reviewer Checklist**

### DIFF
--- a/test/e2e/crd_e2e_test.go
+++ b/test/e2e/crd_e2e_test.go
@@ -10,6 +10,8 @@ import (
 	. "github.com/onsi/gomega"
 	operatorsv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/registry"
+	"github.com/operator-framework/operator-lifecycle-manager/test/e2e/ctx"
+
 	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
@@ -299,13 +301,12 @@ var _ = Describe("CRD Versions", func() {
 
 		// old CRD has been installed onto the cluster - now upgrade the subscription to point to the channel with the new CRD
 		// installing the new CSV should fail with a warning about data loss, since a storage version is missing in the new CRD
-		sub, err := crc.OperatorsV1alpha1().Subscriptions(testNamespace).Get(context.TODO(), subscription.GetName(), metav1.GetOptions{})
-		Expect(err).ToNot(HaveOccurred(), "could not get subscription")
-
-		// update channel on sub
-		sub.Spec.Channel = alphaChannel
-		_, err = crc.OperatorsV1alpha1().Subscriptions(testNamespace).Update(context.TODO(), sub, metav1.UpdateOptions{})
-		Expect(err).ToNot(HaveOccurred(), "could not update subscription")
+		// use server-side apply to apply the update to the subscription point to the alpha channel
+		Eventually(Apply(subscription, func(s *operatorsv1alpha1.Subscription) error {
+			s.Spec.Channel = alphaChannel
+			return nil
+		})).Should(Succeed())
+		ctx.Ctx().Logf("updated subscription to point to alpha channel")
 
 		subscriptionAtLatestWithDifferentInstallPlan := func(v *operatorsv1alpha1.Subscription) bool {
 			return subscriptionStateAtLatestChecker(v) && v.Status.InstallPlanRef != nil && v.Status.InstallPlanRef.Name != fetchedInstallPlan.Name


Bowenislandsong:
The PR **failed tests for 0 times** with 0 individual failed tests and 0 skipped tests.
<details>

 totaltestcount: 0
flaktestcount: 0
skippedtestcount: 0
flaktests: []
skippedtests: []

Bowenislandsong:
The PR **failed tests for 0 times** with 0 individual failed tests and 0 skipped tests.
<details>

 totaltestcount: 0
flaktestcount: 0
skippedtestcount: 0
flaktests: []
skippedtests: []

test/e2e/crd_e2e_test.go
```diff
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"time"
 
+	"k8s.io/apimachinery/pkg/types"
```
ecordell, line 8: nit: goimports

test/e2e/crd_e2e_test.go
```diff
@@ -304,8 +307,21 @@ var _ = Describe("CRD Versions", func() {
...
-		Expect(err).ToNot(HaveOccurred(), "could not update subscription")
+		sub.TypeMeta.APIVersion = "operators.coreos.com/v1alpha1"
+		sub.TypeMeta.Kind = "Subscription"
+		sub.ManagedFields = []metav1.ManagedFieldsEntry{}
+
+		// use server-side apply to apply the update to the subscription point to the alpha channel
+		forceApply := true
+		opts := metav1.PatchOptions{
+			Force:        &forceApply,
+			FieldManager: "operator-lifecycle-manager-registry-reconciler",
```
ecordell, line 318: Shoudl the field manager be "tests" since this client is not OLM?

test/e2e/crd_e2e_test.go
```diff
@@ -304,8 +307,21 @@ var _ = Describe("CRD Versions", func() {
...
-		_, err = crc.OperatorsV1alpha1().Subscriptions(testNamespace).Update(context.TODO(), sub, metav1.UpdateOptions{})
-		Expect(err).ToNot(HaveOccurred(), "could not update subscription")
+		sub.TypeMeta.APIVersion = "operators.coreos.com/v1alpha1"
+		sub.TypeMeta.Kind = "Subscription"
+		sub.ManagedFields = []metav1.ManagedFieldsEntry{}
+
+		// use server-side apply to apply the update to the subscription point to the alpha channel
+		forceApply := true
+		opts := metav1.PatchOptions{
+			Force:        &forceApply,
```
ecordell, line 317: The apply should only need to be forced if anyone else is writing into the spec, right? Which in this test we know isn't happening.

exdx:
```
              Message: "Apply failed with 1 conflict: conflict with \"e2e.test\" using operators.coreos.com/v1alpha1: .spec.channel",
```
Looks like force is required?

ecordell: that makes sense.  I think if we used apply to create the subscription in the first place, then it would succeed, but this works.

test/e2e/crd_e2e_test.go
```diff
@@ -304,8 +307,21 @@ var _ = Describe("CRD Versions", func() {
...
+			FieldManager: "tests",
+		}
+		newSub, err := client.Apply.Data(sub)
+		Expect(err).ToNot(HaveOccurred(), "could not serialize updated subscription to bytes")
+
+		_, err = crc.OperatorsV1alpha1().Subscriptions(testNamespace).Patch(context.TODO(), sub.GetName(), types.ApplyPatchType, newSub, opts)
```
njhale, line 323: @exdx I built a [test utility](https://github.com/operator-framework/operator-lifecycle-manager/blob/c3bcd2eaacd0b6ea326b0a521f97d10b8d18d483/test/e2e/util_test.go#L916) that handles server-side apply generically, sans boilerplate. Is there a reason we shouldn't/can't use that here too?

Ex.

```go
Eventually(Apply(sub, func(s *operatorsv1alpha1.Subscription) error {
    s.Spec.Channel = alphaChannel
    s.TypeMeta.APIVersion = "operators.coreos.com/v1alpha1"
    s.TypeMeta.Kind = "Subscription"
    s.ManagedFields = []metav1.ManagedFieldsEntry{}
})).Should(Succeed())
```

exdx: Should this be returning an error somehow?

njhale: typo, sorry

test/e2e/crd_e2e_test.go
```diff
@@ -302,10 +304,15 @@ var _ = Describe("CRD Versions", func() {
 		sub, err := crc.OperatorsV1alpha1().Subscriptions(testNamespace).Get(context.TODO(), subscription.GetName(), metav1.GetOptions{})
```
njhale, line 304: You don't need to do this. `Apply` handles it. you should be able to use the `subscription` variable declared above.

exdx: Nice

PR #1642, opened 2020-07-15
Bug 1857424: fix(resolver): Exclude all installed packages in dependency search

dinhxuanvu:
Currently, resolver only excludes one installed package at a time during
dependency search. However, this is not enough as there is a corner case
where there are two operators that are requiring the same CRD(s). As a
result, resolver only excludes one package but not the other(s).

Now, the list of installed packages in the same namespace will be added
to list of packages that are not considered during dependency
search.

**Reviewer Checklist**

### DIFF
--- a/test/e2e/subscription_e2e_test.go
+++ b/test/e2e/subscription_e2e_test.go
@@ -1244,9 +1244,9 @@ var _ = Describe("Subscription", func() {
 		require.Len(GinkgoT(), installPlan.Status.CatalogSources, 1)
 	})
 
-	// CatSrc:
+	// CatSrc1:
 	//
-	// Package A (apackage)
+	// Package 1 (apackage)
 	// Default Channel: Stable
 	// Channel Stable:
 	// Operator A (Requires: CRD 1, CRD 2 )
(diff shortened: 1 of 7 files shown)


dinhxuanvu:
@ecordell Ready for review. PTAL. Thanks

dinhxuanvu:
@ecordell @awgreene Ready for review.

test/e2e/subscription_e2e_test.go
```diff
@@ -1423,6 +1441,187 @@ var _ = Describe("Subscription", func() {
...
+	// Create subscription for csvB lead to installation of csvB and csvA
+	// Update catsrc to upgrade csvA to csvNewA which now requires CRD1
+	// csvNewA can't be installed due to no other operators provide CRD1 for it
+	// (Note: OLM can't pick csvA as dependency for csvNewA as it is from the same
+	// same package)
+	// Update catsrc again to upgrade csvB to csvNewB which now owns both CRD1 and
+	// CRD2.
+	// Now csvNewA and csvNewB are installed successfully as csvNewB provides CRD1
+	// that csvNewA requires
+	It("creation in case of transferring providedAPIs", func() {
```
awgreene, line 1455: This is an awesome test!

ecordell:
Validated that runaway installplans are not generated: 

image

PR #1667, opened 2020-07-22
Build per-test bundles and index images

ankitathomas:
Adding support for creating bundles and index images from test bundle directories to remove dependency of tests on pre-built images from Quay

Building test bundles and indexes in the context of an e2e test will help avoid test flakes due to Quay connectivity issues. This should also make it easier for contributors to add tests without having to rely on availability of pre-built bundles and index images. Further details at: https://docs.google.com/document/d/1CxabarOVMmjyl4zyE1_8z9Oq8JPne6MKGuI3NnPbYb0/edit#

**Reviewer Checklist**


test/e2e/skopeo.go
```diff
@@ -13,17 +13,17 @@ import (
...
 	insecure              = "--insecure-policy=true"
-	skopeo                = "skopeo"
+	Skopeo                = "Skopeo"
 	debug                 = "--debug"
 	skipTLS               = "--dest-tls-verify=false"
 	skipCreds             = "--dest-no-creds=true"
 	destCreds             = "--dest-creds="
 	v2format              = "--format=v2s2"
-	skopeoImage           = "quay.io/olmtest/skopeo:0.1.40"
+	skopeoImage           = "quay.io/olmtest/Skopeo:0.1.40"
```
exdx, line 22: do we want to change the name of the skopeo image tag? 

In general, I can see the motivation of exporting all this code out of the `e2e` package so that the `utils` package you wrote can access it. I think its better if we simply moved all this existing skopeo/registry code out of the `e2e` package into `utils` and export from `utils` into `e2e` only whats necessary. 

The e2e package is way too big already :)

test/e2e/skopeo.go
```diff
@@ -99,7 +99,7 @@ func deleteSkopeoPod(client operatorclient.ClientInterface, namespace string) er
 func skopeoLocalCopy(newImage, newTag string, oldImage, oldTag string) (string, error) {
 	newImageName := fmt.Sprint(newImage, newTag)
 	oldImageName := fmt.Sprint(oldImage, oldTag)
-	cmd := exec.Command(skopeo, debug, insecure, "copy", skipTLS, v2format, skipCreds, oldImageName, newImageName)
+	cmd := exec.Command(Skopeo, debug, insecure, "copy", skipTLS, v2format, skipCreds, oldImageName, newImageName)
```
exdx, line 102: I don't think this exec will work if we are calling skopeo `Skopeo`

test/e2e/utils/bundles.go
```diff
@@ -0,0 +1,164 @@
...
+	"github.com/containers/storage/pkg/unshare"
+	"os"
+	"os/exec"
+	"path"
+	"sort"
+	"strings"
+	"time"
+)
+
+type Bundle struct {
```
exdx, line 20: nit: I think it's potentially fine to define a one-off bundle type for our purposes here, but we can also use the `api.Bundle` type that is already used widely within OLM to define a bundle. If that works then I recommend we use that

test/e2e/utils/bundles.go
```diff
@@ -0,0 +1,164 @@
...
+		sort.Strings(b.Channels)
+		b.DefaultChannel = b.Channels[0]
+	}
+	bundlePath := b.BundlePath
+	if len(bundlePath) == 0 {
+		bundlePath = fmt.Sprintf("%s/operator", b.PackageName)
+	}
+	bundleReference = fmt.Sprintf("%s/%s:%s", r.URL, bundlePath, b.Version)
+
+	cmd := exec.Command("opm", "alpha", "bundle", "build", "-d", b.BundleManifestDirectory, "-b", "docker", "-e", b.DefaultChannel, "-c", strings.Join(b.Channels, ","), "--tag", bundleReference)
```
exdx, line 48: I don't think docker is available as a container-tool within the e2e test-execution context. 

The container runtime on openshift is CRI-O -- I think we could maybe include podman in the e2e image and use that as the container-tool if we go down this route

Did we consider trying daemonless builds via containerd (ie using `none` as the container-tool)?

test/e2e/utils/bundles.go
```diff
@@ -0,0 +1,164 @@
...
+	_, _, _, err = b.Commit(context.TODO(), dest, buildah.CommitOptions{
+		ReportWriter:          os.Stderr,
+		SystemContext: &types.SystemContext{
+		},
+		OmitTimestamp:         true,
+	})
+	if err != nil {
+		return "", fmt.Errorf("error creating image from container %s: %v", containerName, err)
+	}
+
```
exdx, line 160: would potentially be cool to validate the image via `opm validate ...` somewhere here as well

test/e2e/utils/imageregistry.go
```diff
@@ -0,0 +1,181 @@
...
+		e2e.DeleteDockerRegistry(c, testNamespace)
+	}
+	return nil
+}
+
+
+// podCheckFunc describes a function that returns true if the given Pod meets some criteria; false otherwise.
+type podCheckFunc func(pod *corev1.Pod) bool
+
+func awaitPod(c operatorclient.ClientInterface, namespace, name string, checkPod podCheckFunc) (*corev1.Pod, error) {
```
exdx, line 88: nit: I think it's best we don't duplicate this function in the the `e2e` and `utils` package. Let's pull it out into the `utils` package in a separate file and access it that way, along with the various `podCheckFunc`s

test/e2e/utils/imageregistry.go
```diff
@@ -0,0 +1,181 @@
...
+		pod = p
+		return checkPod(pod), nil
+	})
+	if err != nil {
+		return nil, err
+	}
+	return pod, nil
+}
+
+func Local(client operatorclient.ClientInterface) (bool, error) {
```
exdx, line 104: this is not directly relevant to the imageregistry -- could we have it live in another file?

test/e2e/utils/imageregistry.go
```diff
@@ -0,0 +1,181 @@
...
+				return "", err
+			}
+			if err := r.skopeoCopy(fmt.Sprintf("%s/%s:", r.URL, b.BundlePath), b.Version, imageRef, ""); err != nil {
+				return "", err
+			}
+			bundleRefs = append(bundleRefs, imageRef)
+		}
+	}
+
+	indexReference, err := r.CreateAndUploadIndex(indexName, bundleRefs)
```
exdx, line 176: one gotcha here is that we need to add the built bundle images to the index in the correct order - if one bundle is after a prior bundle, but the CSV has a `replaces` reference to a CSV that is not yet in the index - opm will error out.

I think we did some work to address this (the ability to add multiple bundles, not necessarily in the right order), but not sure if it's still an issue. Locally I ran into this issue recently. @kevinrizza this is still an issue right? 

tl;dr we should be sure we are adding the bundles in the correct order

test/e2e/utils/labels.go
```diff
@@ -0,0 +1,28 @@
+package utils
+
+import "strings"
+
+const (
```
exdx, line 5: is there a way to import these annotations from the `api` library instead of hardcoding them here?

Bowenislandsong:
<details>

 totaltestcount: 1
flaketestcount: 1
skippedtestcount: 4
flaketests:
  - classname: End-to-end
    name: '**CSV create requirements met CRD**'
    counts: 1
    details:
      - count: 1
        error: |4-


            /home/runner/work/operator-lifecycle-manager/operator-lifecycle-manager/test/e2e/csv_e2e_test.go:458

            	Error Trace:	csv_e2e_test.go:700
            	            				runner.go:113
            	            				runner.go:64
            	            				it_node.go:26
            	            				spec.go:215
            	            				spec.go:138
            	            				spec_runner.go:200
            	            				spec_runner.go:170
            	            				spec_runner.go:66
            	            				suite.go:62
            	            				ginkgo_dsl.go:226
            	            				ginkgo_dsl.go:214
            	            				e2e_test.go:54

    meandurationsec: 332.241255
skippedtests:
  - classname: End-to-end
    counts: 1
    details: []
    meandurationsec: 0.020766
  - classname: End-to-end
        prometheus objects**'
    counts: 1
    details: []
    meandurationsec: 2.213104
  - classname: End-to-end
    counts: 1
    details: []
    meandurationsec: 1.659119
  - classname: End-to-end
        successfully**'
    counts: 1
    details: []
    meandurationsec: 2.175028

test/e2e/skopeo.go
ankitathomas: This was a refactoring error, got that fixed. Looks like buildah does a great job of pushing bundles on its own, so we won't need this.

test/e2e/utils/bundles.go
ankitathomas: api.Bundle is missing a few fields that we need, mainly the default channel and the upload url. We can generate a url from the bundle name and a random string but default channel information is bundle specific.

test/e2e/utils/bundles.go
ankitathomas: The reason for using docker here was just to allow it to run locally, opm wasn't passing along the skipTLS option while pulling index images. Docker was the only container-tool enabling insecure pull by default. We can look at where the test is running before choosing a pull tool.

test/e2e/utils/imageregistry.go
ankitathomas: I've moved it over to utils.go

test/e2e/utils/imageregistry.go
ankitathomas: We're using the multiple bundles in one go approach here, each time it creates a new index. I had issues with bundle order when adding one at a time, and this seemed quicker than a build-upload loop, so switched to it.

test/e2e/utils/labels.go
ankitathomas: I'm switching over to importing the annotation labels

test/e2e/utils/imageregistry.go
ankitathomas: I've moved the common functions to a separate location in the utils package

test/e2e/utils/bundles.go
ankitathomas: Added validation support

base.Dockerfile
```diff
@@ -3,7 +3,7 @@
 FROM openshift/origin-release:golang-1.13
 
 # Install test dependencies
-RUN yum install -y skopeo && \
+RUN yum install -y skopeo gpgme-devel && \
     export OS=$(go env GOOS) && \
```
ecordell, line 7: I think we discovered that this change would need to go into master before the changes take effect, fyi

PR #1669, opened 2020-07-22
Bug 1859178: fix(installplans): GC older installplans

ecordell:
This adds a GC check in the installplan controller.

If it finds more than 5 installplans for a namespace, it will delete the oldest ones. Age is determined first by `Generation` and second via `CreationTimestamp` (to account for any bugs that result in installplans with the same generation, which should never happen).

It was possible in the current resolver to construct a catalog that would cause installplans to be created endlessly. This change protects against potential problems like this.

The new resolver will begin to output installplans even on failed resolution attempts, so this limit will support that feature as well.

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
+	"sort"
 	"strings"
 	"sync"
 	"time"
@@ -22,6 +23,7 @@ import (
 	extinf "k8s.io/apiextensions-apiserver/pkg/client/informers/externalversions"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	utilclock "k8s.io/apimachinery/pkg/util/clock"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
@@ -67,6 +69,8 @@ const (
 	roleKind               = "Role"
 	roleBindingKind        = "RoleBinding"
 	generatedByKey         = "olm.generated-by"
+	maxInstallPlanCount    = 5
+	maxDeletesPerSweep     = 5
 )
 
 // Operator represents a Kubernetes operator that executes InstallPlans by
@@ -777,6 +781,8 @@ func (o *Operator) syncResolvingNamespace(obj interface{}) error {
 		"id":        queueinformer.NewLoopID(),
 	})
 
+	o.gcInstallPlans(logger, namespace)
+
 	// get the set of sources that should be used for resolution and best-effort get their connections working
 	logger.Debug("resolving sources")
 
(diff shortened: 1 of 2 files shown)


pkg/controller/operators/catalog/operator.go
```diff
@@ -1180,6 +1183,64 @@ func (o *Operator) unpackBundles(plan *v1alpha1.InstallPlan) (bool, *v1alpha1.In
 	return unpacked, out, nil
 }
 
+func (o *Operator) gcInstallPlans(log logrus.FieldLogger, namespace string) {
```
anik120, line 1186: What does `gc` stand for?

pkg/controller/operators/catalog/operator.go
```diff
@@ -1180,6 +1183,64 @@ func (o *Operator) unpackBundles(plan *v1alpha1.InstallPlan) (bool, *v1alpha1.In
...
+			return g[i].CreationTimestamp.Before(&g[j].CreationTimestamp)
+		})
+		toDelete = append(toDelete, g[:len(ips)-len(toDelete)-maxInstallPlanCount]...)
+	}
+
+	for _, i := range toDelete {
+		if err := o.client.OperatorsV1alpha1().InstallPlans(namespace).Delete(context.TODO(), i.GetName(), metav1.DeleteOptions{}); err != nil {
+			log.WithField("deleting", i.GetName()).WithError(err).Warn("error GCing old installplan")
+		}
+	}
```
njhale, line 1241: Should we aggregate and return transient errors we care about so this can be requeued?

pkg/controller/operators/catalog/operator.go
```diff
@@ -1196,6 +1257,8 @@ func (o *Operator) syncInstallPlans(obj interface{}) (syncError error) {
 
 	logger.Info("syncing")
 
+	o.gcInstallPlans(logger, plan.GetNamespace())
```
njhale, line 1260: I think performing this GC on namespace resolve sync would result in less overall churn. Is there a reason to perform this for every InstallPlan event?

njhale, line 1260: I think we want to bail out if this InstallPlan isn't the latest generation -- or has been deleted -- right?

pkg/controller/operators/catalog/operator.go
njhale: garbage collect

anik120: Could we document that somehow in the code?

pkg/controller/operators/catalog/operator.go
exdx: What happens when this fix goes out to clusters with 1000s of install plans? They will chug along and delete a lot of install plans all at once? Do we want to implement a semaphore or some kind of rate limiting in this deletion step?

ecordell: I wanted to avoid doing that so that problems with GC don't block otherwise normal installplan processing.

As it stands, I'm a little uncomfortable with this implementation, which stops installplan processing to go GC installplans. On clusters that have hit the installplan bug, this could potentially take a long time. 

I will build this controller into an image, trigger the bug, and test it out. It might be that this needs to be in its own controller entirely to be safe.

pkg/controller/operators/catalog/operator.go
ecordell: That's already handled by the rest of the installplan loop

ecordell:
/hold

I want to perform some additional testing and consider comments here before this merges

pkg/controller/operators/catalog/operator.go
```diff
@@ -1180,6 +1183,64 @@ func (o *Operator) unpackBundles(plan *v1alpha1.InstallPlan) (bool, *v1alpha1.In
 	return unpacked, out, nil
 }
 
+func (o *Operator) gcInstallPlans(log logrus.FieldLogger, namespace string) {
+	ips, err := o.lister.OperatorsV1alpha1().InstallPlanLister().InstallPlans(namespace).List(labels.Everything())
```
exdx, line 1187: should we ensure that we are getting installplans that are in some kind of terminal state only? we could end up listing and deleting installplans that are actively being resolved on-cluster

ecordell: this happens per-namespace, and we shouldn't be creating more than one "active" installplan at a time

pkg/controller/operators/catalog/operator.go
ecordell: > They will chug along and delete a lot of install plans all at once?

OLM will issue a huge set of `delete` requests, yes.

>  Do we want to implement a semaphore or some kind of rate limiting in this deletion step?

This would make me feel better about having the gc run as part of an otherwise "real" control loop. I will look into options here (would be nice to page through the cache) and also consider separating GC entirely.

pkg/controller/operators/catalog/operator.go
ecordell: I reasoned that this hedges against issues with catalogs / the resolver pinning the namespace sync for too long. 

But lots of good ideas here + in other comments, I'll update this PR.

pkg/controller/operators/catalog/operator.go
exdx: oh so if we make two subscriptions to two different operators in the same catalog (all in the same namespace) OLM will sequentially resolve the installplans (only one being active at a time?) good to know. i thought maybe OLM worked on them incrementally

njhale: @exdx only the latest InstallPlan counts, and that will contain the resolution for _all_ the Subscriptions in the namespace.

pkg/controller/operators/catalog/operator.go
```diff
@@ -1180,6 +1186,73 @@ func (o *Operator) unpackBundles(plan *v1alpha1.InstallPlan) (bool, *v1alpha1.In
...
+	}
+
+	if len(allIps) <= maxInstallPlanCount {
+		return
+	}
+
+	// we only consider maxDeletesPerSweep more than the allowed number of installplans for delete at one time
+	ips := allIps
+	if len(ips) > maxInstallPlanCount + maxDeletesPerSweep {
+		ips = allIps[:maxInstallPlanCount+maxDeletesPerSweep]
```
exdx, line 1205: so here we limit the number of extra installplans removed to 8 at a time? Makes sense, but for these clusters with 1000s of installplans they could be waiting a while. I was thinking of a more async process that goes through and deletes these extra ips, eventually reporting status back to OLM. Its definitely more work though and this is an urgent bugfix so idk

pkg/controller/operators/catalog/operator.go
```diff
@@ -1180,6 +1186,73 @@ func (o *Operator) unpackBundles(plan *v1alpha1.InstallPlan) (bool, *v1alpha1.In
...
+
+		// sort by creation time
+		sort.Slice(g, func(i, j int) bool {
+			return g[i].CreationTimestamp.Before(&g[j].CreationTimestamp)
+		})
+		toDelete = append(toDelete, g[:len(ips)-len(toDelete)-maxInstallPlanCount]...)
+	}
+
+	for _, i := range toDelete {
+		if err := o.client.OperatorsV1alpha1().InstallPlans(namespace).Delete(context.TODO(), i.GetName(), metav1.DeleteOptions{}); err != nil {
```
exdx, line 1250: I'm thinking we want to set the `DeleteOption` here?

exdx: If the `InstallPlan` has no dependents then an `Orphan` PropagationPolicy be more efficient?

ecordell:
/hold cancel

tested in a 4.4 cluster with lots of installplans generated, and this kept them down to the five latest.

pkg/controller/operators/catalog/operator.go
```diff
@@ -1180,6 +1186,77 @@ func (o *Operator) unpackBundles(plan *v1alpha1.InstallPlan) (bool, *v1alpha1.In
...
+	}
+
+	if len(allIps) <= maxInstallPlanCount {
+		return
+	}
+
+	// we only consider maxDeletesPerSweep more than the allowed number of installplans for delete at one time
+	ips := allIps
+	if len(ips) > maxInstallPlanCount + maxDeletesPerSweep {
+		ips = allIps[:maxInstallPlanCount+maxDeletesPerSweep]
```
dinhxuanvu, line 1205: I don't fully understand the slicing with indices here. Is this array ordered/sorted? Otherwise, why just randomly pick a bunch of InstallPlans from the first 10 in the array?

njhale, line 1205: This depends on the lister returning `InstallPlans` in descending `createdAt` order, which [doesn't look like the case](https://github.com/operator-framework/operator-lifecycle-manager/blob/cc2112c70498f7a08c84c8886b5c360657568de1/vendor/k8s.io/client-go/tools/cache/thread_safe_store.go#L109).

dinhxuanvu: Also, there are a few clusters with InstallPlans in order of 1000s. I know one of them has at least 1200. Given the sync period of 15 mins (though in reality the namespace sync happens more often due to resources changes) and the limit of 5 IPs per sweep, would this take a bit too much time to clean up especially in the case of idle cluster? In fact, I would prefer to keep the InstallPlans that are currently referenced in the Subscriptions and clean up the others in a higher rate. Maybe this can be changed in the future instead of now due to the urgent need of getting this in.

njhale: forgot this is being run multiple times; i.e. successive sweeps will clean these up.

njhale: IIUC, for each sweep we take up to `maxDeletesPerSweep` InstallPlans to delete, and leave the rest for the next sweep. We're essentially only sorting a chunk of the list on each sweep, and after X sweeps we'll have sorted the whole thing.

PR #1670, opened 2020-07-22
wip(operators): watch preferred kinds dynamically

njhale:

- Allow associating resources of **all** preferred GVKs in a cluster with an `Operator` resource.
- Enable automatic adoption of components for **all** preferred GVKs

This means that **any** resource on the cluster can be labelled as an `Operator` component.

The set of possible resource types that can comprise an operator changes dynamically as a cluster's control-plane is extended; w/ CRDs and extension APIServices. Often, an operator author considers certain custom resources of kinds they provide to be a component of their operator deployment ( e.g. a config CR).

### DIFF
--- a/pkg/controller/operators/adoption_controller.go
+++ b/pkg/controller/operators/adoption_controller.go
@@ -6,18 +6,15 @@ import (
 	"sync"
 
 	"github.com/go-logr/logr"
-	appsv1 "k8s.io/api/apps/v1"
-	corev1 "k8s.io/api/core/v1"
-	rbacv1 "k8s.io/api/rbac/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
-	apiregistrationv1 "k8s.io/kube-aggregator/pkg/apis/apiregistration/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
@@ -27,6 +24,7 @@ import (
 	operatorsv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 	operatorsv2alpha1 "github.com/operator-framework/api/pkg/operators/v2alpha1"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/operators/decorators"
+	libsource "github.com/operator-framework/operator-lifecycle-manager/pkg/lib/controller-runtime/source"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/ownerutil"
 )
 
@@ -37,6 +35,7 @@ type AdoptionReconciler struct {
 	log     logr.Logger
 	mu      sync.RWMutex
 	factory decorators.OperatorFactory
+	source  *libsource.Dynamic
 }
 
 // +kubebuilder:rbac:groups=operators.coreos.com,resources=operators,verbs=create;update;patch;delete
@@ -64,20 +63,7 @@ func (r *AdoptionReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	}
 	err = ctrl.NewControllerManagedBy(mgr).
 		For(&operatorsv1alpha1.ClusterServiceVersion{}).
-		Watches(&source.Kind{Type: &appsv1.Deployment{}}, enqueueCSV).
-		Watches(&source.Kind{Type: &corev1.Namespace{}}, enqueueCSV).
-		Watches(&source.Kind{Type: &corev1.Service{}}, enqueueCSV).
-		Watches(&source.Kind{Type: &corev1.ServiceAccount{}}, enqueueCSV).
-		Watches(&source.Kind{Type: &corev1.Secret{}}, enqueueCSV).
-		Watches(&source.Kind{Type: &corev1.ConfigMap{}}, enqueueCSV).
-		Watches(&source.Kind{Type: &rbacv1.Role{}}, enqueueCSV).
-		Watches(&source.Kind{Type: &rbacv1.RoleBinding{}}, enqueueCSV).
-		Watches(&source.Kind{Type: &rbacv1.ClusterRole{}}, enqueueCSV).
-		Watches(&source.Kind{Type: &rbacv1.ClusterRoleBinding{}}, enqueueCSV).
-		Watches(&source.Kind{Type: &apiextensionsv1.CustomResourceDefinition{}}, enqueueCSV).
-		Watches(&source.Kind{Type: &apiregistrationv1.APIService{}}, enqueueCSV).
-		Watches(&source.Kind{Type: &operatorsv1alpha1.Subscription{}}, enqueueCSV).
-		Watches(&source.Kind{Type: &operatorsv1alpha1.InstallPlan{}}, enqueueCSV).
+		Watches(r.source, enqueueCSV).
 		Complete(reconcile.Func(r.ReconcileClusterServiceVersion))
 	if err != nil {
 		return err
@@ -104,6 +90,7 @@ func NewAdoptionReconciler(cli client.Client, log logr.Logger, scheme *runtime.S
 
 		log:     log,
 		factory: factory,
+		source:  &libsource.Dynamic{},
 	}, nil
 }
 
@@ -238,6 +225,8 @@ func (r *AdoptionReconciler) adoptComponents(ctx context.Context, csv *operators
 		return nil
 	}
 
+	r.log.Info("adopting for", "operator count", len(operators))
+
 	// Label (adopt) prospective components
 	var errs []error
 	// TODO(njhale): parallelize
@@ -247,6 +236,7 @@ func (r *AdoptionReconciler) adoptComponents(ctx context.Context, csv *operators
 			errs = append(errs, err)
 			continue
 		}
+		r.log.Info("retrieved components", "component count", len(components))
 
 		for _, component := range components {
 			candidate := component.DeepCopyObject()
@@ -263,6 +253,7 @@ func (r *AdoptionReconciler) adoptComponents(ctx context.Context, csv *operators
 				continue
 			}
 
+			r.log.Info("adopted", "component labels", component.(metav1.Object).GetLabels())
 			// Patch the component to adopt
 			if err = r.Patch(ctx, component, client.MergeFrom(candidate)); err != nil {
 				errs = append(errs, err)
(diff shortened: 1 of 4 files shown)


pkg/controller/operators/adoption_controller.go
```diff
@@ -64,20 +63,7 @@ func (r *AdoptionReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
-		Watches(&source.Kind{Type: &corev1.ConfigMap{}}, enqueueCSV).
-		Watches(&source.Kind{Type: &rbacv1.Role{}}, enqueueCSV).
-		Watches(&source.Kind{Type: &rbacv1.RoleBinding{}}, enqueueCSV).
-		Watches(&source.Kind{Type: &rbacv1.ClusterRole{}}, enqueueCSV).
-		Watches(&source.Kind{Type: &rbacv1.ClusterRoleBinding{}}, enqueueCSV).
-		Watches(&source.Kind{Type: &apiextensionsv1.CustomResourceDefinition{}}, enqueueCSV).
-		Watches(&source.Kind{Type: &apiregistrationv1.APIService{}}, enqueueCSV).
-		Watches(&source.Kind{Type: &operatorsv1alpha1.Subscription{}}, enqueueCSV).
-		Watches(&source.Kind{Type: &operatorsv1alpha1.InstallPlan{}}, enqueueCSV).
+		Watches(r.source, enqueueCSV).
```
njhale, line 66: I think we should see if we can share caches/watches between controllers under a single manager. I don't like the idea watching and storing `k*|cluster resources|`.

pkg/controller/operators/adoption_controller.go
```diff
@@ -274,24 +265,17 @@ func (r *AdoptionReconciler) adoptComponents(ctx context.Context, csv *operators
...
+	if err != nil {
+		return nil, err
+	}
+
+	var componentLists []runtime.Object
+	for _, gvk := range informable {
+		gvk.Kind = gvk.Kind + "List"
+		ul := &unstructured.UnstructuredList{}
+		ul.SetGroupVersionKind(gvk)
+		componentLists = append(componentLists, ul)
```
njhale, line 278: Needs to be deduplicated across the operator and adoption controller. This is probably generic enough to be it's own layer on top of the controller-runtime client.

pkg/controller/operators/decorators/operator.go
```diff
@@ -218,6 +218,7 @@ func (o *Operator) AdoptComponent(component runtime.Object) (adopted bool, err e
 		labels[labelKey] = ""
 		adopted = true
 	}
+	m.SetLabels(labels)
```
njhale, line 221: Makes the changes stick for `unstructured.Unstructured`.

pkg/lib/controller-runtime/source/dynamic.go
```diff
@@ -0,0 +1,196 @@
...
+	return nil
+}
+
+var (
+	knownInformables []schema.GroupVersionKind
+	informablesLock  sync.RWMutex
+)
+
+// InformableGVKs returns the set of GVKs available in discovery that support the list and watch verbs (informable).
+func (d *Dynamic) InformableGVKs() (informables []schema.GroupVersionKind, err error) {
```
njhale, line 95: Going to factor this out of `Dynamic`.

pkg/lib/controller-runtime/source/dynamic.go
```diff
@@ -0,0 +1,196 @@
...
+	if d.client == nil {
+		return nil, fmt.Errorf("no client")
+	}
+
+	var resourceLists []*metav1.APIResourceList
+	resourceLists, err = d.client.ServerPreferredResources()
```
njhale, line 101: Fails when discovery breaks for **any** resource.

pkg/lib/controller-runtime/source/dynamic.go
```diff
@@ -0,0 +1,196 @@
...
+	if err != nil {
+		informablesLock.RLock()
+		defer informablesLock.RUnlock()
+		log.Error(err, "could not get preferred resources, returning latest known gvks")
+		return knownInformables, nil
```
njhale, line 106: Going to memoize the last successful `APIResourceList` and whittle it down with direct checks against the discovery client instead. This will let us weed out the problem resource types before returning.

pkg/lib/controller-runtime/source/dynamic.go
```diff
@@ -0,0 +1,196 @@
...
+				log.V(4).Info("dynamic source already started", "gvk", gvk)
+				continue
+			}
+
+			if err := d.startSource(gvk, handle, queue, predicates...); err != nil {
+				log.Error(err, "failed to start dynamic source", "gvk", gvk)
+			}
+		}
+
+	}, 15*time.Second, d.stop)
```
njhale, line 82: What do we think about triggering this when APIServices and CRDs change?

njhale, line 82: Make this configurable?

pkg/lib/controller-runtime/source/dynamic.go
```diff
@@ -0,0 +1,196 @@
...
+
+	synced.Wait()
```
njhale, line 84: Block until we set up the initial set of sources.

pkg/lib/controller-runtime/source/dynamic.go
```diff
@@ -0,0 +1,196 @@
...
+	defer d.mu.Unlock()
+
+	if d.sources == nil {
+		d.sources = map[schema.GroupVersionKind]source.Source{}
+	}
+
+	r := &unstructured.Unstructured{}
+	r.SetGroupVersionKind(gvk)
+	s := &source.Kind{Type: r}
+	if err = s.InjectCache(d.cache); err != nil {
```
njhale, line 144: controller-runtime doesn't expose a way to stop individual sources/informers, see:

- https://github.com/kubernetes-sigs/controller-runtime/issues/935
- https://github.com/kubernetes-sigs/controller-runtime/pull/936
- https://github.com/gardener/gardener-resource-manager/issues/61

I'm curious to see if anyone has suggestions for how to get around this short of reimplementing controller-runtime's [informerCache](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/cache/informer_cache.go).

Bowenislandsong:
<details>

 totaltestcount: 1
flaketestcount: 1
skippedtestcount: 4
flaketests:
  - classname: End-to-end
    name: '**Subscription skip range**'
    counts: 1
    details:
      - count: 1
        error: |4-


            /home/runner/work/operator-lifecycle-manager/operator-lifecycle-manager/test/e2e/subscription_e2e_test.go:92

            	Error Trace:	subscription_e2e_test.go:164
            	            				runner.go:113
            	            				runner.go:64
            	            				it_node.go:26
            	            				spec.go:215
            	            				spec.go:138
            	            				spec_runner.go:200
            	            				spec_runner.go:170
            	            				spec_runner.go:66
            	            				suite.go:62
            	            				ginkgo_dsl.go:226
            	            				ginkgo_dsl.go:214
            	            				e2e_test.go:54
            	            	etcdserver: request timed out

    meandurationsec: 28.239263
skippedtests:
  - classname: End-to-end
    counts: 1
    details: []
    meandurationsec: 0.056757
  - classname: End-to-end
    counts: 1
    details: []
    meandurationsec: 0.307005
  - classname: End-to-end
        prometheus objects**'
    counts: 1
    details: []
    meandurationsec: 2.184881
  - classname: End-to-end
        successfully**'
    counts: 1
    details: []
    meandurationsec: 3.170511

pkg/lib/controller-runtime/source/dynamic.go
anik120: `getInformableGVKs` might make the code more readable at the site this function is called. 

https://dmitripavlutin.com/coding-like-shakespeare-practical-function-naming-conventions/#2-prefer-explanatory-names

pkg/lib/controller-runtime/source/dynamic.go
```diff
@@ -0,0 +1,196 @@
...
+func (d *Dynamic) InjectStopChannel(stop <-chan struct{}) error {
+	if d.stop == nil {
+		d.stop = stop
+	}
+
+	return nil
+}
+
+// InjectConfig injects the the REST Config dependency.
+func (d *Dynamic) InjectConfig(config *rest.Config) (err error) {
```
anik120, line 188: Is `InjectRESTConfigDependency` too big of a name? It'll would get rid of the comment, and be self explanatory in the calling site too.

njhale: I don't disagree, but we don't control the [inject method signatures](https://github.com/kubernetes-sigs/controller-runtime/blob/49e1044b3b6ac6b3064cd348604e5b28e9d5c786/pkg/runtime/inject/inject.go#L61).

njhale: > It'll would get rid of the comment

Also, golangci-lint generally complains about exports w/o comments.

pkg/lib/controller-runtime/source/dynamic.go
```diff
@@ -0,0 +1,196 @@
...
+
+	// TODO(njhale): Stop sources when GVKs are removed
+	var (
+		synced sync.WaitGroup
+		once   sync.Once
+	)
+	synced.Add(1)
+
+	go wait.Until(func() {
+		defer once.Do(func() {
```
exdx, line 62: The intention here is to only call `synced.Done()` once after the function exits - but since the function is called within the `Until()` scope every 15 seconds are you sure this doesn't just get triggered the first time the function is executed within the `Until` loop? 

This is probably standard code, just want to confirm.

pkg/lib/controller-runtime/source/dynamic.go
exdx: Makes sense. Maybe we can expand the `InformableGVKs()` function to not only memoize on known GVKs in case of discovery errors but also use that as a cache so that when new CRDs (GVKs) become available we can use the diff instead of going through the `ServerPreferredResources` again

pkg/lib/controller-runtime/source/dynamic.go
exdx: It's actually quite a useful function in a standalone way

pkg/lib/controller-runtime/source/dynamic.go
```diff
@@ -0,0 +1,196 @@
...
+		}
+
+		for _, r := range rl.APIResources {
+			gvk := schema.GroupVersionKind{Group: gv.Group, Version: gv.Version, Kind: r.Kind}
+			informables = append(informables, gvk)
+		}
+	}
+
+	if len(informables) > 0 {
+		// Memoize the latest informables in case discovery fails on successive calls
```
exdx, line 124: I like this touch a lot. May be interesting to play around with an e2e test that fails discovery calls intentionally and sees how the operator API behaves

pkg/controller/operators/operator_controller.go
```diff
@@ -168,23 +152,17 @@ func (r *OperatorReconciler) updateComponents(ctx context.Context, operator *dec
...
-		&rbacv1.RoleList{},
-		&rbacv1.RoleBindingList{},
-		&rbacv1.ClusterRoleList{},
-		&rbacv1.ClusterRoleBindingList{},
-		&apiextensionsv1.CustomResourceDefinitionList{},
-		&apiregistrationv1.APIServiceList{},
-		&operatorsv1alpha1.SubscriptionList{},
-		&operatorsv1alpha1.InstallPlanList{},
-		&operatorsv1alpha1.ClusterServiceVersionList{},
+	informable, err := r.source.InformableGVKs()
```
exdx, line 155: nit: this may take a while on an Openshift cluster with a ton of GVKs...maybe adding a log message indicating that this process is happening could be insightful

exdx:
I think a lot of this work will be useful in the "OLM supports arbitrary kube types" feature upcoming in the next release. In fact if there's anything here that we could expose to that effort that would be super helpful.

pkg/lib/controller-runtime/source/dynamic.go
exdx: > `getInformableGVKs` might make the code more readable at the site this function is called.
> 
> https://dmitripavlutin.com/coding-like-shakespeare-practical-function-naming-conventions/#2-prefer-explanatory-names

I actually disagree, I think it's Go code convention not to have things like `Get...` in function names, unlike say JavaScript
https://golang.org/doc/effective_go.html#Getters

exdx: It's a pet peeve of mine 😄

pkg/lib/controller-runtime/source/dynamic.go
njhale: > The intention here is to only call synced.Done() once after the function exits - but since the function is called within the Until() scope every 15 seconds are you sure this doesn't just get triggered the first time the function is executed within the Until loop?

I'm not understanding the delineation you're making here. Could you rephrase?

pkg/lib/controller-runtime/source/dynamic.go
anik120: Alas.

pkg/lib/controller-runtime/source/dynamic.go
anik120: Okay fair point. I concede.

njhale:
Closing this out to pursue an alternative: for each API group, whenever the default version set changes, spin up a new pod to watch the resources of that set.

PR #1673, opened 2020-07-22
feat(resolver): filtering deprecated bundles in resolver

gallettilance:

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/registry/resolver/cache.go
+++ b/pkg/controller/registry/resolver/cache.go
@@ -416,6 +416,17 @@ func WithPackage(pkg string) OperatorPredicate {
 	}
 }
 
+func WithoutDeprecatedProperty() OperatorPredicate {
+	return func(o *Operator) bool {
+		for _, p := range o.bundle.GetProperties() {
+			if p.GetType() == string(opregistry.DeprecatedType) {
+				return false
+			}
+		}
+		return true
+	}
+}
+
 func WithVersionInRange(r semver.Range) OperatorPredicate {
 	return func(o *Operator) bool {
 		return o.version != nil && r(*o.version)
--- a/pkg/controller/registry/resolver/resolver.go
+++ b/pkg/controller/registry/resolver/resolver.go
@@ -73,6 +73,8 @@ func (r *SatResolver) SolveOperators(namespaces []string, csvs []*v1alpha1.Clust
 			predicates = append(predicates, WithChannel(sub.Spec.Channel))
 		}
 
+		predicates = append(predicates, WithoutDeprecatedProperty())
+
 		// find the currently installed operator (if it exists)
 		var current *Operator
 		for _, csv := range csvs {
@@ -297,6 +299,7 @@ func (r *SatResolver) getBundleInstallables(catalog registry.CatalogKey, predica
 			errs = append(errs, err)
 			continue
 		}
+		dependencyPredicates = append(dependencyPredicates, WithoutDeprecatedProperty())
 		for _, d := range dependencyPredicates {
 			// errors ignored; this will build an empty/unsatisfiable dependency if no candidates are found
 			candidateBundles, _ := AtLeast(1, namespacedCache.FindPreferred(&bundle.sourceInfo.Catalog, d))
--- a/pkg/controller/registry/resolver/util_test.go
+++ b/pkg/controller/registry/resolver/util_test.go
@@ -73,7 +73,7 @@ func NewFakeOperatorSurface(name, pkg, channel, replaces, src, startingCSV strin
 	b := bundle(name, pkg, channel, replaces, providedCRDAPISet, requiredCRDAPISet, providedAPIServiceAPISet, requiredAPIServiceAPISet)
 
 	deps := apiSetToDependencies(requiredCRDAPISet, requiredAPIServiceAPISet)
-	props := apiSetToProperties(requiredCRDAPISet, requiredAPIServiceAPISet)
+	props := apiSetToProperties(requiredCRDAPISet, requiredAPIServiceAPISet, false)
 	deps = append(deps, dependencies...)
 
 	return &Operator{
@@ -277,7 +277,7 @@ func apiSetToDependencies(crds, apis APISet) (out []*api.Dependency) {
 	return
 }
 
-func apiSetToProperties(crds, apis APISet) (out []*api.Property) {
+func apiSetToProperties(crds, apis APISet, deprecated bool) (out []*api.Property) {
 	out = make([]*api.Property, 0)
 	for a := range crds {
 		val, err := json.Marshal(opregistry.GVKProperty{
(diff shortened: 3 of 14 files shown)


gallettilance:
waiting on https://github.com/operator-framework/operator-registry/pull/397

pkg/controller/registry/resolver/operators.go
```diff
@@ -442,9 +440,24 @@ func PredicateForDependency(dependency *api.Dependency) (OperatorPredicate, erro
...
+			predicates = append(predicates, predicateForDeprecated(prop))
+		}
+	}
+	return
+}
+
+func predicateForDeprecated(prop *api.Property) OperatorPredicate {
+	return WithoutDeprecatedProperty()
+}
+
```
ecordell, line 456: You should be able to remove these, right? 

I don't see them used, and we don't need to build predicates dynamically based on the content of properties (OLM will just include the `WithoutDeprecated` predicate when filtering.

pkg/controller/registry/resolver/operators.go
```diff
@@ -442,9 +440,24 @@ func PredicateForDependency(dependency *api.Dependency) (OperatorPredicate, erro
...
 var predicates = map[string]func(string) (OperatorPredicate, error){
-	opregistry.GVKType:     predicateForGVKDependency,
-	opregistry.PackageType: predicateForPackageDependency,
+	opregistry.GVKType:        predicateForGVKDependency,
+	opregistry.PackageType:    predicateForPackageDependency,
+	opregistry.DeprecatedType: predicateForDeprecated,
```
ecordell, line 460: This can go as well, I think

pkg/controller/registry/resolver/resolver.go
```diff
@@ -73,6 +73,8 @@ func (r *SatResolver) SolveOperators(namespaces []string, csvs []*v1alpha1.Clust
 			predicates = append(predicates, WithChannel(sub.Spec.Channel))
 		}
 
+		predicates = append(predicates, WithoutDeprecatedProperty())
```
ecordell, line 76: this applies for finding updates / new installs, but we also need this predicate when searching for dependency candidates - see line 302 in getBundleInstallables

dinhxuanvu:
Looking this is missing some `go mod vendor` action because I don't see any new changes from registry being pulled into this PR.

dinhxuanvu:
This PR can be merged when it is lgtm and approve. I remove the hold.
/hold cancel

pkg/controller/registry/resolver/cache.go
```diff
@@ -416,6 +416,18 @@ func WithPackage(pkg string) OperatorPredicate {
 	}
 }
 
+func WithoutDeprecatedProperty() OperatorPredicate {
+	return func(o *Operator) bool {
+		props := o.bundle.GetProperties()
```
anik120, line 421: `nit`: `properties`, since you've already used `properties` [here](https://github.com/operator-framework/operator-lifecycle-manager/pull/1673/files#diff-4ad630e08ab9fb3fc440005ecf4cc9aaR261)

pkg/controller/registry/resolver/util_test.go
```diff
@@ -307,6 +307,16 @@ func apiSetToProperties(crds, apis APISet) (out []*api.Property) {
 			Value: string(val),
 		})
 	}
+	if deprecated {
+		val, err := json.Marshal(opregistry.DeprecatedProperty{})
+		if err != nil {
+			panic(err)
```
anik120, line 313: `Panic` seems drastic. Can we just return the error instead?

pkg/controller/registry/resolver/operators.go
```diff
@@ -297,8 +297,6 @@ func NewOperatorFromBundle(bundle *api.Bundle, startingCSV string, sourceKey reg
 		return op, nil
 	}
 
-
```
anik120, line 300: Looks like this file hasn't really been changed and just has noises.

pkg/controller/registry/resolver/util_test.go
gallettilance: I'm staying consistent with the pattern https://github.com/operator-framework/operator-lifecycle-manager/pull/1673/files/4988c89d50d10f84b1d2d90df81594cc5121db37#diff-6c0a163d01fd638f4e9b516fee8ad852L302

pkg/controller/registry/resolver/operators.go
gallettilance: different go formatting I guess

pkg/controller/registry/resolver/cache.go
gallettilance: there is only one deprecated property per bundle

anik120: Isn't `GetProperties` returning `properties` (irrespective of whether those properties are deprecated or not)? And there's a for loop after this that's iterating on the `properties` we're getting from `GetProperties`  

```
return func(o *Operator) bool {
		properties := o.bundle.GetProperties()
		for _, property := range properties {
			if property is deprecated {
				return false
			}
		}
		return true
	}
```
reads much better

pkg/controller/registry/resolver/util_test.go
anik120: Not sure what that link is pointing towards, but it's bad to panic from inside a running pod.

pkg/controller/registry/resolver/operators.go
anik120: Would reduce noise in this PR if this was reverted back.

pkg/controller/registry/resolver/util_test.go
benluddy: Maybe not ideal, but it is test code. Reading into the intention behind the existing panics, I'm guessing that the choice was between panicking here or changing the function's signature, and thereby making it unusable from within struct literals.

Big :+1: from me in general in investing in test code, but it may not need to block this PR. I don't think changing the function signature is strictly an improvement, so it would take some time to find a way to handle errors gracefully without introducing more boilerplate at every call site.

pkg/controller/registry/resolver/cache.go
benluddy: Alternatively, you could eliminate the variable `properties` completely:
```go
for _, p := range o.bundle.GetProperties()
```

pkg/controller/registry/resolver/util_test.go
anik120: IMHO it completely slipped me that this is a test. This is fine.

pkg/controller/registry/resolver/cache.go
gallettilance: @anik120 I thought you were talking about the function name - this makes sense to me

benluddy:
```
pkg/controller/registry/resolver/cache_test.go:310:35: not enough arguments in call to apiSetToProperties
	have (map["github.com/operator-framework/operator-registry/pkg/registry".APIKey]struct {}, nil)
	want (APISet, APISet, bool) 
```
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/operator-framework_operator-lifecycle-manager/1673/pull-ci-operator-framework-operator-lifecycle-manager-master-unit/1289276602584141824#1:build-log.txt%3A39

If these are passing locally, you probably need to rebase and check this out.

dinhxuanvu:
Replaced by #1699

PR #1682, opened 2020-07-27
feat(catalog): Add catalog weighting to dependency resolver

harishsurf:
Signed-off-by: Harish <hgovinda@redhat.com>

This PR includes `Priority` field (on the catalog source) to sort catalog source based on priority which is later consumed by the dependency resolver. 
- Higher priority value is preferred over lower priority
- The default priority for custom `catsrc` is 0.
- Sorting precedence for `catsrc` is namespace > priority > name

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -865,7 +865,7 @@ func (o *Operator) syncResolvingNamespace(obj interface{}) error {
 	// resolve a set of steps to apply to a cluster, a set of subscriptions to create/update, and any errors
 	steps, bundleLookups, updatedSubs, err := o.resolver.ResolveSteps(namespace, querier)
 	if err != nil {
-		go o.recorder.Event(ns, corev1.EventTypeWarning,"ResolutionFailed", err.Error())
+		go o.recorder.Event(ns, corev1.EventTypeWarning, "ResolutionFailed", err.Error())
 		return err
 	}
 
--- a/pkg/controller/registry/resolver/cache.go
+++ b/pkg/controller/registry/resolver/cache.go
@@ -9,12 +9,13 @@ import (
 	"time"
 
 	"github.com/blang/semver"
+	"github.com/sirupsen/logrus"
+
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/listers/operators/v1alpha1"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/registry"
 	"github.com/operator-framework/operator-registry/pkg/api"
 	"github.com/operator-framework/operator-registry/pkg/client"
 	opregistry "github.com/operator-framework/operator-registry/pkg/registry"
-	"github.com/sirupsen/logrus"
-
-	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/registry"
 )
 
 type RegistryClientProvider interface {
@@ -43,27 +44,33 @@ type OperatorCacheProvider interface {
 }
 
 type OperatorCache struct {
-	logger    logrus.FieldLogger
-	rcp       RegistryClientProvider
-	snapshots map[registry.CatalogKey]*CatalogSnapshot
-	ttl       time.Duration
-	sem       chan struct{}
-	m         sync.RWMutex
+	logger       logrus.FieldLogger
+	rcp          RegistryClientProvider
+	catsrcLister v1alpha1.CatalogSourceLister
+	snapshots    map[registry.CatalogKey]*CatalogSnapshot
+	ttl          time.Duration
+	sem          chan struct{}
+	m            sync.RWMutex
 }
+
+const defaultCatalogSourcePriority int = 0
+
+type catalogSourcePriority int
 
 var _ OperatorCacheProvider = &OperatorCache{}
 
-func NewOperatorCache(rcp RegistryClientProvider, log logrus.FieldLogger) *OperatorCache {
+func NewOperatorCache(rcp RegistryClientProvider, log logrus.FieldLogger, catsrcLister v1alpha1.CatalogSourceLister) *OperatorCache {
 	const (
 		MaxConcurrentSnapshotUpdates = 4
 	)
 
 	return &OperatorCache{
-		logger:    log,
-		rcp:       rcp,
-		snapshots: make(map[registry.CatalogKey]*CatalogSnapshot),
-		ttl:       5 * time.Minute,
-		sem:       make(chan struct{}, MaxConcurrentSnapshotUpdates),
+		logger:       log,
+		rcp:          rcp,
+		catsrcLister: catsrcLister,
+		snapshots:    make(map[registry.CatalogKey]*CatalogSnapshot),
+		ttl:          5 * time.Minute,
+		sem:          make(chan struct{}, MaxConcurrentSnapshotUpdates),
 	}
 }
 
@@ -151,11 +158,20 @@ func (c *OperatorCache) Namespaced(namespaces ...string) MultiCatalogOperatorFin
 
 	for _, miss := range misses {
 		ctx, cancel := context.WithTimeout(context.Background(), CachePopulateTimeout)
+
+		catsrcPriority := defaultCatalogSourcePriority
+		// Ignoring error and treat catsrc priority as 0 if not found.
+		catsrc, err := c.catsrcLister.CatalogSources(miss.Namespace).Get(miss.Name)
+		if err == nil {
+			catsrcPriority = catsrc.Spec.Priority
+		}
+
 		s := CatalogSnapshot{
-			logger: c.logger.WithField("catalog", miss),
-			key:    miss,
-			expiry: now.Add(c.ttl),
-			pop:    cancel,
+			logger:   c.logger.WithField("catalog", miss),
+			key:      miss,
+			expiry:   now.Add(c.ttl),
+			pop:      cancel,
+			priority: catalogSourcePriority(catsrcPriority),
 		}
 		s.m.Lock()
 		c.snapshots[miss] = &s
@@ -222,13 +238,13 @@ func ensurePackageProperty(o *Operator, name, version string) {
 		PackageName: name,
 		Version:     version,
 	}
-	byte, err := json.Marshal(prop)
+	bytes, err := json.Marshal(prop)
 	if err != nil {
 		return
 	}
 	o.properties = append(o.properties, &api.Property{
 		Type:  opregistry.PackageType,
-		Value: string(byte),
+		Value: string(bytes),
 	})
 }
 
@@ -277,6 +293,7 @@ type CatalogSnapshot struct {
 	operators []*Operator
 	m         sync.RWMutex
 	pop       context.CancelFunc
+	priority  catalogSourcePriority
 }
 
 func (s *CatalogSnapshot) Cancel() {
@@ -354,10 +371,14 @@ func (s SortableSnapshots) Less(i, j int) bool {
 		return false
 	}
 
-	// the rest are sorted first in namespace preference order, then by name
+	// the rest are sorted first on priority, namespace and then by name
+	if s.snapshots[i].priority != s.snapshots[j].priority {
+		return s.snapshots[i].priority > s.snapshots[j].priority
+	}
 	if s.snapshots[i].key.Namespace != s.snapshots[j].key.Namespace {
 		return s.namespaces[s.snapshots[i].key.Namespace] < s.namespaces[s.snapshots[j].key.Namespace]
 	}
+
 	return s.snapshots[i].key.Name < s.snapshots[j].key.Name
 }
 
--- a/pkg/controller/registry/resolver/operators.go
+++ b/pkg/controller/registry/resolver/operators.go
@@ -258,7 +258,7 @@ func NewOperatorFromBundle(bundle *api.Bundle, startingCSV string, sourceKey reg
 
 	// legacy support - if the api doesn't contain properties/dependencies, build them from required/provided apis
 	properties := bundle.Properties
-	if properties == nil || len(properties) == 0{
+	if properties == nil || len(properties) == 0 {
 		properties, err = apisToProperties(provided)
 		if err != nil {
 			return nil, err
@@ -296,8 +296,6 @@ func NewOperatorFromBundle(bundle *api.Bundle, startingCSV string, sourceKey reg
 		op.bundle = bundle
 		return op, nil
 	}
-
-
 
 	return &Operator{
 		name:         bundle.CsvName,
--- a/pkg/controller/registry/resolver/resolver_test.go
+++ b/pkg/controller/registry/resolver/resolver_test.go
@@ -306,6 +306,151 @@ func TestSolveOperators_FindLatestVersionWithNestedDependencies(t *testing.T) {
 	}
 }
 
+func TestSolveOperators_CatsrcPrioritySorting(t *testing.T) {
+	opToAddVersionDeps := []*api.Dependency{
+		{
+			Type:  "olm.package",
+			Value: `{"packageName":"packageB","version":"0.0.1"}`,
+		},
+	}
+
+	namespace := "olm"
+	customCatalog := registry.CatalogKey{"community", namespace}
+	newSub := newSub(namespace, "packageA", "alpha", customCatalog)
+	subs := []*v1alpha1.Subscription{newSub}
+
+	fakeNamespacedOperatorCache := NamespacedOperatorCache{
+		snapshots: map[registry.CatalogKey]*CatalogSnapshot{
+			registry.CatalogKey{
+				Namespace: "olm",
+				Name:      "community",
+			}: {
+				key: registry.CatalogKey{
+					Namespace: "olm",
+					Name:      "community",
+				},
+				operators: []*Operator{
+					genOperator("packageA.v1", "0.0.1", "packageA.v1", "packageA", "alpha", "community", namespace, nil,
+						nil, opToAddVersionDeps, ""),
+				},
+			},
+			registry.CatalogKey{
+				Namespace: "olm",
+				Name:      "community-operator",
+			}: {
+				key: registry.CatalogKey{
+					Namespace: "olm",
+					Name:      "community-operator",
+				},
+				operators: []*Operator{
+					genOperator("packageB.v1", "0.0.1", "", "packageB", "alpha", "community-operator",
+						namespace, nil, nil, nil, ""),
+				},
+			},
+			registry.CatalogKey{
+				Namespace: "olm",
+				Name:      "high-priority-operator",
+			}: {
+				key: registry.CatalogKey{
+					Namespace: "olm",
+					Name:      "high-priority-operator",
+				},
+				priority: catalogSourcePriority(100),
+				operators: []*Operator{
+					genOperator("packageB.v1", "0.0.1", "", "packageB", "alpha", "high-priority-operator",
+						namespace, nil, nil, nil, ""),
+				},
+			},
+		},
+	}
+
+	// operators sorted by priority.
+	satResolver := SatResolver{
+		cache: getFakeOperatorCache(fakeNamespacedOperatorCache),
+	}
+
+	operators, err := satResolver.SolveOperators([]string{"olm"}, []*v1alpha1.ClusterServiceVersion{}, subs)
+	assert.NoError(t, err)
+	expected := OperatorSet{
+		"packageA.v1": genOperator("packageA.v1", "0.0.1", "packageA.v1", "packageA", "alpha", "community", "olm",
+			nil, nil, opToAddVersionDeps, ""),
+		"packageB.v1": genOperator("packageB.v1", "0.0.1", "", "packageB", "alpha", "high-priority-operator", "olm",
+			nil, nil, nil, ""),
+	}
+	assert.Equal(t, 2, len(operators))
+	for k, e := range expected {
+		assert.EqualValues(t, e, operators[k])
+	}
+
+	// Catsrc with the same priority, ns, different name
+	fakeNamespacedOperatorCache.snapshots[registry.CatalogKey{
+		Namespace: "olm",
+		Name:      "community-operator",
+	}] = &CatalogSnapshot{
+		key: registry.CatalogKey{
+			Namespace: "olm",
+			Name:      "community-operator",
+		},
+		priority: catalogSourcePriority(100),
+		operators: []*Operator{
+			genOperator("packageB.v1", "0.0.1", "", "packageB", "alpha", "community-operator",
+				namespace, nil, nil, nil, ""),
+		},
+	}
+
+	satResolver = SatResolver{
+		cache: getFakeOperatorCache(fakeNamespacedOperatorCache),
+	}
+
+	operators, err = satResolver.SolveOperators([]string{"olm"}, []*v1alpha1.ClusterServiceVersion{}, subs)
+	assert.NoError(t, err)
+	expected = OperatorSet{
+		"packageA.v1": genOperator("packageA.v1", "0.0.1", "packageA.v1", "packageA", "alpha", "community", "olm",
+			nil, nil, opToAddVersionDeps, ""),
+		"packageB.v1": genOperator("packageB.v1", "0.0.1", "", "packageB", "alpha", "community-operator", "olm",
+			nil, nil, nil, ""),
+	}
+	assert.Equal(t, 2, len(operators))
+	for k, e := range expected {
+		assert.EqualValues(t, e, operators[k])
+	}
+
+	// operators from the same catalogs source should be prioritized.
+	fakeNamespacedOperatorCache.snapshots[registry.CatalogKey{
+		Namespace: "olm",
+		Name:      "community",
+	}] = &CatalogSnapshot{
+		key: registry.CatalogKey{
+			Namespace: "olm",
+			Name:      "community",
+		},
+		operators: []*Operator{
+			genOperator("packageA.v1", "0.0.1", "packageA.v1", "packageA", "alpha", "community", namespace, nil,
+				nil, opToAddVersionDeps, ""),
+			genOperator("packageB.v1", "0.0.1", "", "packageB", "alpha", "community",
+				namespace, nil, nil, nil, ""),
+		},
+	}
+
+	satResolver = SatResolver{
+		cache: getFakeOperatorCache(fakeNamespacedOperatorCache),
+	}
+
+	operators, err = satResolver.SolveOperators([]string{"olm"}, []*v1alpha1.ClusterServiceVersion{}, subs)
+	assert.NoError(t, err)
+	expected = OperatorSet{
+		"packageA.v1": genOperator("packageA.v1", "0.0.1", "packageA.v1", "packageA", "alpha", "community", "olm",
+			nil, nil, opToAddVersionDeps, ""),
+		"packageB.v1": genOperator("packageB.v1", "0.0.1", "", "packageB", "alpha", "community", "olm",
+			nil, nil, nil, ""),
+	}
+	assert.Equal(t, 2, len(operators))
+	for k, e := range expected {
+		assert.EqualValues(t, e, operators[k])
+	}
+
+}
+
 func TestSolveOperators_WithDependencies(t *testing.T) {
 	APISet := APISet{opregistry.APIKey{"g", "v", "k", "ks"}: struct{}{}}
 	Provides := APISet
@@ -596,7 +741,7 @@ func TestSolveOperators_PreferCatalogInSameNamespace(t *testing.T) {
 
 	csv := existingOperator(namespace, "packageA.v1", "packageA", "alpha", "", Provides, nil, nil, nil)
 	csvs := []*v1alpha1.ClusterServiceVersion{csv}
-	sub := existingSub(namespace,"packageA.v1", "packageA", "alpha", catalog)
+	sub := existingSub(namespace, "packageA.v1", "packageA", "alpha", catalog)
 	subs := []*v1alpha1.Subscription{sub}
 
 	fakeNamespacedOperatorCache := NamespacedOperatorCache{
@@ -789,7 +934,7 @@ func TestSolveOperators_SubscriptionlessOperatorsSatisfyDependencies(t *testing.
 	assert.NoError(t, err)
 	assert.Equal(t, 2, len(operators))
 	expected := OperatorSet{
-		"packageA.v1": stripBundle(genOperator("packageA.v1", "", "", "packageA", "alpha", "@existing", catalog.Namespace, nil, Provides, nil, "")),
+		"packageA.v1":     stripBundle(genOperator("packageA.v1", "", "", "packageA", "alpha", "@existing", catalog.Namespace, nil, Provides, nil, "")),
 		"packageB.v1.0.1": genOperator("packageB.v1.0.1", "1.0.1", "packageB.v1.0.0", "packageB", "alpha", catalog.Name, catalog.Namespace, Provides, nil, apiSetToDependencies(Provides, nil), ""),
 	}
 	for k := range expected {
@@ -903,8 +1048,8 @@ func TestSolveOperators_TransferApiOwnership(t *testing.T) {
 	catalog := registry.CatalogKey{Name: "community", Namespace: namespace}
 
 	phases := []struct {
-		subs  []*v1alpha1.Subscription
-		catalog *CatalogSnapshot
+		subs     []*v1alpha1.Subscription
+		catalog  *CatalogSnapshot
 		expected OperatorSet
 	}{
 		{
@@ -1023,7 +1168,7 @@ func genOperator(name, version, replaces, pkg, channel, catalogName, catalogName
 			Properties:   apiSetToProperties(providedAPIs, nil),
 		},
 		dependencies: dependencies,
-		properties: apiSetToProperties(providedAPIs, nil),
+		properties:   apiSetToProperties(providedAPIs, nil),
 		sourceInfo: &OperatorSourceInfo{
 			Catalog: registry.CatalogKey{
 				Name:      catalogName,
--- a/test/e2e/subscription_e2e_test.go
+++ b/test/e2e/subscription_e2e_test.go
@@ -1366,6 +1366,409 @@ var _ = Describe("Subscription", func() {
 		require.ElementsMatch(GinkgoT(), []string{"csv-a", "csv-b", "csv-e"}, got)
 	})
 
+	Context("with dependencies from different catsrc with priorities", func() {
+		var kubeClient operatorclient.ClientInterface
+		var crClient versioned.Interface
+		var crd apiextensions.CustomResourceDefinition
+		var packageMain, packageDepRight, packageDepWrong registry.PackageManifest
+		var csvsMain, csvsRight, csvsWrong []v1alpha1.ClusterServiceVersion
+		var catsrcMain, catsrcDepRight, catsrcDepWrong *v1alpha1.CatalogSource
+		var cleanup, cleanupSubscription cleanupFunc
+		const (
+			mainCSVName  = "csv-main"
+			rightCSVName = "csv-right"
+			wrongCSVName = "csv-wrong"
+		)
+
+		BeforeEach(func() {
+			kubeClient = ctx.Ctx().KubeClient()
+			crClient = ctx.Ctx().OperatorClient()
+			crd = newCRD(genName("ins"))
+
+			packageMain = registry.PackageManifest{PackageName: genName("PkgMain-")}
+			csv := newCSV(mainCSVName, testNamespace, "", semver.MustParse("0.1.0"), nil,
+				[]apiextensions.CustomResourceDefinition{crd}, nil)
+			packageMain.DefaultChannelName = stableChannel
+			packageMain.Channels = append(packageMain.Channels, registry.PackageChannel{Name: stableChannel, CurrentCSVName: csv.GetName()})
+
+			csvsMain = []v1alpha1.ClusterServiceVersion{csv}
+			csvsRight = []v1alpha1.ClusterServiceVersion{}
+			csvsWrong = []v1alpha1.ClusterServiceVersion{}
+		})
+
+		Context("creating CatalogSources with different names", func() {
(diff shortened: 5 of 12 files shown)


Bowenislandsong:
This PR **failed tests for 1 times** with 2 individual failed tests and 4 skipped tests. A test is considered flaky if failed on multiple commits. 
<details>

 totaltestcount: 1
flaketestcount: 2
skippedtestcount: 4
flaketests:
  - classname: End-to-end
    name: '**Subscription creation in case of transferring providedAPIs**'
    counts: 3
    details:
      - count: 3
        error: |4-


            /home/runner/work/operator-lifecycle-manager/operator-lifecycle-manager/test/e2e/subscription_e2e_test.go:1379

            	Error Trace:	subscription_e2e_test.go:1517
            	            				runner.go:113
            	            				runner.go:64
            	            				it_node.go:26
            	            				spec.go:215
            	            				spec.go:138
            	            				spec_runner.go:200
            	            				spec_runner.go:170
            	            				spec_runner.go:66
            	            				suite.go:62
            	            				ginkgo_dsl.go:226
            	            				ginkgo_dsl.go:214
            	            				e2e_test.go:54

    meandurationsec: 385.68525033333333
  - classname: End-to-end
    name: '**Garbage collection for dependent resources when a bundle with a new configmap
        is installed when the subscription is updated to a later CSV with a configmap
        with a new name should have removed the old configmap and put the new configmap
        in place**'
    counts: 1
    details:
      - count: 1
        error: |4-


            /home/runner/work/operator-lifecycle-manager/operator-lifecycle-manager/test/e2e/gc_e2e_test.go:507
            could not get installplan at complete phase
            Unexpected error:
                <*errors.errorString | 0xc0001f82a0>: {
                    s: "timed out waiting for the condition",
                }
            occurred
            /home/runner/work/operator-lifecycle-manager/operator-lifecycle-manager/test/e2e/gc_e2e_test.go:547
    meandurationsec: 312.100609
skippedtests:
  - classname: End-to-end
        successfully**'
    counts: 1
    details: []
    meandurationsec: 20.075543
  - classname: End-to-end
        prometheus objects**'
    counts: 1
    details: []
    meandurationsec: 15.079386
  - classname: End-to-end
    counts: 1
    details: []
    meandurationsec: 0.02003
  - classname: End-to-end
    counts: 1
    details: []
    meandurationsec: 0.151669

pkg/controller/registry/resolver/cache.go
```diff
@@ -151,11 +158,20 @@ func (c *OperatorCache) Namespaced(namespaces ...string) MultiCatalogOperatorFin
 
 	for _, miss := range misses {
 		ctx, cancel := context.WithTimeout(context.Background(), CachePopulateTimeout)
+
+		catsrcPriority := 0
+		// Ignoring error and treating catalog source priority as 0 if not found
+		catsrc, err := c.crClient.OperatorsV1alpha1().CatalogSources(miss.Namespace).Get(ctx, miss.Name, v1.GetOptions{})
```
benluddy, line 164: Is it relatively safe to get this value from the cache instead of hitting the API server directly?

pkg/controller/registry/resolver/cache.go
```diff
@@ -45,22 +46,28 @@ type OperatorCacheProvider interface {
...
+	crClient  versioned.Interface
 	snapshots map[registry.CatalogKey]*CatalogSnapshot
 	ttl       time.Duration
 	sem       chan struct{}
 	m         sync.RWMutex
 }
 
+type Priority struct {
+	catsrcPriority int
+}
```
benluddy, line 58:
```suggestion
type CatalogPriority int32
```

pkg/controller/registry/resolver/cache.go
```diff
@@ -354,10 +371,15 @@ func (s SortableSnapshots) Less(i, j int) bool {
 		return false
 	}
 
-	// the rest are sorted first in namespace preference order, then by name
+	// the rest are sorted first in namespace preference order, priority, then by name
```
benluddy, line 374: I think we want ... > "same namespace as the subscription" > catalog priority > lexical namespace name > ..., to preserve the existing behavior. We definitely don't want lexical namespace name > catalog priority in general, because the motivation for introducing catalog priorities is to make preferences less arcane.

pkg/controller/registry/resolver/cache.go
Bowenislandsong: dont think this is needed. since the API we merged is int typed

benluddy: It's fine to use int here since its values are only populated from API resources, but my suggestion is primarily to not make this a struct with a single unexported field. If nothing outside of this package will ever use it, then the type itself should be unexported.

Bowenislandsong: that i agree.

pkg/controller/registry/resolver/cache.go
Bowenislandsong: we don't have the value in cache yet, this is the first time getting it. Once we do, they are be resued until snapshot expires.

benluddy: I'm referring to the client-go cache of CatalogSources here, not the resolver's operator cache. Shouldn't the catalog operator always have a CatalogSource cached by the time it's needed here? Making a real request to the API server may be unnecessary.

Bowenislandsong: good call, taking a look

Bowenislandsong: https://github.com/operator-framework/operator-lifecycle-manager/pull/1688/commits/6c2f414c727a3f98804adcea255ef2876b91561e

Bowenislandsong:
This PR **failed tests for 1 times** with 2 individual failed tests and 4 skipped tests. A test is considered flaky if failed on multiple commits. 
<details>

 totaltestcount: 1
flaketestcount: 2
skippedtestcount: 4
flaketests:
  - classname: End-to-end
    name: '**Subscription puppeting CatalogSource health status when missing target
        catalog should surface the missing catalog**'
    counts: 3
    details:
      - count: 3
        error: |4-


            /home/runner/work/operator-lifecycle-manager/operator-lifecycle-manager/test/e2e/subscription_e2e_test.go:588
            Unexpected error:
                <*errors.errorString | 0xc0012dfe70>: {
                    s: "E2E bug detected: configmaps \"mock-ocs\" already exists",
                }
            occurred
            /home/runner/work/operator-lifecycle-manager/operator-lifecycle-manager/test/e2e/subscription_e2e_test.go:590
    meandurationsec: 0.03879866666666667
  - classname: End-to-end
    name: '**Subscription with dependencies from different catsrc with priorities**'
    counts: 2
    details:
      - count: 2
        error: |4-


            /home/runner/work/operator-lifecycle-manager/operator-lifecycle-manager/test/e2e/subscription_e2e_test.go:1369

            	Error Trace:	subscription_e2e_test.go:1521
            	            				runner.go:113
            	            				runner.go:64
            	            				it_node.go:26
            	            				spec.go:215
            	            				spec.go:138
            	            				spec_runner.go:200
            	            				spec_runner.go:170
            	            				spec_runner.go:66
            	            				suite.go:62
            	            				ginkgo_dsl.go:226
            	            				ginkgo_dsl.go:214
            	            				e2e_test.go:54
            	            	clusterserviceversions.operators.coreos.com "csv-main" not found

    meandurationsec: 26.818526
skippedtests:
  - classname: End-to-end
        prometheus objects**'
    counts: 1
    details: []
    meandurationsec: 23.258269
  - classname: End-to-end
    counts: 1
    details: []
    meandurationsec: 0.071045
  - classname: End-to-end
        successfully**'
    counts: 1
    details: []
    meandurationsec: 10.226736
  - classname: End-to-end
    counts: 1
    details: []
    meandurationsec: 0.553019

pkg/controller/registry/resolver/cache.go
harishsurf: git cherry-picked those cache changes.

pkg/controller/registry/resolver/cache.go
```diff
@@ -9,12 +9,12 @@ import (
 	"time"
 
 	"github.com/blang/semver"
+	v1alpha1listers "github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/listers/operators/v1alpha1"
```
Bowenislandsong, line 12: import format nit

pkg/controller/registry/resolver/cache.go
```diff
@@ -354,10 +368,14 @@ func (s SortableSnapshots) Less(i, j int) bool {
 		return false
 	}
 
-	// the rest are sorted first in namespace preference order, then by name
+	// the rest are sorted first on priority, namespace and then by name
```
Bowenislandsong, line 371: Not really objecting this, but based on the previous comments, I thought we want to prioritize same ns, then priority, then ns lexicographic,...

I am not objecting the idea to have priority going before ns, but it does not seem like what we agreed on. unless @ecordell changed his mind?

Bowenislandsong: resolved. priority goes first because it gives user more control

test/e2e/subscription_e2e_test.go
```diff
@@ -1366,6 +1366,175 @@ var _ = Describe("Subscription", func() {
...
+		Eventually(func() []string {
+			ip, err := crClient.OperatorsV1alpha1().InstallPlans(testNamespace).Get(context.TODO(), subscription.Status.InstallPlanRef.Name, metav1.GetOptions{})
+			if err != nil {
+				return nil
+			}
+			got = ip.Spec.ClusterServiceVersionNames
+			return got
+		}).ShouldNot(BeNil())
+		Expect(got).Should(ConsistOf([]string{"csv-main", "csv-a", "csv-b", "csv-c", "csv-d"}))
+		//require.ElementsMatch(GinkgoT(), []string{"csv-main", "csv-a", "csv-b", "csv-c", "csv-d"}, got)
```
benluddy, line 1535:
```suggestion
		Eventually(func() ([]string, error) {
			ip, err := crClient.OperatorsV1alpha1().InstallPlans(testNamespace).Get(context.TODO(), subscription.Status.InstallPlanRef.Name, metav1.GetOptions{})
			if err != nil {
				return nil, err
			}
			return ip.Spec.ClusterServiceVersionNames, nil
		}).Should(ConsistOf([]string{"csv-main", "csv-a", "csv-b", "csv-c", "csv-d"}))
```

No need for the intermediate variable `got` here, since the assertion can match the return value directly. Also, including the error as a second return value will make failures more useful when the assertion times out with an error (http://onsi.github.io/gomega/#handling-errors).

test/e2e/subscription_e2e_test.go
```diff
@@ -1366,6 +1366,175 @@ var _ = Describe("Subscription", func() {
 		require.ElementsMatch(GinkgoT(), []string{"csv-a", "csv-b", "csv-e"}, got)
 	})
 
+	It("with dependencies from different catsrc with priorities", func() {
```
benluddy, line 1369: I would be interested in ways to simplify this test. A couple ideas:

- Split into multiple, smaller specs that involve choices between two operators in different catalogs. That would open up the possibility of a table-based test that parameterizes catalog namespace, name, and priority.
- Can catalog preferences be tested without defining any operator-to-operator dependencies at all? Or, can a single subscription that omits catalog namespace and name exercise the same behavior?

If those two are possible, now all cases are reduced to two operator versions, two packages, two catalogs, and one subscription.

test/e2e/subscription_e2e_test.go
Bowenislandsong: nice

test/e2e/subscription_e2e_test.go
Bowenislandsong: the 4 priority senarios are completely separable. so table testing is possible. or we could separate this test into 4 smaller tests. 

I can't think of a way to test this without operator-to-operator dependencies. besides, as e2e, I actually prefer to have this dependency because that is their primary function.

deploy/chart/crds/0000_50_olm_00-clusterserviceversions.crd.yaml
```diff
@@ -662,13 +662,19 @@ spec:
...
+                          description: StrategyDeploymentSpec contains the name, spec
+                            and labels for the deployment ALM should create
                           type: object
                           required:
                           - name
                           - spec
                           properties:
+                            label:
+                              description: Set is a map of label:value. It implements
+                                Labels.
```
anik120, line 674: Isn't this suppose to be a user readable message? i.e It's a help message for someone who's not looking at the code at all, who's possibly looking at this by using `kubectl describe`, and therefore should not contain any `implement`ation details. 
Also not sure why this is being added in this PR?

pkg/controller/operators/catalog/operator.go
```diff
@@ -865,7 +865,7 @@ func (o *Operator) syncResolvingNamespace(obj interface{}) error {
 	// resolve a set of steps to apply to a cluster, a set of subscriptions to create/update, and any errors
 	steps, bundleLookups, updatedSubs, err := o.resolver.ResolveSteps(namespace, querier)
 	if err != nil {
-		go o.recorder.Event(ns, corev1.EventTypeWarning,"ResolutionFailed", err.Error())
+		go o.recorder.Event(ns, corev1.EventTypeWarning, "ResolutionFailed", err.Error())
```
anik120, line 868: Could you get rid of this noise? Clutters the PR a bit.

pkg/controller/registry/resolver/operators.go
```diff
@@ -258,7 +258,7 @@ func NewOperatorFromBundle(bundle *api.Bundle, startingCSV string, sourceKey reg
 
 	// legacy support - if the api doesn't contain properties/dependencies, build them from required/provided apis
 	properties := bundle.Properties
-	if properties == nil || len(properties) == 0{
+	if properties == nil || len(properties) == 0 {
```
anik120, line 261: Same here.

pkg/controller/registry/resolver/resolver_test.go
```diff
@@ -306,6 +306,151 @@ func TestSolveOperators_FindLatestVersionWithNestedDependencies(t *testing.T) {
...
+			},
+			registry.CatalogKey{
+				Namespace: "olm",
+				Name:      "high-priority-operator",
+			}: {
+				key: registry.CatalogKey{
+					Namespace: "olm",
+					Name:      "high-priority-operator",
+				},
+				priority: catalogPriority(100),
```
anik120, line 358: `100` should be defined as a constant in pkg `resolver` and that constant should be used here.

pkg/controller/registry/resolver/resolver_test.go
```diff
@@ -1041,4 +1186,4 @@ func genOperator(name, version, replaces, pkg, channel, catalogName, catalogName
 func stripBundle(o *Operator) *Operator {
 	o.bundle = nil
 	return o
-}
\ No newline at end of file
+}
```
anik120, line 1189: lot of noise towards the end of this file too.

deploy/chart/crds/0000_50_olm_00-clusterserviceversions.crd.yaml
awgreene: The description here is autogenerated based on [comments](https://github.com/kubernetes/apimachinery/blob/master/pkg/labels/labels.go#L34) above the k8s.io/apimachinery/pkg/labels.Set type.

pkg/controller/registry/resolver/cache.go
awgreene: If you search the OLM code for `github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/listers/operators/v1alpha1`, it's typically imported as `v1alpha1, listers, or operatorslisters`, which `v1alpha1` being the most common approach.

pkg/controller/registry/resolver/cache.go
```diff
@@ -43,27 +44,31 @@ type OperatorCacheProvider interface {
...
+	logger       logrus.FieldLogger
+	rcp          RegistryClientProvider
+	catsrcLister v1alpha1listers.CatalogSourceLister
+	snapshots    map[registry.CatalogKey]*CatalogSnapshot
+	ttl          time.Duration
+	sem          chan struct{}
+	m            sync.RWMutex
 }
 
+type catalogPriority int
```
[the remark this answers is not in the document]
awgreene: My understanding is that this only needs to be exported if the type is used outside of the package.

pkg/controller/registry/resolver/cache.go
```diff
@@ -151,11 +156,20 @@ func (c *OperatorCache) Namespaced(namespaces ...string) MultiCatalogOperatorFin
 
 	for _, miss := range misses {
 		ctx, cancel := context.WithTimeout(context.Background(), CachePopulateTimeout)
+
+		catsrcPriority := 0
```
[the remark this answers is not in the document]
awgreene: The name of the constant should denote that it is the default catsrc value as well.

pkg/controller/registry/resolver/resolver_test.go
Bowenislandsong: no, this is test only contant.

pkg/controller/registry/resolver/resolver_test.go
Bowenislandsong: some format should be fine right. nothing wrong with formating it when it was not properly formated the first time.

pkg/controller/registry/resolver/cache.go
Bowenislandsong: in here i think its ok to convert to v1alpha1, but there is another place that as v1alpha1 already for another package.

pkg/controller/registry/resolver/cache.go
```diff
@@ -151,11 +156,20 @@ func (c *OperatorCache) Namespaced(namespaces ...string) MultiCatalogOperatorFin
...
+		// Ignoring error and treat catsrc priority as 0 if not found.
+		catsrc, err := c.catsrcLister.CatalogSources(miss.Namespace).Get(miss.Name)
```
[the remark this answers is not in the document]
benluddy: I don't know, "miss" has a very specific meaning in a cache implementation.

pkg/controller/registry/resolver/cache.go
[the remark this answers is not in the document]
Bowenislandsong: right, therefore it does not need to be exported.

pkg/controller/registry/resolver/cache.go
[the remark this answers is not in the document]
Bowenislandsong: it's an initialization, not a constant.

pkg/controller/registry/resolver/cache.go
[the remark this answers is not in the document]
benluddy: I wouldn't export this identifier unless there's a good reason.

pkg/controller/registry/resolver/cache.go
```diff
@@ -354,10 +369,14 @@ func (s SortableSnapshots) Less(i, j int) bool {
 		return false
 	}
 
-	// the rest are sorted first in namespace preference order, then by name
+	// the rest are sorted first on priority, namespace and then by name
```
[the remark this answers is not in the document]
Bowenislandsong: Maybe could be substitued for more structured comment for the entire `Less` function to explain the rules.

dinhxuanvu:
This PR can be merged as soon as it is lgtm and approve. So I'm removing the hold.
/hold cancel

deploy/chart/crds/0000_50_olm_00-clusterserviceversions.crd.yaml
```diff
@@ -662,13 +662,19 @@ spec:
                       deployments:
                         type: array
                         items:
-                          description: StrategyDeploymentSpec contains the name and
-                            spec for the deployment ALM should create
+                          description: StrategyDeploymentSpec contains the name, spec
+                            and labels for the deployment ALM should create
```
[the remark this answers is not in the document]
Bowenislandsong: aren't these generated?

pkg/controller/registry/resolver/cache.go
anik120: `olmv1alpha1` should be solid enough to avoid confusion. 
At the end of the day we should adopt a convention that is followed throughout the entire project, but that's a discussion for later.

deploy/chart/crds/0000_50_olm_00-clusterserviceversions.crd.yaml
anik120: I see. That's unfortunate coz as an user this would mean nothing to me. Will we be breaking the convention  if we override the description here?

pkg/controller/registry/resolver/cache.go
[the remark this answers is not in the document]
anik120: `cacheHit` and `cacheMiss`?

pkg/controller/registry/resolver/cache.go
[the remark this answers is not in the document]
anik120: @Bowenislandsong I was suggesting an initialization like: 

`catsrcPriority := resolver.DefaultPriority`

Don't quote me on the package the constant should be initialized,  but generally seeing numbers in the middle of the code would confuse a reader later on: "why 100? Why not another number? What does this number mean?". Using a constant helps provide context.

pkg/controller/registry/resolver/cache.go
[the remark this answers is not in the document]
anik120: True.

pkg/controller/registry/resolver/resolver_test.go
anik120: Can we initialize a `defaultPriorityTest`? It's used in 2 different places, again, a future reader with no context has no way of knowing why `100` is used here.

pkg/controller/registry/resolver/resolver_test.go
anik120: The problem with this different people have different formatting settings in their code editor. So every time this code base is touched, the formatting will just go back and forth between these options. 
This PR isn't about correcting the formatting is my point. 
If we want to adopt a convention for the project so that everyone contributing to this project have the same formatting settings set for their code editor, that can be done in a separate PR.

pkg/controller/registry/resolver/cache.go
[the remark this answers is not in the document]
anik120: +1 to that. 
Quote from Clean code by Robert Martin: 

"The proper use of comments is to compensate our failure to express ourself  in code" 

"Code changes and evolves. Chunks of it move from here to there. Those chunks bifurcate and reproduce and come together again to form chimeras. Unfortunately the comments don't always follow them, can't always follow them." 

Further reading on comments: https://blog.usejournal.com/stop-writing-code-comments-28fef5272752

pkg/controller/registry/resolver/resolver_test.go
anik120: Also, whoever formatted the code the way it was before must have been of the opinion that that was the right formatting for the code, while we're saying no I disagree and it has to be changed in this PR.

Bowenislandsong: it's not a settign thing if there no new line at the end. other format input, yes.

pkg/controller/registry/resolver/cache.go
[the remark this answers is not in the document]
Bowenislandsong: yeah, doing something like this is good.

pkg/controller/registry/resolver/cache.go
awgreene: We don't even need to give it a special name - no other import in this file users `v1alpha1`.

pkg/controller/registry/resolver/resolver_test.go
anik120: Right my comment is on the other formatting bits.

ecordell:
@harishsurf once you rebase you should get the latest `api` package, so we don't need those changes anymore

pkg/controller/registry/resolver/cache.go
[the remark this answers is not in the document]
benluddy: Is that really helpful in a file named `cache.go` in a function called `New*Cache`?

kevinrizza:
I think some of Anik's nits around committing files that have no changes except for `gofmt` are reasonable, but I don't think it should block this pr.

/lgtm

test/e2e/subscription_e2e_test.go
```diff
@@ -1366,6 +1366,258 @@ var _ = Describe("Subscription", func() {
...
+			packageMain.DefaultChannelName = stableChannel
+			packageMain.Channels = append(packageMain.Channels, registry.PackageChannel{Name: stableChannel, CurrentCSVName: csv.GetName()})
+			csvsMain = []v1alpha1.ClusterServiceVersion{csv}
+
+			csvsRight = []v1alpha1.ClusterServiceVersion{}
+			csvsWrong = []v1alpha1.ClusterServiceVersion{}
+		})
+
+		AfterEach(func() {
+			cleanupSubscription()
```
benluddy, line 1402: This could be nil, depending on the behavior of each spec, and the panic would be misleading.

test/e2e/subscription_e2e_test.go
```diff
@@ -1366,6 +1366,258 @@ var _ = Describe("Subscription", func() {
...
+			csvsWrong = append(csvsWrong, csv)
+
+			By("creating catalogsources", func() {
+				var c1, c2 cleanupFunc
+				catsrcMain, c1 = createInternalCatalogSource(kubeClient, crClient, genName("catsrc"), testNamespace,
+					[]registry.PackageManifest{packageDepRight, packageMain}, []apiextensions.CustomResourceDefinition{crd}, csvsMain)
+
+				catsrcDepWrong, c2 = createInternalCatalogSource(kubeClient, crClient, genName("catsrc"), testNamespace,
+					[]registry.PackageManifest{packageDepWrong}, []apiextensions.CustomResourceDefinition{crd}, csvsWrong)
+				catsrcDepWrong.Spec.Priority = 100
```
benluddy, line 1506: This priority assignment never gets sent back to the API server. Is this test valid?

test/e2e/subscription_e2e_test.go
```diff
@@ -1366,6 +1366,258 @@ var _ = Describe("Subscription", func() {
...
+				}
+			})
+
+			By("waiting for catalogsources to be ready", func() {
+				_, err := fetchCatalogSourceOnStatus(crClient, catsrcMain.GetName(), testNamespace, catalogSourceRegistryPodSynced)
+				Expect(err).ToNot(HaveOccurred())
+				_, err = fetchCatalogSourceOnStatus(crClient, catsrcDepWrong.GetName(), testNamespace, catalogSourceRegistryPodSynced)
+				Expect(err).ToNot(HaveOccurred())
+			})
+			By("choose the dependent package from the same catsrc as the installing operator", func() {
```
benluddy, line 1520: The "By" descriptions in each of these tests are more like what I would expect the "It" descriptions to say -- they are describing the expected behavior.

test/e2e/subscription_e2e_test.go
```diff
@@ -1366,6 +1366,258 @@ var _ = Describe("Subscription", func() {
...
+				Expect(err).ToNot(HaveOccurred())
+			})
+
+			By("choose the dependent package from the same catsrc that has a higher priority value", func() {
+				ensureInstalledCSVs()
+			})
+		})
+
+		It("the same dependency exists on both the catsrc of the installing operator and another catsrc with a higher"+
+			" priority", func() {
```
benluddy, line 1486: Nit: Splitting the spec description across two lines is funky. There's no hard line length rule, and if there were, this would probably be an exception to it.

test/e2e/subscription_e2e_test.go
```diff
@@ -1366,6 +1366,258 @@ var _ = Describe("Subscription", func() {
...
+			csvsWrong = append(csvsWrong, csv)
+
+			By("creating catalogsources", func() {
+				var c1, c2, c3 cleanupFunc
+				catsrcMain, c1 = createInternalCatalogSource(kubeClient, crClient, genName("catsrc"), testNamespace,
+					[]registry.PackageManifest{packageMain}, nil, csvsMain)
+
+				catsrcDepRight, c2 = createInternalCatalogSource(kubeClient, crClient, "catsrc1", testNamespace,
+					[]registry.PackageManifest{packageDepRight}, []apiextensions.CustomResourceDefinition{crd}, csvsRight)
+				catsrcDepRight.Spec.Priority = 100
```
benluddy, line 1458: Same here as below, where are these priorities being sent to the API server?

benluddy:
Putting a temporary hold as a precaution while investigating my question about the end-to-end tests.

/hold

test/e2e/subscription_e2e_test.go
Bowenislandsong: lol, goland, i sorta need to disable that character count function.

test/e2e/subscription_e2e_test.go
Bowenislandsong: they are more like steps within the It, i dont think they should be it within an it.

test/e2e/subscription_e2e_test.go
Bowenislandsong: I don't think the tests hit panic or nil right? they all passed. the func value were assigned within the test.

harishsurf:
```suggestion
if cleanupSubscription !=nil {
     cleanupSubscription()
}
```

benluddy: > I don't think the tests hit panic or nil right? they all passed. the func value were assigned within the test.

Yup, they won't panic if they all pass. But if any of the tests end early (like if an assertion fails), the assignment to cleanupSubscription will not happen and this line will panic.

test/e2e/subscription_e2e_test.go
benluddy: The string passed to `It` is essentially the "Then" in the BDD "GivenWhenThen", and it represents the behavior that the program being tested is expected to demonstrate. If one of these tests fails before it reaches the end, the test report will not tell you which behavior was broken.

test/e2e/subscription_e2e_test.go
benluddy: Haha, it was only trying to help! :pencil2:

test/e2e/subscription_e2e_test.go
Bowenislandsong: thanks, you are right.

test/e2e/subscription_e2e_test.go
Bowenislandsong: so you are suggesting to convert all the By within the It to be It?

Bowenislandsong: we can do that

test/e2e/subscription_e2e_test.go
Bowenislandsong: i see

test/e2e/subscription_e2e_test.go
benluddy: > so you are suggesting to convert all the By within the It to be It?

No, only to change the descriptions being passed to the existing `It`s

test/e2e/subscription_e2e_test.go
```diff
@@ -1366,6 +1366,409 @@ var _ = Describe("Subscription", func() {
...
+				})
+			})
+		})
+
+		Context("creating CatalogSources with same name but different priority values", func() {
+			var catsrcCleanup1, catsrcCleanup2 cleanupFunc
+
+			BeforeEach(func() {
+
+				//same dependency exists on both the catsrc of the installing operator and another catsrc with a higher priority", func() {
```
ecordell, line 1497: there's a func() in your comment

Bowenislandsong:
/hold cancel
e2e test comments addressed.

test/e2e/subscription_e2e_test.go
```diff
@@ -1366,6 +1366,409 @@ var _ = Describe("Subscription", func() {
...
+				})
+				AfterEach(func() {
+					if cleanupSubscription != nil {
+						cleanupSubscription()
+					}
+					if cleanup != nil {
+						cleanup()
+					}
+				})
+				It("choose the dependent package from the same catsrc as the installing operator", func() {
```
ecordell, line 1564:
```suggestion
				It("chooses the dependent package from the same catsrc as the installing operator", func() {
```

test/e2e/subscription_e2e_test.go
```diff
@@ -1366,6 +1366,409 @@ var _ = Describe("Subscription", func() {
...
+			csvsRight = []v1alpha1.ClusterServiceVersion{}
+			csvsWrong = []v1alpha1.ClusterServiceVersion{}
+		})
+
+		Context("creating CatalogSources with different names", func() {
+			var catsrcCleanup1, catsrcCleanup2, catsrcCleanup3 cleanupFunc
+
+			BeforeEach(func() {
+
+				packageDepRight = registry.PackageManifest{PackageName: "PackageDependent"}
```
ecordell, line 1404: I don't know that this should block, but I'm finding I have to do a lot of work to figure out what initial state this test is configuring.

Notes I'm taking as I go:

```
package main
  - stable - "csv-main" - requires CRD

package depRight
 - alpha - "csv-right" - provides CRD

package depWrong
 - alpha - "csv-wrong" - provides CRD

CatSrcMain - 
 - package main

CatSrcDepRight
 - priority: 100
 - package depRight

CatSrcDepWrong
 - priority: 100
 - package depWrong
```

test/e2e/subscription_e2e_test.go
```diff
@@ -1366,6 +1366,409 @@ var _ = Describe("Subscription", func() {
...
+				packageDepWrong = packageDepRight
+				packageDepWrong.Channels = []registry.PackageChannel{{Name: alphaChannel, CurrentCSVName: csv.GetName()}}
+				csvsWrong = append(csvsWrong, csv)
+
+				catsrcMain, catsrcCleanup1 = createInternalCatalogSource(kubeClient, crClient, genName("catsrc"), testNamespace,
+					[]registry.PackageManifest{packageMain}, nil, csvsMain)
+
+				catsrcDepRight, catsrcCleanup2 = createInternalCatalogSource(kubeClient, crClient, "catsrc1", testNamespace,
+					[]registry.PackageManifest{packageDepRight}, []apiextensions.CustomResourceDefinition{crd}, csvsRight)
+				updateCatSrcPriority(crClient, testNamespace, catsrcDepRight, 100)
```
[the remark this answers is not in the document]
ecordell: if this test is testing lexicographic ordering I don't think we need to update the priorities at all

test/e2e/subscription_e2e_test.go
```diff
@@ -1366,6 +1366,409 @@ var _ = Describe("Subscription", func() {
...
+				packageDepWrong = packageDepRight
+				packageDepWrong.Channels = []registry.PackageChannel{{Name: alphaChannel, CurrentCSVName: csv.GetName()}}
+				csvsWrong = append(csvsWrong, csv)
+
+				catsrcMain, catsrcCleanup1 = createInternalCatalogSource(kubeClient, crClient, genName("catsrc"), testNamespace,
+					[]registry.PackageManifest{packageDepRight, packageMain}, []apiextensions.CustomResourceDefinition{crd}, csvsMain)
+
+				catsrcDepWrong, catsrcCleanup2 = createInternalCatalogSource(kubeClient, crClient, genName("catsrc"), testNamespace,
+					[]registry.PackageManifest{packageDepWrong}, []apiextensions.CustomResourceDefinition{crd}, csvsWrong)
+				updateCatSrcPriority(crClient, testNamespace, catsrcDepWrong, 100)
```
ecordell, line 1515: this is setting up almost the exact same catalogs each time - could we just create them in a context surrounding all of these tests, and update their priorities in the child contexts?

or at the very least factor our the definitions, even if we're going to create new catalogs for each test case

test/e2e/subscription_e2e_test.go
ecordell: we also already have a test for this - we can probably delete the other one.

test/e2e/subscription_e2e_test.go
```diff
@@ -1366,6 +1366,409 @@ var _ = Describe("Subscription", func() {
...
+				packageDepRight.DefaultChannelName = alphaChannel
+				packageDepRight.Channels = []registry.PackageChannel{{Name: alphaChannel, CurrentCSVName: csv.GetName()}}
+				csvsRight = append(csvsRight, csv)
+
+				csv.Name = wrongCSVName
+				packageDepWrong = packageDepRight
+				packageDepWrong.Channels = []registry.PackageChannel{{Name: alphaChannel, CurrentCSVName: csv.GetName()}}
+				csvsWrong = append(csvsWrong, csv)
+
+				dNS, err := kubeClient.KubernetesInterface().CoreV1().Namespaces().Create(context.TODO(),
```
ecordell, line 1690: I think this test is intending to check for namespace ordering in the sort, right?

If that's the case, the catalog in this generated namespace won't even be in the set that's being sorted. The only namespaces that are considered for sorting are: 1. the namespace where the subscription is and 2. the namespace that is configured for global catalogs

ecordell:
/lgtm

there are a couple of improvements we should make to the tests, but we can do that in a separate PR.

test/e2e/subscription_e2e_test.go
Bowenislandsong: mb, I forgot about this.

Bowenislandsong: definately need to have a fix next week

test/e2e/subscription_e2e_test.go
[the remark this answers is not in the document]
Bowenislandsong: you are right, I forgot why priority is set here "different people refactored this".

test/e2e/subscription_e2e_test.go
Bowenislandsong: yeah, this needs fixing.

dinhxuanvu:
Replaced by #1705

test/e2e/subscription_e2e_test.go
Bowenislandsong: i'll try to write context better in there

PR #1684, opened 2020-07-28
Bug 1860185: fix(install): account for installplans when calculating generation

njhale:
Account for existing InstallPlans while calculating the maximum
generation. This ensures new Subscriptions get a new generation and
are never associated with an existing InstallPlan in cases where
previous Subscriptions were deleted.


tlwu2013:
Hi @njhale, @dinhxuanvu, @ecordell  I can also reproduce this bug on OCP 4.4.14 (cluster-bot). 
We might need to backport this to 4.4.z as well. Can you guys help double-check? thanks!

njhale:
> Hi @njhale, @dinhxuanvu, @ecordell I can also reproduce this bug on OCP 4.4.14 (cluster-bot).
> We might need to backport this to 4.4.z as well. Can you guys help double-check? thanks!

@tlwu2013 we are planning to backport this to 4.4.z.

PR #1687, opened 2020-07-29
feat(resolver): Add label dependency type to new resolver

dinhxuanvu:
Introduce the new label dependency as predicate for the resolver.
Also add new function to match label dependency against operator
properties.

Add an option for operator authors to fine-tune the dependency selection in case of conflicts. The label dependency can be used as a mark/filter for the resolver to select the desirable operator during dependency resolution.

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/registry/resolver/resolver_test.go
+++ b/pkg/controller/registry/resolver/resolver_test.go
@@ -419,6 +419,107 @@ func TestSolveOperators_WithGVKDependencies(t *testing.T) {
 	}
 }
 
+func TestSolveOperators_WithLabelDependencies(t *testing.T) {
+	namespace := "olm"
+	catalog := registry.CatalogKey{"community", namespace}
+
+	newSub := newSub(namespace, "packageA", "alpha", catalog)
+	subs := []*v1alpha1.Subscription{newSub}
+
+	deps := []*api.Dependency{
+		{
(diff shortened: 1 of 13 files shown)


dinhxuanvu:
/hold
Put a hold here pending further review and approval.

ecordell:
looks good - can you add a test case to the `resolver_tests`?

dinhxuanvu:
@anik120 @ecordell Test case added. PTAL

pkg/controller/registry/resolver/resolver_test.go
```diff
@@ -419,6 +419,75 @@ func TestSolveOperators_WithGVKDependencies(t *testing.T) {
...
+	Provides := APISet
+
+	namespace := "olm"
+	catalog := registry.CatalogKey{"community", namespace}
+
+	newSub := newSub(namespace, "packageA", "alpha", catalog)
+	subs := []*v1alpha1.Subscription{newSub}
+
+	deps := []*api.Dependency{
+		{
```
benluddy, line 433: This test relies on many more behaviors than just satisfying label dependencies. How about two test cases: both with two operators (A and B) available, a subscription to A, and a label dependency for A. Then in the first test case, B has the label and both are installed, and in the second, B does not have the label and resolution fails.

PR #1689, opened 2020-07-29
[release-4.5] Bug 1862407: fix(resolver): be smarter about the way downgrades are attempted

ecordell:
"downgrade" here refers to downgrading potential updates back to what
has already been installed.

this fixes issues where OLM would otherwise not have been able to
determine a safe set of operators to update to - the case where, after
tentative updates to the current generation, there are required apis
missing from the set.

the previous algorithm solves this by downgrading the updates that
require the missing apis, which has the potential to downgrade back
to the initial set of operators.

in the case where the api was provided by some operator in the previous
generation, it would be possible to not downgrade all requirers, but
rather downgrade the operator that previously provided the api.

this commit implements an exhaustive search of potential downgrades for
a generation, and uses one without missing required apis if one is found

if none is found, the old behavior takes place, and the set may still
contract back to what is currently installed.

the new resolver avoids this issue entirely, so to avoid complicating
the implementation a limit of 64 updated operators per generation has
been added.

the implementation uses a bitmask to select either the old or new
operator from the set. finding all possible combinations of old/new
versions is then a simple integer count from 0 to the max.

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/registry/resolver/evolver.go
+++ b/pkg/controller/registry/resolver/evolver.go
@@ -17,6 +17,7 @@ type Evolver interface {
 type NamespaceGenerationEvolver struct {
 	querier SourceQuerier
 	gen     Generation
+	replacements map[OperatorSurface]OperatorSurface
 }
 
 func NewNamespaceGenerationEvolver(querier SourceQuerier, gen Generation) Evolver {
@@ -44,6 +45,11 @@ func (e *NamespaceGenerationEvolver) Evolve(add map[OperatorSourceInfo]struct{})
 		return err
 	}
 
+	// if apis are missing, attempt to contract back to a good set by trying all combinations of rollbacks to updates
+	if err := e.downgradeUpdates(); err != nil {
+		return err
+	}
+
 	// for any remaining missing APIs, attempt to downgrade the operator that required them
 	// this may contract the generation back to the original set!
 	e.downgradeAPIs()
@@ -51,6 +57,8 @@ func (e *NamespaceGenerationEvolver) Evolve(add map[OperatorSourceInfo]struct{})
 }
 
 func (e *NamespaceGenerationEvolver) checkForUpdates() error {
+	e.replacements = make(map[OperatorSurface]OperatorSurface)
+
 	// maps the old operator identifier to the new operator
 	updates := EmptyOperatorSet()
 
@@ -74,6 +82,7 @@ func (e *NamespaceGenerationEvolver) checkForUpdates() error {
 		}
 		o.SetReplaces(op.Identifier())
 		updates[op.Identifier()] = o
+		e.replacements[op] = o
 	}
 
 	// remove any operators we found updates for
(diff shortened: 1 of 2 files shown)


pkg/controller/registry/resolver/evolver.go
```diff
@@ -153,7 +162,65 @@ func (e *NamespaceGenerationEvolver) queryForRequiredAPIs() error {
 	return nil
 }
 
+func (e *NamespaceGenerationEvolver) downgradeUpdates() error {
+	// no need to attempt downgrades
+	if len(e.gen.MissingAPIs()) == 0 {
```
kevinrizza, line 167: Super nit: the caller for this function might be more obvious if this check happened outside the method itself.

pkg/controller/registry/resolver/evolver.go
```diff
@@ -153,7 +162,65 @@ func (e *NamespaceGenerationEvolver) queryForRequiredAPIs() error {
...
+	}
+
+	// smart downgrades are only supported if fewer than 64 updates are resolved at the same time
+	// (this should be all but pathological cases)
+	if len(e.replacements) > 64 {
+		return nil
+	}
+
+	// updates[1][x] = old
+	// updates[0][x] = new
```
kevinrizza, line 178: Definitely non blocking, but I'm curious: This feels a little odd to reason about. Why not just have two slices? The bitwise operation is very nice, but do we save more than the cost of the cognitive load here -- especially given that we are restricting this to a relatively small constraint set?

ecordell: updated to your suggestion

pkg/controller/registry/resolver/evolver.go
ecordell: this is just keeping in line with the rest of the object, which has a series of methods which may or may not affect the generation - adding the check in `Evolve` would make it read weirdly, IMO

Bowenislandsong:
<details>

 totaltestcount: 1
flaketestcount: 1
skippedtestcount: 4
flaketests:
  - classname: End-to-end
    name: '**Subscription creation using existing CSV**'
    counts: 1
    details:
      - count: 1
        error: |4-


            /home/runner/work/operator-lifecycle-manager/operator-lifecycle-manager/test/e2e/subscription_e2e_test.go:70

            	Error Trace:	subscription_e2e_test.go:77
            	            				runner.go:113
            	            				runner.go:64
            	            				it_node.go:26
            	            				spec.go:215
            	            				spec.go:138
            	            				spec_runner.go:200
            	            				spec_runner.go:170
            	            				spec_runner.go:66
            	            				suite.go:62
            	            				ginkgo_dsl.go:226
            	            				ginkgo_dsl.go:214
            	            				e2e_test.go:54

    meandurationsec: 0.055855
skippedtests:
  - classname: End-to-end
    counts: 1
    details: []
    meandurationsec: 0.163795
  - classname: End-to-end
        prometheus objects**'
    counts: 1
    details: []
    meandurationsec: 2.183249
  - classname: End-to-end
    counts: 1
    details: []
    meandurationsec: 0.024478
  - classname: End-to-end
        successfully**'
    counts: 1
    details: []
    meandurationsec: 2.116935

dinhxuanvu:
The PR looks good to me. Plus, I tested it against OCS and it works.
/lgtm

ecordell:
/bugzilla refresh

(retargeted this to 4.5 since the new resolver doesn't have this issue)

sdodson:
Overrode bugzilla/valid-bug after reviewing the upstream bug, QE seems convinced that the implementation is not relevant to 4.6. This needs a fresh lgtm though.

PR #1693, opened 2020-07-30
Upstream PPC64LE: changing arch flags to use GOARCH env variable

aemperador:
Makefile update: changing arch flags to use GOARCH env variable to enable upstreaming for PPC64LE

**Reviewer Checklist**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,6 @@ CONTROLLER_GEN := go run $(MOD_FLAGS) ./vendor/sigs.k8s.io/controller-tools/cmd/
 YQ_INTERNAL := go run $(MOD_FLAGS) ./vendor/github.com/mikefarah/yq/v2/
 KUBEBUILDER_ASSETS := $(or $(or $(KUBEBUILDER_ASSETS),$(dir $(shell command -v kubebuilder))), /usr/local/kubebuilder/bin)
 export KUBEBUILDER_ASSETS
-KUBEBUILDER_ASSETS ?= /usr/local/kubebuilder/bin
 
 ifeq ($(shell go env GOARCH), ppc64le)
 	GOARCH := ppc64le
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,12 @@ YQ_INTERNAL := go run $(MOD_FLAGS) ./vendor/github.com/mikefarah/yq/v2/
 KUBEBUILDER_ASSETS := $(or $(or $(KUBEBUILDER_ASSETS),$(dir $(shell command -v kubebuilder))), /usr/local/kubebuilder/bin)
 export KUBEBUILDER_ASSETS
 KUBEBUILDER_ASSETS ?= /usr/local/kubebuilder/bin
-GOARCH := $(shell go env GOARCH)
+
+ifeq ($(shell go env GOARCH), ppc64le)
+	GOARCH := ppc64le
+else
+	GOARCH := 386
+endif
 
 # ART builds are performed in dist-git, with content (but not commits) copied 
 # from the source repo. Thus at build time if your code is inspecting the local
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -25,6 +25,13 @@ CONTROLLER_GEN := go run $(MOD_FLAGS) ./vendor/sigs.k8s.io/controller-tools/cmd/
 YQ_INTERNAL := go run $(MOD_FLAGS) ./vendor/github.com/mikefarah/yq/v2/
 KUBEBUILDER_ASSETS := $(or $(or $(KUBEBUILDER_ASSETS),$(dir $(shell command -v kubebuilder))), /usr/local/kubebuilder/bin)
 export KUBEBUILDER_ASSETS
+KUBEBUILDER_ASSETS ?= /usr/local/kubebuilder/bin
```
ecordell, line 28: This can be removed, the definitions above should cover this.

Makefile
```diff
@@ -84,15 +91,15 @@ build-coverage: build_cmd=test -c -covermode=count -coverpkg ./pkg/controller/..
 build-coverage: clean $(CMDS)
 
 build-linux: build_cmd=build
-build-linux: arch_flags=GOOS=linux GOARCH=386
+build-linux: arch_flags=GOOS=linux GOARCH=$(GOARCH)
```
ecordell, line 94: These targets are only used for local development / test configuration. The build linux/amd64 binaries on the host, and are then copied into container images which run via kind or minikube.

If you want ppc64le binaries, you should get those just with `make build` with no changes the Makefile. Can you explain more what the goal of these changes is?

Makefile
aemperador: I will remove this. It got pulled in when I rebased.

Makefile
aemperador: when running the targets for run-local and run-console-local those do not get a build for the ppc64le arch. We would like to be able to do local development with ppc64le so upstreaming these local targets is the goal

aemperador: @ecordell is that ok to enable us to move forward?

aemperador: @ecordell wanted to see if this was still on your radar

Makefile
```diff
@@ -26,6 +26,12 @@ YQ_INTERNAL := go run $(MOD_FLAGS) ./vendor/github.com/mikefarah/yq/v2/
 KUBEBUILDER_ASSETS := $(or $(or $(KUBEBUILDER_ASSETS),$(dir $(shell command -v kubebuilder))), /usr/local/kubebuilder/bin)
 export KUBEBUILDER_ASSETS
 
+ifeq ($(shell go env GOARCH), ppc64le)
```
ecordell, line 29: Should this just be `GOARCH := $(shell go env GOARCH)`?

aemperador: so the GOARCH for amd64 devices will be amd64 instead of 386, so I had to use the if to make sure other than ppc64le it'd be 386. this was found when the tests were ran. it seems these prow arches are amd64

aemperador: @ecordell please advise on how you would like this to be handled if that's not satisfactory

aemperador: @ecordell bump :)

njhale:
@aemperador 

It looks like this PR fell off our radar at some point. It's been quite a long time, so I'm going to close this out.

Reviewing the comments, I believe the use of `run-local` has been misinterpreted:

> when running the targets for run-local and run-console-local those do not get a build for the ppc64le arch. 

to rephrase an earlier response, the `run-local` target is only intended to produce binaries for use in a test container that is __exclusively__ linux/amd64; i.e. no matter what your host environment is, the target environment is the same test container that will be run in kind/minikube.

PR #1702, opened 2020-08-03
Remove spurious check for duplicate operators in resolver.

benluddy:
Resolution fails when more than one operator in a catalog share the
same name. Since the registry ListBundles endpoint returns one entry
per channel, this cardinality test is overly aggressive. The resolver
should be able to distinguish operators that share a name as long as
their channel is different (and the solver will generate an error if
its input contains duplicate identifiers).


benluddy:
@ecordell This is expected after the recent ListBundles change, right?

ecordell:
This fixes the immediate issue but can introduce some nondeterminism (the list of bundles will need to be sorted in a reasonable way). This will unblock other PRs and we can address the nondeterminism separately.

kevinrizza:
Do we need a bugzilla at this point to merge this?

Bowenislandsong:
> Do we need a bugzilla at this point to merge this?

not really, at least not according to the label requirements

ecordell:
/override e2e-gcp

e2e-gcp seems to be stalling out, and currently the OLM sanity test is disabled anyway.

PR #1708, opened 2020-08-04
Bug 1856526: check API discovery before attempting to install a CR in a bundle.

exdx:
Support for the gardener feature allowed packaging VPAs in the bundle. If the VPA CRD is not installed in the cluster, the CRD could be included in the bundle. 

Since this is the first time OLM allows a CRD and CR related to that CRD be installed simultaneously, a race condition in the InstallPlan occurs. The CRD is installed first, but if it takes a while to be established and accepted then the CR creation will fail. This fix ensures that the CR will not be created before the corresponding CRD is available via discovery. 

Also adds a unit test to the new resolver to check that an operator can satisfy its own dependencies. 

This race affects gardener objects as well as prometheus objects 

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -102,12 +102,13 @@ type Operator struct {
 	clientAttenuator         *scoped.ClientAttenuator
 	serviceAccountQuerier    *scoped.UserDefinedServiceAccountQuerier
 	bundleUnpacker           bundle.Unpacker
+	dynamicResourceTimeout   time.Duration
 }
 
 type CatalogSourceSyncFunc func(logger *logrus.Entry, in *v1alpha1.CatalogSource) (out *v1alpha1.CatalogSource, continueSync bool, syncError error)
 
 // NewOperator creates a new Catalog Operator.
-func NewOperator(ctx context.Context, kubeconfigPath string, clock utilclock.Clock, logger *logrus.Logger, resync time.Duration, configmapRegistryImage, utilImage string, operatorNamespace string) (*Operator, error) {
+func NewOperator(ctx context.Context, kubeconfigPath string, clock utilclock.Clock, logger *logrus.Logger, resync time.Duration, configmapRegistryImage, utilImage string, operatorNamespace string, dynamicResourceTimeout time.Duration) (*Operator, error) {
 	resyncPeriod := queueinformer.ResyncWithJitter(resync, 0.2)
 	config, err := clientcmd.BuildConfigFromFlags("", kubeconfigPath)
 	if err != nil {
@@ -160,6 +161,7 @@ func NewOperator(ctx context.Context, kubeconfigPath string, clock utilclock.Clo
 		catalogSubscriberIndexer: map[string]cache.Indexer{},
 		serviceAccountQuerier:    scoped.NewUserDefinedServiceAccountQuerier(logger, crClient),
 		clientAttenuator:         scoped.NewClientAttenuator(logger, config, opClient, crClient, dynamicClient),
+		dynamicResourceTimeout:   dynamicResourceTimeout,
 	}
 	op.sources = grpc.NewSourceStore(logger, 10*time.Second, 10*time.Minute, op.syncSourceState)
 	op.reconciler = reconciler.NewRegistryReconcilerFactory(lister, opClient, configmapRegistryImage, op.now)
@@ -1432,6 +1434,11 @@ func transitionInstallPlanState(log *logrus.Logger, transitioner installPlanTran
 			out.Status.SetCondition(v1alpha1.ConditionMet(v1alpha1.InstallPlanInstalled, &now))
 			out.Status.Phase = v1alpha1.InstallPlanPhaseComplete
 		}
+
+		// add resolved condition if no conditions exist - the installplan is still installing (waiting for APIs, etc)
+		if len(out.Status.Conditions) == 0 {
+			out.Status.SetCondition(v1alpha1.ConditionMet(v1alpha1.InstallPlanResolved, &now))
+		}
 		return out, nil
 	default:
 		return out, nil
@@ -1578,9 +1585,9 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
 		}
 
 		switch step.Status {
-		case v1alpha1.StepStatusPresent, v1alpha1.StepStatusCreated, v1alpha1.StepStatusWaitingForAPI:
+		case v1alpha1.StepStatusPresent, v1alpha1.StepStatusCreated:
 			continue
-		case v1alpha1.StepStatusUnknown, v1alpha1.StepStatusNotPresent:
+		case v1alpha1.StepStatusUnknown, v1alpha1.StepStatusNotPresent, v1alpha1.StepStatusWaitingForAPI:
 			manifest, err := r.ManifestForStep(step)
 			if err != nil {
 				return err
@@ -1857,11 +1864,19 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
 					return errorwrap.Wrapf(err, "error decoding %s object to an unstructured object", step.Resource.Name)
 				}
 
-				// Get the resource from the GVK.
+				// Get the GVK from the resource.
 				gvk := unstructuredObject.GroupVersionKind()
+				// Then check if the API for this dynamic resource is available on-cluster, before creating the resource, by querying discovery.
+				// This ensures CRs created via the dynamic client are created after the corresponding CRD is accepted and established on the cluster to avoid potential race conditions.
+				// (Default) 2 minute timeout for the required API to appear in discovery before failing the installplan
+				if timeout(resolvedTimestamp(plan), o.dynamicResourceTimeout) {
+					return fmt.Errorf("required GVK %s not found via discovery: deadline exceeded", gvk.String())
+				}
 				r, err := o.apiresourceFromGVK(gvk)
 				if err != nil {
-					return err
+					o.logger.Infof("failed to find required server resources: %s", err)
+					plan.Status.Plan[i].Status = v1alpha1.StepStatusWaitingForAPI
+					continue
 				}
 
 				// Create the GVR
@@ -2026,6 +2041,8 @@ func getCSVNameSet(plan *v1alpha1.InstallPlan) map[string]struct{} {
 	return csvNameSet
 }
 
+// apiresourceFromGVK looks for the presence of a GVK on the api-server in a non-blocking way.
+// It checks the installplan timestamp to determine whether or not to continue.
 func (o *Operator) apiresourceFromGVK(gvk schema.GroupVersionKind) (metav1.APIResource, error) {
 	logger := o.logger.WithFields(logrus.Fields{
 		"group":   gvk.Group,
@@ -2036,15 +2053,17 @@ func (o *Operator) apiresourceFromGVK(gvk schema.GroupVersionKind) (metav1.APIRe
 	resources, err := o.opClient.KubernetesInterface().Discovery().ServerResourcesForGroupVersion(gvk.GroupVersion().String())
 	if err != nil {
 		logger.WithField("err", err).Info("could not query for GVK in api discovery")
-		return metav1.APIResource{}, err
+		return metav1.APIResource{}, fmt.Errorf("could not query for GVK in api discovery: %s", err)
 	}
 	for _, r := range resources.APIResources {
 		if r.Kind == gvk.Kind {
+			logger.WithField("kind", r.Kind).Infof("found GVK %s in api discovery", gvk.String())
 			return r, nil
 		}
 	}
-	logger.Info("couldn't find GVK in api discovery")
-	return metav1.APIResource{}, olmerrors.GroupVersionKindNotFoundError{gvk.Group, gvk.Version, gvk.Kind}
+
+	logger.Infof("could not find GVK %s in api discovery", gvk.String())
+	return metav1.APIResource{}, fmt.Errorf("could not find GVK %s in api discovery", gvk.String())
 }
 
 const (
@@ -2067,4 +2086,26 @@ var supportedKinds = map[string]struct{}{
 func isSupported(kind string) bool {
 	_, ok := supportedKinds[kind]
 	return ok
+}
+
+func timeout(timestamp time.Time, timeout time.Duration) bool {
+	if timestamp.IsZero() {
+		return false
+	}
+	if time.Now().Sub(timestamp) > timeout {
+		return true
+	}
+	return false
+}
+
+// resolvedTimestamp returns the time at which the installplan had a resolved condition.
+// This condition is representative of the first time the installplan went through a sync.
+// The timestamp is used for timing out the installplan if dynamic resources are missing APIs on the server.
+func resolvedTimestamp(plan *v1alpha1.InstallPlan) time.Time {
+	for _, cond := range plan.Status.Conditions {
+		if cond.Type == v1alpha1.InstallPlanResolved {
+			return cond.LastUpdateTime.Time
+		}
+	}
+	return time.Time{}
 }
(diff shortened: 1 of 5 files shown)


test/e2e/bundle_e2e_test.go
```diff
@@ -165,6 +165,88 @@ var _ = Describe("Installing bundles with new object types", func() {
...
+			packageName = "busybox"
+			channelName = "alpha"
+			subName     = "test-subscription"
+			vpaName     = "busybox-vpa"
+		)
+
+		BeforeEach(func() {
+			const (
+				sourceName = "test-catalog"
+				imageName  = "quay.io/olmtest/single-bundle-index:vpa-race"
```
benluddy, line 181: The test looks good to me, but it's not ideal to take a another dependency on an external catalog image.

pkg/controller/operators/catalog/operator.go
```diff
@@ -1861,6 +1863,20 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
...
 				if err != nil {
+					// Check if the API for this dynamic resource is available on cluster before creating the resource by checking discovery.
+					// Ensures CRs created via the dynamic client are created after the corresponding CRD is accepted and established on the cluster to avoid potential race conditions.
+					if _, ok := err.(olmerrors.GroupVersionKindNotFoundError); ok {
+						// API not yet established on cluster
+						o.logger.Infof("failed to find server resource for GVK %s via discovery", gvk.String())
+						plan.Status.Plan[i].Status = v1alpha1.StepStatusUnknown
+						continue
+					}
+					if k8serrors.IsNotFound(err) {
```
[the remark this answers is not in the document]
exdx: Yeah, I agree -- originally I just wrote my own logic for this check, then scrolled down and saw this function and decided to reuse existing code (to limit the complexity of the PR). But I agree having it return two error types is potentially confusing. 

Yep there's another function that does the same thing haha -- it's just in the olm operator and not in the catalog operator. I'd prefer not to touch ` isGVKRegistered` and look into refactoring `apiresourceFromGVK` after looking at its usages as you suggested. Although since they are essentially the same function I'm wondering if it's worth consolidating them both as part of this PR?

test/e2e/bundle_e2e_test.go
exdx: Agreed -- but unfortunately I don't see another option until our bundle building in tests effort merges

pkg/controller/operators/catalog/operator.go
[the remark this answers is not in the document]
benluddy: I'd like to see at least `apiresourceFromGVK` refactored, because it's directly impacting the way this change is implemented. There don't seem to be too many uses of its twin, so it might be an easy win to unify them now, but that's more of a tangent and I don't feel as strongly about it.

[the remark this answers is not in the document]
exdx: Yep, sounds good -- I looked and the only actual use of `apiresourceFromGVK` function seems to be in a unit test -- and it doesn't make use of the custom error type at all. So my plan is to simply remove the return of the custom error (change it to a simple `fmt.Errorf()` error) in `apiresourceFromGVK` and have the simpler `if err != nil` condition in my code only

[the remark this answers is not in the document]
exdx: `isGVKRegistered` also doesn't seem to actually require the use of the custom-error type at any of its call-sites, but I agree changing it is not entirely within the scope of this PR

exdx:
What I'm wondering now is the scenario where a CR in a bundle requires a CRD that is not provided by another operator or within the bundle itself -- in this case the current code would have that CR step  in the InstallPlan stay in an `Unknown` state forever. Potentially the installplan would get requeued since it's not in a failed state but still in installing I believe. There is some kind of backoff to the queue built-in I think, so maybe over time it would not get requeued as much? 

Just wondering if we don't end up introducing something where an incomplete installation can cause OLM to spin on one particular installplan indefinitely.

pkg/controller/operators/catalog/operator.go
```diff
@@ -1857,11 +1857,15 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
...
-				// Get the resource from the GVK.
+				// Get the GVK from the resource.
+				// Then check if the API for this dynamic resource is available on-cluster, before creating the resource, by querying discovery.
+				// This ensures CRs created via the dynamic client are created after the corresponding CRD is accepted and established on the cluster to avoid potential race conditions.
 				gvk := unstructuredObject.GroupVersionKind()
 				r, err := o.apiresourceFromGVK(gvk)
 				if err != nil {
-					return err
+					o.logger.Infof("failed to find any server resources for GVK %s in api discovery", gvk.String())
+					plan.Status.Plan[i].Status = v1alpha1.StepStatusUnknown
```
exdx, line 1867: I'm not even sure this is necessary - isn't the step in an `Unknown` state at this point anyway? Maybe just the `continue` is enough

exdx:
I've moved the installplan step for the dynamic resource into `StepStatusWaitingForAPI` status in the case where the API is not reachable via discovery -- this makes it more clear that the installplan is waiting for the API for that CR before proceeding. This status is only used here and in the CRD logic so I think it's safe to move it around a little bit here. 

However if the API for the CR is never discoverable (maybe the CRD is just missing) - the installplan will continue to be requeued. If there is an expontential backoff on the requeuing then eventually the installplan will be left in a `Installing` state indefinitely. If we are ok with that then conceptually this is what we are looking for. Could also add another e2e test for this case if its helpful.

pkg/controller/operators/catalog/operator.go
```diff
@@ -2040,11 +2044,12 @@ func (o *Operator) apiresourceFromGVK(gvk schema.GroupVersionKind) (metav1.APIRe
 	}
 	for _, r := range resources.APIResources {
 		if r.Kind == gvk.Kind {
+			logger.WithField("kind", r.Kind).Infof("found GVK %s in api discovery", gvk.String())
 			return r, nil
 		}
 	}
 	logger.Info("couldn't find GVK in api discovery")
-	return metav1.APIResource{}, olmerrors.GroupVersionKindNotFoundError{gvk.Group, gvk.Version, gvk.Kind}
```
awgreene, line 2047: Any reason we're not using the `GroupVersionKindNotFound` error?

pkg/controller/operators/catalog/operator.go
```diff
@@ -2040,11 +2044,12 @@ func (o *Operator) apiresourceFromGVK(gvk schema.GroupVersionKind) (metav1.APIRe
 	}
 	for _, r := range resources.APIResources {
 		if r.Kind == gvk.Kind {
+			logger.WithField("kind", r.Kind).Infof("found GVK %s in api discovery", gvk.String())
```
awgreene, line 2047: :+1:

test/e2e/bundle_e2e_test.go
```diff
@@ -165,6 +165,87 @@ var _ = Describe("Installing bundles with new object types", func() {
 			}).Should(Succeed())
 		})
 	})
+
+	When("A bundle is installed with a CR that is associated with a provided CRD", func() {
```
awgreene, line 169: Does this test always hit the race condition?

pkg/controller/operators/catalog/operator.go
```diff
@@ -2040,11 +2044,12 @@ func (o *Operator) apiresourceFromGVK(gvk schema.GroupVersionKind) (metav1.APIRe
...
 			return r, nil
 		}
 	}
 	logger.Info("couldn't find GVK in api discovery")
-	return metav1.APIResource{}, olmerrors.GroupVersionKindNotFoundError{gvk.Group, gvk.Version, gvk.Kind}
+	return metav1.APIResource{}, fmt.Errorf("could not find GVK %s in api discovery", gvk.String())
```
awgreene, line 2052: IT looks like OLM will keep trying to reconcile the CR if the CRD is not present. What happens if the CR's CRD is not present on cluster and is not included in the installPlan. Should OLM fail the CSV?

pkg/controller/operators/catalog/operator.go
```diff
@@ -1578,9 +1578,9 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
 		}
 
 		switch step.Status {
-		case v1alpha1.StepStatusPresent, v1alpha1.StepStatusCreated, v1alpha1.StepStatusWaitingForAPI:
+		case v1alpha1.StepStatusPresent, v1alpha1.StepStatusCreated:
```
awgreene, line 1581: To make sure that I understand this change, if ` v1alpha1.StepStatusWaitingForAPI` was here, OLM would just skip the step ,correct?

exdx: Yes

pkg/controller/operators/catalog/operator.go
exdx: That's essentially more concern in https://github.com/operator-framework/operator-lifecycle-manager/pull/1708#issuecomment-670140773

How can we tell whether want to fail the CSV or wait longer for the GVK to become available

test/e2e/bundle_e2e_test.go
exdx: yes I believe so

A similar test failed like 99/100 times in the first gardener PR. It depends on the api-server a bit but basically yes it would fail consistently without the code change.

pkg/controller/operators/catalog/operator.go
exdx: @benluddy objected to the fact that this function returns different error types for no reason, and the caller had to be aware of that fact. I generally agree since this custom error type serves no purpose in our code ie we don't look for it

pkg/controller/operators/catalog/operator.go
exdx: we could set a `context.WithTimeout()` set to like 5 minutes?

pkg/controller/operators/catalog/operator.go
awgreene: SGTM

pkg/controller/operators/catalog/operator.go
exdx: Prototyped this out in https://github.com/exdx/operator-lifecycle-manager/pull/3

exdx:
Do we want to have a different timeout for the test binary versus the actual live catalog operator? Also, 2 minutes is fairly arbitrary, idk if there are any feelings about a different timeout.

pkg/controller/operators/catalog/operator.go
```diff
@@ -1857,11 +1859,19 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
 					return errorwrap.Wrapf(err, "error decoding %s object to an unstructured object", step.Resource.Name)
 				}
 
-				// Get the resource from the GVK.
+				// Get the GVK from the resource.
 				gvk := unstructuredObject.GroupVersionKind()
+				// Then check if the API for this dynamic resource is available on-cluster, before creating the resource, by querying discovery.
+				// This ensures CRs created via the dynamic client are created after the corresponding CRD is accepted and established on the cluster to avoid potential race conditions.
+				// (Default) 2 minute timeout for the required API to appear in discovery before failing the installplan
+				if timeout(plan.CreationTimestamp, o.dynamicResourceTimeout) {
```
njhale, line 1867: Should the timeout begin when we first attempt to create the CR (or CRD), rather than the plan itself?

pkg/controller/operators/catalog/operator.go
```diff
@@ -2026,6 +2036,8 @@ func getCSVNameSet(plan *v1alpha1.InstallPlan) map[string]struct{} {
 	return csvNameSet
 }
 
+// apiresourceFromGVK looks for the presence of a GVK on the api-server in a non-blocking way.
+// It checks the installplan timestamp to determine whether or not to continue.
```
njhale, line 2040: nit: out of date comment?

pkg/controller/registry/resolver/resolver_test.go
```diff
@@ -1198,3 +1198,41 @@ func TestSolveOperators_WithoutDeprecated(t *testing.T) {
...
+
+	newSub := newSub(namespace, "packageA", channel, catalog)
+	subs := []*v1alpha1.Subscription{newSub}
+
+	fakeNamespacedOperatorCache := NamespacedOperatorCache{
+		snapshots: map[registry.CatalogKey]*CatalogSnapshot{
+			catalog: {
+				operators: []*Operator{
+					// Default channel is stable in this case
+					genOperator("packageA.v0.0.2", "0.0.2", "", "packageA", channel, catalog.Name, catalog.Namespace, APISet, APISet, nil, channel, false),
```
njhale, line 1219: I wouldn't think you'd list the API as both provided and required, since if it's provided, there's no need for the API to be resolved as a dependency. This is an interesting case though, and to really make sure the resolver doesn't choke on this, we'd need to have another operator in the catalog that provides the same API (but I'm pretty sure that's an invariant). We may even want to enforce that provided and required are disjoint sets.

test/e2e/bundle_e2e_test.go
```diff
@@ -165,6 +165,142 @@ var _ = Describe("Installing bundles with new object types", func() {
...
+					Namespace: testNamespace,
+					Labels:    map[string]string{"olm.catalogSource": sourceName},
+				},
+				Spec: v1alpha1.CatalogSourceSpec{
+					SourceType: v1alpha1.SourceTypeGrpc,
+					Image:      imageName,
+				},
+			}
+
+			source, err := operatorClient.OperatorsV1alpha1().CatalogSources(source.GetNamespace()).Create(context.TODO(), source, metav1.CreateOptions{})
```
njhale, line 204: nit: I recommend using the [controller-runtime client](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/test/e2e/ctx/ctx.go#L78), since it's less verbose.

pkg/controller/registry/resolver/resolver_test.go
benluddy: That may have been a problem with the old stepwise resolver, but allowing one operator to satisfy its own constraints is natural for the current resolution system. Even thought it means that pointless (tautological) constraints are allowed, they are harmless. A special restriction for this case ends up being extra complexity without a real upside.

Full disclosure: I specifically requested the test case to enshrine this behavior as valid.

pkg/controller/operators/catalog/operator.go
exdx: That probably is more accurate. I'll look into using a timestamp on a per-installplan basis

pkg/controller/registry/resolver/resolver_test.go
exdx: Yes, this is a testcase that I added just to see how the new resolver would behave in this case as @benluddy mentioned. Since we don't really want to file a separate bug for an additional test I think it's ok to include this test here.

exdx: The `APISet` is both provided and required on this operator in this catalog as-is I believe

pkg/controller/operators/catalog/operator.go
exdx: Per installplan step

exdx: This is a little tricky to implement though. Simply using a local timestamp in `ExecutePlan()` wouldn't work because it would get GC'd everytime the function exits, and so the timeout would never occur. This timestamp needs to persist somehow between reconciliation cycles. I see two possible options:
* use a `map[*v1alpha1.InstallPlan]time.Time` at the package level to set/lookup timestamps per installplan. This would add package-level state though and the map would grow linearly in the number of processed installplans. Even though I implemented this locally I don't think its a good solution. 
* add an optional field to the InstallPlan, or specifically the `InstallPlan.Status.Step` resource called `Timestamp` that we can set/look at from this portion of the code. 

Does the second option sound correct? Are there any other options I'm missing @njhale?

exdx:
Adding a new `InstallPlanConditionReason` called `InstallPlanReasonWaitingForDiscovery` to have a more accurate basis for the timeout
https://github.com/operator-framework/api/pull/61

exdx:
Need to investigate consistent CI failure (CRD storage version upgrade test fails)

PR #1715, opened 2020-08-07
Bug 1867169: Fix stale content in configmap due to stale bundle image in the node

dinhxuanvu:
If the bundle is already present in the node, during the unpack process
the configmap may potentially use a stale bundle image. The default
setting for ImagePullPolicy is `IfNotPresent` and should be changed to
`Always`.

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/bundle/bundle_unpacker.go
+++ b/pkg/controller/bundle/bundle_unpacker.go
@@ -106,9 +106,10 @@ func (c *ConfigMapUnpacker) job(cmRef *corev1.ObjectReference, bundlePath string
 							},
 						},
 						{
-							Name:    "pull",
-							Image:   bundlePath,
-							Command: []string{"/util/cpb", "/bundle"}, // Copy bundle content to its mount
+							Name:            "pull",
+							Image:           bundlePath,
+							ImagePullPolicy: "Always",
+							Command:         []string{"/util/cpb", "/bundle"}, // Copy bundle content to its mount
 							VolumeMounts: []corev1.VolumeMount{
 								{
 									Name:      "bundle",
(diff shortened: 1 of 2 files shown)


exdx:
This does potentially cause more egress from the cluster but the fix is necessary for the update feature to work correctly. 
Also I assume this will also work in disconnected environments (since all bundle images are hosted internally) and shouldn't break anything else.
/lgtm

pkg/controller/bundle/bundle_unpacker.go
```diff
@@ -106,9 +106,10 @@ func (c *ConfigMapUnpacker) job(cmRef *corev1.ObjectReference, bundlePath string
 							},
 						},
 						{
-							Name:    "pull",
-							Image:   bundlePath,
-							Command: []string{"/util/cpb", "/bundle"}, // Copy bundle content to its mount
+							Name:            "pull",
+							Image:           bundlePath,
+							ImagePullPolicy: "Always",
```
awgreene, line 111: Should we be worried about the consequences of pulling an image in a bundle everytime an operator is deployed? If operators are installed in multiple namespaces this is expensive.

exdx: Yes, this is essentially the tradeoff

dinhxuanvu: I don't think this is such a big deal given the bundle image is just scratch image so it is small. Plus, unless there are multiple new installations of the same operator in multiple namespaces (that is different with MultiNamespace mode), then this scenario doesn't happen. The common pattern of operator installation so far has been pretty much singleton.

PR #1716, opened 2020-08-07
Bug 1860035: Fix SubscriptionConfig NodeSelector field

awgreene:
Problem: OLM's Subscription CRD allows cluster admins to set
nodeSelectors on Operators they deploy. Currently this API is exposed
but performs no action.

Solution: Wire the NodeSelector specified in the SubscriptionConfig into
the pods deployed when installing the operator.

### DIFF
--- a/doc/design/subscription-config.md
+++ b/doc/design/subscription-config.md
@@ -16,7 +16,7 @@ The `env` field defines a list of [Environment Variables](https://kubernetes.io/
 
 Increase log verbosity on an Operator's container that utilizes the `ARGS` variable:
 
-```
+```yaml
 kind: Subscription
 metadata:
   name: prometheus
@@ -39,7 +39,7 @@ The `envFrom` field defines a [list of sources to populate Environment Variables
 
 Inject a license key residing in a Secret to unlock Operator features:
 
-```
+```yaml
 kind: Subscription
 metadata:
   name: my-operator
@@ -68,7 +68,7 @@ The `volumeMounts` field defines a list of [VolumeMounts](https://kubernetes.io/
 
 Mount a ConfigMap as a Volume that contains configuration information that can change default Operator behavior. Modifications to the content of the ConfigMap should appear within the container's VolumeMount.
 
-```
+```yaml
 kind: Subscription
 metadata:
   name: my-operator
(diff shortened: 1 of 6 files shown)


doc/design/subscription-config.md
```diff
@@ -133,3 +133,23 @@ spec:
         memory: "128Mi"
         cpu: "500m"
 ```
+
+### NodeSelector
+
+The `nodeSelector` field defines a [NodeSelector](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/) for the Pod created by OLM.
```
exdx, line 139: nit: this is a bit vague. Could we say "for the pods associated with the operator deployment" instead?

This also future-proofs it a bit where in the future OLM supports more than just deploying operators, something we discussed recently

pkg/controller/operators/olm/overrides/inject.go
```diff
@@ -206,3 +206,17 @@ func InjectResourcesIntoDeployment(podSpec *corev1.PodSpec, resources corev1.Res
 
 	return nil
 }
+
+// InjectNodeSelectorIntoDeployment injects the provided NodeSelector
+// into the container(s) of the given PodSpec.
+//
+// If any Container in PodSpec already defines a NodeSelector it will
+// be overwritten.
+func InjectNodeSelectorIntoDeployment(podSpec *corev1.PodSpec, nodeSelector map[string]string) error {
```
exdx, line 215: nit: this function is called `InjectNodeSelectorIntoDeployment` but injects the nodeselector into a `PodSpec`

maybe calling it `InjectNodeSelectorIntoPodSpec` makes it a little clearer?

awgreene: This makes sense, but none of the other methods in this file distinguish injecting into the deployment or podSpec. I suspect this was either an oversight when the code was written or assuming its fine given that the podspec is part of the deployment. Should we change the distinction in all places or follow convention.

doc/design/subscription-config.md
awgreene: /SGTM

awgreene: Technically this config is applied to all deployments created by OLM. OLM also creates deployments for webhooks, which could be stored in the same deployment as the operator but isn't guaranteed, so I don't know how OLM could only create deployments for operators.

awgreene:
The e2e test was failing because the CSV never reached the suceeded state. The pod associated with the operator was never assigned to a node that had the `nodeSelector` specified in the subscription config because no node had the labels. I have created a new e2e test that checks that the deployment has the nodeSelector instead.

PR #1718, opened 2020-08-10
Bug 1856134: bumping golang.org/x/text to v0.3.3

ankitathomas:
Adding golang.org/x/text as an anonymous top level import and setting version to 0.3.3.

See https://github.com/golang/go/issues/39491

**Reviewer Checklist**


njhale:
why are we taking on this import directly? it looks like the [transitive dependency in master](https://github.com/operator-framework/operator-lifecycle-manager/blob/63a809b03d27e5e00ec14249d629ecd65e53b7aa/go.sum#L855) is already at v0.3.3

benluddy:
Doesn't OLM already depend (transitively) on v0.3.3?

```
$ go list -m all | grep x/text
golang.org/x/text v0.3.3
```

The library fix appears to be present in the vendored copy: https://github.com/operator-framework/operator-lifecycle-manager/blob/master/vendor/golang.org/x/text/encoding/unicode/unicode.go#L371-L377 / https://go-review.googlesource.com/c/text/+/238238/2/encoding/unicode/unicode.go.

PR #1722, opened 2020-08-11
fix(reconciler): always use the current opm to serve a `grpc` type

ecordell:
Uses the `opm` from the cluster to serve the registry content, instead of what happens to be in the index image.

This allows us to ship bugfixes to the registry and have it apply to running indexes. It also enables a migration path for indexes built for older clusters without manual intervention to update the base image.

**Reviewer Checklist**


pkg/controller/registry/reconciler/configmap.go
```diff
@@ -90,9 +91,56 @@ func (s *configMapCatalogSourceDecorator) Service() *v1.Service {
 }
 
 func (s *configMapCatalogSourceDecorator) Pod(image string) *v1.Pod {
-	pod := Pod(s.CatalogSource, "configmap-registry-server", image, s.Labels(), 5, 2)
-	pod.Spec.ServiceAccountName = s.GetName() + ConfigMapServerPostfix
-	pod.Spec.Containers[0].Command = []string{"configmap-server", "-c", s.Spec.ConfigMap, "-n", s.GetNamespace()}
+	pod := &v1.Pod{
```
exdx, line 94: should we update the configmap based catalogs if these store raw manifests and don't require `opm` to unpack bundles IIRC (since they don't support bundles)?

util/cpb/main.go
```diff
@@ -128,3 +161,107 @@ func getMetadata() (m *metadata, err error) {
 
 	return
 }
+
+// copyIndex searches for the index in an image and then copies it to dest
+// it will create intermediate directories for dest if they do not exist
+func copyIndex(dest string) (err error) {
```
exdx, line 167: nit: unnecessary parentheses around the return error

util/cpb/main.go
```diff
@@ -128,3 +161,107 @@ func getMetadata() (m *metadata, err error) {
...
+	if err = out.Sync(); err != nil {
+		return
+	}
+
+	status := reconciler.CopyIndexResult{Digest: fmt.Sprintf("%x", hash.Sum(nil))}
+	b, err := json.Marshal(status)
+	if err != nil {
+		return
+	}
+	return ioutil.WriteFile(reconciler.CopyIndexStatusLocation, b, os.ModePerm)
```
exdx, line 210: nit: since we have a `defer()` above that overwrites the named error return type, and here we are returning directly from the function, there is a chance that a potential error here gets overwritten by the `cerr` in the `defer` func. Could we handle the `io.WriteFile` error here via the typical `if err != nil {...descriptive error}`?

util/cpb/main.go
```diff
@@ -128,3 +161,107 @@ func getMetadata() (m *metadata, err error) {
...
+}
+
+// getIndexPath searches standard paths for a db file and then falls back to walk the fs.
+// it is possible to get a false positive if there is more than one sqlite file in the image,
```
exdx, line 214: does the bundle have an annotation that provides the extract db location? or just its name. 

I guess as you said it doesn't matter, in reality only one sqlite db will be present in the index at any time. Intuitively that makes sense but I don't recall if we have index-level validation that enforces that rule.

util/cpb/main.go
```diff
@@ -128,3 +161,107 @@ func getMetadata() (m *metadata, err error) {
...
+		}
+		if info.IsDir() {
+			return nil
+		}
+		if !info.Mode().IsRegular() {
+			return nil
+		}
+		f, err := os.Open(path)
+		if err != nil {
+			fmt.Println("err opening: ", err)
```
exdx, line 252: nit: should this be `fmt.Printf` instead? In general these prints like they should be `Printf` but since the argument is a string it works as-is as well

ecordell:
/hold 
for more discussion

util/cpb/main.go
njhale: they are necessary since we're using a named return parameter

PR #1723, opened 2020-08-11
Bug 1866444: remove duplicate catalog source pod when polling

exdx:
This PR modifies the `ensureUpdatePod()` portion of the catalog syncing loop to always delete an update pod in the case where polling is enabled. Either:

- The update pod has a different digest than the serving pod, in which case we delete the old serving pod immediately
OR
- The update pod has the same digest as the serving pod, in which case we immediately delete the update pod

This requires requeueing the catalog source sync until the update pod becomes running on the cluster to obtain the imageID of the running container in the pod. 

The result is that two catalog source pods are seen on the cluster for only a minimal amount of time. 

May require a rebase on #1624 

Having two catalog pods on the cluster for one OLM sync cycle was considered bad UX

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -5,7 +5,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"k8s.io/apimachinery/pkg/runtime"
 	"reflect"
 	"sort"
 	"strings"
@@ -25,6 +24,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	utilclock "k8s.io/apimachinery/pkg/util/clock"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
@@ -622,10 +622,17 @@ func (o *Operator) syncRegistryServer(logger *logrus.Entry, in *v1alpha1.Catalog
 	// Registry pod hasn't been created or hasn't been updated since the last configmap update, recreate it
 	logger.Debug("ensuring registry server")
 
-	if err := srcReconciler.EnsureRegistryServer(out); err != nil {
-		syncError = fmt.Errorf("couldn't ensure registry server - %v", err)
-		out.SetError(v1alpha1.CatalogSourceRegistryServerError, syncError)
-		return
+	err = srcReconciler.EnsureRegistryServer(out)
+	if err != nil {
+		if _, ok := err.(reconciler.UpdateNotReadyErr); ok {
+			logger.Debug("requeueing registry server for catalog update check: update pod not yet ready")
+			o.catsrcQueueSet.RequeueAfter(out.GetNamespace(), out.GetName(), reconciler.CatalogPollingRequeuePeriod)
+			return
+		} else {
+			syncError = fmt.Errorf("couldn't ensure registry server - %v", err)
+			out.SetError(v1alpha1.CatalogSourceRegistryServerError, syncError)
+			return
+		}
 	}
 
 	logger.Debug("ensured registry server")
--- a/pkg/controller/registry/reconciler/grpc.go
+++ b/pkg/controller/registry/reconciler/grpc.go
@@ -219,7 +219,15 @@ func (c *GrpcRegistryReconciler) ensureUpdatePod(source grpcCatalogSourceDecorat
 	}
 
 	// check if update pod is ready - if not requeue the sync
+	// if update pod failed (potentially due to a bad catalog image) delete it
 	for _, p := range currentUpdatePods {
+		fail, err := c.podFailed(p)
+		if err != nil {
+			return err
+		}
+		if fail {
+			return fmt.Errorf("update pod %s in a %s state: deleted update pod", p.GetName(), p.Status.Phase)
+		}
 		if !podReady(p) {
 			return UpdateNotReadyErr{catalogName: source.GetName(), podName: p.GetName()}
 		}
@@ -358,4 +366,17 @@ func swapLabels(pod *corev1.Pod, labelKey, updateKey string) *corev1.Pod {
 	pod.Labels[CatalogSourceLabelKey] = labelKey
 	pod.Labels[CatalogSourceUpdateKey] = updateKey
 	return pod
+}
+
+// podFailed checks whether the pod status is in a failed or unknown state, and deletes the pod if so.
+func (c *GrpcRegistryReconciler) podFailed(pod *corev1.Pod) (bool, error) {
+	if pod.Status.Phase == corev1.PodFailed || pod.Status.Phase == corev1.PodUnknown {
+		logrus.WithField("UpdatePod", pod.GetName()).Infof("catalog polling result: update pod %s failed to start", pod.GetName())
+		err := c.removePods([]*corev1.Pod{pod}, pod.GetNamespace())
+		if err != nil {
+			return true, errors.Wrapf(err, "error deleting failed catalog polling pod: %s", pod.GetName())
+		}
+		return true, nil
+	}
+	return false, nil
 }
--- a/pkg/controller/registry/reconciler/grpc.go
+++ b/pkg/controller/registry/reconciler/grpc.go
@@ -3,8 +3,11 @@ package reconciler
 import (
 	"context"
 	"fmt"
-	"github.com/operator-framework/api/pkg/operators/v1alpha1"
+	"time"
+
 	controllerclient "github.com/operator-framework/operator-lifecycle-manager/pkg/lib/controller-runtime/client"
+
+	"github.com/operator-framework/api/pkg/operators/v1alpha1"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/operatorclient"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/operatorlister"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/ownerutil"
@@ -17,11 +20,23 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
-const CatalogSourceUpdateKey = "catalogsource.operators.coreos.com/update"
+const (
+	CatalogSourceUpdateKey      = "catalogsource.operators.coreos.com/update"
+	CatalogPollingRequeuePeriod = 30 * time.Second
+)
 
 // grpcCatalogSourceDecorator wraps CatalogSource to add additional methods
 type grpcCatalogSourceDecorator struct {
 	*v1alpha1.CatalogSource
+}
+
+type UpdateNotReadyErr struct {
+	catalogName string
+	podName     string
+}
+
+func (u UpdateNotReadyErr) Error() string {
+	return fmt.Sprintf("catalog polling: %s not ready for update: update pod %s has not yet reported ready", u.catalogName, u.podName)
 }
 
 func (s *grpcCatalogSourceDecorator) Selector() labels.Selector {
@@ -141,6 +156,9 @@ func (c *GrpcRegistryReconciler) EnsureRegistryServer(catalogSource *v1alpha1.Ca
 		return errors.Wrapf(err, "error ensuring pod: %s", source.Pod().GetName())
 	}
 	if err := c.ensureUpdatePod(source); err != nil {
+		if _, ok := err.(UpdateNotReadyErr); ok {
+			return err
+		}
 		return errors.Wrapf(err, "error ensuring updated catalog source pod: %s", source.Pod().GetName())
 	}
 	if err := c.ensureService(source, overwrite); err != nil {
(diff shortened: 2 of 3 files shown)


pkg/controller/registry/reconciler/grpc.go
```diff
@@ -185,15 +186,12 @@ func (c *GrpcRegistryReconciler) ensureUpdatePod(source grpcCatalogSourceDecorat
 	if !source.Poll() {
 		return nil
 	}
-	var updateFlag bool
 	currentLivePods := c.currentPods(source)
 	currentUpdatePods := c.currentUpdatePods(source)
 
 	for _, updatePod := range currentUpdatePods {
 		if updatePodByDigest(updatePod, currentLivePods) {
```
dinhxuanvu, line 193: We need to rework this loop a bit due to a few potential problems here:
1. There is a possibility that there are multiple update pods that have different digest compared to live pod. You either need to break out (hence the original `break` statement) of the loop when you find one or you risk promoting multiple update pods into serving pods.
2. The line #205 may fail on the second iteration if there are multiple qualified update pods and you delete the living pod on the first iteration because the living pod is gone.

My suggestion here is NOT to delete or promote anything during the loop. Instead, keep track of which update pod is the current candidate and check for the creation timestamp to see which one is created the latest in case there are multiple candidate and update the current candidate to latest candidate as the loop goes on. After the loop is done, promoting the current candidate to living pod, remove it from the update pods slice, deleting previous living pod and then delete all remaining update pods in the slice.

exdx: Hey, so I reworked the PR immediately after I pushed -- sorry about that, I didn't think you would review so fast...

@ecordell said he was comfortable with a synchronous update loop that checks if the image has changed and immediately deletes the old catalog. this makes the logic much easier. see the 2nd commit for the rework.

dinhxuanvu: I thought without some sort of waiting time, the pod digest is often empty as it does take a bit of time for crio to pull the image down. The index image is unfortunately not as small as bundle due to included binaries. We did run into this issue back then during testing due to the issue I just described. Maybe you can run this test on an actual cluster to verify.

exdx: Yes -- in the latest commit I basically put a blocking loop that waits for the container to be have an `imageID` that we are comparing. My takeaway from our meeting was that this was ok given that we want to have 2 catalog source pods on the cluster for a minimal amount of time and are OK with waiting.

I do have a `TODO` to bail out somehow if that update pod never gets scheduled and running successfully though

pkg/controller/registry/reconciler/grpc.go
```diff
@@ -287,17 +282,8 @@ func updatePodByDigest(updatePod *corev1.Pod, servingPods []*corev1.Pod) bool {
...
-	if pod.Status.ContainerStatuses != nil && len(pod.Status.ContainerStatuses) > 0 {
-		podImageID = pod.Status.ContainerStatuses[0].ImageID
-	} else {
-		logrus.WithField("CatalogSource", pod.GetName()).Warn("pod status unknown")
-	}
-
-	logrus.WithField("CatalogSource", pod.GetName()).Infof("ImageID %s", podImageID)
-	return podImageID
+func imageID(pod *corev1.Pod) string {
+	return pod.Status.ContainerStatuses[0].ImageID
```
dinhxuanvu, line 286: I think this was one of issues we did bump into the last time that when the ImageID is not there, this will panic out due to null value.

pkg/controller/registry/reconciler/grpc.go
dinhxuanvu: Maybe @ecordell can clarify but what I got from Evan was we can't wait around and block the queue. So now I'm not sure if I misunderstood what he said or not.

pkg/controller/registry/reconciler/grpc.go
exdx: Right so that's why I call https://github.com/operator-framework/operator-lifecycle-manager/pull/1723/files#diff-8f8ed949ede33539a8c472480f3d4f25R330 first to ensure its non-nil. This makes this function extremely simple, almost unnecessary really

pkg/controller/registry/reconciler/grpc.go
exdx: I think conceptually if we are syncing a catalog source and we see that its ready for an update, holding up the reconciliation loop for a small amount of time to see if an update is actually ready to roll out is a fair tradeoff. The user wants their updated catalog to be in use as soon as possible. It does effect other catalogs in the loop but I think its ok

pkg/controller/registry/reconciler/grpc.go
```diff
@@ -132,15 +133,15 @@ func (c *GrpcRegistryReconciler) EnsureRegistryServer(catalogSource *v1alpha1.Ca
...
+	// recreate the pod if no existing pod is serving the latest imageID
 	overwritePod := overwrite || len(c.currentPodsWithCorrectImage(source)) == 0
 
 	//TODO: if any of these error out, we should write a status back (possibly set RegistryServiceStatus to nil so they get recreated)
 	if err := c.ensurePod(source, overwritePod); err != nil {
 		return errors.Wrapf(err, "error ensuring pod: %s", source.Pod().GetName())
 	}
 	if err := c.ensureUpdatePod(source); err != nil {
-		return errors.Wrapf(err, "error ensuring updated catalog source pod: %s", source.Pod().GetName())
+		return errors.Wrapf(err, "error ensuring imageChanged catalog source pod: %s", source.Pod().GetName())
```
ecordell, line 144: This looks like an unintended find/replace?

pkg/controller/registry/reconciler/grpc.go
```diff
@@ -185,57 +186,51 @@ func (c *GrpcRegistryReconciler) ensureUpdatePod(source grpcCatalogSourceDecorat
...
+	if err != nil {
+		return errors.Wrapf(err, "creating update catalog source pod")
 	}
+	source.SetLastUpdateTime()
 
-	if updateFlag {
-		// Clean up outdated serving pod(s)
-		err := c.removePods(currentLivePods, source.GetNamespace())
+	// TODO bail out at some point if the pod never becomes ready
+	for {
```
ecordell, line 203: We shouldn't spin like this in the sync loop -  

Instead we should check for the exit condition we want (update pod ready) and if it's there, do the rest of the work, and if not, requeue the current object to check again in a bit. This keeps the controller running and processing other objects

pkg/controller/registry/reconciler/grpc.go
exdx: yes, fixed

pkg/controller/registry/reconciler/grpc.go
exdx: Looks like we should requeue until the pod is ready per evan's comment -- will update

pkg/controller/registry/reconciler/grpc.go
njhale: In general, this is the pattern we should use for all eventually consistent work. Since the queue that feeds the reconciliation has a constant throughput (e.g. GOMAXPROC), blocking any one sync can prevent events on other resources from being reconciled. In this case, we could prevent other CatalogSources from being noticed until a pod becomes ready. Even with a timeout on this `for`, we would see that latency reflected down the queue.

exdx: Ok I think I addressed this in the latest commit by adding a requeue based on the type of error the `ensureUpdatePod()` function returns. 

The only concern I have is if the polling interval is extremely small this code will always spin up a new update pod. That needs to be fixed. But I think its close.

exdx: I fixed this by adding a `&& len(currentUpdatePods) == 0` to the creation logic. But this basically makes those really small polling intervals not work

exdx:
Also turned some of the logging down, by removing extraneous logs that were printing image IDs -- these were useful during debugging but make the OLM logs really noisy on 4.6 clusters.

pkg/controller/registry/reconciler/grpc.go
```diff
@@ -180,62 +193,54 @@ func (c *GrpcRegistryReconciler) ensurePod(source grpcCatalogSourceDecorator, ov
...
-			updatePod.Labels[CatalogSourceUpdateKey] = ""
-
-			// Update the update pod to promote it to serving pod
-			_, err := c.OpClient.KubernetesInterface().CoreV1().Pods(source.GetNamespace()).Update(context.TODO(), updatePod, metav1.UpdateOptions{})
-			if err != nil {
-				return errors.Wrapf(err, "error creating new pod: %s", source.Pod().GetName())
-			}
-
-			break
+	if source.Update() && len(currentUpdatePods) == 0 {
```
awgreene, line 205: Is there ever a time where the source is ready for an update but the number of currentUpdatePods is greater than 1? Could this happen if there was a CatSrc with a bad image?

pkg/controller/registry/reconciler/grpc_test.go
```diff
@@ -374,11 +374,6 @@ func TestGetPodImageID(t *testing.T) {
 		pod         *corev1.Pod
 		result      string
 	}{
-		{
-			description: "no pod status: return nothing",
-			pod:         &corev1.Pod{Status: corev1.PodStatus{ContainerStatuses: []corev1.ContainerStatus{}}},
-			result:      "",
-		},
```
awgreene, line 381: This was removed since we moved the input validation from `GetPodImageID` correct?

pkg/controller/operators/catalog/operator.go
```diff
@@ -615,10 +615,16 @@ func (o *Operator) syncRegistryServer(logger *logrus.Entry, in *v1alpha1.Catalog
...
 
-	if err := srcReconciler.EnsureRegistryServer(out); err != nil {
-		syncError = fmt.Errorf("couldn't ensure registry server - %v", err)
-		out.SetError(v1alpha1.CatalogSourceRegistryServerError, syncError)
-		return
+	err = srcReconciler.EnsureRegistryServer(out)
+	if err != nil {
+		if _, ok := err.(reconciler.UpdateNotReadyErr); ok {
+			logger.Debug("requeueing registry server for catalog update check: update pod not yet ready")
+			o.catsrcQueueSet.RequeueAfter(out.GetNamespace(), out.GetName(), 10*time.Second)
```
awgreene, line 622: Why don't we return if the Update is not ready?

pkg/controller/registry/reconciler/grpc.go
```diff
@@ -325,3 +320,28 @@ func (c *GrpcRegistryReconciler) CheckRegistryServer(catalogSource *v1alpha1.Cat
...
+// containerReady returns true if the given Pod has a ready status for its container.
+func containerReady(pod *corev1.Pod) bool {
+	if pod.Status.ContainerStatuses != nil {
+		for _, status := range pod.Status.ContainerStatuses {
+			if status.Ready == true {
+				return true
+			}
+		}
+	}
+	return false
```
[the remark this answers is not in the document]
awgreene: The existing code obviously works with our current implementation of catalogsources, but it would be nice to be bullet proof in the future.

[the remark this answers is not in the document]
exdx: So this pod is the catalog pod, we only spin up one container inside the pod per https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/controller/registry/reconciler/reconciler.go#L88 so the distinction is pretty much a design choice. It's possible in the future to have more than one container in the catalog and this would break, so I can check the suggested way instead.

pkg/controller/registry/reconciler/grpc.go
exdx: This is a scenario we definitely want to avoid and I think we have it in our code currently if the polling interval is really small. Basically with this additional check we only create an update pod if there is not one present on the cluster. 

If the catsrc has a bad image, it never reports ready (so we can't detect the imageID of the update pod). That sounds like an additional case that we should handle  (but do not currently). I can open up another BZ to handle this case.

pkg/controller/registry/reconciler/grpc_test.go
exdx: yes

pkg/controller/operators/catalog/operator.go
exdx: Wasn't sure if we wanted to the code below ie the `err := o.sources.Remove(sourceKey)` to run, but looking at it now I'm pretty sure its better to return early

pkg/controller/registry/reconciler/grpc.go
awgreene: Thanks for opening the BZ, marking this resolved.

pkg/controller/operators/catalog/operator.go
```diff
@@ -622,10 +622,17 @@ func (o *Operator) syncRegistryServer(logger *logrus.Entry, in *v1alpha1.Catalog
...
```
dinhxuanvu, line 629: Is there a reason for this 10 seconds requeue in term of duration? Would 10 seconds be too short? It does take a bit of time to pull the index image as it can be pretty big with all the binaries inside. Not sure if we are concerned about requeueing too much? If that is not a concern, than I guess we can ignore this.

pkg/controller/registry/reconciler/grpc.go
```diff
@@ -181,63 +196,54 @@ func (c *GrpcRegistryReconciler) ensurePod(source grpcCatalogSourceDecorator, ov
...
-			err := c.SSAClient.Apply(context.TODO(), updatePod, func(p *v1.Pod) error {
-				p.Labels[CatalogSourceLabelKey] = source.GetName()
-				p.Labels[CatalogSourceUpdateKey] = ""
-				return nil
-			})()
-
-			if err != nil {
-				return errors.Wrapf(err, "error updating catalog source pod: %s", source.Pod().GetName())
-			}
+	if source.Update() && len(currentUpdatePods) == 0 {
```
dinhxuanvu, line 208: Is this `len(currentUpdatePods)` check necessary? When it is time to upgrade, we should just cleanup all update pod(s) and create a new one.

pkg/controller/registry/reconciler/grpc.go
```diff
@@ -327,3 +324,34 @@ func (c *GrpcRegistryReconciler) CheckRegistryServer(catalogSource *v1alpha1.Cat
...
+		p.Labels[CatalogSourceUpdateKey] = ""
+		return nil
+	})()
+
+	return err
+}
+
+// containerReady returns true if the given Pod has a ready status for its container.
+func containerReady(pod *corev1.Pod) bool {
+	if pod.Status.ContainerStatuses != nil {
```
dinhxuanvu, line 343: I'm a bit confused by this as if there is no container statuses (== nil), should we consider it as NotReady?

dinhxuanvu: I also wonder if we need to check `containerStatuses` for readiness reason. We technically care if the pod is ready. So maybe we should check for `Status.Conditions` for the` type: Ready` as `true`. We only care about container for its `imageID`. This way we won't need to change the code if the pod contains multiple containers in the future.

pkg/controller/operators/catalog/operator.go
dinhxuanvu: Also, maybe make this duration a constant so we can adjust it if needed instead of hardcoding.

pkg/controller/registry/reconciler/grpc.go
exdx: I think this is safer because it ensures only 1 update pod is ever created on the cluster at a time. If the polling interval is really small (say 5 seconds) without this check we would end up spinning up a lot of extra pods and deleting them. Since we don't enforce a minimum polling interval, I think this is good to keep in-place. 

We delete immediately after checking the podID instead, to ensure the extra 1 pod doesn't hang around. The logic moved around a bit from the initial implementation, but I think the code is more intuitive this way.  

`if time to update and no update pod -> create update pod`
`...wait for update pod...`
`check update pod and immediately delete it, or delete the outdated pod`

pkg/controller/registry/reconciler/grpc.go
exdx: We don't care about the pod as much as the container in the pod but I agree this reasoning is easier to follow. Will update.

dinhxuanvu (approved):
/lgtm

dinhxuanvu (approved):
/lgtm

pkg/controller/registry/reconciler/grpc.go
```diff
@@ -17,13 +20,25 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
-const CatalogSourceUpdateKey = "catalogsource.operators.coreos.com/update"
+const (
+	CatalogSourceUpdateKey      = "catalogsource.operators.coreos.com/update"
+	CatalogPollingRequeuePeriod = 30 * time.Second
```
ecordell, line 25: this seems on the long side if waiting for a catalog pod to start - I would think 1-5s would be more typical

pkg/controller/registry/reconciler/grpc.go
```diff
@@ -141,6 +156,9 @@ func (c *GrpcRegistryReconciler) EnsureRegistryServer(catalogSource *v1alpha1.Ca
 		return errors.Wrapf(err, "error ensuring pod: %s", source.Pod().GetName())
 	}
 	if err := c.ensureUpdatePod(source); err != nil {
+		if _, ok := err.(UpdateNotReadyErr); ok {
```
ecordell, line 159: nit: could use errors.Is

pkg/controller/registry/reconciler/grpc.go
```diff
@@ -327,3 +335,48 @@ func (c *GrpcRegistryReconciler) CheckRegistryServer(catalogSource *v1alpha1.Cat
 	healthy = true
 	return
 }
+
+// promoteCatalog swaps the labels on the update pod so that the update pod is now reachable by the catalog service.
+// By updating the catalog on cluster it promotes the update pod to act as the new version of the catalog on-cluster.
+func (c *GrpcRegistryReconciler) promoteCatalog(updatePod *corev1.Pod, key string) error {
+	// Update the update pod to promote it to serving pod via the SSA client
```
ecordell, line 342: should this drop the label from the old pod before adding to the new? there might be a small period of time where the service is routing to both pods

pkg/controller/registry/reconciler/grpc.go
exdx: I had set it originally to 10 seconds but then we upped it because as @dinhxuanvu pointed out it takes time for the new image to be pulled down onto the node. Of course several layers would already be cached and shared between the new/old image but I guess we wanted to stay on the safe side here.

pkg/controller/registry/reconciler/grpc.go
exdx: nice, forgot about this func. much cleaner. will use it next time

pkg/controller/registry/reconciler/grpc.go
exdx: this is a good point - in our existing implementation we promote the update pod and then delete the old serving pod. so there is a period of overlap, and this PR does not change that design. I've thought about the issue myself back when traveling to our F2F. 

right now its
```
update labels on update pod -> (both pods are serving) -> delete serving pod -> only update pod left serving
```
we could have
```
remove labels on serving pod -> (no pod is serving) -> update labels on update pod -> only update pod serving -> delete serving pod
```

The latter is probably a bit cleaner, especially if we can apply both label changes via SSA so they occur basically simultaneously. 

However, from the point of view of the catalog client (the subscription essentially), it doesn't really know when the catalog has been updated, and so a small period where it gets either the new or the old content doesn't really affect it. There's no clear way for the user to know "ok, now my catalog has been updated, i can make a new sub". They just have a rough view of it based on their polling interval. Since the feature is more meant to provide a means for updating existing subs the new operator will get onto the cluster eventually. 

tl;dr i think this could be optimized but I'm not sure it improves much based on the current state of the feature

PR #1725, opened 2020-08-13
Bug 1868497: Fix install plan creation for subscriptions that omit channel.

benluddy:


pkg/controller/registry/resolver/installabletypes.go
```diff
@@ -77,12 +77,5 @@ func (r SubscriptionInstallable) Constraints() []solver.Constraint {
 func (r *SubscriptionInstallable) AddDependency(dependencies []solver.Identifier) {
 	if len(dependencies) > 0 {
 		r.constraints = append(r.constraints, solver.Dependency(dependencies...))
-		// "dependencies" may contain bundles that are not mutually exclusive
-		// currently, the constraint types:
-		// - packagename+version range
-		// - provided gvk
-		// - ANDs of either of the above
-		// all happen to be mutually exclusive, but this won't always be the case
-		r.constraints = append(r.constraints, solver.AtMost(1, dependencies...))
```
exdx, line 86: so what's the reasoning behind this change if we want at most one solution based on the dependencies? If now we are adding several dependencies we expect to only get one answer in a deterministic way, without explicitly specifying it in the SAT solver?

pkg/controller/registry/resolver/step_resolver.go
```diff
@@ -103,21 +98,36 @@ func (r *OperatorStepResolver) ResolveSteps(namespace string, _ SourceQuerier) (
...
+		)
 		sourceInfo := *op.SourceInfo()
-		sourceInfo.DefaultChannel = false
-		_, isAdded := add[sourceInfo]
-		existingSubscription, subExists := subMap[sourceInfo]
+		for _, sub := range subs {
+			if sub.Spec.Package != sourceInfo.Package {
+				continue
+			}
+			if sub.Spec.Channel != "" && sub.Spec.Channel != sourceInfo.Channel {
```
exdx, line 111: is `sub.Spec.Channel != ""` helpful here, if we only care about if the subs are the same? If they are both blank then we know both are pointing to the default channel.

Unless you're accounting for the case where the default channel changed in-between subscriptions, which makes sense.

pkg/controller/registry/resolver/step_resolver.go
```diff
@@ -103,21 +98,36 @@ func (r *OperatorStepResolver) ResolveSteps(namespace string, _ SourceQuerier) (
...
+			subCatalogKey := registry.CatalogKey{
+				Name:      sub.Spec.CatalogSource,
+				Namespace: sub.Spec.CatalogSourceNamespace,
+			}
+			if !subCatalogKey.Empty() && !subCatalogKey.Equal(sourceInfo.Catalog) {
+				continue
+			}
+			alreadyExists, err = r.hasExistingCurrentCSV(sub)
+			if err != nil {
+				return nil, nil, nil, fmt.Errorf("unable to determine whether subscription has a preexisting CSV")
```
exdx, line 123: nit: this isn't relevant to your PR, but it would be nice if we could create another abstraction for the results of this function so we don't return 4 different things, which looks a bit funny. Something like a `ResolutionResult` which encapsulates all the existing types. Just a thought.

pkg/controller/registry/resolver/installabletypes.go
benluddy: Constraints that prevent duplicate package entries and duplicate GVKs are constructed here:
https://github.com/operator-framework/operator-lifecycle-manager/blob/3b086a72f96ff5e5ad44d2b7337cafd60c7965b0/pkg/controller/registry/resolver/resolver.go#L115
That makes the constraint, "no more than one of the operators that satisfies a subscription's constraints may be installed" redundant.

This one was causing problems because the varargs slice wasn't guaranteed to contain only unique entries, which produced constraints like `AtMost(1, {A, A})` that could never cause anything to be installed.

pkg/controller/registry/resolver/step_resolver.go
benluddy: If the subscription doesn't specify a channel, it could be resolved from any channel (although the default channel is preferred). In that case, the operator's source would be the resolved channel.

pkg/controller/registry/resolver/step_resolver.go
benluddy: I agree with you, I would like to see the output eventually be a set of manifests to apply -- so the contract would basically be current state in, resolved state out.

PR #1728, opened 2020-08-19
Bug 1869523: Fix nodeSelector subscription config override

awgreene:
Problem: Empty nodeSelector overrides defined in the subscription are
overwriting existing nodeSelectors in the CSV when empty / not set.

Solution: Only override the nodeSelectors defined in the CSV when the
provided nodeSelector list is not nil.

### DIFF
--- a/pkg/controller/operators/olm/overrides/inject_test.go
+++ b/pkg/controller/operators/olm/overrides/inject_test.go
@@ -746,6 +746,30 @@ func TestInjectNodeSelectorIntoDeployment(t *testing.T) {
 				NodeSelector: map[string]string{"foo": "bar"},
 			},
 		},
+		{
+			// Existing PodSpec is left alone if nodeSelector is nil
+			// Expected: PodSpec is not changed
+			name: "WithNilNodeSelector",
+			podSpec: &corev1.PodSpec{
+				NodeSelector: defaultNodeSelector,
+			},
+			nodeSelector: nil,
+			expected: &corev1.PodSpec{
+				NodeSelector: defaultNodeSelector,
+			},
+		},
+		{
+			// Existing PodSpec is set to an empty map if the nodeSelector is an empty map
+			// Expected: PodSpec nodeSelector is set to an empty map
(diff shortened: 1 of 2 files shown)


pkg/controller/operators/olm/overrides/inject_test.go
```diff
@@ -746,6 +746,30 @@ func TestInjectNodeSelectorIntoDeployment(t *testing.T) {
...
+			podSpec: &corev1.PodSpec{
+				NodeSelector: defaultNodeSelector,
+			},
+			nodeSelector: nil,
+			expected: &corev1.PodSpec{
+				NodeSelector: defaultNodeSelector,
+			},
+		},
+		{
+			// Existing PodSpec is left alone if nodeSelector is an empty map
```
benluddy, line 762: This comment's out of sync with the test expectation. FWIW, I agree with the behavior as it is right now (nil doesn't override, empty does).

awgreene: Oops - addressed in the latest version of the PR.

pkg/controller/operators/olm/overrides/inject_test.go
```diff
@@ -746,6 +746,30 @@ func TestInjectNodeSelectorIntoDeployment(t *testing.T) {
...
+			podSpec: &corev1.PodSpec{
+				NodeSelector: defaultNodeSelector,
+			},
+			nodeSelector: nil,
+			expected: &corev1.PodSpec{
+				NodeSelector: defaultNodeSelector,
+			},
+		},
+		{
+			// Existing PodSpec is set to an empty map if the nodeSelector is an empty map
```
njhale, line 762: how would this be serialized?

njhale: ah, forgetting YAML is a superset of JSON...

```yaml
emptyMap: {}
```

awgreene: I don't believe it can.

PR #1729, opened 2020-08-20
Bug 1870453: Should not compare the digest if cannot the new update pod's imageID

jianzhangbjz:

During the CatalogSource pods polling, OLM compare the digest of the update pod and the current served pods. But, OLM cannot get the imageID of the update pod once it's created. But, OLM still compare the digest, it leads to unnecessary updates operations. As follows:

```console
 330 time="2020-08-19T05:28:57Z" level=warning msg="pod status unknown" CatalogSource=qe-app-registry-v5xz4
 331 time="2020-08-19T05:28:57Z" level=info msg="ImageID " CatalogSource=qe-app-registry-v5xz4
 332 time="2020-08-19T05:28:57Z" level=info msg="Update Pod ImageID " CatalogSource=qe-app-registry-v5xz4
```

When the imageID of the update pod is empty, don't run the updates operations.

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/registry/reconciler/grpc.go
+++ b/pkg/controller/registry/reconciler/grpc.go
@@ -292,7 +292,10 @@ func (c *GrpcRegistryReconciler) createUpdatePod(source grpcCatalogSourceDecorat
 // checkUpdatePodDigest checks update pod to get Image ID and see if it matches the serving (live) pod ImageID
 func imageChanged(updatePod *corev1.Pod, servingPods []*corev1.Pod) bool {
 	updatedCatalogSourcePodImageID := imageID(updatePod)
-
+	if updatedCatalogSourcePodImageID == "" {
+		logrus.WithField("CatalogSource", updatePod.GetName()).Warn("pod status unknown, cannot get the pod's imageID")
+		return false
+	}
 	for _, servingPod := range servingPods {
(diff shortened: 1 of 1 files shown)


dinhxuanvu:
@exdx Dan, this PR impacts your on-going polling PR. May want to see if we can consolidate the effort to avoid potential code conflict here.

exdx:
@jianzhangbjz thanks for your contribution. we are doing some refactoring in this space, take a look at #1733 which is also implementing the same bugfix but in the newer codepath. i would like to add an e2e test to this PR, which is something we can collaborate on and could also use your review.

jianzhangbjz:
@exdx Thanks for your explanation! I get it, feel free to do any operations on this PR, thank you!

jianzhangbjz:
/retitle Bug 1870453: Should not compare the digest if cannot the new update pod's imageID

jianzhangbjz:
@exdx Could you help have a review again? Thanks!

exdx:
With the new logic in place in `podReady()` this should never happen (the digest being `""`) but it's still helpful to have a log here just in case. 

/hold cancel
/lgtm

jianzhangbjz:
Thanks for reviewing!

PR #1732, opened 2020-08-20
Bug 1829735: fix(operator): map top-level csv phase to component conditions 

njhale:

Map the top-level status of a CSV to a single status condition of an `Operator` component reference. A corollary is that a CSV component reference will only have the latest phase recorded as a condition.

The CSV status condition type doesn't match component status conditions, resulting in a bad projection:

```yaml
- lastTransitionTime: "2020-04-30T07:05:13Z"
  lastUpdateTime: "2020-04-30T07:05:13Z"
  message: all requirements found, attempting install
  reason: AllRequirementsMet
  status: ""
  type: ""
```

Additionally, CSV conditions are effectively a log of past phases rather than a summary of the current status - mapping 1:1 produces noisy `Operator` resources.


pkg/controller/operators/decorators/operator.go
```diff
@@ -356,9 +372,23 @@ func (c *Component) Reference() (ref *operatorsv1.RichReference, err error) {
...
+	query := componentConditionsJQ
+	switch c.GroupVersionKind() {
+	case csvGVK:
+		query = csvConditionsJQ
+	}
+	iter := query.Run(c.UnstructuredContent())
+
+	var out interface{}
+	for {
+		v, ok := iter.Next()
```
exdx, line 384: if the query returns no results, so `ok` is false here and we break -- is there an issue passing a blank `interface{}` into the `Decode` call below? Should we check it's length?

exdx (commented):
/lgtm
gojq looks like a very handy library to have in OLM

njhale:
https://github.com/operator-framework/operator-registry/pull/437 has merged and should rectify the test timeouts.

/retest

PR #1733, opened 2020-08-20
Bug 1870367: gracefully handling bad catalog image when polling 

exdx:
Updates to the catalog polling code ensure that only 1 update pod is running on the cluster. When it becomes ready, we check the imageID against the running catalog imageID to see if an update occured.

In the case where a bad catalog source image is pushed, the container would never come up ready, and the check would fail. Since we only allow 1 update pod, this would break the upgrade logic. 

A bad catalog image break the updated polling logic by continually requeuing the catalog. 

Dependent on #1723 

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -5,7 +5,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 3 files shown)


dinhxuanvu:
Looking good. I will add the lgtm after the commit from #1723 is removed.

exdx:
Closing this PR to merge the fix as part of #1723

PR #1745, opened 2020-08-31
Bug 1874938: Set RevisionHistoryLimit per Deployment

awgreene:
Problem: OLM does not set the revision history limit for deployments it
creates for an operator. This value defaults to 10. If OLM has to update
a deployment multiple times when reconciling a CSV, up to 11 replicaSets
may exist.

Solution: Have OLM set the revision history limit on deployments to 1,
keeping only the current and previous replicaSets for a deployment.


awgreene:
/hold
This PR still needs to address an issue where the deployment hash is updated a few times during installation.

PR #1746, opened 2020-09-02
Bug 1851095: Delete subscription metric when an operator is uninstalled

anik120:

When an operator was subscribed to using a Subscription Object,
the subscription_sync_total metric was emitted whenever the Subscription
Object was created/updated/deleted. This PR updates that behaviour to emit
the metric only when the Subscription object is created/updated, and deletes
the time series for that particular subscription when the subscription object
is deleted.

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/catalog/subscription/syncer.go
+++ b/pkg/controller/operators/catalog/subscription/syncer.go
@@ -50,7 +50,7 @@ func (s *subscriptionSyncer) Sync(ctx context.Context, event kubestate.ResourceE
 		return err
 	}
 
-	s.recordMetrics(res)
+	metrics.EmitSubMetric(res)
 
 	logger := s.logger.WithFields(logrus.Fields{
 		"reconciling": fmt.Sprintf("%T", res),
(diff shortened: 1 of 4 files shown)


benluddy:
Does this work if a subscription is renamed?

anik120:
> Does this work if a subscription is renamed?

Unfortunately not. We could introduce the fix to use `uid` as key instead of `name` of the subscription to handle this case from the [original bug fix](https://github.com/operator-framework/operator-lifecycle-manager/pull/1519) and backport it until this release. 

/retest

benluddy:
> > Does this work if a subscription is renamed?
> 
> @benluddy unfortunately not. We could introduce the fix to use `uid` as key instead of `name` of the subscription to handle this case from the [original bug fix](https://github.com/operator-framework/operator-lifecycle-manager/pull/1519) and backport it until this release.
> 
> /retest

OK.

/approve

dinhxuanvu (approved):
/lgtm

PR #1748, opened 2020-09-03
Bug 1875247: Add package check for the channel head querying

jianzhangbjz:

OLM supports multi packages in a channel, so OLM should check the head bundle based on the package. Fix bug: https://bugzilla.redhat.com/show_bug.cgi?id=1875247
Any comments are welcome! Thanks!

Supports multi packages for the channel head querying

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/registry/resolver/resolver.go
+++ b/pkg/controller/registry/resolver/resolver.go
@@ -555,11 +555,11 @@ func (r *SatResolver) sortChannel(bundles []*Operator) ([]*Operator, error) {
 	}
 
 	if len(headCandidates) > 1 {
-		errInfo := "found more than one head for channel: "
+		var names []string
 		for _, v := range headCandidates {
-			errInfo = errInfo + fmt.Sprintf("%s.%s-", v.sourceInfo.Channel, v.sourceInfo.Package)
+			names = append(names, v.Identifier())
 		}
-		return nil, fmt.Errorf(fmt.Sprintf("%s, please check the `replaces`/`skipRange` fields of the operator bundles", errInfo))
+		return nil, fmt.Errorf("found multiple channel heads: %v, please check the `replaces`/`skipRange` fields of the operator bundles", names)
 
 	} else if len(headCandidates) < 1 {
 		return nil, fmt.Errorf("head of channel not found")
--- a/pkg/controller/registry/resolver/resolver.go
+++ b/pkg/controller/registry/resolver/resolver.go
@@ -554,9 +554,15 @@ func (r *SatResolver) sortChannel(bundles []*Operator) ([]*Operator, error) {
 		}
 	}
 
-	if len(headCandidates) != 1 {
-		// TODO: more context in error
-		return nil, fmt.Errorf("found more than one head for channel")
+	if len(headCandidates) > 1 {
+		errInfo := "found more than one head for channel: "
+		for _, v := range headCandidates {
+			errInfo = errInfo + fmt.Sprintf("%s.%s-", v.sourceInfo.Channel, v.sourceInfo.Package)
+		}
+		return nil, fmt.Errorf(fmt.Sprintf("%s, please check the `replaces`/`skipRange` fields of the operator bundles", errInfo))
+
+	} else if len(headCandidates) < 1 {
+		return nil, fmt.Errorf("head of channel not found")
 	}
 
 	head := headCandidates[0]


pkg/controller/registry/resolver/resolver.go
```diff
@@ -548,20 +548,32 @@ func (r *SatResolver) sortChannel(bundles []*Operator) ([]*Operator, error) {
...
+			if v.sourceInfo.Package != p {
+				p = v.sourceInfo.Package
+			} else {
+				// if find multi heads in the same package, return error
+				return nil, fmt.Errorf("found more than one head for package: %s, please check the `replaces` or `skipRange` fields of the operator bundles", p)
+			}
+		}
+
+	} else if len(headCandidates) < 1 {
+		return nil, fmt.Errorf("Not found the head for channel")
```
kevinrizza, line 563: nit: can this error message be `head of channel not found`?

pkg/controller/registry/resolver/resolver.go
```diff
@@ -548,20 +548,32 @@ func (r *SatResolver) sortChannel(bundles []*Operator) ([]*Operator, error) {
...
 	}
 
-	head := headCandidates[0]
-	current := head
-	for {
-		channel = append(channel, current)
-		next, ok := replaces[current]
-		if !ok {
-			break
+	for _, v := range headCandidates {
```
kevinrizza, line 566: Is this change just for the case where `headCandidates == nil`? Is it necessary since above we are just already returning if that's the case?

pkg/controller/registry/resolver/resolver.go
jianzhangbjz: Sure, thanks!

pkg/controller/registry/resolver/resolver.go
jianzhangbjz: Not really, this change is for the situation of the `len(headCandidates) > 1`.  If there are multi packages in a channel, there will multi `headCandidates`, return all of them here.

jianzhangbjz:
/retitle Bug 1875247: Add package check for the channel head querying

pkg/controller/registry/resolver/resolver.go
```diff
@@ -554,27 +554,32 @@ func (r *SatResolver) sortChannel(bundles []*Operator) ([]*Operator, error) {
...
-	if len(headCandidates) != 1 {
-		// TODO: more context in error
-		return nil, fmt.Errorf("found more than one head for channel")
+	if len(headCandidates) > 1 {
+		var p string
+		for _, v := range headCandidates {
+			if v.sourceInfo.Package != p {
+				p = v.sourceInfo.Package
+			} else {
+				// if find multi heads in the same package, return error
```
benluddy, line 563: This would only fire if two _consecutive_ candidates were from the same package (`[A A B]` vs. `[A B A]`), which I don't think is the intention.

benluddy (changes requested):
Thanks for opening this PR.

I don't think we should be checking at runtime for a condition that should never occur -- that is, the arguments to `sortChannel` should all be in the same package and channel.

Improving the errors that are returned when there are more or fewer than 1 head candidates would be great, though. Do you mind updating this change so that it only affects the errors that are returned?

ecordell:
I think most of this was addressed with #1735, do we want to keep this open to address error messages / etc?

pkg/controller/registry/resolver/resolver.go
jianzhangbjz: @benluddy Thanks for your review and suggestions! I agreed. Update the error logs.

jianzhangbjz:
@ecordell @benluddy Thanks! I removed other parts and updated the error logs output.

pkg/controller/registry/resolver/resolver.go
```diff
@@ -554,9 +554,15 @@ func (r *SatResolver) sortChannel(bundles []*Operator) ([]*Operator, error) {
...
 
-	if len(headCandidates) != 1 {
-		// TODO: more context in error
-		return nil, fmt.Errorf("found more than one head for channel")
+	if len(headCandidates) > 1 {
+		errInfo := "found more than one head for channel: "
+		for _, v := range headCandidates {
+			errInfo = errInfo + fmt.Sprintf("%s.%s-", v.sourceInfo.Channel, v.sourceInfo.Package)
+		}
+		return nil, fmt.Errorf(fmt.Sprintf("%s, please check the `replaces`/`skipRange` fields of the operator bundles", errInfo))
```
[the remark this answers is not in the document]
jianzhangbjz: Ok, got it. I misunderstood `v.sourceInfo.Channel` and `v.sourceInfo.Package`. I thought they provides the names.
```console
entry_id    channel_na  package_name                    operatorbundle_name             replaces    depth     
----------  ----------  ------------------------------  ------------------------------  ----------  ----------
...     
9           quay-v3.3   quay-bridge-operator            quay-bridge-operator.v3.3.1     10          0         
10          quay-v3.3   quay-bridge-operator            quay-bridge-operator.v3.3.0                 1         
15          quay-v3.3   quay-operator                   red-hat-quay.v3.3.1             16          0         
16          quay-v3.3   quay-operator                   red-hat-quay.v3.3.0                         1         
17          quay-v3.3   container-security-operator     container-security-operator.v3  18          0         
18          quay-v3.3   container-security-operator     container-security-operator.v3              1 
```

benluddy:
/retest

Looks good to me. Thanks for the contribution! This is just waiting for a second review and for the master branch to unfreeze (should happen in about a week).

ecordell (approved):
/lgtm

PR #1755, opened 2020-09-10
Bug 1877835: update catalog image pull policy

exdx:
Set catalog image pull policy to Always when the image is not digest-based

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/registry/reconciler/reconciler.go
+++ b/pkg/controller/registry/reconciler/reconciler.go
@@ -2,6 +2,8 @@
 package reconciler
 
 import (
+	"strings"
+
 	"github.com/operator-framework/api/pkg/operators/v1alpha1"
 	controllerclient "github.com/operator-framework/operator-lifecycle-manager/pkg/lib/controller-runtime/client"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/operatorclient"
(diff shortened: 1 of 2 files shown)


pkg/controller/registry/reconciler/reconciler.go
```diff
@@ -148,3 +152,17 @@ func Pod(source *v1alpha1.CatalogSource, name string, image string, labels map[s
 	}
 	return pod
 }
+
+// digest returns true when the image name contains a valid sha-256 digest.
+// for example, true if the image is quay.io/operator-framework/olm@sha256:b9d011c0fbfb65b387904f8fafc47ee1a9479d28d395473341288ee126ed993b
+func digest(image string) bool {
```
benluddy, line 158: You could use https://godoc.org/github.com/docker/distribution/reference for dealing with image refs, and type assert on the Digested interface.

exdx: Thanks, will take a look. I suppose this check will be more robust this way, which makes up for adding a dependency

ecordell: technically algorithms other than sha256 can be used

based on the [grammar](https://github.com/docker/distribution/blob/749f6afb4572201e3c37325d0ffedb6f32be8950/reference/regexp.go#L68-L70) it looks like scanning for `@` would be enough to know that there's a digest in the image.

exdx: I was wondering if that was the case, I only found examples of sha256 but its good to know others are also valid. 

do we simply want to reduce the check to if `@` is present in the image name?

PR #1764, opened 2020-09-17
[release-4.5] Bug 1878307: shorten catalog sync interval if polling is enabled

exdx:
**Description of the change:**
Fix of #1759 

**Motivation for the change:**

**Reviewer Checklist**
- [ ] Implementation matches the proposed design, or proposal is updated to match implementation
- [ ] Sufficient unit test coverage 
- [ ] Sufficient end-to-end test coverage
- [ ] Docs updated or added to `/docs` 
- [ ] Commit messages sensible and descriptive

PR #1765, opened 2020-09-17
[release-4.5] Bug 1878297: remove duplicate catalog source pod when polling

exdx:
Fix of #1758 

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -11,8 +11,6 @@ import (
 	"sync"
 	"time"
 
-	"k8s.io/apimachinery/pkg/runtime"
-
 	errorwrap "github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"google.golang.org/grpc/connectivity"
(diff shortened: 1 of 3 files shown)


dinhxuanvu (approved):
/lgtm

markmc:
(patch manager) this seems like quite a large and risky change for a relatively low-impact issue. What am I missing?

exdx:
The following 4.5 backports should be brought in if we want to support better catalog polling UX for customers in 4.5.z. Polling will continue to work without these fixes but will not work as effectively. 

* #1760 (merged)
* #1764 (merged) 
* #1780 
* This PR #1765 


Just depends how much we expect customers to use this feature in 4.5, versus in 4.6 where catalog polling is clearly being used as the default red hat operator catalogs will ship with polling enabled.

exdx:
Also worth pointing out that all the fixes work together to provide better UX, we can't really pick and choose which ones we want. In particular #1780 unlocks the rest of the fixes to work as intended.

exdx:
@operator-framework/patch-managers 

Catalog polling is not going to work properly in 4.5.z for Red Hat catalogs as well as for the IBM teams that are publishing catalogs for their customers without this fix going in.

If we don't backport, shorter polling intervals won't work correctly (minimum will always be 15 minutes), there will always be multiple catalog pods running on the cluster per catalog (only one serves traffic), and edge cases where the updated catalog fails to start will not be handled well. So polling will still work in the base case, but the UX will be worse. As customers move more to the new index-based catalog format and adopt polling in their existing clusters this can be an issue. 

cc @kevinrizza

exdx:
Also in terms of risk all the changes in this PR are to one file that deals exclusively with polling logic. Catalog polling is an opt-in feature and these changes do not affect any other behavioral aspects of OLM.

PR #1781, opened 2020-09-28
Bug 1881222: Annotate CSVs with the properties used during dependency resolution.

benluddy:
This ensures that property information is not lost after an operator
is installed, and that the properties of an operator are the same
before and after installation. Preserving properties also allows
installed operators to satisfy dependencies on properties that cannot
be inferred from a ClusterServiceVersion spec, and is a step toward
unifying installed and available operators from the perspective of
resolution.

**Reviewer Checklist**


pkg/controller/registry/resolver/cache.go
```diff
@@ -436,14 +436,27 @@ func WithChannel(channel string) OperatorPredicate {
 
 func WithPackage(pkg string) OperatorPredicate {
 	return func(o *Operator) bool {
+		for _, p := range o.Properties() {
```
ecordell, line 439: Shouldn't this be in `NewOperator*FromCSV`?

pkg/controller/registry/resolver/resolver.go
```diff
@@ -397,6 +408,14 @@ func (r *SatResolver) newSnapshotForNamespace(namespace string, subs []*v1alpha1
 		installables = append(installables, &i)
 	}
 
+	if len(csvsMissingProperties) > 0 {
+		names := make([]string, len(csvsMissingProperties))
+		for i, csv := range csvsMissingProperties {
+			names[i] = csv.GetName()
+		}
+		r.log.Infof("considered csvs without properties annotation during resolution: %v", names)
```
exdx, line 416: I'm wondering if it makes more sense to emit this log per CSV, so its more human readable.

For example, instead of 
`considered csvs without properties annotation during resolution: CSVA, CSVB, CSVC` 
we have 
`considered CSVA without properties annotation during resolution` 
`considered CSVB without properties annotation during resolution` 
`considered CSVC without properties annotation during resolution` 

WDYT?

benluddy: I am worried about the noise level. Under this change, it would be normal/harmless for users on pre-existing clusters to see this message on every resolution cycle. What might eventually make sense is to persist the inferred annotations up-front, so there would be a single message per inferred CSV. I just didn't want to do that until I was confident that there was nothing else we wanted to infer from an unannotated CSV.

exdx: ok, makes sense

pkg/controller/registry/resolver/cache.go
njhale: If an `Operator` can have more than one `PackageType` property, then the location here makes sense; currently, we don't return more than one package from `o.Package()`.

benluddy: I didn't move this because I hope that we will be externalizing constraint filtering soon :tm: , and didn't want to touch `struct Operator` in the meantime.

benluddy:
> @benluddy: The following test **failed**, say `/retest` to rerun all failed tests:
> Test name 	Commit 	Details 	Rerun command
> ci/prow/e2e-aws-olm 	[bcf0ace](https://github.com/operator-framework/operator-lifecycle-manager/commit/bcf0acee7715a9d4806fc8a931be95d7a15eb054) 	[link](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/operator-framework_operator-lifecycle-manager/1781/pull-ci-operator-framework-operator-lifecycle-manager-master-e2e-aws-olm/1311768040697761792) 	`/test e2e-aws-olm`
> 
> [Full PR test history](https://prow.ci.openshift.org/pr-history?org=operator-framework&repo=operator-lifecycle-manager&pr=1781). [Your PR dashboard](https://prow.ci.openshift.org/pr?query=is:pr+state:open+author:benluddy).

/hold

This looks like a problem with the index image fixture. It was last pushed July 22, and it's exhibiting a bug in its ListBundles response that was fixed a month later in https://github.com/operator-framework/operator-registry/pull/417.

PR #1797, opened 2020-10-06
Bug 1885398: Retrieve CA from conversion webhooks for CA Hash

awgreene:
Problem: If a CSV defines a Conversion Webhook but does not include an
API Service or an Admission Webhook the CSV will cycle through the
Pending, ReadyToInstall, and Installing phases indefinitely.

Cause: When deploying an operator with an API Service, Conversion
Webhook, or an Admission Webhook, OLM should retrieve the CA from an
existing resource in order to calculate a CA Hash annotation. This
annotation influences a Deployment Hash that OLM relies on to confirm
that the deployment is installed correctly. OLM currently does not
retrieve the CA from Conversion Webhooks, resulting in a bad
Deployment Hash which causes OLM to attempt to reinstall the CSV.

Solution: Update OLM so it will use the existing Conversion Webhook to
retrieve that value of the CA and correctly calculate the Deployment
Hash.

### DIFF
--- a/pkg/controller/operators/olm/operator_test.go
+++ b/pkg/controller/operators/olm/operator_test.go
@@ -21,6 +21,7 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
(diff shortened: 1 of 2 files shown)


pkg/controller/operators/olm/operator_test.go
```diff
@@ -3091,6 +3092,74 @@ func TestTransitionCSV(t *testing.T) {
...
+	type expected struct {
+		caBundle []byte
+		err      error
+	}
+	tests := []struct {
+		name     string
+		initial  initial
+		expected expected
+	}{
+		{
```
kevinrizza, line 3112: Do we want to have a test case here for a case where more than just a conversion webhook is included? Or error cases?

Otherwise this lgtm

awgreene: Great idea - I added cases for validating and mutating webhooks and errors getting the CA.

PR #1803, opened 2020-10-08
Bug 1885403: Improve transitionCSVState error logs

awgreene:
Problem: OLM has received complaints that a while transitioning a CSV
from state to state, OLM suppresses a number of helpful error logs.
Currently, OLM surfaces errors in the reconciler that can be addressed
with additional syncs. Errors that cannot be addressed with additional
syncs should be logged.

Solution: Log errors that are not surfaced in OLM logs via the
reconciler.

### DIFF
--- a/pkg/controller/operators/olm/operator.go
+++ b/pkg/controller/operators/olm/operator.go
@@ -1278,6 +1278,7 @@ func (a *Operator) operatorGroupForCSV(csv *v1alpha1.ClusterServiceVersion, logg
 }
 
 // transitionCSVState moves the CSV status state machine along based on the current value and the current cluster state.
+// SyncError should be returned when an additional reconcile of the CSV might fix the issue.
 func (a *Operator) transitionCSVState(in v1alpha1.ClusterServiceVersion) (out *v1alpha1.ClusterServiceVersion, syncError error) {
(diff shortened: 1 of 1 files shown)


hasbro17 (approved):
/lgtm

pkg/controller/operators/olm/operator.go
```diff
@@ -1327,7 +1328,6 @@ func (a *Operator) transitionCSVState(in v1alpha1.ClusterServiceVersion) (out *v
 
 	modeSet, err := v1alpha1.NewInstallModeSet(out.Spec.InstallModes)
 	if err != nil {
-		syncError = err
 		logger.WithError(err).Warn("csv has invalid installmodes")
 		out.SetPhaseWithEventIfChanged(v1alpha1.CSVPhaseFailed, v1alpha1.CSVReasonInvalidInstallModes, syncError.Error(), now, a.recorder)
```
ankitathomas, line 1332: If syncError is being removed here, shouldn't this be changed to err?

awgreene: Yes - great catch.

awgreene: Resolved in the latest version of the PR!

ankitathomas (approved):

PR #1815, opened 2020-10-15
WIP: prevent no-op hotlooping on Operators

sjenning:
xref https://bugzilla.redhat.com/show_bug.cgi?id=1888073

@njhale 

General idea is to cache the `resourceVersion` of the `Operator` the last time we updated it.  If a component resource changes, unset the cached `resourceVersion` for force a reconcile.  If the `Operator` itself changes (which might have been caused by us updating it), check the last `resourceVersion` we observed for it.  If it is the same, the reconcile was caused by our own update and we can bail out, breaking the reconciliation hot loop.

### DIFF
--- a/pkg/controller/operators/operator_controller.go
+++ b/pkg/controller/operators/operator_controller.go
@@ -45,11 +45,11 @@ type OperatorReconciler struct {
 	client.Client
 
 	log     logr.Logger
-	mu      sync.RWMutex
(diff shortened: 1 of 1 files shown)


sjenning:
opened https://github.com/operator-framework/operator-lifecycle-manager/pull/1816 for `master`.  We can review there and pick back from that PR.

PR #1816, opened 2020-10-15
Bug 1888073: prevent no-op hotlooping on Operators

sjenning:
xref https://bugzilla.redhat.com/show_bug.cgi?id=1888073

The `Operator` Reconciler does not detect when a reconciliation is triggered by its own Update and thus hotloops until it gets lucky enough to create the exact same rendering of the `Operator` two times in a row.  For larger `Operator` resources, this is a unlikely and will hotloop forever.

This hotloop creates a lot of strain on the kube-apiserver and etcd as well, since some `Operator` resources are very large.  I observe about 500m of additional CPU cores usage per master in addition to 1 core of CPU used directly by `olm-operator` hotlooping. `olm-operator` also generates 1-2MB/s of network traffic in this state.

Before patch with ACM Operator installed

olm-operator

After patch.  The usage spike is during ACM installation, but it falls back down after installation is compelete, as expected.

olm-operator-patched

@njhale @derekwaynecarr @dinhxuanvu

### DIFF
--- a/pkg/controller/operators/operator_controller.go
+++ b/pkg/controller/operators/operator_controller.go
@@ -45,11 +45,11 @@ type OperatorReconciler struct {
 	client.Client
 
 	log     logr.Logger
-	mu      sync.RWMutex
 	factory decorators.OperatorFactory
 
-	// operators contains the names of Operators the OperatorReconciler has observed exist.
-	operators map[types.NamespacedName]struct{}
+	// last observed resourceVersion for known Operators
+	lastResourceVersion map[types.NamespacedName]string
+	mu                  sync.RWMutex
 }
 
 // +kubebuilder:rbac:groups=operators.coreos.com,resources=operators,verbs=create;update;patch;delete
@@ -101,9 +101,9 @@ func NewOperatorReconciler(cli client.Client, log logr.Logger, scheme *runtime.S
 	return &OperatorReconciler{
 		Client: cli,
 
-		log:       log,
-		factory:   factory,
-		operators: map[types.NamespacedName]struct{}{},
+		log:                 log,
+		factory:             factory,
+		lastResourceVersion: map[types.NamespacedName]string{},
 	}, nil
 }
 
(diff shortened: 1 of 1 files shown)


sjenning:
/hold

Still some things to work out. It isn't recreating the Operator when it is deleted manually.  And the unit test failure I think is due to not requeuing on failure.

pkg/controller/operators/operator_controller.go
```diff
@@ -144,11 +150,19 @@ func (r *OperatorReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 
 	}
 
-	if err := r.Status().Update(ctx, operator.Operator); err != nil {
-		log.Error(err, "Could not update Operator status")
-		return ctrl.Result{}, err
+	if create {
+		if err := r.Create(context.Background(), operator); err != nil && !apierrors.IsAlreadyExists(err) {
```
njhale, line 154:
```suggestion
		if err := r.Create(context.Background(), operator.Operator); err != nil && !apierrors.IsAlreadyExists(err) {
```

pkg/controller/operators/operator_controller.go
```diff
@@ -144,11 +150,19 @@ func (r *OperatorReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+			r.log.Error(err, "Could not create Operator", "operator", name)
```
njhale, line 155: Is there a reason we don't want to return an error here and requeue? Any error at this point should be transient, and the work queue should rate limit to prevent hot loops.

pkg/controller/operators/operator_controller.go
```diff
@@ -144,11 +150,19 @@ func (r *OperatorReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+		}
+	} else {
+		if err := r.Status().Update(ctx, operator.Operator); err != nil {
+			log.Error(err, "Could not update Operator status")
+			return ctrl.Result{}, err
+		}
 	}
 
+	r.setLastResourceVersion(req.NamespacedName, operator.Operator.ResourceVersion)
```
njhale, line 164: are we sure that the Create and Status.Update() methods bump the ResourceVersion on the resource argument?

njhale, line 164:
```suggestion
	r.setLastResourceVersion(req.NamespacedName, operator.GetResourceVersion())
```

pkg/controller/operators/operator_controller.go
njhale: Note: this change makes the patch pass e2e and unit tests

pkg/controller/operators/operator_controller.go
sjenning: Yes, I checked this.  That is why Create/Update take pointers to the resource so they can return the resource after the server has processed it.

pkg/controller/operators/operator_controller.go
sjenning: Yes, we do want to requeue.  Updated.

sjenning:
`Operator` is now recreating on delete and I think the unit test should be fixed now as well
/hold cancel

pkg/controller/operators/operator_controller.go
```diff
@@ -115,40 +115,55 @@ func (r *OperatorReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+		}
 	}
-	r.observe(req.NamespacedName)
 
 	// Wrap with convenience decorator
 	operator, err := r.factory.NewOperator(in)
 	if err != nil {
 		log.Error(err, "Could not wrap Operator with convenience decorator")
-		return reconcile.Result{}, nil
+		return reconcile.Result{Requeue: true}, nil
```
njhale, line 144: I don't think an error like this is transient; more likely it would be a programming error. If that's the case, wouldn't requeuing just congest the queue with something that would never reconcile correctly?

related: if we agree that an error here would mean the program is incorrect, then is a panic more appropriate?

njhale (commented):
Wow, awesome work! Thanks again for finding this and patching it so quickly!

I have one last comment before I sign off. Let me know what you think.

njhale:
looks like there's some flakiness in the unit tests

/hold

sjenning:
`e2e-aws-olm` seems to have a <50% pass rate over all PRs

/test e2e-aws-olm

sjenning:
The `End-to-end: CSV update with owned API service` e2e test is very flaky
/retest

sjenning:
clean run :tada:

pkg/controller/operators/operator_controller.go
ecordell: this function (currently) can only error out when `in` is nil, which will never be the case here.

ecordell (approved):
/lgtm

Issue #1774, opened 2020-09-21
[RFE] Provide procedure for hot-fixing an operator


cdjohnson:
## Feature Request
Give support a way to patch an operator for individual customers without requiring a new version of the operator.

**Is your feature request related to a problem? Please describe.**
Many times, customers will run into a problem with the operator or operand and support must enable diagnostic patches, or provide replacement images or pod specifications.  There is no clear path to support this today other than to publish a new Catalog Index Image with a new Bundle version, which is very expensive and time consuming to deploy.  It is also visible to all customers.

**Describe the solution you'd like**
A simple approach is to allow the customer to patch a deployed CSV file.

This means that OLM needs to know that it can't reconcile the CSV.

To patch:
1.  Modify the subscription to Manual install mode.
2.  Modify the CSV (pod spec, related images, etc) per the support team's instructions
3.  Add metadata to the CSV such that the customer remembers that the CSV has been patched.

To move to the official fix:
1.  Approve a new install plan, which rebuilds the CSV from the Catalog, overlaying the changes.

**Out of scope:**
1.  Provide a generalized template option to continually patch/merge changes into the CSV during upgrades.

ecordell:
Have you run into issues modifying a CSV directly? Any changes made there should be synced to the deployment correctly. I believe the workflow you're looking for is already supported, but I'd like to hear if there's something you've tried that didn't work.

cdjohnson:
We think that this works today.  We just want "official" support for this.  The concern is that we're using an undocumented side effect of OLM's lack of a reconciler.

kevinrizza:
Result of this issue being resolved is documentation and a test describing current behavior. For the future, we should create a new issue (as an RFE) that removes this as the default behavior.

jeyaramashok:
I have created a PR https://github.com/operator-framework/operator-lifecycle-manager/pull/1817  to add an e2e test case , which covers the hotfix scenario described in this issue

cdjohnson:
@jeyaramashok Can we also update the OLM Documentation?

For example:  Add a new topic after this one:

https://olm.operatorframework.io/docs/tasks/install-operator-with-olm/

Called:  Patching an Operator

PR #1817, opened 2020-10-17
test: patch existing CSV for a hotfix

jeyaramashok:

This is a e2e test case for hot fixing a CSV. In this scenario an existing CSV is patched directly to update the deployment spec.

hot fix scenario described in https://github.com/operator-framework/operator-lifecycle-manager/issues/1774

**Reviewer Checklist**

### DIFF
--- a/test/e2e/csv_e2e_test.go
+++ b/test/e2e/csv_e2e_test.go
@@ -2915,7 +2915,144 @@ var _ = Describe("ClusterServiceVersion", func() {
 		err = waitForDeploymentToDelete(c, strategy.DeploymentSpecs[0].Name)
 		Expect(err).ShouldNot(HaveOccurred())
 	})
+	It("update deployment spec in an existing CSV for a hotfix", func() {
 
+		c := newKubeClient()
+		crc := newCRClient()
+
+		// Create dependency first (CRD)
+		crdPlural := genName("ins")
+		crdName := crdPlural + ".cluster.com"
+		cleanupCRD, err := createCRD(c, apiextensions.CustomResourceDefinition{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: crdName,
+			},
+			Spec: apiextensions.CustomResourceDefinitionSpec{
+				Group:   "cluster.com",
+				Version: "v1alpha1",
+				Names: apiextensions.CustomResourceDefinitionNames{
+					Plural:   crdPlural,
+					Singular: crdPlural,
+					Kind:     crdPlural,
+					ListKind: "list" + crdPlural,
+				},
+				Scope: "Namespaced",
+			},
+		})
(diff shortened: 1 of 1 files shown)


ankitathomas (approved):
/ok-to-test

test/e2e/csv_e2e_test.go
```diff
@@ -2664,7 +2664,136 @@ var _ = Describe("CSV", func() {
...
+		// Fetch the current csv
+		fetchedCSV, err := fetchCSV(crc, csv.Name, testNamespace, csvSucceededChecker)
+		require.NoError(GinkgoT(), err)
+
+		// Update csv with modified deployment spec
+		fetchedCSV.Spec.InstallStrategy.StrategySpec = strategyNew
+
+		// Update the current csv
+		_, err = crc.OperatorsV1alpha1().ClusterServiceVersions(testNamespace).Update(context.TODO(), fetchedCSV, metav1.UpdateOptions{})
+		require.NoError(GinkgoT(), err)
```
awgreene, line 2785: Could this all be done in an Eventually() command? As it is, the update command may occasionally fail if another resource modifies the CSV before it is fetched.

awgreene (changes requested):
One change request, looks good otherwise.

test/e2e/csv_e2e_test.go
```diff
@@ -2664,7 +2664,136 @@ var _ = Describe("CSV", func() {
 		err = waitForDeploymentToDelete(GinkgoT(), c, strategy.DeploymentSpecs[0].Name)
 		require.NoError(GinkgoT(), err)
 	})
+	It("update deployment spec in an existing CSV for a hotfix", func() {
```
timflannagan, line 2667: It looks like you may need to rebase against master. When I look at the raw view of this file, that testify/require package is present in the list of package imports, but not present in this same file in master: https://github.com/operator-framework/operator-lifecycle-manager/blob/master/test/e2e/csv_e2e_test.go#L3-L35, which explains why the e2e tests are failing due `test/e2e/csv_e2e_test.go:3045:3: undefined: require`.

test/e2e/csv_e2e_test.go
jeyaramashok: sorry I missed this msg, I have update to include eventually() block https://github.com/operator-framework/operator-lifecycle-manager/pull/1817/files#diff-e94bc9e5ebd2b6db20f851288184630432208a5a772cf0515935707ad3660799R3034

test/e2e/csv_e2e_test.go
jeyaramashok: rebased the PR with latest on master

jeyaramashok: I verified that the test passes on my local machine.

PR #1828, opened 2020-10-23
Specify cluster name when there is only one cluster in kind

### LINKED ISSUE #1827: no nodes found for cluster "kind" while making run-local
## Bug Report

**What did you do?**
Running `make run-local` in project

OLM runs in my kind cluster

ERROR: no nodes found for cluster "kind"

**Environment**
* operator-lifecycle-manager version: 0e65642458a1e44e77bd5509dceee3a41da35b76

* Kubernetes version information: v1.19.1

* Kubernetes cluster kind: deployed using kind

**Possible Solution**
Specify the cluster name when there is only one cluster in kind

**Additional context**
I have created a kubernetes cluster with a specific name using kind, and kind will use the cluster called "kind" if you run any commands without specifying the cluster name. So it would be convenient to specify the cluster name when there is only one cluster.

zkyle95:
Closes #1827 
Specify the cluster name in kind while running build-local if there is only one cluster. Otherwise it will use the default cluster name "kind".

More flexibility to run OLM with kind in local

**Reviewer Checklist**

### DIFF
--- a/scripts/build_local.sh
+++ b/scripts/build_local.sh
@@ -17,7 +17,15 @@ docker build -f e2e.Dockerfile -t quay.io/operator-framework/olm:local -t quay.i
 docker build -f test/e2e/hang.Dockerfile -t hang:10 ./bin
 
 if [ -n "$KIND" ]; then
-  kind load docker-image quay.io/operator-framework/olm:local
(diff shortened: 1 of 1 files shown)


dinhxuanvu (approved):
/lgtm

PR #1832, opened 2020-10-26
add current profile annotations to CVO manifests

guillaumerose:
This is matches openshift/enhancements#510 and doesn't change existing behavior

**Reviewer Checklist**


ecordell:
this looks good, but the manifests are generated from `/deploy`, so the manifests will need to change there in order to merge.

guillaumerose:
Thanks for the advice @ecordell !

I hope what I did is not too far from what you think. I tried to keep the annotation as close as possible from ocp and not leak it in upstream. 

This script adds the annotations at the end of the generation. The main issue we that is we need to deserialize/serialize the YAML. It changes the indentation, etc. With yq v3, comments are preserved but indents changed in some places.

Let me know what you think!

guillaumerose:
Almost all green. How can I relaunch the failed github action?

PR #1834, opened 2020-10-26
Enable debug messgaes via klog

errordeveloper:

At present any libraries using klog do not log debug messages when
the `--debug` flag is set. This change chains klog to `--debug` flag
and enables debugging of all dependencies that use klog, e.g. various
packages under `pkg/lib/operatorclient` and `pkg/lib/kubernetes`.

I was faced with a challenge while trying to undetstand what is wrong
with CSV object  that I have setup and I kept getting unsatisfied RBAC
dependencies. I have looked at the code and found the following log
message:

https://github.com/operator-framework/operator-lifecycle-manager/blob/0e65642458a1e44e77bd5509dceee3a41da35b76/pkg/lib/kubernetes/plugin/pkg/auth/authorizer/rbac/rbac.go#L118

However, having looked at how `klog` is initialised, I learnt that the
log level cannot be controlled in any way, so I had to make this
change and I hope it will be useful to others.

**Reviewer Checklist**

### DIFF
--- a/cmd/olm/main.go
+++ b/cmd/olm/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"crypto/tls"
+	"flag"
 	"fmt"
 	"net/http"
 	"os"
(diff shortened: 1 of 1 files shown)


errordeveloper:
Ping!

errordeveloper:
@ecordell thank for the approval, looks like test failures are unrelated - wdyt?

PR #1835, opened 2020-10-27
Bug 1891898: Set default, min, and max values on Webhook ports

awgreene:
Problem: Operator Authors can create CSVs that define Webhooks with
containerPorts set outside the range of 1-65535, which prevents the
ValidatingWebhookConfiguration or MutatingWebhookConfiguration from
being created due to failed validation.

Solution: Update the CSV's CRD Validation to include the proper min and
max values for webhookDescription ContainerPort field. If the
containerPort is not defined, default to 443.


dinhxuanvu (approved):
/lgtm

PR #1843, opened 2020-10-29
Bug 1841175: Recreate pending installplan if deleted before approval

dinhxuanvu:
At the moment, OLM will not recreate the pending-approval
installplan. The subscription will be stuck in UpgradePending forever
as a result. This PR will enable installplan recreation if it is deleted
before approval.

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -831,6 +831,12 @@ func (o *Operator) syncResolvingNamespace(obj interface{}) error {
 		return err
 	}
 
+	ips, err := o.listInstallPlansMap(namespace)
+	if err != nil {
+		logger.WithError(err).Debug("couldn't list installplan")
+		return err
+	}
+
 	// TODO: parallel
 	maxGeneration := 0
 	subscriptionUpdated := false
@@ -847,7 +853,7 @@ func (o *Operator) syncResolvingNamespace(obj interface{}) error {
 		}
 
 		// ensure the installplan reference is correct
-		sub, changedIP, err := o.ensureSubscriptionInstallPlanState(logger, sub)
+		sub, changedIP, err := o.ensureSubscriptionInstallPlanState(logger, sub, ips)
 		if err != nil {
 			logger.Debugf("error ensuring installplan state: %v", err)
 			return err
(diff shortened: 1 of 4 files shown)


pkg/controller/operators/catalog/operator.go
```diff
@@ -940,13 +940,26 @@ func (o *Operator) syncSubscriptions(obj interface{}) error {
...
 	}
-	if sub.Status.InstallPlanRef != nil && sub.Status.State == v1alpha1.SubscriptionStateUpgradePending {
-		logger.Debugf("skipping update: installplan already created")
+	return false
+}
+
+// isInstallPlanMissing checks if the installplan is missing or not
+func (o *Operator) isInstallPlanMissing(sub *v1alpha1.Subscription) bool {
+	_, err := o.client.OperatorsV1alpha1().InstallPlans(sub.GetNamespace()).Get(context.TODO(), sub.Status.Install.Name, metav1.GetOptions{})
+	if err != nil {
```
benluddy, line 962: An error here doesn't imply that the plan is missing (for example, the API server could be unreachable).

benluddy (commented):
Did you consider handling this by verifying the install plan reference as part of reconciling subscriptions, then making `ensureSubscriptionInstallPlanState` bail if a subscription has a missing reference? I think this fix will be more resilient if it doesn't require the resolver to change its behavior.

dinhxuanvu:
> Did you consider handling this by verifying the install plan reference as part of reconciling subscriptions, then making `ensureSubscriptionInstallPlanState` bail if a subscription has a missing reference? I think this fix will be more resilient if it doesn't require the resolver to change its behavior.

I have considered to do that but to me the state of the subscription doesn't change. It is the IP that is missing which we simply need to recreate. Changing the state can be a bit confusing to the UI. However, I think it may require less code if we change the subscription state which will force the `nothingToUpdate` to return false. That's something I can do.

test/e2e/subscription_e2e_test.go
```diff
@@ -217,8 +217,34 @@ var _ = Describe("Subscription", func() {
...
+		newInstallPlan, err := fetchInstallPlan(GinkgoT(), crc, ipName, buildInstallPlanPhaseCheckFunc(v1alpha1.InstallPlanPhaseRequiresApproval))
+		require.NoError(GinkgoT(), err)
+		require.NotNil(GinkgoT(), newInstallPlan)
+
+		require.NotEqual(GinkgoT(), installPlan.Name, newInstallPlan.Name, "expected new installplan recreated")
+		require.Equal(GinkgoT(), v1alpha1.ApprovalManual, newInstallPlan.Spec.Approval)
+		require.Equal(GinkgoT(), v1alpha1.InstallPlanPhaseRequiresApproval, newInstallPlan.Status.Phase)
+
+		newInstallPlan.Spec.Approved = true
+		_, err = crc.OperatorsV1alpha1().InstallPlans(testNamespace).Update(context.Background(), newInstallPlan, metav1.UpdateOptions{})
```
benluddy, line 247: Is this going to flake occasionally if the plan is updated between the GET and the PUT?

dinhxuanvu: It is possible but then that scenario is possible for any update that we do to any OLM objects really due to reconciliation process. This is the original code so I didn't change this line. No flakes so far on the test run but do you have something in mind to improve this? Wrapping this around `eventually`?

benluddy: Does https://github.com/operator-framework/operator-lifecycle-manager/blob/master/test/e2e/util_test.go#L875 help?

dinhxuanvu: Sure. I can use that.

dinhxuanvu:
@benluddy May I have a lgtm and approval? :)

pkg/controller/operators/catalog/subscription/state.go
```diff
@@ -385,23 +385,33 @@ func (i *installPlanReferencedState) InstallPlanNotFound(now *metav1.Time, clien
...
+		cond := out.Status.GetCondition(v1alpha1.SubscriptionInstallPlanMissing)
+		cond.Status = corev1.ConditionTrue
+		cond.Reason = v1alpha1.ReferencedInstallPlanNotFound
+		cond.LastTransitionTime = now
+		out.Status.SetCondition(cond)
+	}
 
 	// Build missing state
-	missingState := &installPlanMissingState{
+	state := &installPlanMissingState{
```
kevinrizza, line 406: This is a really really minor nit, but I don't love naming a variable `state`.

kevinrizza (commented):
/lgtm

PR #1848, opened 2020-11-02
Bug 1891995: Add spec hash to service's label to ensure service is correct

dinhxuanvu:
Calculate has from service spec and add it to service's labels.
Will use this hash info to ensure service during catalog source
reconciliation. This feature will force OLM to recreate new
service if the hash doesn't match to avoid stale service with
the same name.

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/registry/reconciler/configmap.go
+++ b/pkg/controller/registry/reconciler/configmap.go
@@ -85,6 +85,11 @@ func (s *configMapCatalogSourceDecorator) Service() *v1.Service {
 			Selector: s.Selector(),
 		},
 	}
+
+	labels := map[string]string{}
+	hash := HashServiceSpec(svc.Spec)
+	labels[ServiceHashLabelKey] = hash
+	svc.SetLabels(labels)
 	ownerutil.AddOwner(svc, s.CatalogSource, false, false)
 	return svc
 }
@@ -363,8 +368,9 @@ func (c *ConfigMapRegistryReconciler) ensurePod(source configMapCatalogSourceDec
 
 func (c *ConfigMapRegistryReconciler) ensureService(source configMapCatalogSourceDecorator, overwrite bool) error {
 	service := source.Service()
-	if c.currentService(source) != nil {
-		if !overwrite {
+	svc := c.currentService(source)
+	if svc != nil {
+		if !overwrite && ServiceHashMatch(svc, service) {
 			return nil
 		}
 		if err := c.OpClient.DeleteService(service.GetNamespace(), service.GetName(), metav1.NewDeleteOptions(0)); err != nil {
--- a/pkg/controller/registry/reconciler/grpc.go
+++ b/pkg/controller/registry/reconciler/grpc.go
@@ -3,25 +3,28 @@ package reconciler
 import (
 	"context"
 	"fmt"
+	"hash/fnv"
 	"time"
 
 	controllerclient "github.com/operator-framework/operator-lifecycle-manager/pkg/lib/controller-runtime/client"
 
 	"github.com/operator-framework/api/pkg/operators/v1alpha1"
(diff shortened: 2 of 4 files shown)


pkg/controller/registry/reconciler/configmap.go
```diff
@@ -85,6 +85,11 @@ func (s *configMapCatalogSourceDecorator) Service() *v1.Service {
 			Selector: s.Selector(),
 		},
 	}
+
+	labels := map[string]string{}
```
exdx, line 89: Could we use 
```
labels := svc.GetLabels()
if labels == nil {
    labels = map[string]string{}
}
```
That way we don't end up overwriting user-defined labels on services, they may have some set

pkg/controller/registry/reconciler/grpc.go
```diff
@@ -3,25 +3,28 @@ package reconciler
...
-	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/apimachinery/pkg/util/rand"
 )
 
 const (
 	CatalogSourceUpdateKey      = "catalogsource.operators.coreos.com/update"
+	ServiceHashLabelKey         = "olm.service-spec-hash"
```
exdx, line 27: should the label  key format be consistent? ie `catalogsource.operators.coreos.com/service-spec-hash`

pkg/controller/registry/reconciler/grpc.go
```diff
@@ -74,11 +77,16 @@ func (s *grpcCatalogSourceDecorator) Service() *v1.Service {
 			Selector: s.Labels(),
 		},
 	}
+
+	labels := map[string]string{}
```
exdx, line 81: same comment as above regarding non-nil labels

pkg/controller/registry/reconciler/grpc.go
```diff
@@ -274,6 +283,33 @@ func (c *GrpcRegistryReconciler) ensureService(source grpcCatalogSourceDecorator
...
+	labels := existing.GetLabels()
+	if len(labels) == 0 {
+		return false
+	}
+	existingSvcSpecHash, ok := labels[ServiceHashLabelKey]
+	if !ok {
+		return false
+	}
+
+	hash, _ := new.GetLabels()[ServiceHashLabelKey]
```
exdx, line 298: should we check that the key exists here (as above) or are we saying we will always have a hash on the `new` service? That would be an error based on this logic.

exdx (commented):
looks good, just left some small comments. I think an e2e test could also be helpful.

pkg/controller/registry/reconciler/configmap.go
dinhxuanvu: This is a brand-new Service that we create for the first time here so there isn't any user-defined labels. You can see there is no input because this Service is created and managed by OLM.

pkg/controller/registry/reconciler/grpc.go
dinhxuanvu: The current active label in the pod is `olm.catalogSource: certified-operators` and the label for deployment follows the same pattern `olm.deployment-spec-hash`. That's why I use this `olm.something` for this reason. The label you mention is for update and temporary use.

pkg/controller/registry/reconciler/grpc.go
dinhxuanvu: The `new` Service is coming from `source.Service()` which with the code change will guarantee there is a label. However, I can see your point that this func may be used else where so the provided Service may not have labels. I can change that.

ecordell:
/approve

looks good, but missing tests

PR #1855, opened 2020-11-09
Bug 1819457: Services should not have duplicate ownerrefs

ecordell:
This change only adds a unit test to verify that services created for apiservices installed via OLM do not have multiple ownerrefs added.

The test harness is bulky, but because this is intended to be backported, I did not attempt to refactor the related apiservice management to make testing simpler.

This is to verify that the changes from this commit (https://github.com/operator-framework/operator-lifecycle-manager/commit/1476ff19c6a0c0a37a2fcdb47aa982683d37fee6) are included in master.

This set of changes + the above commit will be backported to older versions, to address a report of this bug in 4.5

**Reviewer Checklist**


pkg/lib/operatorlister/lister.go
```diff
@@ -11,6 +11,18 @@ import (
...
+//go:generate go run github.com/maxbrunsfeld/counterfeiter/v6 -o ../../fakes/client-go/listers/fake_v1_secret_namespace_lister.go k8s.io/client-go/listers/core/v1.SecretNamespaceLister
+//go:generate go run github.com/maxbrunsfeld/counterfeiter/v6 -o ../../fakes/client-go/listers/fake_rbac_v1_role_lister.go k8s.io/client-go/listers/rbac/v1.RoleLister
+//go:generate go run github.com/maxbrunsfeld/counterfeiter/v6 -o ../../fakes/client-go/listers/fake_rbac_v1_role_namespace_lister.go k8s.io/client-go/listers/rbac/v1.RoleNamespaceLister
+//go:generate go run github.com/maxbrunsfeld/counterfeiter/v6 -o ../../fakes/client-go/listers/fake_rbac_v1_rolebinding_lister.go k8s.io/client-go/listers/rbac/v1.RoleBindingLister
+//go:generate go run github.com/maxbrunsfeld/counterfeiter/v6 -o ../../fakes/client-go/listers/fake_rbac_v1_rolebinding_namespace_lister.go k8s.io/client-go/listers/rbac/v1.RoleBindingNamespaceLister
+//go:generate go run github.com/maxbrunsfeld/counterfeiter/v6 -o ../../fakes/client-go/listers/fake_rbac_v1_clusterrolebinding_lister.go k8s.io/client-go/listers/rbac/v1.ClusterRoleBindingLister
```
njhale, line 24: 👍

pkg/controller/install/certresources_test.go
```diff
@@ -0,0 +1,615 @@
...
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+		},
+		{
+			name: "doesn't add duplicate service ownerrefs",
```
njhale, line 367: I'm having a difficult time finding where the service is actually checked for duplicate ownerreferences in the test case. @ecordell could you point it out?

pkg/controller/install/certresources_test.go
```diff
@@ -0,0 +1,615 @@
...
+			mockExternal: func(mockOpClient *operatorclientmocks.MockClientInterface, fakeLister *operatorlisterfakes.FakeOperatorLister, namespace string, args args) {
+				mockOpClient.EXPECT().DeleteService(namespace, "test-service", &metav1.DeleteOptions{}).Return(nil)
+				service := corev1.Service{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      "test-service",
+						Namespace: owner.GetNamespace(),
+						OwnerReferences: []metav1.OwnerReference{
+							ownerutil.NonBlockingOwner(owner),
```
ecordell, line 375: this is a service that has the ownerref set

pkg/controller/install/certresources_test.go
```diff
@@ -0,0 +1,615 @@
...
+				authReaderRoleBinding.SetNamespace(KubeSystem)
+
+				mockOpClient.EXPECT().UpdateRoleBinding(authReaderRoleBinding).Return(authReaderRoleBinding, nil)
+			},
+			state: fakeState{
+				existingService: &corev1.Service{
+					ObjectMeta: metav1.ObjectMeta{
+						Namespace: owner.GetNamespace(),
+						OwnerReferences: []metav1.OwnerReference{
+							ownerutil.NonBlockingOwner(owner),
```
ecordell, line 494: this sets the "existing" service on the cluster to already have the ownerref

pkg/controller/install/certresources_test.go
```diff
@@ -0,0 +1,615 @@
...
+						},
+					},
+					Spec: corev1.ServiceSpec{
+						Ports:    args.ports,
+						Selector: selector(t, "test=label").MatchLabels,
+					},
+				}
+				mockOpClient.EXPECT().CreateService(&service).Return(&service, nil)
```
ecordell, line 383: and then this mock is expected to be called with a `service` object with that exact spec - i.e. only one owner ref

if you remove the commit that fixes the bug, this mock fails because the service passed to `CreateService` will have two ownerrefs and fail mock validation.

njhale: Ah, so if you don't pass a single ownerref service, the mock will return nil and I assume that would cause a failure -- subtle.

Issue #1237, opened 2020-01-16
Duplicate owner reference in packageserver service


bryanl:
## Bug Report

**What did you do?**

Installed OLM using kubectl

**What did you expect to see?**

The service created should have a single owner reference to the CSV. 

**What did you see instead? Under which circumstances?**

The service has multiple owner references to the same CSV.

```yaml
---
apiVersion: v1
kind: Service
metadata:
  creationTimestamp: "2020-01-15T13:49:15Z"
  name: v1-packages-operators-coreos-com
  namespace: olm
  ownerReferences:
  - apiVersion: operators.coreos.com/v1alpha1
    blockOwnerDeletion: false
    controller: false
    kind: ClusterServiceVersion
    name: packageserver
    uid: e2bdc207-8f58-4ef8-afe4-212704dbe060
  - apiVersion: operators.coreos.com/v1alpha1
    blockOwnerDeletion: false
    controller: false
    kind: ClusterServiceVersion
    name: packageserver
    uid: e2bdc207-8f58-4ef8-afe4-212704dbe060
  resourceVersion: "201176"
  selfLink: /api/v1/namespaces/olm/services/v1-packages-operators-coreos-com
  uid: 7a1f2c82-e2ae-4adb-a10d-dee40c8a76ae
spec:
  clusterIP: 10.96.68.225
  ports:
  - port: 443
    protocol: TCP
    targetPort: 5443
  selector:
    app: packageserver
  sessionAffinity: None
  type: ClusterIP
status:
  loadBalancer: {}
```

**Environment**
* operator-lifecycle-manager version:

https://github.com/operator-framework/operator-lifecycle-manager/tree/976f2b981c326ee51748045f6df20317d6d95bd5

* Kubernetes version information:

```
Client Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.0", GitCommit:"70132b0f130acc0bed193d9ba59dd186f0e634cf", GitTreeState:"clean", BuildDate:"2019-12-07T21:20:10Z", GoVersion:"go1.13.4", Compiler:"gc", Platform:"darwin/amd64"}
[...]

PR #1856, opened 2020-11-09
Bug 1896051: Services should not have duplicate ownerrefs

ecordell:
This change only adds a unit test to verify that services created for apiservices installed via OLM do not have multiple ownerrefs added.

The test harness is bulky, but because this is intended to be backported, I did not attempt to refactor the related apiservice management to make testing simpler.

This is to verify that the changes from this commit (https://github.com/operator-framework/operator-lifecycle-manager/commit/1476ff19c6a0c0a37a2fcdb47aa982683d37fee6) are included in 4.6.

This set of changes + the above commit will be backported to older versions, to address a report of this bug in 4.5

**Reviewer Checklist**


exdx:
Addresses #1237

exdx:
This looks great and it looks like https://github.com/kubernetes/kubernetes/pull/96185 landing upstream will resolve this issue on the api-server side. 

/lgtm

PR #1863, opened 2020-11-11
Don't create default ServiceAccounts

### LINKED ISSUE #1752: Installation of the service account does not keep pull image secrets
## Type of question

**Are you asking about community best practices, how to implement a specific feature, or about general context and help around the operator-sdk?**

## Question

**What did you do?**
I had multiple operators developed by us scaffolded by operator-sdk 1.0 installed into the same namespace, the `ServiceAccount` `default` was overwritten multiple times and the field `pullImageSecrets` was removed, and the field `metadata.ownerReferences` was set to one of the CSV.

- do not overwrite the service account if it already exists
- or preserve `pullImageSecrets` previously set to the `ServiceAccount`.

The `pullImageSecrets` field was removed from the `ServiceAccount` `default` after the installation.

**Environment**
v0.15.1

v1.16.13
 
* Kubernetes cluster kind: 
GKE

**Additional context**

### LINKED ISSUE #1820: Operator installation fails when using imagePullSecret linked to the default service account
## Bug Report

**What did you do?**
We are working on an operator that installs via OLM and the images are from a private registry. The operator uses `default` Service Account (in CSV). Hence, we create a (secret) `imagePullSecret` to our private image registry and link it to the 'default' service account before we install the operator. The deployment of the operator, updates the default Service Account and removes the linked `imagePullSecret` from the ServiceAccount. This causes the deployment of the operator to fail. Once we recreate the secret and the link to the Service Account and following it up with pod recreation [delete the pod (for reconcile to happen), the recreated pod just picks it up right] it works. When we define a Service Account and use the one that we created, the issue doesn't appear

Operator deployment succeeds with the given `imagePullSecret` in `default` service account

Deployment of the operator failed

**Environment**
v1.0.0

```
  oc v3.11.0+0cbc58b
    kubernetes v1.11.0+d4cacc0
    features: Basic-Auth
    kubernetes v1.18.3+6c42de8
```

* Kubernetes cluster kind: 
Openshift 

**Possible Solution**

**Additional context**

jeloba:

Don't create steps to create ServiceAccounts named "default" in install plans.

[According to the k8s docs](https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/#serviceaccount-controller) the ServiceAccount controller is responsible for creating these.

Re-creating the "default" service account may remove privileges or pull secrets granted to it before the operator was installed.

In our case these are pull secrets that are needed for fetching the controller image.

Closes #1752
Closes #1820

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/registry/resolver/step_resolver_test.go
+++ b/pkg/controller/registry/resolver/step_resolver_test.go
@@ -831,6 +831,19 @@ func TestNamespaceResolverRBAC(t *testing.T) {
 		},
 	}
 	bundle := bundleWithPermissions("a.v1", "a", "alpha", "", nil, nil, nil, nil, simplePermissions, simplePermissions)
+	defaultServiceAccountPermissions := []v1alpha1.StrategyDeploymentPermissions{
+		{
+			ServiceAccountName: "default",
+			Rules: []rbacv1.PolicyRule{
+				{
+					Verbs:     []string{"get", "list"},
+					APIGroups: []string{""},
+					Resources: []string{"configmaps"},
+				},
+			},
+		},
+	}
+	bundleWithDefaultServiceAccount := bundleWithPermissions("a.v1", "a", "alpha", "", nil, nil, nil, nil, defaultServiceAccountPermissions, defaultServiceAccountPermissions)
 	type out struct {
 		steps [][]*v1alpha1.Step
 		subs  []*v1alpha1.Subscription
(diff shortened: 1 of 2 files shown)


EronWright:
@jeloba wouldn't the issue of overwriting the pull secrets apply to non-default service accounts also?  

Maybe the fix should be that the OLM controller patches the `ServiceAccount` when it exists in a non-destructive way; that is, update the `ownerRefs` but preserve any secrets, annotations, labels, etc that might have been applied.  This would improve interoperability and solve https://github.com/operator-framework/operator-lifecycle-manager/issues/1931.

jeloba:
@EronWright since this was my first contribution I wanted to keep it minimal.

The re-creation of the default `ServiceAccount` seemed to be in obvious contradiction to the K8s docs and easy to address. It also seemed to be the main issue in the tickets linked and it's the account that the Operator SDK used by default.

I think #1931 would be addressed in #1904 or #1881.

PR #1875, opened 2020-11-20
Create operatorcondition for operator

awgreene:

**Reviewer Checklist**


pkg/controller/registry/resolver/steps.go
```diff
@@ -6,11 +6,13 @@ import (
 	"fmt"
 	"strings"
 
+	v1 "github.com/operator-framework/api/pkg/operators/v1"
```
njhale, line 9: nit: import order, and extra `v1`.

I suggest just aliasing it to `operatorsv1`, and `v1alpha1` to `operatorsv1alpha1` if we feel it makes sense to do in this PR.

pkg/controller/registry/resolver/rbac.go
```diff
@@ -90,6 +90,21 @@ func RBACForClusterServiceVersion(csv *v1alpha1.ClusterServiceVersion) (map[stri
 			permissions[permission.ServiceAccountName] = NewOperatorPermissions(serviceAccount)
 		}
 
+		operatorConditionRBAC := []rbacv1.PolicyRule{
+			{
+				Verbs:         []string{"get"},
```
njhale, line 95: Is it reasonable for an operator to update the spec or change the metadata (e.g. annotations, labels)?

pkg/controller/registry/resolver/rbac.go
```diff
@@ -90,6 +90,21 @@ func RBACForClusterServiceVersion(csv *v1alpha1.ClusterServiceVersion) (map[stri
...
+				APIGroups:     []string{"operators.coreos.com"},
+				Resources:     []string{"operatorconditions"},
+				ResourceNames: []string{csv.GetName()},
+			},
+			{
+				Verbs:         []string{"get,update"},
```
njhale, line 101: +patch?

pkg/controller/registry/resolver/steps.go
```diff
@@ -136,6 +154,12 @@ func NewStepResourceFromBundle(bundle *api.Bundle, namespace, replaces, catalogS
 	}
 	steps := []v1alpha1.StepResource{step}
 
+	step, err = NewOperatorConditionStepResource(csv, catalogSourceName, catalogSourceName)
```
njhale, line 157: What happens if/when an operator author includes an `OperatorCondition` in their bundle?

pkg/lib/ownerutil/util.go
```diff
@@ -406,6 +406,12 @@ func InferGroupVersionKind(obj runtime.Object) error {
 			Version: apiextensionsv1beta1.SchemeGroupVersion.Version,
 			Kind:    "CustomResourceDefinition",
 		})
+	case *operatorsv1.OperatorCondition:
```
njhale, line 409: I wonder if it's possible to get the GVK from a properly initialized `runtime.Scheme`.

njhale, line 409: nit: could we move this case before both CRD versions?

test/e2e/subscription_e2e_test.go
```diff
@@ -75,6 +75,12 @@ var _ = Describe("Subscription", func() {
 		csv, err := fetchCSV(crc, subscription.Status.CurrentCSV, testNamespace, buildCSVConditionChecker(v1alpha1.CSVPhaseSucceeded))
 		require.NoError(GinkgoT(), err)
 
+		// Ensure that the OperatorCondition was created for the CSV.
```
njhale, line 78: I think we should also be interested in ensuring the existence of the automagically generated RBAC for OperatorConditions.

njhale (commented):
Nice work @awgreene!

Aside from some more tests, I do want us to think about what this looks like if we implement `OperatorCondition` generation as a net-new controller (in the style of the Operator API's controllers). I think we'll want to add features as new controllers when we can, so we can reduce the amount we need to rewrite as we move away from OLM's original controllers.

pkg/controller/registry/resolver/rbac.go
awgreene: At this time no, we can always expand these privileges in the future

pkg/controller/registry/resolver/rbac.go
awgreene: I think that's fair, I'll add it in an update.

pkg/controller/registry/resolver/steps.go
awgreene: I'm glad you brought that up. Since each `OperatorCondition` is created for a `CSV` in the installplan I believe that it makes sense to update the `isSupported` feature to make sure that the operatorCondition name matches a CSV included in the installplan.

pkg/lib/ownerutil/util.go
awgreene: Sure!

test/e2e/subscription_e2e_test.go
awgreene: Agreed - I'm working on that test locally.

awgreene:
Thanks for the detailed review @njhale 

> Aside from some more tests, I do want us to think about what this looks like if we implement OperatorCondition generation as a net-new controller (in the style of the Operator API's controllers). I think we'll want to add features as new controllers when we can, so we can reduce the amount we need to rewrite as we move away from OLM's original controllers.

Are you proposing that instead of generating the `OperatorConditions` and RBAC as part of the installplan we instead create a new controller that generates these resources itself when a CSV is created?

njhale:
> Are you proposing that instead of generating the OperatorConditions and RBAC as part of the installplan we create a new controller that generates these resources itself when a CSV is created?

Exactly.

pkg/lib/ownerutil/util.go
```diff
@@ -45,6 +45,17 @@ func IsOwnedBy(object metav1.Object, owner Owner) bool {
 	return false
 }
 
+func HasOwnerReference(object metav1.Object, owner metav1.OwnerReference) bool {
+	for _, oref := range object.GetOwnerReferences() {
+		if oref.UID == owner.UID &&
```
gallettilance, line 50: would just UID be enough?

gallettilance (commented):
looks great!

pkg/controller/operators/operatocondition_controller.go
```diff
@@ -0,0 +1,262 @@
...
+		return nil, err
+	}
+
+	return &OperatorConditionReconciler{
+		Client: cli,
+		log:    log,
+	}, nil
+}
+
+const (
```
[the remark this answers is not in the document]
awgreene: Sure.

pkg/controller/operators/operatocondition_controller.go
```diff
@@ -0,0 +1,262 @@
...
+	err = r.Client.Create(context.TODO(), &role)
+	if err != nil {
+		if !k8serrors.IsAlreadyExists(err) {
+
+			return err
+		}
+		existingRole := rbacv1.Role{}
+		err := r.Client.Get(context.TODO(), client.ObjectKey{Name: role.GetName(), Namespace: role.GetNamespace()}, &existingRole)
+		if err != nil {
+			r.log.V(1).Info("Alex: error getting role")
```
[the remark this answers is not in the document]
awgreene: Nope - applies in all places.

pkg/controller/operators/operatoconditiongenerator_controller.go
```diff
@@ -0,0 +1,126 @@
...
+
+	operatorsv1 "github.com/operator-framework/api/pkg/operators/v1"
+	operatorsv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/operators/decorators"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/ownerutil"
+)
+
+// OperatorReconciler reconciles a Operator object.
+type OperatorConditionGenReconciler struct {
+	client.Client
```
benluddy, line 23: Why is `Client` embedded?

pkg/controller/operators/operatocondition_controller.go
```diff
@@ -0,0 +1,259 @@
...
+			return err
+		}
+		existingRole := rbacv1.Role{}
+		err := r.Client.Get(context.TODO(), client.ObjectKey{Name: role.GetName(), Namespace: role.GetNamespace()}, &existingRole)
+		if err != nil {
+			return err
+		}
+		existingRole.OwnerReferences = role.OwnerReferences
+		existingRole.Rules = role.Rules
+		err = r.Client.Update(context.TODO(), &existingRole)
```
benluddy, line 135: Will this generate a lot of write load if the update call is made whether or not there's a change? I'm especially concerned with the case where an operator spams writes to its condition resource and this controller amplifies that traffic.

pkg/controller/operators/operatocondition_controller.go
```diff
@@ -0,0 +1,259 @@
...
+
+	return nil
+}
+
+func hasOwnerFunc(ownerReference metav1.OwnerReference) func(object metav1.Object) bool {
+	return func(object metav1.Object) bool {
+		if ownerutil.HasOwnerReference(object, ownerReference) {
+			return true
+		}
+		return false
```
benluddy, line 181:
```suggestion
		return ownerutil.HasOwnerReference(object, ownerReference)
```

pkg/controller/operators/operatocondition_controller.go
```diff
@@ -0,0 +1,259 @@
...
+	}
+}
+
+func filterServiceAccounts(serviceAccounts []corev1.ServiceAccount, f func(object metav1.Object) bool) []corev1.ServiceAccount {
+	// // filter based on OwnerReferences
```
benluddy, line 186: Very consequential suggestion:
```suggestion
	// filter based on OwnerReferences
```

pkg/controller/operators/operatoconditiongenerator_controller.go
```diff
@@ -0,0 +1,126 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	operatorsv1 "github.com/operator-framework/api/pkg/operators/v1"
+	operatorsv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/operators/decorators"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/ownerutil"
+)
+
+// OperatorReconciler reconciles a Operator object.
+type OperatorConditionGenReconciler struct {
```
benluddy, line 22: Maybe `OperatorConditionGenerationController`? Calling it a "condition reconciler" is a bit misleading, since it's not technically watching conditions.

pkg/controller/operators/operatoconditiongenerator_controller.go
```diff
@@ -0,0 +1,126 @@
...
+		// Get CSV Namespace
+		csvList := operatorsv1alpha1.ClusterServiceVersionList{}
+		err = r.Client.List(context.TODO(), &csvList, &client.ListOptions{LabelSelector: selector})
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+
+		// If you cannot identify the CSV for the operator do not proceed.
+		if len(csvList.Items) != 1 {
+			return ctrl.Result{Requeue: true}, err
```
benluddy, line 102: `err` should always be nil here, right? Returning nil directly would help others recognize that at a glance.

pkg/controller/operators/operatocondition_controller.go
```diff
@@ -0,0 +1,247 @@
...
+			found = true
+			result = append(result, envVar)
+		}
+	}
+
+	if !found {
+		result = append(result, envVar)
+	}
+
+	return result
```
benluddy, line 246: If I understand, this can be simplified a bit (and more if the duplication isn't important):

```suggestion
	dupe := make([]corev1.EnvVar, len(envVars))
	copy(dupe, envVars)

	for _, each := range dupe {
		if each.Name == envVar.Name {
			return dupe
		}
	}

	return append(dupe, envVar)
```

benluddy (commented):
Nice work!

pkg/controller/operators/operatocondition_controller.go
awgreene: Good point - I'll update the code to check if any of the fields that the controller sets were modified before updating the resource.

awgreene: ^applies to deps, roles, and roleBindings

pkg/controller/operators/operatoconditiongenerator_controller.go
awgreene: Coppied from reference code, I'll unembed it.

pkg/controller/operators/operatoconditiongenerator_controller.go
```diff
@@ -0,0 +1,126 @@
...
+
+// NewOperatorReconciler constructs and returns an OperatorReconciler.
+// As a side effect, the given scheme has operator discovery types added to it.
+func NewOperatorConditionGenReconciler(cli client.Client, log logr.Logger, scheme *runtime.Scheme) (*OperatorConditionGenReconciler, error) {
+	// Add watched types to scheme.
+	if err := AddToScheme(scheme); err != nil {
+		return nil, err
+	}
+
+	factory, err := decorators.NewSchemedOperatorFactory(scheme)
```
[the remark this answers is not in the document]
awgreene: So this factory code was [copied from the operator controller](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/controller/operators/operator_controller.go#L96-L99), should we change in both places?

pkg/controller/operators/operatocondition_controller.go
awgreene: Addressed in the latest version of the PR.

pkg/controller/operators/operatoconditiongenerator_controller.go
awgreene: Addressed in the latest version of the PR.

pkg/controller/operators/operatocondition_controller.go
awgreene: Addressed in the latest version of the PR.

pkg/controller/operators/operatocondition_controller.go
awgreene: Addressed in the latest version of the PR.

pkg/controller/operators/operatocondition_controller.go
awgreene: Addressed in the latest version of the PR.

pkg/controller/operators/operatoconditiongenerator_controller.go
awgreene: Addressed in the latest version of the PR.

pkg/controller/operators/operatoconditiongenerator_controller.go
```diff
@@ -0,0 +1,126 @@
...
+	if err != nil {
+		return reconcile.Result{Requeue: true}, nil
+	}
+	operatorConditionList := operatorsv1.OperatorConditionList{}
+
+	err = r.Client.List(context.TODO(), &operatorConditionList, &client.ListOptions{LabelSelector: selector})
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	if len(operatorConditionList.Items) == 0 {
```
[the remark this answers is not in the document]
awgreene: Addressed in the latest version of the PR.

pkg/controller/operators/operatoconditiongenerator_controller.go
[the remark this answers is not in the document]
awgreene: Spoke offline, will be addressed in all places in a future PR.

pkg/controller/operators/operatocondition_controller.go
```diff
@@ -0,0 +1,259 @@
...
+		Complete(r)
+}
+
+// NewOperatorReconciler constructs and returns an OperatorReconciler.
+// As a side effect, the given scheme has operator discovery types added to it.
+func NewOperatorConditionReconciler(cli client.Client, log logr.Logger, scheme *runtime.Scheme) (*OperatorConditionReconciler, error) {
+	// Add watched types to scheme.
+	if err := AddToScheme(scheme); err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
awgreene: Spoke offline, will be addressed in all places in a future PR.

pkg/controller/operators/operatocondition_controller.go
```diff
@@ -0,0 +1,259 @@
+package operators
```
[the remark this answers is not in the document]
awgreene: Addressed in the latest version of the PR.

pkg/lib/ownerutil/util.go
awgreene: UID might be enough but I don't see why we shouldn't check multiple fields.

test/e2e/subscription_e2e_test.go
awgreene: Addressed in the latest version of the PR.

pkg/controller/registry/resolver/steps.go
awgreene: No longer relevant.

pkg/controller/registry/resolver/rbac.go
awgreene: Added in the latest version of the PR.

pkg/controller/registry/resolver/steps.go
awgreene: No longer applies but taking the lesson and applying it to work in the future.

pkg/controller/operators/operatorconditiongenerator_controller.go
```diff
@@ -0,0 +1,203 @@
...
+	for _, serviceAccount := range filterServiceAccounts(serviceAccountList.Items, hasOwnerFunc(ownerutil.NonBlockingOwner(&csv))) {
+		serviceAccounts = append(serviceAccounts, serviceAccount.GetName())
+	}
+	if len(serviceAccounts) == 0 {
+		serviceAccounts = []string{"default"}
+	}
+
+	// get all Service Accounts
+	deploymentList := appsv1.DeploymentList{}
+	err = r.Client.List(context.TODO(), &deploymentList, client.InNamespace(csv.GetNamespace()))
```
ecordell, line 132: these calls should all be hitting the cache as well

pkg/controller/operators/operatorconditiongenerator_controller.go
```diff
@@ -0,0 +1,203 @@
...
+		if f(&deployment) {
+			filteredDeployments = append(filteredDeployments, deployment)
+		}
+	}
+	return filteredDeployments
+}
+
+func hasOwnerFunc(ownerReference metav1.OwnerReference) func(object metav1.Object) bool {
+	return func(object metav1.Object) bool {
+		return ownerutil.HasOwnerReference(object, ownerReference)
```
ecordell, line 190: there should be an `IsOwnedBy` already there

pkg/controller/operators/operatorconditiongenerator_controller.go
```diff
@@ -0,0 +1,203 @@
...
+	if err != nil {
+		return ctrl.Result{Requeue: true}, err
+	}
+	if len(operatorConditionList.Items) != 0 {
+		return ctrl.Result{}, nil
+	}
+	r.log.V(1).Info("Creating OperatorCondition")
+
+	// Get CSV Namespace
+	csvList := operatorsv1alpha1.ClusterServiceVersionList{}
```
ecordell, line 97: should this just reconcile on CSVs instead of on operators?

pkg/controller/operators/operatorconditiongenerator_controller.go
```diff
@@ -0,0 +1,203 @@
...
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	// If you cannot identify the CSV for the operator do not proceed.
+	if len(csvList.Items) != 1 {
+		return ctrl.Result{Requeue: true}, nil
+	}
+
+	csv := csvList.Items[0]
```
ecordell, line 108: you should be able to derive CSV -> serviceaccount names and deployments names from existing machinery in the `install` package, right? no need to query the namespace.

pkg/controller/operators/operatorconditiongenerator_controller.go
```diff
@@ -0,0 +1,203 @@
...
+			Labels:          map[string]string{labelKey: ""},
+			OwnerReferences: []metav1.OwnerReference{ownerutil.NonBlockingOwner(&csv)},
+		},
+		Spec: operatorsv1.OperatorConditionSpec{
+			ServiceAccounts: serviceAccounts,
+			Deployments:     deployments,
+		},
+	}
+
+	err = r.Client.Create(context.TODO(), &operatorCondition)
```
ecordell, line 156: this should only create/update if it doesn't already exist, right? 

also - does every operator automatically get an OperatorCondition, even if nothing is configured to write to it?

pkg/controller/operators/operatorcondition_controller.go
```diff
@@ -0,0 +1,237 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+
+	operatorsv1 "github.com/operator-framework/api/pkg/operators/v1"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/ownerutil"
+)
+
+const (
+	OperatorConditionEnvVarKey = "OPERATOR_CONDITION_NAME"
```
ecordell, line 26: Just to clarify - there's only one operator condition per operator, but multiple deployments/serviceaccounts have permission to write to it?

manifests/0000_50_olm_00-operatorconditions.crd.yaml
```diff
@@ -0,0 +1,199 @@
...
+  creationTimestamp: null
+  name: operatorconditions.operators.coreos.com
+spec:
+  group: operators.coreos.com
+  names:
+    kind: OperatorCondition
+    listKind: OperatorConditionList
+    plural: operatorconditions
+    singular: operatorcondition
+  scope: Namespaced
```
ecordell, line 15: I thought this was supposed to be a cluster scoped resource?

awgreene: @kevinrizza and I discussed this earlier and we decided to keep it namespaced.

pkg/controller/operators/operatorcondition_controller.go
awgreene: At this time yes. In the future, OLM can be updated to allow operator authors to define the deployments and serviceAccounts that should be able to update the resource by providing the operatorcondition in the bundle and specifying said resources in the spec.

pkg/controller/operators/operatorconditiongenerator_controller.go
awgreene: >this should only create/update if it doesn't already exist, right?

Yes - but the code will never reach this point because of an earlier[ if statement:](https://github.com/operator-framework/operator-lifecycle-manager/pull/1875/files#diff-2827b0191f1f690c9de2ac31e63e2f3a3b087392633a4ac555f0cd9bfe357ac9R91).

>also - does every operator automatically get an OperatorCondition, even if nothing is configured to write to it?

Yes.

pkg/controller/operators/operatorconditiongenerator_controller.go
awgreene: I went with operators because I needed to derive the component label from the operator. Given [your comment](https://github.com/operator-framework/operator-lifecycle-manager/pull/1875#discussion_r533663666) this could be rewritten to use CSVs.

pkg/controller/operators/operatorconditiongenerator_controller.go
awgreene: That's a good idea.

manifests/0000_50_olm_00-operatorconditions.crd.yaml
ecordell: Can you record the reasoning either here or in the enhancement so that we don't forget?

pkg/controller/operators/operatorconditiongenerator_controller.go
awgreene: Actually - would we be able to get the name of the deployment this way? What happens if a deployment uses a generateName?

pkg/controller/operators/operatorcondition_controller.go
```diff
@@ -0,0 +1,237 @@
...
+			if !containedEnvVar {
+				deploymentNeedsUpdate = true
+			}
+			deployment.Spec.Template.Spec.Containers[i].Env = envVars
+		}
+		if !deploymentNeedsUpdate {
+			r.log.V(1).Info("Existing deployment does not need to be updated")
+			continue
+		}
+		err = r.Client.Update(context.TODO(), deployment)
```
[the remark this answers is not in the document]
awgreene: Tested and there hasn't been an issue.

pkg/controller/operators/operatorcondition_controller.go
```diff
@@ -0,0 +1,237 @@
...
+	}, nil
+}
+
+// Implement reconcile.Reconciler so the controller can reconcile objects
+var _ reconcile.Reconciler = &OperatorConditionReconciler{}
+
+func (r *OperatorConditionReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
+	// Set up a convenient log object so we don't have to type request over and over again
+	log := r.log.WithValues("request", req)
+	log.V(1).Info("reconciling operatorcondition")
```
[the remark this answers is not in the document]
awgreene: I'll bump it up a few levels.

manifests/0000_50_olm_00-operatorconditions.crd.yaml
awgreene: I'll update the enhancement.

pkg/controller/operators/operatorconditiongenerator_controller.go
ecordell: All deployments in csvs have names given in the template

pkg/controller/operators/operatorcondition_controller.go
```diff
@@ -0,0 +1,237 @@
...
+			{
+				Verbs:         []string{"get,update,patch"},
+				APIGroups:     []string{"operators.coreos.com"},
+				Resources:     []string{"operatorconditions/status"},
+				ResourceNames: []string{operatorCondition.GetName()},
+			},
+		},
+	}
+	ownerutil.AddOwner(role, operatorCondition, false, true)
+	err := r.Client.Create(context.TODO(), role)
```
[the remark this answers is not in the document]
kevinrizza: Doesn't this also pretty much apply to any of the steps in this func? Shouldn't we just check to see that things are already there, and only create / update them if needed?

pkg/controller/operators/operatorconditiongenerator_controller.go
awgreene: Perfect, will move forward with the suggested approach.

pkg/controller/operators/operatorconditiongenerator_controller.go
awgreene: Per [this comment](https://github.com/operator-framework/operator-lifecycle-manager/pull/1875#discussion_r533663666), yes.

pkg/controller/operators/operatorconditiongenerator_controller.go
awgreene: The latest version of the OperatorConditionGenerator Controller reconciles CSVs now. The list of ServiceAccounts and Deployments added to the OperatorCondition's Spec is now based on the contents of the CSV.

pkg/controller/operators/operatorconditiongenerator_controller.go
awgreene: The latest version of the OperatorConditionGenerator Controller reconciles CSVs now.

pkg/controller/operators/operatorcondition_controller.go
```diff
@@ -0,0 +1,237 @@
...
+		err = r.Client.Update(context.TODO(), deployment)
+		if err != nil {
+			r.log.V(1).Info("error updating deployments", "err", err)
+			return err
+		}
+	}
+	return nil
+}
+
+func containsEnvVar(envVars []corev1.EnvVar, envVar corev1.EnvVar) ([]corev1.EnvVar, bool) {
```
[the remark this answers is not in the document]
awgreene: Updated to `ensureEnvVarIsPresent`

pkg/controller/operators/operatorcondition_controller.go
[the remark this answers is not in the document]
awgreene: I have updated the logging levels in the latest version of this PR.

pkg/controller/operators/operatorcondition_controller.go
```diff
@@ -0,0 +1,237 @@
...
+	}
+	ownerutil.AddOwner(roleBinding, operatorCondition, false, true)
+
+	err = r.Client.Create(context.TODO(), roleBinding)
+	if err != nil {
+		if !k8serrors.IsAlreadyExists(err) {
+			return err
+		}
+		existingRoleBinding := &rbacv1.RoleBinding{}
+		err := r.Client.Get(context.TODO(), client.ObjectKey{Name: roleBinding.GetName(), Namespace: roleBinding.GetNamespace()}, existingRoleBinding)
```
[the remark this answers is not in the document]
awgreene: Controller runtime uses an inline cache, will likely need to enable it.

pkg/controller/operators/operatorconditiongenerator_controller.go
```diff
@@ -0,0 +1,203 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	operatorsv1 "github.com/operator-framework/api/pkg/operators/v1"
+	operatorsv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/operators/decorators"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/ownerutil"
+)
+
+// OperatorReconciler reconciles a Operator object.
```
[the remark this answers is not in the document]
awgreene: Addressed in the latest version of the PR.

pkg/controller/operators/operatorconditiongenerator_controller.go
awgreene: I have removed the `owneruti.lHasOwnerReference` code in the latest version of the PR.

pkg/lib/ownerutil/util.go
awgreene: Done in the latest version of the PR.

pkg/lib/ownerutil/util.go
awgreene: Removed this function in the latest version of the PR, resolving this conversation as it is no longer relevant.

pkg/controller/operators/operatorcondition_controller.go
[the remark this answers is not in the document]
awgreene: Agreed - updated so the calls get - create / update.

pkg/controller/operators/adoption_controller.go
```diff
@@ -82,6 +82,7 @@ func (r *AdoptionReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Watches(&source.Kind{Type: &apiextensionsv1.CustomResourceDefinition{}}, enqueueProviders).
 		Watches(&source.Kind{Type: &apiregistrationv1.APIService{}}, enqueueCSV).
 		Watches(&source.Kind{Type: &operatorsv1alpha1.Subscription{}}, enqueueCSV).
+		Watches(&source.Kind{Type: &operatorsv1.OperatorCondition{}}, enqueueCSV).
```
awgreene, line 85: @njhale can you confirm that this is the correct handler?

awgreene:
/retest

Infra Issues...

dinhxuanvu (approved):
/lgtm

awgreene:
infra issues
/retest

dinhxuanvu (approved):
/lgtm

awgreene:
Failed to acquire leases.

/retest

Issue #1505, opened 2020-05-07
Catalogsource cannot be created via private docker registry


gyliu513:
## Bug Report

**What did you do?**
A clear and concise description of the steps you took (or insert a code snippet).

```yaml
apiVersion: operators.coreos.com/v1alpha1
kind: CatalogSource
metadata:
  name: my-operators
  namespace: openshift-marketplace
spec:
  displayName: My Operators
  sourceType: grpc
  image: my/catalog-test:0.0.1
```

If I push the image `my/catalog-test:0.0.1` to a private registry, then there is no place to enable me to set a imagepullsecrets for this image, and this will cause the catalogsource deploy failed.

**What did you expect to see?**
A clear and concise description of what you expected to happen (or insert a code snippet).

**What did you see instead? Under which circumstances?**
A clear and concise description of what you expected to happen (or insert a code snippet).

**Environment**
* operator-lifecycle-manager version:

* Kubernetes version information:

* Kubernetes cluster kind: 

**Possible Solution**

**Additional context**

gyliu513:
Can any expert show some comments for this? Thanks!

skuffe:
Take a look at #1294

greglanthier:
I'm stumbling on this as well with OLM v 0.16.1 in a plain vanilla `kind` cluster & `operator-sdk` v 1.0.1.

In my case I declared a CatalogSource like this:

```yaml
apiVersion: operators.coreos.com/v1alpha1
kind: CatalogSource
metadata:
  name: cloudcar-catalog
  namespace: foo
spec:
  displayName: Cloudcar Operators
  sourceType: grpc
  publisher: cloudcar
  image: us.icr.io/ca-sci/cloudcar-operators-index:latest
  secrets:
    - registry-read
  updateStrategy:
    registryPoll:
      interval: 10m
```

assuming that the pre-existing `registry-read` pull secret in the `foo` namespace would be used to pull the index image from my private registry.  Instead the image pulls failed with events like this:

```bash
Failed to pull image "us.icr.io/ca-sci/cloudcar-operators-index:latest": rpc error: code = Unknown desc = failed to pull and unpack image "us.icr.io/ca-sci/cloudcar-operators-index:latest": failed to resolve reference "us.icr.io/ca-sci/cloudcar-operators-index:latest": failed to authorize: failed to fetch anonymous token: unexpected status: 401 Unauthorized
```

[...]

PR #1878, opened 2020-11-24
Add authentication for private index images

### LINKED ISSUE #1536: secret does not work with catalogsource
## Bug Report

**What did you do?**
A clear and concise description of the steps you took (or insert a code snippet).

```yaml
kind: CatalogSource
metadata:
  name: my-catalog
  namespace: default
spec:
  name: my-catalog
  sourceType: grpc
  image: <your catalog image>
  secrets:
  - my-pull-secret
```

I found the secrets does not work for my cluster, please checkout https://github.com/operator-framework/operator-lifecycle-manager/issues/1294#issuecomment-625570581 for detail.

FYI @hassenius @sethpeterson @jeyaramashok 

A clear and concise description of what you expected to happen (or insert a code snippet).

A clear and concise description of what you expected to happen (or insert a code snippet).

**Environment**

* Kubernetes cluster kind: 

**Possible Solution**

**Additional context**

anik120:

Closes #1536  #1505 

This PR attaches the secrets in the `spec.Secrets` field of the 
CatalogSource to the default service account in the CatalogSource's
namespace, to allow for pulling of index images that are backed 
by authentication.

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/registry/reconciler/reconciler.go
+++ b/pkg/controller/registry/reconciler/reconciler.go
@@ -91,7 +91,7 @@ func NewRegistryReconcilerFactory(lister operatorlister.OperatorLister, opClient
 	}
 }
 
-func Pod(source *v1alpha1.CatalogSource, name string, image string, labels map[string]string, readinessDelay int32, livenessDelay int32) *v1.Pod {
+func Pod(source *v1alpha1.CatalogSource, name string, image string, saName string, labels map[string]string, readinessDelay int32, livenessDelay int32) *v1.Pod {
 	// Ensure the catalog image is always pulled if the image is not based on a digest, measured by whether an "@" is included.
 	// See https://github.com/docker/distribution/blob/master/reference/reference.go for more info.
 	// This means recreating non-digest based catalog pods will result in the latest version of the catalog content being delivered on-cluster.
@@ -148,6 +148,7 @@ func Pod(source *v1alpha1.CatalogSource, name string, image string, labels map[s
 			NodeSelector: map[string]string{
 				"kubernetes.io/os": "linux",
 			},
+			ServiceAccountName: saName,
 		},
 	}
 	return pod
--- a/pkg/controller/operators/catalog/operator_test.go
+++ b/pkg/controller/operators/catalog/operator_test.go
@@ -1424,7 +1424,7 @@ func toManifest(t *testing.T, obj runtime.Object) string {
 }
 
 func pod(s v1alpha1.CatalogSource) *corev1.Pod {
-	pod := reconciler.Pod(&s, "registry-server", s.Spec.Image, s.GetLabels(), 5, 10)
+	pod := reconciler.Pod(&s, "registry-server", s.Spec.Image, s.GetName(), s.GetLabels(), 5, 10)
 	ownerutil.AddOwner(pod, &s, false, false)
 	return pod
 }
--- a/pkg/controller/registry/reconciler/configmap.go
+++ b/pkg/controller/registry/reconciler/configmap.go
@@ -95,7 +95,7 @@ func (s *configMapCatalogSourceDecorator) Service() *v1.Service {
 }
 
 func (s *configMapCatalogSourceDecorator) Pod(image string) *v1.Pod {
-	pod := Pod(s.CatalogSource, "configmap-registry-server", image, s.Labels(), 5, 2)
+	pod := Pod(s.CatalogSource, "configmap-registry-server", image, "", s.Labels(), 5, 2)
 	pod.Spec.ServiceAccountName = s.GetName() + ConfigMapServerPostfix
 	pod.Spec.Containers[0].Command = []string{"configmap-server", "-c", s.Spec.ConfigMap, "-n", s.GetNamespace()}
 	ownerutil.AddOwner(pod, s.CatalogSource, false, false)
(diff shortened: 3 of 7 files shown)


pkg/controller/registry/reconciler/reconciler.go
```diff
@@ -150,5 +150,12 @@ func Pod(source *v1alpha1.CatalogSource, name string, image string, labels map[s
 			},
 		},
 	}
+	if len(source.Spec.Secrets) > 0 {
+		pod.Spec.ImagePullSecrets = []v1.LocalObjectReference{
+			{
+				Name: source.Spec.Secrets[0],
```
exdx, line 156: Per https://github.com/operator-framework/api/blob/master/crds/operators.coreos.com_catalogsources.yaml#L110 the secrets field is an array, I'm not sure we have explicitly documented that the first secret name in the array is the secret for the catalog pod specifically. I think it makes sense, but it could break existing users of the `secrets` field who rely on it to copy secrets from the OLM namespace to the operator's namespace for a variety of uses. So we have to be careful about backwards compatibility with a change like this

anik120: @exdx exactly what I had in mind too. Which is why I think it's better to accept this secret with a new `spec.imageSecret` field, for which I have opened up a PR here https://github.com/operator-framework/api/pull/79

pkg/controller/registry/reconciler/reconciler.go
```diff
@@ -150,5 +150,12 @@ func Pod(source *v1alpha1.CatalogSource, name string, image string, labels map[s
 			},
 		},
 	}
+	if len(source.Spec.Secrets) > 0 {
+		pod.Spec.ImagePullSecrets = []v1.LocalObjectReference{
```
njhale, line 154: I'd like to see what it looks like if we bind the secrets to the pod's service account rather than the pod itself: https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#add-imagepullsecrets-to-a-service-account

pkg/controller/registry/reconciler/reconciler.go
njhale: > I'm not sure we have explicitly documented that the first secret name in the array is the secret for the catalog pod specifically.

We should actually be adding all the secrets rather than just the first one.

njhale (changes requested):
Thanks @anik120 for picking this up.

Please see https://github.com/operator-framework/api/pull/79#discussion_r530544680 for my thoughts on the approach and an alternate solution.

pkg/controller/registry/reconciler/reconciler.go
exdx: From what I can see from the doc, if we were to bind image pull secrets to the default SA in the catalog pod namespace all pods using that default SA would get the `imagePullSecrets` field set automatically, regardless of whether they are catalogs or just other random pods in that namespace. Is that accurate? Per https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#verify-imagepullsecrets-was-added-to-pod-spec

Would we use the default SA or a new individual SA for each custom catalog?

anik120: >I'd like to see what it looks like if we bind the secrets to the pod's service account rather than the pod itself:

Which service account do you mean exactly by the `pod's service account`? This is what I can see is created when I create a new CatalogSource: 

```
$ kubectl get pods -n olm 
NAME                                                READY   STATUS    RESTARTS   AGE
catalog-operator-7fb7fb69d9-6r8hf       1/1        Running             0          25h
my-catalog-sqbf2                                  1/1        Running             0          25h
olm-operator-69f46ff69d-5kvg7            1/1        Running             0          25h
operatorhubio-catalog-9wb2x               1/1        Running             0          25h
packageserver-59dc8f9447-v4xpn       1/1        Running             0          25h 

$ k get serviceaccount -n olm       
NAME                                            SECRETS   AGE
default                                                  1            24h
olm-operator-serviceaccount               1            24h

$ k get service -n olm 
NAME                                TYPE        CLUSTER-IP      EXTERNAL-IP   PORT(S)     AGE
my-catalog                      ClusterIP     10.96.199.188         <none>        50051/TCP    25h
operatorhubio-catalog    ClusterIP     10.96.141.232          <none>        50051/TCP    25h
packageserver-service   ClusterIP     10.96.95.185            <none>        5443/TCP      25h
```

test.txt
```diff
@@ -0,0 +1,3467 @@
+KUBEBUILDER_ASSETS=/usr/local/kubebuilder/bin go test -mod=vendor  -tags "json1" -v -race -count=1 ./pkg/...
```
kevinrizza, line 1: Was this file included by accident? It seems unnecessary

anik120: Oops

anik120: @kevinrizza removed.

exdx:
I think the implementation here looks good, if this is the way we choose to go. 

The concern I raised in https://github.com/operator-framework/operator-lifecycle-manager/pull/1878#discussion_r530579642 or as @ecordell put it 

> adding them [catalog secrets] to the [default] serviceaccount means that every pod using the default serviceaccount will need to rotate through the list of credentials in order to pull

is a valid concern. Is creating a per-catalog SA and using it too much of a change in behavior? It at least isolates the secrets a bit more. 
/lgtm

anik120:
>The concern I raised in #1878 (comment) or as @ecordell put it
>
>adding them [catalog secrets] to the [default] serviceaccount means that every pod using the default serviceaccount will need to rotate through the list of credentials in order to pull
>
>is a valid concern. Is creating a per-catalog SA and using it too much of a change in behavior? It at least isolates the secrets a >bit more.
@exdx @ecordell updated the PR to use per catalog SA instead of the default SA. PTAL, thanks!

exdx:
/lgtm

per the docs
> The service account has to exist at the time the pod is created, or it will be rejected.
So just to be clear, for the openshift default catalogs, those are still managed by the marketplace operator right? Those have their own authentication not based on SAs. So this change is for custom catalogs only. 

I think this is good, once we get some good docs on the `Secrets` field and its use people should have an easier time with using private catalogs and their contents.

PR #1883, opened 2020-12-02
Add OLM support for the Upgradeable OperatorCondition

dinhxuanvu:
OLM will check for OperatorCondition CR for OperatorUpgradeable status.
The pending CSV will not transition to Succeeded status until the operator
is upgradeable based on the OperatorCondition. If the operator doesn't use
OperatorCondition, the normal transition is expected.

**Reviewer Checklist**


dinhxuanvu:
Need #1875 to merge first.

pkg/controller/operators/olm/operatorconditions.go
```diff
@@ -0,0 +1,35 @@
...
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
+			return true, nil
+		}
+		return false, err
+	}
+
+	if len(cond.Status.Conditions) < 1 {
+		return false, fmt.Errorf("No operator conditions are available")
+	}
```
awgreene, line 29: If the upgradeable condition is not present I believe we should fallback on default behavior.

awgreene (commented):
Nice work @dinhxuanvu!

dinhxuanvu:
Close in favor of new PR #1885

PR #1884, opened 2020-12-02
Bug 1898500: Support InstallPlan steps upgrading existing ClusterIP Services.

benluddy:
If a ClusterIP-type Service manifest omits the immutable spec field
"ClusterIP", a value is selected automatically when the Service is
created. The catalog operator can patch existing Service resources if
the updated manifest either omits "ClusterIP" or sets "ClusterIP" to
the same value as that of the existing Service.

### DIFF
--- a/pkg/lib/operatorclient/service.go
+++ b/pkg/lib/operatorclient/service.go
@@ -28,11 +28,22 @@ func (c *Client) DeleteService(namespace, name string, options *metav1.DeleteOpt
 // UpdateService will update the given Service resource.
 func (c *Client) UpdateService(service *v1.Service) (*v1.Service, error) {
 	klog.V(4).Infof("[UPDATE Service]: %s", service.GetName())
-	oldSa, err := c.GetService(service.GetNamespace(), service.GetName())
+	old, err := c.GetService(service.GetNamespace(), service.GetName())
 	if err != nil {
 		return nil, err
 	}
-	patchBytes, err := createPatch(oldSa, service)
+	normalized, err := cloneAndNormalizeObject(old)
+	if err != nil {
+		return nil, fmt.Errorf("failed to normalize existing Service resource for patch: %w", err)
+	}
+	if service.Spec.ClusterIP == old.Spec.ClusterIP {
+		// Support updating to manifests that specify a
+		// ClusterIP when its value is the same as that of the
+		// existing Service.
+		service = service.DeepCopy()
+		service.Spec.ClusterIP = ""
+	}
+	patchBytes, err := createPatch(normalized, service)
 	if err != nil {
 		return nil, fmt.Errorf("error creating patch for Service: %v", err)
 	}
--- a/pkg/lib/operatorclient/service_test.go
+++ b/pkg/lib/operatorclient/service_test.go
@@ -1,0 +1,187 @@
+package operatorclient
+
+import (
+	"testing"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/kubernetes/fake"
+	clienttesting "k8s.io/client-go/testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestUpdateService(t *testing.T) {
+	gvr := schema.GroupVersionResource{Group: "", Version: "v1", Resource: "services"}
+
+	// In a test expectation, matches any single fake client action.
+	var wildcard clienttesting.Action = clienttesting.ActionImpl{Verb: "wildcard!"}
+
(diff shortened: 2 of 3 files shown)


pkg/lib/operatorclient/service_test.go
```diff
@@ -0,0 +1,160 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/kubernetes/fake"
+	clienttesting "k8s.io/client-go/testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestUpdateService(t *testing.T) {
+	gvr := schema.GroupVersionResource{Group: "", Version: "v1", Resource: "services"}
+	var wildcard clienttesting.Action = clienttesting.ActionImpl{Verb: "wildcard!"}
```
exdx, line 18: what's the reasoning behind this var? looks interesting.

pkg/lib/operatorclient/service.go
```diff
@@ -28,11 +28,22 @@ func (c *Client) DeleteService(namespace, name string, options *metav1.DeleteOpt
...
+	normalized, err := cloneAndNormalizeObject(old)
+	if err != nil {
+		return nil, fmt.Errorf("failed to normalize existing Service resource for patch: %w", err)
+	}
+	if service.Spec.ClusterIP == old.Spec.ClusterIP {
+		// Support updating to manifests that specify a
+		// ClusterIP when its value is the same as that of the
+		// existing Service.
+		service = service.DeepCopy()
+		service.Spec.ClusterIP = ""
```
exdx, line 44: A little confused -- the objective is to ensure the `ClusterIP` remains the same for the new service as the old service right? Even if other things about the service changed. 

`cloneAndNormalizeObject` strips the `ClusterIP` from the old service. This line removes the `ClusterIP` from the new service. This way, when we create the patch, we get a Service with no `ClusterIP` set. Then when we go to patch the existing service on the cluster with the patch, it preserves the existing `ClusterIP` of the service on cluster? But OLM can potentially update other Service fields. Makes sense.

pkg/lib/operatorclient/service.go
```diff
@@ -28,11 +28,22 @@ func (c *Client) DeleteService(namespace, name string, options *metav1.DeleteOpt
...
+	}
+	patchBytes, err := createPatch(normalized, service)
```
exdx, line 46: Would there be an issue applying the patch in the case where the new service specifies a `ClusterIP` whereas the old one does not?

benluddy: Yes, that should fail. My thinking was that there are two scenarios where a patch shouldn't touch `ClusterIP`:

- v1 and v2 omit `ClusterIP`
- v1 and v2 both set the same `ClusterIP` value

For the case where v1 omits `ClusterIP` and v2 does not, I'm not convinced that there is a clear automatic decision to be made. We could go ahead and ignore the fact that v2 specifies a certain IP, but it's plausible that for some reason the operator depends on the v2 service having exactly that IP, so I felt that it was best in that case to force a person to step in and decide what to do.

pkg/lib/operatorclient/service.go
benluddy: That's right -- my intention was to limit this to only allow thetwo cases I mentioned in my other [comment](https://github.com/operator-framework/operator-lifecycle-manager/pull/1884#discussion_r534483523).

pkg/lib/operatorclient/service_test.go
benluddy: If this is referenced in one of the test cases, it matches any single call to the fake client. I wanted to avoid maintaining test expectations for the uninteresting Get calls.

pkg/lib/operatorclient/service_test.go
```diff
@@ -0,0 +1,187 @@
...
+					ClusterIP: "1.2.3.4",
+				},
+			},
+			Expected: []clienttesting.Action{
+				wildcard,
+				clienttesting.NewPatchAction(gvr, "namespace", "name", types.StrategicMergePatchType, []byte(`{}`)),
+			},
+		},
+		{
+			Name: "clusterip patched if changed", // even though the patch will be rejected due to field immutability
```
benluddy, line 115: @exdx I just added this test case for [the scenario you mentioned ](https://github.com/operator-framework/operator-lifecycle-manager/pull/1884/files#r534467007) so that it's clear that this is intentional.

pkg/lib/operatorclient/service_test.go
exdx: Interesting, will have to check that out. Haven't seen it before

pkg/lib/operatorclient/service.go
exdx: I think that makes sense. I feel like the behavior could be somewhat dependent on the service `Type` -- maybe a loadbalanced or external service would expect different upgrade behavior than a cluster-internal one. I think we can keep an eye on this and build out more robust service upgrades upstream (if it arises) or downstream.

PR #1885, opened 2020-12-03
Add OLM support for the Upgradeable OperatorCondition and Admin overrides

dinhxuanvu:
OLM will check for OperatorCondition CR for OperatorUpgradeable status.
The pending CSV will not transition to Succeeded status until the operator
is upgradeable based on the OperatorCondition. If the operator doesn't use
OperatorCondition, the normal transition is expected.

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/olm/operator.go
+++ b/pkg/controller/operators/olm/operator.go
@@ -220,6 +220,22 @@ func newOperatorWithConfig(ctx context.Context, config *operatorConfig) (*Operat
 			return nil, err
 		}
 
+		// Register OperatorCondition QueueInformer
+		opConditionInformer := extInformerFactory.Operators().V1().OperatorConditions()
+		op.lister.OperatorsV1().RegisterOperatorConditionLister(namespace, opConditionInformer.Lister())
+		opConditionQueueInformer, err := queueinformer.NewQueueInformer(
+			ctx,
+			queueinformer.WithLogger(op.logger),
+			queueinformer.WithInformer(opConditionInformer.Informer()),
+			queueinformer.WithSyncer(k8sSyncer),
+		)
+		if err != nil {
+			return nil, err
+		}
+		if err := op.RegisterQueueInformer(opConditionQueueInformer); err != nil {
+			return nil, err
+		}
+
 		subInformer := extInformerFactory.Operators().V1alpha1().Subscriptions()
 		op.lister.OperatorsV1alpha1().RegisterSubscriptionLister(namespace, subInformer.Lister())
 		subQueueInformer, err := queueinformer.NewQueueInformer(
@@ -1426,6 +1442,15 @@ func (a *Operator) transitionCSVState(in v1alpha1.ClusterServiceVersion) (out *v
 		logger.Info("scheduling ClusterServiceVersion for requirement verification")
 		out.SetPhaseWithEvent(v1alpha1.CSVPhasePending, v1alpha1.CSVReasonRequirementsUnknown, "requirements not yet checked", now, a.recorder)
 	case v1alpha1.CSVPhasePending:
+		// Check previous version's Upgradeable condition
+		replacedCSV := a.isReplacing(out)
+		if replacedCSV != nil {
+			operatorUpgradeable, condErr := a.isOperatorUpgradeable(replacedCSV)
+			if !operatorUpgradeable {
+				out.SetPhaseWithEventIfChanged(v1alpha1.CSVPhasePending, v1alpha1.CSVReasonOperatorConditionNotUpgradeable, fmt.Sprintf("operator is not upgradeable: %s", condErr), now, a.recorder)
+				return
+			}
+		}
 		met, statuses, err := a.requirementAndPermissionStatus(out)
 		if err != nil {
 			// TODO: account for Bad Rule as well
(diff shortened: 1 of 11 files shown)


pkg/controller/operators/olm/operatorconditions.go
```diff
@@ -0,0 +1,35 @@
...
+
+	cond, err := a.lister.OperatorsV1().OperatorConditions(csv.GetNamespace()).Get(context.TODO(), csv.GetName(), metav1.GetOptions{})
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
+			return true, nil
+		}
+		return false, err
+	}
+
+	if len(cond.Status.Conditions) < 1 {
```
ecordell, line 38: @dinhxuanvu @awgreene Just want you to sanity check the interaction of these two parts.

It looks like:

1. An OperatorCondition is generated for every operator installed
2. There's a requirement that the condition be set to upgradeable=true in order to install successfully

Doesn't that mean that for operators that don't support / know how to update an OperatorCondition, their CSV's will never go to succeeded?

awgreene: This if statement returns `true`.

pkg/controller/operators/olm/operator.go
```diff
@@ -1824,6 +1818,12 @@ func (a *Operator) updateInstallStatus(csv *v1alpha1.ClusterServiceVersion, inst
 		return strategyErr
 	}
 
+	if condErr != nil {
+		a.logger.WithError(condErr).Debug("operator is not upgradeable")
+		csv.SetPhaseWithEventIfChanged(v1alpha1.CSVPhasePending, v1alpha1.CSVReasonOperatorConditionNotUpgradeable, "operator upgradeable condition is false", now, a.recorder)
```
ecordell, line 1823: I don't think we want to flip back to pending - that will attempt to recreate / reinstall components. Shouldn't it stay succeeded, just not upgrade?

awgreene: I agree with @ecordell

pkg/controller/operators/olm/operatorconditions.go
ecordell: And what ensures that an operator has a chance to set `Upgradeable=false` before this check?

ecordell:
Don't we also want something that can prevent the resolver from looking for updates if an operator is not upgradable?

dinhxuanvu:
> Don't we also want something that can prevent the resolver from looking for updates if an operator is not upgradable?

We talked about this (Ben asked about this) and we agreed that at this point we don't want the resolver to be involved into this. This is meant to be a separate part of the resolver and it simply prevents the operator proceed to Succeeded state but not blocking the entire resolution altogether.

pkg/controller/operators/olm/operatorconditions.go
dinhxuanvu: We can use the timeStamp for this purpose.

ecordell: Synced offline - 

There will be an expectation that any operator that opts into this feature will update the OperatorCondition before their readiness probe returns true.

pkg/controller/operators/olm/operatorconditions.go
```diff
@@ -0,0 +1,48 @@
+package olm
+
+import (
+	"fmt"
+
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	meta "k8s.io/apimachinery/pkg/api/meta"
```
awgreene, line 7: Can we just import meta or metav1?

awgreene (commented):
Great work @dinhxuanvu !

pkg/controller/operators/olm/operatorconditions.go
```diff
@@ -0,0 +1,45 @@
...
+	cond, err := a.lister.OperatorsV1().OperatorConditionLister().OperatorConditions(csv.GetNamespace()).Get(csv.GetName())
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
+			return true, nil
+		}
+		return false, err
+	}
+
+	// Check condition overrides
+	for _, override := range cond.Spec.Overrides {
```
ecordell, line 28: I think a better ux (requiring a webhook) would be to have `spec.overrides` actually change the value of `status` to match the overridden value, and then reject any external changes to status while the override is set (this is the part that requires a webhook).

Otherwise anyone else that is not OLM (including e.g. the UI) needs to re-implement this union logic. And the problem will grow as more supported conditions are added.

pkg/lib/operatorclient/customresources.go
```diff
@@ -154,6 +154,49 @@ func (c *Client) UpdateCustomResourceRaw(apiGroup, version, namespace, resourceP
 	return nil
 }
 
+// UpdateCustomResourceStatus updates the custom resource's status.
+func (c *Client) UpdateCustomResourceStatus(item *unstructured.Unstructured) error {
```
ecordell, line 158: The operatorclient is more or less deprecated in favor of direct calls via the generated clients or the unstructured client. Can we just remove this? (less to test, too)

dinhxuanvu: Oh this code is not in use anymore and I should discard this. I fully use the OC client to create/update CR.

pkg/controller/operators/olm/operatorconditions.go
dinhxuanvu: Acked.

PR #1887, opened 2020-12-03
Bug 1907329: Update /manifests with default cluster profile annotation

guillaumerose:
* Add option to yq to modify all documents in a single file
* Updated with `make release`

### DIFF
--- a/manifests/0000_50_olm_00-operatorgroups.crd.yaml
+++ b/manifests/0000_50_olm_00-operatorgroups.crd.yaml
@@ -3,6 +3,7 @@ kind: CustomResourceDefinition
 metadata:
   annotations:
     controller-gen.kubebuilder.io/version: v0.4.1
+    include.release.openshift.io/self-managed-high-availability: "true"
   creationTimestamp: null
   name: operatorgroups.operators.coreos.com
 spec:
@@ -230,4 +231,3 @@ spec:
       storage: false
       subresources:
         status: {}
-
--- a/manifests/0000_50_olm_07-olm-operator.deployment.yaml
+++ b/manifests/0000_50_olm_07-olm-operator.deployment.yaml
@@ -5,6 +5,8 @@ metadata:
   namespace: openshift-operator-lifecycle-manager
   labels:
     app: olm-operator
+  annotations:
+    include.release.openshift.io/self-managed-high-availability: "true"
 spec:
   strategy:
     type: RollingUpdate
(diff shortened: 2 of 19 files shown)


manifests/0000_50_olm_07-olm-operator.deployment.yaml
```diff
@@ -22,18 +24,18 @@ spec:
...
-          - $(OPERATOR_NAMESPACE)
-          - --writeStatusName
-          - operator-lifecycle-manager
-          - --writePackageServerStatusName
-          - operator-lifecycle-manager-packageserver
-          - --tls-cert
-          - /var/run/secrets/serving-cert/tls.crt
-          - --tls-key
-          - /var/run/secrets/serving-cert/tls.key
+            - --namespace
```
kevinrizza, line 29: Seems like there's a lot of churn on spacing here? Can we avoid that?

manifests/0000_50_olm_00-operatorgroups.crd.yaml
```diff
@@ -3,303 +3,231 @@ kind: CustomResourceDefinition
...
     kind: OperatorGroup
     listKind: OperatorGroupList
     plural: operatorgroups
     shortNames:
-    - og
+      - og
     singular: operatorgroup
   scope: Namespaced
   versions:
-  - name: v1
```
kevinrizza, line 21: This also seems like a lot of churn

kevinrizza (changes requested):

guillaumerose:
GH actions are green, thanks! Now, we need the openshift CI 🙏

eparis:
```
pods/openshift-cluster-version_cluster-version-operator-54bb4c667c-kncbw_cluster-version-operator.log.gz:E1214 22:16:36.916095       1 runtime.go:78] Observed a panic: &errors.errorString{s:"v1.ClusterOperator.ObjectMeta: v1.ObjectMeta.Annotations: ReadString: expects \" or n, but found t, error found in #10 byte of ...|ability\":true},\"name|..., bigger context ...|ase.openshift.io/self-managed-high-availability\":true},\"name\":\"operator-lifecycle-manager\"},\"status\"|..."} (v1.ClusterOperator.ObjectMeta: v1.ObjectMeta.Annotations: ReadString: expects " or n, but found t, error found in #10 byte of ...|ability":true},"name|..., bigger context ...|ase.openshift.io/self-managed-high-availability":true},"name":"operator-lifecycle-manager"},"status"|...)
```

It is because you are trying to set `true` instead of `"true"` so it's being treated as a boolean, not as a string, and it can't parse correctly?

eparis:
/hold
no need to keep wasting CI cycles trying to parse a bool as a strong. feel free to unhold when you get CI fixed

guillaumerose:
I added the double quote. Let's see how it goes now. Thanks @eparis for spotting the error!

guillaumerose:
/hold

CI is in trouble right now.

dinhxuanvu:
@guillaumerose There are a few e2e test cases that are failing consistently. Can you run those test locally to see they are passing or not?

guillaumerose:
@dinhxuanvu these failures are the same I observed in others PRs. For example, https://github.com/operator-framework/operator-lifecycle-manager/pull/1913

guillaumerose:
All green! Thanks @awgreene for fixing the CI.

PR #1888, opened 2020-12-03
Bug 1913085: use OLM client when installing CRDs

exdx:
Always use the OLM client when installing CRDs for operators using a scoped service account referenced in the operatorgroup

Regression when adding support for v1 CRDs
See https://docs.openshift.com/container-platform/4.5/operators/admin/olm-creating-policy.html for expected behavior

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -1578,7 +1578,16 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
 
 	ensurer := newStepEnsurer(kubeclient, crclient, dynamicClient)
 	r := newManifestResolver(plan.GetNamespace(), o.lister.CoreV1().ConfigMapLister(), o.logger)
-	b := newBuilder(kubeclient, dynamicClient, o.csvProvidedAPIsIndexer, r, o.logger)
+
+	// CRDs should be installed via the default OLM (cluster-admin) client and not the scoped client specified by the AttenuatedServiceAccount
+	// the StepBuilder is currently only implemented for CRD types
+	// TODO give the StepBuilder both OLM and scoped clients when it supports new scoped types
+	builderKubeClient, _, builderDynamicClient, err := o.clientAttenuator.AttenuateClientWithServiceAccount(nil)
+	if err != nil {
+		o.logger.Errorf("failed to get a client for plan execution- %v", err)
+		return err
+	}
+	b := newBuilder(builderKubeClient, builderDynamicClient, o.csvProvidedAPIsIndexer, r, o.logger)
 
 	for i, step := range plan.Status.Plan {
 		doStep := true
--- a/test/e2e/installplan_e2e_test.go
+++ b/test/e2e/installplan_e2e_test.go
@@ -22,6 +22,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -2803,7 +2804,8 @@ var _ = Describe("Install Plan", func() {
 			Expect(ref.Kind).To(Equal("ConfigMap"))
 		}
 	})
-	It("uses the OLM client when installing CRDs from a bundle", func() {
+
+	It("limits installed resources if the scoped serviceaccount has no permissions", func() {
 		c := newKubeClient()
 		crc := newCRClient()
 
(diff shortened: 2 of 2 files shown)


pkg/controller/operators/catalog/operator.go
```diff
@@ -1578,7 +1578,16 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
 
 	ensurer := newStepEnsurer(kubeclient, crclient, dynamicClient)
 	r := newManifestResolver(plan.GetNamespace(), o.lister.CoreV1().ConfigMapLister(), o.logger)
-	b := newBuilder(kubeclient, dynamicClient, o.csvProvidedAPIsIndexer, r, o.logger)
+
+	// CRDs should be installed via the default OLM (cluster-admin) client and not the scoped client specified by the AttenuatedServiceAccount
+	// the StepBuilder is currently only implemented for CRD types
+	// TODO give the StepBuilder both OLM and scoped clients when it supports new scoped types
+	builderKubeClient, _, builderDynamicClient, err := o.clientAttenuator.AttenuateClientWithServiceAccount(nil)
```
ecordell, line 1585: this can be a follow up - 

is there a related bug where we won't grant the operator permission to manage its own types because the scoped serviceaccount doesn't have rbac for the new types?

exdx: I can check but I thought this would be expected behavior instead of a bug? Or am I missing something

exdx:
Will move forward without the automatic BZ automation via an override for backports.

test/e2e/installplan_e2e_test.go
```diff
@@ -2715,6 +2715,273 @@ var _ = Describe("Install Plan", func() {
...
+		// create role that includes permission for installing the kiali operator, but not CRDs
+		// see https://github.com/operator-framework/operator-lifecycle-manager/blob/master/doc/design/scoped-operator-install.md
+		role := &rbacv1.ClusterRole{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: genName("role-"),
+			},
+			Rules: []rbacv1.PolicyRule{
+				{
+					APIGroups: []string{"operators.coreos.com"},
+					Resources: []string{"subscriptions", "clusterserviceversions"},
```
njhale, line 2750: This brings up another good question: shouldn't resolvable OLM kinds _always_ be applied with OLM's privileges? 

Requiring explicit permissions for OLM resources (of the OperatorGroup ServiceAccount) seems like _YetAnotherHoop_ ™️

pkg/controller/operators/catalog/operator.go
njhale, line 1585: It might be cool long-term if we had an attenuated client that could conditionally proxy to other attenuated clients based on some rules; e.g. GroupKind -> Client. This would let us make attenuation opaque to consumers (like the thing applying the steps).

But this is something to follow up on if anything.

njhale (commented):
Looking good!

Only nits and some questions on the e2e test.

test/e2e/installplan_e2e_test.go
```diff
@@ -2715,6 +2715,273 @@ var _ = Describe("Install Plan", func() {
...
+				Name:      genName("sa-"),
+				Namespace: ns.GetName(),
+			},
+		}
+		_, err = c.KubernetesInterface().CoreV1().ServiceAccounts(ns.GetName()).Create(context.TODO(), sa, metav1.CreateOptions{})
+		Expect(err).ToNot(HaveOccurred())
+
+		// create role that includes permission for installing the kiali operator, but not CRDs
+		// see https://github.com/operator-framework/operator-lifecycle-manager/blob/master/doc/design/scoped-operator-install.md
+		role := &rbacv1.ClusterRole{
```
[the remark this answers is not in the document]
exdx: I decided to go with this kiali bundle because I didn't want to introduce a new image dependency for this test. It's also representative of an actual operator installation so I think it's potentially helpful in future tests.

I agree that the RBAC is rather bloated, but I'd rather not rewrite the test solely for this reason. Since this is test code it's not in the compiled binary anyway.  

I could move the RBAC to another file if that's helpful

test/e2e/installplan_e2e_test.go
exdx: I totally agree. Why does an operator's SA need knowledge of OLM's APIs to be installed by OLM is unclear -- it would make more sense if this was implied, as you said by using the OLM client for these types. 

I got these from https://docs.openshift.com/container-platform/4.5/operators/admin/olm-creating-policy.html#olm-policy-fine-grained-permissions_olm-creating-policy 

Seems like using your own SA in the OG means you must acknowledge the OLM types - since this is more advanced behavior I can see why this makes sense as is, but yeah changing it could make sense

dinhxuanvu (approved):
/lgtm

test/e2e/installplan_e2e_test.go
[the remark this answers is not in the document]
benluddy: > write a CRD to an InstallPlan's status directly (forgo resolution)

I'd like to see this, too. If the behavior verified by this test is CRD creation in the presence of a scoped SA, then doing resolution and unpacking makes the test itself more fragile (i.e., it can fail in many ways that do not indicate a problem with the behavior under test).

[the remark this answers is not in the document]
exdx: Alright, I will rewrite the e2e test to make an installplan instead

exdx:
Depending on https://github.com/openshift/console/pull/7835 for console tests to pass

PR #1891, opened 2020-12-04
fix: use annotation on the operatorgroup to determine csv copying

exdx:
Enabled CSV copying behavior to be modified at runtime via an annotation on the operatorgroup

Fine-grained CSV copying behavior 

**Reviewer Checklist**


exdx:
Closing for an alternate implementation - update the copied CSV only based on spec changes, and add a small amount of status on the copy - phase, reason, message indicating the CSV was copied. 

Part of the reasoning is to encourage users to use the new operator API to discover operators on the cluster (and their associated status) instead of looking at the copied csvs in their namespace

PR #1892, opened 2020-12-04
Bug 1905599: Preserve original .status.lastUpdateTime in copied CSVs.

benluddy:
Currently, CSVs are mirrored into the namespaces that are watched by
the operator they represent. At the moment a copied CSV is updated,
its .status.lastUpdateTime field is set to the current time, rather
than using the same timestamp that appears on the original CSV's
status. As the size of a cluster grows, it becomes increasingly
unlikely that these two timestamps will have the same value. This
means that there's almost always a non-empty diff between the original
and any given copy, thereby triggering an update to the copy even if
nothing else has changed in the original.


dinhxuanvu:
This is a good stop-gap solution for the copied CSV issue.

dinhxuanvu (approved):
/lgtm

benluddy:
Patch for unrelated test flake is up: https://github.com/operator-framework/operator-lifecycle-manager/pull/1896.

/retest

PR #1893, opened 2020-12-07
add 'single-node-production-edge' annotations to CVO manifests.

osherdp:
This adds annotations for the single-node-production-edge cluster profile. There's a growing requirement from several customers to enable creation of single-node (not high-available) Openshift clusters.
In stage one (following openshift/enhancements#504) there should be no implication on components logic.
In the next stage, the component's behavior will match a non high-availability profile if the customer is specifically interested in one.
This PR is separate from the 'single-node-developer' work, which will implement a different behavior and is currently on another stage of implementation.

For more info, please refer to the enhancement link and participate in the discussion.


osherdp:
Shifting strategy based on https://github.com/openshift/enhancements/pull/560

PR #1899, opened 2020-12-09
Bug 1906134: Don't create OperatorConditions for copied CSVs

awgreene:
Problem: OLM is currently creating an OperatorCondition for copied CSVs.
As a result, OLM is reporting that the deployments defined in the copied
CSVs cannot be found.

Solution: OLM should not be creating OperatorCondtions for copied CSVs.
Update the OperatorConditionGenerator Reconciler to include predicate
functions that exclude copied CSVs.

### DIFF
--- a/pkg/controller/operators/operatorconditiongenerator_controller.go
+++ b/pkg/controller/operators/operatorconditiongenerator_controller.go
@@ -9,8 +9,11 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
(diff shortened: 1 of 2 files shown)


pkg/controller/operators/operatorconditiongenerator_controller.go
```diff
@@ -34,9 +37,35 @@ func (r *OperatorConditionGeneratorReconciler) SetupWithManager(mgr ctrl.Manager
...
+			if _, ok := e.Meta.GetLabels()[operatorsv1alpha1.CopiedLabelKey]; ok {
+				return false
+			}
+			return true
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			if _, ok := e.MetaOld.GetLabels()[operatorsv1alpha1.CopiedLabelKey]; ok {
+				return false
+			}
+			return e.ObjectOld != e.ObjectNew
```
benluddy, line 57: Does `!=` have the right semantics for this?

awgreene: I had copied this bit from the [example code](controller ) in kubebuilder, I'm fine returning true though.

benluddy: Yeah, I suppose even if for whatever reason you have two value-equal objects for which `!=` evaluates to false, nothing bad happens. Returning `true` should be practically the same, and might save people like me from pausing to think about it, but I'm good with it either way. Thanks!

awgreene: In the interest of improved clarity I am updating this return statement to always return true.

kevinrizza:
Tests?

awgreene:
>Tests?

I can create a test that ensures an OpeneratorCondition wasn't created for a Copied CSV over x seconds but there is not a way to distinguish that the controller simply hasn't reconciled the Copied CSV yet.

awgreene:
@kevinrizza I added the test described above which at least sounds an alarm if a Copied CSV is reconciled.

PR #1900, opened 2020-12-09
Bug 1910160: Create OperatorConditions for CSVs being installed

awgreene:
Problem: If OLM creates an OperatorConditions for an object before the
CSV is being installed it could update the existing deployment which is
not upgradeable, killing the pod.

Solution: Do not create the OperatorCondition until the CSV is being
installed.

### DIFF
--- a/pkg/controller/operators/operatorcondition_controller.go
+++ b/pkg/controller/operators/operatorcondition_controller.go
@@ -19,6 +19,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	operatorsv1 "github.com/operator-framework/api/pkg/operators/v1"
+	operatorsv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/ownerutil"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/controller/operators/operatorcondition_controller.go
```diff
@@ -201,6 +202,13 @@ func (r *OperatorConditionReconciler) ensureDeploymentEnvVars(operatorCondition
 		if err != nil {
 			return err
 		}
+
+		// Check the deployment is owned by a CSV with the same name as the OperatorCondition.
+		deploymentOwner := ownerutil.GetOwnerByKind(deployment, operatorsv1alpha1.ClusterServiceVersionKind)
```
dinhxuanvu, line 207: Would it be easier to use `IsOwnedBy` instead of this whole if condition?

dinhxuanvu (changes requested):
Looking good. Just a nit comment.

pkg/controller/operators/operatorcondition_controller.go
awgreene: I would rather not have to make a get call against the api server to check this.

dinhxuanvu: Fair enough.

dinhxuanvu (approved):
/lgtm

PR #1901, opened 2020-12-09
Bug 1903034:  Reduce log noise from OLM

awgreene:
Problem: OLM produce a lot of noise in its logs while printing debug
logs. Additionally, a few info logs could better be classified as debug
logs and are contributing to the noise as well.

Solution: Ensure that controllers created by OLM are not printing debug
level logs and reclassify a few info logs as debug logs.

### DIFF
--- a/cmd/olm/main.go
+++ b/cmd/olm/main.go
@@ -169,7 +169,7 @@ func main() {
 		}()
 	}
 
-	mgr, err := Manager(ctx)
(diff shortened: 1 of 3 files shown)


awgreene:
If you are interested in seeing the effect of this change please review [this comment](https://bugzilla.redhat.com/show_bug.cgi?id=1903034#c8).

PR #1903, opened 2020-12-10
Bug 1908471: Bump deps k8s 1.20

awgreene:


test/e2e/go.mod
```diff
@@ -0,0 +1,39 @@
+module github.com/operator-framework/operator-lifecycle-manager/test/e2e
+
+go 1.14
+
+require (
+	github.com/blang/semver v3.5.1+incompatible
```
kevinrizza, line 6: Does the e2e module really need all of these deps?

awgreene: Sadly yes - but helm recently shipped a PR that supports 1.20 so I decided to avoid creating a new module as we hit other dependency issues. We should however investigate how we can reduce our e2e tests on code implemented in the OLM project.

pkg/controller/registry/resolver/operators_test.go
```diff
@@ -6,7 +6,7 @@ import (
 
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/registry"
 
-	"github.com/blang/semver"
+	semver "github.com/blang/semver/v4"
```
ecordell, line 9: nit: you don't need the alias for a versioned module like this

cmd/olm/manager.go
```diff
@@ -13,7 +13,7 @@ import (
 var log = ctrl.Log.WithName("manager")
 
 func Manager(ctx context.Context, debug bool) (ctrl.Manager, error) {
-	ctrl.SetLogger(zap.Logger(debug))
+	ctrl.SetLogger(zap.New(zap.UseDevMode(debug)))
```
kevinrizza, line 16: Do we want debug mode?

awgreene: I believe so, the value is based on a flag provided via at runtime.

dinhxuanvu (approved):
/lgtm

dinhxuanvu (approved):
/lgtm

dinhxuanvu (approved):
/lgtm

dinhxuanvu (approved):
/lgtm

dinhxuanvu (commented):
/lgtm

test/e2e/installplan_e2e_test.go
```diff
@@ -2206,6 +2212,7 @@ var _ = Describe("Install Plan", func() {
 					}
 					return true, nil
 				})
+				require.NoError(GinkgoT(), err)
```
njhale, line 2215: Might as well switch over to `Eventually` here as well, no?

njhale, line 2215: nit: since `err` is a free variable within this loop body, you can check this once after both `if` statements.

njhale:
Looking really good. I think we can punt changing function signatures to use controller-runtime's `Object` interface and address it a follow up which only leaves the `go.mod` related comments to address.

pkg/controller/operators/adoption_controller.go
```diff
@@ -265,15 +258,20 @@ func (r *AdoptionReconciler) adopt(ctx context.Context, operator *decorators.Ope
 		return nil
 	}
 
-	if err := r.Get(ctx, types.NamespacedName{Namespace: m.GetNamespace(), Name: m.GetName()}, component); err != nil {
+	cObj, ok := component.(client.Object)
```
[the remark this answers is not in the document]
awgreene: Captured in https://github.com/operator-framework/operator-lifecycle-manager/issues/1917

pkg/controller/operators/adoption_controller.go
```diff
@@ -282,14 +280,21 @@ func (r *AdoptionReconciler) adopt(ctx context.Context, operator *decorators.Ope
...
+			return fmt.Errorf("Unable to typecast runtime.Object to client.Object")
+		}
+		return r.Patch(ctx, pCObj, client.MergeFrom(cObj))
 	}
 
 	return nil
 }
 
 func (r *AdoptionReconciler) disown(ctx context.Context, operator *decorators.Operator, component runtime.Object) error {
+	cObj, ok := component.(client.Object)
```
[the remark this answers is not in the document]
awgreene: Captured in https://github.com/operator-framework/operator-lifecycle-manager/issues/1917

pkg/controller/operators/adoption_controller.go
```diff
@@ -282,14 +280,21 @@ func (r *AdoptionReconciler) adopt(ctx context.Context, operator *decorators.Ope
...
+	if !ok {
+		return fmt.Errorf("Unable to typecast runtime.Object to client.Object")
+	}
 	candidate := component.DeepCopyObject()
```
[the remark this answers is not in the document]
awgreene: Can be done in https://github.com/operator-framework/operator-lifecycle-manager/issues/1917

test/e2e/installplan_e2e_test.go
awgreene: I don't believe that this is true, if the first wait failed and had an error but the second did not the first error would be overwritten.

PR #1904, opened 2020-12-11
Bug 1905299: fix(olm): Verify ServiceAccount ownership before installing deployment

dinhxuanvu:
Currently, there is a timing during upgrade, the new SA is not yet being
created but the deployment is successfully updated and the new pod
is coming up using the old SA. As a result, the pod fails to succeed
due to permission failure.

When CSV is in Pending state, a permission and requirement check list
is performed. A check for SA ownership is added to that check list
to verify if new SA is in place before moving CSV to Installing state.
If new SA is not yet in place, CSV can continue to be Pending which
will prevent the deployment to be created/updated.

**Reviewer Checklist**


ankitathomas (commented):
/lgtm

PR #1911, opened 2020-12-16
Bug 1907792: Change the OperatorUpgradeable condition to Upgradeable

dinhxuanvu:
The original design intends to use the name `Upgradeable` only.

Gate operator upgrade on the "Upgradeable" OperatorCondition status type instead of "OperatorUpgradeable".

The [OperatorConditions enhancement](https://github.com/operator-framework/enhancements/blob/master/enhancements/operator-conditions.md#example-upgradeable-condition) indicates that OLM gates upgrades on a status condition type of "Upgradeable", not "OperatorUpgradeable".

**Reviewer Checklist**

### DIFF
--- a/test/e2e/operator_condition_e2e_test.go
+++ b/test/e2e/operator_condition_e2e_test.go
@@ -22,9 +22,9 @@ var _ = Describe("Operator Condition", func() {
 		TearDown(testNamespace)
 	})
 
-	It("OperatorUpgradeable condition and overrides", func() {
+	It("OperatorCondition Upgradeable type and overrides", func() {
 		By("This test proves that an operator can upgrade successfully when" +
-			"OperatorUpgrade condition is set in OperatorCondition CR. Plus, an operator" +
+			"Upgrade condition type is set in OperatorCondition CR. Plus, an operator" +
 			"chooses not to use OperatorCondition, the upgrade process will proceed as" +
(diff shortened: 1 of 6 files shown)


dinhxuanvu:
/hold
until this #1903 gets merged first

test/e2e/operator_condition_e2e_test.go
```diff
@@ -23,9 +23,9 @@ var _ = Describe("Operator Condition", func() {
 		TearDown(testNamespace)
 	})
 
-	It("OperatorUpgradeable condition and overrides", func() {
+	It("OperatorCondition Upgradeable type and overrides", func() {
 		By("This test proves that an operator can upgrade successfully when" +
-			"OperatorUpgrade condition is set in OperatorCondition CR. Plus, an operator" +
+			"Upgrade condition rype is set in OperatorCondition CR. Plus, an operator" +
```
awgreene, line 28:
```suggestion
			"Upgrade condition type is set in OperatorCondition CR. Plus, an operator" +
```

njhale (approved):
I think we got lucky that we didn't cut a release -- in both upstream and OCP -- when this was an option. If we had and we were being strict about it, I think we would have needed to continue support for "OperatorUpgradeable" until we could formally deprecate it.

/approve

Issue #1868, opened 2020-11-18
CSV deployment annotations not showing in final deployment/pods


groeges:
## Bug Report

Any deployment template annotations defined in a CSV do not get included in the deployment when the operator is installed. Not sure whether this is a bug or something that  is not currently possible. Any help/guidance on providing annotation for the deployment template would be greatly received.

**What did you do?**
Generated a CSV (not full csv definition show here) that contained some annotations, 
spec.install.spec.deployments.spec.template.metadata.annotations
```
apiVersion: operators.coreos.com/v1alpha1
kind: ClusterServiceVersion
metadata:
  annotations:
    capabilities: Basic Install
    operators.operatorframework.io/builder: operator-sdk-v1.0.0
    operators.operatorframework.io/project_layout: go.kubebuilder.io/v2
  name: my-operator.v0.0.1
  namespace: placeholder
spec:
  apiservicedefinitions: {}
  customresourcedefinitions:
    owned:
      - description: Elasticsearch is the Schema for the elasticsearches API
        displayName: Elasticsearch
        kind: Elasticsearch
        name: elasticsearches.elastic.com
        version: v1alpha1
  description: my operator
  displayName: my-operator
  icon:
    - base64data: ""
      mediatype: ""
  install:
    spec:
      clusterPermissions:
        - rules:
            - apiGroups:
                - apps
              resources:
                - deployments
              verbs:
                - get
      deployments:
        - name: mydeployment
          spec:
            replicas: 1
            selector:
              matchLabels:
                app.kubernetes.io/component: elasticsearch
                control-plane: controller-manager
            strategy: {}
            template:
              metadata:
                annotations:
                  myAnnnotation1: ALL
                  myAnnotation2: none
                labels:
                  app.kubernetes.io/component: mycomponent
                  control-plane: controller-manager
              spec:
.....
```

**What did you expect to see?**
When the operator was installed I expected to see the 
```
   myAnnotation1: ALL
   myAnnotation2: none
```
annotations on the "Deployment" under spec.template.metadata.anotations, for example:
```
kind: Deployment
apiVersion: apps/v1
metadata:
  annotations:
    deployment.kubernetes.io/revision: '2'
  resourceVersion: '25276567'
  name: mydeployment
  uid: 2aba1f32-5ae2-4854-9dca-500820a14900
  creationTimestamp: '2020-11-18T10:41:06Z'
  generation: 2
  managedFields: {}
  namespace: default
  labels:
[...]

PR #1913, opened 2020-12-17
Bug 1907381: CSV Annotations override pod annotations

awgreene:
Problem: If a pod template defined in a CSV's StrategyDetailsDeployment
includes an annotation defined for the CSV OLM will fail to install the
operator because the pod's annotations do not match those defined in
the CSV.

Solution: Pod template annotations defined in a StrategyDetailsDeployment are
overwritten by those defined in the CSV.


dinhxuanvu (approved):
/lgtm

exdx:
Closes #1868

PR #1915, opened 2020-12-21
[wip] tweak: remove reconciling status on copied csvs

exdx:
Removes continuous reconciliation of status changes between the original and copied CSVs. This disables status reconciliation both ways, between the original and the copy and the copy and the original. 

Note: there is still an initial update of the status on the copy (setting status fields that indicate the copied CSV is in fact a copy) which will cause a spike in resource utilization in the api-server when the CSV is installed across many namespaces. However the update is small, and after that initial spike no further changes to the status in the original CSV will be reflected in the copy. 

Reconciling CSV statuses between original <-> copy has been observed to cause performance issues on clusters with large numbers of CSVs across multiple namespaces. 

**Reviewer Checklist**


pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -725,17 +725,13 @@ func (a *Operator) copyToNamespace(csv *v1alpha1.ClusterServiceVersion, namespac
...
-		newCSV.Status.Message = fmt.Sprintf("The operator is running in %s but is managing this namespace", csv.GetNamespace())
-
-		if !reflect.DeepEqual(fetchedCSV.Status, newCSV.Status) {
-			logger.Debug("updating status")
-			// Must use fetchedCSV because UpdateStatus(...) checks resource UID.
-			fetchedCSV.Status = newCSV.Status
-			if fetchedCSV, err = a.client.OperatorsV1alpha1().ClusterServiceVersions(namespace).UpdateStatus(context.TODO(), fetchedCSV, metav1.UpdateOptions{}); err != nil {
-				logger.WithError(err).Error("status update for target CSV failed")
+		logger.Debug("checking spec")
+		if !reflect.DeepEqual(fetchedCSV.Spec, newCSV.Spec) {
```
njhale, line 729: This is a net new event hook, but IMO we've probably been missing this for a while.

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -725,17 +725,13 @@ func (a *Operator) copyToNamespace(csv *v1alpha1.ClusterServiceVersion, namespac
...
+			// update based on changes to the spec only
+			// Note: PUT verbs on objects (such as via Update()) ignore status values
+			fetchedCSV.Spec = newCSV.Spec
+			if fetchedCSV, err = a.client.OperatorsV1alpha1().ClusterServiceVersions(namespace).Update(context.TODO(), fetchedCSV, metav1.UpdateOptions{}); err != nil {
```
njhale, line 733: It would be nice to wrap changes for both updated labels and spec into a single `Update` call.

test/e2e/csv_e2e_test.go
```diff
@@ -98,7 +99,7 @@ var _ = Describe("ClusterServiceVersion", func() {
 			}
 		})
 
-		It("is synchronized with the original csv", func() {
```
njhale, line 101: I do think we want a test to make sure the copied CSV status is level with the "minimal copied CSV status" we've defined in this PR.

test/e2e/csv_e2e_test.go
```diff
@@ -130,7 +131,43 @@ var _ = Describe("ClusterServiceVersion", func() {
...
+					return err
+				}
+
+				key.Namespace = target.Name
+				if err := ctx.Ctx().Client().Get(context.Background(), key, &original); err != nil {
+					return err
+				}
+
+				// add extra port (443) to the deployment spec on the CSV
+				original.Spec.InstallStrategy.StrategySpec.DeploymentSpecs[0].Spec.Template.Spec.Containers[0].Ports = []corev1.ContainerPort{{ContainerPort: 80}, {ContainerPort: 443}}
```
njhale, line 149: It almost feels like we should be eliding everything but the metadata useful to operator consumers. For example, the `InstallStrategy` is irrelevant to operator consumers. In general, I think we need to keep everything used directly by Console.

njhale (changes requested):
Thanks @exdx! Looking great!

I'm still expecting to see some changes to the OperatorGroup unit tests for this. It also looks like this may be cause some issues with Console e2e tests.

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -746,6 +742,11 @@ func (a *Operator) copyToNamespace(csv *v1alpha1.ClusterServiceVersion, namespac
 		newCSV.SetNamespace(namespace)
 		newCSV.SetResourceVersion("")
 		newCSV.SetLabels(utillabels.AddLabel(newCSV.GetLabels(), v1alpha1.CopiedLabelKey, csv.GetNamespace()))
+		// the copied CSV should only contain the original CSV spec and a minimal amount of status indicating it's a copy
+		// updating based on changes to the status causes performance issues on clusters with large numbers of CSVs across namespaces
+		// the operator API is the preferred method of discovering operators on-cluster versus checking CSVs in a namespace
+		// Note: POST verbs on objects (such as via Create()) ignore status values, the below line is just for clarity
+		newCSV.Status = v1alpha1.ClusterServiceVersionStatus{}
```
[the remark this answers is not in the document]
exdx: I agree its superfluous -- just added it for reviewing purposes. Removed.

test/e2e/csv_e2e_test.go
exdx: This is a good point but beyond the scope of what we originally envisioned when we came up with the bugfix, which was just the minimal "this CSV is copied" properties. Is there an easy way to determine all the metadata the console looks at for a CSV?

dinhxuanvu:
We will eventually get rid of copying CSV altogether so this PR is no longer needed.
/close

PR #1881, opened 2020-11-27
Bug 1857877: check the service account owner in the requirement

horis233:
**Description of the change:**

This PR is to check the owner of the service account when OLM controller checks the requirements

Closes #1879 
**Motivation for the change:**

This PR is for fixing this issue https://github.com/operator-framework/operator-lifecycle-manager/issues/1879.

This can help to avoid the issue that during the upgrade, the service account is deleted together with the old CSV.

**Reviewer Checklist**
- [ ] Implementation matches the proposed design, or proposal is updated to match implementation
- [ ] Sufficient unit test coverage 
- [ ] Sufficient end-to-end test coverage
- [ ] Docs updated or added to `/docs` 
- [ ] Commit messages sensible and descriptive

PR #1916, opened 2020-12-21
Revert "Bug 1905299: fix(olm): Verify ServiceAccount ownership before installing deployment"

dinhxuanvu:
Reverts operator-framework/operator-lifecycle-manager#1904

Reason: The fix in this PR is a dup from another merged PR: #1881

### DIFF
--- a/pkg/controller/operators/olm/requirements.go
+++ b/pkg/controller/operators/olm/requirements.go
@@ -269,14 +269,6 @@ func (a *Operator) permissionStatus(strategyDetailsDeployment *v1alpha1.Strategy
 				statusesSet[saName] = status
 				continue
 			}
-			// Check SA's ownership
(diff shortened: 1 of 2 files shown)


dinhxuanvu:
/hold
Will wait until master is open again.

njhale:
/approve
/retest

I wonder if there's some procedure we can adopt when writing or submitting bug fixes that will help us catch duplicates like this? e.g. split the patch into two commits: failing test in the first, fix in the second.

dinhxuanvu:
> /approve
> /retest
> 
> I wonder if there's some procedure we can adopt when writing or submitting bug fixes that will help us catch duplicates like this? e.g. split the patch into two commits: failing test in the first, fix in the second.

The two PRs were merged a few days apart so I believe at the time I was working on this, my local master might have not had the other fix. The test case I added was a negative test case meaning it was supposed to fail.

dinhxuanvu:
No longer needed and replaced by #2029 
/close

PR #1934, opened 2020-12-31
Use client.Object for methods in the adoption and operator controllers

### LINKED ISSUE #1917: Update Adoption and Operator Controllers to use client.Object for methods
## Feature Request

**Is your feature request related to a problem? Please describe.**
The adoption and operator controllers have not updated their method signatures to use the client.Object instead of the runtime.Object. 

The methods in the controllers mentioned above should be updated to use the client.Object interface.

horis233:
Fix #1917

This PR is for fixing issue #1917.

- The `runtime.Object` in the adoption and operator controllers are replaced to `client.Object` and `client.ObjectList`. 

- Function `flatten` is updated to extract `client.ObjectList` slice to `client.Object` slice.

- The object flatten logic is removed from `func (o *Operator) AddComponents`, components are going to be flattened before they are passed to `AddComponents`.

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/adoption_controller.go
+++ b/pkg/controller/operators/adoption_controller.go
@@ -262,7 +262,7 @@ func (r *AdoptionReconciler) adoptComponents(ctx context.Context, csv *operators
 func (r *AdoptionReconciler) adopt(ctx context.Context, operator *decorators.Operator, component client.Object) error {
 	if err := r.Get(ctx, types.NamespacedName{Namespace: component.GetNamespace(), Name: component.GetName()}, component); err != nil {
 		if apierrors.IsNotFound(err) {
-			r.log.V(1).Info("not found", "component", cObj)
+			r.log.V(1).Info("not found", "component", component)
 			err = nil
 		}
 
--- a/pkg/controller/operators/adoption_controller.go
+++ b/pkg/controller/operators/adoption_controller.go
@@ -368,7 +368,13 @@ func (r *AdoptionReconciler) adoptees(ctx context.Context, operator decorators.O
 		components []client.Object
 		errs       []error
 	)
-	for _, candidate := range flatten(componentLists) {
+
+	componentCandidates, err := flatten(componentLists)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, candidate := range componentCandidates {
 		m, err := meta.Accessor(candidate)
 		if err != nil {
 			errs = append(errs, err)
@@ -524,12 +530,15 @@ func (r *AdoptionReconciler) mapToProviders(obj client.Object) (requests []recon
 }
 
 // flatten extracts slice of ObjectList to the slice of Object
-func flatten(lists []client.ObjectList) (flattened []client.Object) {
+func flatten(lists []client.ObjectList) (flattened []client.Object, err error) {
 	for _, list := range lists {
-		_ = meta.EachListItem(list, func(obj runtime.Object) error {
+		err = meta.EachListItem(list, func(obj runtime.Object) error {
 			flattened = append(flattened, obj.(client.Object))
 			return nil
 		})
+		if err != nil {
+			return
+		}
 	}
 	return
 }
--- a/pkg/controller/operators/adoption_controller.go
+++ b/pkg/controller/operators/adoption_controller.go
@@ -259,18 +259,8 @@ func (r *AdoptionReconciler) adoptComponents(ctx context.Context, csv *operators
 	return utilerrors.NewAggregate(errs)
 }
 
-func (r *AdoptionReconciler) adopt(ctx context.Context, operator *decorators.Operator, component runtime.Object) error {
-	m, err := meta.Accessor(component)
-	if err != nil {
-		return nil
-	}
-
-	cObj, ok := component.(client.Object)
-	if !ok {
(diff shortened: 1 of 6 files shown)


pkg/controller/operators/adoption_controller.go
```diff
@@ -252,26 +252,25 @@ func (r *AdoptionReconciler) adoptComponents(ctx context.Context, csv *operators
 	return utilerrors.NewAggregate(errs)
 }
 
-func (r *AdoptionReconciler) adopt(ctx context.Context, operator *decorators.Operator, component runtime.Object) error {
+func (r *AdoptionReconciler) adopt(ctx context.Context, operator *decorators.Operator, component client.Object) error {
 	m, err := meta.Accessor(component)
 	if err != nil {
 		return nil
 	}
```
joelanford, line 259: There's no need to use `m` and `meta.Accessor` on `client.Object` instances. The `client.Object` interface includes the `metav1.Object` interface.

pkg/controller/operators/adoption_controller.go
```diff
@@ -512,15 +498,31 @@ func (r *AdoptionReconciler) mapToProviders(obj client.Object) (requests []recon
...
+		if len(nested) == 0 {
+			continue
+		}
 
+		if meta.IsListType(nested[0]) {
+			cNested := make([]client.ObjectList, len(nested))
+			for index := range nested {
+				cNested[index] = nested[index].(client.ObjectList)
+			}
+			flattened = append(flattened, flatten(cNested)...)
```
joelanford, line 518: Is is possible for individual items within an ObjectList to be more ObjectLists? I don't _think_ so, but I could be wrong.

In the previous implementation, we didn't know if the `obj []runtime.Object`s were lists or just regular objects. Hence the need to try to call `meta.ExtractList` (and if that didn't work, just add the object directly).

Now that we know that the objects being passed in are lists, we don't need to handle both cases.

Would this simpler version work instead:
```go
func flatten(lists []client.ObjectList) (flattened []client.Object) {
	for _, list := range lists {
		_ = meta.EachListItem(list, func(obj runtime.Object) error {
			flattened = append(flattened, obj.(client.Object))
			return nil
		})
	}
	return
}
```

pkg/controller/operators/adoption_controller.go
horis233: Thanks @joelanford, I will remove that section.

pkg/controller/operators/adoption_controller.go
horis233: @joelanford Thank you for the explanation! 

Yes, what you say makes sense to me. I will use the function you provide in the above comment

pkg/controller/operators/adoption_controller.go
```diff
@@ -543,15 +524,13 @@ func (r *AdoptionReconciler) mapToProviders(obj client.Object) (requests []recon
...
-		}
-
-		flattened = append(flattened, obj)
+// flatten extracts slice of ObjectList to the slice of Object
+func flatten(lists []client.ObjectList) (flattened []client.Object) {
+	for _, list := range lists {
+		_ = meta.EachListItem(list, func(obj runtime.Object) error {
+			flattened = append(flattened, obj.(client.Object))
+			return nil
+		})
```
timflannagan, line 533: Are we discounting any errors that this inline function may produce? Do we care about handling errors that may occur when iterating over the list of client objects?

pkg/controller/operators/decorators/operator.go
```diff
@@ -62,10 +63,10 @@ func OperatorNames(labels map[string]string) (names []types.NamespacedName) {
 
 type OperatorFactory interface {
 	// NewOperator returns an Operator decorator that wraps the given external Operator representation.
-	// An error is returned if the decorator cannon be instantiated.
+	// An error is returned if the decorator cannot be instantiated.
 	NewOperator(external *operatorsv1.Operator) (*Operator, error)
 
-	// NewPackageOperator returns an Operator decorator for a package and install namespace.
+	// NewPackageOperator returns an Operator decorator for a package and installs namespace.
```
njhale, line 69: This was not a typo. It should be `install namespace`.

pkg/controller/operators/adoption_controller.go
njhale: Hmm, I think it's probably okay.

In this case, the variables are constrained to `[]client.ObjectList`, which we know contains elements that of `[]client.Object`, so we know the assertion would never panic. We also know that the function literal will always return nil.

pkg/controller/operators/decorators/operator.go
horis233: Thanks for pointing it out. I reverted it back to `install`

awgreene:
Hello,

This repository contains a number of PRs that have been open for an excessively long period of time, in some cases these PRs have been open for more than two years. In an effort to improve the health of our project, we are closing PRs that haven't been updated in the past 30 days or include features that we do not wish to adopt. This PR has been identified as a candidate meeting the requirements above and will be closed. If you believe that this PR should still be merged, please reopen the PR and explain why.

We will eventually release an official policy detailing our process for handling "stale" PRs and issues. If you would like to be involved in the conversation, please engage through [this Issue](https://github.com/operator-framework/rukpak/pull/553).

Thanks you,

Alex

PR #1941, opened 2021-01-06
Bug 1909992: Allow private bundle images within private indexes

anik120:
In #1878, the secrets passed in `spec.secrets` field of a catalogsource
were attached to the catsrc's corresponding SA that was used by the registry
pod, thereby having access to the secrets. This allowed for pulling of
private index images. However, the job that unpacks the bundle images did
not have access to the secrets, and as a result private bundle image included
in the catsrc were not installable using the secrets.
This PR attaches the secrets from the catsrc to the job that unpacks the bundles,
thereby allowing the inclusion of private bundle images within index from private
indexes.

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/bundle/bundle_unpacker.go
+++ b/pkg/controller/bundle/bundle_unpacker.go
@@ -65,7 +65,7 @@ func newBundleUnpackResult(lookup *operatorsv1alpha1.BundleLookup) *BundleUnpack
 	}
 }
 
-func (c *ConfigMapUnpacker) job(cmRef *corev1.ObjectReference, bundlePath string) *batchv1.Job {
+func (c *ConfigMapUnpacker) job(cmRef *corev1.ObjectReference, bundlePath string, secrets []corev1.LocalObjectReference) *batchv1.Job {
 	job := &batchv1.Job{
 		Spec: batchv1.JobSpec{
 			//ttlSecondsAfterFinished: 0 // can use in the future to not have to clean up job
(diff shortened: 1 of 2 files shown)


pkg/controller/bundle/bundle_unpacker.go
```diff
@@ -65,7 +65,7 @@ func newBundleUnpackResult(lookup *operatorsv1alpha1.BundleLookup) *BundleUnpack
 	}
 }
 
-func (c *ConfigMapUnpacker) job(cmRef *corev1.ObjectReference, bundlePath string) *batchv1.Job {
+func (c *ConfigMapUnpacker) job(cmRef *corev1.ObjectReference, bundlePath string, secrets []corev1.LocalObjectReference) *batchv1.Job {
```
exdx, line 68: This looks good, but do we want to go with the same approach of using the SA instead of passing the secret directly so that the `imagePullSecrets` field on the pod spec would get set automatically?

Or are we unable to get the catalog SA since it's generated randomly. 

Maybe its better to use the same SA for pulling an index and unpacking a bundle from that same index via a map of some kind

anik120: @exdx the issue with that approach is that the serviceaccount is created when the registry pod is created, but no record of the serviceaccount is persisted in the CatalogSource itself (it does not have any `spec.Serviceaccount` field. The SA is attached to the registry pod after both are created). We could query for the registry pod of the CatalogSource and extract the SA name from there, but that seems like an unnecessary extra client call when the CatalogSource is already being fetched once.

exdx: We could get around this by annotating the catalogsource when creating it with the service account that is being used by the underlying pod and then using it in this context. I just feel like if we went through this effort of getting per-catalog SAs shouldn't we use them for the underlying bundles as well? Seems like a clean solution but would like @njhale to weigh in here

njhale (approved):
This looks good, thanks!

At some level, I would eventually like us to use one technique for secret binding everywhere, but this seems like the best solution at the moment.

njhale:
/retest

we really need to fix up these flakes to get some meaningful signals

njhale:
/hold

looks like the console tests are consistently failing.

anik120:
Looks like the PR to fix the test was merged: https://github.com/openshift/console/pull/7835
/hold cancel
/retest

estroz:
Would love to see this backported to 0.17.

PR #1942, opened 2021-01-06
Bug 1910160: Update OperatorCondition controller to watch deps

awgreene:
Problem: Updates to deployments can remove the OPERATOR_CONDITION_NAME
environment variable.

Solution: Update the OperatorCondition controller to watch for
Deployment events, if the deployment is owned by a CSV, trigger the
reconciler.

### DIFF
--- a/pkg/controller/operators/operatorcondition_controller_test.go
+++ b/pkg/controller/operators/operatorcondition_controller_test.go
@@ -287,8 +287,44 @@ var _ = Describe("OperatorCondition", func() {
 					Expect(k8sClient.Create(ctx, operatorCondition)).To(Succeed())
 				})
 
-				It("injects the OperatorCondition name into the deployment's Environment Variables", func() {
+				It("should always inject the OperatorCondition Environment Variable into containers defined in the deployment", func() {
 					deployment := &appsv1.Deployment{}
+					Eventually(func() error {
+						err := k8sClient.Get(ctx, types.NamespacedName{Name: operatorCondition.Spec.Deployments[0], Namespace: namespace.GetName()}, deployment)
+						if err != nil {
+							return err
+						}
+						if len(deployment.Spec.Template.Spec.Containers) != 1 {
+							return fmt.Errorf("Deployment should contain a single container")
+						}
+						for _, container := range deployment.Spec.Template.Spec.Containers {
+							if len(container.Env) == 0 {
+								return fmt.Errorf("env vars should exist")
+							}
+						}
+						return nil
+					}, timeout, interval).Should(BeNil())
+
+					Expect(len(deployment.Spec.Template.Spec.Containers)).Should(Equal(1))
+					for _, container := range deployment.Spec.Template.Spec.Containers {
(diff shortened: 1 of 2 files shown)


awgreene:
gcp failures.

/retest

pkg/controller/operators/operatorcondition_controller_test.go
```diff
@@ -313,6 +313,44 @@ var _ = Describe("OperatorCondition", func() {
 							Value: operatorCondition.GetName(),
 						}))
 					}
+
+					Context("if the OPERATOR_CONDITION_NAME environment variable is removed", func() {
```
joelanford, line 317: I don't think `Context` is typically found inside `It` blocks. Is is possible to refactor this to be more [idiomatic](https://onsi.github.io/ginkgo/#organizing-specs-with-containers-describe-and-context)?

pkg/controller/operators/operatorcondition_controller_test.go
```diff
@@ -313,6 +313,44 @@ var _ = Describe("OperatorCondition", func() {
...
+						deployment := &appsv1.Deployment{}
+						Eventually(func() error {
+							err := k8sClient.Get(ctx, types.NamespacedName{Name: operatorCondition.Spec.Deployments[0], Namespace: namespace.GetName()}, deployment)
+							if err != nil {
+								return err
+							}
+							deployment.Spec.Template.Spec.Containers[0].Env = nil
+							return k8sClient.Update(ctx, deployment)
+						}, timeout, interval).Should(BeNil())
```
joelanford, line 326: Should this setup code to delete the env be in a `BeforeEach` block to ensure it runs before each test in this context (even if there is only one now)?

pkg/controller/operators/operatorcondition_controller_test.go
```diff
@@ -313,6 +313,44 @@ var _ = Describe("OperatorCondition", func() {
...
+						Context("The OPERATOR_CONDITION_NAME environment variable is injected once again", func() {
```
joelanford, line 327:
```suggestion
						It("should re-inject the OPERATOR_CONDITION_NAME environment variable", func() {
```

pkg/controller/operators/operatorcondition_controller_test.go
awgreene: Can do!

pkg/controller/operators/operatorcondition_controller_test.go
```diff
@@ -313,6 +313,42 @@ var _ = Describe("OperatorCondition", func() {
 							Value: operatorCondition.GetName(),
 						}))
 					}
+
+					// Remove the container's Environment Variables defined in the deployment
```
joelanford, line 317: Non-blocking nit: instead of using comments, you can use [`By()`](https://onsi.github.io/ginkgo/#documenting-complex-its-by) so that the step that fails within the `It` block can be printed in the test output.

joelanford (approved):
One minor nit. Otherwise
/lgtm

PR #1943, opened 2021-01-06
Bug 1913132: Create CSV Deployments with OpCond EnvVar

awgreene:
Problem: The OLM controller creates deployments defined in the CSV
without including the OPERATOR_CONDITION_NAME environment variable. The
OperatorCondition reconciller then immediately injects said environment
variable into the deployment.

Solution: Update the OLM controller to create the deployment with the
OPERATOR_CONDITION_NAME environment variable.


dinhxuanvu (approved):
/lgtm

PR #1951, opened 2021-01-11
Bug1906056: fix(resolver): Allow skipped versions to be installed initially

dinhxuanvu:
Currently, the resolver will filter all skipped versions so they
cannot be installed under any conditions. This behavior is a slight
change from the older resolver which allows the skipped versions to
be installed if they are chosen in `startingCSV`.

With this change, the resolver will allow skipped versions to be
installed if specified in `startingCSV` during the initial
installation.

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/registry/resolver/resolver.go
+++ b/pkg/controller/registry/resolver/resolver.go
@@ -97,6 +97,7 @@ func (r *SatResolver) SolveOperators(namespaces []string, csvs []*v1alpha1.Clust
 
 		// if no operator is installed and we have a startingCSV, filter for it
 		if current == nil && len(sub.Spec.StartingCSV) > 0 {
+			predicates = append(predicates, WithCSVName(sub.Spec.StartingCSV))
 			channelFilter = append(channelFilter, WithCSVName(sub.Spec.StartingCSV))
 			startingCSVs[sub.Spec.StartingCSV] = struct{}{}
 		}
--- a/pkg/controller/registry/resolver/resolver_test.go
+++ b/pkg/controller/registry/resolver/resolver_test.go
@@ -1354,14 +1354,14 @@ func TestSolveOperators_WithoutDeprecated(t *testing.T) {
 	assert.IsType(t, solver.NotSatisfiable{}, err)
 }
 
-func TestSolveOperators_WithSkips(t *testing.T) {
+func TestSolveOperators_WithSkipsAndStartingCSV(t *testing.T) {
 	APISet := APISet{opregistry.APIKey{"g", "v", "k", "ks"}: struct{}{}}
 	Provides := APISet
 
 	namespace := "olm"
 	catalog := registry.CatalogKey{"community", namespace}
 
-	newSub := newSub(namespace, "packageB", "alpha", catalog)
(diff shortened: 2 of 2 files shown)


pkg/controller/registry/resolver/resolver.go
```diff
@@ -99,10 +100,11 @@ func (r *SatResolver) SolveOperators(namespaces []string, csvs []*v1alpha1.Clust
 		if current == nil && len(sub.Spec.StartingCSV) > 0 {
 			channelFilter = append(channelFilter, WithCSVName(sub.Spec.StartingCSV))
 			startingCSVs[sub.Spec.StartingCSV] = struct{}{}
+			allowSkipped = true
```
ecordell, line 103: instead of a `bool` flag that gets threaded all the way down, could we come up with a `predicate` or `channelFilter` that we could pass to the existing `getSubscriptionInstallables`?

If the resolver only sees the filtered data with one option per channel, the operator will get ignored because it's skipped (because all the operators that skip it will have been filtered out)

exdx: I agree, this also would help with keeping the signatures of the various helper functions involved in resolution simpler

dinhxuanvu: I have tried to do so but the issue here is being-skipped is not a property that can be constructed into `predicate`. You can filter out an operator with a property that owns (or doesn't). Currently you can filter out an operator that skips/replaces something but you can't do the other way around.
At the moment, we filter out the skipped version during sorting bundles within a channel via a utility function. In order to filter out skipped versions, we will need to let the registry to add some label manually when the bundle is added to an index.

ecordell: I don't think we need to filter out "things that skip startingCSV"

I think we need to filter out "anything that is not startingCSV" - if you've specified startingCSV and there's not already an operator installed for the subscription, you should either get startingCSV or nothing, right?

It might be as simple as changing

```go
channelFilter = append(channelFilter, WithCSVName(sub.Spec.StartingCSV))
```
to 
```go
predicates =  append(predicates, WithCSVName(sub.Spec.StartingCSV))
```

(but I haven't looked at this in a while - I'm a little unclear about why the channelFilter doesn't take care of this already)

dinhxuanvu: After discussed with Evan, we agreed to proceed with the `predicate` addition to `cachepredicate` to filter out the ineligible versions in advance before it reaches the `sortChannel` stage.

pkg/controller/registry/resolver/resolver_test.go
```diff
@@ -1361,7 +1361,7 @@ func TestSolveOperators_WithSkips(t *testing.T) {
 	namespace := "olm"
```
ecordell, line 1361: Should this be a separate test case, so that we have two? Skips without startingCSV, skips with startingCSV?

dinhxuanvu: The non-startingCSV is implied with the dependent operator but I made 2 test cases now just to be more clear.

njhale (approved):
Nice!

/lgtm

PR #1972, opened 2021-01-25
update makefile, workflow, and docs

jchunkins:

The purpose of this PR is to:
- migrate logic of `x.mk` into main `Makefile` and remove `x.mk`
- update `.github/workflows/e2e-tests.yml` workflow to use main `Makefile`
- update .gitignore to avoid tarfile checkin
- add Makefile documentation
- update readme documentation with usage and examples
- enhance scripts/package_release.sh to run sed correctly on multi platform (linux and Mac)

As a prelude to adding new end to end tests, I wanted the ability to run individual tests. At first glance the main makefile seemed like it had not been updated to use the new BDD tests, and it was pointed out that `x.mk` makefile was serving this purpose. During the last weekly call it was suggested that we move the make target logic in `x.mk` to the main makefile. 

**Reviewer Checklist**


scripts/package_release.sh
```diff
@@ -17,7 +17,14 @@ charttmpdir=$(mktemp -d 2>/dev/null || mktemp -d -t charttmpdir)
...
 cp -R deploy/chart/ "${charttmpdir}"
-sed -i "s/^[Vv]ersion:.*\$/version: ${version}/" "${charttmpdir}/Chart.yaml"
+
+# A space character is needed for -i argument in mac BSD sed.
+# GNU sed has a --version argument, and BSD sed does not. 
+# Check for both conditions to be safe for darwin users that override sed with gsed
+if [ -z "${OSTYPE##*"darwin"*}" ] && ! sed --version >/dev/null 2>&1; then
+    SP=" "
+fi
+sed -i"${SP}"'' "s/^[Vv]ersion:.*\$/version: ${version}/" "${charttmpdir}/Chart.yaml"
```
benluddy, line 27: Since we have a separate source and destination file, how about avoiding `-i` altogether with something like this?

```suggestion
sed "s/^[Vv]ersion:.*\$/version: ${version}/" deploy/chart/Chart.yaml > "${charttmpdir}/Chart.yaml"
```

Makefile
```diff
@@ -128,9 +131,35 @@ setup-bare: clean e2e.namespace
...
+FORCE:
+
+# main entry point for running end to end tests. used by .github/workflows/e2e-tests.yml See test/e2e/README.md for details
+.PHONY: e2e-local
+e2e-local: bin/e2e-local.test test/e2e-local.image.tar
+	$(GINKGO) -nodes $(or $(NODES),1) -flakeAttempts 3 -randomizeAllSpecs $(if $(TEST),-focus '$(TEST)') -v -timeout 90m $< -- -namespace=operators -olmNamespace=operator-lifecycle-manager -dummyImage=bitnami/nginx:latest -kind.images=../test/e2e-local.image.tar
+
+# this target updates the zz_chart.go file with files found in deploy/chart
+# will only fire if a file in deploy/chart has been changed
+test/e2e/assets/chart/zz_chart.go: $(shell find deploy/chart -type f)
```
benluddy, line 147: Forcing this recipe to run always might save some future headaches. The prerequisite list can miss things like removed files or modification time shenanigans. I forget the details, but it definitely frustrated someone at some point -- the quick patch at the time was to avoid committing the generated file (https://github.com/operator-framework/operator-lifecycle-manager/pull/1454).

benluddy (commented):
Great, thank you for doing this!

/ok-to-test

scripts/package_release.sh
jchunkins: good point... this is easier and accomplishes the same goal. Fixed.

Makefile
jchunkins: Makes sense. I've updated the target to make if "phony" which forces its execution.

jchunkins:
@benluddy Thanks for the quick review. I made the changes to the PR, but I see that `ci/prow/e2e-upgrade` and `ci/prow/e2e-aws-console-olm` failed. Are these the result of common build failures? Also note that DCO is failed... I added a signature to my last commit, but apparently that's not enough. Just wondering if these items are critical, and if so, what would need to be done to resolve them.

.github/workflows/e2e-tests.yml
```diff
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - run: make -f x.mk e2e-local NODES=2 JUNIT_DIRECTORY=./artifacts/
+      - run: make -f Makefile e2e-local NODES=2 JUNIT_DIRECTORY=./artifacts/
```
ecordell, line 12: You can just remove `-f Makefile` entirely

benluddy:
> @benluddy Thanks for the quick review. I made the changes to the PR, but I see that `ci/prow/e2e-upgrade` and `ci/prow/e2e-aws-console-olm` failed. Are these the result of common build failures? Also note that DCO is failed... I added a signature to my last commit, but apparently that's not enough. Just wondering if these items are critical, and if so, what would need to be done to resolve them.

Neither test failure is related to your changes. The console failure looks to have been a flake (it hit a 503 attempting to download a test dependency), and the upgrade failure is a problem with the downstream test infrastructure.

The DCO check requires all commits in a PR to be "signed off," so you'll need to amend the first commit and rebase in order to satisfy that one.

.github/workflows/e2e-tests.yml
jchunkins: happy to remove it... my rationale for leaving it was to make it explicit

jchunkins: fixed

jchunkins:
@benluddy Don't want to pester you, but I am not sure on the process here... I read through https://github.com/kubernetes/community/blob/master/contributors/guide/owners.md#the-code-review-process but have questions. Should I preemptively do an assignment of an approver or should I wait for a looks good label and then do the assignment of the approver?

benluddy:
@jchunkins Most of the people in OWNERS tend to check in on PRs in need of review from time to time, so I don't think it's strictly necessary to assign a reviewer in order to get a PR approved.

Opened https://github.com/operator-framework/operator-lifecycle-manager/issues/1976 for an apparent flake in some of the test teardown behavior.

/retest

jchunkins:
I think folks have reviewed, but there has not been a LGTM or Approve yet. Just wondering if there is any further work I need to do before this can be merged?

jchunkins:
Github showed that I had a branch conflict, and I resolved it (just a readme update) and re-ran tests which removed the LGTM comment.
@kevinrizza Can I get a repeat LGTM?
@benluddy Can you approve?

PR #1989, opened 2021-02-03
[release-4.6] Bug 1922919: Fix zero-delay resyncs for certain registry update policies.

benluddy:
If the registry update polling interval is greater than and a multiple
of the operator's default resync period (15m), and there are fewer
than 15 minutes remaining until the next poll is due, the
catalog-operator will continuously resync its associated CatalogSource
until the next poll time is reached. This is not beneficial to update
polling and can generate excessive load that negatively impacts useful
work on the cluster.


benluddy:
https://www.githubstatus.com/incidents/5zb8gfbl7qkt

PR #1990, opened 2021-02-03
Bug 1921954: Clarify subscription constraint strings in resolution failures.

benluddy:
ResolutionFailed events exposed internal terminology (i.e. "mandatory"
and "dependency") directly to users. Common constraints applied to
subscriptions appeared as "PACKAGE is mandatory" or "PACKAGE has a
dependency without any candidates to satisfy it", which in practice
meant "a subscription to PACKAGE exists in the namespace" and "no
operators found matching the criteria of the subscription to PACKAGE",
respectively. The language of these Subscription-related constraint
strings now better reflects what they mean to users.


ecordell (approved):
/approve

PR #1993, opened 2021-02-04
Bug 1921953: Infer package name property for unannotated CSVs, if possible.

benluddy:
If a CSV is missing the properties annotation, but it is referenced in
a Subscription's .status.installedCSV and is available in the same
Subscription's referenced catalog, then the CSV can be treated as
though it has a package annotation.

### DIFF
--- a/pkg/controller/registry/resolver/resolver.go
+++ b/pkg/controller/registry/resolver/resolver.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"sort"
 
+	"github.com/blang/semver/v4"
 	"github.com/sirupsen/logrus"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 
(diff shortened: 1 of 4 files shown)


benluddy:
The end-to-end test for this relies on https://github.com/operator-framework/operator-lifecycle-manager/pull/1994. The existing test CatalogSource functionality would need to be reworked to express a package dependency, so inference is tested by ensuring that an unannotated CSV causes a package uniqueness clash with a second subscription for the same package. https://github.com/operator-framework/operator-lifecycle-manager/pull/1994 makes subscription identity (for the purposes of resolution) based on subscription name rather than package name. Without that change, only one subscription to a given package is respected.

dinhxuanvu:
Replaced by #2005

dinhxuanvu (approved):
/lgtm

horis233:
@dinhxuanvu Will this fix be picked for OCP4.6?

dinhxuanvu:
@horis233 It will.

PR #1994, opened 2021-02-05
Add meaningful string representations for resolution invariants.

benluddy:
During dependency resolution, invariants are enforced for
one-operator-per-package and one-provider-per-api. Their user-facing
string representation (shown when they prevent resolution from
succeeding) had re-used the same representation as constraints on
Subscriptions out of convenience.

Recently, the messaging of Subscription-related constraints was
changed to be more descriptive which inadvertently made the invariant
messaging misleading. This patch improves constraint messaging for all
dependency relationships except bundle dependencies.


dinhxuanvu:
Replaced by #2005

PR #2005, opened 2021-02-12
Add meaningful string representations for resolution invariants

ecordell:

Original description:
> During dependency resolution, invariants are enforced for
one-operator-per-package and one-provider-per-api. Their user-facing
string representation (shown when they prevent resolution from
succeeding) had re-used the same representation as constraints on
Subscriptions out of convenience.

> Recently, the messaging of Subscription-related constraints was
changed to be more descriptive which inadvertently made the invariant
messaging misleading. This patch improves constraint messaging for all
dependency relationships except bundle dependencies.

This rebase of #1994 adds additional tests when multiple subscriptions 
exist for the same package. Those tests uncovered a weird UX for an
edge case: if two subscriptions are present for the same package, and 
the resolver manages to find an operator that satisfies both.

This is not likely or common today, but since it is a potential source of
confusion I updated the resolver to ensure that in that case, both 
subscriptions are updated to point to the operator that satisfies both.

Prior to that change, only one of the subscriptions that matched would 
be picked (whichever came first in the query).

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/registry/resolver/installabletypes.go
+++ b/pkg/controller/registry/resolver/installabletypes.go
@@ -5,7 +5,6 @@ import (
 	"strings"
 
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/registry"
-
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/registry/resolver/solver"
 )
 
@@ -37,7 +36,7 @@ func (i *BundleInstallable) AddDependency(dependencies []solver.Identifier) {
 }
 
 func (i *BundleInstallable) BundleSourceInfo() (string, string, registry.CatalogKey, error) {
-	info := strings.Split(string(i.identifier), "/")
+	info := strings.Split(i.identifier.String(), "/")
 	// This should be enforced by Kube naming constraints
 	if len(info) != 4 {
 		return "", "", registry.CatalogKey{}, fmt.Errorf("Unable to parse identifier %s for source info", i.identifier)
@@ -52,7 +51,7 @@ func (i *BundleInstallable) BundleSourceInfo() (string, string, registry.Catalog
 }
 
 func bundleId(bundle, channel string, catalog registry.CatalogKey) solver.Identifier {
-	return solver.Identifier(fmt.Sprintf("%s/%s/%s", catalog.String(), channel, bundle))
+	return solver.IdentifierFromString(fmt.Sprintf("%s/%s/%s", catalog.String(), channel, bundle))
 }
 
 func NewBundleInstallable(bundle, channel string, catalog registry.CatalogKey, constraints ...solver.Constraint) BundleInstallable {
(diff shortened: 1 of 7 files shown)


pkg/controller/registry/resolver/installabletypes.go
```diff
@@ -62,58 +61,100 @@ func NewBundleInstallable(bundle, channel string, catalog registry.CatalogKey, c
...
+	} else {
+		req = fmt.Sprintf("at least one of %s or %s", strings.Join(s[:len(s)-1], ", "), s[len(s)-1])
+	}
+	result.constraints = append(result.constraints, PrettyConstraint(solver.Dependency(dependencies...), fmt.Sprintf("subscription %s requires %s", name, req)))
+
+	return result
+}
+
+func NewSingleAPIProviderInstallable(group, version, kind string, providers []solver.Identifier) solver.Installable {
+	gvk := fmt.Sprintf("%s (%s/%s)", kind, group, version)
```
kevinrizza, line 106: nit: this seems like something that should be factored out if possible

pkg/controller/registry/resolver/installabletypes.go
```diff
@@ -62,58 +61,100 @@ func NewBundleInstallable(bundle, channel string, catalog registry.CatalogKey, c
...
+	result := GenericInstallable{
+		identifier: solver.IdentifierFromString(pkg),
+	}
+	if len(providers) <= 1 {
+		// The constraints are pointless without more than one provider.
+		return result
+	}
+	result.constraints = append(result.constraints, PrettyConstraint(solver.Mandatory(), fmt.Sprintf("there can be only one operator from package %s", pkg)))
+
+	var s []string
```
kevinrizza, line 136: super  nit: `providersString` maybe? This loop would probably be nicer as a closure but there's not good syntactic sugar to do that in go

pkg/controller/registry/resolver/resolver.go
```diff
@@ -449,12 +448,7 @@ func (r *SatResolver) addInvariants(namespacedCache MultiCatalogOperatorFinder,
 			if err != nil {
 				continue
 			}
-			key := fmt.Sprintf("gvkunique/%s/%s/%s", prop.Group, prop.Version, prop.Kind)
-			_, ok := conflictToInstallable[key]
-			if !ok {
-				conflictToInstallable[key] = make([]solver.Installable, 0)
-			}
-			conflictToInstallable[key] = append(conflictToInstallable[key], installable)
+			gvkConflictToInstallable[prop] = append(gvkConflictToInstallable[prop], installable.Identifier())
```
kevinrizza, line 451: nice

kevinrizza (commented):
/lgtm

ecordell:
/hold

The failure in the unit tests looks unrelated to this change, but it seems to happen consistently

dinhxuanvu:
The unit test passed locally for me

PR #2011, opened 2021-02-16
Bug 1925614: chore(adoption): use debug level to log missing components

njhale:
Use verbosity level 1 (debug) to log when a missing component is
encountered during adoption. This reduces log noise in situations where
the missing component is tolerable under normal operation; e.g. an
InstallPlan referenced by a Subscription has been garbage collected.


joelanford:
/lgtm

Looks like my operator deletion test is flaky :disappointed:

PR #2016, opened 2021-02-19
pkg/controller/operators/operator_controller_test.go: fix flaky test

joelanford:
Fixes a flake in the operator controller unit tests.

Flaky tests are no fun.

**Reviewer Checklist**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,7 +44,7 @@ all: test build
 test: clean cover.out
 
 unit: kubebuilder
-	KUBEBUILDER_ASSETS=$(KUBEBUILDER_ASSETS) go test $(MOD_FLAGS) $(SPECIFIC_UNIT_TEST) -tags "json1" -v -race -count=1 ./pkg/...
+	KUBEBUILDER_ASSETS=$(KUBEBUILDER_ASSETS) go test $(MOD_FLAGS) $(SPECIFIC_UNIT_TEST) -tags "json1" -race -count=1 ./pkg/...
 
 # Ensure kubebuilder is installed before continuing
 KUBEBUILDER_ASSETS_ERR := not detected in $(KUBEBUILDER_ASSETS), to override the assets path set the KUBEBUILDER_ASSETS environment variable, for install instructions see https://book.kubebuilder.io/quick-start.html
--- a/pkg/controller/operators/operator_controller_test.go
+++ b/pkg/controller/operators/operator_controller_test.go
@@ -45,12 +45,10 @@ var _ = Describe("Operator Controller", func() {
 	})
 
 	Describe("operator deletion", func() {
+		var originalUID types.UID
 		JustBeforeEach(func() {
+			originalUID = operator.GetUID()
 			Expect(k8sClient.Delete(ctx, operator)).To(Succeed())
-			Eventually(func() bool {
-				err := k8sClient.Get(ctx, name, operator)
-				return apierrors.IsNotFound(err)
-			}, timeout, interval).Should(BeTrue())
 		})
 		Context("with components bearing its label", func() {
 			var (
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -44,7 +44,7 @@ all: test build
 test: clean cover.out
 
 unit: kubebuilder
-	KUBEBUILDER_ASSETS=$(KUBEBUILDER_ASSETS) go test $(MOD_FLAGS) $(SPECIFIC_UNIT_TEST) -tags "json1" -v -race -count=1 ./pkg/...
```
joelanford, line 47: I removed `-v` here because it was _really_ hard to find the test failure in the log with all of the verbose output. If `-v` is here for a good reason, I can add it back.

dinhxuanvu (approved):
/lgtm

pkg/controller/operators/operator_controller_test.go
```diff
@@ -77,16 +75,62 @@ var _ = Describe("Operator Controller", func() {
...
+				//
+				// This race condition means we need to test for either of two possibilities when getting
+				// the operator:
+				//   a. We see a "not found" error followed by a nil error. This covers 1) and 2) above.
+				//   b. We eventually timeout looking for a), always seeing a nil error. This covers 3) above.
+
+				// This section waits until we've seen a not found error followed by a nil error. It
+				// returns any other unexpected error immediately.
+				gotNotFound := false
+				err := wait.PollImmediate(interval, timeout, func() (bool, error) {
```
njhale, line 93: Could we use [Gomega's Eventually](https://onsi.github.io/gomega/#eventually) here instead of `wait.PollImmediate`?

pkg/controller/operators/operator_controller_test.go
```diff
@@ -77,16 +75,62 @@ var _ = Describe("Operator Controller", func() {
...
-					return k8sClient.Get(ctx, name, operator)
-				}).Should(Succeed())
+				// There's a race condition between this test and the controller. By the time,
+				// this function is running, we may be in one of three states.
+				//   1. The original deletion in the test setup has not yet finished, so the original
+				//      operator resource still exists.
+				//   2. The operator doesn't exist, and the controller has not yet re-created it.
+				//   3. The operator has already been deleted and re-created.
+				//
+				// This race condition means we need to test for either of two possibilities when getting
```
njhale, line 85: Could we add a value to original resource that would change if it was recreated? This would make the check much simpler: just wait until a resource with the same name exists that has a changed "sentinel value". 

Initially, I was thinking that we could use the `meta.id` field, but since this is a unit test, I'm not sure if the behavior of the mock apiserver will be the same (i.e. id fields aren't generated or unique). If using the resource ID doesn't work, we could always add an annotation or label, then check if it no longer exists.

njhale (commented):
Great, thanks for fixing this!

I've also added a few comments for some potential changes.

pkg/controller/operators/operator_controller_test.go
joelanford: I tried several different ways of making `Eventually` work, but I was always foiled by the fact that this test basically is `Eventually(notFoundThenFound) OR Consistently(found)`.

There doesn't seem to be a way with Gomega to compose Assertions in this way, but I'd be happy to be proven wrong because that would make this test more clear.

pkg/controller/operators/operator_controller_test.go
joelanford: This is a good idea. This is running in envtest with a real etcd and kube-apiserver, so I think the meta.id should be different. If this works, then you're right, all we need to do is wait for a different UID.

joelanford:
/hold

I'm going to try the approach @njhale suggested here: https://github.com/operator-framework/operator-lifecycle-manager/pull/2016#discussion_r579422936

joelanford:
/hold cancel

Ok I think it's ready to go (assuming everything passes).

dinhxuanvu (approved):
/lgtm

PR #2021, opened 2021-03-01
Bug 1933779: Support InstallPlan steps upgrading existing ClusterIP Services.

benluddy:
If a ClusterIP-type Service manifest omits the immutable spec field
"ClusterIP", a value is selected automatically when the Service is
created. The catalog operator can patch existing Service resources if
the updated manifest either omits "ClusterIP" or sets "ClusterIP" to
the same value as that of the existing Service.


dinhxuanvu (approved):
/lgtm

benluddy:
The retest bot was spinning out because one of the (non-prow) upstream tests hit a flake, and /retest doesn't restart those. Putting a temporary hold to reduce bot spam until the upstream retest finishes.

/hold

derekwaynecarr:
(Patch-Manager) met criteria, approve.

benluddy:
The test failures don't appear connected to this change. I looked at the historical success rate for the 4.5 test job and it's been very low at least as far back as last September and the 4.6+ test jobs are significantly better.

PR #2023, opened 2021-03-01
Specify user in catalog pod securityContext

### LINKED ISSUE #1969: Set securityContext on pod created for a catalogSource
Can we get the pod created when there is a catalogSource to have its security context set to either run as a user or ensure the pod runs as non-root? The `configmap-operator-registry` image has already been set to run as a user via an [operator-registry PR](https://github.com/operator-framework/operator-registry/pull/425), but we also need the pod's securityContext set. (See https://kubernetes.io/docs/tasks/configure-pod-container/security-context/ for the kube docs)

At a glance, it looks like the pod definition that needs the securityContext set is [here](https://github.com/operator-framework/operator-lifecycle-manager/blob/5e982c5ef6157ab29140afcdea1606ee9b0ed205/pkg/controller/registry/reconciler/reconciler.go#L105)

pamelachristie:
Specify `RunAsUser` in catalog pod `SecurityContext` to match User specified in https://github.com/operator-framework/operator-registry/pull/425

Improve security for catalogs running on non-openshift distros

**Reviewer Checklist**

Closes #1969


exdx:
@pamelachristie there are some ongoing issues with the e2e-aws-console-olm series of tests that are being triaged by the console team -- not related to your PR in any way.

pamelachristie:
@exdx are these tests supposed to be fixed?

pamelachristie:
@exdx I only see one of the tests failing now, but it doesn't seem to re-try when I comment to retest. Is there anything I should be doing to get this looked at?

timflannagan:
@pamelachristie Can you rebase against the latest master? It doesn't look like we're able to kick that fail test on our side as well.

timflannagan:
@pamelachristie Oh, nice closing, and then re-opening worked. Disregard me last message :+1:

pkg/controller/registry/reconciler/reconciler_test.go
```diff
@@ -74,3 +74,20 @@ func TestPullPolicy(t *testing.T) {
...
+
+func TestPodRunAsUser(t *testing.T) {
+	source := &v1alpha1.CatalogSource{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "test",
+			Namespace: "test-ns",
+		},
+	}
+
+	p := Pod(source, "hello", "busybox", "", map[string]string{}, int32(0), int32(0))
```
timflannagan, line 86: It looks like you're missing arguments according to that failed unit test:

```
INFO[2021-05-03T18:49:03Z] pkg/controller/registry/reconciler/reconciler_test.go:86:10: not enough arguments in call to Pod 
INFO[2021-05-03T18:49:03Z] 	have (*"github.com/operator-framework/api/pkg/operators/v1alpha1".CatalogSource, string, string, string, map[string]string, int32, int32) 
INFO[2021-05-03T18:49:03Z] 	want (*"github.com/operator-framework/api/pkg/operators/v1alpha1".CatalogSource, string, string, string, map[string]string, map[string]string, int32, int32) 
```

pamelachristie: Should be updated in https://github.com/operator-framework/operator-lifecycle-manager/pull/2023/commits/f8df548373274250ae72e248f8639707998adccb

pkg/controller/registry/reconciler/reconciler_test.go
```diff
@@ -74,3 +74,20 @@ func TestPullPolicy(t *testing.T) {
...
+		},
+	}
+
+	p := Pod(source, "hello", "busybox", "", map[string]string{}, map[string]string{}, int32(0), int32(0))
+	user := p.Spec.SecurityContext.RunAsUser
+	if user == nil {
+		t.Fatal("expected RunAsUser value for pod expected to not be nil")
+	} else if *user != int64(1001) {
+		t.Fatalf("expected RunAsUser value for pod expected to be %v, not %v", int64(1001), *user)
+	}
```
timflannagan, line 92: Minor nit, but I think we can use the `require` package here to validate the Pod matches our expectations:

```go
require.NotNil(t, user, "expected RunAsUser ...")
require.EqualValues(t, ...)
```

pamelachristie: Should be updated in https://github.com/operator-framework/operator-lifecycle-manager/pull/2023/commits/60e36c0c998b77c40fa6af6759eca4bdeac478e6

pkg/controller/registry/reconciler/reconciler_test.go
```diff
@@ -1,7 +1,10 @@
 package reconciler
 
 import (
+	"fmt"
```
exdx, line 4: nit: imports

exdx: imports from the std-lib should be grouped together

pamelachristie: is there a make command or something to check that? this is how VSCode automatically imported it for me

exdx: there is a [goimports](https://pkg.go.dev/golang.org/x/tools/cmd/goimports) command that will sort the imports. I'm not sure if it will break them out into groups correctly, but that's something we like to do in our repos -- it makes external dependencies easier to see. 

Per @joelanford we can breakup imports into three groups:
* standard library packages
* other packages from within the project itself
* external dependencies

pamelachristie: let me know if https://github.com/operator-framework/operator-lifecycle-manager/pull/2023/commits/fe24a8ad57434ee19341291c30da53852320fd1a has the correct grouping

kevinrizza:
@pamelachristie Could you go into a bit of detail about the high level goal and value that this pull request is attempting to provide?

I took a look at the related pull request in the operator registry (https://github.com/operator-framework/operator-registry/pull/425), and I'm trying to piece together how they are related. That pull request sets the USER to 1001 for the configmap registry image builder, but this pull requests seems to expect that image to be used universally.

The operator registry does provide some tools and scaffolding that enable users to generate catalog images to present content to OLM on cluster, but the only contract that OLM actually expects from those catalog images is that they have the right consumable catalog content (either over the grpc API or in the right format that can be dropped into a configmap). That dockerfile definitely isn't what the majority of operator registry and OLM users are using to build catalog images -- much more likely, they have their own dockerfile or they scaffold it on every build using the command line tool `opm`.

I'm concerned that this PR actually just breaks any user who doesn't use that specific configmap dockerfile or doesn't have that USER in their container image.

Separately, I'm also wondering if its really the place of OLM to define strict requirements about running containers as root. I imagine for some upstream users, they may want to run as root for one reason or another. This sort of security configuration feels to me like it would be better suited for something like a pod security policy or even maybe something like the Gatekeeper project, at least while the greater kubernetes community is still using Docker as a container runtime.

All that being said, I am making some assumptions about the purpose and goals of this pull request given that there's not a lot of detail in the PR description or commit message. Do you have any thoughts on the above?

exdx:
It's hard to understand what happens when a catalog container image without the uid 1001 is run in a pod with the uid set to 1001 -- does the pod fail to start, silently switch to another uid, etc? I couldn't find any docs on this. 

There is also the `runAsNonRoot` that is mentioned in the [pod security context docs](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.21/#podsecuritycontext-v1-core) which might be a better option to enforce this. But as @kevinrizza mentioned, some users may want to run their catalogs as root and it's not necessarily up to OLM to stop them from doing so.

PR #2024, opened 2021-03-01
Bug 1932626: Gracefully handle service unavailable errors from kube-apiserver

joelanford:
Signed-off-by: Joe Lanford <joe.lanford@gmail.com>

This PR changes the CSV install status update logic to avoid failing a CSV when there are there are transient errors connecting to the API server.

OLM should not fire alerts during cluster upgrades when there are expected transient control plane connection issues.

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/olm/operator.go
+++ b/pkg/controller/operators/olm/operator.go
@@ -1782,6 +1782,11 @@ func (a *Operator) updateInstallStatus(csv *v1alpha1.ClusterServiceVersion, inst
 		return nil
 	}
 
+	if err := findFirstError(k8serrors.IsServiceUnavailable, strategyErr, apiServiceErr, webhookErr); err != nil {
+		csv.SetPhaseWithEventIfChanged(v1alpha1.CSVPhasePending, v1alpha1.CSVReasonDetectedClusterChange, err.Error(), now, a.recorder)
+		return err
+	}
+
 	// installcheck determined we can't progress (e.g. deployment failed to come up in time)
 	if install.IsErrorUnrecoverable(strategyErr) {
 		csv.SetPhaseWithEventIfChanged(v1alpha1.CSVPhaseFailed, v1alpha1.CSVReasonInstallCheckFailed, fmt.Sprintf("install failed: %s", strategyErr), now, a.recorder)
@@ -1826,6 +1831,15 @@ func (a *Operator) updateInstallStatus(csv *v1alpha1.ClusterServiceVersion, inst
 		return strategyErr
 	}
 
+	return nil
+}
+
+func findFirstError(f func(error) bool, errs ...error) error {
+	for _, err := range errs {
+		if f(err) {
+			return err
+		}
+	}
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


pkg/controller/operators/olm/operator.go
```diff
@@ -1782,6 +1782,11 @@ func (a *Operator) updateInstallStatus(csv *v1alpha1.ClusterServiceVersion, inst
 		return nil
 	}
 
+	if err := findFirstError(k8serrors.IsServiceUnavailable, strategyErr, apiServiceErr, webhookErr); err != nil {
+		csv.SetPhaseWithEventIfChanged(v1alpha1.CSVPhasePending, v1alpha1.CSVReasonDetectedClusterChange, err.Error(), now, a.recorder)
```
joelanford, line 1786: I'm not sure if this Phase and Reason makes sense. If a CSV is in phase Successful when this error occurs (which I think is generally the case during cluster upgrade scenarios), it looks like resetting to Pending will cause OLM to step through InstallReady and Installing before getting back to Succeeded.

Is there a better Phase and/or Reason to use? I couldn't find any other areas of OLM that deal with transient issues like this. Is it possible we need a new Phase or Reason?

pkg/controller/operators/olm/requirements.go
```diff
@@ -408,15 +407,15 @@ func (a *Operator) isGVKRegistered(group, version, kind string) error {
...
 	for _, r := range resources.APIResources {
 		if r.Kind == kind {
-			return nil
+			return true, nil
 		}
 	}
 
 	logger.Info("couldn't find GVK in api discovery")
-	return olmErrors.GroupVersionKindNotFoundError{group, version, kind}
+	return false, nil
```
joelanford, line 420: I could not find any usage of `isGVKRegistered` where the actual content of the error was used, so I changed the function signature to separate the found/not found question from any errors that may have occurred trying to arrive at the answer.

Any caller of `isGVKRegistered` could trivially re-construct the previous behavior with:
```go
func (a *Operator) oldIsGVKRegistered(group, version, kind string) error {
	ok, err := a.isGVKRegistered(group, version, kind)
	if err != nil {
		return err
	}
	if !ok {
		return olmErrors.GroupVersionKindNotFoundError{group, version, kind}
	}
	return nil
}

pkg/controller/operators/olm/operator.go
benluddy: I don't think we want transient errors to result in any updates to the CSV. Instead, the sync function can return a non-nil error and should be automatically retried later.

joelanford: That seems reasonable. The only concern I have with this approach is handling a "transient" error that's actually permanent for some reason.

In that case, there would never be an indication on the CSV that anything is wrong.

In practice a kube-apiserver that's unavailable would probably mean no user could actually query the CSV to know there was an issue AND they would probably be more worried about the kube-apiserver than the CSV.

joelanford: @benluddy Updated to return the error without changing CSV phase/reason.

smarterclayton:
This is still blocking us landing the "no alerts during upgrades" test: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/25904/pull-ci-openshift-origin-master-e2e-gcp-upgrade/1369043944154861568

dinhxuanvu (approved):
/lgtm

PR #2025, opened 2021-03-02
Bug 1932182: Support jittering relatively small resync intervals.

benluddy:
The previous jitter computation truncated intervals shorter than
around 1 minute to 0. Intervals that short are atypical for resyncs,
but still need to jitter to a reasonable value to avoid hot resync
loops.

### DIFF
--- a/pkg/lib/queueinformer/jitter.go
+++ b/pkg/lib/queueinformer/jitter.go
@@ -1,32 +1,45 @@
 package queueinformer
 
 import (
+	"math"
 	"math/rand"
 	"time"
 )
 
 const DefaultResyncPeriod = 15 * time.Minute
 
+type float64er interface {
+	// Float64 returns a float64 in range [0.0, 1.0).
+	Float64() float64
+}
+
+type realFloat64er struct{}
+
+func (realFloat64er) Float64() float64 {
+	return rand.Float64()
+}
+
 // ResyncWithJitter takes a resync interval and adds jitter within a percent difference.
 // factor is a value between 0 and 1 indicating the amount of jitter
 // a factor of 0.2 and a period of 10m will have a range of 8 to 12 minutes (20%)
 func ResyncWithJitter(resyncPeriod time.Duration, factor float64) func() time.Duration {
+	return resyncWithJitter(resyncPeriod, factor, realFloat64er{})
+}
+
+func resyncWithJitter(period time.Duration, factor float64, rand float64er) func() time.Duration {
 	return func() time.Duration {
-		if factor < 0.0 || factor > 1.0 {
-			return resyncPeriod
-		}
-		if resyncPeriod < 0.0 {
+		if period < 0.0 {
 			return DefaultResyncPeriod
 		}
-
-		// if we would wrap around, return resyncPeriod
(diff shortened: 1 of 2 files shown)


pkg/lib/queueinformer/jitter.go
```diff
@@ -1,32 +1,42 @@
...
+	return resyncWithJitter(resyncPeriod, factor, realFloat64er{})
+}
+
+func resyncWithJitter(period time.Duration, factor float64, rand float64er) func() time.Duration {
 	return func() time.Duration {
-		if factor < 0.0 || factor > 1.0 {
-			return resyncPeriod
-		}
-		if resyncPeriod < 0.0 {
+		if period < 0.0 {
```
exdx, line 30: Should this maybe be `period <= 0.0`? This would help with the extreme case where the duration comes out to zero that we have been seeing.

pkg/lib/queueinformer/jitter.go
```diff
@@ -1,32 +1,42 @@
...
-			return resyncPeriod
+		if factor < 0.0 || factor > 1.0 {
+			return period
 		}
-
-		min := resyncPeriod.Minutes() * (1 - factor)
-		max := resyncPeriod.Minutes() * (1 + factor)
-
-		return time.Duration(min)*time.Minute + time.Duration(rand.Float64()*(max-min))*time.Minute
+		if period > math.MaxInt64/2 { // 1281023h53m38.427387903s
```
exdx, line 36: nit: could we move this `period` comparison up a few lines right after the previous `period` comparison? Makes it a little easier to read.

pkg/lib/queueinformer/jitter.go
```diff
@@ -1,32 +1,42 @@
...
+			// avoid overflowing time.Duration
+			return period
+		}
+		return time.Duration((1 - factor + 2*rand.Float64()*factor) * float64(period))
```
exdx, line 40: Just an observation -- this is harder to reason about than the previous return statement that uses the `min` and `max` variables explicitly. Is there a way to make this return calculation a little more clear for future users?

exdx (commented):
Nice, I think this will definitely help with those CPU spikes.

I'm wondering if there is anything on the e2e or performance side we can add tests for to ensure that OLM performance is adequate (I think we may have even discussed this at one point). Not for this PR, but I think there are projects that do something like utilization-based profiling tests to ensure that these types of regressions don't occur. We tried adding resource limits to the OLM pods themselves but those didn't really work as intended

pkg/lib/queueinformer/jitter.go
benluddy: It might be better to set a higher floor on this because 1ns is effectively as bad as 0. I left it alone for now.

pkg/lib/queueinformer/jitter.go
benluddy: Added a comment here to explain how this relies on the range of values allowed for `factor` and returned by `rand.Float64()`.

benluddy:
> relerase-4.7

D'oh!

/cherry-pick release-4.7

PR #2029, opened 2021-03-04
Bug 1935909: Allow non-CSV-owned ServiceAccounts to satisfy CSV requirements.

benluddy:
The existing check, intended to prevent a race condition that could
occur during CSV upgrades, marks a ServiceAccount as NotPresent when
it has at least one owner but is not owned by the current CSV. It's
expected for the ServiceAccount "default" not to be have an
ownerreference to a CSV, so the check can be relaxed to only consider
owners of kind ClusterServiceVersion.

This also combines two ServiceAccount checks that were independently
added to address the same race condition into a single check.

### DIFF
--- a/pkg/controller/operators/olm/requirements.go
+++ b/pkg/controller/operators/olm/requirements.go
@@ -272,17 +272,8 @@ func (a *Operator) permissionStatus(strategyDetailsDeployment *v1alpha1.Strategy
 			// Check SA's ownership
 			if ownerutil.IsOwnedByKind(sa, v1alpha1.ClusterServiceVersionKind) && !ownerutil.IsOwnedBy(sa, csv) {
 				met = false
-				status.Status = v1alpha1.RequirementStatusReasonNotPresent
-				status.Message = "Service account is stale"
-				statusesSet[saName] = status
-				continue
-			}
-
-			// Check if the ServiceAccount is owned by CSV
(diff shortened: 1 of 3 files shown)


pkg/controller/operators/olm/requirements.go
```diff
@@ -273,16 +273,7 @@ func (a *Operator) permissionStatus(strategyDetailsDeployment *v1alpha1.Strategy
 			if ownerutil.IsOwnedByKind(sa, v1alpha1.ClusterServiceVersionKind) && !ownerutil.IsOwnedBy(sa, csv) {
 				met = false
 				status.Status = v1alpha1.RequirementStatusReasonNotPresent
```
estroz, line 275: Would this make more sense as `v1alpha1.RequirementStatusReasonPresentNotSatisfied` since the SA exists but the condition of use (not owned by another CSV) isn't satisfied?

benluddy: I'm not totally clear on the the intended semantic difference, but your interpretation seems reasonable to me. The tests aren't verifying the expected status per requirement, so I'll go ahead and fix that too.

benluddy:
/test e2e-aws-console-olm

This test failure should now be resolved by https://github.com/openshift/console/pull/8289.

njhale:
/lgtm cancel

@benluddy said he was going to add some e2e tests.

njhale (approved):
/lgtm

PR #2047, opened 2021-03-16
Bug 1939294: Avoid setting metadata.GracePeriodSeconds to zero seconds

timflannagan:
Update any references where we set the metadata.GracePeriodSeconds field
to zero seconds. If we need a Pod to be deleted as quick as possible,
use a single second.

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/registry/reconciler/grpc.go
+++ b/pkg/controller/registry/reconciler/grpc.go
@@ -226,7 +226,7 @@ func (c *GrpcRegistryReconciler) ensurePod(source grpcCatalogSourceDecorator, sa
 			return nil
 		}
 		for _, p := range currentLivePods {
-			if err := c.OpClient.KubernetesInterface().CoreV1().Pods(source.GetNamespace()).Delete(context.TODO(), p.GetName(), *metav1.NewDeleteOptions(0)); err != nil {
+			if err := c.OpClient.KubernetesInterface().CoreV1().Pods(source.GetNamespace()).Delete(context.TODO(), p.GetName(), *metav1.NewDeleteOptions(1)); err != nil {
 				return errors.Wrapf(err, "error deleting old pod: %s", p.GetName())
 			}
 		}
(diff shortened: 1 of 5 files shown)


timflannagan:
Hmm, I searched for that upgrade error in the openshift CI search tool and it looks like a known flake. Going to try and retest again before pinging the monitoring team.

/retest

pkg/controller/registry/reconciler/grpc.go
```diff
@@ -226,7 +226,7 @@ func (c *GrpcRegistryReconciler) ensurePod(source grpcCatalogSourceDecorator, sa
 			return nil
 		}
 		for _, p := range currentLivePods {
-			if err := c.OpClient.KubernetesInterface().CoreV1().Pods(source.GetNamespace()).Delete(context.TODO(), p.GetName(), *metav1.NewDeleteOptions(0)); err != nil {
+			if err := c.OpClient.KubernetesInterface().CoreV1().Pods(source.GetNamespace()).Delete(context.TODO(), p.GetName(), *metav1.NewDeleteOptions(1)); err != nil {
```
timflannagan, line 229: Probably easiest if I also add a package constant set to 1 and just reference that variable in `NewDeleteOptions(...)` instead of hardcoding the value of 1. I need to double-check what the default grace period seconds are, and whether we can just pass an empty metav1.DeleteOptions object instead of explicitly setting the grace period every time.

njhale: I'm partial to a package scoped constant _that is also_ the zero value for `metav1.DeleteOptions`: `metav1.DeleteOptions{}`.

According to the [type docs](https://pkg.go.dev/k8s.io/apimachinery@v0.21.0-beta.1/pkg/apis/meta/v1#DeleteOptions), as long as the `GracePeriodSeconds` field is `nil`:

> ... the default grace period for the specified type will be used.

njhale, line 229: Does this block for non-zero graceperiods? Should we set a deadline so we don't wait forever?

ecordell: The grace period is a setting for the apiserver, not the client

timflannagan:
That last upgrade test looks somewhat concerning, but going to retest anyway just in case a flake was produced.

/retest

PR #2054, opened 2021-03-19
Set labels for the operator Deployment created via the ClusterServiceVersion

### LINKED ISSUE #1522: Set labels for the operator Deployment created via the ClusterServiceVersion
## Feature Request

Would like the ability to set labels for the operator Deployment created via the ClusterServiceVersion.

As far as I understand, only owner labels are set by OLM, and the operator Deployment is limited to setting the DeploymentSpec, with no ability to set any metadata such as labels

https://github.com/operator-framework/operator-lifecycle-manager/blob/release-4.3/pkg/api/apis/operators/v1alpha1/clusterserviceversion_types.go#L62
https://github.com/operator-framework/operator-lifecycle-manager/blob/release-4.3/pkg/controller/install/deployment.go#L84

dmvolod:
With this changes is possible to set custom labels for Deployment in addition to the system labels set by the OLM. This labels should be present in the `labels` fields of the `deployments` section.  

Would like the ability to set labels for the operator Deployment created via the ClusterServiceVersion.
Only owner labels are set by OLM, and the operator Deployment is limited to setting the DeploymentSpec, with no ability to set any metadata such as labels.
Custom operator deployment labels is very desirable feature, for example when we need to lookup operator deployment from the operator code itself and the name of this deployment could be change potentially and we don't want to change the logic every update. Only custom CSV based labels could help in this case.

**Reviewer Checklist**

Closes #1522

### DIFF
--- a/doc/design/building-your-csv.md
+++ b/doc/design/building-your-csv.md
@@ -314,6 +314,8 @@ Ensure that the `serviceAccountName` used in the `deployment` spec matches one o
 
 Multiple Roles should be described to reduce the scope of any actions needed containers that the Operator may run on the cluster. For example, if you have a component that generates a TLS Secret upon start up, a Role that allows `create` but not `list` on Secrets is more secure than using a single all-powerful Service Account.
 
+It is also possible to set custom labels for Deployment in addition to the system labels set by the OLM. This labels should be present in the `labels` fields of the `deployments` section.  
+
 Here’s a full example:
 
 ```yaml
@@ -321,6 +323,9 @@ Here’s a full example:
     spec:
       deployments:
         - name: example-operator
+          labels:
+            application: example-operator
+            technology: general
           spec:
             replicas: 1
             selector:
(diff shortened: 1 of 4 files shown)


dmvolod:
@kevinrizza , @ecordell is there any chance to backport this feature to one of the release 4.6, 4.7, 4.8 or 4.9 or backports are only for bugs?
Thanks in advance.

kevinrizza:
@dmvolod for openshift backports are exclusively for bugs unfortuntely

prabhjotbawa:
Has this feature been rolled out to any of the 4.10 OpenShift releases? I followed the instructions and added the labels to the CSV but when it gets applied on the cluster, I don't see the labels getting persisted. 
Openshit release: 4.10.45
I checked the olm version from old-operator as shown below:
```
$ olm --version
OLM version: 0.19.0
git commit: 7837ef293f6d68073efdddab1f8af7309ebdec94
```
Is this feature released yet?

Thank you

PR #2060, opened 2021-03-26
Bump the remaining images to Go 1.16

timflannagan:
The recent ART/CI Dockerfile(s) sync updated the repository to build and run e2e tests against Go 1.16, but the rest of the repository, namely the upstream images and the unit tests, still need to be manually bumped.

**Reviewer Checklist**

### DIFF
--- a/base.Dockerfile
+++ b/base.Dockerfile
@@ -1,6 +1,7 @@
 # Dockerfile to bootstrap build and test in openshift-ci
 
-FROM openshift/origin-release:golang-1.15
+# TODO: do we need to use quay due to dockerhub rate limiting?
+FROM openshift/origin-release:golang-1.16
 
 # Install test dependencies
 RUN yum install -y skopeo && \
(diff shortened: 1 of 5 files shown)


base.Dockerfile
```diff
@@ -1,6 +1,7 @@
 # Dockerfile to bootstrap build and test in openshift-ci
 
-FROM openshift/origin-release:golang-1.15
+# TODO: do we need to use quay due to dockerhub rate limiting?
+FROM openshift/origin-release:golang-1.16
```
timflannagan, line 4: A quick note on this:

- No changes made to the build root Dockerfile will propagate to tests until it's been merged to master first, so we won't get any CI feedback from this, which should be fine but something to call out.
- We may want to avoid using an image hosted on docker.io due to potential flakes from rate limitations.

kevinrizza:
Can we update the upstream dockerfile to use 1.16 as well?

timflannagan:
@kevinrizza It looks like I had that changed locally, but didn't push those changes. Updated to include that upstream dockerfile.

timflannagan:
Need to wait for #2064 to land before CI is back to green.

/hold

PR #2064, opened 2021-03-29
Dockerfile: Add a symbolically linked release Dockerfile

timflannagan:
Introduce the operator-lifecycle-manager.Dockerfile that's symbolically
linked from the root Dockerfile.

This is likely only needed in the short term while we iron out the
upstream/downstream split, and needed as the openshift/release
ci-operator is expecting to build the OLM container image using the
operator-lifecycle-manager.Dockerfile (what's currently present in
openshfit/ocp-build-data for the OLM 4.8 images) and that file doesn't
exist anywhere in the repository.

/hold

**Reviewer Checklist**

### DIFF
--- a/operator-lifecycle-manager.Dockerfile
+++ b/operator-lifecycle-manager.Dockerfile
@@ -1,0 +1,1 @@
+Dockerfile


timflannagan:
This is passing the image test now.

/hold cancel

timflannagan:
This should be a no-op for CI, so I'm overriding the upgrades test as CI will permafail until this lands.

/override ci/prow/e2e-upgrade

PR #2068, opened 2021-03-30
Bug 1945261: Fix inconsistent dependency candidate order.

benluddy:
After bundle dependency candidates are sorted, a second query is made
to the catalog cache. This query filters by CSV name alone, so it's
ambiguous when the same CSV name appears in multiple channels (or
packages). The query appears to be redundant, so it has been removed.


benluddy:
/hold

There should be a regression test for this.

ecordell (commented):
/lgtm

PR #2069, opened 2021-03-30
phase one for making catalog system tests

jchunkins:

- update kind server to v0.10.0
- establish a git workflow that installs dependent tools and caches them
- add tools.md to be used as a git workflow cache key
- add a local docker registry for tests to use & hook up to kind server
- change kind provisioner to access the local docker registry
- update makefile to pass in tool binary location
- add toolsBin argument to e2e test
- add initial sample e2e tests...
  1. illustrate how catalog generation works
  2. test subscription with no dependencies
- add test/e2e/catalogutil package as home for catalog operations
- update api to v0.7.1 to pick up related images & skips
- execute make vendor to clean up dependencies

Signed-off-by: John Hunkins <jhunkins@us.ibm.com>

This is a first pass at creating "system level" e2e tests that leverages actual tools used to build catalogs and ensure that OLM handles those catalogs correctly. Effectively this PR creates a "framework" that allows you to create a catalog in a variety of different ways, which can subsequently be used in test cases where we create catalog sources using those catalogs, create subscriptions and then check for artifacts that would be generated by OLM to make sure the catalog is working as expected. The goal here is to create overarching e2e tests that exercise both the code from the registry and olm repos.

**Reviewer Checklist**

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -6,10 +6,33 @@ on:
   pull_request:
 jobs:
   run-e2e-tests:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
+    env:
+      TOOLS_BIN_RELATIVE_PATH: "tools/binary"
     steps:
       - uses: actions/checkout@v2
-      - run: make e2e-local NODES=2 JUNIT_DIRECTORY=./artifacts/
+
+      - name: Cache required binary files
+        uses: actions/cache@v2
+        id: cache
+        with:
+          path: ${{ env.TOOLS_BIN_RELATIVE_PATH }}
+          key: ${{ runner.os }}-${{ hashFiles(format('{0}/tools.md', env.TOOLS_BIN_RELATIVE_PATH)) }}
+      
+      - name: Install Dependencies
+        if: steps.cache.outputs.cache-hit != 'true'
+        run: "${GITHUB_WORKSPACE}/scripts/installDependencies.sh \"${GITHUB_WORKSPACE}/${TOOLS_BIN_RELATIVE_PATH}\""
+
+      - name: Test Installed
+        run: |
+          echo "[oc output] " && "${GITHUB_WORKSPACE}/${TOOLS_BIN_RELATIVE_PATH}/ocv4.5.0" version
+          echo "[opm upstream output] " && "${GITHUB_WORKSPACE}/${TOOLS_BIN_RELATIVE_PATH}/opmupv1.15.1" version
+
+      - name: start local registry and begin e2e-local tests
+        run: |
+          "${GITHUB_WORKSPACE}/scripts/start_registry.sh"
+          make e2e-local NODES=2 JUNIT_DIRECTORY=./artifacts/ TOOLS_BIN="${GITHUB_WORKSPACE}/${TOOLS_BIN_RELATIVE_PATH}"
+
       - name: Archive Test Artifacts # test results, failed or not, are always uploaded.
         if: ${{ always() }}
         uses: actions/upload-artifact@v2
--- a/scripts/installDependencies.sh
+++ b/scripts/installDependencies.sh
@@ -1,0 +1,136 @@
+#!/usr/bin/env bash
+
+
+TOOLS_BIN="${1?"Need to set full path to the location where binary tools will be downloaded"}"
+
+
+# the username used to access registry.redhat.io (OPTIONAL)
+RH_USER="${2:-}"
+# the password used to access registry.redhat.io (OPTIONAL)
+RH_PASSWORD="${3:-}"
+
+mkdir -p "${TOOLS_BIN}"
+
+# install upstream version of opm from github
+function installUptreamOpm()
+{
+    local version=${1:-"v1.15.1"}
+    local opmPrefix="opmup"
+    local platform=""
+    local architecture=""
+
+    # translate OSTYPE to expected platform name
+    if [[ "${OSTYPE}" == "linux-gnu"* ]]; then
+        platform="linux"
+    elif [[ "${OSTYPE}" == "darwin"* ]]; then
+        platform="darwin"
+    else
+        # unsupported os
+        return 1
+    fi
+
+    # translate HOSTTYPE to expected architecture name
+    if [[ "${HOSTTYPE}" == "x86_64" ]]; then
+        architecture="amd64"
+    elif [[ "${HOSTTYPE}" == "s390x" ]]; then
+        architecture="s390x"
+    elif [[ "${HOSTTYPE}" == "powerpc64le" ]]; then
+        architecture="ppc64le-opm"
+    else
+        # unsupported architecture
+        return 1
+    fi
+    
+    # download the upstream opm taking into account the version, platform, and architecture we're running on
+    assetURL=$(curl -s "https://api.github.com/repos/operator-framework/operator-registry/releases/tags/${version}" | jq --argjson platform "\"${platform}\"" --argjson architecture "\"${architecture}\"" -r '.assets[] | select(.browser_download_url | contains( ($platform + "-" + $architecture) ) ) | .url')
+    if ! "${TOOLS_BIN}/${opmPrefix}${version}" version 2>/dev/null && [[ -n ${assetURL} ]] ; then
+        curl -sL -H "Accept: application/octet-stream" -o "${TOOLS_BIN}/${opmPrefix}${version}" "${assetURL}" && chmod 755 "${TOOLS_BIN}/${opmPrefix}${version}"
+    fi
+
+}
+
+# installs downstream opm which requires docker for image extraction
+function installDownstreamOpm() 
+{
+    local opmCatalogBinaryImage="${1:-"registry.redhat.io/openshift4/ose-operator-registry:v4.6.0"}"
+    local version="${2:-"v1.14.3"}"
+    local opmPrefix="opmdown"
+    local executableName=""
+
+    if [[ "${OSTYPE}" == "linux-gnu"* ]]; then
+        executableName="opm"
+    elif [[ "${OSTYPE}" == "darwin"* ]]; then
+        executableName="darwin-amd64-opm"
+    else
+        # unsupported os
+        return 1
+    fi
+
+    # check if we already have it
+    if ! "${TOOLS_BIN}/${opmPrefix}${version}" version 2>/dev/null; then
+        # docker login to redhat registry
+        if echo "${RH_PASSWORD}" | docker login registry.redhat.io -u "\"${RH_USER}\"" --password-stdin &>/dev/null; then
+            # create a offline container
+            if containerReference=$(docker create "${opmCatalogBinaryImage}"); then
+                # extract binary to tools bin and make it executable
+                docker cp --follow-link "${containerReference}:/usr/bin/${executableName}" "${TOOLS_BIN}" && \
+                    mv "${TOOLS_BIN}/${executableName}" "${TOOLS_BIN}/${opmPrefix}${version}" && \
+                    chmod 755 "${TOOLS_BIN}/${opmPrefix}${version}"
+            else
+                # can't create offline container
+                return 1
+            fi
+        else
+            # can't login
+            return 1
+        fi
+    fi
+}
+
+# installs the "oc" executable
+function installOC() 
+{
+    local version=${1:-"4.5.0"}
+    local tarfile="oc.tar.gz"
+    local platform=""
+    local linux_wildcard_flag=""
+    local stripcomp=1
+    local ocprefix="ocv"
+
+    # linux tar requires a wildcard flag, but darwin does not
+    # also set platform specific URL 
+    if [[ "${OSTYPE}" == "linux-gnu"* ]]; then
+        linux_wildcard_flag="yes"
+        platform="linux"
+    elif [[ "${OSTYPE}" == "darwin"* ]]; then
+        linux_wildcard_flag=""
+        platform="mac"
+    else
+        # unsupported os
+        return 1
+    fi
+    # check if we already have it
+    if ! "${TOOLS_BIN}/${ocprefix}${version}" version 2>/dev/null; then
+        # download specific tar for version and platform, and make sure the download succeeds
+        if curl -sL -H "Accept: application/octet-stream" "https://mirror.openshift.com/pub/openshift-v4/clients/ocp/${version}/openshift-client-${platform}-${version}.tar.gz" -o "${tarfile}" && [[ -f "${tarfile}" ]]; then
+            # figure out if we need to strip off leading path elements
+            [[ $(tar tvf ${tarfile} | grep -E -v kubectl | grep -E '/oc$| oc$' | rev | cut -f1 -d' ' | rev) =~ ^oc$ ]] && stripcomp=0 || stripcomp=1
+            # extract oc executable, move it to tools bin with different name and cleanup tarfile
+            tar xvf ${tarfile} --strip-components=$stripcomp --directory "${TOOLS_BIN}" ${linux_wildcard_flag:+'--wildcards'} '*oc' && \
+                mv "${TOOLS_BIN}/oc" "${TOOLS_BIN}/${ocprefix}${version}" && \
+                rm "${tarfile}"
+        else
+            return 1
+        fi
+    fi
+}
+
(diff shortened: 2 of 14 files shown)


jchunkins:
Note: we intend to submit further tests beyond what is included here, but wanted to get this in front of folks before we invest more time in creating these tests. This PR illustrates the overall flow for the tests we intend to create.

jchunkins:
@ecordell This pull request downloads dependent tools and caches them. One of the dependent tools is downstream opm which requires credentials to download.  So in I have the values defined as GitHub secrets:

https://github.com/operator-framework/operator-lifecycle-manager/pull/2069/files#diff-194218c48b9a0cdd03974145733804c2d992ca818529fe2fa69a501d8b5b1cc3R22-R25

```
      - name: Install Dependencies
         env:
           RH_PASSWORD: ${{ secrets.RH_PASSWORD }}
           RH_USER: ${{ secrets.RH_USER }}
```

Here's an example of the secrets defined in my fork https://github.com/jchunkins/operator-lifecycle-manager/settings/secrets/actions

image

I'd like to be able to get credentials added to secrets in this repo and then get a `ok to test` label to make sure everything works as expected.

Thanks!

jchunkins:
I removed the reliance on GitHub secrets but left in the code to allow for the possibility of downloading the downstream version of opm if we ever need to. That way a downstream fork of the tests could use the same script.

At any rate, the tests all pass so it should be in a good state to review.

.github/workflows/e2e-tests.yml
```diff
@@ -6,10 +6,33 @@ on:
   pull_request:
 jobs:
   run-e2e-tests:
-    runs-on: ubuntu-latest
```
exdx, line 9: Hi @jchunkins, 

First off thank you for putting this big effort to improve the OLM CI system by testing across various different components. This PR captures exactly the sort of thing that we would want to run against every OLM and registry PR. 

That being said, I propose the idea of moving this "OLM test infrastructure" into a standalone repo, so that we can make specific github actions that can run these infrastructure tests against any PR in operator-framework. 

Here are some reasons why we don't necessarily want test infrastructure like this in the OLM repo:

- We recently simplified the OLM CI greatly by removing all the openshift specific CI runs which cut down a lot of flakes. Now the OLM CI is quick and easy to run locally as it runs entirely on a local k8s cluster. We think keeping it this way makes the most sense, and want to avoid encoding a lot of steps into our github actions. Tests should be entirely self contained and not require documentation to be able to run. 
- The way this PR changes test cluster provisioning adds new expectations for additional test setup before running go test, which adds more complexity when developing locally. 
- Making a test catalog in this PR is complex. We've struggled ourselves with the problem of building catalogs locally in e2e tests versus hardcoding quay images. I think we want to solve this by using a declarative config test fixture that can then be converted into a catalogsource. But this is still an open problem.

So that being said, what do you think about moving this PR to a standalone repo, importing all the necessary bits, and running the tests this way. From there we can iterate and integrate this into the OLM CI when it stabilizes. I can also check to see if we can make a designated operator-framework repository for this infrastructure to live in once its stabilized. 

More tests is not strictly better, but a few (3-5) of these overarching tests could be valuable and provide more confidence in the OLM suite working as intended.

scripts/installDependencies.sh
```diff
@@ -0,0 +1,136 @@
...
+    if ! "${TOOLS_BIN}/${opmPrefix}${version}" version 2>/dev/null && [[ -n ${assetURL} ]] ; then
+        curl -sL -H "Accept: application/octet-stream" -o "${TOOLS_BIN}/${opmPrefix}${version}" "${assetURL}" && chmod 755 "${TOOLS_BIN}/${opmPrefix}${version}"
+    fi
+
+}
+
+# installs downstream opm which requires docker for image extraction
+function installDownstreamOpm() 
+{
+    local opmCatalogBinaryImage="${1:-"registry.redhat.io/openshift4/ose-operator-registry:v4.6.0"}"
```
exdx, line 55: we should take care not to pull downstream specific content into the upstream tests

scripts/installDependencies.sh
```diff
@@ -0,0 +1,136 @@
...
+        executableName="darwin-amd64-opm"
+    else
+        # unsupported os
+        return 1
+    fi
+
+    # check if we already have it
+    if ! "${TOOLS_BIN}/${opmPrefix}${version}" version 2>/dev/null; then
+        # docker login to redhat registry
+        if echo "${RH_PASSWORD}" | docker login registry.redhat.io -u "\"${RH_USER}\"" --password-stdin &>/dev/null; then
```
exdx, line 72: I think we discussed this in our call, but having tests depend on secrets like this would not allow users on forks to run these tests (unless they checked in their own credentials on their fork). The test would only run on commits to master in the upstream repository. We should avoid requiring secrets like this if possible.

scripts/installDependencies.sh
```diff
@@ -0,0 +1,136 @@
...
+            fi
+        else
+            # can't login
+            return 1
+        fi
+    fi
+}
+
+# installs the "oc" executable
+function installOC()
```
exdx, line 91: `oc` is a openshift-specific tool and we don't necessarily want to depend on it in our OLM upstream repository

scripts/start_registry.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+set -o errexit
+
+# create registry container unless it already exists
+reg_name='kind-registry'
+reg_port='5000'
+running="$(docker inspect -f '{{.State.Running}}' "${reg_name}" 2>/dev/null || true)"
+if [ "${running}" != 'true' ]; then
+  docker run \
```
exdx, line 9: is it possible to spin up this registry from within the cluster instead?

test/e2e/catalogutil/catalog.go
```diff
@@ -0,0 +1,1537 @@
...
+
+• catalogEntries: a slice of catalog entries that need to be added into a catalog
+
+• stack: contains arguments needed to generate a catalog for a specific environment
+
+Returns:
+
+error is non nil if anything prevented the catalog from being generated and pushed to the target registry
+*/
+func CreateTemporaryCatalog(toolsBin string, catalogEntries []CatalogEntry, stack Stack) error {
```
exdx, line 722: This is great, and it's frankly impressive that you got this to work. With the move to declarative config based indexes though, I wonder if there is a way to greatly simplify this by simply using config test fixtures.

test/e2e/catalogutil/catalog.go
```diff
@@ -0,0 +1,1537 @@
...
+
+• destinationDir: The "root destination" directory that contains the manifest files (where the bundles.db is located)
+
+• stack: contains arguments needed to generate a catalog for a specific environment
+
+Returns:
+
+error is non nil if anything prevented the catalog image from being created (NOTE: pushing occurs independently)
+*/
+func CreateIndexFromDatabase(destinationDir string, stack Stack) error {
```
exdx, line 1025: This assumes a database is present in the catalog image, which is something we are actively trying to move away from. Testing both existing catalogs and the newer declarative config-based ones would be challenging, but it could be extremely useful. In cases like that, we probably don't want to assume the index is one way or another until we actually unpack and inspect its metadata contents.

test/e2e/catalogutil/catalog.go
```diff
@@ -0,0 +1,1537 @@
...
+				Singular: ownedCRD.SingluarName,
+			},
+			Scope:    apiextensionsv1.NamespaceScoped,
+			Versions: versions,
+		},
+	}
+}
+
+// createV1Beta1CRD creates a v1beta1 CRD
+func createV1Beta1CRD(ownedCRD CRDInformation, versions []apiextensionsv1beta1.CustomResourceDefinitionVersion) *apiextensionsv1beta1.CustomResourceDefinition {
```
exdx, line 1265: We probably don't want to create any new v1beta1 CRDs at all since they are being deprecated in kube 1.22. 

We need to overhaul our own e2e tests to move to v1 CRDs entirely at some point in the near future

test/e2e/ctx/provisioner_kind.go
```diff
@@ -74,8 +79,22 @@ func Provision(ctx *TestContext) (func(), error) {
...
+	// 2. wait for 5 minutes for control plane
+	// 3. use explicit kube config path
+	registry := "kind-registry"
+	// registry := "localhost"
+	registryPort := 5000
 	if err := provider.Create(
 		name,
+		cluster.CreateWithV1Alpha4Config(&v1alpha4.Cluster{
+			ContainerdConfigPatches: []string{
+				fmt.Sprintf(`[plugins."io.containerd.grpc.v1.cri".registry.mirrors."localhost:%[2]d"]
```
exdx, line 94: this is interesting...what does this do exactly? It creates the cluster and mirrors content from a registry outside the cluster to be on all the kind "nodes" within the cluster? Does it continuously sync as new images are pushed to the external registry?

test/e2e/sample_e2e_test.go
```diff
@@ -0,0 +1,174 @@
...
+			TargetCatalogType: cu.Image,
+		}
+
+		err := cu.CreateTemporaryCatalog(*toolsBin, catalogEntry, stack)
+		_ = err
+	})
+
+	Describe("test with catalog index image created in replaces mode", func() {
+
+		It("create subscription with no dependency", func() {
```
exdx, line 109: I know this is just an example test, but I just want to point out that we do accomplish the same functionality by testing each discrete unit instead. For example, given an installplan, do we get the expected objects on cluster. From a subscription to a catalog source, do we get the expected resolution and an installplan with the right items in it on-cluster. 

These overarching tests can be extremely powerful if we come up with some scenarios that are truly beyond what testing each individual piece does. Most of the e2e test cases that we write as part of bug-fixing are extremely specific and tied to edge cases in the individual APIs versus the overall interaction of the system. The overall interaction is interesting, we should brainstorm some interesting scenarios that we can test.

exdx (commented):
Great job, really respect the amount of effort that went into this. This review is largely based on feedback from the OLM team regarding this PR, others can chime in as they see fit. 

Look forward to discussing this more in upcoming OLM community meetings.

scripts/installDependencies.sh
jchunkins: From previous comment:
> I removed the reliance on GitHub secrets but left in the code to allow for the possibility of downloading the downstream version of opm if we ever need to. That way a downstream fork of the tests could use the same script.

Was reluctant to pull out working code when it could be useful for a downstream fork to use when the time comes.

scripts/installDependencies.sh
jchunkins: From previous comment:
> I removed the reliance on GitHub secrets but left in the code to allow for the possibility of downloading the downstream version of opm if we ever need to. That way a downstream fork of the tests could use the same script.

This code won't fire unless you feed it credentials so its effectively a no-op. As I said in the other comment, I was reluctant to pull out working code when it could be useful for a downstream fork to use when the time comes.

scripts/installDependencies.sh
jchunkins: Our usage of this tool is currently non-existent... I had it in here since some of our existing bash based test scripts used oc and thought we might need it. Using kubectl (if necessary at some point) is probably sufficient. @cdjohnson might have ideas on why this might be needed.

scripts/start_registry.sh
jchunkins: This is a major rabbit hole, but here we go...

Testing with either SSL or non-SSL registries has its own set of problems due to the way docker daemon needs to be configured. Note that the operator-registry tests skip over this whole problem for docker because of the issues involved. Note that operator-registry has its own in-process docker registry that can start up and run and it can create a CA cert that it can hand to the containerd tooling to make things work with SSL. This won't work for docker tooling because the CA cert has to be available to the docker daemon. If you want to configure a per-registry SSL configuration there is a mechanism for this, but you have to configure it and then restart the docker daemon (see 
https://docs.docker.com/engine/security/certificates/). If you want to setup insecure registries via the `insecure-registry` configuration, you can do that, but again it requires the docker daemon to be restarted (see https://docs.docker.com/registry/insecure/). 

Bottom line for SSL is if you can get the system to trust your CA, then things go easier. Note that using a local docker registry from **within a container** works around a bunch of different problems relating to networking and getting SSL to work. The docker network can allow you to bridge networks between containers (which is what's happening for the kind server in this PR). I've also seen that it makes insecure registries work properly with docker push/pull (I presume this is working because of the out of box host networking configuration).

Note that we put a variation of this script, which is far more capable, into https://github.com/operator-framework/operator-registry/blob/master/scripts/start_registry.sh where we spin up a registry. Note that it will correctly handle setting up CA certs in a cross platform way since it is using [mkcert](https://github.com/FiloSottile/mkcert) under the hood to do this "right". 

One idea I just had would be to make use of this more capable shell script but execute it when the testing actually starts, which would remove the extra step that a user would have to do to execute the tests. You can also just change the make file call the script too, which accomplishes the same thing.

test/e2e/catalogutil/catalog.go
jchunkins: There's definitely work to do for handling the new catalog style, but I think if we start with this PR we at least cover the existing catalogs. My thinking on this is making the catalog creation function `CreateTemporaryCatalog` use another path for handling the new catalog type by adding a new `TargetCatalogType` with the `Stack`.

test/e2e/catalogutil/catalog.go
jchunkins: I think the idea for using v1beta1 is to handle the complexity of what happens when we migrate from one to the other when your operator gets upgraded. There's a whole series of test cases we were thinking of creating around this.

test/e2e/ctx/provisioner_kind.go
jchunkins: This configures the kind cluster configuration to use the local registry.  See these pages:

https://kind.sigs.k8s.io/docs/user/local-registry/
https://github.com/containerd/containerd/blob/master/docs/cri/registry.md#configure-registry-endpoint

Note that the current configuration in the PR uses `http` since that's how the registry is currently setup.

test/e2e/sample_e2e_test.go
jchunkins: I know that @cdjohnson has a whole set of test plans written down that we were going to implement.

.github/workflows/e2e-tests.yml
jchunkins: To make this work in a separate repo, we'd need access to opm to create catalogs, but then also OLM builds to run inside a cluster... Maybe getting the [quickstart](https://github.com/operator-framework/operator-lifecycle-manager/tree/master/deploy/upstream/quickstart) files and running it against a minikube/kind cluster? Not sure where the test code comes into play here either. Open to suggestions on how you'd make this work.

scripts/installDependencies.sh
cdjohnson: `kubectl` is sufficient.  `oc` is a legacy artifact of my tool and has no dependencies on oc itself.

test/e2e/catalogutil/catalog.go
cdjohnson: We started this test suite MANY months ago to explicitly test out the `opm` imperative design.  I think we should finish this quickly and then move on to another declarative approach.

.github/workflows/e2e-tests.yml
cdjohnson: Because we've already been working on this for many months, the OLM teams initial suggestion that we in fact put it here, I would rather keep with the original design and then tackle a new PR that carves it out into something new and could be enhanced for the new new declarative index design.  If we spend too much more time chasing the ball around, this will never be finished.

scripts/installDependencies.sh
jchunkins: Thanks for clarification Chris. I'll pull the installOC portions out then.

test/e2e/catalogutil/catalog.go
cdjohnson: I agree, @jchunkins

timflannagan:
Thanks for tackling this work John - it looks like we've lost some steam with reviewing this PR, and we're now running into merge conflicts. I think pursuing this could be valuable, but it would be nice to offload this kind of thing to it's own repository, or house a common operator-framework/test-infra so we have a more consistent testing story throughout the OLM-controlled repositories. I'm going to close this out in the meantime.

PR #2073, opened 2021-04-01
Bump k8s dependencies to v0.20.4

timflannagan:
Avoid pinning against an untagged version of the k8s.io/apiserver
dependency, which contains the webhook authorization panic fix.

Instead, pin against the v0.20.4 tag, which contains that same upstream
fix.

Bump the remaining k8s.io/* dependencies to that v0.20.4 version

**Reviewer Checklist**


go.mod
```diff
@@ -72,7 +72,7 @@ replace (
 
 	// pinned for webhook authorizer panic fix
 	// https://github.com/kubernetes/kubernetes/pull/97820
```
ecordell, line 74: can just remove the `replace` entirely here for apiserver

timflannagan: I think the potential problem here is that the go.sum will then contain multiple k8s.io/apiserver versions, including that problematic v0.20.x minor version(s) that don't contain the panic fix:

```bash
$ cat go.sum | grep k8s.io/apiserver
k8s.io/apiserver v0.20.4 h1:zMMKIgIUDIFiwK3LyY7qOV4Z4wKsHVYExL6vXY9fPX4=
k8s.io/apiserver v0.20.4/go.mod h1:Mc80thBKOyy7tbvFtB4kJv1kbdD0eIH8k8vianJcbFM=
...
# Remove the replaces directive for the k8s.io/apiserver
$ make vendor
$ k8s.io/apiserver v0.18.2/go.mod h1:Xbh066NqrZO8cbsoenCwyDJ1OSi8Ag8I2lezeHxzwzw=
k8s.io/apiserver v0.20.0/go.mod h1:6gRIWiOkvGvQt12WTYmsiYoUyYW0FXSiMdNl4m+sxY8=
k8s.io/apiserver v0.20.1/go.mod h1:ro5QHeQkgMS7ZGpvf4tSMx6bBOgPfE+f52KwvXfScaU=
k8s.io/apiserver v0.20.4 h1:zMMKIgIUDIFiwK3LyY7qOV4Z4wKsHVYExL6vXY9fPX4=
k8s.io/apiserver v0.20.4/go.mod h1:Mc80thBKOyy7tbvFtB4kJv1kbdD0eIH8k8vianJcbFM=
...
```

benluddy: That's ok. From https://golang.org/ref/mod#go:

> The go.sum file may contain hashes for multiple versions of a module. The go command may need to load go.mod files from multiple versions of a dependency in order to perform minimal version selection. go.sum may also contain hashes for module versions that aren't needed anymore (for example, after an upgrade). go mod tidy will add missing hashes and will remove unnecessary hashes from go.sum.

So as long as you are running `make vendor` (which runs `go mod tidy`) then there shouldn't be a problem if there are multiple versions of k8s.io/apiserver in go.sum.

timflannagan: Thanks for diving into that. I can remove the replace pin.

dinhxuanvu (approved):
/lgtm

PR #2076, opened 2021-04-05
Preserve existing ServiceAccount owner references during installs.

benluddy:
A ServiceAccount requirement in a ClusterServiceVersion is considered
unsatisfied if it is not owned by the CSV _and_ it is owned by a
different CSV. During InstallPlan execution, owner references in
ServiceAccount steps replace existing owner references. As a result, a
CSV that is being replaced may transition to Failed (requirements not
met) and block replacement.

### DIFF
--- a/pkg/controller/operators/catalog/step_ensurer.go
+++ b/pkg/controller/operators/catalog/step_ensurer.go
@@ -156,6 +156,7 @@ func (o *StepEnsurer) EnsureServiceAccount(namespace string, sa *corev1.ServiceA
 		return
 	}
 	sa.Secrets = preSa.Secrets
+	sa.OwnerReferences = mergedOwnerReferences(preSa.OwnerReferences, sa.OwnerReferences)
 
 	sa.SetNamespace(namespace)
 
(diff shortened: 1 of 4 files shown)


timflannagan:
I'll hold just in case someone else wants to take a look before the e2e-aws-olm test passes.

/lgtm
/hold

pkg/controller/operators/catalog/step_ensurer.go
```diff
@@ -349,3 +350,20 @@ func (o *StepEnsurer) EnsureConfigMap(namespace string, configmap *corev1.Config
 	status = v1alpha1.StepStatusPresent
 	return
 }
+
+func mergedOwnerReferences(in ...[]metav1.OwnerReference) []metav1.OwnerReference {
+	uniques := make(map[metav1.OwnerReference]struct{})
+	for _, refs := range in {
+		for _, ref := range refs {
+			uniques[ref] = struct{}{}
```
timflannagan, line 358: The only potential concern I have is whether we need to check if that `ref` key already exists, but I'm almost positive we're safe when we use this implementation?

timflannagan: Yeah, this looks like it would be safe re-reading this again as we're only initializing that uniques map.

timflannagan:
We can remove the hold if this needs another reviewer before merging.

/lgtm

ecordell:
/lgtm
/hold cancel

discussed offline, but there's a similar issue that can happen with roles/rolebindings, so we should file a bug as a followup before anyone hits it.

timflannagan:
No need to respin clusters and take up leases while we wait for those checks to be removed in an inflight openshift/release PR.

/override ci/prow/e2e-aws-console-olm
/override ci/prow/e2e-gcp

PR #2077, opened 2021-04-06
Indicate invalid OperatorGroup on InstallPlan status

hasbro17:
The InstallPlan reconciler/sync will now update the InstallPlan phase as `Failed` along with a status condition message
if it sees an invalid OperatorGroup.
An invalid OperatorGroup is one of:
- No OperatorGroups in the InstallPlan's namespace
- Multiple OperatorGroups in the InstallPlan's namespace
- An incorrect or non-existent ServiceAccount name specified on the OperatorGroup

Previously an InstallPlan would stay in the `Installing` phase indefinitely if the installplan sync encountered an invalid OperatorGroup.
With this change, the failure is more readily apparent and the InstallPlan status condition is also propagated to the Subscription's status condition.

```Yaml
kind: InstallPlan
status:
  conditions:
    - lastTransitionTime: '2021-04-06T01:10:01Z'
      lastUpdateTime: '2021-04-06T01:10:01Z'
      message: >-
        invalid operator group - no operator group found that
        is managing this namespace
      reason: InstallCheckFailed
      status: 'False'
      type: Installed
  phase: Failed
```

```Yaml
kind: Subscription
...
status:
  ...
  conditions:
    ...
    - lastTransitionTime: '2021-04-06T01:29:16Z'
      reason: InstallCheckFailed
      status: 'True'
      type: InstallPlanFailed
```

See feature request: https://issues.redhat.com/browse/OLM-2116

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -1331,10 +1331,44 @@ func (o *Operator) syncInstallPlans(obj interface{}) (syncError error) {
 		return
 	}
 
+	if plan.Status.Phase == v1alpha1.InstallPlanPhaseFailed {
+		return
+	}
+
 	querier := o.serviceAccountQuerier.NamespaceQuerier(plan.GetNamespace())
 	ref, err := querier()
 	if err != nil {
-		syncError = fmt.Errorf("attenuated service account query failed - %v", err)
+
+		// Retry sync if non-fatal error
+		if !scoped.IsOperatorGroupError(err) {
+			syncError = fmt.Errorf("attenuated service account query failed - %v", err)
+			return
+		}
+
+		// Mark the InstallPlan as failed for a fatal Operator Group related error
+		ipFailError := fmt.Errorf("attenuated service account query failed - %v", err)
+		logger.Infof("InstallPlan failed: %v", ipFailError)
+
+		now := o.now()
+		out := plan.DeepCopy()
+		out.Status.SetCondition(v1alpha1.ConditionFailed(v1alpha1.InstallPlanInstalled,
+			v1alpha1.InstallPlanReasonInstallCheckFailed, ipFailError.Error(), &now))
+		out.Status.Phase = v1alpha1.InstallPlanPhaseFailed
+
+		logger.Info("Transitioning InstallPlan to failed")
+		if _, err := o.client.OperatorsV1alpha1().InstallPlans(plan.GetNamespace()).UpdateStatus(context.TODO(), out, metav1.UpdateOptions{}); err != nil {
+			updateErr := errors.New("error updating InstallPlan status: " + err.Error())
+			logger = logger.WithField("updateError", updateErr)
+			logger.Info("error transitioning InstallPlan to failed")
+
+			// retry sync with error to update InstallPlan status
+			syncError = fmt.Errorf("InstallPlan failed: %s and error updating InstallPlan status as failed: %s", ipFailError, updateErr)
+			return
+		}
+
+		// Requeue subscription to propagate SubscriptionInstallPlanFailed condtion to subscription
+		o.requeueSubscriptionForInstallPlan(plan, logger)
+
 		return
 	}
 
@@ -1394,19 +1428,7 @@ func (o *Operator) syncInstallPlans(obj interface{}) (syncError error) {
 		defer o.ipQueueSet.RequeueAfter(outInstallPlan.GetNamespace(), outInstallPlan.GetName(), time.Second*5)
 	}
 
-	defer func() {
-		// Notify subscription loop of installplan changes
-		if owners := ownerutil.GetOwnersByKind(plan, v1alpha1.SubscriptionKind); len(owners) > 0 {
-			for _, owner := range owners {
-				logger.WithField("owner", owner).Debug("requeueing installplan owner")
-				if err := o.subQueueSet.Requeue(plan.GetNamespace(), owner.Name); err != nil {
-					logger.WithError(err).Warn("error requeuing installplan owner")
-				}
-			}
-			return
-		}
-		logger.Trace("no installplan owner subscriptions found to requeue")
-	}()
+	defer o.requeueSubscriptionForInstallPlan(plan, logger)
 
 	// Update InstallPlan with status of transition. Log errors if we can't write them to the status.
 	if _, err := o.client.OperatorsV1alpha1().InstallPlans(plan.GetNamespace()).UpdateStatus(context.TODO(), outInstallPlan, metav1.UpdateOptions{}); err != nil {
@@ -1421,6 +1443,20 @@ func (o *Operator) syncInstallPlans(obj interface{}) (syncError error) {
 	}
 
 	return
+}
+
+func (o *Operator) requeueSubscriptionForInstallPlan(plan *v1alpha1.InstallPlan, logger *logrus.Entry) {
+	// Notify subscription loop of installplan changes
+	if owners := ownerutil.GetOwnersByKind(plan, v1alpha1.SubscriptionKind); len(owners) > 0 {
+		for _, owner := range owners {
+			logger.WithField("owner", owner).Debug("requeueing installplan owner")
+			if err := o.subQueueSet.Requeue(plan.GetNamespace(), owner.Name); err != nil {
+				logger.WithError(err).Warn("error requeuing installplan owner")
+			}
+		}
+		return
+	}
+	logger.Trace("no installplan owner subscriptions found to requeue")
 }
 
 type installPlanTransitioner interface {
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -1331,7 +1331,8 @@ func (o *Operator) syncInstallPlans(obj interface{}) (syncError error) {
 		return
 	}
 
-	if plan.Status.Phase == v1alpha1.InstallPlanPhaseFailed {
+	// Complete and Failed are terminal phases
+	if plan.Status.Phase == v1alpha1.InstallPlanPhaseFailed || plan.Status.Phase == v1alpha1.InstallPlanPhaseComplete {
 		return
 	}
 
@@ -1446,16 +1447,19 @@ func (o *Operator) syncInstallPlans(obj interface{}) (syncError error) {
 
 func (o *Operator) requeueSubscriptionForInstallPlan(plan *v1alpha1.InstallPlan, logger *logrus.Entry) {
 	// Notify subscription loop of installplan changes
-	if owners := ownerutil.GetOwnersByKind(plan, v1alpha1.SubscriptionKind); len(owners) > 0 {
-		for _, owner := range owners {
-			logger.WithField("owner", owner).Debug("requeueing installplan owner")
-			if err := o.subQueueSet.Requeue(plan.GetNamespace(), owner.Name); err != nil {
-				logger.WithError(err).Warn("error requeuing installplan owner")
-			}
-		}
+	owners := ownerutil.GetOwnersByKind(plan, v1alpha1.SubscriptionKind)
+
+	if len(owners) == 0 {
+		logger.Trace("no installplan owner subscriptions found to requeue")
 		return
 	}
-	logger.Trace("no installplan owner subscriptions found to requeue")
+
+	for _, owner := range owners {
+		logger.WithField("owner", owner).Debug("requeueing installplan owner")
+		if err := o.subQueueSet.Requeue(plan.GetNamespace(), owner.Name); err != nil {
+			logger.WithError(err).Warn("error requeuing installplan owner")
+		}
+	}
 }
 
 type installPlanTransitioner interface {
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -1346,20 +1346,19 @@ func (o *Operator) syncInstallPlans(obj interface{}) (syncError error) {
 		}
 
 		// Mark the InstallPlan as failed for a fatal Operator Group related error
-		ipFailError := fmt.Errorf("attenuated service account query failed - %v", err)
-		logger.Infof("InstallPlan failed: %v", ipFailError)
-
+		logger.Infof("attenuated service account query failed - %v", err)
+		ipFailError := fmt.Errorf("invalid operator group - %v", err)
 		now := o.now()
 		out := plan.DeepCopy()
 		out.Status.SetCondition(v1alpha1.ConditionFailed(v1alpha1.InstallPlanInstalled,
 			v1alpha1.InstallPlanReasonInstallCheckFailed, ipFailError.Error(), &now))
 		out.Status.Phase = v1alpha1.InstallPlanPhaseFailed
 
-		logger.Info("Transitioning InstallPlan to failed")
+		logger.Info("transitioning InstallPlan to failed")
 		if _, err := o.client.OperatorsV1alpha1().InstallPlans(plan.GetNamespace()).UpdateStatus(context.TODO(), out, metav1.UpdateOptions{}); err != nil {
 			updateErr := errors.New("error updating InstallPlan status: " + err.Error())
 			logger = logger.WithField("updateError", updateErr)
-			logger.Info("error transitioning InstallPlan to failed")
+			logger.Errorf("error transitioning InstallPlan to failed")
 
 			// retry sync with error to update InstallPlan status
 			syncError = fmt.Errorf("InstallPlan failed: %s and error updating InstallPlan status as failed: %s", ipFailError, updateErr)
--- a/pkg/controller/operators/catalog/operator_test.go
+++ b/pkg/controller/operators/catalog/operator_test.go
@@ -162,33 +162,33 @@ func TestSyncInstallPlanUnhappy(t *testing.T) {
 	)
 
 	tests := []struct {
-		testName       string
-		in             *v1alpha1.InstallPlan
-		err            error
-		expectedPhase  v1alpha1.InstallPlanPhase
-		operatorGroups []*operatorsv1.OperatorGroup
-		clientObjs     []runtime.Object
+		testName      string
+		err           error
+		in            *v1alpha1.InstallPlan
+		expectedPhase v1alpha1.InstallPlanPhase
+
+		clientObjs []runtime.Object
 	}{
 		{
 			testName:      "NoStatus",
-			in:            installPlan("p", namespace, v1alpha1.InstallPlanPhaseNone),
 			err:           nil,
 			expectedPhase: v1alpha1.InstallPlanPhaseNone,
+			in:            installPlan("p", namespace, v1alpha1.InstallPlanPhaseNone),
 		},
 		{
 			// This checks that an installplan is marked as failed when no operatorgroup is present
 			testName:      "HasSteps/NoOperatorGroup",
-			in:            ipWithSteps,
 			err:           nil,
 			expectedPhase: v1alpha1.InstallPlanPhaseFailed,
+			in:            ipWithSteps,
 		},
 		{
 			// This checks that an installplan is marked as failed when multiple operator groups are present for the same namespace
 			testName:      "HasSteps/TooManyOperatorGroups",
-			in:            ipWithSteps,
 			err:           nil,
 			expectedPhase: v1alpha1.InstallPlanPhaseFailed,
-			operatorGroups: []*operatorsv1.OperatorGroup{
+			in:            ipWithSteps,
+			clientObjs: []runtime.Object{
 				operatorGroup("og1", "sa", namespace,
 					&corev1.ObjectReference{
 						Kind:      "ServiceAccount",
@@ -206,10 +206,10 @@ func TestSyncInstallPlanUnhappy(t *testing.T) {
 		{
 			// This checks that an installplan is marked as failed when no service account is synced for the operator group, i.e the service account ref doesn't exist
 			testName:      "HasSteps/NonExistentServiceAccount",
-			in:            ipWithSteps,
 			err:           nil,
 			expectedPhase: v1alpha1.InstallPlanPhaseFailed,
-			operatorGroups: []*operatorsv1.OperatorGroup{
+			in:            ipWithSteps,
+			clientObjs: []runtime.Object{
 				operatorGroup("og", "sa1", namespace, nil),
 			},
 		},
@@ -221,9 +221,6 @@ func TestSyncInstallPlanUnhappy(t *testing.T) {
 			defer cancel()
 
 			tt.clientObjs = append(tt.clientObjs, tt.in)
-			for _, og := range tt.operatorGroups {
-				tt.clientObjs = append(tt.clientObjs, og)
-			}
 
 			op, err := NewFakeOperator(ctx, namespace, []string{namespace}, withClientObjs(tt.clientObjs...))
 			require.NoError(t, err)
--- a/pkg/controller/operators/catalog/operator_test.go
+++ b/pkg/controller/operators/catalog/operator_test.go
@@ -41,6 +41,7 @@ import (
 	apiregistrationv1 "k8s.io/kube-aggregator/pkg/apis/apiregistration/v1"
 	apiregistrationfake "k8s.io/kube-aggregator/pkg/client/clientset_generated/clientset/fake"
 
+	operatorsv1 "github.com/operator-framework/api/pkg/operators/v1"
 	"github.com/operator-framework/api/pkg/operators/v1alpha1"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/clientset/versioned/fake"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/informers/externalversions"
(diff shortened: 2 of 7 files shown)


hasbro17:
Looking into the e2e test failures and going to add some e2e tests myself. But the implementation should be okay for a preliminary review I think.

pkg/controller/operators/catalog/operator.go
```diff
@@ -1331,10 +1331,44 @@ func (o *Operator) syncInstallPlans(obj interface{}) (syncError error) {
 		return
 	}
 
+	if plan.Status.Phase == v1alpha1.InstallPlanPhaseFailed {
+		return
```
hasbro17, line 1335: This seemed like an obvious step to me but I'm probably missing something if it wasn't here before.

Is the InstallPlan sync ever expected to recover back from a `Failed` InstallPlan? I thought that was terminal and you have to start over with a new one.

pkg/controller/operators/catalog/operator.go
```diff
@@ -1331,10 +1331,44 @@ func (o *Operator) syncInstallPlans(obj interface{}) (syncError error) {
...
-		syncError = fmt.Errorf("attenuated service account query failed - %v", err)
+
+		// Retry sync if non-fatal error
+		if !scoped.IsOperatorGroupError(err) {
+			syncError = fmt.Errorf("attenuated service account query failed - %v", err)
+			return
+		}
+
+		// Mark the InstallPlan as failed for a fatal Operator Group related error
+		ipFailError := fmt.Errorf("attenuated service account query failed - %v", err)
```
hasbro17, line 1349: Right now, I'm propagating this error message to the InstallPlan status conditions. Wondering if this should be trimmed down a bit.
```Yaml
kind: InstallPlan
status:
  conditions:
    - lastTransitionTime: '2021-04-06T01:10:01Z'
      lastUpdateTime: '2021-04-06T01:10:01Z'
      message: >-
        attenuated service account query failed - no operator group found that
        is managing this namespace
      reason: InstallCheckFailed
      status: 'False'
      type: Installed
  phase: Failed
```

hasbro17: Updated the status condition message to be more relevant to the user:
```YAML
kind: InstallPlan
status:
  conditions:
    - lastTransitionTime: '2021-04-06T01:10:01Z'
      lastUpdateTime: '2021-04-06T01:10:01Z'
      message: >-
        invalid operator group - no operator group found that
        is managing this namespace
      reason: InstallCheckFailed
      status: 'False'
      type: Installed
  phase: Failed
```

pkg/controller/operators/catalog/operator.go
```diff
@@ -1423,6 +1444,20 @@ func (o *Operator) syncInstallPlans(obj interface{}) (syncError error) {
...
+	if owners := ownerutil.GetOwnersByKind(plan, v1alpha1.SubscriptionKind); len(owners) > 0 {
+		for _, owner := range owners {
+			logger.WithField("owner", owner).Debug("requeueing installplan owner")
+			if err := o.subQueueSet.Requeue(plan.GetNamespace(), owner.Name); err != nil {
+				logger.WithError(err).Warn("error requeuing installplan owner")
+			}
+		}
+		return
+	}
+	logger.Trace("no installplan owner subscriptions found to requeue")
```
timflannagan, line 1458: (the actual implementation seems reasonable to me)

Minor nit: I think we can also just reverse the conditional check to be `if owners := ...; len(owners) <= 0` and just move that log message and return early in that conditional, which allows us to move that for loop further to the right for better readability.

pkg/controller/operators/catalog/operator.go
```diff
@@ -1331,10 +1331,43 @@ func (o *Operator) syncInstallPlans(obj interface{}) (syncError error) {
...
+			syncError = fmt.Errorf("attenuated service account query failed - %v", err)
+			return
+		}
+
+		// Mark the InstallPlan as failed for a fatal Operator Group related error
+		logger.Infof("attenuated service account query failed - %v", err)
+		ipFailError := fmt.Errorf("invalid operator group - %v", err)
+		now := o.now()
+		out := plan.DeepCopy()
+		out.Status.SetCondition(v1alpha1.ConditionFailed(v1alpha1.InstallPlanInstalled,
```
benluddy, line 1353: This could result in a Complete -> Failed transition, which seems confusing. I think you'll either need to expand the values of Phase that are considered terminal above or somehow push more of this into transitionInstallPlanState to avoid that.

test/e2e/installplan_e2e_test.go
```diff
@@ -2708,13 +2708,96 @@ var _ = Describe("Install Plan", func() {
...
-		log(fmt.Sprintf("Install plan %s fetched with status %s", fetchedInstallPlan.GetName(), fetchedInstallPlan.Status.Phase))
+		log(fmt.Sprintf("Install plan %s fetched with phase %s", fetchedInstallPlan.GetName(), fetchedInstallPlan.Status.Phase))
+		log(fmt.Sprintf("Install plan %s fetched with conditions %+v", fetchedInstallPlan.GetName(), fetchedInstallPlan.Status.Conditions))
+	})
+
+	It("should fail an InstallPlan when multiple operatorgroups are present", func() {
+
+		log := func(s string) {
+			GinkgoT().Logf("%s: %s", time.Now().Format("15:04:05.9999"), s)
+		}
```
benluddy, line 2721: This should probably use https://github.com/operator-framework/operator-lifecycle-manager/blob/master/test/e2e/ctx/ctx.go#L49 so that we can tweak test logging in one place.

test/e2e/installplan_e2e_test.go
```diff
@@ -2708,13 +2708,96 @@ var _ = Describe("Install Plan", func() {
...
+
+		ns := &corev1.Namespace{}
+		ns.SetName(genName("ns-"))
+
+		c := newKubeClient()
+		crc := newCRClient()
+
+		// Create a namespace
+		ns, err := c.KubernetesInterface().CoreV1().Namespaces().Create(context.TODO(), ns, metav1.CreateOptions{})
+		require.NoError(GinkgoT(), err)
```
benluddy, line 2731: Could you change the assertions to use Gomega matchers instead of testify? The adjacent tests were partially converted to be run by Ginkgo, but they still contain some cruft.

test/e2e/installplan_e2e_test.go
```diff
@@ -2708,13 +2708,96 @@ var _ = Describe("Install Plan", func() {
...
+		subscriptionCleanup := createSubscriptionForCatalog(crc, ns.GetName(), subscriptionName, mainCatalogName, mainPackageName, stableChannel, "", operatorsv1alpha1.ApprovalAutomatic)
+		defer subscriptionCleanup()
+
+		subscription, err := fetchSubscription(crc, ns.GetName(), subscriptionName, subscriptionHasInstallPlanChecker)
+		require.NoError(GinkgoT(), err)
+		require.NotNil(GinkgoT(), subscription)
+
+		installPlanName := subscription.Status.InstallPlanRef.Name
+
+		fetchedInstallPlan, err := fetchInstallPlanWithNamespace(GinkgoT(), crc, installPlanName, ns.GetName(), buildInstallPlanPhaseCheckFunc(operatorsv1alpha1.InstallPlanPhaseFailed))
```
benluddy, line 2797: Instead of creating a CatalogSource and Subscription, this test could be more specific (and faster) if it were to create the InstallPlan directly.

test/e2e/installplan_e2e_test.go
```diff
@@ -3122,6 +3217,18 @@ var _ = Describe("Install Plan", func() {
...
 
+		// Wait for the OperatorGroup to be synced and have a status.ServiceAccountRef
+		// before moving on. Otherwise the catalog operator treats it as an invalid OperatorGroup
+		// and fails the InstallPlan
+		Eventually(func() (bool, error) {
+			outOG, err := crc.OperatorsV1().OperatorGroups(ns.GetName()).Get(context.TODO(), og.Name, metav1.GetOptions{})
+			if err != nil {
+				return false, err
+			}
+			fmt.Fprintf(GinkgoWriter, "[DEBUG] Operator Group Status: %+v\n", outOG.Status)
```
benluddy, line 3228: Similar comment here, the test context has a Logf interface that does approximately the same thing.

test/e2e/installplan_e2e_test.go
```diff
@@ -3122,6 +3217,18 @@ var _ = Describe("Install Plan", func() {
...
+			return outOG.Status.ServiceAccountRef != nil, nil
+		}).Should(BeTrue())
```
benluddy, line 3230:
```suggestion
		Eventually(func() (*corev1.ObjectReference, error) {
			outOG, err := crc.OperatorsV1().OperatorGroups(ns.GetName()).Get(context.TODO(), og.Name, metav1.GetOptions{})
			if err != nil {
				return nil, err
			}
			return outOG.Status.ServiceAccountRef, nil
		}).ShouldNot(BeNil())
```

It's nice to use the more specific Gomega matcher for this kind of thing instead of converting to a boolean first -- then when the test fails the report will have better information about what went wrong.

pkg/controller/operators/catalog/operator_test.go
```diff
@@ -142,38 +143,75 @@ func TestTransitionInstallPlan(t *testing.T) {
...
 	tests := []struct {
-		testName string
-		in       *v1alpha1.InstallPlan
-		err      error
+		testName       string
+		in             *v1alpha1.InstallPlan
+		err            error
+		expectedPhase  v1alpha1.InstallPlanPhase
+		operatorGroups []*operatorsv1.OperatorGroup
+		clientObjs     []runtime.Object
```
benluddy, line 170: I think `*OperatorGroup` implements `runtime.Object`. Can these fields be merged?

pkg/controller/operators/catalog/operator.go
hasbro17: I think I would rather consider `Complete` as terminal as well, similar to `Failed` and just end the sync early.
For all intents and purposes `Complete` is treated as terminal anyway right now by the InstallPlan sync, or more specifically [`transitionInstallPlanState()`](https://github.com/operator-framework/operator-lifecycle-manager/blob/bc37ac4e271272a8f8fbac4d1b08157b4883ec9f/pkg/controller/operators/catalog/operator.go#L1495-L1497).
That's more clear and separate to me than trying to handle all of this in `transitionInstallPlanState()` and passing it the results of the NamespaceQuerier.

Although please correct me if there's ever a case where we need to reconcile something more after the InstallPlan hits `Complete`.

pkg/controller/operators/catalog/operator.go
hasbro17: Yeah I just moved this block of code as is to reuse it, but I can refactor it to reduce nesting as well.

test/e2e/installplan_e2e_test.go
hasbro17: This coming over from an existing test but I can update these in my tests.
I'll do a follow up after this PR to make sure we're consistently using the same test logging util in the entire e2e suite.

pkg/controller/operators/catalog/operator.go
benluddy: As far as I know the InstallPlan is effectively a historical record once it has reached either Failed or Complete. Returning early in both cases makes sense to me (feels related to https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88).

kuiwang02:
@hasbro17 for "An incorrect or non-existent ServiceAccount name specified on the OperatorGroup", the error message seems not correct. we expect "please make sure the service account exists...", but it is "invalid operator group - no operator group found that is managing this namespace". here is the example
```console
[root@preserve-olm-env OCP-40958]# cat teiidcatsrc.yaml
kind: CatalogSource
metadata:
  name: teiid
  namespace: default
spec:
  displayName: "teiid Operators"
  image: quay.io/kuiwang/teiid-index:1906056
  publisher: QE
  sourceType: grpc
[root@preserve-olm-env OCP-40958]# oc apply -f teiidcatsrc.yaml
catalogsource.operators.coreos.com/teiid created
[root@preserve-olm-env OCP-40958]# oc get pod
NAME          READY   STATUS    RESTARTS   AGE
teiid-t2rtq   0/1     Running   0          7s
[root@preserve-olm-env OCP-40958]# cat ogwrongsa.yaml
kind: OperatorGroup
metadata:
  name: ogwrongsa
  namespace: default
spec:
  serviceAccountName: foo
  targetNamespaces:
  - default
[root@preserve-olm-env OCP-40958]# oc apply -f ogwrongsa.yaml
operatorgroup.operators.coreos.com/ogwrongsa created
[root@preserve-olm-env OCP-40958]# oc get og ogwrongsa -o yaml
kind: OperatorGroup
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"operators.coreos.com/v1","kind":"OperatorGroup","metadata":{"annotations":{},"name":"ogwrongsa","namespace":"default"},"spec":{"serviceAccountName":"foo","targetNamespaces":["default"]}}
  creationTimestamp: "2021-04-08T03:09:14Z"
  generation: 1
  ...
  name: ogwrongsa
  namespace: default
  resourceVersion: "45889"
  uid: a5420f23-d7d9-46a0-95ca-97586300de68
spec:
  serviceAccountName: foo
  targetNamespaces:
  - default
[root@preserve-olm-env OCP-40958]# cat teiidsub.yaml
kind: Subscription
metadata:
  name: teiid
  namespace: default
spec:
  source: teiid
  sourceNamespace: default

  channel: beta
  name: teiid
  startingCSV: teiid.v0.4.0
[root@preserve-olm-env OCP-40958]# oc apply -f teiidsub.yaml
subscription.operators.coreos.com/teiid created
[root@preserve-olm-env OCP-40958]# oc get sub
NAME    PACKAGE   SOURCE   CHANNEL
teiid   teiid     teiid    beta
[root@preserve-olm-env OCP-40958]# oc get ip
NAME            CSV            APPROVAL    APPROVED
install-27nd7   teiid.v0.4.0   Automatic   true
[root@preserve-olm-env OCP-40958]# oc get ip install-27nd7 -o=jsonpath='{.status.conditions}'|jq .
[
  {
    "lastTransitionTime": "2021-04-08T03:10:01Z",
    "lastUpdateTime": "2021-04-08T03:10:01Z",
    "message": "invalid operator group - no operator group found that is managing this namespace",
    "reason": "InstallCheckFailed",
    "status": "False",
    "type": "Installed"
  }
]

[root@preserve-olm-env OCP-40958]# oc get sa -A|grep foo
```

pkg/controller/operators/catalog/operator_test.go
hasbro17: Done

test/e2e/installplan_e2e_test.go
hasbro17: Yeah that was much faster and cleaner, but I have to do the subscription sync step myself to populate the InstallPlan status  as well.

pkg/controller/operators/catalog/operator.go
hasbro17: TIL the official term for "return early to reduce nesting".

hasbro17:
@kuiwang02 That's actually fine I think. Specifying a non-existent ServiceAccount would mean the OperatorGroup never gets synced properly to have its `status.Namespaces` populated. Which means that the OperatorGroup is not managing that namespace, and so the correct error would be to see `invalid operator group - no operator group found that is managing this namespace`

https://github.com/operator-framework/operator-lifecycle-manager/blob/5cd78221443d9f67a06b717d26f9211f5538ab2d/pkg/controller/operators/olm/operatorgroup.go#L65-L69

https://github.com/operator-framework/operator-lifecycle-manager/blob/5cd78221443d9f67a06b717d26f9211f5538ab2d/pkg/lib/scoped/syncer.go#L74-L79

And that's what you can see from your OperatorGroup as well, that it has no status:

```console
[root@preserve-olm-env OCP-40958]# oc get og ogwrongsa -o yaml
kind: OperatorGroup
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"operators.coreos.com/v1","kind":"OperatorGroup","metadata":{"annotations":{},"name":"ogwrongsa","namespace":"default"},"spec":{"serviceAccountName":"foo","targetNamespaces":["default"]}}
  creationTimestamp: "2021-04-08T03:09:14Z"
  generation: 1
  ...
  name: ogwrongsa
  namespace: default
  resourceVersion: "45889"
  uid: a5420f23-d7d9-46a0-95ca-97586300de68
spec:
  serviceAccountName: foo
  targetNamespaces:
  - default
```

The proper way to test the case of the missing ServiceAccountRef would be to create an OperatorGroup with the `status.namespaces` populated but `status.serviceAccountRef` missing, e.g:
```YAML
kind: OperatorGroup
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"operators.coreos.com/v1","kind":"OperatorGroup","metadata":{"annotations":{},"name":"ogwrongsa","namespace":"default"},"spec":{"serviceAccountName":"foo","targetNamespaces":["default"]}}
  creationTimestamp: "2021-04-08T03:09:14Z"
  generation: 1
  ...
  name: ogwrongsa
  namespace: default
  resourceVersion: "45889"
  uid: a5420f23-d7d9-46a0-95ca-97586300de68
spec:
  serviceAccountName: foo
  targetNamespaces:
  - default
status:
  lastUpdated: "2021-04-06T01:00:44Z"
  namespaces:
  - default
```

I've just added an e2e test for that as well.

kuiwang02:
> @kuiwang02 That's actually fine I think. Specifying a non-existent ServiceAccount would mean the OperatorGroup never gets synced properly to have its `status.Namespaces` populated. Which means that the OperatorGroup is not managing that namespace, and so the correct error would be to see `invalid operator group - no operator group found that is managing this namespace`
> 
> https://github.com/operator-framework/operator-lifecycle-manager/blob/5cd78221443d9f67a06b717d26f9211f5538ab2d/pkg/controller/operators/olm/operatorgroup.go#L65-L69
> 
> https://github.com/operator-framework/operator-lifecycle-manager/blob/5cd78221443d9f67a06b717d26f9211f5538ab2d/pkg/lib/scoped/syncer.go#L74-L79
> 
> And that's what you can see from your OperatorGroup as well, that it has no status:
> 
> ```
> [root@preserve-olm-env OCP-40958]# oc get og ogwrongsa -o yaml
> apiVersion: operators.coreos.com/v1
> kind: OperatorGroup
> metadata:
>   annotations:
>     kubectl.kubernetes.io/last-applied-configuration: |
>       {"apiVersion":"operators.coreos.com/v1","kind":"OperatorGroup","metadata":{"annotations":{},"name":"ogwrongsa","namespace":"default"},"spec":{"serviceAccountName":"foo","targetNamespaces":["default"]}}
>   creationTimestamp: "2021-04-08T03:09:14Z"
>   generation: 1
>   ...
>   name: ogwrongsa
>   namespace: default
>   resourceVersion: "45889"
>   uid: a5420f23-d7d9-46a0-95ca-97586300de68
> spec:
>   serviceAccountName: foo
>   targetNamespaces:
>   - default
> ```
> 
> The proper way to test the case of the missing ServiceAccountRef would be to create an OperatorGroup with the `status.namespaces` populated but `status.serviceAccountRef` missing, e.g:
> 
> ```yaml
> apiVersion: operators.coreos.com/v1
> kind: OperatorGroup
> metadata:
>   annotations:
>     kubectl.kubernetes.io/last-applied-configuration: |
>       {"apiVersion":"operators.coreos.com/v1","kind":"OperatorGroup","metadata":{"annotations":{},"name":"ogwrongsa","namespace":"default"},"spec":{"serviceAccountName":"foo","targetNamespaces":["default"]}}
>   creationTimestamp: "2021-04-08T03:09:14Z"
>   generation: 1
>   ...
>   name: ogwrongsa
>   namespace: default
>   resourceVersion: "45889"
>   uid: a5420f23-d7d9-46a0-95ca-97586300de68
> spec:
>   serviceAccountName: foo
>   targetNamespaces:
>   - default
> status:
>   lastUpdated: "2021-04-06T01:00:44Z"
>   namespaces:
>   - default
> ```
> 
> I've just added an e2e test for that as well.

@hasbro17 you are right. thanks.

test/e2e/installplan_e2e_test.go
```diff
@@ -2728,40 +2696,13 @@ var _ = Describe("Install Plan", func() {
 
 		// Create a namespace
 		ns, err := c.KubernetesInterface().CoreV1().Namespaces().Create(context.TODO(), ns, metav1.CreateOptions{})
+		Expect(err).NotTo(HaveOccurred())
 		require.NoError(GinkgoT(), err)
```
hasbro17, line 2700: Missed this one. Will remove it in the next update.

pkg/controller/operators/catalog/operator.go
```diff
@@ -1331,10 +1331,44 @@ func (o *Operator) syncInstallPlans(obj interface{}) (syncError error) {
 		return
 	}
 
+	// Complete and Failed are terminal phases
+	if plan.Status.Phase == v1alpha1.InstallPlanPhaseFailed || plan.Status.Phase == v1alpha1.InstallPlanPhaseComplete {
```
exdx, line 1335: This makes sense logically but looking at https://issues.redhat.com/browse/OLM-2114 coming in as well -- if there is a transient error in `ExecutePlan()` we will not mark the installplan as `Failed` but keep it in the `Installing` phase for a certain period of time right?

The new behavior would be that in the `ExecutePlan()` function that runs the installplan steps any error returned from that function effectively causes installplan reconciliation to stop. We do not have that behavior today (the installplan stays in the `Failed` state but continues to be reconciled). 

The reason I bring this up is because of https://github.com/operator-framework/operator-lifecycle-manager/blob/master/doc/design/scoped-operator-install.md#when-operator-install-fails
I believe we have a feature where if the ServiceAccount used to install the operator lacks certain permissions, we encourage the user to modify the SA to get to the permission level they need and OLM will automatically requeue up the installplan. This changes that behavior. cc @ecordell if I'm mistaken.

pkg/controller/operators/catalog/operator.go
```diff
@@ -1423,6 +1445,23 @@ func (o *Operator) syncInstallPlans(obj interface{}) (syncError error) {
...
+func (o *Operator) requeueSubscriptionForInstallPlan(plan *v1alpha1.InstallPlan, logger *logrus.Entry) {
+	// Notify subscription loop of installplan changes
+	owners := ownerutil.GetOwnersByKind(plan, v1alpha1.SubscriptionKind)
+
+	if len(owners) == 0 {
+		logger.Trace("no installplan owner subscriptions found to requeue")
+		return
+	}
+
+	for _, owner := range owners {
```
exdx, line 1457: can an installplan ever be owned by more than 1 subscription -- maybe we want to explicitly check `len(owners)`?

test/e2e/installplan_e2e_test.go
hasbro17: Forgot I had two of these blocks. Coming in the next update.

pkg/controller/operators/catalog/operator.go
benluddy: > the installplan stays in the Failed state but continues to be reconciled

If it's in Failed, then when it reconciles again it hits the default case in `transitionInstallPlanState` and remains Failed. The in-progress retry feature is going to prevent the transition from Installing -> Failed and requeue as long as a timeout hasn't been reached.

pkg/controller/operators/catalog/operator.go
benluddy: It can (https://github.com/operator-framework/operator-lifecycle-manager/blob/e5a160aa2a69b4afe3a8da0add98b4083975096c/pkg/controller/operators/catalog/operator.go#L1143), since all Subscriptions in a namespace are resolved together.

hasbro17: This is coming over from an existing block in the sync that I didn't really modify.

But even if we have >1 owners for an InstallPlan, it's okay to requeue them all right? Since they need to update their status to reflect the change in the InstallPlan phase.

So @exdx not sure I understood what you meant by explicitly checking `len(owners)`.

hasbro17:
Fixed the failing e2e test that had an incorrect gomega assertion in the previous commit.

test/e2e/installplan_e2e_test.go
```diff
@@ -2664,57 +2660,141 @@ var _ = Describe("Install Plan", func() {
...
+		// The status gets ignored on create so we need to update it else the InstallPlan sync ignores
+		// InstallPlans without any steps or bundle lookups
+		outIP.Status = ip.Status
+		_, err = crc.OperatorsV1alpha1().InstallPlans(ns.GetName()).UpdateStatus(context.TODO(), outIP, metav1.UpdateOptions{})
+		Expect(err).NotTo(HaveOccurred())
+
+		fetchedInstallPlan, err := fetchInstallPlanWithNamespace(GinkgoT(), crc, installPlanName, ns.GetName(), buildInstallPlanPhaseCheckFunc(operatorsv1alpha1.InstallPlanPhaseFailed))
+		Expect(err).NotTo(HaveOccurred())
+		ctx.Ctx().Logf(fmt.Sprintf("Install plan %s fetched with phase %s", fetchedInstallPlan.GetName(), fetchedInstallPlan.Status.Phase))
+		ctx.Ctx().Logf(fmt.Sprintf("Install plan %s fetched with conditions %+v", fetchedInstallPlan.GetName(), fetchedInstallPlan.Status.Conditions))
```
benluddy, line 2686: It looks like `fetchedInstallPlan` could be nil and cause a panic here.

test/e2e/installplan_e2e_test.go
```diff
@@ -2664,57 +2660,141 @@ var _ = Describe("Install Plan", func() {
...
+		c := newKubeClient()
+		crc := newCRClient()
+
+		// Create a namespace
+		ns, err := c.KubernetesInterface().CoreV1().Namespaces().Create(context.TODO(), ns, metav1.CreateOptions{})
+		Expect(err).NotTo(HaveOccurred())
+		deleteOpts := &metav1.DeleteOptions{}
 		defer func() {
-			require.NoError(GinkgoT(), crc.OperatorsV1alpha1().Subscriptions(ns.GetName()).DeleteCollection(context.TODO(), metav1.DeleteOptions{}, metav1.ListOptions{}))
+			require.NoError(GinkgoT(), c.KubernetesInterface().CoreV1().Namespaces().Delete(context.TODO(), ns.GetName(), *deleteOpts))
```
benluddy, line 2702: There are a few remaining testify + GinkgoT holdouts from the adjacent tests, but it should be fine to follow up on them later as part of ongoing test consistency improvements.

pkg/controller/operators/catalog/operator.go
exdx: Ok. I seem to recall the scenario I described above as desirable (if the SA isn't quite right it can be modified and the InstallPlan can eventually succeed from a conversation with Evan). But it does seem like an anti-pattern in general so I am ok moving ahead as-is.

pkg/controller/operators/catalog/operator.go
exdx: That sounds fine. I guess I am conceptually confused by the case where an installplan would have multiple subscriptions associated with it -- does it introduce any weird edge cases? But it looks like our code has no issues with that. 

As @hasbro17 it's fine to requeue them all since they would all need to be updated with the new installplan status

PR #2081, opened 2021-04-06
Block OpenShift upgrades while incompatible operators are installed

njhale:

Set OLM's `ClusterOperator` resource to Upgradeable=false when any OLM managed operator on the cluster has a maxOpenShiftVersion property _less than_ the next minor OCP release.

Implements [OLM-2074](https://issues.redhat.com/browse/OLM-2074).

I decided to implement this as a new controller so we could more easily interleave new and preexisting ClusterOperator behavior.


pkg/controller/operators/openshift/clusteroperator_controller.go
```diff
@@ -0,0 +1,247 @@
+package openshift
```
exdx, line 1: Taking a brief look and my thought is, should this just be a downstream PR and not merged into this repository? Or is the plan to merge, sync, and then take this out. This is the most openshift specific code I've ever seen proposed to be checked into this repo. 

I know this is a bit annoying for you as the author, as we have to iron the split out, but I'm just wondering.

pkg/controller/operators/openshift/helpers.go
```diff
@@ -0,0 +1,234 @@
...
+	if !ok {
+		return nil, nil
+	}
+
+	properties, err := projection.PropertyListFromPropertiesAnnotation(annotation)
+	if err != nil {
+		return nil, err
+	}
+
+	// Take the highest semver if there's more than one max version specified
```
benluddy, line 153: Would it be a safer choice to go with the lowest version?

pkg/controller/operators/openshift/helpers.go
```diff
@@ -0,0 +1,234 @@
...
+	return max, nil
+}
+
+func olmOperatorRelatedObjects(ctx context.Context, cli client.Client, namespace string) ([]configv1.ObjectReference, error) {
+	csvList := &operatorsv1alpha1.ClusterServiceVersionList{}
+	if err := cli.List(ctx, csvList, client.InNamespace(namespace)); err != nil {
+		return nil, err
+	}
+
+	// TODO: Is there a better way to filter (server-side maybe)?
```
benluddy, line 216: There should be an `olm.copiedFrom` label on copied CSVs. I've had this thought too, and I'm not sure why there are no examples of a CSV label selector referencing it.

pkg/controller/operators/openshift/clusteroperator_controller.go
```diff
@@ -0,0 +1,247 @@
...
+	return nil
+}
+
+func (r *ClusterOperatorReconciler) setDegraded(_ context.Context, co *ClusterOperator) error {
+	desired := &configv1.ClusterOperatorStatusCondition{
+		Type:               configv1.OperatorDegraded,
+		LastTransitionTime: r.Now(),
+	}
+
+	if r.syncTracker.SuccessfulSyncs() > 0 && reflect.DeepEqual(co.Status.Versions, r.TargetVersions) {
```
benluddy, line 190: This seems fine, but do we need to lean on reflection? Since OperandVersion is `struct{Name,Version string}` it would be a tiny loop.

pkg/controller/operators/openshift/clusteroperator_controller.go
njhale: it needs to be backported to release-4.6, which we're not capable of doing completely downstream-only yet.

pkg/controller/operators/openshift/helpers.go
njhale: I was going at this from the perspective that folks would always make a conscious decision to add the property, but over several releases may forget to remove older ones.

pkg/controller/operators/openshift/clusteroperator_controller.go
njhale: we can do an `ElementsMatch` type thing easily enough.

pkg/controller/operators/openshift/helpers.go
```diff
@@ -0,0 +1,234 @@
...
+		return false
+	}
+
+	return a.Type == b.Type && a.Status == b.Status && a.Message == b.Message && a.Reason == b.Reason
+}
+
+type skews []skew
+
+func (s skews) String() string {
+	// TODO: Use a string builder.
```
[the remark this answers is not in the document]
njhale: that works too.

pkg/controller/operators/openshift/clusteroperator_controller.go
njhale: I went ahead and made this change.

pkg/controller/operators/openshift/helpers.go
njhale: I added a label selector.

pkg/controller/operators/openshift/helpers.go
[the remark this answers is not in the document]
njhale: done

pkg/controller/operators/openshift/clusteroperator_controller.go
```diff
@@ -0,0 +1,247 @@
...
+
+	// AddToScheme adds all types necessary for the controller to operate.
+	AddToScheme = localSchemeBuilder.AddToScheme
+)
+
+type ClusterOperatorReconciler struct {
+	*ReconcilerConfig
+
+	delayRequeue reconcile.Result
+	mutate       MutateFunc
```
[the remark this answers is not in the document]
njhale: done

pkg/controller/operators/openshift/helpers.go
```diff
@@ -0,0 +1,275 @@
...
+		}
+
+		max = &version
+	}
+
+	// Return an error if THE max version has a duplicate (i.e. equivalent version)
+	// Note: This may not be a problem since there should be no difference as far as blocking upgrades is concerned.
+	// This is more for clear status messages.
+	for _, dup := range dups {
+		if max.EQ(dup) && max.String() != dup.String() { // "1.0.0" vs "1.0.0" is fine, but not "1.0.0" vs "1.0.0+1"
```
exdx, line 223: Should this be a validation check at the bundle build time instead of at runtime? Like if I'm a cluster admin running an update, and I see that there is an operator that violates this check (it has a weird max ocp version semver string) how could I recover? 

There's no great way around this, but I'm wondering if we put the admin in an untenable position here -- go delete the operator or update it to a version with a better formatted maxOCP property?

exdx (commented):
one nit
/lgtm
/hold

pkg/controller/operators/openshift/helpers.go
exdx: Ok, as long as we surface it somehow.. I suppose it's better being safe than sorry in this type of scenario

PR #2086, opened 2021-04-08
Give more precise reasons for Subscriptions with no candidates.

benluddy:
A Subscription specifies certain criteria for selecting operators from
catalogs. The granularity of the Subscription criteria is, from least
to most specific: catalog, package, channel, startingCSV (if
applicable). If a user creates a Subscription and its criteria does
not match a single operator, the failure message now indicates the
highest level of granularity that excludes all available
operators. For example, if a Subscription specifies a package name
that doesn't exist in the referenced, catalog, the specified package
name will appear in the failure text.

### DIFF
--- a/pkg/controller/registry/resolver/resolver.go
+++ b/pkg/controller/registry/resolver/resolver.go
@@ -67,16 +67,6 @@ func (r *SatResolver) SolveOperators(namespaces []string, csvs []*v1alpha1.Clust
 
 	// build constraints for each Subscription
 	for _, sub := range subs {
-		pkg := sub.Spec.Package
-		catalog := registry.CatalogKey{
-			Name:      sub.Spec.CatalogSource,
-			Namespace: sub.Spec.CatalogSourceNamespace,
-		}
-		predicates := []OperatorPredicate{WithPackage(pkg)}
-		if sub.Spec.Channel != "" {
-			predicates = append(predicates, WithChannel(sub.Spec.Channel))
-		}
-
 		// find the currently installed operator (if it exists)
 		var current *Operator
 		for _, csv := range csvs {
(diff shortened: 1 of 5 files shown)


pkg/controller/registry/resolver/resolver.go
```diff
@@ -183,11 +163,53 @@ func (r *SatResolver) SolveOperators(namespaces []string, csvs []*v1alpha1.Clust
...
+			CountingPredicate(WithChannel(sub.Spec.Channel), &nch),
+			CountingPredicate(csvPredicate, &ncsv),
+		))
+		bundles = namespacedCache.Catalog(catalog).Find(cachePredicates...)
+
+		var si solver.Installable
+		switch {
+		case nall == 0:
+			si = NewInvalidSubscriptionInstallable(sub.GetName(), fmt.Sprintf("no operators found from catalog %s in namespace %s referenced by subscription %s", sub.Spec.CatalogSource, sub.Spec.CatalogSourceNamespace, sub.GetName()))
+		case npkg == 0:
```
exdx, line 200: nit: is this actually `package %s not found in the catalog %s referenced by subscription %s`?

If we're saying the package was not found versus no operator (bundles) were found in that package. Because they are two different types of errors, empty package versus no package.

pkg/controller/registry/resolver/resolver.go
```diff
@@ -183,11 +163,53 @@ func (r *SatResolver) SolveOperators(namespaces []string, csvs []*v1alpha1.Clust
...
+			si = NewInvalidSubscriptionInstallable(sub.GetName(), fmt.Sprintf("no operators found in package %s in the catalog referenced by subscription %s", sub.Spec.Package, sub.GetName()))
+		case nch == 0:
```
exdx, line 202: same here, is it no channel found versus no operators found in channel? 

`channel %s not found in package %s in the catalog referenced by subscription %s`

pkg/controller/registry/resolver/resolver.go
benluddy: Good observation -- we only know that a package (or channel) exists at this point if there is a bundle claiming to be in that package, so the wording is intentional.

PR #2090, opened 2021-04-08
Add time-limited retries to InstallPlan execution.

benluddy:
In order to recover from transient errors while applying step
resources, errors at this stage will be retried until success or
timeout, and .status.message will indicate that retries are
occurring. Configuring a timeout of zero is equivalent to disabling
the retry behavior.


benluddy:
Opened https://github.com/operator-framework/operator-lifecycle-manager/issues/2091 for what looks like a test flake.

pkg/controller/operators/catalog/operator.go
```diff
@@ -1424,49 +1426,52 @@ func (o *Operator) syncInstallPlans(obj interface{}) (syncError error) {
...
 			out.Status.Phase = v1alpha1.InstallPlanPhaseInstalling
+			out.Status.Message = ""
+			log.Debugf("approved, setting to %s", out.Status.Phase)
 		} else {
 			log.Debug("not approved, skipping sync")
 		}
 		return out, nil
 
 	case v1alpha1.InstallPlanPhaseInstalling:
+		if out.Status.StartTime == nil {
```
hasbro17, line 1449: Can we use [IsZero()](https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/time.go#L58-L64) which checks for nil and zero.
Just thinking if StartTime could ever be 0, which would fail this check?
```suggestion
		if out.Status.StartTime.IsZero() {
```

hasbro17: Or I guess we're considering if this field is ever unset which is not quite the same as 0.

benluddy: This should never be modified after it's set the first time, even if it was set to zero. I can image two general scenarios where it might be zero: an admin manually patches it or the system clock is wrong. The result would be that retries wouldn't happen. So it'd be inconvenient, but assigning a new value would either be fighting with an admin or relying on the same broken clock.

cmd/catalog/main.go
```diff
@@ -69,6 +69,8 @@ var (
 
 	profiling = flag.Bool(
 		"profiling", false, "serve profiling data (on port 8080)")
+
+	installPlanTimeout = flag.Duration("install-plan-retry-timeout", 1*time.Minute, "time since first attempt at which plan execution errors are considered fatal")
```
exdx, line 73: should the default live as an exported constant somewhere like the default resync interval of 15 minutes in the `queueinformer` package?

pkg/controller/operators/catalog/operator.go
```diff
@@ -1424,49 +1426,52 @@ func (o *Operator) syncInstallPlans(obj interface{}) (syncError error) {
...
+			out.Status.StartTime = &now
+		}
 		log.Debug("attempting to install")
 		if err := transitioner.ExecutePlan(out); err != nil {
-			out.Status.SetCondition(v1alpha1.ConditionFailed(v1alpha1.InstallPlanInstalled,
-				v1alpha1.InstallPlanReasonComponentFailed, err.Error(), &now))
-			out.Status.Phase = v1alpha1.InstallPlanPhaseFailed
+			if now.Sub(out.Status.StartTime.Time) >= timeout {
+				out.Status.SetCondition(v1alpha1.ConditionFailed(v1alpha1.InstallPlanInstalled,
+					v1alpha1.InstallPlanReasonComponentFailed, err.Error(), &now))
```
exdx, line 1456: Do we want to mention that the error failed even with retries in the status?

For example, say installplan failed due to a transient connection error with the api-server (that persisted for some amount of time). Is it more helpful to say "installplan failed due to connection error with api-server: timeout <interval>" versus just "installplan failed due to connection error"? 

This leaks to the user that we retried the plan, but maybe that's what we would want in cases like this?

pkg/controller/operators/catalog/operator.go
```diff
@@ -1424,49 +1426,52 @@ func (o *Operator) syncInstallPlans(obj interface{}) (syncError error) {
...
 	case v1alpha1.InstallPlanPhaseInstalling:
+		if out.Status.StartTime == nil {
+			out.Status.StartTime = &now
+		}
 		log.Debug("attempting to install")
 		if err := transitioner.ExecutePlan(out); err != nil {
-			out.Status.SetCondition(v1alpha1.ConditionFailed(v1alpha1.InstallPlanInstalled,
-				v1alpha1.InstallPlanReasonComponentFailed, err.Error(), &now))
-			out.Status.Phase = v1alpha1.InstallPlanPhaseFailed
+			if now.Sub(out.Status.StartTime.Time) >= timeout {
```
exdx, line 1454: I know we are extremely pressed on time here but I envisioned that we would actually check the error type as well. For example if it's a 409 or something, we certainly retry. If it's "a CSV with the same name already exists" then we don't. But it's not worth trying to implement some kind of retriable error function that returns a bool?

exdx (commented):
Nice! Just some general feedback

cmd/catalog/main.go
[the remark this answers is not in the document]
benluddy: It seemed reasonable to me to keep it fairly short until we can identify errors that are definitely unrecoverable.

benluddy: I don't think any calling code needs to depend on it. IMO the fact that there is a default is a concern of the CLI at this point.

pkg/controller/operators/catalog/operator.go
benluddy: It actually does do this!

pkg/controller/operators/catalog/operator.go
benluddy: I figured we would do the opposite, since plan execution itself is supposed to be idempotent, and eventually tag specific error cases that definitely won't resolve with retries. But I think that's a future enhancement, since we don't have a complete list.

cmd/catalog/main.go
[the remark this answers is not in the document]
exdx: Right, that makes sense

pkg/controller/operators/catalog/operator.go
exdx: sweet

pkg/controller/operators/catalog/operator.go
exdx: It's definitely something for the future, yes. I see what you mean about doing the opposite. Instead of catching errors that are requauable (which can be large in number) track those errors which are definitely terminal and fail-fast on those. Makes sense. 

At my last company we started trying to build something like an `IsRequeuable(error) bool` type thing, not sure it really went anywhere

horis233:
@benluddy Do we have any plan to cherry this PR to previous versions?

benluddy:
> @benluddy Do we have any plan to cherry this PR to previous versions?

No, this depends on the new `startTime` field in the InstallPlan API, so it won't be backported. If you're an OpenShift user, there is https://bugzilla.redhat.com/show_bug.cgi?id=1923111 which tracks creating a limited backportable patch for the same problem.

horis233:
@benluddy Thank you for the information!

PR #2093, opened 2021-04-09
Fail InstallPlan on bundle unpack timeout

hasbro17:
The InstallPlan sync can stay stalled on the Installing phase if the bundle cannot be successfully unpacked.
Adding a configurable timeout for the duration of the bundle unpack Job helps identify if an unpack Job is
stalled, and the InstallPlan is then transitioned to Failed with the unpack Job's failure condition
propagated to the InstallPlan condition.
- InstallPlan will fail after the unpack job's `ActiveDeadlineSeconds`
- InstallPlan will fail after the unpack Job's pods exit in error/crash more than the [`BackoffLimit`](https://kubernetes.io/docs/concepts/workloads/controllers/job/#pod-backoff-failure-policy), currently 3
- For a non-existent image the InstallPlan stays in phase `Installing` but the `BundleLookupPending` condition is update with the reason for why the unpack Job's pods are in a pending state. This shows the `ErrImagePull`.

Example of Failed InstallPlan due to a bundle unpack timeout:
```YAML
kind: InstallPlan
status:
  conditions:
    - lastTransitionTime: '2021-04-06T01:10:01Z'
      lastUpdateTime: '2021-04-06T01:10:01Z'
      message: >-
        Bundle extract Job failed with Reason: DeadlineExceeded, 
        and Message: Job was active longer than specified deadline
      reason: InstallCheckFailed
      status: 'False'
      type: Installed
  phase: Failed
```
Example of Failed InstallPlan due to an invalid bundle causing repeated pod failures:
```YAML
kind: InstallPlan
status:
  bundleLookups:
    - lastTransitionTime: "2021-04-21T20:19:27Z"
      message: Job has reached the specified backoff limit
      reason: BackoffLimitExceeded
      status: "True"
      type: BundleLookupFailed
    identifier: foobar.v0.0.1
    path: alpine:latest
  conditions:
  - lastTransitionTime: "2021-04-21T20:19:27Z"
    lastUpdateTime: "2021-04-21T20:19:27Z"
    message: 'Bundle unpacking failed. Reason: BackoffLimitExceeded, and Message:
      Job has reached the specified backoff limit'
    reason: InstallCheckFailed
    status: "False"
    type: Installed
  phase: Failed
```

Example of an InstallPlan that is stalled on a non-existent bundle image lookup:
```YAML
kind: InstallPlan
status:
  bundleLookups:
    conditions:
    - lastTransitionTime: "2021-04-09T00:44:55Z"
      message: "unpack job not completed: Unpack pod(default/8c3e265337115bdeff1ab7a092b9f531bffc580830a021ffe04d5214eenb2xr)
        container(pull) is pending. Reason: ErrImagePull, Message: rpc error: code
        = Unknown desc = failed to pull and unpack image \"quay.io/foo/bar:latest\":
        failed to resolve reference \"quay.io/foo/bar:latest\": unexpected status
        code [manifests latest]: 401 UNAUTHORIZED \n"
      reason: JobIncomplete
      status: "True"
      type: BundleLookupPending
    path: quay.io/foo/bar:latest
``` 

**Reviewer Checklist**


cmd/catalog/main.go
```diff
@@ -69,6 +69,9 @@ var (
 
 	profiling = flag.Bool(
 		"profiling", false, "serve profiling data (on port 8080)")
+
+	// TODO(haseeb): Change to a more realistic timeout for bundle unpack jobs
+	bundleUnpackTimeout = flag.Duration("bundle-unpack-timeout", 60*time.Second, "The time duration after which the bundle unpack Job for an installplan will be aborted and the installplan will be Failed. 0 is considered as having no timeout.")
```
hasbro17, line 74: @benluddy Any idea what a sane default might be here?

Xia-Zhao-rh:
@hasbro17 Would you please help to rebase it?

hasbro17:
@Xia-Zhao-rh If this is with regards to testing, please hold off that for a while. I've just added the WIP label to indicate that this is still WIP and will update/rebase accordingly once it's ready for testing.

Xia-Zhao-rh:
> @Xia-Zhao-rh If this is with regards to testing, please hold that off for a while. I've just added the WIP label to indicate that this is still WIP and will update/rebase accordingly once it's ready for testing.

OK, got it, thanks a lot!

pkg/controller/bundle/bundle_unpacker.go
```diff
@@ -361,7 +375,22 @@ func (c *ConfigMapUnpacker) UnpackBundle(lookup *operatorsv1alpha1.BundleLookup)
 		return
 	}
 
-	if !jobConditionTrue(job, batchv1.JobComplete) && (cond.Status != corev1.ConditionTrue || cond.Reason != JobIncompleteReason) {
+	// Check if bundle unpack job has failed due a timeout
+	// Return a BundleJobError so we can mark the InstallPlan as Failed
+	isFailed, jobCond := jobConditionTrue(job, batchv1.JobFailed)
+	if isFailed {
+		cond.Status = corev1.ConditionTrue
```
benluddy, line 382: If the job has failed, shouldn't LookupPending become False?

pkg/controller/operators/catalog/operator.go
```diff
@@ -1447,6 +1462,28 @@ func (o *Operator) syncInstallPlans(obj interface{}) (syncError error) {
 	return
 }
 
+func (o *Operator) transitionInstallPlanToFailed(plan *v1alpha1.InstallPlan, logger *logrus.Entry, reason v1alpha1.InstallPlanConditionReason, message string) (syncError error) {
```
benluddy, line 1465:
```suggestion
func (o *Operator) transitionInstallPlanToFailed(plan *v1alpha1.InstallPlan, logger logrus.FieldLogger, reason v1alpha1.InstallPlanConditionReason, message string) error {
```
WDYT?

pkg/controller/bundle/bundle_unpacker.go
```diff
@@ -165,6 +166,12 @@ func (c *ConfigMapUnpacker) job(cmRef *corev1.ObjectReference, bundlePath string
 	job.SetName(cmRef.Name)
 	job.SetOwnerReferences([]metav1.OwnerReference{ownerRef(cmRef)})
 
+	// Don't set a timeout if it is 0
+	if c.unpackTimeout != time.Duration(0) {
+		t := int64(c.unpackTimeout.Seconds())
+		job.Spec.ActiveDeadlineSeconds = &t
```
exdx, line 172: oh nice, I didn't know a `Job` has that field on the spec. very handy for this exact use-case.

cmd/catalog/main.go
```diff
@@ -70,7 +70,8 @@ var (
 	profiling = flag.Bool(
 		"profiling", false, "serve profiling data (on port 8080)")
 
-	installPlanTimeout = flag.Duration("install-plan-retry-timeout", 1*time.Minute, "time since first attempt at which plan execution errors are considered fatal")
+	installPlanTimeout  = flag.Duration("install-plan-retry-timeout", 1*time.Minute, "time since first attempt at which plan execution errors are considered fatal")
+	bundleUnpackTimeout = flag.Duration("bundle-unpack-timeout", catalog.DefaultBundleUnpackTimeout, "The time duration after which the bundle unpack Job for an installplan will be aborted and the installplan will be Failed. 0 is considered as having no timeout.")
```
exdx, line 74: so this `bundleUnpackTimeout` is very similar to the `installPlanTimeout`-- I commented in @benluddy's timeout PR that we should have an exported constant for the default value of the timeout, but he disagreed saying there's no real reason to have that constant in the codebase if it's just used as a default for the flag and that's it. 

Based on that logic, WDYT about also just hardcoding the timeout? I don't see anywhere else in the PR where we use that DefaultBundleUnpackTimeout constant.

pkg/controller/operators/catalog/operator.go
```diff
@@ -77,6 +77,8 @@ const (
 	maxInstallPlanCount    = 5
 	maxDeletesPerSweep     = 5
 	RegistryFieldManager   = "olm.registry"
+	// TODO(haseeb): Change to a more realistic timeout for bundle unpack jobs
+	DefaultBundleUnpackTimeout = 2 * time.Minute
```
exdx, line 81: My impression is that 5 minutes might be a more reasonable deadline, if we consider how long it takes to pull a UBI based image, extract all its contents, and put the manifests we care about into a configmap. 

Might just be worth seeing how long it takes for the official red hat operators catalog (without that image already present on the node) and setting it around that

test/e2e/installplan_e2e_test.go
```diff
@@ -2998,6 +2999,143 @@ var _ = Describe("Install Plan", func() {
...
+			Expect(ctx.Ctx().Client().Create(context.Background(), outIP)).To(Succeed())
+
+			// The status gets ignored on create so we need to update it else the InstallPlan sync ignores
+			// InstallPlans without any steps or bundle lookups
+			outIP.Status = ip.Status
+			Expect(ctx.Ctx().Client().Status().Update(context.Background(), outIP)).To(Succeed())
+
+			// The InstallPlan should be Failed due to the bundle unpack job timeout
+			// We wait for some time over the bundle unpack Job's timeout (i.e ActiveDeadlineSeconds) so that the Job can fail
+			waitFor := catalog.DefaultBundleUnpackTimeout + 30*time.Second
```
exdx, line 3102: oh I see the constant is being used in test code. Still, I'm not sure it's worth keeping around versus just declaring it in the test package here. Making a constant in live code to only use it in test code sounds a little strange

exdx (commented):
Nice implementation, just leaving some thoughts

pkg/controller/operators/catalog/operator.go
benluddy: I've definitely spoken with users who have had unpacking take over 10 minutes. I think ultimately we want to give up fast when there are immediate errors (like `opm` crashing repeatedly), but allow more time when there may be nothing wrong other than slow I/O.

pkg/controller/bundle/bundle_unpacker.go
```diff
@@ -342,7 +356,22 @@ func (c *ConfigMapUnpacker) UnpackBundle(lookup *operatorsv1alpha1.BundleLookup)
...
+	// Return a BundleJobError so we can mark the InstallPlan as Failed
+	isFailed, jobCond := jobConditionTrue(job, batchv1.JobFailed)
+	if isFailed {
+		cond.Status = corev1.ConditionTrue
+		cond.Reason = jobCond.Reason
+		cond.Message = jobCond.Message
+		cond.LastTransitionTime = &now
+		result.SetCondition(cond)
+
+		err = NewBundleJobError(fmt.Sprintf("Bundle extract Job failed with Reason: %v, and Message: %v", jobCond.Reason, jobCond.Message))
```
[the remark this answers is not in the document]
hasbro17: Yeah that should be possible. Although I probably need to introduce a new bundleLookup.status condition type e.g `BundleLookupFailed`. So I can set that when `UnpackBundle()` sees that the Job has failed. And then in the main `syncInstallPlans()` loop I can check for the failed bundleLookup status condition to set the InstallPlan to failed as well.

pkg/controller/bundle/bundle_unpacker.go
hasbro17: Good catch. I need to make `BundleLookupPending==false` a no-op as well. Otherwise it'll continue on to making steps from the bundle.

test/e2e/installplan_e2e_test.go
hasbro17: Yeah my rationale for making it a constant was to always have the test time only after the bundle unpack timeout.

Else if we ever change the default for `--bundle-unpack-timeout` then you could have the test start failing as well unless you tweak the magic number here as well.

pkg/controller/operators/catalog/operator.go
hasbro17: To aid us in failing fast for immediate crashing errors we already have the `spec.backoffLimit` field right now which we don't set but by default is set to 6.
https://kubernetes.io/docs/concepts/workloads/controllers/job/#pod-backoff-failure-policy
```
Failed Pods associated with the Job are recreated by the Job controller with an exponential back-off delay (10s, 20s, 40s ...) capped at six minutes.
```

Although given that it will take `10s + 20s + 40s + 80s + 160s +320s = 10.5 minutes` that's probably not fast enough. So maybe `spec.backoffLimit=4` or `2.5m` gets us closer to failing fast for repeated pod failures.

hasbro17: I think I also need to set the pod failure policy `restartPolicy = "Never"` so the job can spawn a new pod per failure. Otherwise the failed containers go away on restart and it'll be hard to inspect failure logs.

pkg/controller/bundle/bundle_unpacker.go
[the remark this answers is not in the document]
hasbro17: Or actually right now we have `BundleLookupPending.status=true` indicating that we're still waiting for bundle unpacking.
The absence of the `BundleLookupPending` condition indicates a successful unpack of the bundle.
`BundleLookupPending.status=false` isn't being used to signify anything so I could use that to indicate a bundle unpack failure.

So I probably don't need a new condition type for failure.

test/e2e/installplan_e2e_test.go
hasbro17: Removed the constant for now.

pkg/controller/bundle/bundle_unpacker.go
```diff
@@ -165,6 +171,21 @@ func (c *ConfigMapUnpacker) job(cmRef *corev1.ObjectReference, bundlePath string
...
+	}
+
+	// By default the BackoffLimit is set to 6 which with exponential backoff 10s + 20s + 40s ...
+	// translates to ~10m of waiting time.
+	// We want to fail faster than that when we have repeated failures from the bundle unpack pod
+	// so we set it to 4 which is ~2.5m of waiting time
+	// See: https://kubernetes.io/docs/concepts/workloads/controllers/job/#pod-backoff-failure-policy
+	// TODO (haseeb): Should this be configurable as well?
+	backOffLimit := int32(4)
+	job.Spec.BackoffLimit = &backOffLimit
```
hasbro17, line 187: @benluddy Does this seem reasonable if we want to fail fast in the case of repeated crashes, e.g the `pull` or `extract` containers failing due to invalid bundle contents.

Ideally we'd fail after the first time on a fatal error like that, but it's hard to tell if there wasn't some other transient error that might be remedied by a restart.

pkg/controller/bundle/bundle_unpacker.go
```diff
@@ -75,7 +76,12 @@ func (c *ConfigMapUnpacker) job(cmRef *corev1.ObjectReference, bundlePath string
 					Name: cmRef.Name,
 				},
 				Spec: corev1.PodSpec{
-					RestartPolicy:    corev1.RestartPolicyOnFailure,
+					// With restartPolicy = "OnFailure" when the spec.backoffLimit is reached, the job controller will delete all
+					// the job's pods to stop them from crashlooping forever.
+					// By setting restartPolicy = "Never" the pods don't get cleaned up since they're not running after a failure.
+					// Keeping the pods around after failures helps in inspecting the logs of a failed bundle unpack job.
+					// See: https://kubernetes.io/docs/concepts/workloads/controllers/job/#pod-backoff-failure-policy
+					RestartPolicy:    corev1.RestartPolicyNever,
```
hasbro17, line 84: This is all well and good for when a pod's container errors out after it's running (e.g invalid bundle manifests). In that case the job fails fast due to `BackOffLimit=4`, the InstallPlan is marked as failed, and the failed pod sticks around for inspection.

However if the pod never starts running i.e `phase=Pending` and encounters the `ImagePullBackoff` or `ErrImagePull` error then the job won't fail fast. 
The job's `backoffLimit` is never invoked since the container hasn't exited in error yet.
So the job keeps running and will only fail once we hit the larger (10m) timeout from `ActiveDeadlineSeconds` at which point the pod is terminated.
Even more unfortunately the pod is removed after the timeout so you can't inspect it afterwards.

So for the case of a non-existent image in a bundlelookup, we're not really failing fast, or preserving the pod for debugging afterwards. We'll just eventually have the InstallPlan as Failed after a long timeout.

hasbro17: @benluddy I'm not sure how we want to handle this case. We could try and inspect the pod state for the case when we have an image pull failure and then fail fast ourselves:
```YAML
apiVersion: v1
kind: Pod
metadata:
  name: 8c3e265337115bdeff1ab7a092b9f531bffc580830a021ffe04d5214ee4k7kf
  namespace: default
  ownerReferences:
  - apiVersion: batch/v1
    blockOwnerDeletion: true
    controller: true
    kind: Job
    name: 8c3e265337115bdeff1ab7a092b9f531bffc580830a021ffe04d5214eea1419
    uid: 273dfba6-426e-46e1-b586-fdae4b792ad4
status:
  initContainerStatuses:
  ...
  - image: quay.io/foo/bar:latest
    imageID: ""
    lastState: {}
    name: pull
    ready: false
    restartCount: 0
    state:
      waiting:
        message: 'rpc error: code = Unknown desc = failed to pull and unpack image
          "quay.io/foo/bar:latest": failed to resolve reference "quay.io/foo/bar:latest":
          unexpected status code [manifests latest]: 401 UNAUTHORIZED'
        reason: ErrImagePull
  phase: Pending
```

Seems a little brittle though.

pkg/controller/bundle/bundle_unpacker.go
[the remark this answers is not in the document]
benluddy: I don't think that Pending being absent or False indicates anything about the failure state -- that should be absent or False for successfully completed lookups too.

pkg/controller/bundle/bundle_unpacker.go
benluddy: That seems reasonable to me. The extract container needs to talk to the API server to write to a ConfigMap, so there's at least some risk of transient error. I wouldn't object to an even lower limit (three attempts feels good).

I don't think it's worth it, but we could always instrument unpacking to report metrics on bundle unpacking (time to completion, number of attempts, outcome) and try to gauge how useful retries are.

pkg/controller/bundle/bundle_unpacker.go
benluddy: Well, a long timeout is still an improvement over no timeout.

Is there a middle ground where the Pending lookup condition message indicates what is currently happpening? Then it would take a while to fail, but at least if a user inspects the status then she will see that there is a Job with a pod in Pending/ImagePullBackOff. I agree that there may be more we can do to fail fast in special cases, but also agree that it could be brittle. That could be a future improvement?

pkg/controller/bundle/bundle_unpacker.go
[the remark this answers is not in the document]
hasbro17: Yeah in this current update I ended up recycling `BundleLookupPending.status=False` since it was unused. But I agree that's probably not as clear. I'm going to update with a dedicated failure condition e.g `BundleLookupFailed`.

cmd/catalog/main.go
hasbro17: Removed the constant.

pkg/controller/bundle/bundle_unpacker.go
hasbro17: We can peek at the container statuses while we're waiting for bundle lookup to try and add more context to the `BundleLookupPending` condition.
Not sure if that covers anything else but at least it will surface this particular issue to the user. Going to add that in the next update.

hasbro17: Okay so the BundleLookupPending condition should have the container status info for any containers that aren't ready in pending unpack Job pods.
That way while the InstallPlan is stalled on phase Installing with BundleLookupPending for a non-existent bundle image we can see the reason why:
```YAML
kind: InstallPlan
status:
  bundleLookups:
    conditions:
    - lastTransitionTime: "2021-04-09T00:44:55Z"
      message: "unpack job not completed: Unpack pod(default/8c3e265337115bdeff1ab7a092b9f531bffc580830a021ffe04d5214eenb2xr)
        container(pull) is pending. Reason: ErrImagePull, Message: rpc error: code
        = Unknown desc = failed to pull and unpack image \"quay.io/foo/bar:latest\":
        failed to resolve reference \"quay.io/foo/bar:latest\": unexpected status
        code [manifests latest]: 401 UNAUTHORIZED \n"
      reason: JobIncomplete
      status: "True"
      type: BundleLookupPending
    path: quay.io/foo/bar:latest
``` 

Of course it keeps toggles back and forth between `ErrImagePull` and `ImagePullBackOff` as it backs off but that's fine I think:
```YAML
kind: InstallPlan
status:
  bundleLookups:
    conditions:
    - lastTransitionTime: "2021-04-09T00:44:55Z"
      message: "unpack job not completed: Unpack pod(default/8c3e265337115bdeff1ab7a092b9f531bffc580830a021ffe04d5214eenb2xr)
        container(pull) is pending. Reason: ImagePullBackOff, Message: Back-off pulling
        image \"quay.io/foo/bar:latest\" \n"
      reason: JobIncomplete
      status: "True"
      type: BundleLookupPending
    path: quay.io/foo/bar:latest
  phase: Installing
```

Hopefully that makes the non-existent image case more apparent.

pkg/controller/bundle/bundle_unpacker.go
hasbro17: I've bumped the number of retries down to 3.
@benluddy Do you think it might be worth making this configurable? Not sure it would be much help for fatal errors. And I guess 3 is good enough for transient errors as well.

hasbro17:
Updated the PR description with the various failure cases and their expected output on the InstallPlan.

hasbro17:
Intentionally formatted the title so the BZ bot doesn't move https://bugzilla.redhat.com/show_bug.cgi?id=1921264 to Modified once this merges.
That should happen once this is in the [downstream repo](https://github.com/openshift/operator-framework-olm).

pkg/controller/bundle/bundle_unpacker.go
```diff
@@ -394,6 +491,39 @@ func (c *ConfigMapUnpacker) UnpackBundle(lookup *operatorsv1alpha1.BundleLookup)
...
+
+		for _, ic := range pod.Status.InitContainerStatuses {
+			if ic.Ready {
+				// only check non-ready containers for their waiting reasons
+				continue
+			}
+
+			// Aggregate the wait reasons for all pending containers
+			containerStatusMessages = containerStatusMessages +
+				fmt.Sprintf("Unpack pod(%s/%s) container(%s) is pending. Reason: %s, Message: %s \n",
```
benluddy, line 519: Nit: If this is going to end up in a string field on the plan status, does it make sense to end it with a newline?

test/e2e/installplan_e2e_test.go
```diff
@@ -2998,6 +2999,155 @@ var _ = Describe("Install Plan", func() {
...
+
+			// The status gets ignored on create so we need to update it else the InstallPlan sync ignores
+			// InstallPlans without any steps or bundle lookups
+			outIP.Status = ip.Status
+			Expect(ctx.Ctx().Client().Status().Update(context.Background(), outIP)).To(Succeed())
+
+			// The InstallPlan should be Failed due to the bundle unpack job timeout
+			// We wait for some time over the default --bundle-unpack-timeout (i.e ActiveDeadlineSeconds) so that the Job can fail
+			defaultBundleUnpackTimeout := 10 * time.Minute
+			waitFor := defaultBundleUnpackTimeout + 30*time.Second
```
benluddy, line 3139: Can we add an optional timeout field to the InstallPlan spec (using the command-line flag as a default) to make this easier to test?

pkg/controller/bundle/bundle_unpacker.go
```diff
@@ -361,16 +417,57 @@ func (c *ConfigMapUnpacker) UnpackBundle(lookup *operatorsv1alpha1.BundleLookup)
...
+		failedCond.Message = jobCond.Message
+		failedCond.LastTransitionTime = &now
+		result.SetCondition(failedCond)
+
+		// BundleLookupPending is false with reason being job failed
+		cond.Status = corev1.ConditionFalse
+		cond.Reason = JobFailedReason
+		cond.Message = JobFailedMessage
 		cond.LastTransitionTime = &now
 		result.SetCondition(cond)
```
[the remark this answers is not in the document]
hasbro17: Yeah, with two of them it's confusing. I prefixed the other one as `failedCond` so I'll change this one too.

pkg/controller/bundle/bundle_unpacker.go
hasbro17: Yeah I noticed it's not really rendering a newline in the yaml output from `kubectl get installplan ... -o yaml` anyway.
I wanted to demarcate between multiple container status error messages, so maybe `|` is better.

test/e2e/installplan_e2e_test.go
hasbro17: As discussed offline, while a new `spec.bundleUnpackTimeout` won't be back-portable we can specify a per InstallPlan bundle unpack timeout via an annotation for now, e.g:
```YAML
kind: InstallPlan
metadata:
  annotations:
    olm/bundle-unpack-timeout: 2m
```

That way we can set a shorter timeout for this test.

And we can potentially graduate the annotation to a spec field later on.

hasbro17: Also I just realized I have this timeout on the wrong test. It should be on the other one for `should show an error on the bundlelookup condition for a non-existent bundle image`.
That's where we need to wait to expire the 10m `ActiveDeadlineSeconds` because the pods stay pending.

Here we can simply wait for the `BackoffLimit` of 3 retries to fail i.e (10s+20s+40s) or > 1m10s.

I'll update that as well.

test/e2e/installplan_e2e_test.go
```diff
@@ -2998,6 +3000,171 @@ var _ = Describe("Install Plan", func() {
...
+			}
+		})
+
+		AfterEach(func() {
+			Expect(ctx.Ctx().Client().Delete(context.Background(), ns)).To(Succeed())
+		})
+
+		It("should show an error on the bundlelookup condition for a non-existent bundle image", func() {
+			// Create an InstallPlan status.bundleLookups.Path specified for a non-existent bundle image
+			ip.Status.BundleLookups[0].Path = "quay.io/foo/bar:v0.0.1"
```
benluddy, line 3090: This could break if someone publishes to quay.io/foo/bar. Does the test work if this points to something that isn't a registry, like `localhost:0/does/not:exist`?

test/e2e/installplan_e2e_test.go
```diff
@@ -2998,6 +3000,171 @@ var _ = Describe("Install Plan", func() {
...
+
+			// Create the single (kiali) bundle catalog source
+			catsrc := &operatorsv1alpha1.CatalogSource{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      genName("kiali-"),
+					Namespace: ns.GetName(),
+					Labels:    map[string]string{"olm.catalogSource": "kaili-catalog"},
+				},
+				Spec: operatorsv1alpha1.CatalogSourceSpec{
+					Image:      "quay.io/olmtest/single-bundle-index:1.0.0",
```
benluddy, line 3024: Does this need to be a real image ref? It looks like the unpacker only requires that a CatalogSource exists.

test/e2e/installplan_e2e_test.go
```diff
@@ -2998,6 +3000,171 @@ var _ = Describe("Install Plan", func() {
...
+			ip.SetAnnotations(annotations)
+			waitFor := 1*time.Minute + 30*time.Second
+
+			outIP := ip.DeepCopy()
+
+			Expect(ctx.Ctx().Client().Create(context.Background(), outIP)).To(Succeed())
+
+			// The status gets ignored on create so we need to update it else the InstallPlan sync ignores
+			// InstallPlans without any steps or bundle lookups
+			outIP.Status = ip.Status
```
benluddy, line 3106: Is this plus the above clone necessary? There are other InstallPlan tests that call update and status update consecutively that seem to work, which makes me think that Create() doesn't overwrite the status of the provided object.

test/e2e/installplan_e2e_test.go
hasbro17: It does. Will update this.

test/e2e/installplan_e2e_test.go
hasbro17: You're right. Seems like we don't need a running CatalogSource here. Just a dummy one would do.

test/e2e/installplan_e2e_test.go
hasbro17: Hmm I could have sworn that create [updates the object](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/vendor/sigs.k8s.io/controller-runtime/pkg/client/typed_client.go#L51) but from a quick test the status block does not get updated.

So yeah no need for a clone in that case.

hasbro17:
Updated the e2e tests. Added a wait for the OperatorGroup to be synced since that was causing some flakes where the InstallPlan would fail when it would see an OperatorGroup before it was synced.
Previously that was hidden by the fact that we were waiting for the CatalogSource to become ready during which the OperatorGroup would have time to sync.

njhale:
Removed bug title so we don't confuse the BZ-bot.

The bug title should be placed on the downstream PR.


/bugzilla refresh

pkg/controller/bundle/bundle_unpacker.go
```diff
@@ -165,26 +185,62 @@ func (c *ConfigMapUnpacker) job(cmRef *corev1.ObjectReference, bundlePath string
 	job.SetName(cmRef.Name)
 	job.SetOwnerReferences([]metav1.OwnerReference{ownerRef(cmRef)})
 
+	// By default the BackoffLimit is set to 6 which with exponential backoff 10s + 20s + 40s ...
+	// translates to ~10m of waiting time.
+	// We want to fail faster than that when we have repeated failures from the bundle unpack pod
+	// so we set it to 3 which is ~1m of waiting time
+	// See: https://kubernetes.io/docs/concepts/workloads/controllers/job/#pod-backoff-failure-policy
+	// TODO (haseeb): Should this be configurable as well?
```
njhale, line 193: It might help to adopt an idiom like: "present the smallest possible surface to users" -- i.e. if we're unsure of whether a knob is necessary, don't include it.

Anything we expose becomes an API; the less we add to it, the easier it is to support and extend.

pkg/controller/bundle/bundle_unpacker.go
```diff
@@ -66,7 +81,7 @@ func newBundleUnpackResult(lookup *operatorsv1alpha1.BundleLookup) *BundleUnpack
 	}
 }
 
-func (c *ConfigMapUnpacker) job(cmRef *corev1.ObjectReference, bundlePath string, secrets []corev1.LocalObjectReference) *batchv1.Job {
+func (c *ConfigMapUnpacker) job(cmRef *corev1.ObjectReference, bundlePath string, secrets []corev1.LocalObjectReference, ipAnnotations map[string]string) *batchv1.Job {
```
njhale, line 84: I think this should have a `timeout` parameter instead. It's unnecessary to depend on the entire annotations map.

pkg/controller/bundle/bundle_unpacker.go
```diff
@@ -165,26 +185,62 @@ func (c *ConfigMapUnpacker) job(cmRef *corev1.ObjectReference, bundlePath string
...
+	job.Spec.ActiveDeadlineSeconds = &timeoutSeconds
+
 	return job
 }
 
 //go:generate go run github.com/maxbrunsfeld/counterfeiter/v6 . Unpacker
 
 type Unpacker interface {
-	UnpackBundle(lookup *operatorsv1alpha1.BundleLookup) (result *BundleUnpackResult, err error)
+	UnpackBundle(lookup *operatorsv1alpha1.BundleLookup, ipAnnotations map[string]string) (result *BundleUnpackResult, err error)
```
njhale, line 228: I think this should have a `timeout` parameter instead. It's unnecessary to depend on the entire annotations map.

cmd/catalog/main.go
```diff
@@ -70,7 +70,8 @@ var (
 	profiling = flag.Bool(
 		"profiling", false, "serve profiling data (on port 8080)")
 
-	installPlanTimeout = flag.Duration("install-plan-retry-timeout", 1*time.Minute, "time since first attempt at which plan execution errors are considered fatal")
+	installPlanTimeout  = flag.Duration("install-plan-retry-timeout", 1*time.Minute, "time since first attempt at which plan execution errors are considered fatal")
```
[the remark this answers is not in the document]
hasbro17: Kind of yeah but they apply to different stages.
`bundle-unpack-timeout` will only timeout the unpack Job.

`install-plan-retry-timeout` only applies after unpacking succeeds and we fail during the plan execution.
https://github.com/operator-framework/operator-lifecycle-manager/blob/9359b9dd5d5c755a38fc3d79d64cce7bb31b81fe/pkg/controller/operators/catalog/operator.go#L1499-L1508

pkg/controller/bundle/bundle_unpacker.go
```diff
@@ -25,6 +27,19 @@ import (
...
+	// TODO: Move to operator-framework/api/pkg/operators/v1alpha1
+	// BundleLookupFailed describes conditions types for when BundleLookups fail
+	BundleLookupFailed operatorsv1alpha1.BundleLookupConditionType = "BundleLookupFailed"
+
+	// TODO: This can be a spec field
+	// BundleUnpackTimeoutAnnotationKey allows setting a bundle unpack timeout per InstallPlan
+	// and overrides the default specified by the --bundle-unpack-timeout flag
+	// The time duration should be in the same format as accepted by time.ParseDuration()
+	// e.g 1m30s
+	BundleUnpackTimeoutAnnotationKey = "bundle-unpack-timeout"
```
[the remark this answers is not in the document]
hasbro17: This was largely motivated by the e2e test's need to set a shorter timeout without reconfiguring the catalog operator deployment flags.

And I see your point but @benluddy and I thought this might be more easily back-portable than a proper spec field for now.

cmd/catalog/main.go
```diff
@@ -70,7 +70,8 @@ var (
...
+	bundleUnpackTimeout = flag.Duration("bundle-unpack-timeout", 10*time.Minute, "The time duration after which the bundle unpack Job for an installplan will be aborted and the installplan will be Failed. 0 is considered as having no timeout.")
```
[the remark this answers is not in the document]
hasbro17: The timeout is only configurable per InstallPlan if you create the InstallPlan yourself.

In the more normal workflow of creating a subscription and having the catalog operator create the InstallPlan for you, the user won't have the option to retroactively apply the `bundle-unpack-timeout` annotation on the InstallPlan.
By that point the unpack Job would already be created with the default 10m timeout and won't be adjusted after the annotation is applied.

So users would have no way of tweaking the default in the standard workflow without the flag.

If this were a proper spec field then I guess we could reconcile and update/recreate the job after the annotation is applied to reflect the new timeout. But imo it's not great UX to set this on the InstallPlan after it's created, rather than setting it at the start via the Subscription.

So unless we change all of that, I'd argue we keep the flag for now if it's just the annotation.

pkg/controller/bundle/bundle_unpacker.go
hasbro17: Yep, this can just be a timeout. I was trying to keep all the bundle-unpack-timeout annotation logic grouped in one place but I can parse and pass only what we need right here.

pkg/controller/bundle/bundle_unpacker.go
hasbro17: Updated: We're parsing the timeout annotation before calling UnpackBundle() now so we don't pass in the entire annotation set anymore.

hasbro17:
/hold

Seeing a flake in an e2e test with that last change so I'm going to fix that first.

pkg/controller/bundle/bundle_unpacker.go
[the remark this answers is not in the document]
njhale: ah, the e2e test angle makes sense as a tradeoff.

cmd/catalog/main.go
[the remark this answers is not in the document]
njhale: ah, unpack isn't part of execution, thanks for pointing that out.

cmd/catalog/main.go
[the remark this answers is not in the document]
njhale: after having a conversation with @hasbro17 and @benluddy offline, since the annotation is mostly to make our lives easier test-wise, I think it's fine to expose this. Especially since we don't yet have info on what a great default is.

hasbro17:
Okay so the flake is unrelated to the last change. It's actually to do with the job's `BackoffLimit`.
Seems like if `BackoffLimit=n` then you're waiting for n+1 failures, not n failures.

So currently BackoffLimit=3 (with an expected wait of 10s + 20s + 40s=1m10s) is actually going to wait for 4 failures (10s + 20s + 40s + 80s=2m30s).
So I need to bump the timeout for the test `should timeout and fail the InstallPlan for an invalid bundle image` to 3m.

pkg/controller/bundle/bundle_unpacker.go
```diff
@@ -300,25 +367,32 @@ const (
...
-func (c *ConfigMapUnpacker) UnpackBundle(lookup *operatorsv1alpha1.BundleLookup) (result *BundleUnpackResult, err error) {
+func (c *ConfigMapUnpacker) UnpackBundle(lookup *operatorsv1alpha1.BundleLookup, annotationUnpackTimeout time.Duration) (result *BundleUnpackResult, err error) {
+
 	result = newBundleUnpackResult(lookup)
 
-	// if pending condition is missing, bundle has already been unpacked
-	cond := result.GetCondition(operatorsv1alpha1.BundleLookupPending)
-	if cond.Status == corev1.ConditionUnknown {
+	// if bundle lookup failed condition already present, then there is nothing more to do
+	failedCond := result.GetCondition(BundleLookupFailed)
```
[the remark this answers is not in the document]
hasbro17: That's correct. If the bundle unpack job fails (timeout or hits the backoff limit) then we fail the InstallPlan and don't retry.

pkg/controller/operators/catalog/operator.go
```diff
@@ -1179,18 +1183,41 @@ func (o *Operator) unpackBundles(plan *v1alpha1.InstallPlan) (bool, *v1alpha1.In
...
+		// if the failed condition is present it means the bundle unpacking has failed
+		failedCondition := res.GetCondition(bundle.BundleLookupFailed)
+		if failedCondition.Status == corev1.ConditionTrue {
+			unpacked = false
+			continue
+		}
+
+		// if the bundle lookup pending condition is present it means that the bundle has not been unpacked
+		// status=true means we're still waiting for the job to unpack to configmap
+		// status=false means the bundle unpack job has failed
```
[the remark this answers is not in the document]
hasbro17: Good catch. This is a hold over from a previous change where I didn't have an explicit condition (`BundleLookupFailed`) for marking the bundle as failed so I was using `BundleLookupPending.status=false` as the failure condition.
I'll revert this back.

Although we never set `BundleLookupPending.status=false` anywhere since we only consider the absence of `BundleLookupPending` condition to indicate that the look up has finished.

So effectively `BundleLookupPending.status=false` and `BundleLookupPending` being missing are the same thing i.e lookup has finished.

pkg/controller/bundle/bundle_unpacker_test.go
```diff
@@ -38,6 +38,7 @@ func TestConfigMapUnpacker(t *testing.T) {
 	now := func() metav1.Time {
 		return start
 	}
+	backoffLimit := int32(3)
```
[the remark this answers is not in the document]
hasbro17: I'll add some unit tests to ensure failed unpack jobs result in the correct bundlelookup status conditions.

pkg/controller/bundle/bundle_unpacker.go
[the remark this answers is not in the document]
hasbro17: I've seen variants of namespaced `operatorframework.io/` vs prefixed `olm.` annotations on olm resources but yeah namespacing sgtm.

pkg/controller/bundle/bundle_unpacker.go
```diff
@@ -165,26 +185,57 @@ func (c *ConfigMapUnpacker) job(cmRef *corev1.ObjectReference, bundlePath string
...
+	job.Spec.ActiveDeadlineSeconds = &timeoutSeconds
+
 	return job
 }
 
 //go:generate go run github.com/maxbrunsfeld/counterfeiter/v6 . Unpacker
 
 type Unpacker interface {
-	UnpackBundle(lookup *operatorsv1alpha1.BundleLookup) (result *BundleUnpackResult, err error)
+	UnpackBundle(lookup *operatorsv1alpha1.BundleLookup, annotationUnpackTimeout time.Duration) (result *BundleUnpackResult, err error)
```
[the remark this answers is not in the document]
hasbro17: Yeah I intentionally made this name verbose in `func (c *ConfigMapUnpacker) job()` to make it distinct from `c.timeout` which is the flag timeout.

But I can shorten the var names elsewhere.

pkg/controller/bundle/bundle_unpacker_test.go
```diff
@@ -751,6 +756,523 @@ func TestConfigMapUnpacker(t *testing.T) {
...
+												},
+											},
+											Resources: corev1.ResourceRequirements{
+												Requests: corev1.ResourceList{
+													corev1.ResourceCPU:    resource.MustParse("10m"),
+													corev1.ResourceMemory: resource.MustParse("50Mi"),
+												},
+											},
+										},
+										{
```
njhale, line 850: If this pod spec is identical in most test cases, can we factor it out?

pkg/controller/bundle/bundle_unpacker_test.go
```diff
@@ -751,6 +756,523 @@ func TestConfigMapUnpacker(t *testing.T) {
...
+								LastTransitionTime: &start,
+							},
+							{
+								Type:               BundleLookupFailed,
+								Status:             corev1.ConditionTrue,
+								Reason:             "DeadlineExceeded",
+								Message:            "Job was active longer than specified deadline",
+								LastTransitionTime: &start,
+							},
+						},
```
njhale, line 1152: Now that I'm looking at it directly, I could see how having both `Pending` and `Failed` present and `True` can be confusing for users.

pkg/controller/bundle/bundle_unpacker_test.go
```diff
@@ -789,7 +1313,8 @@ func TestConfigMapUnpacker(t *testing.T) {
 			)
 			require.NoError(t, err)
 
-			res, err := unpacker.UnpackBundle(tt.args.lookup)
+			// The annotation unpack timeout arg is negative so it is ignored for these tests
+			res, err := unpacker.UnpackBundle(tt.args.lookup, -1*time.Minute)
```
njhale, line 1317: I would have expected at least one test where the timeout is varied and its inclusion on the job spec is verified

test/e2e/installplan_e2e_test.go
```diff
@@ -2998,6 +3000,168 @@ var _ = Describe("Install Plan", func() {
...
+	When("waiting on the bundle unpacking job", func() {
+		var (
+			ns         *corev1.Namespace
+			catsrcName string
+			ip         *operatorsv1alpha1.InstallPlan
+		)
+		BeforeEach(func() {
+			ns = &corev1.Namespace{}
+			ns.SetName(genName("ns-"))
+			Expect(ctx.Ctx().Client().Create(context.Background(), ns)).To(Succeed())
```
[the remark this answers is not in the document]
hasbro17: I thought `Eventually` was used for polling situations where we expect to see a certain condition eventually.
Are we worried that all CRUD operations could block for a long time? We do have an overall test timeout for that case.

Plus I might be wrong but I don't see that we've been using Eventually for ensuring Create,Update,Delete operations succeed in any of our other tests.

pkg/controller/bundle/bundle_unpacker_test.go
hasbro17: I don't think so. I tried to trim down the Job spec and the problem is that `func (c *ConfigMapUnpacker) UnpackBundle()`
will fail to ensure that the expected Job is present with the desired spec.
https://github.com/operator-framework/operator-lifecycle-manager/blob/a326d83ff42bda14a6b206675f54f092fcec815c/pkg/controller/bundle/bundle_unpacker.go#L411-L430

While testing this I've realized that we could end up with a nil job when we find and delete an unpack job that has the incorrect spec or ownerrefs.
https://github.com/operator-framework/operator-lifecycle-manager/blob/a326d83ff42bda14a6b206675f54f092fcec815c/pkg/controller/bundle/bundle_unpacker.go#L358-L362

So I'm going to update to also return and retry when `job == nil` as well otherwise we could see a nil dereference later on.

test/e2e/installplan_e2e_test.go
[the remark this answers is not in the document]
hasbro17: Done

hasbro17:
Not sure why this unit test is failing all of a sudden:
```
--- FAIL: TestSyncOperatorGroups (1.40s)
    --- FAIL: TestSyncOperatorGroups/AllNamespaces/CSVPresent/InstallModeNotSupported (0.14s)
        operator_test.go:4489: 
            	Error Trace:	operator_test.go:4489
            	            	clusterroles.rbac.authorization.k8s.io "csv-role" already exists
            	Test:       	TestSyncOperatorGroups/AllNamespaces/CSVPresent/InstallModeNotSupported
```

Can't reproduce this locally but still looking.

hasbro17:
Seems like it was a flake.

njhale:
Nice job!

/lgtm

PR #2096, opened 2021-04-13
.github/workflows: Add the unit and verify workflows

timflannagan:
Introduce two additional GH action workflows that match the existing
prow-based unit/verify jobs, with the intention of migrating off prow
entirely for master/4.8+ branch(es).

**Reviewer Checklist**

### DIFF
--- a/.github/workflows/unit.yml
+++ b/.github/workflows/unit.yml
@@ -1,0 +1,25 @@
+name: unit
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    paths:
+      - '**'
+      - '!doc/**'
+jobs:
+  unit:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+        with:
+          go-version: '~1.16'
+      - name: Install Kubebuilder
+        run: |
+          os=$(go env GOOS)
+          arch=$(go env GOARCH)
+          curl -L https://go.kubebuilder.io/dl/2.3.1/${os}/${arch} | tar -xz -C /tmp/
+          sudo mv /tmp/kubebuilder_2.3.1_${os}_${arch} /usr/local/kubebuilder
+      - name: Run unit tests
+        run: make unit
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -1,0 +1,25 @@
+name: verify
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    paths:
+      - '**'
+      - '!doc/**'
+jobs:
+  verify:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+        with:
+          go-version: '~1.16'
+
+        # Note(tflannag): We need to explicitly setup the GOPATH as the generate-internal-groups.sh
+        # codegen script still relies on $GOPATH being set and will return an error otherwise.
+        # The setup-go action does not set this value for us: https://github.com/actions/setup-go/issues/14.
+      - name: Run the verify target
+        run: |
+          export GOPATH=$(go env GOPATH)
+          make verify


.github/workflows/verify.yml
```diff
@@ -0,0 +1,25 @@
...
+      - uses: actions/setup-go@v2
+        with:
+          go-version: '~1.16'
+
+        # Note(tflannag): We need to explicitly setup the GOPATH as the generate-internal-groups.sh
+        # codegen script still relies on $GOPATH being set and will return an error otherwise.
+        # The setup-go action does not set this value for us: https://github.com/actions/setup-go/issues/14.
+      - name: Run the verify target
+        run: |
+          export GOPATH=$(go env GOPATH)
```
timflannagan, line 24: I tried using the recommended [environment files](https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#environment-files) but the following error was produced when configuring the environment in a separate job:

```
$GOPATH/go.mod exists but should not
# Clients, listers, and informers
$GOPATH/go.mod exists but should not
./scripts/update_codegen.sh
$GOPATH/go.mod exists but should not
make: *** [Makefile:203: codegen] Error 1
```

timflannagan:
Corresponding PR in o/release to remove those prow-based jobs entirely: https://github.com/openshift/release/pull/18324.

timflannagan:
```
            	            	clusterroles.rbac.authorization.k8s.io "csv-role" already exists
```

/rerun-workflow unit

.github/workflows/unit.yml
```diff
@@ -0,0 +1,25 @@
+name: unit
```
exdx, line 1: nit: this ends up being called "unit/unit" and "verify/verify" on the github actions on the PR...could we have them be under "test/unit" or something instead?

.github/workflows/unit.yml
```diff
@@ -0,0 +1,25 @@
...
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+        with:
+          go-version: '~1.16'
+      - name: Install Kubebuilder
+        run: |
+          os=$(go env GOOS)
+          arch=$(go env GOARCH)
+          curl -L https://go.kubebuilder.io/dl/2.3.1/${os}/${arch} | tar -xz -C /tmp/
+          sudo mv /tmp/kubebuilder_2.3.1_${os}_${arch} /usr/local/kubebuilder
```
exdx, line 23: A bit surprised we have a hard kubebuilder dependency to run our unit tests, but I seem to recall that being introduced sometime last year. Not really a huge deal I suppose.

.github/workflows/unit.yml
exdx: same names as the registry github actions

timflannagan: That's how the registry GH actions are currently named, and I didn't see any evidence we have any consistent naming of workflows/jobs here. I don't have a strong opinion either way so I'm fine with renaming both of the workflow names to be `test` unless that's problematic having two separate workflows that share the same name.

.github/workflows/unit.yml
timflannagan: I didn't dive too much into why we have that explicit dependency as I was only copying the content in the base.Dockerfile, which is the container image that the unit prow job would be run on. Once this and the o/release PR merges, we can re-evaluate whether we need that build root Dockerfile anymore (along with the root Dockerfile, etc.).

.github/workflows/unit.yml
exdx: Yes, I don't know if we have an explicit convention. Looks like our OLM github actions now are under "upstream-tests/" and "e2e-tests/" so maybe putting these under "upstream-tests/" for now makes sense. 

That being said this is now solely the upstream repository so I feel like we could drop the "upstream-" in a follow-up PR.

timflannagan: @exdx There may be an existing upstream issue for this, or at least I mentioned it in removing the OCP manifests from the repository issue, but I plan on tackling some more updates to the CI pipeline once we migrate off of prow (note: but not tide), so I think it's fine to keep the current workflow/job names with the intention of trying to create a more generalized naming convention in a subsequent PR.

benluddy: +1, let's revisit this topic since there isn't an existing convention. Also agreed that explicitly calling things "upstream" doesn't make sense as we migrate the downstream bits out of operator-framework.

dinhxuanvu (approved):
/lgtm

PR #2098, opened 2021-04-14
fix environment variable merging from subscription.config

joelanford:
OLM's subscription.config environment variable merge algorithm is broken because it only overrides environment variables listed in the CSV if they are listed first in the subscription.config overrides. This PR fixes that problem.
 
Fixes a bug in the way OLM merges environment variable overrides

**Reviewer Checklist**


dinhxuanvu (approved):
/lgtm

PR #2102, opened 2021-04-16
fix(catalog): Reduce namespace resync in resolution failure

dinhxuanvu:
Every resolution error will trigger namespace resync up to 8 times.
This sometimes leads to excessive resync even though the problem
is persistent.

NotSatisfiable error from resolver is often terminal and require
further manual intervetion. This type of error usually won't get
resolved by retrying resolution. As a result, if catalog encounters
this error, it should move on and doesn't force immediate resync.
The resync will happen naturally later when the default resync is
triggered.

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -893,7 +893,7 @@ func (o *Operator) syncResolvingNamespace(obj interface{}) error {
 		// given not-satisfiable error is terminal and most likely require intervention
 		// from users/admins. Resyncing the namespace again is unlikely to resolve
 		// not-satisfiable error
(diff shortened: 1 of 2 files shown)


benluddy:
/hold

:+1:, this is not a sync error and it should not change without a change to the subscription or catalog set. Can you write at least a couple tests for this please?

dinhxuanvu:
@benluddy May I have your blessings on this PR? I would like to move forward with it.

benluddy:
> May I have your blessings on this PR?

Gesundheit!

/approve

hasbro17 (approved):
/lgtm

PR #2103, opened 2021-04-16
fix(openshift): prevent race condition in the clusteroperator controller

njhale:
Use a deep copy of the target ClusterOperator in SyncTracker to avoid unsafe concurrent access.

Signed-off-by: Nick Hale <njohnhale@gmail.com>


dinhxuanvu (approved):
/lgtm

PR #2104, opened 2021-04-19
Fix resolution error if inner entry doesn't provide a required API.

benluddy:
For bundle-to-bundle dependencies (as opposed to
subscription-to-bundle "dependencies"), candidates were filtered first
to only those that satisfy the constraint in question, then sorted in
channel order. Since the relative order of entries within a channel is
currently determined by replaces-distance from the channel head,
sorting would fail if an entry in the middle of a channel's replaces
chain did not satisfy the constraint.

For example, assume an operator "bar" has three available versions
within a single channel: bar-1, bar-2, and bar-3. If bar-1 and bar-3
provide the API "Bar", but bar-2 does not, then resolution will fail
for any operator that requires the API "Bar", because bar-1 and bar-3
are not connected by update edges without bar-2.

Now, all (package, channel, catalog) combinations satisfying a
constraint are enumerated and sorted before filtering to only those
entries that satisfy the constraint.

### DIFF
--- a/pkg/controller/registry/resolver/resolver.go
+++ b/pkg/controller/registry/resolver/resolver.go
@@ -339,15 +339,38 @@ func (r *SatResolver) getBundleInstallables(catalog registry.CatalogKey, predica
 		}
 
 		for _, d := range dependencyPredicates {
-			// errors ignored; this will build an empty/unsatisfiable dependency if no candidates are found
-			candidateBundles, _ := AtLeast(1, namespacedCache.FindPreferred(&bundle.sourceInfo.Catalog, d))
-			sortedBundles, err := r.sortBundles(candidateBundles)
+			sourcePredicate := False()
+			// Build a filter matching all (catalog,
+			// package, channel) combinations that contain
+			// at least one candidate bundle, even if only
+			// a subset of those bundles actually satisfy
+			// the dependency.
+			sources := map[OperatorSourceInfo]struct{}{}
+			for _, b := range namespacedCache.Find(d) {
+				si := b.SourceInfo()
+
+				if _, ok := sources[*si]; ok {
+					// Predicate already covers this source.
+					continue
+				}
+				sources[*si] = struct{}{}
+
+				sourcePredicate = Or(sourcePredicate, And(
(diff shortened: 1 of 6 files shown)


pkg/controller/registry/resolver/resolver_test.go
```diff
@@ -1319,6 +1319,8 @@ func genOperator(name, version, replaces, pkg, channel, catalogName, catalogName
 				Namespace: catalogNamespace,
 			},
 			DefaultChannel: defaultChannel != "" && channel == defaultChannel,
+			Package:        pkg,
```
dinhxuanvu, line 1322: How are these changes relevant to the fix?

dinhxuanvu (commented):
Looking good me except some changes on properties that don't seem to be relevant to the fix. Can you explain? Carry another fix within this?

ecordell:
/approve

approach looks good and fix makes sense.

pkg/controller/registry/resolver/resolver.go
```diff
@@ -469,6 +483,19 @@ func (r *SatResolver) newSnapshotForNamespace(namespace string, subs []*v1alpha1
 		op.sourceInfo = &OperatorSourceInfo{
 			Catalog: existingOperatorCatalog,
 		}
+		for _, p := range op.properties {
```
dinhxuanvu, line 486: Sorry. I should've commented here on the question how these changes relevant to the fix are.

test/e2e/subscription_e2e_test.go
```diff
@@ -55,6 +55,74 @@ var _ = Describe("Subscription", func() {
...
+					}
+				},
+				Equal(operatorsv1alpha1.SubscriptionSpec{
+					CatalogSource:          "test-catalog",
+					CatalogSourceNamespace: testNamespace,
+					Package:                "dependency",
+					Channel:                "channel-dependency",
+				}),
+			)))
+		})
```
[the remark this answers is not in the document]
benluddy: Done.

dinhxuanvu (approved):
/lgtm

PR #2105, opened 2021-04-19
Fix panic due to empty version in the ClusterOperator controller.

benluddy:
If the ClusterVersion singleton "version" has an empty string for the
value of its field .Status.Desired.Version, then the ClusterOperator
controller would panic.

Seen here: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/operator-framework_operator-lifecycle-manager/2104/pull-ci-operator-framework-operator-lifecycle-manager-master-unit/1384220564280840192#1:build-log.txt%3A1165.

### DIFF
--- a/pkg/controller/operators/openshift/helpers_test.go
+++ b/pkg/controller/operators/openshift/helpers_test.go
@@ -219,11 +219,22 @@ func TestIncompatibleOperators(t *testing.T) {
 	}
 	for _, tt := range []struct {
 		description string
+		cv          configv1.ClusterVersion
 		in          skews
 		expect      expect
 	}{
 		{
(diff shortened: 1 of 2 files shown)


pkg/controller/operators/openshift/helpers_test.go
```diff
@@ -219,11 +219,22 @@ func TestIncompatibleOperators(t *testing.T) {
 	}
 	for _, tt := range []struct {
 		description string
+		cv          configv1.ClusterVersion
```
njhale, line 222: nit: the amount of noise in the test cases could be reduced if this was changed `version string`.

PR #2110, opened 2021-04-20
replace dockerhub images in upstream.Dockerfile

exdx:

The upstream image building process is currently broken due to the new ratelimiting that dockerhub imposed for non-authenticated image pulls. To see, head to https://quay.io/repository/operator-framework/olm?tab=builds and see the latest builds all fail (due to image pull ratelimiting). These builds all use the upstream.Dockerfile which uses golang:alpine based images during the build process. 

The only solution is to a) move to images on quay or b) pay for more pulls from dockerhub.

This PR moves the base image to fedora 34 (to easily install latest versions of go) and switches the alpine image to scratch in the second stage of the build 

Size-wise the image remains roughly the same size
```
upstream-olm     scratch                    8c23e249cd84   25 minutes ago   212MB
upstream-olm     dockerhub                  af5fda43325f   21 hours ago     218MB
```

This impacts the ability to cut upstream releases: the tag must have a green build on quay before the release can be cut. 

**Reviewer Checklist**

### DIFF
--- a/upstream.Dockerfile
+++ b/upstream.Dockerfile
@@ -1,8 +1,9 @@
-FROM golang:1.16-alpine as builder
+FROM quay.io/fedora/fedora:34-x86_64 as builder
 LABEL stage=builder
 WORKDIR /build
 
-RUN apk update && apk add bash make git mercurial jq && apk upgrade
+# install dependencies and go 1.16
+RUN dnf update -y && dnf install -y bash make git mercurial jq wget golang && dnf upgrade -y
 
 # copy just enough of the git repo to parse HEAD, used to record version in OLM binaries
 COPY .git/HEAD .git/HEAD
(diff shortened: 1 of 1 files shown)


exdx:
Going to fix by using a centos/scratch based file instead.

exdx:
Might make more sense to use the multi-arch `34` tag instead of targeting x64 in the upstream dockerfile

exdx:
Looks like using the multi-arch image runs into GPG signing issues ala `#6 57.62 Error: GPG check FAILED` and the image build fails. 

The option to use `dnf update --nogpgcheck` exists but I think we would rather do the gpgcheck. Going to revert back to the x86 base image where the gpg key checks pass.

dinhxuanvu (approved):
/lgtm

PR #2114, opened 2021-04-22
Annotate CRDs that are installed alongside CSVs.

benluddy:
In the absence of a way to correlate installed resources back to their
bundle of origin, it's possible for upgrades to complete
early. Upgrades begin when the catalog operator creates a new
CSV (with .spec.replaces set) as part of InstallPlan execution, and
complete when the olm operator sees that the new CSV has .status.phase
Succeeded. When both versions of the operator being upgraded own the
same CRD, the new CSV may use the existing CRD to satisfy its CRD
ownership requirement, thereby transitioning to Succeeded and deleting
the old CSV.

This problem can be avoided for namespace-scoped bundle resources,
which perform the transfer by adding the new CSV as an owner reference
and allowing garbage collection to remove the old owner reference
post-upgrade. This is not possible for CRDs because cluster-scoped
resources cannot have owner references to a namespace-scoped resource
like ClusterServiceVersion.

It may be necessary to make a similar change affecting other
cluster-scoped resources, but CRDs are a common culprit of early
upgrade completion. Ideally, a first-class, cluster-scoped Bundle
resource will obsolete the need to use an annotation to track bundle
of origin.

Upstream fix for: https://bugzilla.redhat.com/show_bug.cgi?id=1947946


dinhxuanvu:
I recall we already have labels attached to CRD a part of Operator API. So does this work improve anything further as I feel this is a bit redundant?

pkg/controller/operators/internal/alongside/alongside.go
```diff
@@ -0,0 +1,80 @@
...
+import (
+	"fmt"
+	"hash/fnv"
+	"strings"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	prefix = "operatorframework.io/installed-alongside-"
```
dinhxuanvu, line 15: I recall we already have labels attached to CRD a part of Operator API. So does this work improve anything further as I feel this is a bit redundant?

benluddy: This associates a resource with a specific bundle installation, not a package, so that we can ensure that an owned CRD requirement is not satisfied until that step of its bundle installation has been applied.

dinhxuanvu: I think this is what you are already doing but I think we can simply add this prefix as key and value as a hash value of either CSV name or CSV name + version. When we are checking for requirement, we can compare that hash value to ensure the resources are what they supposed to be (either created/upgraded by the installplan). I would like to change the key a bit. Maybe `"operatorframework.io/install-id` if we end up using hash value.

benluddy: That's not sufficient because the same object can have multiple annotations of this form. This would be the case during an upgrade from one CSV to another providing the same CRD, or when two CSVs providing the same CRD are installed in two namespaces.

The value is "namespace/name" in order to be able to enumerate all of these CSVs. It would be possible to omit the value and instead list all CSVs on the cluster, then hash them and compare against the annotation key suffixes, but that would be sensitive to things like fixes for hash function bugs, and it's not possible in that case to regenerate these annotations once the InstallPlan has been garbage collected. It would also be possible to do no hashing at all and serialize a list of CSV names into the value of the annotation.

dinhxuanvu: I see what you meant. However, I'm not particularly a big fan of adding more and more annotations/labels into CSV and cascading all CSV names in that annotations. It only makes the in-cluster CSV becomes more bloated than it already is. I find massive CSV is just a painful to debug when it comes to it. Anyway, it is in-cluster representation so I don't think it is a big deal. I'm fine with it.
One more thing though if the descoping operator comes true, then we probably don't need worry about two instances of the same operator (owning the same CRD) anymore.

benluddy: The annotations aren't placed on CSVs, they're placed on things that were installed as part of the same bundle as a CSV.

I agree with that last point. I mention in the commit message that this is a stopgap measure to deal with the lack of a cluster-scoped API that represents the applied bundle itself. That's ideal, but it's not something that can be delivered as a bugfix for users that are seeing this problem today.

exdx:
Just to add more context to the original description: 

The problem of early success of the new CSV (based off the ownership of the same CRDs as the old CSV already on-cluster) is that the old CSV gets deleted too early. The old CSV may have ownerrefs to different resources which would then be GC'd automatically and cause the new CSV to go into a pending state. For example, if the old CSV owned a ServiceAccount, it would be removed before the new CSV could use it, and the new CSV would then be stuck in a Pending phase. 

Early success of the new CSV can also have an impact in the case where CR validation fails, but the CSV is in a succeeded state (since the CRDs it needs are already on-cluster), which the user may not notice unless they check the InstallPlan.

pkg/controller/operators/catalog/step.go
```diff
@@ -3,22 +3,27 @@ package catalog
 import (
 	"context"
 	"fmt"
+	"strings"
 
-	"k8s.io/client-go/dynamic"
-	"k8s.io/client-go/tools/cache"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
```
njhale, line 10: nit: extra space

test/e2e/csv_e2e_test.go
```diff
@@ -39,15 +40,351 @@ var _ = Describe("ClusterServiceVersion", func() {
...
+							Type:      v1alpha1.InstallModeTypeAllNamespaces,
+							Supported: true,
+						},
+					},
+				},
+			}
+			Expect(ctx.Ctx().Client().Create(context.Background(), &new)).To(Succeed())
+
+			Eventually(func() error {
+				return ctx.Ctx().Client().Get(context.Background(), client.ObjectKeyFromObject(&old), &old)
```
njhale, line 204: @benluddy could you check my understanding?

So the transition path for unassociated looks like: created -> pending ("there's an annotation that's not mine on the CRD") -> associated created -> no longer pending ("I'm being replaced") -> deleted

njhale (commented):
Nice job!

I think this is really close.

pkg/controller/operators/olm/requirements.go
```diff
@@ -95,6 +103,16 @@ func (a *Operator) requirementStatus(strategyDetailsDeployment *v1alpha1.Strateg
 			continue
 		}
 
+		if names := othersInstalledAlongside(crd, csv); len(names) > 0 && ownedCRDNames[crd.Name] {
+			if replacement := a.isBeingReplaced(csv, a.csvSet(csv.GetNamespace(), v1alpha1.CSVPhaseAny)); replacement == nil {
```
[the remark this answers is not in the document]
benluddy: That's intentional, not to prevent noise, but to avoid cases where the CSV being replaced transitions from Replacing to Pending (RequirementsNotMet) in the middle of an upgrade. This happens sometimes today and requires manual intervention (usually deleting the old CSV) to unblock the upgrade, even though the upgrade would otherwise have completed successfully.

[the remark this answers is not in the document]
benluddy: On re-read, you're right, good catch. There are at least a couple issues with replacement chain handling that need test cases and updates to this test. Thanks!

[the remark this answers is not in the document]
njhale: To be more clear, consider the following CSVs in namespace `ns`:
- A, A', and B all claim the same CRD
- A' replaces A

In this case, it looks like A would **not** become pending even though B, with no relation, exists. Do we care about this?

pkg/controller/operators/olm/requirements.go
```diff
@@ -410,3 +428,27 @@ func (a *Operator) isGVKRegistered(group, version, kind string) (bool, error) {
...
+
+	for _, nn := range a.FromObject(o) {
+		if nn.Namespace != csv.GetNamespace() {
+			continue
+		}
+		if nn.Name != csv.GetName() {
+			names = append(names, nn.Name)
+			continue
+		}
+		return nil
```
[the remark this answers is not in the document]
benluddy: Good catch, thanks. This was missing a test as well so I will go ahead and add coverage for that case.

pkg/controller/operators/catalog/operator.go
```diff
@@ -1642,7 +1642,7 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
 		o.logger.Errorf("failed to get a client for plan execution- %v", err)
 		return err
 	}
-	b := newBuilder(builderKubeClient, builderDynamicClient, o.csvProvidedAPIsIndexer, r, o.logger)
+	b := newBuilder(plan, o.lister.OperatorsV1alpha1().ClusterServiceVersionLister(), builderKubeClient, builderDynamicClient, o.csvProvidedAPIsIndexer, r, o.logger)
```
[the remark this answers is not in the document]
exdx: Yup, I double checked and it looks like it's not being used anywhere -- a holdover from the v1beta1 -> v1 CRD support refactor

pkg/controller/operators/internal/alongside/alongside.go
```diff
@@ -0,0 +1,83 @@
...
+	annotations := o.GetAnnotations()
+
+	for key := range annotations {
+		if strings.HasPrefix(key, prefix) {
+			delete(annotations, key)
+		}
+	}
+
+	if len(nns) == 0 {
+		if len(annotations) == 0 {
```
exdx, line 61: nit: is this `if` required? the zero value of a map is nil (which always has length zero)

pkg/controller/operators/olm/requirements.go
```diff
@@ -410,3 +428,27 @@ func (a *Operator) isGVKRegistered(group, version, kind string) (bool, error) {
...
+// installed, excluding the given CSV and any CSVs in other
+// namespaces.
+func othersInstalledAlongside(o metav1.Object, csv *v1alpha1.ClusterServiceVersion) []string {
+	var (
+		a     alongside.Annotator
+		names []string
+	)
+
+	for _, nn := range a.FromObject(o) {
+		if nn.Namespace != csv.GetNamespace() {
```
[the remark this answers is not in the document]
njhale: let's punt on this because there are other multitenancy checks.

pkg/controller/operators/internal/alongside/alongside.go
njhale: At this point `annotations` could either be `nil` or the zero value. I think the point is to ensure we always set the annotations on `o` to `nil` in this situation.

pkg/controller/operators/internal/alongside/alongside.go
```diff
@@ -0,0 +1,83 @@
...
+
+const (
+	prefix = "operatorframework.io/installed-alongside-"
+)
+
+// NamespacedName is a reference to an object by namespace and name.
+type NamespacedName struct {
+	Namespace string
+	Name      string
+}
```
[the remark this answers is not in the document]
benluddy: No, I was aware of that type but I wanted to skip the dependency -- especially in conjunction with your other suggestion here (which removes the metav1 import too).

test/e2e/csv_e2e_test.go
benluddy: That's right. This spec should check that replacements do not block when the old CSV is not associated with its CRD, since that will be the state of most CSVs that pre-date this change.

pkg/controller/operators/olm/requirements.go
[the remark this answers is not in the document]
benluddy: Discussed this at length last week. Ultimately, this test is never reached if there are multiple CRD owners in separate replacement chains within the same namespace, but I've updated it to explicitly consider only ancestors within the reconciling CSV's replacement chain.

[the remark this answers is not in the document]
benluddy: I think this should also be resolved by the other replacement chain changes we discussed. This requirement should now only be failed when an ancestor of the owning CSV is referenced by a CRD annotation AND the owning CSV is not.

PR #2125, opened 2021-04-28
Use DeploymentAvailable instead of custom test for CSV status.

benluddy:
CSVs can transition to Failed (reason ComponentUnhealthy) if a single
pod managed by one of its deployments is deleted. This commit changes
the deployment availability test from a comparison between
.status.availableReplicas and .status.updatedReplicas to a check that
passes unless the deployment's Available condition exists and is
False.

### DIFF
--- a/deploy/chart/values.yaml
+++ b/deploy/chart/values.yaml
@@ -38,6 +38,8 @@ catalog:
 
 package:
   replicaCount: 2
+  maxUnavailable: 1
+  maxSurge: 1
   image:
     ref: quay.io/operator-framework/olm:master
     pullPolicy: Always
--- a/pkg/controller/install/status_viewer.go
+++ b/pkg/controller/install/status_viewer.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 
 	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
 )
 
 const TimedOutReason = "ProgressDeadlineExceeded"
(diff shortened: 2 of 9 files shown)


pkg/controller/install/status_viewer.go
```diff
@@ -26,9 +27,8 @@ func DeploymentStatus(deployment *appsv1.Deployment) (string, bool, error) {
 		if deployment.Status.Replicas > deployment.Status.UpdatedReplicas {
 			return fmt.Sprintf("Waiting for rollout to finish: %d old replicas are pending termination...\n", deployment.Status.Replicas-deployment.Status.UpdatedReplicas), false, nil
 		}
-		// waiting for new replicas to report as available
-		if deployment.Status.AvailableReplicas < deployment.Status.UpdatedReplicas {
-			return fmt.Sprintf("Waiting for rollout to finish: %d of %d updated replicas are available...\n", deployment.Status.AvailableReplicas, deployment.Status.UpdatedReplicas), false, nil
+		if c := getDeploymentCondition(deployment.Status, appsv1.DeploymentAvailable); c != nil && c.Status == corev1.ConditionFalse {
```
benluddy, line 30: I'm not sure under what conditions the Available condition could be missing or Unknown. This treats a deployment as available _unless_ it is present and False.

njhale: The [go docs](https://github.com/kubernetes/kubernetes/blob/db70c2a96ac0c85cd44f5a9f533ffedaf6df004c/staging/src/k8s.io/api/apps/v1/types.go#L421) for the field make me think that the condition is present in the steady state. I also have anecdotal evidence to back this interpretation.

To that end, this needs to treat a deployment as available _only_ when it is both present and True.

deploy/chart/values.yaml
```diff
@@ -38,6 +38,8 @@ catalog:
 
 package:
   replicaCount: 2
+  maxUnavailable: 1
+  maxSurge: 1
```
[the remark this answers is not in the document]
njhale: maxUnavailable turns out to be `0` in the default case, so I was wrong.

PR #2133, opened 2021-05-03
fix(bundles): symlink cp at expected path

njhale:
Symlink cp at /bin/cp to satisfy bundle unpack Job assumptions. The
previous switch to gcr.io/distroless as a base image effectively moved
cp from /bin/cp to /busybox/cp, resulting in unpack failures when cp is mounted
in unpack init containers.

### DIFF
--- a/upstream.Dockerfile
+++ b/upstream.Dockerfile
@@ -3,9 +3,9 @@ LABEL stage=builder
 WORKDIR /build
 
 # install dependencies and go 1.16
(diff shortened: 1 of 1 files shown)


exdx:
Going to check locally that this fixes bundle unpacking

njhale:
> Going to check locally that this fixes bundle unpacking

@exdx I just verified the patch locally, so please double check my findings.

exdx:
LGTM to me locally as well, subscription flow works as expected

Issue #2092, opened 2021-04-08
Flake: TestConnectionEvents deadlock?


benluddy:
Saw this here https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/operator-framework_operator-lifecycle-manager/2090/pull-ci-operator-framework-operator-lifecycle-manager-master-unit/1380242533942038528#1:build-log.txt%3A48, looks like a standoff:

```INFO[2021-04-08T19:50:20Z] ?   	github.com/operator-framework/operator-lifecycle-manager/pkg/controller/registry	[no test files] 
INFO[2021-04-08T19:50:20Z] panic: test timed out after 10m0s            
INFO[2021-04-08T19:50:20Z]                                              
INFO[2021-04-08T19:50:20Z] goroutine 185 [running]:                     
INFO[2021-04-08T19:50:20Z] testing.(*M).startAlarm.func1()              
INFO[2021-04-08T19:50:20Z] 	/usr/local/go/src/testing/testing.go:1701 +0x11f 
INFO[2021-04-08T19:50:20Z] created by time.goFunc                       
INFO[2021-04-08T19:50:20Z] 	/usr/local/go/src/time/sleep.go:180 +0x52   
INFO[2021-04-08T19:50:20Z]                                              
INFO[2021-04-08T19:50:20Z] goroutine 1 [chan receive, 10 minutes]:      
INFO[2021-04-08T19:50:20Z] testing.(*T).Run(0xc00039ec00, 0x23879d5, 0x14, 0x2434c88, 0x0) 
INFO[2021-04-08T19:50:20Z] 	/usr/local/go/src/testing/testing.go:1240 +0x610 
INFO[2021-04-08T19:50:20Z] testing.runTests.func1(0xc00039ec00)         
INFO[2021-04-08T19:50:20Z] 	/usr/local/go/src/testing/testing.go:1512 +0xa7 
INFO[2021-04-08T19:50:20Z] testing.tRunner(0xc00039ec00, 0xc0007bfce0)  
INFO[2021-04-08T19:50:20Z] 	/usr/local/go/src/testing/testing.go:1194 +0x203 
INFO[2021-04-08T19:50:20Z] testing.runTests(0xc0002824c8, 0x329c9b0, 0x1, 0x1, 0xc013f57d9455a068, 0x8c300fb3ee, 0x333d600, 0xc0000ae1a8) 
[...]

PR #2134, opened 2021-05-03
Remove OpenShift-specific Dockerfiles.

benluddy:


Dockerfile
```diff
@@ -1,56 +1,36 @@
-FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.16-openshift-4.8 AS builder
```
timflannagan, line 1: Do we want to wait until all the prow-based jobs, which rely on this Dockerfile and we substitute images based on known base image names, until after they all have been migrated to GH actions?

timflannagan: Yeah, same comment w.r.t to the base.Dockerfile, which is listed as the build root Dockerfile in the o/release prow configuration. I'm working towards removing the prow tests after I get the marketplace k8s bump off my plate.

benluddy: Definitely, let me know if I can help with that.

/hold

timflannagan: @benluddy I have a PR open for adding the unit/verify tests to GH actions in https://github.com/operator-framework/operator-lifecycle-manager/pull/2096 and once that merges we can merge the [o/release PR](https://github.com/openshift/release/pull/18324) for removing these OCP specific Dockerfile(s).

timflannagan: Both of those aformentioned PRs merged - I'm on the fence whether we want to just remove the OCP-specific Dockerfile(s) here (e.g. Dockerfile, base.Dockerfile) and keep the upstream.Dockerfile w.o. deflating/merging with a single Dockerfile. I think it's fine to iterate on this going forward as our upstream CI doesn't test using the same container image we attempt to successfully build.

timflannagan:
I'd like to get another person to sign off on the changes, but this seems reasonable to me and is largely a no-op now that we're removed prow from this repository.

/lgtm
/hold

benluddy:
> /rerun-workflow unit

Dockerfile
benluddy: > just remove the OCP-specific Dockerfile(s) here

That's all that is happening here -- Dockerfile and base.Dockerfile are removed, and upstream.Dockerfile is renamed to Dockerfile with no changes. The diff is just misleading.

PR #2136, opened 2021-05-04
Fix CR validation bug on go 1.16

ecordell:
See https://github.com/kubernetes/kube-openapi/pull/230 for details


dinhxuanvu (approved):
/lgtm

pkg/controller/operators/catalog/operator.go
```diff
@@ -1600,7 +1600,7 @@ func validateExistingCRs(dynamicClient dynamic.Interface, gvr schema.GroupVersio
 		}
 		err = validation.ValidateCustomResource(field.NewPath(""), cr.UnstructuredContent(), validator).ToAggregate()
 		if err != nil {
-			return fmt.Errorf("error validating custom resource against new schema %#v: %s", newCRD.Spec.Validation, err)
+			return fmt.Errorf("error validating custom resource against new schema: %v", err)
```
arjunrn, line 1603: When I hit this error in OLM I thought it would be useful if the error message indicated which specific resource failed the validation.

PR #2140, opened 2021-05-05
Set reason/message for Available condition in packageserver co

anik120:

**Reviewer Checklist**

### DIFF
--- a/pkg/lib/operatorstatus/csv_reporter.go
+++ b/pkg/lib/operatorstatus/csv_reporter.go
@@ -11,6 +11,12 @@ import (
 const (
 	// versionName reflects the name of the version CVO expects in Status.
 	versionName = "operator"
+
+	reasonCSVSucceeded = "ClusterServiceVersionSucceeded"
+
+	reasonCSVNotSucceeded = "ClusterServiceVersionNotSucceeded"
+
+	reasonCSVDeleted = "ClusterServiceVersionDeleted"
 )
 
 // newCSVStatusReporter returns a new instance of CSVStatusReporter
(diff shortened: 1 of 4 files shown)


pkg/lib/operatorstatus/csv_reporter.go
```diff
@@ -80,7 +84,7 @@ func (r *csvStatusReporter) GetNewStatus(existing *configv1.ClusterOperatorStatu
 
 		if context.WorkingToward == nil {
 			builder.WithProgressing(configv1.ConditionFalse, fmt.Sprintf("Uninstalled version %s", csv.Spec.Version)).
-				WithAvailable(configv1.ConditionFalse, "")
+				WithAvailable(configv1.ConditionFalse, "", "")
```
benluddy, line 87: These other two callsites should set an informative message, too.

pkg/lib/operatorstatus/csv_reporter.go
```diff
@@ -98,9 +102,9 @@ func (r *csvStatusReporter) GetNewStatus(existing *configv1.ClusterOperatorStatu
 
 	switch phase {
 	case v1alpha1.CSVPhaseSucceeded:
-		builder.WithAvailable(configv1.ConditionTrue, "")
+		builder.WithAvailable(configv1.ConditionTrue, fmt.Sprintf("ClusterServiceVersion %v/%v is in phase %v", csv.Namespace, csv.Name, csv.Status.Phase), reasonCSVSucceeded)
 	default:
-		builder.WithAvailable(configv1.ConditionFalse, "")
+		builder.WithAvailable(configv1.ConditionFalse, fmt.Sprintf("ClusterServiceVersion %v/%v is in phase %v", csv.Namespace, csv.Name, csv.Status.Phase), reasonCSVNotSucceeded)
```
benluddy, line 107: I'd add the CSV's .status.reason and/or .status.message at the end of this message. The person reading this message may have been paged while away from work/home and will appreciate having a clearer picture of what might be wrong before opening their laptop and investigating.

anik120: Done.

pkg/lib/operatorstatus/csv_reporter.go
anik120: @benluddy updated this calling site, but for the other calling site, i.e when the co is being initialized and is waiting for events in the CSV, looks like the Progressing condition has the skeleton message "waiting for events - source=csv.Name". That should be enough right? Coz the Available condition will probably have to spew out the same message otherwise. Also, since this in only during init, I'm guessing no one's getting paged when the cluster is being created/upgraded.

hasbro17 (commented):
/lgtm

PR #2147, opened 2021-05-07
Simplify deployment status check to reduce flapping.

benluddy:
The deployment status logic (originally based on kubectl's rollout
status) can impact its ClusterServiceVersion's status (e.g. by sending
it to Failed/ComponentUnhealthy or triggering reinstallation). If
these ClusterServiceVersion status updates are generated when no
changes have been made to the Deployment spec and when the Deployment
itself has condition Available: True, then the usefulness of
ClusterServiceVersion statuses as a signal is reduced.

This patch changes the deployment status logic so that it is
considered "done" when a) there are no replicas from earlier
generations, and b) the Deployment has condition Available: true.

### DIFF
--- a/pkg/controller/install/status_viewer.go
+++ b/pkg/controller/install/status_viewer.go
@@ -1,7 +1,5 @@
 package install
 
-// See kubernetes/pkg/kubectl/rollout_status.go
-
 import (
 	"fmt"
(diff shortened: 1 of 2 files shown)


benluddy:

PR #2152, opened 2021-05-10
feat(metrics): Emit metrics for CatalogSource state

anik120:

This PR introduces a prometheus gauage vector `catalogsource_ready`
that is emitted by each CatalogSource to indicate the connectivity.State
of the CatalogSource object. A value of 1 for the vector indicates that
the CatalogSource object is in a READY state, while a value of 0
indicates that the CatalogSource object is in one of IDLE/CONNECTING/
TRANSIENT_FAILURE/SHUTDOWN state. If/When the CatalogSource object
eventually reaches a READY state, the value for the vector is set
to 0.

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -396,6 +396,7 @@ func (o *Operator) syncSourceState(state grpc.SourceState) {
 	o.sourcesLastUpdate.Set(o.now().Time)
 
 	o.logger.Infof("state.Key.Namespace=%s state.Key.Name=%s state.State=%s", state.Key.Namespace, state.Key.Name, state.State.String())
+	metrics.RegisterCatalogSourceState(state.Key.Name, state.Key.Namespace, state.State)
 
 	switch state.State {
 	case connectivity.Ready:
@@ -513,6 +514,8 @@ func (o *Operator) handleCatSrcDeletion(obj interface{}) {
 		o.logger.WithError(err).Warn("error closing client")
 	}
 	o.logger.WithField("source", sourceKey).Info("removed client for deleted catalogsource")
+
+	metrics.DeleteCatalogSourceStateMetric(catsrc.GetName(), catsrc.GetNamespace())
 }
 
 func validateSourceType(logger *logrus.Entry, in *v1alpha1.CatalogSource) (out *v1alpha1.CatalogSource, continueSync bool, _ error) {
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -144,7 +144,7 @@ var (
 
 	catalogSourceReady = prometheus.NewGaugeVec(
 		prometheus.GaugeOpts{
-			Name: "catalogSource_ready",
+			Name: "catalogsource_ready",
 			Help: "State of a CatalogSource. 1 indicates that the CatalogSource is in a READY state. 0 indicates CatalogSource is in a Non READY state.",
 		},
 		[]string{NAMESPACE_LABEL, NAME_LABEL},
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -4,6 +4,7 @@ import (
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
+	"google.golang.org/grpc/connectivity"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 
@@ -141,6 +142,14 @@ var (
 		},
 	)
 
+	catalogSourceReady = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "catalogSource_ready",
+			Help: "State of a CatalogSource. 1 indicates that the CatalogSource is in a READY state. 0 indicates CatalogSource is in a Non READY state.",
+		},
+		[]string{NAMESPACE_LABEL, NAME_LABEL},
+	)
+
 	// exported since it's not handled by HandleMetrics
 	CSVUpgradeCount = prometheus.NewCounter(
 		prometheus.CounterOpts{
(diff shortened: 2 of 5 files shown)


pkg/metrics/metrics.go
```diff
@@ -141,6 +142,14 @@ var (
 		},
 	)
 
+	catalogSourceState = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "catalogSource_state",
```
benluddy, line 147:
```suggestion
			Name: "catalogsource_unready",
```

pkg/metrics/metrics.go
```diff
@@ -214,6 +224,19 @@ func CounterForSubscription(name, installedCSV, channelName, packageName, planAp
 	return SubscriptionSyncCount.WithLabelValues(name, installedCSV, channelName, packageName, planApprovalStrategy)
 }
 
+func RegisterCatalogSourceState(name, namespace string, state connectivity.State) {
+	switch state {
+	case connectivity.Ready:
+		catalogSourceState.WithLabelValues(namespace, name).Set(0)
+	default:
+		catalogSourceState.WithLabelValues(namespace, name).Inc()
```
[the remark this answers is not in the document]
joelanford: Yeah, I actually like this more than the comment in my previous suggestion. But could we invert it ( 0 == not ready, 1 == ready ) and name it `catalog_source_ready`?

https://www.robustperception.io/booleans-logic-and-math

pkg/metrics/metrics.go
benluddy: :+1: I was assuming here that my other suggestion about changing semantics was accepted.

test/e2e/util_test.go
```diff
@@ -521,6 +521,32 @@ func buildServiceAccountCleanupFunc(t GinkgoTInterface, c operatorclient.ClientI
...
+			Kind:       v1alpha1.CatalogSourceKind,
+			APIVersion: v1alpha1.CatalogSourceCRDAPIVersion,
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+		},
+		Spec: v1alpha1.CatalogSourceSpec{
+			SourceType: "grpc",
+			Image:      "quay.io/my-namespace/invalid-catsrc:latest",
```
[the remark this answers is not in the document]
anik120: I see the point, updated the ref to use `localhost:0/not/exists:catsrc` instead.

pkg/controller/operators/catalog/operator.go
```diff
@@ -488,6 +489,7 @@ func (o *Operator) handleDeletion(obj interface{}) {
 
 	o.requeueOwners(metaObj)
 
+	metrics.DeleteCatalogSourceStateMetric(metaObj.GetNamespace(), metaObj.GetName())
```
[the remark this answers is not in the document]
anik120: That's a great point. This was supposed to be in `handleCatsrcDeletion` and not `handleDeletetion(obj interface{})`
Moved it to `handleCatsrcDeletion`.

test/e2e/metrics_e2e_test.go
```diff
@@ -266,6 +270,109 @@ var _ = Describe("Metrics are generated for OLM managed resources", func() {
 			})
 		})
 	})
+
+	Context("Metrics emitted by CatalogSources", func() {
+		When("A valid CatalogSource object is created", func() {
+			var (
+				name        = "metrics-catsrc-valid"
+				cleanup     func()
+				cleanupDone = false
```
benluddy, line 279: You'll want to set this to `true` in a `BeforeEach` -- this assignment will only happen once at package initialization time.

anik120: Right, that's what happens here https://github.com/operator-framework/operator-lifecycle-manager/pull/2152/files#diff-d5a3dea398f2933b707ed901c793832f40784c0999a861847356f823e536c5d4R324-R326

anik120: @benluddy do you mean like it's being set here https://github.com/operator-framework/operator-lifecycle-manager/pull/2152/files#diff-d5a3dea398f2933b707ed901c793832f40784c0999a861847356f823e536c5d4R324-R327 or do you have something else in mind?

jan--f:
I think this looks good. A second set of eyes would be good though cc @paulfantom

test/e2e/metrics_e2e_test.go
```diff
@@ -266,6 +271,110 @@ var _ = Describe("Metrics are generated for OLM managed resources", func() {
...
+					ContainElement(LikeMetric(
+						WithFamily("catalogsource_ready"),
+						WithName(name),
+						WithNamespace(testNamespace),
+						WithValue(0),
+					)),
+				))
+				Consistently(func() []Metric {
+					return getMetricsFromPod(c, getPodWithLabel(c, "app=catalog-operator"), "8081")
+				}, "3m").Should(And(
```
timflannagan, line 367: @anik120 Any idea on why "3m" was chosen as the consistently poll here? I was poking around this package trying to rundown some flakes and noticed specific test spec takes roughly 75-80% of the total runtime for this metrics e2e package.

anik120: @timflannagan I can't seem to remember why I chose 3m, but I think I wanted to prove out that the metrics with a particular value is emitted for a considerable amount of time. Looking back at it, I don't think `Eventually` with a `pollAfter` duration set is a bad idea either. That'll reduce the run time since I'm assuming the process will sleep till it's time to poll.

anik120: Actually, there are rules that depend on this metric being having a consistent value over a period of time, eg https://github.com/operator-framework/operator-marketplace/blob/master/manifests/12_prometheus_rule.yaml#L23-L24. We could reduce the value to 2m too if that'll help bring that percentage down to ~50%

anik120: Opened a PR that's a middle ground solution (I think) https://github.com/operator-framework/operator-lifecycle-manager/pull/2739

Issue #2091, opened 2021-04-08
Flake: TestSyncOperatorGroups/AllNamespaces/CSVPresent/InstallModeNotSupported


benluddy:
```INFO[2021-04-08T19:18:51Z] --- FAIL: TestSyncOperatorGroups (1.90s)     
INFO[2021-04-08T19:18:51Z]     --- FAIL: TestSyncOperatorGroups/AllNamespaces/CSVPresent/InstallModeNotSupported (0.17s) 
INFO[2021-04-08T19:18:51Z]         operator_test.go:4477:               
INFO[2021-04-08T19:18:51Z]             	Error Trace:	operator_test.go:4477 
INFO[2021-04-08T19:18:51Z]             	Error:      	Received unexpected error: 
INFO[2021-04-08T19:18:51Z]             	            	clusterroles.rbac.authorization.k8s.io "csv-role" already exists 
INFO[2021-04-08T19:18:51Z]             	Test:       	TestSyncOperatorGroups/AllNamespaces/CSVPresent/InstallModeNotSupported
```

Seen this a couple times now, most recently in https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/operator-framework_operator-lifecycle-manager/2090/pull-ci-operator-framework-operator-lifecycle-manager-master-unit/1380237274532286464#1:build-log.txt%3A969.

benluddy:
Another one: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/operator-framework_operator-lifecycle-manager/2104/pull-ci-operator-framework-operator-lifecycle-manager-master-unit/1384220564280840192#1:build-log.txt%3A1064.

benluddy:
Seen downstream here: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_operator-framework-olm/62/pull-ci-openshift-operator-framework-olm-master-unit-olm/1384544819082170368#1:build-log.txt%3A1341.

benluddy:
Saw a new failure in https://github.com/operator-framework/operator-lifecycle-manager/runs/3243630835?check_suite_focus=true#step:5:1356:

```
--- FAIL: TestSyncOperatorGroups (1.42s)
    --- FAIL: TestSyncOperatorGroups/AllNamespaces/CSVPresent/InstallModeNotSupported (0.13s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
[...]

PR #2153, opened 2021-05-10
Remove the root list of YAML manifests stored in the manifests directory

### LINKED ISSUE #2130: Remove the root manifests directory and OCP/downstream helm charts
The root manifests are no longer needed now that a new repository was created to house the downstream OCP builds, and this repository was decoupled from the OCP release cycle.

We can similarly investigate removing the OCP-related helm charts entirely from upstream.

timflannagan:
Partially fixes #2130 

Remove the root list of YAML manifests stored in manifests/

The root manifest directory was needed by the openshift cluster-version-operator (CVO) to deploy OLM in OCP cluster
environments. During the 4.8 release, the operator-framework/operator-lifecycle-manager repository was decoupled from the OCP release cadence, and a downstream repository was created in the openshift organization. As a result of this effort, this list of manifests is no longer needed and instead will live in the downstream repository.


benluddy:

hasbro17 (commented):
/lgtm

PR #2154, opened 2021-05-10
Deprecated inner channel entries cause resolution error.

benluddy:
Similar to https://github.com/operator-framework/operator-lifecycle-manager/pull/2104, but this one happens because deprecated bundles are filtered out before they are returned from the catalog query layer: https://github.com/operator-framework/operator-lifecycle-manager/blob/f40e8e5a92f216aa5b84e7d7097b5221b925e43b/pkg/controller/registry/resolver/cache.go#L404


ecordell (approved):
/lgtm

PR #2155, opened 2021-05-10
Emit metrics on deprecated APIs dynamically

exdx:

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1958296

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/adoption_controller.go
+++ b/pkg/controller/operators/adoption_controller.go
@@ -27,7 +27,9 @@ import (
 	operatorsv1 "github.com/operator-framework/api/pkg/operators/v1"
 	operatorsv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/operators/decorators"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/deprecated"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/ownerutil"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/metrics"
 )
 
 // AdoptionReconciler automagically associates Operator components with their respective operator resource.
@@ -235,6 +237,14 @@ func (r *AdoptionReconciler) adoptComponents(ctx context.Context, csv *operators
 			}()
 		}
 
+		// alert on components that are deprecated in upcoming kubernetes versions
+		for _, component := range components {
+			if deprecated.Is(component.GetObjectKind().GroupVersionKind()) {
+				metrics.EmitComponentDeprecatedWarning("deprecation warning for operator component",
+					operator.Name, component.GetObjectKind().GroupVersionKind().String())
+			}
+		}
+
 		for _, component := range components {
 			var (
 				// Copy variables into iteration scope
@@ -336,7 +346,7 @@ func (r *AdoptionReconciler) disownFromAll(ctx context.Context, component runtim
 		}
 		operators = append(operators, *operator)
 	}
-	errs := make([]error,0)
+	errs := make([]error, 0)
 	for _, operator := range operators {
 		if err := r.disown(ctx, &operator, component); err != nil {
 			errs = append(errs, err)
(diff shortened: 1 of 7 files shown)


timflannagan:
I'm going to remove the "Bug <...>" from the title so the BZ automation doesn't pick this up before a downstream PR has been created. I'll update your description to add a link to the BZ instead.

/retitle Emit metrics on deprecated APIs dynamically

pkg/controller/operators/catalog/step.go
```diff
@@ -84,6 +88,10 @@ func (b *builder) create(step v1alpha1.Step) (Stepper, error) {
 		case crdlib.V1Version:
 			return b.NewCRDV1Step(b.opclient.ApiextensionsInterface().ApiextensionsV1(), &step, manifest), nil
 		case crdlib.V1Beta1Version:
+			// Emit error about deprecated API
+			// This codepath will be removed entirely in 1.22
+			metrics.EmitInstallPlanWarning(v1beta1DeprecationWarning, step.Resource.Name, "",
```
benluddy, line 93: Why is this special case necessary?

exdx: We install CRDs with OLM's client and the scoped client RoundTripper would not apply

pkg/lib/scopedclient/clientfactory.go
```diff
@@ -26,6 +26,7 @@ type Factory struct {
 // token specified.
 func (f *Factory) NewOperatorClient(token string) (client operatorclient.ClientInterface, err error) {
 	scoped := copy(f.config, token)
+	scoped = SetWarningRoundTripper(scoped)
```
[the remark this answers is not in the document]
exdx: Yep, had the same idea 👍

pkg/metrics/metrics.go
```diff
@@ -182,6 +184,14 @@ var (
 		[]string{Outcome},
 	)
 
+	installPlanWarningCount = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "install_plan_warnings_count",
+			Help: "monotonic count of resources created with warnings (for example, deprecation warnings)",
+		},
+		[]string{WARNING_LABEL, NAME_LABEL, NAMESPACE_LABEL, GVK_LABEL},
```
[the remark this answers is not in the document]
exdx: It's certainly a concern, but since we are just alerting here on deprecated objects on creation I'm wondering where the cardinality would truly explode? Is it cases where the installplan includes additional subscriptions from dependency resolution and you basically have an N dimensional install 

I was looking to embed things like the installplan name or even better the operator package in the WarningRoundTripper struct, instantiate it in the `ExcutePlan()` step with all the info, and then pass it down when creating the clients

pkg/lib/scopedclient/warning.go
```diff
@@ -0,0 +1,53 @@
...
+type WarningRoundTripper struct{}
+
+func SetWarningRoundTripper(config *rest.Config) *rest.Config {
+	if config != nil {
+		config.Transport = new(WarningRoundTripper)
+	}
+	return config
+}
+
+func (w *WarningRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
```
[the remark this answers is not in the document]
exdx: Ah, I was hoping we could do that. But it makes sense, since this is just part of the path of the response, we should not read or close it directly (we could copy the data and do whatever on that though). 

I can checkout the idea of reversing the URL to get the resource info (not to unmarshal the response directly). 

Putting code into the stepper would work too, but feels like quite a bit of complexity

pkg/controller/operators/adoption_controller.go
```diff
@@ -235,6 +237,14 @@ func (r *AdoptionReconciler) adoptComponents(ctx context.Context, csv *operators
 			}()
 		}
 
+		// alert on components that are deprecated in upcoming kubernetes versions
```
kevinrizza, line 240: Do we need to emit this on every reconciliation? Are we potentially going to start spewing metrics far too frequently in some cases? Can we batch these somehow?

I'm wondering if we should be aggregating this kind of metric and have a separate process that emits them.

exdx: Yup, I realized this was an issue with this implementation -- we probably need to move it around somewhere. 

Aggregating the metric sounds interest as well, so as not to explode the cardinality?

exdx:
Closing in favor of another PR.

PR #2158, opened 2021-05-12
.github/workflows: Separate out the e2e jobs into workflows

timflannagan:
Update the .github/workflows/test-scripts.yml GH action workflow and
move all the jobs listed there into separate workflows. This allows us
to retest a single workflow at a time as a workaround to GH not allowing
you to retest individual jobs in a workflow.

**Reviewer Checklist**


.github/workflows/build.yml
```diff
@@ -0,0 +1,19 @@
+name: build
+on:
+  pull_request:
+    paths:
+    - '**'
+    - '!doc/contributors/**'
+    - '!doc/design/**'
```
timflannagan, line 7: This still needs to be updated to handle the periodic testing case if I'm reading the following section in (the now deleted) test-scripts.yml workflow definition:

```
  schedule:
  - cron: '0 0 * * *' # daily to pick up releases
```

/hold

timflannagan:
@exdx This is probably a good time to revisit how we're naming these workflow/jobs instead of that other PR I posted last week. I think removing the upstream-* prefix from anything is a step in the right direction - any opinion on naming conventions, besides that?

exdx:
> @exdx This is probably a good time to revisit how we're naming these workflow/jobs instead of that other PR I posted last week. I think removing the upstream-* prefix from anything is a step in the right direction - any opinion on naming conventions, besides that?

This looks great @timflannagan. The other thing I'd like to change is the duplicate name as I brought up in the prior PR, things like `build/build` and `unit/unit` -- could we have those under a `test/*` or `ci/*` action instead?

njhale (approved):
Thanks for cleaning this up.

/approve

dinhxuanvu (approved):
/lgtm

PR #2160, opened 2021-05-13
Add OperatorCondition status sync and update operator upgradeable check

dinhxuanvu:
OperatorCondition controller will update its status to refect the
changes on its spec regarding the operator conditions that are reported
by the operators themselves. In turn, operators can read the status
to confirm if OLM has processed the spec changes.

OLM will not take actions on upgradeable condition if the status is
stale using ObservedGeneration and Generation check.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1927340.

**Reviewer Checklist**


timflannagan:
The BZ bot will automatically move the BZ state to POST once this PR lands but we only want that behavior for downstream PRs. I'm going to add a link to the BZ in your PR description instead.

/retitle Add OperatorCondition status sync and update operator upgradeable check

dinhxuanvu:
The proposal: https://hackmd.io/9wG20hu5TU-y1HrkhvcsZQ?view FYI

test/e2e/operator_condition_e2e_test.go
```diff
@@ -22,11 +22,11 @@ var _ = Describe("Operator Condition", func() {
 		TearDown(testNamespace)
 	})
 
-	It("OperatorCondition Upgradeable type and overrides", func() {
+	FIt("OperatorCondition Upgradeable type and overrides", func() {
```
awgreene, line 25:
```suggestion
	It("OperatorCondition Upgradeable type and overrides", func() {
```

awgreene (changes requested):
Please remove the `FIT` test, looks good otherwise.

dinhxuanvu:
@ecordell PTAL

pkg/controller/operators/operatorcondition_controller.go
```diff
@@ -252,6 +253,27 @@ func (r *OperatorConditionReconciler) ensureDeploymentEnvVars(operatorCondition
 	return nil
 }
 
+func (r *OperatorConditionReconciler) syncOperatorConditionStatus(operatorCondition *operatorsv2.OperatorCondition) error {
+	r.log.V(4).Info("Sync operatorcondition status")
+	currentGen := operatorCondition.ObjectMeta.GetGeneration()
+	changed := false
+	for _, cond := range operatorCondition.Spec.Conditions {
+		if c := meta.FindStatusCondition(operatorCondition.Status.Conditions, cond.Type); c != nil {
+			if cond.Status == c.Status && c.ObservedGeneration == currentGen {
```
awgreene, line 262: Why do we check that `cond.Status == c.Status` here?

pkg/controller/operators/operatorcondition_controller.go
```diff
@@ -119,27 +120,27 @@ func (r *OperatorConditionReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		return ctrl.Result{Requeue: true}, err
 	}
 
+	err = r.syncOperatorConditionStatus(operatorCondition)
+	if err != nil {
+		log.V(1).Error(err, "Error syncing OperatorCondition Status")
+		return ctrl.Result{Requeue: true}, err
```
awgreene, line 126: Wouldn't this be requeued because of the error?

pkg/lib/ownerutil/util.go
```diff
@@ -394,10 +395,10 @@ func InferGroupVersionKind(obj runtime.Object) error {
 			Version: operatorsv1.GroupVersion.Version,
 			Kind:    "OperatorGroup",
 		})
-	case *operatorsv1.OperatorCondition:
+	case *operatorsv2.OperatorCondition:
```
awgreene, line 398: I believe this change is fine, but wanted to invite others to consider possible scenarios where OLM would be unable to infer the GVK for the v1 operator condition. I don't believe it is possible to OLM to hit this scenario because all getters are at v2.

test/e2e/operator_condition_e2e_test.go
```diff
@@ -134,34 +117,39 @@ var _ = Describe("Operator Condition", func() {
...
 			Status:             metav1.ConditionTrue,
 			Reason:             "test",
 			Message:            "test",
 			LastTransitionTime: metav1.Now(),
 		}
 		Eventually(func() error {
-			cond, err = crc.OperatorsV1().OperatorConditions(testNamespace).Get(context.TODO(), csvA.GetName(), metav1.GetOptions{})
-			if err != nil {
+			cond, err = crc.OperatorsV2().OperatorConditions(testNamespace).Get(context.TODO(), csvA.GetName(), metav1.GetOptions{})
+			if err != nil || currentGen == cond.ObjectMeta.GetGeneration() {
```
awgreene, line 128: Is this just to check that the last update worked? It seems like we don't need this given that the spec was successfully earlier.

test/e2e/operator_condition_e2e_test.go
```diff
@@ -134,34 +117,39 @@ var _ = Describe("Operator Condition", func() {
...
-			if err != nil {
+			cond, err = crc.OperatorsV2().OperatorConditions(testNamespace).Get(context.TODO(), csvB.GetName(), metav1.GetOptions{})
+			if err != nil || currentGen == cond.ObjectMeta.GetGeneration() {
 				return err
 			}
-			meta.SetStatusCondition(&cond.Status.Conditions, upgradeableFalseCondition)
-			_, err = crc.OperatorsV1().OperatorConditions(testNamespace).UpdateStatus(context.TODO(), cond, metav1.UpdateOptions{})
+			currentGen = cond.ObjectMeta.GetGeneration()
+			upgradeableTrueCondition.ObservedGeneration = currentGen + 1
+			meta.SetStatusCondition(&cond.Status.Conditions, upgradeableTrueCondition)
```
awgreene, line 151: Just to confirm: you chose to change the `cond.Status.Condition` entry because doing so in the Spec would just cause OLM to observe a new generation eh?

awgreene (changes requested):
A few nits, but it looks great!

test/e2e/operator_condition_e2e_test.go
dinhxuanvu: This is simply to ensure all of reconciliation of OperatorCondition is completed and we get the latest version. This is to reduce flakes that tend to happen with e2e with all processes running in parallel.

test/e2e/operator_condition_e2e_test.go
dinhxuanvu: Correct. Updating Status won't increment generation.

pkg/lib/ownerutil/util.go
dinhxuanvu: We can keep v1 here and simply add v2 if that's what we prefer.

pkg/controller/operators/operatorcondition_controller.go
dinhxuanvu: Correct. I don't see the problem here. If there is an internal error somewhere, we want to requeue to ensure we sync the status properly in the end. Otherwise, the status is not synced and the operator may be stuck in whatever state it may be.

pkg/controller/operators/operatorcondition_controller.go
dinhxuanvu: This is a strict comparison that will storm out any possible manipulation the status in the object.Status while maintaining the same ObservedGeneration. We can relax this by only comparing ObservedGeneration. That's totally fine. That will allow an escape hatch if something happens with OperationCondition that causes the generation out of sync.

test/e2e/operator_condition_e2e_test.go
awgreene: Typically I would go for a longer timeout, but this is fine.

pkg/controller/operators/operatorcondition_controller.go
awgreene: Sorry - I might not have been clear. Based on [this documentation](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/vendor/sigs.k8s.io/controller-runtime/pkg/reconcile/reconcile.go#L91-L92) I believe that you can just return an error to requeue the request.

awgreene:
Overall this all looks good to me, most of the points I raised are nits so I'll add the /approve label.

pkg/controller/operators/olm/operatorconditions.go
```diff
@@ -7,35 +7,45 @@ import (
...
+	})
+
 	// Check condition overrides
-	for _, override := range cond.Spec.Overrides {
-		if override.Type == operatorsv1.Upgradeable {
-			if override.Status == metav1.ConditionTrue {
-				return true, nil
-			}
-			return false, fmt.Errorf("The operator is not upgradeable: %s", override.Message)
+	if o := meta.FindStatusCondition(cond.Spec.Overrides, operatorsv2.Upgradeable); o != nil {
```
timflannagan, line 34: :+1:

pkg/controller/operators/operatorcondition_controller.go
timflannagan: Agreed, I would just return an error without specifying the `Requeue: true` here.

dinhxuanvu: Again, nothing wrong with being explicit either. This is borderline bikeshedding.

pkg/controller/operators/operatorcondition_controller.go
```diff
@@ -252,6 +253,27 @@ func (r *OperatorConditionReconciler) ensureDeploymentEnvVars(operatorCondition
...
+				continue
+			}
+		}
+		cond.ObservedGeneration = currentGen
+		meta.SetStatusCondition(&operatorCondition.Status.Conditions, cond)
+		changed = true
+	}
+
+	if changed {
+		return r.Client.Status().Update(context.TODO(), operatorCondition)
```
timflannagan, line 272: Do we need to also log that we're attempting to update the `operatorcondition`, maybe at a higher verbosity? I guess it's possible that any encountered error will be logged at the call site, but may help down the line for debuggability.

dinhxuanvu: There is a log line above this to indicate that we are "syncing operatorcondition status" at level 4. That's significant. Are you recommending to log more info on this area?

pkg/controller/operators/operatorcondition_controller.go
timflannagan: Fair enough - I don't necessarily disagree, it's just not as intuitive without diving into the vendor code. It looks like the error check is the first conditional check under the hood, and the requeue check is last in the chain, so this is fine with me. It's likely that conditional structure won't change in the future also.

pkg/controller/operators/operatorcondition_controller.go
timflannagan: I was more thinking aloud about whether we'd want to log at a lower verbosity that we encountered a status condition (or conditions) that requires firing off an update request.

PR #2162, opened 2021-05-17
Make ClusterOperator Available condition sticky.

benluddy:
The intermediate CSV phases (those other than Succeeded and Failed)
don't on their own provide availability information, so they should
not be used to make decisions about a corresponding ClusterOperator's
Available status condition.


hasbro17 (commented):
/lgtm

PR #2165, opened 2021-05-19
Filter secret names for registry pod's sa

hasbro17:
During the registry server sync the image pull secrets from the catalogsource's spec.secrets are passed
unfiltered to the serviceaccount for the registry pod.
```YAML
kind: CatalogSource
metadata:
  name: ditto-index
  namespace: default
spec:
  sourceType: grpc
  image: quay.io/haseeb_tariq/index-image:ditto-index-0.1
  secrets:
    - ""
```
Passing an empty string in the secrets list breaks serverside apply for the registry pod which gets created without the `metadata.managedFields` when it has an empty element in the `imagePullSecrets` list:
```YAML
kind: Pod
apiVersion: v1
metadata:
  generateName: ditto-index-
  ...
  # No managedFields on resource
spec:
  imagePullSecrets:
    - {}
    - name: ditto-index-dockercfg-djmz7
...
```

This prevents the registry pod from being promoted via the SSA client when
there is an update to the index image:
```console
    couldn't ensure registry server - error ensuring updated catalog source pod:
    : detected imageID change: error during update: failed to create manager for
    existing fields: failed to convert new object (/v1, Kind=Pod) to smd typed:
    .spec.imagePullSecrets: element 0: associative list with keys has an element
    that omits key field "name" (and doesn't have default value)
```

To fix this, the image pull secrets list is filtered for empty strings
before being set on the serviceaccount.

Fix for: https://bugzilla.redhat.com/show_bug.cgi?id=1945548

Similar to https://github.com/kubernetes-sigs/structured-merge-diff/issues/130

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/registry/reconciler/grpc.go
+++ b/pkg/controller/registry/reconciler/grpc.go
@@ -98,6 +98,9 @@ func (s *grpcCatalogSourceDecorator) ServiceAccount() *corev1.ServiceAccount {
 	blockOwnerDeletion := true
 	isController := true
 	for _, secretName := range s.CatalogSource.Spec.Secrets {
+		if secretName == "" {
+			continue
+		}
 		secrets = append(secrets, corev1.LocalObjectReference{Name: secretName})
 	}
 	return &corev1.ServiceAccount{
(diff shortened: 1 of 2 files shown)


joelanford:
/lgtm

Kudos on the description as well! Super helpful!

PR #2167, opened 2021-05-19
Explicitly set `readOnlyRootFilesystem: false` on created registry pods.

joelanford:
When running in environments that impose strict security (e.g.
OpenShift with unknown SCC defaults), it is important to be explicit
about the security context requirements of the pods we run to avoid
runtime failures due to missing permissions.

This commit explicitly sets `readOnlyRootFilesystem: false` because
the registry pods need to copy the registry pod's databile file to
a separate read-only copy.

Signed-off-by: Joe Lanford <joe.lanford@gmail.com>

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/registry/reconciler/reconciler.go
+++ b/pkg/controller/registry/reconciler/reconciler.go
@@ -4,13 +4,14 @@ package reconciler
 import (
 	"strings"
 
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
 	"github.com/operator-framework/api/pkg/operators/v1alpha1"
 	controllerclient "github.com/operator-framework/operator-lifecycle-manager/pkg/lib/controller-runtime/client"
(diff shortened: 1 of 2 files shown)


pkg/controller/registry/reconciler/reconciler.go
```diff
@@ -4,13 +4,14 @@ package reconciler
 import (
 	"strings"
 
+	v1 "k8s.io/api/core/v1"
```
timflannagan, line 7: I'd like to see us standardize these package import names and have all of these be `corev1` at some point.

I'll briefly hold in case someone else wants to review but these changes sound reasonable to me.

/lgtm
/hold

ankitathomas (approved):
/lgtm

timflannagan:
JK, I realized we already have an approval

/hold cancel

pkg/controller/registry/reconciler/reconciler.go
joelanford: I blindly use https://github.com/AanZee/goimportssort as a file watcher in Goland, so I don't pay much attention to the imports block. It probably wouldn't be too hard to write a simple check that looks for some known api packages and checks that the import name follows that pattern.

PR #2178, opened 2021-05-27
Add polling interval for default catalogsource

kevinrizza:
Update the operatorhub catalog template to include a polling interval of 15 minutes


exdx:
/hold
this seems to be flaking really hard for some reason...

PR #2180, opened 2021-06-02
Surface API server warnings when applying plan steps.

benluddy:
Responses from the API server may include RFC2616 14.46 warnings (for
example, due to deprecation). Since the catalog-operator creates and
updates cluster resources on behalf of users, these warnings should be
surfaced to users.

When at least one warning is returned during the creation or update of
a step resource, catalog-operator now creates an Event with reason
"AppliedWithWarnings", involvedObject and fieldPath set to the plan
and step that generated the warning, respectively, and its message
containing the warning text.

In addition, a new metric, "installplan_warnings_total", counts the
number of steps that have been applied with warnings.


test/e2e/installplan_e2e_test.go
```diff
@@ -60,6 +63,140 @@ var _ = Describe("Install Plan", func() {
...
+		)
+
+		BeforeEach(func() {
+			dc, err := discovery.NewDiscoveryClientForConfig(ctx.Ctx().RESTConfig())
+			Expect(err).ToNot(HaveOccurred())
+
+			v, err := dc.ServerVersion()
+			Expect(err).ToNot(HaveOccurred())
+
+			if minor, err := strconv.Atoi(v.Minor); err == nil && minor < 21 {
```
exdx, line 80: this check makes sense -- are you describing that in 1.25 the API under deprecation notice would potentially be removed entirely, so these tests would not be relevant in 1.25+? Should we do `21 < minor < 25` in that case?

test/e2e/installplan_e2e_test.go
```diff
@@ -60,6 +63,140 @@ var _ = Describe("Install Plan", func() {
...
+			for _, metric := range getMetricsFromPod(ctx.Ctx().KubeClient(), getPodWithLabel(ctx.Ctx().KubeClient(), "app=catalog-operator"), "8081") {
+				if metric.Family == "installplan_warnings_total" {
+					counter = metric.Value
+				}
+			}
+
+			csv := newCSV("test-csv", testNamespace, "", semver.Version{}, nil, nil, nil)
+			Expect(ctx.Ctx().Client().Create(context.TODO(), &csv)).To(Succeed())
+
+			pdb = policyv1beta1.PodDisruptionBudget{
```
exdx, line 101: using a PDB makes sense. I'm wondering if it's worth adding a case for v1beta1 CRDs (since it's really what we care about) or if we are confident enough in the new codepath with a PDB in the installplan only? 

CRDs have some additional validation logic when applying to a cluster with the CRD already present -- stuff like schema validation, etc. If that fails, I don't think the alert would fire, which I believe is what we want.

pkg/lib/clients/transformers.go
```diff
@@ -0,0 +1,12 @@
+package clients
+
+import (
+	"k8s.io/client-go/rest"
+)
+
+func SetWarningHandler(wh rest.WarningHandler) ConfigTransformer {
+	return ConfigTransformerFunc(func(config *rest.Config) *rest.Config {
+		config.WarningHandler = wh
```
exdx, line 9: nit: should we check for a nil config here?

exdx (commented):
Really awesome, I like the `ConfigTransformer` and the new `Factory` interfaces. Very elegant solution. This is a quick review -- will take a look again.

pkg/lib/clients/transformers.go
benluddy: I updated the doc comment to indicate that the argument will never be nil and that implementations should never return nil.

test/e2e/installplan_e2e_test.go
benluddy: This assumes we're going to revisit this in 2021 as part of the Bundle API work in order to avoid doing the work required to maintain a test webhook or to add support for arbitrary CRs to InstallPlans. If not, then this test should fail and we will be forced to deal with it then.

benluddy:
I suspect https://github.com/operator-framework/operator-lifecycle-manager/pull/2183 will improve the end-to-end test flakiness. Doesn't seem related to this change.

benluddy:
@exdx I rebased to pick up a fix for some test flakiness, PTAL.

PR #2181, opened 2021-06-03
Registry ensure methods error out if resource is not found

awgreene:
Problem: When calling a registry's ensure methods, if the resource is
deleted between the get and delete calls the ensure method fails due
to a 404 error.

Solution: When deleting an object, ignore 404 errors.

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1967621

### DIFF
--- a/pkg/controller/registry/reconciler/grpc.go
+++ b/pkg/controller/registry/reconciler/grpc.go
@@ -234,7 +234,7 @@ func (c *GrpcRegistryReconciler) ensurePod(source grpcCatalogSourceDecorator, sa
 			return nil
 		}
 		for _, p := range currentLivePods {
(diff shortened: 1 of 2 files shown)


pkg/controller/registry/reconciler/grpc.go
```diff
@@ -11,6 +11,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/api/core/v1"
 	k8serror "k8s.io/apimachinery/pkg/api/errors"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
```
timflannagan, line 14: nit: there's a `k8error` declare above

awgreene: Thanks, removed it.

timflannagan:
Feel free to remove the hold. Just adding in case someone else wants to take a look.

/lgtm
/hold

kevinrizza:
/hold cancel

lgtm

PR #2183, opened 2021-06-04
Fix nil pointer dereference while reporting bundle unpack status.

benluddy:


benluddy:
All the spam is from /retest not affecting Github Actions plus https://github.com/operator-framework/operator-lifecycle-manager/issues/2091, not from 100 test failures.

PR #2190, opened 2021-06-08
Check for conflicting install

kevinrizza:
Update the install script so that known conflictingdownstream installations of OLM don't cause conflicts and hotloop

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -14,6 +14,12 @@ if [[ ${#@} -lt 1 || ${#@} -gt 2 ]]; then
     exit 1
 fi
 
+kubectl get deployment olm-operator -n openshift-operator-lifecycle-manager -o=jsonpath='{.spec}' > /dev/null 2>&1
+if [[ $? -eq 0 ]]; then
+    echo "OLM is already installed in a different configuration. This is common if you are not running a vanilla Kubernetes cluster. Exiting..."
+    exit 1
+fi
+
 release="$1"
 base_url="${2:-${default_base_url}}"
 url="${base_url}/${release}"


scripts/install.sh
```diff
@@ -14,6 +14,12 @@ if [[ ${#@} -lt 1 || ${#@} -gt 2 ]]; then
     exit 1
 fi
 
+openshift_install_check=$(kubectl get deployment olm-operator -n openshift-operator-lifecycle-manager -o=jsonpath='{.spec}' 2>/dev/null)
+if [[ $openshift_install_check != "" ]]; then
```
timflannagan, line 18: You probably can just use the `z` flag to determine if this variable is empty (and/or set) but this looks fine to me.

I'll hold in case anyone else has any comments.

/lgtm
/hold

scripts/install.sh
```diff
@@ -14,6 +14,12 @@ if [[ ${#@} -lt 1 || ${#@} -gt 2 ]]; then
     exit 1
 fi
 
+openshift_install_check=$(kubectl get deployment olm-operator -n openshift-operator-lifecycle-manager -o=jsonpath='{.spec}' 2>/dev/null)
```
benluddy, line 17: Is the string test necessary? `kubectl get` should exit with a nonzero status code if it doesn't find the resource, IIRC.

scripts/install.sh
```diff
@@ -14,6 +14,12 @@ if [[ ${#@} -lt 1 || ${#@} -gt 2 ]]; then
     exit 1
 fi
 
+kubectl get deployment olm-operator -n openshift-operator-lifecycle-manager -o=jsonpath='{.spec}' > /dev/null 2>&1
```
joelanford, line 17: Check in all namespaces?

timflannagan: This seems more reasonable to me the more I think about it. Ideally, we don't have any downstream-specific knowledge embedded upstream, so checking the openshift-prefixed OLM namespace should be sufficient in the short term to prevent users from installing upstream OLM on managed clusters, but I think we'd like to attempt to start diving out downstream in the next couple of releases.

If we're querying for the olm-operator deployment in all of the namespaces, is there a chance we're casting a wide net here? I doubt there are user-defined deployments that match this name but I'm wondering if we need to also add a label selector.

jmazzitelli:
I think we can fix this via my suggestion here: https://github.com/operator-framework/operator-lifecycle-manager/issues/2210#issuecomment-865105374

PR #2205, opened 2021-06-11
Resurrect deleted metrics Service templates.

benluddy:
The olm-operator-metrics and catalog-operator-metrics services were
deleted as part of a removal of OpenShift-specific content, but they
are still necessary when monitoring is enabled.

### DIFF
--- a/deploy/chart/templates/0000_50_olm_02-services.yaml
+++ b/deploy/chart/templates/0000_50_olm_02-services.yaml
@@ -1,0 +1,39 @@
+{{ if .Values.monitoring.enabled }}
+apiVersion: v1
+kind: Service
+metadata:
+  name: olm-operator-metrics
+  namespace: {{ .Values.namespace }}
+  annotations:
+    service.alpha.openshift.io/serving-cert-secret-name: olm-operator-serving-cert
+  labels:
+    app: olm-operator
+spec:
+  type: ClusterIP
+  ports:
+  - name: https-metrics
+    port: 8081
+    protocol: TCP
+    targetPort: metrics
+  selector:
+    app: olm-operator
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: catalog-operator-metrics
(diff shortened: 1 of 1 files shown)


deploy/chart/templates/0000_50_olm_02-services.yaml
```diff
@@ -0,0 +1,39 @@
...
+  namespace: {{ .Values.namespace }}
+  annotations:
+    service.alpha.openshift.io/serving-cert-secret-name: olm-operator-serving-cert
+  labels:
+    app: olm-operator
+spec:
+  type: ClusterIP
+  ports:
+  - name: https-metrics
+    port: 8081
```
timflannagan, line 15: I need to double-check whether we expose a way to configure the metrics port (e.g. through flags) for the catalog and olm operator binaries. Do we want to update the monitoring helm chart value properties and add a section for configuring this port that defaults to 8081?

timflannagan: It looks like it's hardcoded in the olm operator main package. I think exposing a configuration knob makes sense upstream, but it's out of scope for this PR anyways.

deploy/chart/templates/0000_50_olm_02-services.yaml
```diff
@@ -0,0 +1,39 @@
+{{ if .Values.monitoring.enabled }}
+apiVersion: v1
+kind: Service
+metadata:
+  name: olm-operator-metrics
+  namespace: {{ .Values.namespace }}
+  annotations:
+    service.alpha.openshift.io/serving-cert-secret-name: olm-operator-serving-cert
```
timflannagan, line 8: Do we want to remove any OCP-specific annotations? IIRC, this annotation has graduated to beta as well.

benluddy: I think so, yes. There's some other work that has to happen to make enabling metrics work out-of-the-box on upstream though.

PR #2213, opened 2021-06-23
Bug 1952576: csv_succeeded metric not present

josefkarasek:
`csv_succeeded` metric is lost between pod restarts.
This is because this metric is only emitted when CSV.Status is changed.

Emit `csv_succeeded`/`csv_abnormal` metric during every CSV sync loop.

`csv_succeeded` metric is lost between pod restarts.

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/olm/operator.go
+++ b/pkg/controller/operators/olm/operator.go
@@ -1115,10 +1115,11 @@ func (a *Operator) syncClusterServiceVersion(obj interface{}) (syncError error) 
 			} else {
 				syncError = fmt.Errorf("error transitioning ClusterServiceVersion: %s and error updating CSV status: %s", syncError, updateErr)
 			}
-		} else {
-			metrics.EmitCSVMetric(clusterServiceVersion, outCSV)
 		}
 	}
+
+	// always emit csv metrics
+	metrics.EmitCSVMetric(clusterServiceVersion, outCSV)
 
 	operatorGroup := a.operatorGroupFromAnnotations(logger, clusterServiceVersion)
 	if operatorGroup == nil {


pkg/controller/operators/olm/operator.go
```diff
@@ -1115,11 +1115,12 @@ func (a *Operator) syncClusterServiceVersion(obj interface{}) (syncError error)
 			} else {
 				syncError = fmt.Errorf("error transitioning ClusterServiceVersion: %s and error updating CSV status: %s", syncError, updateErr)
 			}
-		} else {
-			metrics.EmitCSVMetric(clusterServiceVersion, outCSV)
 		}
 	}
 
+	// always emit csv metrics
+	metrics.EmitCSVMetric(clusterServiceVersion, outCSV)
```
timflannagan, line 1122: Are there any cardinality concerns with always emitting CSV metrics?

josefkarasek: I think the metric is using a good, qualified name, which is always unique for one CSV
```
csv_succeeded{name="etcdoperator.v0.9.4",namespace="olm",version="0.9.4"} 1
```

timflannagan: Agreed, that doesn't look crazy to me, but it looks like we're emitting other metrics besides that csv_succeeded one when I was poking around that metrics package - I'm trying to wrap my head around how all those CSV-related metrics, emitting on a per-step basis, can lead to problems affecting the core monitoring stack.

timflannagan: Ah, I found https://github.com/operator-framework/operator-lifecycle-manager/pull/1099 - @awgreene any idea on whether we can safely emit metrics on a per-step basis now?

josefkarasek: Other approach to fixing this bug could be to emit the metric for all CSVs during pod startup and then update it only when a change happens

josefkarasek:
How can I fix `bugzilla/invalid-bug` label on the PR?

timflannagan:
@josefkarasek The bug bot is complaining about the current state of the BZ: https://github.com/operator-framework/operator-lifecycle-manager/pull/2213#issuecomment-866827324:

```
expected the bug to target the "4.9.0" release, but it targets "---" instead
```

In order to fix this, update the BZ's "Target Release" dropdown button to the 4.9.0, instead of the default value (empty release, "---"), save the BZ, and then comment `/bugzilla refresh` here.

pkg/controller/operators/olm/operator.go
anik120: @timflannagan I don't think there's any cardinality concern here. `csv_succeeded` is a prometheus gauge, and within `EmitCSVMetrics` we're always first deleting the old metric for the csv being synced, and then emitting a new metric and setting the gauge value to 1 or 0 (succeeded/did not succeeded). Even if we were not deleting the old metric, iirc metric points for a unique set of values are only emitted once, i.e they're always unique data points in the set of emitted metrics, which is what @josefkarasek clarified in the first comment.  

However, I'm not convinced this actually solves the problem. @josefkarasek the original problem was that we were only edge triggering this metric, i.e whenever the controller syncs a clusterserviceversion (`syncClusterServiceVersion` holds the logic for when that does happen), and that happens only when there's a change in the CSV object on cluster. But we need some way to level drive this metric too, which is what the first part of your last comment is. 

> update it only when a change happens
edge triggered

> emit the metric for all CSVs during pod startup 
level driven

anik120 (changes requested):
We'd also want to add test for this as proof of concept.

anik120:
Also, @timflannagan we don't need the bug number in the PR title right? We'll only need it when we downstream the PR?

pkg/controller/operators/olm/operator.go
josefkarasek: I'm assuming that all CSVs are queued up during pod start and reconciled. My assumption is that this approach is edge+level driven at the same time. From what you're saying it sounds like this assumption doesn't hold.

anik120: Although CSVs are queued up during pod start, that is still edge trigger, the trigger here being queuing of the CSV. True level driven is when you query the state of the cluster and attempt to reconcile with the desired state. There's always that chance with edge triggers that we'll miss an event, so querying for existing CSVs and emitting metrics for them on pod restart is the most full proof method to solve this problem.

josefkarasek:
Closing in favor of #2216

PR #2214, opened 2021-06-23
Resolve InstallPlanStepAppliedWithWarnings alert after some time.

benluddy:
There is an error in the alert expression that causes the alert to
persist indefinitely. The intention of the alert was to fire for
several minutes to inform the administration of the warnings. The
appropriate response to the alert is to stop installing problematic
bundles going forward.


timflannagan:
Feel free to remove the hold.

/lgtm
/hold

benluddy:

PR #2215, opened 2021-06-24
(fix)InstallPlan: Do not tranisition IP to failed on OG/SA failure

### LINKED ISSUE #2207: invalid operator group - no operator group found that is managing this namespace
Install plan fails with the following error:

`   Message:               invalid operator group - no operator group found that is managing this namespace      
`

I got an operator group for the namespace `istio-system`:

```
Name:         operator-group
Namespace:    istio-system
Labels:       gitops.jenkins-x.io/pipeline=namespaces
Annotations:  meta.helm.sh/release-name: istio-system
API Version:  operators.coreos.com/v1
Kind:         OperatorGroup
Metadata:
  Creation Timestamp:  2021-06-15T07:58:41Z
  Generation:          1
  Managed Fields:
    API Version:  operators.coreos.com/v1
    Fields Type:  FieldsV1
    fieldsV1:
      f:metadata:
        f:annotations:
          .:
          f:kubectl.kubernetes.io/last-applied-configuration:
          f:meta.helm.sh/release-name:
        f:labels:
          .:
          f:gitops.jenkins-x.io/pipeline:
      f:spec:
        .:
        f:targetNamespaces:
          .:
          v:"istio-system":
    Manager:      kubectl
    Operation:    Update
    Time:         2021-06-15T07:58:41Z
    API Version:  operators.coreos.com/v1
    Fields Type:  FieldsV1
    fieldsV1:
      f:status:
        .:
        f:lastUpdated:
        f:namespaces:
          .:
          v:"istio-system":
    Manager:         olm
    Operation:       Update
    Time:            2021-06-15T08:12:37Z
  Resource Version:  115903077
  UID:               04cba808-9a2e-487b-ba0d-959b1b83cde4
Spec:
  Target Namespaces:
    istio-system
Status:
  Last Updated:  2021-06-15T08:12:37Z
  Namespaces:
    istio-system
Events:  <none>
```

and here the installation plan:

```
Name:         install-lq52r
Namespace:    istio-system
Labels:       operators.coreos.com/kiali.istio-system=
Annotations:  <none>
API Version:  operators.coreos.com/v1alpha1
Kind:         InstallPlan
Metadata:
  Creation Timestamp:  2021-06-15T07:40:37Z
  Generate Name:       install-
  Generation:          1
  Managed Fields:
    API Version:  operators.coreos.com/v1alpha1
    Fields Type:  FieldsV1
    fieldsV1:
      f:metadata:
        f:generateName:
        f:ownerReferences:
          .:
          k:{"uid":"91870dd8-17c6-4ba9-a05a-c714eeae3ab8"}:
            .:
            f:apiVersion:
            f:blockOwnerDeletion:
            f:controller:
            f:kind:
            f:name:
            f:uid:
      f:spec:
        .:
        f:approval:
        f:approved:
        f:clusterServiceVersionNames:
        f:generation:
      f:status:
        .:
        f:bundleLookups:
        f:catalogSources:
        f:conditions:
        f:phase:
    Manager:      catalog
    Operation:    Update
    Time:         2021-06-15T07:40:37Z
    API Version:  operators.coreos.com/v1alpha1
    Fields Type:  FieldsV1
    fieldsV1:
      f:metadata:
        f:labels:
          .:
          f:operators.coreos.com/kiali.istio-system:
    Manager:    olm
    Operation:  Update
    Time:       2021-06-15T08:12:44Z
  Owner References:
    API Version:           operators.coreos.com/v1alpha1
    Block Owner Deletion:  false
    Controller:            false
    Kind:                  Subscription
    Name:                  istio-kiali
    UID:                   91870dd8-17c6-4ba9-a05a-c714eeae3ab8
  Resource Version:        115903226
  UID:                     c486e04f-838d-4ed9-9f9c-c94659467283
Spec:
  Approval:  Automatic
  Approved:  true
  Cluster Service Version Names:
    kiali-operator.v1.36.0
  Generation:  1
Status:
  Bundle Lookups:
    Catalog Source Ref:
      Name:       operatorhubio-catalog
      Namespace:  operator-lifecycle-manager
    Conditions:
      Message:   bundle contents have not yet been persisted to installplan status
      Reason:    BundleNotUnpacked
      Status:    True
      Type:      BundleLookupNotPersisted
      Message:   unpack job not yet started
      Reason:    JobNotStarted
      Status:    True
      Type:      BundleLookupPending
    Identifier:  kiali-operator.v1.36.0
    Path:        quay.io/operatorhubio/kiali:v1.36.0
    Properties:  {"properties":[{"type":"olm.gvk","value":{"group":"kiali.io","kind":"Kiali","version":"v1alpha1"}},{"type":"olm.gvk","value":{"group":"monitoring.kiali.io","kind":"MonitoringDashboard","version":"v1alpha1"}},{"type":"olm.package","value":{"packageName":"kiali","version":"1.36.0"}}]}
    Replaces:    kiali-operator.v1.35.0
  Catalog Sources:
  Conditions:
    Last Transition Time:  2021-06-15T07:40:37Z
    Last Update Time:      2021-06-15T07:40:37Z
    Message:               invalid operator group - no operator group found that is managing this namespace
    Reason:                InstallCheckFailed
    Status:                False
    Type:                  Installed
  Phase:                   Failed
Events:                    <none>
```
I am with kubernetes 1.18.2

anik120:

In #2077, a new phase `Failed` was introduced for InstallPlans, and failure in
detecting a valid OperatorGroup(OG) or a Service Account(SA) for the namespace
the InstallPlan was being created in would transition the InstallPlan to the
`Failed` state, i.e failure to detected these resources when the InstallPlan was
reconciled the first time was considered a permanant failure. This is a regression
from the previous behavior of InstallPlans where failure to detect OG/SA would
requeue the InstallPlan for reconciliation, so creating the required resources before
the retry limit of the informer queue was reached would transition the InstallPlan
from the `Installing` phase to the `Complete` phase(unless the bundle unpacking step
failed, in which case #2093 introduced transitioning the InstallPlan to the `Failed`
phase).

This regression introduced oddities for users who has infra built that applies a
set of manifests simultaneously to install an operator that includes a Subscription to
an operator (that creates InstallPlans) along with the required OG/SAs. In those cases,
whenever there was a delay in the reconciliation of the OG/SA, the InstallPlan would
be transitioned to a state of permanant faliure.

This PR:
* Removes the logic that transitioned the InstallPlan to `Failed`. Instead, the
InstallPlan will again be requeued for any reconciliation error.

* Introduces logic to bubble up reconciliation error through the InstallPlan's
status.Conditions, eg:

When no OperatorGroup is detected:

```
conditions:
    - lastTransitionTime: "2021-06-23T18:16:00Z"
      lastUpdateTime: "2021-06-23T18:16:16Z"
      message: attenuated service account query failed - no operator group found that
        is managing this namespace
      reason: InstallCheckFailed
      status: "False"
      type: Installed
```

Then when a valid OperatorGroup is created:

```
conditions:
    - lastTransitionTime: "2021-06-23T18:33:37Z"
      lastUpdateTime: "2021-06-23T18:33:37Z"
      status: "True"
      type: Installed
```

**Reviewer Checklist**

Closes #2207

### DIFF
--- a/doc/design/architecture.md
+++ b/doc/design/architecture.md
@@ -102,11 +102,12 @@ None --> Planning +------>------->------> Installing +---> Complete
 | Phase            | Description                                                                                    |
 |------------------|------------------------------------------------------------------------------------------------|
 | None             | initial phase, once seen by the Operator, it is immediately transitioned to `Planning`         |
-| Planning         | dependencies between resources are being resolved, to be stored in the InstallPlan `Status` |
+| Planning         | dependencies between resources are being resolved, to be stored in the InstallPlan `Status`    |
 | RequiresApproval | occurs when using manual approval, will not transition phase until `approved` field is true    |
-| Installing       | resolved resources in the InstallPlan `Status` block are being created                      |
+| Installing       | waiting for reconciliation of required resource(OG/SA etc), or resolved resources in the 
+                     InstallPlan `Status` block are being created                                                   |
 | Complete         | all resolved resources in the `Status` block exist                                             |
-| Failed           | occurs when resources fail to install or there is an invalid OperatorGroup                     |
+| Failed           | occurs when resources fail to install or when bundle unpacking fails                           |
 
 ### Subscription Control Loop
 
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -1387,28 +1387,36 @@ func (o *Operator) syncInstallPlans(obj interface{}) (syncError error) {
 
 	querier := o.serviceAccountQuerier.NamespaceQuerier(plan.GetNamespace())
 	ref, err := querier()
+
+	out := plan.DeepCopy()
 	if err != nil {
-
-		// Retry sync if non-fatal error
-		if !scoped.IsOperatorGroupError(err) {
-			syncError = fmt.Errorf("attenuated service account query failed - %v", err)
-			return
-		}
-
-		// Mark the InstallPlan as failed for a fatal Operator Group related error
-		logger.Infof("attenuated service account query failed - %v", err)
-		ipFailError := fmt.Errorf("invalid operator group - %v", err)
-
-		if err := o.transitionInstallPlanToFailed(plan, logger, v1alpha1.InstallPlanReasonInstallCheckFailed, ipFailError.Error()); err != nil {
-			// retry for failure to update status
+		// Set status condition/message and retry sync if any error
+		ipFailError := fmt.Errorf("attenuated service account query failed - %v", err)
+		logger.Infof(ipFailError.Error())
+		_, err := o.setInstallPlanInstalledCond(out, v1alpha1.InstallPlanReasonInstallCheckFailed, err.Error(), logger)
+		if err != nil {
 			syncError = err
 			return
 		}
-
-		// Requeue subscription to propagate SubscriptionInstallPlanFailed condtion to subscription
-		o.requeueSubscriptionForInstallPlan(plan, logger)
-
+		syncError = ipFailError
 		return
+	} else {
+		// reset condition/message if it had been set in previous sync. This condition is being reset since any delay in the next steps
+		// (bundle unpacking/plan step errors being retried for a duration) could lead to this condition sticking around, even after
+		// the serviceAccountQuerier returns no error since the error has been resolved (by creating the required resources), which would
+		// be confusing to the user
+
+		// NOTE: this makes the assumption that the InstallPlanInstalledCheckFailed reason is only set in the previous if clause, which is
+		// true in the current iteration of the catalog operator. Any future implementation change that aims at setting the reason as
+		// InstallPlanInstalledCheckFailed must make sure that either this assumption is not breached, or the condition being set elsewhere
+		// is not being unset here unintentionally.
+		if cond := out.Status.GetCondition(v1alpha1.InstallPlanInstalled); cond.Reason == v1alpha1.InstallPlanReasonInstallCheckFailed {
+			plan, err = o.setInstallPlanInstalledCond(out, v1alpha1.InstallPlanConditionReason(corev1.ConditionUnknown), "", logger)
+			if err != nil {
+				syncError = err
+				return
+			}
+		}
 	}
 
 	if ref != nil {
@@ -1551,6 +1559,18 @@ func (o *Operator) requeueSubscriptionForInstallPlan(plan *v1alpha1.InstallPlan,
 			logger.WithError(err).Warn("error requeuing installplan owner")
 		}
 	}
+}
+
+func (o *Operator) setInstallPlanInstalledCond(ip *v1alpha1.InstallPlan, reason v1alpha1.InstallPlanConditionReason, message string, logger *logrus.Entry) (*v1alpha1.InstallPlan, error) {
+	now := o.now()
+	ip.Status.SetCondition(v1alpha1.ConditionFailed(v1alpha1.InstallPlanInstalled, reason, message, &now))
+	outIP, err := o.client.OperatorsV1alpha1().InstallPlans(ip.GetNamespace()).UpdateStatus(context.TODO(), ip, metav1.UpdateOptions{})
+	if err != nil {
+		logger = logger.WithField("updateError", err.Error())
+		logger.Errorf("error updating InstallPlan status")
+		return nil, nil
+	}
+	return outIP, nil
 }
 
 type installPlanTransitioner interface {
--- a/pkg/controller/operators/catalog/operator_test.go
+++ b/pkg/controller/operators/catalog/operator_test.go
@@ -166,12 +166,12 @@ func TestSyncInstallPlanUnhappy(t *testing.T) {
 	)
 
 	tests := []struct {
-		testName      string
-		err           error
-		in            *v1alpha1.InstallPlan
-		expectedPhase v1alpha1.InstallPlanPhase
-
-		clientObjs []runtime.Object
+		testName          string
+		err               error
+		in                *v1alpha1.InstallPlan
+		expectedPhase     v1alpha1.InstallPlanPhase
+		expectedCondition *v1alpha1.InstallPlanCondition
+		clientObjs        []runtime.Object
 	}{
 		{
 			testName:      "NoStatus",
@@ -180,18 +180,22 @@ func TestSyncInstallPlanUnhappy(t *testing.T) {
 			in:            installPlan("p", namespace, v1alpha1.InstallPlanPhaseNone),
 		},
 		{
-			// This checks that an installplan is marked as failed when no operatorgroup is present
+			// This checks that an installplan's status.Condition contains a condition with error message when no operatorgroup is present
 			testName:      "HasSteps/NoOperatorGroup",
-			err:           nil,
-			expectedPhase: v1alpha1.InstallPlanPhaseFailed,
-			in:            ipWithSteps,
+			err:           fmt.Errorf("attenuated service account query failed - no operator group found that is managing this namespace"),
+			expectedPhase: v1alpha1.InstallPlanPhaseInstalling,
+			expectedCondition: &v1alpha1.InstallPlanCondition{Type: v1alpha1.InstallPlanInstalled, Status: corev1.ConditionFalse, Reason: v1alpha1.InstallPlanReasonInstallCheckFailed,
+				Message: "no operator group found that is managing this namespace"},
+			in: ipWithSteps,
 		},
 		{
-			// This checks that an installplan is marked as failed when multiple operator groups are present for the same namespace
+			// This checks that an installplan's status.Condition contains a condition with error message when multiple operator groups are present for the same namespace
 			testName:      "HasSteps/TooManyOperatorGroups",
-			err:           nil,
-			expectedPhase: v1alpha1.InstallPlanPhaseFailed,
+			err:           fmt.Errorf("attenuated service account query failed - more than one operator group(s) are managing this namespace count=2"),
+			expectedPhase: v1alpha1.InstallPlanPhaseInstalling,
 			in:            ipWithSteps,
+			expectedCondition: &v1alpha1.InstallPlanCondition{Type: v1alpha1.InstallPlanInstalled, Status: corev1.ConditionFalse, Reason: v1alpha1.InstallPlanReasonInstallCheckFailed,
+				Message: "more than one operator group(s) are managing this namespace count=2"},
 			clientObjs: []runtime.Object{
 				operatorGroup("og1", "sa", namespace,
 					&corev1.ObjectReference{
@@ -208,11 +212,13 @@ func TestSyncInstallPlanUnhappy(t *testing.T) {
 			},
 		},
 		{
-			// This checks that an installplan is marked as failed when no service account is synced for the operator group, i.e the service account ref doesn't exist
+			// This checks that an installplan's status.Condition contains a condition with error message when no service account is synced for the operator group, i.e the service account ref doesn't exist
 			testName:      "HasSteps/NonExistentServiceAccount",
-			err:           nil,
-			expectedPhase: v1alpha1.InstallPlanPhaseFailed,
-			in:            ipWithSteps,
+			err:           fmt.Errorf("attenuated service account query failed - please make sure the service account exists. sa=sa1 operatorgroup=ns/og"),
+			expectedPhase: v1alpha1.InstallPlanPhaseInstalling,
+			expectedCondition: &v1alpha1.InstallPlanCondition{Type: v1alpha1.InstallPlanInstalled, Status: corev1.ConditionFalse, Reason: v1alpha1.InstallPlanReasonInstallCheckFailed,
+				Message: "please make sure the service account exists. sa=sa1 operatorgroup=ns/og"},
+			in: ipWithSteps,
 			clientObjs: []runtime.Object{
 				operatorGroup("og", "sa1", namespace, nil),
 			},
@@ -236,6 +242,11 @@ func TestSyncInstallPlanUnhappy(t *testing.T) {
 			require.NoError(t, err)
 
 			require.Equal(t, tt.expectedPhase, ip.Status.Phase)
+
+			if tt.expectedCondition != nil {
+				require.True(t, hasExpectedCondition(ip, *tt.expectedCondition))
+			}
+
 		})
 	}
 }
@@ -1829,4 +1840,13 @@ func operatorGroup(ogName, saName, namespace string, saRef *corev1.ObjectReferen
 			ServiceAccountRef: saRef,
 		},
 	}
+}
+
+func hasExpectedCondition(ip *v1alpha1.InstallPlan, expectedCondition v1alpha1.InstallPlanCondition) bool {
+	for _, cond := range ip.Status.Conditions {
+		if cond.Type == expectedCondition.Type && cond.Message == expectedCondition.Message && cond.Status == expectedCondition.Status {
+			return true
+		}
+	}
+	return false
 }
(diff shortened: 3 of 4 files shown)


anik120:
Note: Removed the [e2e test cases](https://github.com/operator-framework/operator-lifecycle-manager/pull/2215/files#diff-eddfb8217b77c532b776defca62[…]b8bf8a4d8de408df3a6a4L3148-L3303) that checked if the InstallPlan was being transitioned to the `Failed` state, instead of modifying them to checked if not state transition was happening, and if condition `Installed:False` condition is present because the [unit test cases](https://github.com/operator-framework/operator-lifecycle-manager/pull/2215/files#diff-1b127b54452c61c5f43789606e4[…]7acdfef97e3272b415f834dR169-R249) are doing the exact same thing, making the e2e test cases superfluous.

pkg/controller/operators/catalog/operator.go
```diff
@@ -1389,25 +1389,23 @@ func (o *Operator) syncInstallPlans(obj interface{}) (syncError error) {
...
-		ipFailError := fmt.Errorf("invalid operator group - %v", err)
+		// Set status condition/message and retry sync if any error
+		ipFailError := fmt.Errorf("attenuated service account query failed - %v", err)
+		logger.Infof(ipFailError.Error())
 
-		if err := o.transitionInstallPlanToFailed(plan, logger, v1alpha1.InstallPlanReasonInstallCheckFailed, ipFailError.Error()); err != nil {
-			// retry for failure to update status
+		now := o.now()
+		out := plan.DeepCopy()
+		out.Status.SetCondition(v1alpha1.ConditionFailed(v1alpha1.InstallPlanInstalled,
```
benluddy, line 1398: Is this condition reset anywhere after the OperatorGroup issue is resolved, but before all plan steps succeed?

doc/design/architecture.md
```diff
@@ -106,7 +106,7 @@ None --> Planning +------>------->------> Installing +---> Complete
 | RequiresApproval | occurs when using manual approval, will not transition phase until `approved` field is true    |
 | Installing       | resolved resources in the InstallPlan `Status` block are being created                      |
 | Complete         | all resolved resources in the `Status` block exist                                             |
-| Failed           | occurs when resources fail to install or there is an invalid OperatorGroup                     |
+| Failed           | occurs when bundle unpacking fails                     |
```
benluddy, line 109: Resources failing to install will still produce phase "Failed".

pkg/controller/operators/catalog/operator_test.go
```diff
@@ -182,16 +182,20 @@ func TestSyncInstallPlanUnhappy(t *testing.T) {
 		{
 			// This checks that an installplan is marked as failed when no operatorgroup is present
 			testName:      "HasSteps/NoOperatorGroup",
-			err:           nil,
-			expectedPhase: v1alpha1.InstallPlanPhaseFailed,
-			in:            ipWithSteps,
+			err:           fmt.Errorf("attenuated service account query failed - no operator group found that is managing this namespace"),
+			expectedPhase: v1alpha1.InstallPlanPhaseInstalling,
+			expectedCondition: &v1alpha1.InstallPlanCondition{Type: v1alpha1.InstallPlanInstalled, Status: corev1.ConditionFalse, Reason: v1alpha1.InstallPlanReasonInstallCheckFailed,
+				Message: "attenuated service account query failed - no operator group found that is managing this namespace"},
```
benluddy, line 188: Will users have any idea what an "attenuated service account query" is? As far as I know, that language is only used internally.

test/e2e/installplan_e2e_test.go
```diff
@@ -3895,18 +3716,6 @@ var _ = Describe("Install Plan", func() {
 		_, err = crc.OperatorsV1().OperatorGroups(ns.GetName()).Create(context.TODO(), og, metav1.CreateOptions{})
 		Expect(err).ToNot(HaveOccurred())
 
-		// Wait for the OperatorGroup to be synced and have a status.ServiceAccountRef
-		// before moving on. Otherwise the catalog operator treats it as an invalid OperatorGroup
-		// and fails the InstallPlan
-		Eventually(func() (*corev1.ObjectReference, error) {
```
benluddy, line 3901: Won't removing these make test flakes more likely? My understanding was that InstallPlan reconciliation will retry 8 times in the span of about one second, and if the OperatorGroup hasn't been reconciled by the eighth try, the InstallPlan reconciliation won't be attempted again for as long as 6 minutes. So anything that ties up olm-operator for 1 second could cause a timeout in a test that waits for an InstallPlan to execute.

doc/design/architecture.md
```diff
@@ -106,7 +106,7 @@ None --> Planning +------>------->------> Installing +---> Complete
 | RequiresApproval | occurs when using manual approval, will not transition phase until `approved` field is true    |
 | Installing       | resolved resources in the InstallPlan `Status` block are being created                      |
```
benluddy, line 107: Worth updating this while you're in the neighborhood -- it may still be resolving bundles, or, now, waiting for a valid OperatorGroup.

pkg/controller/operators/catalog/operator_test.go
anik120: That's a good point. Changing this to `no operator group found that is managing this namespace` only.

pkg/controller/operators/catalog/operator.go
anik120: The next step is bundle unpacking, and if any error is encountered there, this condition is reset there (tested [here](https://github.com/operator-framework/operator-lifecycle-manager/pull/2215/files#R3245)). The step right after that is to indicate that all plan steps have succeeded. So the possible unhappy paths looks covered.

test/e2e/installplan_e2e_test.go
anik120: > Won't removing these make test flakes more likely?

I can see how that could happen. I was mainly looking at it as a proof that we don't necessarily have to wait around for OG reconciliation for the IP to reach it's `Complete` phase, but I guess since other tests prove that OG reconciliation error keeps the IPs in the `Installing` phase, it's safer to poll for successful OG reconciliation to reduce test flakes.

pkg/controller/operators/catalog/operator.go
benluddy: Bundle unpacking could take a long time (several minutes, timing out after 10 minutes by default), and plan step errors are retried for up to 1 minute (by default). It could be confusing for users who see the OperatorGroup error, resolve the problem, then continue to see the same error for another 5 minutes.

anik120: @benluddy updated the PR to reset the condition, along with a note about assumptions made.

pkg/controller/operators/catalog/operator.go
```diff
@@ -1387,28 +1387,36 @@ func (o *Operator) syncInstallPlans(obj interface{}) (syncError error) {
 
 	querier := o.serviceAccountQuerier.NamespaceQuerier(plan.GetNamespace())
 	ref, err := querier()
-	if err != nil {
-
-		// Retry sync if non-fatal error
-		if !scoped.IsOperatorGroupError(err) {
```
hasbro17, line 1393: We can remove the associated error util here since it's only being used here.
https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/lib/scoped/error.go

But I'm okay with making this a follow up too.

pkg/controller/operators/catalog/operator.go
```diff
@@ -1387,28 +1387,36 @@ func (o *Operator) syncInstallPlans(obj interface{}) (syncError error) {
...
+		// NOTE: this makes the assumption that the InstallPlanInstalledCheckFailed reason is only set in the previous if clause, which is
+		// true in the current iteration of the catalog operator. Any future implementation change that aims at setting the reason as
+		// InstallPlanInstalledCheckFailed must make sure that either this assumption is not breached, or the condition being set elsewhere
+		// is not being unset here unintentionally.
+		if cond := out.Status.GetCondition(v1alpha1.InstallPlanInstalled); cond.Reason == v1alpha1.InstallPlanReasonInstallCheckFailed {
+			err := o.setInstallPlanInstalledCond(out, v1alpha1.InstallPlanConditionReason(corev1.ConditionUnknown), "", logger)
+			if err != nil {
+				syncError = err
+				return
+			}
```
hasbro17, line 1418: After updating the InstallPlan for resetting the condition we're proceeding in the reconcile loop with an outdated (resourceVersion) IP.
And the next step right after this could be to update the status again for the service account reference

https://github.com/operator-framework/operator-lifecycle-manager/blob/2e16b7e83f463f364627f7245adfe07ae631393f/pkg/controller/operators/catalog/operator.go#L1423-L1430

That would result in a resource conflict, sync error, and retry.

So to avoid that, we could either requeue and return here, or update the IP from the return value before proceeding:
```Go
plan, err := o.setInstallPlanInstalledCond(out, v1alpha1.InstallPlanConditionReason(corev1.ConditionUnknown), "", logger)
```

pkg/controller/operators/catalog/operator.go
```diff
@@ -1387,28 +1387,36 @@ func (o *Operator) syncInstallPlans(obj interface{}) (syncError error) {
...
-		// Requeue subscription to propagate SubscriptionInstallPlanFailed condtion to subscription
-		o.requeueSubscriptionForInstallPlan(plan, logger)
-
+		syncError = ipFailError
 		return
+	} else {
+		// reset condition/message if it had been set in previous sync. This condition is being reset since any delay in the next steps
+		// (bundle unpacking/plan step errors being retried for a duration) could lead to this condition sticking around, even after
+		// the serviceAccountQuerier returns no error since the error has been resolved (by creating the required resources), which would
+		// be confusing to the user
```
hasbro17, line 1407: We should probably add a unit test case to test this path and check that this holds, i.e pass an IP with the condtion`Installing=false` and reason `InstallPlanInstalledCheckFailed` and check that the condition gets reset with an OG present.

pkg/controller/operators/catalog/operator_test.go
```diff
@@ -1830,3 +1841,15 @@ func operatorGroup(ogName, saName, namespace string, saRef *corev1.ObjectReferen
 		},
 	}
 }
+
+func hasExpectedCondition(ip *v1alpha1.InstallPlan, expectedCondition v1alpha1.InstallPlanCondition) bool {
+
```
hasbro17, line 1846: Nit
```suggestion
```

pkg/controller/operators/catalog/operator_test.go
```diff
@@ -1830,3 +1841,15 @@ func operatorGroup(ogName, saName, namespace string, saRef *corev1.ObjectReferen
...
+	hasCondition := false
+	for _, cond := range ip.Status.Conditions {
+		if cond.Type == expectedCondition.Type && cond.Message == expectedCondition.Message && cond.Status == expectedCondition.Status {
+			hasCondition = true
+			break
+		}
+	}
+	return hasCondition
```
hasbro17, line 1854: Also nit:
```suggestion
	for _, cond := range ip.Status.Conditions {
		if cond.Type == expectedCondition.Type && cond.Message == expectedCondition.Message && cond.Status == expectedCondition.Status {
		         return true
		}
	}
	return false
```

pkg/controller/operators/catalog/operator.go
anik120: @hasbro17 that's a really good point. Although, I was trying to read through what we have in the main branch currently, and doesn't look like we handle this concern anywhere, i.e whenever there's an update, we don't update `plan` to have the latest object. Which means we're always running into the scenario that you've mentioned. And if that's the case, will just updating the plan at this step make any dent at all?

pkg/controller/operators/catalog/operator.go
anik120: It looks harmless + we could end up needing it later on. Feels like it's better to leave that method in there just for the sake of completeness of the type definition.

pkg/controller/operators/catalog/operator.go
hasbro17: I'd have to check the rest of the loop but at least for this part previously we were returning after we had to update the IP to avoid the conflict.
https://github.com/operator-framework/operator-lifecycle-manager/blob/20ded32d2260a8f1eeb594b9ec2147ad0134cfc6/pkg/controller/operators/catalog/operator.go#L1402-L1411

Given that it's an additional request to resolve the conflict I would still lean towards updating or returning.

But it's a trivial issue so we probably don't need to split hairs on it and can punt on this for now 🙂

pkg/controller/operators/catalog/operator.go
anik120: Added an e2e test for this instead since an e2e test feels more natural for this.

pkg/controller/operators/catalog/operator.go
anik120: I see, no it makes sense to me. I've updated `plan` with the returned IP after status update to avoid that additional request 👍🏽

test/e2e/installplan_e2e_test.go
```diff
@@ -3145,160 +3145,94 @@ var _ = Describe("Install Plan", func() {
...
+			// The status gets ignored on create so we need to update it else the InstallPlan sync ignores
+			// InstallPlans without any steps or bundle lookups
+			outIP.Status = ip.Status
+			_, err = crc.OperatorsV1alpha1().InstallPlans(ns.GetName()).UpdateStatus(context.TODO(), outIP, metav1.UpdateOptions{})
+			Expect(err).NotTo(HaveOccurred())
+		})
 
-	It("should fail an InstallPlan when an OperatorGroup specifies an unsynced ServiceAccount with no ServiceAccountRef", func() {
+		It("should clear clear up the condition in the InstallPlan status that contains an error message when a valid OperatorGroup is created", func() {
+			// Create an operatorgroup for the same namespace
```
hasbro17, line 3183: Before we create the OG to trigger the removal of the `Installed`/`InstallCheckFailed`  condition, shouldn't we check if it's present before a valid OG is present?
Since we essentially want to test for the change of that condition.

test/e2e/installplan_e2e_test.go
```diff
@@ -3145,160 +3145,94 @@ var _ = Describe("Install Plan", func() {
...
+					return false, err
+				}
+				if fetchedInstallPlan == nil {
+					return false, err
+				}
+				unwantedCondition := v1alpha1.InstallPlanCondition{Type: v1alpha1.InstallPlanInstalled, Status: corev1.ConditionFalse, Reason: v1alpha1.InstallPlanReasonInstallCheckFailed,
+					Message: "no operator group found that is managing this namespace"}
+				if fetchedInstallPlan.Status.Phase != v1alpha1.InstallPlanPhaseInstalling {
+					return false, nil
+				}
```
hasbro17, line 3221: Is this necessary?
I think when we get the IP above with the phase check, it's going to stall (poll 20 mins) until it finds the IP in phase `Installing`.
```Go
fetchInstallPlanWithNamespace(GinkgoT(), crc, installPlanName, ns.GetName(), buildInstallPlanPhaseCheckFunc(operatorsv1alpha1.InstallPlanPhaseInstalling))
```

https://github.com/operator-framework/operator-lifecycle-manager/blob/9723d5d205afcefdc228d32a19dc64d220a25df1/test/e2e/installplan_e2e_test.go#L4052-L4058

So the fetched IP will always be in phase `Installing` at this point.

test/e2e/installplan_e2e_test.go
anik120: That's a good point, added a check to see if the InstallPlan has the expected condition/message before it is being cleared up.

test/e2e/installplan_e2e_test.go
anik120: Not a big fan of the `buildInstallPlanPhaseCheckFunc`, I think it's an artifact of the pre gikgo test framework we use now. `fetchInstallPlan` should do one thing, i.e fetch the IP, and the checking of the phase is a separate task. But looks like we're using it in a lot of places, so I'm okay with refactoring the tests later on, and leaving this check out of here for now.

hasbro17 (approved):
/lgtm

PR #2216, opened 2021-06-28
Emit CSV metric on startup

josefkarasek:
`csv_succeeded` metric is lost between pod restarts.
This is because this metric is only emitted when CSV.Status is changed.

Emit `csv_succeeded`/`csv_abnormal` metric during OLM startup

csv_succeeded metric is lost between pod restarts.

**Reviewer Checklist**

### DIFF
--- a/cmd/olm/main.go
+++ b/cmd/olm/main.go
@@ -171,6 +171,11 @@ func main() {
 	op.Run(ctx)
 	<-op.Ready()
 
+	// Emit CSV metric
+	if err = op.EnsureCSVMetric(); err != nil {
+		logger.WithError(err).Fatalf("error emitting metrics for existing CSV")
+	}
+
 	if *writeStatusName != "" {
 		reconciler, err := openshift.NewClusterOperatorReconciler(
 			openshift.WithClient(mgr.GetClient()),
--- a/test/e2e/metrics_e2e_test.go
+++ b/test/e2e/metrics_e2e_test.go
@@ -16,6 +16,7 @@ import (
 	. "github.com/onsi/gomega"
 	io_prometheus_client "github.com/prometheus/client_model/go"
 	"github.com/prometheus/common/expfmt"
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -113,6 +114,44 @@ var _ = Describe("Metrics are generated for OLM managed resources", func() {
 						ContainElement(LikeMetric(WithFamily("csv_abnormal"), WithName(failingCSV.Name))),
 						ContainElement(LikeMetric(WithFamily("csv_succeeded"), WithName(failingCSV.Name))),
 					))
+				})
+			})
+		})
+
+		When("a CSV is created", func() {
+			var (
+				cleanupCSV cleanupFunc
+				csv        v1alpha1.ClusterServiceVersion
+			)
+			BeforeEach(func() {
+				packageName := genName("csv-test-")
+				packageStable := fmt.Sprintf("%s-stable", packageName)
+				csv = newCSV(packageStable, testNamespace, "", semver.MustParse("0.1.0"), nil, nil, nil)
+
+				var err error
+				_, err = createCSV(c, crc, csv, testNamespace, false, false)
+				Expect(err).ToNot(HaveOccurred())
+				_, err = fetchCSV(crc, csv.Name, testNamespace, csvSucceededChecker)
+				Expect(err).ToNot(HaveOccurred())
+			})
+			AfterEach(func() {
+				if cleanupCSV != nil {
+					cleanupCSV()
+				}
+			})
+			It("emits a CSV metrics", func() {
+				Expect(getMetricsFromPod(c, getPodWithLabel(c, "app=olm-operator"))).To(
+					ContainElement(LikeMetric(WithFamily("csv_succeeded"), WithName(csv.Name), WithValue(1))),
+				)
+			})
+			When("the OLM pod restarts", func() {
+				BeforeEach(func() {
+					restartDeploymentWithLabel(c, "app=olm-operator")
+				})
+				It("CSV metric is preserved", func() {
+					Expect(getMetricsFromPod(c, getPodWithLabel(c, "app=olm-operator"))).To(
+						ContainElement(LikeMetric(WithFamily("csv_succeeded"), WithName(csv.Name), WithValue(1))),
+					)
 				})
 			})
 		})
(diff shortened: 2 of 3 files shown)


josefkarasek:
This is an alternative solution along with #2213

cmd/olm/main.go
```diff
@@ -257,10 +258,30 @@ func main() {
...
+
 	// Start the controller manager
 	if err := mgr.Start(ctx); err != nil {
 		logger.WithError(err).Fatal("controller manager stopped")
 	}
 
 	<-op.Done()
 }
+
+func ensureCSVMetric(logger *logrus.Logger, c *versioned.Clientset) error {
```
anik120, line 274: This logic looks perfect, although I'm not sure if the main function is the write place for it. Is there a reason including this logic when a new olm controller is created instead might be a bad idea?

anik120 (changes requested):
/ok-to-test

cmd/olm/main.go
josefkarasek: The informer cache is not yet populated at this moment - I see that as an alternative place for this call. Just when the cache gets populated.

timflannagan: Nothing immediate comes to mind as to why this implementation may be a bad idea, but maybe others have stronger opinions on this - we could also wrap this in a Runnable instance to abstract away some of the prerequisites before starting the controller/informers/etc. if we're worried about control flow in the driver function.

timflannagan: I guess one main downsides is that performing a list call on a non-cached client may be inefficient in the context of a cluster with a significant amount of CSVs.

josefkarasek: I think I was wrong assuming that the List call is from a cold cache.
[This](https://github.com/operator-framework/operator-lifecycle-manager/blob/6a52c02a42d8623a369e6501520919be066d1644/cmd/olm/main.go#L210) is where the main function waits for the cache sync and [this](https://github.com/operator-framework/operator-lifecycle-manager/blob/6a52c02a42d8623a369e6501520919be066d1644/pkg/lib/queueinformer/queueinformer_operator.go#L182) is the deferred call that signals that the cache has synced.

ad Placing this call to main.go: What is the alternative? The way I look at it is that it shouldn't happen before the informer cache has synced and not after the controllers have started.

anik120: > not after the controllers have started

Could you elaborate a little more on why not?

josefkarasek: 1) Then the controller could emit the metric during CSV reconciliation
2) It prolongs the period when the metric is not accurately set - thus we don't target the core problem

cmd/olm/main.go
```diff
@@ -257,10 +258,30 @@ func main() {
...
+	logger.Debug("Emitting CSV metric")
```
anik120, line 275: `nit`: "Emitting metrics for existing CSVs" might be a more appropriate log message

cmd/olm/main.go
```diff
@@ -257,10 +258,30 @@ func main() {
 		go monitor.Run(op.Done())
 	}
 
+	// Emit CSV metric
+	if err = ensureCSVMetric(logger, crClient); err != nil {
+		logger.WithError(err).Fatalf("error emitting CSV metric")
```
anik120, line 263: `nit`: similar to the other comment: "error emitting metrics for existing CSV" might be more readable.

test/e2e/metrics_e2e_test.go
```diff
@@ -112,6 +113,41 @@ var _ = Describe("Metrics are generated for OLM managed resources", func() {
...
+				_, err = fetchCSV(crc, csv.Name, testNamespace, csvSucceededChecker)
+				Expect(err).ToNot(HaveOccurred())
+			})
+			AfterEach(func() {
+				if cleanupCSV != nil {
+					cleanupCSV()
+				}
+			})
+			It("csv metric is preserved", func() {
+				Expect(getMetricsFromPod(c, getPodWithLabel(c, "app=olm-operator"), "8081")).To(
```
anik120, line 140: `nit`: I would expect this in `BeforeEach`: create a CSV, AND make sure metrics is emitted in BeforeEach 

Also, the DCO target is failing because the commits need to be signed off. The target has more details.

anik120 (approved):
/lgtm

dinhxuanvu (approved):
/lgtm

cmd/olm/main.go
```diff
@@ -257,10 +258,29 @@ func main() {
 		go monitor.Run(op.Done())
 	}
 
+	// Emit CSV metric
+	if err = ensureCSVMetric(logger, op); err != nil {
```
ecordell, line 262: Why is this in `main` and not the operator package?

cmd/olm/main.go
ecordell: https://github.com/operator-framework/operator-lifecycle-manager/blob/6a52c02a42d8623a369e6501520919be066d1644/pkg/lib/queueinformer/queueinformer_operator.go#L213-L218

couldn't it go in between here? after caches have synced but before controllers have started?

josefkarasek: That looks to be a fairly generic function. [Catalog operator](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/cmd/catalog/main.go#L139)(and other clients of that lib) would then be exposing CSV metric.

[OLM operator](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/controller/operators/olm/operator.go#L61) doesn't have a sense when the cache has been synced. But it could be done - we could pass the ready channel to NewOperator() and spin up a goroutine that would emit the metric once the cache is synced.
It sounds quite complicated, but the emit logic would be in the operator pkg.

cmd/olm/main.go
anik120: +1 I have the same question.

test/e2e/metrics_e2e_test.go
```diff
@@ -116,6 +117,45 @@ var _ = Describe("Metrics are generated for OLM managed resources", func() {
 				})
 			})
 		})
+
+		When("the OLM pod restars", func() {
```
estroz, line 121:
```suggestion
		When("the OLM pod restarts", func() {
```

estroz:
I don't see necessarily a "better" place for the EnsureCSVMetrics call since queueinformer.Operator.Run can't easily be wrapped, so this looks fine to me.

/lgtm

test/e2e/metrics_e2e_test.go
anik120: The setup of the test feels a little unnatural to me. I would expect a `When` block inside this [`Context`](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/test/e2e/metrics_e2e_test.go#L121). And the layout ideally would something be: 

```
Context("Metrics emitted by objects during operator installation", func() {

  When("A CSV is created", func() {
			BeforeEach(func() {
                                   // create CSV
                       }
                      It("emits a CSV metrics", func() {
                              // check for CSV metrics
                      }
                     When("The OLM pod restarts", func() {
                         BeforeEach(func(){
                                // kill the deployment and wait for it to come up again
                         }
                        It("The csv metrics are preserved" func(){
                                 // check for CSV metrics again
                       }) 
                     })
  }
```

This leaves room for extending this portion of the test suite in the future

anik120 (changes requested):

anik120:
This looks good to me. If Evan's concerns have been assuaged (I don't see any conversation here that suggests that they have been yet), we can remove the hold and lgtm it.

dinhxuanvu (approved):
/lgtm

PR #2222, opened 2021-06-29
feat(og): Project OperatorGroup errors to status conditions

dinhxuanvu:

Currently, errors associated with ServiceAccount not found or multiple
OperatorGroup in the same namespace are not recorded in OperatorGroup's
status. This PR will project those errors to OG's status conditions to
improve the UX.

**Reviewer Checklist**


pkg/lib/scoped/querier.go
```diff
@@ -75,11 +81,33 @@ func queryServiceAccountFromNamespace(logger *logrus.Entry, crclient versioned.I
 	}
 
 	if len(groups) > 1 {
+		// Add to all OG's status conditions to indidate they're multiple OGs in the
```
DhritiShikhar, line 84: typo

indidate -> indicate

pkg/lib/scoped/querier.go
```diff
@@ -6,12 +6,18 @@ import (
 
 	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
+	meta "k8s.io/apimachinery/pkg/api/meta"
```
timflannagan, line 9: Nit: nothing pressing, but there's an existing meta package import (metav1) right before before this declaration.

pkg/controller/operators/olm/operator.go
```diff
@@ -827,6 +828,42 @@ func (a *Operator) syncNamespace(obj interface{}) error {
 		return err
 	}
 
+	// Filter the operator associated with this namespace
+	groups := make([]*v1.OperatorGroup, 0)
+	for _, og := range operatorGroupList {
+		if og.GetNamespace() == namespace.GetName() {
+			groups = append(groups, og)
+		}
+	}
```
awgreene, line 837: Non-Blocking Nit: Given that the lister retrieves the `OperatorGroups` from a cache it might be cleaner to filter by namespace above. An additional call to the cache can be made below for the bit that requires all `OperatorGroups`.

pkg/lib/scoped/syncer.go
```diff
@@ -74,6 +78,19 @@ func (s *UserDefinedServiceAccountSyncer) SyncOperatorGroup(in *v1.OperatorGroup
...
+			Type:    v1.OperatorGroupServiceAccountCondition,
+			Status:  metav1.ConditionTrue,
+			Reason:  v1.OperatorGroupServiceAccountReason,
+			Message: fmt.Sprintf("ServiceAccount %s not found", serviceAccountName),
+		}
+
+		meta.SetStatusCondition(&in.Status.Conditions, cond)
+		_, err = s.update(in, nil)
+		if err != nil {
+			logger.Warnf("fail to upgrade operator group status condition og=%s: %s", in.GetName(), err.Error())
```
awgreene, line 92: Same as above.

pkg/controller/operators/olm/operator.go
```diff
@@ -827,6 +828,42 @@ func (a *Operator) syncNamespace(obj interface{}) error {
...
+			Status:  metav1.ConditionTrue,
+			Reason:  v1.MultipleOperatorGroupsReason,
+			Message: "Multiple OperatorGroup found in the same namespace",
+		}
+		for _, og := range groups {
+			meta.SetStatusCondition(&og.Status.Conditions, cond)
+			_, err = a.client.OperatorsV1().OperatorGroups(namespace.GetName()).UpdateStatus(context.TODO(), og, metav1.UpdateOptions{})
+			if err != nil {
+				logger.Warnf("fail to upgrade operator group status condition og=%s: %s", og.GetName(), err.Error())
+			}
```
awgreene, line 863:
```suggestion
if _, err = a.client.OperatorsV1().OperatorGroups(namespace.GetName()).UpdateStatus(context.TODO(), og, metav1.UpdateOptions{}); err != nil {
	logger.Warnf("error updating OperatorGroup status condition og=%s: %s", in.GetName(), err.Error())
}
```
It may be worthwhile to log the condition as well.

awgreene (changes requested):
Nice work, a few nits but its looking good.

awgreene:
Seems like a few files were removed in the `deploy/chart/crds/` dir

ecordell:
are you running a different version of the code generator locally? it looks like the committed files have different names than the `verify` script expects

pkg/lib/scoped/syncer.go
```diff
@@ -74,6 +78,19 @@ func (s *UserDefinedServiceAccountSyncer) SyncOperatorGroup(in *v1.OperatorGroup
 	// A service account has been specified, we need to update the status.
 	sa, err := s.client.KubernetesInterface().CoreV1().ServiceAccounts(namespace).Get(context.TODO(), serviceAccountName, metav1.GetOptions{})
 	if err != nil {
+		// Set OG's status condition to indicate SA is not found
+		cond := metav1.Condition{
+			Type:    v1.OperatorGroupServiceAccountCondition,
+			Status:  metav1.ConditionTrue,
+			Reason:  v1.OperatorGroupServiceAccountReason,
+			Message: fmt.Sprintf("ServiceAccount %s not found", serviceAccountName),
```
awgreene, line 86: Is it worth including the type of error?

awgreene: Or at least logging it?

awgreene (changes requested):
One nit but looks good otherwise.

pkg/lib/scoped/syncer.go
```diff
@@ -74,6 +78,19 @@ func (s *UserDefinedServiceAccountSyncer) SyncOperatorGroup(in *v1.OperatorGroup
...
+		}
+
+		meta.SetStatusCondition(&in.Status.Conditions, cond)
+		_, err = s.update(in, nil)
```
awgreene: This actually overrides the Get call error, which is then used on line 94, so the error returned by the function may be nil or refer to an issue setting the Condition.

dinhxuanvu: I already corrected this. The condition is added when SA is not found (according to the error). When there is a different type of error, then log and return that error.

awgreene:
/approve

Nit: Squash the commits.

awgreene (commented):
/approve

pkg/controller/operators/olm/operator.go
```diff
@@ -827,6 +828,37 @@ func (a *Operator) syncNamespace(obj interface{}) error {
 		return err
 	}
 
+	// Query OG in this namespace
+	groups, err := a.lister.OperatorsV1().OperatorGroupLister().OperatorGroups(namespace.GetName()).List(labels.Everything())
```
ecordell, line 832: err isn't checked

pkg/controller/operators/olm/operator.go
```diff
@@ -827,6 +828,37 @@ func (a *Operator) syncNamespace(obj interface{}) error {
...
+
+	// Check if there is a stale multiple OG condition and clear it if existed.
+	if len(groups) == 1 {
```
ecordell, line 835: why do this in `syncNamespace` instead of `syncOperatorGroup`? The only time this condition changes is when an operatorgroup has been created or deleted, right?

dinhxuanvu: Because in `syncNamespace`, I can query all OG in that namespace to determine if there are multiple OG in the same namespace and add the conditions for MultipleOG accordingly. It is a namespace level check so I think that is an appropriate to check for it. `syncOperatorGroup` is to sync individual OG. If I query all OG in the same namespace for every OG in the same namespace, then there will be redundant calls. I know it is not a big deal given we use cache but still I think this way is more effective. Plus, when a OG is updated, the `syncOperatorGroups` will trigger `syncNamespace` so that will take care of the OG is updated as well.

pkg/controller/operators/olm/operator.go
dinhxuanvu: Fixed.

PR #2224, opened 2021-06-29
(fix)registry: fix configmap registry server liveness probe timeouts

anik120:

In #1378, the readiness probe initial delay and timeout were bumped
to account for node resource constraints during pod startup. This
PR bumps the liveness probe delay and timeout values too, to account
for scenarios where node resources continue to be overloaded causing
the liveness probe to fail temporarily.

Failing readiness/liveness probe fires prometheus alerts for
CatalogSources backed by configmap registry servers.

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/registry/reconciler/configmap.go
+++ b/pkg/controller/registry/reconciler/configmap.go
@@ -101,7 +101,7 @@ func (s *configMapCatalogSourceDecorator) Service() *v1.Service {
 }
 
 func (s *configMapCatalogSourceDecorator) Pod(image string) *v1.Pod {
-	pod := Pod(s.CatalogSource, "configmap-registry-server", image, "", s.Labels(), s.Annotations(), 5, 2)
+	pod := Pod(s.CatalogSource, "configmap-registry-server", image, "", s.Labels(), s.Annotations(), 5, 5)
 	pod.Spec.ServiceAccountName = s.GetName() + ConfigMapServerPostfix
 	pod.Spec.Containers[0].Command = []string{"configmap-server", "-c", s.Spec.ConfigMap, "-n", s.GetNamespace()}
 	ownerutil.AddOwner(pod, s.CatalogSource, false, false)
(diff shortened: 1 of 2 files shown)


pkg/controller/registry/reconciler/configmap.go
```diff
@@ -101,7 +101,7 @@ func (s *configMapCatalogSourceDecorator) Service() *v1.Service {
 }
 
 func (s *configMapCatalogSourceDecorator) Pod(image string) *v1.Pod {
-	pod := Pod(s.CatalogSource, "configmap-registry-server", image, "", s.Labels(), s.Annotations(), 5, 2)
+	pod := Pod(s.CatalogSource, "configmap-registry-server", image, "", s.Labels(), s.Annotations(), 5, 5)
```
timflannagan, line 104: Minor nit: This seems like a magic number for people that are new to the codebase - I wonder if it's worth making this a constant in the reconcile package so we don't hardcode that value in two places.

anik120: I don't mind creating two constants for `readinessDelay` and `livenessDelay`, but for 

> magic number for people that are new to the codebase

if you look at the function signature
```
Pod(source *v1alpha1.CatalogSource, name string, image string, saName string, labels map[string]string, annotations map[string]string, readinessDelay int32, livenessDelay int32)
```

it looks pretty self-explanatory. 

Creating two constants `LivenessDelay` and `ReadinessDelay` seems superfluous when we're only creating this pod in one place.

timflannagan: That's fine with me - I wasn't recommending you change it, but throwing out the suggestion as I think it improves readability.

That function signature is brutal though, so moving a couple of parameters from hardcoded numbers to constants probably doesn't move the needle all that much. Refactoring that function to use a type, or reducing the number of parameters, doesn't seem worth the effort for a function with one call site.

dinhxuanvu:
Rerun the unit test due to flakes.

dinhxuanvu (approved):
/lgtm

PR #2225, opened 2021-06-29
Require valid client certificate for profiling endpoints.

benluddy:
One HTTP server is used for healthz, metrics, and profiling.

TLS can still be enabled via flags, but it now applies to all three
APIs since there is a single HTTP server per operator process.

The -profiling flag is deprecated and does nothing. The profiling APIs
are now always enabled, but they refuse to serve clients that do not
present a verifiable certificate (see the new flag -client-ca). This
effectively disables the profiling APIs when served over HTTP.

### DIFF
--- a/cmd/catalog/main.go
+++ b/cmd/catalog/main.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"context"
-	"crypto/tls"
 	"flag"
 	"fmt"
 	"net/http"
@@ -10,7 +9,6 @@ import (
 	"time"
 
 	configv1client "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
-	"github.com/prometheus/client_golang/prometheus/promhttp"
 	log "github.com/sirupsen/logrus"
 	utilclock "k8s.io/apimachinery/pkg/util/clock"
 	k8sscheme "k8s.io/client-go/kubernetes/scheme"
@@ -18,10 +16,9 @@ import (
 
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/client"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/operators/catalog"
-	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/filemonitor"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/operatorclient"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/operatorstatus"
-	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/profile"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/server"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/signals"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/metrics"
 	olmversion "github.com/operator-framework/operator-lifecycle-manager/pkg/version"
@@ -67,8 +64,9 @@ var (
 	tlsCertPath = flag.String(
 		"tls-cert", "", "Path to use for certificate key (requires tls-key)")
 
-	profiling = flag.Bool(
-		"profiling", false, "serve profiling data (on port 8080)")
+	profiling = flag.Bool("profiling", false, "deprecated")
+
+	clientCAPath = flag.String("client-ca", "", "path to watch for client ca bundle")
 
 	installPlanTimeout  = flag.Duration("install-plan-retry-timeout", 1*time.Minute, "time since first attempt at which plan execution errors are considered fatal")
 	bundleUnpackTimeout = flag.Duration("bundle-unpack-timeout", 10*time.Minute, "The time limit for bundle unpacking, after which InstallPlan execution is considered to have failed. 0 is considered as having no timeout.")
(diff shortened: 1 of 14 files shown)


cmd/catalog/main.go
```diff
@@ -106,60 +109,59 @@ func main() {
...
+		logger.Fatal("both --tls-key and --tls-crt must be provided for TLS to be enabled")
 	} else {
-		go func() {
-			err := http.ListenAndServe(":8081", metricsMux)
-			if err != nil {
-				logger.Errorf("Metrics (http) serving failed: %v", err)
-			}
-		}()
+		logger.Info("TLS keys not set, using non-https for metrics")
 	}
```
awgreene, line 157: This code reappears in [olm/main.go](https://github.com/operator-framework/operator-lifecycle-manager/pull/2225/files#diff-52ebf4e1c25c87743e2d504e8015be1767d6140aadf638304142cdbb54216b3eR139-R171), can we create a function shared between the two?

awgreene (changes requested):
One nit, but looks good otherwise.

kevinrizza:
Overall this looks good to me, but I think we should think about how to  provide a more explicit path on the templating changes when you want to specify the CA and TLS stuff, given that they aren't referenced at all in the values.yaml file

benluddy:
/hold

Going to rebase/fixup.

deploy/chart/values.yaml
```diff
@@ -55,3 +61,6 @@ package:
 monitoring:
   enabled: false
   namespace: monitoring
+
+pprof:
+  enabled: false
```
benluddy, line 66: Isn't this technically always enabled now (though effectively disabled if no client CA bundle is configured)?

deploy/chart/templates/0000_50_olm_02-services.yaml
```diff
@@ -31,9 +31,48 @@ spec:
...
+spec:
+  type: ClusterIP
+  ports:
+  - name: https-pprof
+    port: {{ .Values.catalog.service.internalPort }}
+    protocol: TCP
+    targetPort: {{ .Values.catalog.service.externalPort }}
   selector:
     app: catalog-operator
 {{ end }}
```
benluddy, line 78: These should live in OCP, since they rely on service-ca-operator.

deploy/chart/values.yaml
```diff
@@ -13,7 +13,10 @@ olm:
     ref: quay.io/operator-framework/olm:master
     pullPolicy: Always
   service:
-    internalPort: 8080
+    internalPort: 8443
+    externalPort: 8443
+  tlsSecret: olm-serving-cert
+  clientCASecret: pprof-serving-cert
```
benluddy, line 19: Shouldn't this remain HTTP by default? Users will need to customize this on their own based on the way they manage certificates.

deploy/chart/values.yaml
awgreene: Fair - we can remove this and rely on the `{olm|catalog}.clientCASecret`.

deploy/chart/values.yaml
awgreene: Isn't upstream built using the [deploy/upstream/values.yaml](https://github.com/operator-framework/operator-lifecycle-manager/blob/fffd2207fa0fac57bb99c59afe98823945c6b4b7/deploy/upstream/values.yaml) file? I agree with the sentiment behind your statement though.

deploy/chart/templates/0000_50_olm_02-services.yaml
awgreene: Good point. I'll remove these here.

deploy/chart/values.yaml
awgreene: This actually isn't even needed if we remove the pprof services.

deploy/chart/templates/0000_50_olm_02-services.yaml
```diff
@@ -31,9 +31,48 @@ spec:
...
+    app: catalog-operator
+{{ end }}
+{{ if .Values.pprof.enabled }}
+apiVersion: v1
+kind: Service
+metadata:
+  name: olm-operator-pprof
+  namespace: {{ .Values.namespace }}
+  annotations:
+    service.alpha.openshift.io/serving-cert-secret-name: olm-operator-serving-cert
```
timflannagan, line 47: Just a quick note: I believe this annotation has graduated to beta when I was implementing a downstream solution for the PackageServer CSV to be deployed as a deployment/APIService/etc. set of resources.

timflannagan: (Ah, nevermind didn't read the conversation about these manifests being removed -- disregard).

deploy/chart/templates/0000_50_olm_07-olm-operator.deployment.yaml
```diff
@@ -43,29 +69,30 @@ spec:
...
+          - --client-ca
+          - /profile-collector-cert/tls.crt
           {{- end }}
           image: {{ .Values.olm.image.ref }}
           imagePullPolicy: {{ .Values.olm.image.pullPolicy }}
           ports:
             - containerPort: {{ .Values.olm.service.internalPort }}
-            - containerPort: 8081
-              name: metrics
-              protocol: TCP
```
awgreene, line 60: Removed given that this should always be 8080 or 8081, as set in the values.yaml

benluddy:
> /lgtm

/lgtm

PR #2237, opened 2021-07-06
Bug 1989731: Registry ensure methods error out if resource is not found

awgreene:
Problem: When calling a registry's ensure methods, if the resource is
deleted between the get and delete calls the ensure method fails due
to a 404 error.

Solution: When deleting an object, ignore 404 errors.


awgreene:
The team has decided that this bug is not sever enough to warrant backporting, closing this PR.

PR #2239, opened 2021-07-07
Fix cpb traverses unneeded paths

eramoto:
- Excludes device filesystems (`/dev`) and pseudo filesystems (`/proc` and `/sys`) from filesystems looking for metadata.
- Outputs absolute path of skipped file of directory.

`cpb`, it is run on `/` of several images, outputs many unuseful messages and spoils log aggregation systems (`elasticsearch`, `loki` etc.).

**Reviewer Checklist**

### DIFF
--- a/util/cpb/main.go
+++ b/util/cpb/main.go
@@ -84,6 +84,13 @@ func getMetadata() (m *metadata, err error) {
 		manifestDir:     "/manifests",
 	}
 
+	// Exclude device filesystems and pseudo filesystems from filesystems looking for metadata
+	excludeDir := []string{
+		"/dev",
+		"/proc",
+		"/sys",
+	}
+
 	// Traverse the filesystem looking for metadata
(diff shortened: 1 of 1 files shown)


eramoto:
Added my sign-off to this commit

util/cpb/main.go
```diff
@@ -84,14 +84,32 @@ func getMetadata() (m *metadata, err error) {
...
 			return nil
 		}
 		if info.IsDir() {
-			fmt.Printf("skipping a dir without errors: %+v \n", info.Name())
+			absPath, err := filepath.Abs(path)
+			if err != nil {
+				fmt.Printf("couldn't get the absolute path %q: %v\n", path, err)
+				return nil
+			}
+			for _, v := range excludeDir {
```
njhale, line 106: nit: We could use a hash set for an O(1) lookup here, but since we have a relatively small and constant number of exclusions, I'm not too worried about being ultra efficient here.

njhale (approved):
/lgtm

PR #2241, opened 2021-07-07
Remove cleanup from startup

ecordell:
Remove cleanup from startup

a series of cleanup steps run on start, but they are only required
when updating from very old versions of olm. it doesn't make sense to
carry them forever.

**Reviewer Checklist**


njhale (approved):
This seems fine. The versions of OLM that the cleanup addresses are obsolete.

Users can step through an intermediate version if they want a seamless upgrade anyway.

/lgtm

ecordell:
/hold

want to double check that we don't need https://github.com/operator-framework/operator-lifecycle-manager/commit/05f5417a37c6b567c70bf98fbe5197fa707b958a#diff-4ce7e916862e18b439f8ce8edbf0b0a85f384589f0ef2b8c5ff32237cb907da9

ecordell:
/hold cancel

that one looks fine - hasn't been needed since ocp 4.4

timflannagan:
Need to investigate why that e2e-tests action is continuously timing out so the openshift-bot doesn't perma-retest.

/hold

PR #2242, opened 2021-07-08
feat(sub): Include IP failure condition message in sub status condition

anik120:

Currently when an InstallPlan
1. fails, due to bundle unpacking error
2. is pending due to errors like invalid operatorgroup, or pending bundle unpacking job

the reason is propagated to the Subscription that owns it. The message for the failure is
however missing from the Subscription condition.

eg
```
kind: Subscription
status:
conditions:
    - lastTransitionTime: "2021-07-07T17:55:20Z"
      reason: Installing
      status: "True"
      type: InstallPlanPending
```

This PR propagates the message assosiated with the reason in the InstallPlan
condition when an InstallPlan is either pending, or has failed permanantly.

eg
```
kind: Subscription
status:
conditions:
    - lastTransitionTime: "2021-07-07T17:55:20Z"
      message: no operator group found that is managing this namespace
      reason: Installing
      status: "True"
      type: InstallPlanPending
```

**Reviewer Checklist**


anik120:
Closed in favor of https://github.com/operator-framework/operator-lifecycle-manager/pull/2262

PR #2228, opened 2021-06-30
feat: remove support for deprecated APIs ahead of kube 1.22

exdx:
**Description of the change:**
A series of deprecated APIs are being removed entirely in kube 1.22 per https://kubernetes.io/docs/reference/using-api/deprecation-guide/#v1-22. This PR removes support for them explicitly in OLM, and also updates all OLM e2e tests to ensure they are not using deprecated APIs that are being removed in 1.22. 

The following GroupVersions are no longer supported by OLM:
```
	schema.GroupVersion{Group: "admissionregistration.k8s.io", Version: "v1beta1"}:      {},
	schema.GroupVersion{Group: "apiextensions.k8s.io/v1beta1", Version: "v1beta1"}:      {},
	schema.GroupVersion{Group: "apiregistration.k8s.io/v1beta1", Version: "v1beta1"}:    {},
	schema.GroupVersion{Group: "rbac.authorization.k8s.io/v1beta1", Version: "v1beta1"}: {},
	schema.GroupVersion{Group: "scheduling.k8s.io/v1beta1", Version: "v1beta1"}:         {},
```

**Motivation for the change:**
Eventual kube 1.22 rebase 

**Reviewer Checklist**
- [ ] Implementation matches the proposed design, or proposal is updated to match implementation
- [ ] Sufficient unit test coverage
- [ ] Sufficient end-to-end test coverage
- [ ] Docs updated or added to `/doc`
- [ ] Commit messages sensible and descriptive

PR #2247, opened 2021-07-09
[WIP] Remove all v1beta1 crds from tests

kevinrizza:
Still needs to run e2e tests, throwing this wip up to test with


test/e2e/csv_e2e_test.go
```diff
@@ -16,7 +16,6 @@ import (
 	rbacv1 "k8s.io/api/rbac/v1"
 	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions"
```
timflannagan, line 17: I'm surprised this package wasn't removed - it looks like it's still used to register the sheme:

```
	if err := apiextensions.AddToScheme(scheme); err != nil {
		return nil, err
	}
```

Is that scheme still needed or is fine to keep in place?

test/e2e/testdata/vpa/crd.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
```
exdx, line 1: This needs to be updated further with schema changes to be a valid v1 CRD, see https://github.com/operator-framework/operator-lifecycle-manager/pull/2228/files#diff-c3d03b07312e2fc84b0bd43696d3b855b171253a3f1e48dc58b7b131595014dc

In general for most of these tests simply bumping the version won't be enough, they won't be valid v1 CRDs without some sort of schema validation. I have been using dummy schemas in my PR and they seem to work fine. Things like https://github.com/operator-framework/operator-lifecycle-manager/pull/2228/files#diff-ba504d941f6b7c7c6d826ab787f8531904b4bc5d1edb0658a28a8dbb04ff17a8R296

PR #2119, opened 2021-04-26
Don't cache secrets not managed by olm

ecordell:
**Description of the change:**
Omits secrets that aren't owned by olm (via label-selector in cache)

**Motivation for the change:**

Before:
Screen Shot 2021-04-26 at 5 15 13 PM

After: 
Screen Shot 2021-04-26 at 5 27 12 PM

**Reviewer Checklist**
- [ ] Implementation matches the proposed design, or proposal is updated to match implementation
- [ ] Sufficient unit test coverage
- [ ] Sufficient end-to-end test coverage
- [ ] Docs updated or added to `/doc`
- [ ] Commit messages sensible and descriptive

PR #2248, opened 2021-07-09
 Filter secrets from Operator reconciler

ecordell:
#2119 changes the client-go based controllers to omit secrets from the cache that OLM didn't create. 

This attempts to do the same for the controller-runtime based controllers. It is likely not backportable due to the newer controller-runtime required

Minimize cache.

**Reviewer Checklist**


dinhxuanvu (approved):
/lgtm

PR #2249, opened 2021-07-09
Remove test dependencies on the v1beta1 CRD API.

benluddy:
Remove/update tests that are dependent on the v1beta CRD API because that version is being removed with Kubernetes 1.22.

Forked from https://github.com/operator-framework/operator-lifecycle-manager/pull/2228 but omitted the changes to on-cluster handling of deprecated GVKs.


benluddy:
The E2E tests pass on my machine against a Kubernetes 1.22 cluster.

PR #2250, opened 2021-07-09
Remove crd v1beta1 tests

ecordell:
Remove/update tests that are dependent on the v1beta CRD API because that version is being removed with Kubernetes 1.22.

- Forked from #2249 but using a newer mock-extension-apiserver that doesn't hit removed webhook apis.
  - Forked from #2228 but omitted the changes to on-cluster handling of deprecated GVKs.


benluddy:
Folded back into https://github.com/operator-framework/operator-lifecycle-manager/pull/2249.

PR #2251, opened 2021-07-12
check if server resource exists when installing OLM supported resources 

exdx:
Adds a check during installplan execution that checks for 404 errors. By querying discovery, we can establish if the GroupVersionKind if available on cluster. If not (for example, that GVK has been deprecated and removed) return a helpful error message. 

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/catalog/not_found.go
+++ b/pkg/controller/operators/catalog/not_found.go
@@ -1,0 +1,69 @@
+package catalog
+
+import (
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/discovery"
+
+	operatorsv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
+)
+
+// gvkNotFoundError is returned from installplan execution when a step contains a GVK that is not found on cluster.
+type gvkNotFoundError struct {
+	gvk  schema.GroupVersionKind
+	name string
+}
+
+func (g gvkNotFoundError) Error() string {
+	return fmt.Sprintf("api-server resource not found installing %s %s: GroupVersionKind %s not found on the cluster. %s", g.gvk.Kind, g.name, g.gvk,
+		"This API may have been deprecated and removed, see https://kubernetes.io/docs/reference/using-api/deprecation-guide/ for more information.")
+}
+
+type DiscoveryQuerier interface {
+	QueryForGVK() error
+}
+
+type DiscoveryQuerierFunc func() error
+
+func (d DiscoveryQuerierFunc) QueryForGVK() error {
+	return d()
+}
+
+type discoveryQuerier struct {
+	client discovery.DiscoveryInterface
+}
+
+func newDiscoveryQuerier(client discovery.DiscoveryInterface) *discoveryQuerier {
+	return &discoveryQuerier{client: client}
+}
+
+// WithStepResource returns a DiscoveryQuerier which uses discovery to query for supported APIs on the server based on the provided step's GVK.
+func (d *discoveryQuerier) WithStepResource(stepResource operatorsv1alpha1.StepResource) DiscoveryQuerier {
+	var f DiscoveryQuerierFunc = func() error {
+		gvk := schema.GroupVersionKind{Group: stepResource.Group, Version: stepResource.Version, Kind: stepResource.Kind}
+
+		resourceList, err := d.client.ServerResourcesForGroupVersion(gvk.GroupVersion().String())
+		if err != nil {
+			if errors.IsNotFound(err) {
+				return gvkNotFoundError{gvk: gvk, name: stepResource.Name}
+			}
+			return err
+		}
+
+		if resourceList == nil {
+			return gvkNotFoundError{gvk: gvk, name: stepResource.Name}
+		}
+
+		for _, resource := range resourceList.APIResources {
+			if resource.Kind == stepResource.Kind {
+				// this kind is supported for this particular GroupVersion
+				return nil
+			}
+		}
+
+		return gvkNotFoundError{gvk: gvk, name: stepResource.Name}
+	}
+	return f
+}
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -11,6 +11,8 @@ import (
 	"sync"
 	"time"
 
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+
 	errorwrap "github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"google.golang.org/grpc/connectivity"
@@ -1774,6 +1776,8 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
 	ensurer := newStepEnsurer(kubeclient, crclient, dynamicClient)
 	r := newManifestResolver(plan.GetNamespace(), o.lister.CoreV1().ConfigMapLister(), o.logger)
 
+	discoveryQuerier := newDiscoveryQuerier(o.opClient.KubernetesInterface().Discovery())
+
 	// CRDs should be installed via the default OLM (cluster-admin) client and not the scoped client specified by the AttenuatedServiceAccount
 	// the StepBuilder is currently only implemented for CRD types
 	// TODO give the StepBuilder both OLM and scoped clients when it supports new scoped types
@@ -2173,6 +2177,14 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
 			}
 			return nil
 		}(i, step); err != nil {
+			if k8serrors.IsNotFound(err) {
+				// Check for APIVersions present in the installplan steps that are not available on the server.
+				// The check is made via discovery per step in the plan. Transient communication failures to the api-server are handled by the plan retry logic.
+				notFoundErr := discoveryQuerier.WithStepResource(step.Resource).QueryForGVK()
+				if notFoundErr != nil {
+					return notFoundErr
+				}
+			}
 			return err
 		}
 	}
--- a/test/e2e/deprecated_e2e_test.go
+++ b/test/e2e/deprecated_e2e_test.go
@@ -1,6 +1,89 @@
 package e2e
 
-// TODO
+import (
+	"context"
+	"time"
 
-// v1beta1 CRD in installplan fails
-// v1beta1 RBAC in an installplan fails
+	"github.com/blang/semver/v4"
+	. "github.com/onsi/ginkgo"
+	"github.com/onsi/ginkgo/extensions/table"
+	. "github.com/onsi/gomega"
+	operatorsv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/operator-framework/operator-lifecycle-manager/test/e2e/ctx"
+)
+
+var missingAPI = `{"apiVersion":"verticalpodautoscalers.autoscaling.k8s.io/v1","kind":"VerticalPodAutoscaler","metadata":{"name":"my.thing","namespace":"foo"}}`
+
+var _ = Describe("Not found APIs", func() {
+	BeforeEach(func() {
+		csv := newCSV("test-csv", testNamespace, "", semver.Version{}, nil, nil, nil)
+		Expect(ctx.Ctx().Client().Create(context.TODO(), &csv)).To(Succeed())
+	})
+	AfterEach(func() {
+		TearDown(testNamespace)
+	})
+
+	When("objects with APIs that are not on-cluster are created in the installplan", func() {
+		// each entry is an installplan with a deprecated resource
+		type payload struct {
+			name       string
+			ip         *operatorsv1alpha1.InstallPlan
+			errMessage string
+		}
+
+		var tableEntries []table.TableEntry
+		tableEntries = []table.TableEntry{
+			table.Entry("contains an entry with a missing API not found on cluster ", payload{
+				name: "installplan contains a missing API",
+				ip: &operatorsv1alpha1.InstallPlan{
+					ObjectMeta: metav1.ObjectMeta{
+						Namespace: *namespace, // this is necessary due to ginkgo table semantics, see https://github.com/onsi/ginkgo/issues/378
+						Name:      "test-plan-api",
+					},
+					Spec: operatorsv1alpha1.InstallPlanSpec{
+						Approval:                   operatorsv1alpha1.ApprovalAutomatic,
+						Approved:                   true,
+						ClusterServiceVersionNames: []string{},
+					},
+					Status: operatorsv1alpha1.InstallPlanStatus{
+						Phase:          operatorsv1alpha1.InstallPlanPhaseInstalling,
+						CatalogSources: []string{},
+						Plan: []*operatorsv1alpha1.Step{
+							{
+								Resolving: "test-csv",
+								Status:    operatorsv1alpha1.StepStatusUnknown,
+								Resource: operatorsv1alpha1.StepResource{
+									Name:     "my.thing",
+									Group:    "verticalpodautoscalers.autoscaling.k8s.io",
+									Version:  "v1",
+									Kind:     "VerticalPodAutoscaler",
+									Manifest: missingAPI,
+								},
+							},
+						},
+					},
+				},
+				errMessage: "api-server resource not found installing VerticalPodAutoscaler my.thing: GroupVersionKind " +
+					"verticalpodautoscalers.autoscaling.k8s.io/v1, Kind=VerticalPodAutoscaler not found on the cluster",
+			}),
+		}
+
+		table.DescribeTable("the ip enters a failed state with a helpful error message", func(tt payload) {
+			Expect(ctx.Ctx().Client().Create(context.Background(), tt.ip)).To(Succeed())
+			Expect(ctx.Ctx().Client().Status().Update(context.Background(), tt.ip)).To(Succeed())
+
+			// The IP sits in the Installing phase with the GVK missing error
+			Eventually(func() (*operatorsv1alpha1.InstallPlan, error) {
+				return tt.ip, ctx.Ctx().Client().Get(context.Background(), client.ObjectKeyFromObject(tt.ip), tt.ip)
+			}).Should(And(HavePhase(operatorsv1alpha1.InstallPlanPhaseInstalling)), HaveMessage(tt.errMessage))
+
+			// Eventually the IP fails with the GVK missing error, after installplan retries, which is by default 1 minute.
+			Eventually(func() (*operatorsv1alpha1.InstallPlan, error) {
+				return tt.ip, ctx.Ctx().Client().Get(context.Background(), client.ObjectKeyFromObject(tt.ip), tt.ip)
+			}, 2*time.Minute).Should(And(HavePhase(operatorsv1alpha1.InstallPlanPhaseFailed)), HaveMessage(tt.errMessage))
+		}, tableEntries...)
+	})
+})
--- a/test/e2e/installplan_e2e_test.go
+++ b/test/e2e/installplan_e2e_test.go
@@ -16,7 +16,6 @@ import (
 	"github.com/onsi/ginkgo/extensions/table"
 	. "github.com/onsi/gomega"
 	. "github.com/onsi/gomega/gstruct"
-	"github.com/onsi/gomega/types"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	appsv1 "k8s.io/api/apps/v1"
@@ -53,12 +52,6 @@ import (
 )
 
 var _ = Describe("Install Plan", func() {
-	HavePhase := func(goal operatorsv1alpha1.InstallPlanPhase) types.GomegaMatcher {
-		return WithTransform(func(plan *operatorsv1alpha1.InstallPlan) operatorsv1alpha1.InstallPlanPhase {
-			return plan.Status.Phase
-		}, Equal(goal))
-	}
-
 	AfterEach(func() {
 		TearDown(testNamespace)
 	})
--- a/test/e2e/util_test.go
+++ b/test/e2e/util_test.go
@@ -35,7 +35,9 @@ import (
 	"k8s.io/client-go/rest"
 	k8scontrollerclient "sigs.k8s.io/controller-runtime/pkg/client"
 
+	gtypes "github.com/onsi/gomega/types"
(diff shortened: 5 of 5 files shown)


pkg/controller/operators/catalog/deprecated.go
```diff
@@ -0,0 +1,37 @@
+package catalog
```
exdx, line 1: Since we have IP retries now — deprecation is a terminal error. Should we categorize the error and fail the plan immediately if this comes up? Since we do not have that, the test sits around for a minute until it fails. Not a huge deal, but just thinking about it. 
Also @timflannagan  made the point around alerting — do we alert when an IP fails? should there be an alert in this PR as well? 
Lastly, in the PR now OLM still can watch for v1beta1 CRDs, Operator objects can consist of v1beta1 CRDs, should we strip all this code as well?

exdx (commented):
open questions

pkg/controller/operators/catalog/operator.go
```diff
@@ -1947,6 +1947,11 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
 						return errorwrap.Wrapf(err, "error parsing step manifest: %s", step.Resource.Name)
 					}
 
+					// check GroupVersion before proceeding -- v1beta1 RBAC resources are deprecated
+					if deprecated(cr.TypeMeta.GroupVersionKind().GroupVersion()) {
+						return deprecatedError{name: cr.Name, gv: cr.TypeMeta.GroupVersionKind().GroupVersion(), k: cr.Kind}
+					}
+
```
benluddy, line 1954: Can we check `apiVersion` once at the top instead of once per case?

pkg/controller/operators/catalog/deprecated.go
```diff
@@ -0,0 +1,37 @@
...
+func (d deprecatedError) Error() string {
+	return fmt.Sprintf("resource has been deprecated: %s of Kind %s is in a deprecated GroupVersion %s and is no longer installable with this version of OLM. %s", d.name, d.k, d.gv,
+		"See https://kubernetes.io/docs/reference/using-api/deprecation-guide/ for information on how to update to a supported version of this resource.")
+}
+
+// deprecatedGroupVersions contains a list of OLM-supported GroupVersions that have been deprecated and are now removed in kube 1.22+.
+// Any resources in these GVs are no longer installable on a 1.22+ cluster. OLM will return an install plan error if a resource in any
+// of these group/versions is found in an installplan step before attempting to create it on-cluster.
+// See https://kubernetes.io/docs/reference/using-api/deprecation-guide/ for notes on deprecated resources.
+// As additional resource groups move from deprecated to removed this list may need to be updated.
```
timflannagan, line 25: I'm still slightly hesitant around introducing this kind of failure logic upstream as I feel like we should be less restrictive here and produce warning messages and/or alerts that notify there are deprecated resources being used currently in the cluster, vs. fully blocking the creation of these resources downstream. We likely don't have the infrastructure in place and don't want to deal with the code skew, but something to think about going forward as we split up this repository. Last time I checked, we have no explicit support when it comes to OLM versions being mapped to a specific k8s version, so we might be fine and just need to notify of this change in the release notes.

pkg/controller/operators/catalog/deprecated.go
timflannagan: Ah, I had glossed over this comment when I was reviewing and made a similar comment in https://github.com/operator-framework/operator-lifecycle-manager/pull/2251/files#r669890440. tl;dr I think we shouldn't be failing the installplan resource here, but it would be difficult to deal with skew downstream, so it's an understandable approach and I'm open to hearing other opinions.

Is there a situation where we wouldn't want to immediately fail an InstallPlan? Nothing immediate comes to my mind, but I just wanted to double-check whether that assumption holds up.

pkg/controller/operators/catalog/deprecated.go
benluddy: These are all deprecated *and removed* apiVersions, right? Catalog operator already generate alerts when a user asks it to install a resource that is only deprecated.

Given that we can know at runtime whether or not these API versions are being served, can this error be generated dynamically? Can a 404 status on a POST/create response indicate anything other than the API is not being served (whether due to deprecation and removal or for some other reason)?

exdx: > Given that we can know at runtime whether or not these API versions are being served, can this error be generated dynamically? Can a 404 status on a POST/create response indicate anything other than the API is not being served (whether due to deprecation and removal or for some other reason)?

We could, but the impression I got from @kevinrizza was that we can't let the deprecated objects POST to the api-server in the first place, OLM must stop them from being created ahead of time during installplan exeuction. Otherwise the OLM SA will generate alerts on itself for installing removed APIs. 

> These are all deprecated and removed apiVersions, right?
Yes, I was debating whether to call these GroupVersions _deprecated_ or _removed_ in the code. I think deprecated makes sense because from what I can tell that signifies that they were removed, but maybe not. Can change it if it's less confusing.

pkg/controller/operators/catalog/deprecated.go
exdx: > Is there a situation where we wouldn't want to immediately fail an InstallPlan?

Yes, there are sometimes transient errors (errors connecting to the apiserver) or object has been modified errors that pop up during installplan execution. We added the retries to try to overcome these types of errors. Some errors are terminal though, and don't have to be retried - I think the idea was to build an explicit list of these at some point. 

This PR isn't really that big, like if we wanted to make this downstream only effectively it could not be such a huge burden. But it does increase the complexity going forward somewhat -- one version of OLM supports a different set of APIs than the other.

pkg/controller/operators/catalog/deprecated.go
timflannagan: Yeah, that sounds about right last time I had checked. I know there are some e2e checks in the conformance jobs that check (and count) the number of times a serviceaccount accesses a deprecated/remoted API. Those results _should_ report back as a flake so it's easy to miss that in e2e feedback.

It looks like the `[sig-arch][Late] clients should not use APIs that are removed in upcoming releases` reports which serviceaccounts/users/etc. are attempting to access that list of deprecated APIs.

timflannagan: I guess the point I was attempting to make was whether it made sense for newer/unreleased versions of upstream OLM to attempt to support a wider range of Kubernetes version, vs. making the inference that certain OLM releases only follow the minor version support path that Kubernetes itself follows. I assume we'd follow the latter given the large support burden of the former, but it would be difficult to release new OLM features to users' running older versions of Kubernetes, e.g. managed Kubernetes platforms lagging behind in minor versions.

njhale: > We could, but the impression I got from @kevinrizza was that we can't let the deprecated objects POST to the api-server in the first place, OLM must stop them from being created ahead of time during installplan exeuction. Otherwise the OLM SA will generate alerts on itself for installing removed APIs.

There's always the option of querying API discovery for the GVK to avoid sending a request to the endpoint itself. I'd prefer detecting this dynamically b/c:
- we should already be doing something similar for CR instances included in bundles
- I'm really not a fan of maintaining deny lists and managing skew

pkg/controller/operators/catalog/operator.go
```diff
@@ -2173,6 +2175,15 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
 			}
 			return nil
 		}(i, step); err != nil {
+			s := step
```
exdx, line 2178: is this doing what we expect?

test/e2e/deprecated_e2e_test.go
```diff
@@ -1,6 +1,137 @@
...
+	. "github.com/onsi/gomega"
+	operatorsv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+//go:embed testdata/deprecated/fakeCRD.yaml
+var fakeCRD string
+
+//go:embed testdata/deprecated/fakeRBAC.yaml
```
exdx, line 22: this results in the test panicking if it's called directly

pkg/controller/operators/catalog/not_found.go
```diff
@@ -0,0 +1,39 @@
...
+
+// notFound queries discovery to check supported GroupVersions on the server.
+func notFound(client discovery.DiscoveryInterface, stepResource operatorsv1alpha1.StepResource) error {
+	_, err := client.ServerResourcesForGroupVersion(schema.GroupVersion{Group: stepResource.Group, Version: stepResource.Version}.String())
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return notFoundError{gv: schema.GroupVersion{Group: stepResource.Group, Version: stepResource.Version}, k: stepResource.Kind, name: stepResource.Name}
+		}
+		return err
+	}
```
njhale, line 37: I think we'd cover a greater number of cases if we also checked for the kind.

test/e2e/deprecated_e2e_test.go
```diff
@@ -1,6 +1,137 @@
...
 
-// TODO
+import (
+	"context"
+	_ "embed"
+	"time"
 
-// v1beta1 CRD in installplan fails
-// v1beta1 RBAC in an installplan fails
+	"github.com/operator-framework/operator-lifecycle-manager/test/e2e/ctx"
```
njhale, line 8: nit: import order

test/e2e/deprecated_e2e_test.go
```diff
@@ -1,6 +1,137 @@
...
+		//_, err := crdb.Read(fakeCRD)
+		//Expect(err).To(BeNil())
+		//
+		//var rbacb bytes.Buffer
+		//_, err = crdb.Read(fakeRBAC)
+		//Expect(err).To(BeNil())
+
+		var tableEntries []table.TableEntry
+		tableEntries = []table.TableEntry{
+			table.Entry("contains a v1beta1 CRD", payload{
```
njhale, line 52: the test name no longer matches what's being tested; i.e. it mentions `v1beta1` while it actually creates something else.

test/e2e/deprecated_e2e_test.go
```diff
@@ -1,6 +1,137 @@
...
+					Status: operatorsv1alpha1.InstallPlanStatus{
+						Phase:          operatorsv1alpha1.InstallPlanPhaseInstalling,
+						CatalogSources: []string{},
+						Plan: []*operatorsv1alpha1.Step{
+							{
+								Resolving: "test-csv",
+								Status:    operatorsv1alpha1.StepStatusUnknown,
+								Resource: operatorsv1alpha1.StepResource{
+									Name:     "plums.cluster.com",
+									Version:  "v2beta5",
```
njhale, line 73: it may be far-fetched right now, but if this version is ever introduced into k8s then this test could fail.

could we use a custom group version instead?

test/e2e/deprecated_e2e_test.go
```diff
@@ -1,6 +1,137 @@
...
+		}
+
+		// TODO should we use embed string or []byte
+		//var crdb bytes.Buffer
+		//_, err := crdb.Read(fakeCRD)
+		//Expect(err).To(BeNil())
+		//
+		//var rbacb bytes.Buffer
+		//_, err = crdb.Read(fakeRBAC)
+		//Expect(err).To(BeNil())
```
njhale, line 48: nit: leftover comments

test/e2e/deprecated_e2e_test.go
```diff
@@ -1,6 +1,137 @@
...
+				ip := operatorsv1alpha1.InstallPlan{}
+				err := ctx.Ctx().Client().Get(context.Background(), client.ObjectKeyFromObject(tt.ip), &ip)
+				Expect(err).To(BeNil())
+				return ip.Status.Message
+			}).Should(ContainSubstring("resource may be deprecated"))
+
+			// Eventually the IP fails, after installplan retries, which is by default 1 minute.
+			Eventually(func() (*operatorsv1alpha1.InstallPlan, error) {
+				return tt.ip, ctx.Ctx().Client().Get(context.Background(), client.ObjectKeyFromObject(tt.ip), tt.ip)
+			}, 2*time.Minute).Should(HavePhase(operatorsv1alpha1.InstallPlanPhaseFailed))
```
njhale, line 134: nit: can these be condensed into a single `Eventually`?

test/e2e/deprecated_e2e_test.go
```diff
@@ -1,6 +1,137 @@
...
+						CatalogSources: []string{},
+						Plan: []*operatorsv1alpha1.Step{
+							{
+								Resolving: "test-csv",
+								Status:    operatorsv1alpha1.StepStatusUnknown,
+								Resource: operatorsv1alpha1.StepResource{
+									Name:     "plums.cluster.com",
+									Version:  "v2beta5",
+									Kind:     "CustomResourceDefinition",
+									Manifest: fakeCRD,
```
njhale, line 75: looks like this doesn't set the `Group`

test/e2e/deprecated_e2e_test.go
njhale: hmm, we could sidestep that issue by generating or hardcoding the manifest instead. IIUC, we really only need a manifest containing some metadata to ensure we handle non-existent GVKs; e.g.

```yaml
apiVersion: my.group/v1
kind: MyKind
metadata:
  name: my.thing
```

pkg/controller/operators/catalog/not_found.go
```diff
@@ -0,0 +1,39 @@
...
+
+// notFoundError is returned from installplan execution when a step contains a resource in a GroupVersion that is not found on the cluster.
+type notFoundError struct {
+	gv   schema.GroupVersion
+	k    string
+	name string
+}
+
+func (d notFoundError) Error() string {
+	return fmt.Sprintf("server resource not found: %s of Kind %s in GroupVersion %s is not found on this cluster. %s", d.name, d.k, d.gv,
```
njhale, line 25: The error message needs to clarify whether it's the specific resource doesn't or the entire API (GVK).

njhale (commented):
Thanks for the quick turnaround!

test/e2e/deprecated_e2e_test.go
exdx: this makes sense but the problem is we don't support arbitrary kinds, so instead of testing the behavior we are interested in we get back `v1alpha1.ErrInvalidInstallPlan`

pkg/controller/operators/catalog/not_found.go
```diff
@@ -0,0 +1,46 @@
...
+	name string
+}
+
+func (g gvkNotFoundError) Error() string {
+	return fmt.Sprintf("api-server resource not found installing %s %s: GroupVersionKind %s not found on the cluster. %s", g.gvk.Kind, g.name, g.gvk.String(),
+		"This API may have been deprecated and removed, see https://kubernetes.io/docs/reference/using-api/deprecation-guide/ for more information.")
+}
+
+// notFound queries discovery to check supported APIs on the server.
+func notFound(client discovery.DiscoveryInterface, stepResource operatorsv1alpha1.StepResource) error {
```
kevinrizza, line 25: this is a nit and I don't want to block the pr over it, but creating a global function in another file like this is probably the wrong pattern to use. it would be better if this was wrapped in a type so that we could eventually mock this in the higher order function

pkg/controller/operators/catalog/not_found.go
```diff
@@ -0,0 +1,46 @@
+package catalog
+
+import (
+	"fmt"
+
+	operatorsv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/discovery"
```
njhale, line 10: nit: import order is reversed

pkg/controller/operators/catalog/not_found.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+func (g gvkNotFoundError) Error() string {
+	return fmt.Sprintf("api-server resource not found installing %s %s: GroupVersionKind %s not found on the cluster. %s", g.gvk.Kind, g.name, g.gvk.String(),
+		"This API may have been deprecated and removed, see https://kubernetes.io/docs/reference/using-api/deprecation-guide/ for more information.")
+}
+
+// notFound queries discovery to check supported APIs on the server.
+func notFound(client discovery.DiscoveryInterface, stepResource operatorsv1alpha1.StepResource) error {
+	resourceList, err := client.ServerResourcesForGroupVersion(schema.GroupVersion{Group: stepResource.Group, Version: stepResource.Version}.String())
```
njhale, line 26: nit: I think it would be cleaner if this function either has a `schema.GroupVersionKind` parameter or declares only one up front. The type has some methods that make extracting the `GroupVersion` straight forward too: https://github.com/kubernetes/apimachinery/blob/a644435e2c133d990b624858d9c01985d7f59adf/pkg/runtime/schema/group_version.go#L157

pkg/controller/operators/catalog/not_found.go
```diff
@@ -0,0 +1,46 @@
...
+)
+
+// gvkNotFoundError is returned from installplan execution when a step contains a GVK that is not found on cluster.
+type gvkNotFoundError struct {
+	gvk  schema.GroupVersionKind
+	name string
+}
+
+func (g gvkNotFoundError) Error() string {
+	return fmt.Sprintf("api-server resource not found installing %s %s: GroupVersionKind %s not found on the cluster. %s", g.gvk.Kind, g.name, g.gvk.String(),
```
njhale, line 20: nit: `Sprintf` should automatically call the `String()` method for anything that implements the `Stringer` interface when the `%s` directive is used; i.e. `g.gvk` should suffice here

pkg/controller/operators/catalog/operator.go
njhale: I don't think aliasing this is necessary.

test/e2e/deprecated_e2e_test.go
```diff
@@ -1,6 +1,84 @@
...
+		}
+
+		table.DescribeTable("the ip enters a failed state", func(tt payload) {
+			Expect(ctx.Ctx().Client().Create(context.Background(), tt.ip)).To(Succeed())
+			Expect(ctx.Ctx().Client().Status().Update(context.Background(), tt.ip)).To(Succeed())
+
+			// Eventually the IP fails with the GVK missing error, after installplan retries, which is by default 1 minute.
+			Eventually(func() (*operatorsv1alpha1.InstallPlan, error) {
+				return tt.ip, ctx.Ctx().Client().Get(context.Background(), client.ObjectKeyFromObject(tt.ip), tt.ip)
+			}, 2*time.Minute).Should(And(HavePhase(operatorsv1alpha1.InstallPlanPhaseFailed)), HaveMessage(tt.errMessage))
```
njhale, line 81: So it takes a minute before the status shows the specific message? or does the message show up, just not in a failed phase?

njhale (commented):
Nice work!. I only have one non-nit and one question.

test/e2e/deprecated_e2e_test.go
exdx: The latter

exdx: I suppose we could just check for the existence of the message, it would make the test run a lot faster. Not sure if we also want to ensure the plan fails though

pkg/controller/operators/catalog/not_found.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+func (g gvkNotFoundError) Error() string {
+	return fmt.Sprintf("api-server resource not found installing %s %s: GroupVersionKind %s not found on the cluster. %s", g.gvk.Kind, g.name, g.gvk,
+		"This API may have been deprecated and removed, see https://kubernetes.io/docs/reference/using-api/deprecation-guide/ for more information.")
+}
+
+type DiscoveryQuerier interface {
+	QueryForGVK() error
+}
```
timflannagan, line 26: :+1:

pkg/controller/operators/catalog/not_found.go
```diff
@@ -0,0 +1,69 @@
...
+)
+
+// gvkNotFoundError is returned from installplan execution when a step contains a GVK that is not found on cluster.
+type gvkNotFoundError struct {
+	gvk  schema.GroupVersionKind
+	name string
+}
+
+func (g gvkNotFoundError) Error() string {
+	return fmt.Sprintf("api-server resource not found installing %s %s: GroupVersionKind %s not found on the cluster. %s", g.gvk.Kind, g.name, g.gvk,
```
timflannagan, line 20: not blocking/nit: might be easier for readability to move the inline string definition as it's own type.

Also may need to add a `... not found when installing the %s %s`.

/lgtm

pkg/controller/operators/catalog/not_found.go
```diff
@@ -0,0 +1,69 @@
...
+				return gvkNotFoundError{gvk: gvk, name: stepResource.Name}
+			}
+			return err
+		}
+
+		if resourceList == nil {
+			return gvkNotFoundError{gvk: gvk, name: stepResource.Name}
+		}
+
+		for _, resource := range resourceList.APIResources {
```
timflannagan, line 59: I'm not super familiar with using the ServerResourcesForGroupVersion function -- do we need to check the len of the APIResources field if it's zero elements, and if so exit early, or do we implicitly get an error further down the control flow?

timflannagan:
Feel free to remove the hold if no one else needs to review.

/hold

pkg/controller/operators/catalog/not_found.go
exdx: It's its non-nil and length zero then we get back the error (GVK does not exist)

timflannagan: Makes sense - thanks.

pkg/controller/operators/catalog/not_found.go
```diff
@@ -0,0 +1,69 @@
...
+	gvk  schema.GroupVersionKind
+	name string
+}
+
+func (g gvkNotFoundError) Error() string {
+	return fmt.Sprintf("api-server resource not found installing %s %s: GroupVersionKind %s not found on the cluster. %s", g.gvk.Kind, g.name, g.gvk,
+		"This API may have been deprecated and removed, see https://kubernetes.io/docs/reference/using-api/deprecation-guide/ for more information.")
+}
+
+type DiscoveryQuerier interface {
```
njhale, line 24: Not blocking:

alternative

```go
type GVKChecker interface {
  Exists(schema.GroupVersionKind) (ok bool, err error)
}
```

njhale (approved):
I'm not sure how I feel about the design of the interfaces added in this PR, but I don't think it's worth holding this PR on.

The e2e test is verifying the behavior well, but I'd eventually like a follow up with some unit tests that make use of the newly added interface.

/hold cancel

PR #2254, opened 2021-07-13
Translate legacy "bundle dependencies" to properties.

benluddy:
Operator dependencies are expressed as properties, but there was
initially a notion of dependencies as distinct from properties. For
that reason, catalog-operator must support existing catalog images
that serve separate Bundle.dependencies and Bundle.properties fields
by translating the three legacy dependency types (olm.label,
olm.package, and olm.gvk) into their corresponding properties.


ecordell (commented):
couple of small comments, otherwise looks good.

This change makes me think we want to prioritize a story for "property-only" updates of operators. I think we would want to have existing CSVs get these annotations added to them without any other changes. (and there are other reasons we might want that too with declarative config)

The main blocker for that is just identity - but I think we could probably pick the the properties we use for identity today (package, name, version, channel) and use that as the identifier and potentially work on something a bit more solid in the future.

PR #2258, opened 2021-07-15
feat(resolver): Indicate dependency class in resolution constraint text 

anik120:

This PRintroduces a `String()` method to the Predicate interface, implementations
of which are used to build clauses that are fed to the sat solver. Calling the
predicate.String() method enhances the error messages displayed when dependency
resolution fails.

For example, an error message previously due to an unsatisfyiable contraint was
diplayed as

```
ResolutionFailed' constraints not satisfiable: bundle etcdoperator-community.v0.6.1 has
a dependency without any candidates to satisfy it, subscription etcd requires
my-catalog/olm/alpha/etcdoperator-community.v0.6.1, subscription etcd exists
```

With this commit, the same error is displayed as:

```
ResolutionFailed' constraints not satisfiable: bundle etcdoperator-community.v0.6.1 requires
an operator with package:foo and with version in range:>0.27.0, subscription etcd requires
my-catalog/olm/alpha/etcdoperator-community.v0.6.1, subscription etcd exists
```

**Reviewer Checklist**


ankitathomas (approved):
/lgtm

pkg/controller/registry/resolver/predicates.go
```diff
@@ -0,0 +1,376 @@
...
+
+type OperatorPredicateFunc func(*Operator) bool
+
+func (opf OperatorPredicateFunc) Test(o *Operator) bool {
+	return opf(o)
+}
+
+func (opf OperatorPredicateFunc) String() string {
+	return ""
+}
```
benluddy, line 296: Now that predicates must provide a human-readable string representation, this type can probably be removed altogether. That avoids the risk down the line of a user seeing an empty string and becoming confused.

pkg/controller/registry/resolver/predicates.go
```diff
@@ -0,0 +1,376 @@
...
+		}
+		if prop.Kind == g.api.Kind && prop.Version == g.api.Version && prop.Group == g.api.Group {
+			return true
+		}
+	}
+	return false
+}
+
+func (g gvkPredicate) String() string {
+	return fmt.Sprintf("with group: %s, version: %s, kind: %s", g.api.Group, g.api.Version, g.api.Kind)
```
benluddy, line 186:
```suggestion
	return fmt.Sprintf("providing an API with group: %s, version: %s, kind: %s", g.api.Group, g.api.Version, g.api.Kind)
```

pkg/controller/registry/resolver/predicates.go
```diff
@@ -0,0 +1,376 @@
...
+func CatalogPredicate(key registry.CatalogKey) OperatorPredicate {
+	return catalogPredicate{key: key}
+}
+
+func (c catalogPredicate) Test(o *Operator) bool {
+	return c.key.Equal(o.SourceInfo().Catalog)
+}
+
+func (c catalogPredicate) String() string {
+	return fmt.Sprintf("with catalog: %v/%v", c.key.Namespace, c.key.Name)
```
benluddy, line 155:
```suggestion
	return fmt.Sprintf("from catalog: %v/%v", c.key.Namespace, c.key.Name)
```

pkg/controller/registry/resolver/predicates.go
```diff
@@ -0,0 +1,376 @@
...
+	})
+}
+
+func False() OperatorPredicate {
+	return OperatorPredicateFunc(func(*Operator) bool {
+		return false
+	})
+}
+
+func CountingPredicate(p OperatorPredicate, n *int) OperatorPredicate {
```
benluddy, line 310: If you decide to remove `OperatorPredicateFunc`, I'd expect the String() implementation for a counting predicate to delegate to the predicate that it wraps -- the counter doesn't affect the meaning of the predicate.

benluddy (commented):
Looks great!

pkg/controller/registry/resolver/predicates.go
anik120: @benluddy I'd explored removing this type for a `genericPredicate` instead, but it seemed to do the same thing, with it's string representation being `""`. The Generic Predicate would be required to implement predicates like the [True](https://github.com/operator-framework/operator-lifecycle-manager/pull/2258/files#diff-a5735698ad5d9d46d8568de6aaddf740642e9c19598888768c68db7d1eec1fe1R298), [False](https://github.com/operator-framework/operator-lifecycle-manager/pull/2258/files#diff-a5735698ad5d9d46d8568de6aaddf740642e9c19598888768c68db7d1eec1fe1R304) predicate etc

pkg/controller/registry/resolver/predicates.go
anik120: I'd explored this too, but at the heart of the counting predicate it's calling p.Test() for the predicate it receives, which requires the `resolver.Operator` as it's argument, which is not available statically, i.e the `OperatorPredicate` func type that's implemented within this method does make sense.

PR #2260, opened 2021-07-16
Explain unique replaces chain requirement in channel sort errors.

benluddy:
The "multiple channel head" error that can be returned from channel
sorting has been a source of confusion for users. The error message
now explains that a unique replacement chain is required in order to
define the relative order between channel entries, and it also
provides an abbreviated list of all replacement chains identified, in
the form "head...tail" or "singleton".

### DIFF
--- a/pkg/controller/registry/resolver/resolver_test.go
+++ b/pkg/controller/registry/resolver/resolver_test.go
@@ -1,6 +1,7 @@
 package resolver
 
 import (
+	"errors"
 	"fmt"
 	"testing"
 
@@ -21,11 +22,11 @@ import (
 )
 
 func TestSolveOperators(t *testing.T) {
-	APISet := APISet{opregistry.APIKey{"g", "v", "k", "ks"}: struct{}{}}
+	APISet := APISet{opregistry.APIKey{Group: "g", Version: "v", Kind: "k", Plural: "ks"}: struct{}{}}
 	Provides := APISet
 
-	namespace := "olm"
-	catalog := registry.CatalogKey{"community", namespace}
+	const namespace = "test-namespace"
+	catalog := registry.CatalogKey{Name: "test-catalog", Namespace: namespace}
 
 	csv := existingOperator(namespace, "packageA.v1", "packageA", "alpha", "", Provides, nil, nil, nil)
 	csvs := []*v1alpha1.ClusterServiceVersion{csv}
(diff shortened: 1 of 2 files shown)


benluddy:
/hold

This is a WIP, and may have also identified a bug with one of the new tests.

pkg/controller/registry/resolver/resolver_test.go
```diff
@@ -1742,3 +1743,166 @@ func TestInferProperties(t *testing.T) {
...
+					bundle: &api.Bundle{
+						PackageName: "package",
+						ChannelName: "channel",
+					},
+				},
+			},
+			Err: errors.New(`a cycle exists in the chain of replacement beginning with "a" in channel "channel" of package "package"`),
+		},
+		{
+			Name: "skipped entry omitted",
```
benluddy, line 1849: This test represents my understanding of the intended behavior of skips here, but it's failing. Is this a bug in the sort or am I mistaken about how this should work?

benluddy: Fixed this and added a higher-level test case (TestSolveOperatorsWithSkipsPreventingSelection) to cover the skips omission behavior.

pkg/controller/registry/resolver/resolver_test.go
```diff
@@ -141,10 +129,9 @@ func TestSolveOperators_MultipleChannels(t *testing.T) {
 	operators, err := satResolver.SolveOperators([]string{"olm"}, csvs, subs)
 	assert.NoError(t, err)
 	expected := OperatorSet{
-		"packageA.v1": genOperator("packageA.v1", "0.0.1", "packageA.v1", "packageA", "alpha", "community", "olm", nil, nil, nil, "", false),
 		"packageB.v1": genOperator("packageB.v1", "1.0.0", "", "packageB", "alpha", "community", "olm", nil, nil, nil, "", false),
 	}
-	assert.Equal(t, 2, len(operators))
+	assert.Len(t, operators, 1)
```
ecordell, line 134: discussed offline: this change confused me, but the old assertions were incorrect because packageA.v1 replaced itself

ecordell (approved):
/lgtm

also discussed offline the merits of coupling the message change with the test fixes; but both can probably reasonably be backported the same degree

jianzhangbjz:
@benluddy @kevinrizza Will it be cherry-picked to the downstream repo?  If yes, when? I checked the latest commits(1dc76f08ed05a635458420ffa979aebbe59a3890) of the downstream repo, but not found this PR.  I am really interested in the process of picking up upstream PRs for the downstream.  Could you help explain more about it? Or any doc? Thanks!

PR #2261, opened 2021-07-16
 Correctly omit skipped channel entries during resolution.

benluddy:
A behavior was added with the intention of preventing the installation
of channel entries that are skipped by other entries, but in practice
it was omitting entries that were replaced by skipped entries.

This is fixed. There are now unit test cases and one higher-level test
for this behavior.


benluddy:
This fix applies on top of changes introduced in https://github.com/operator-framework/operator-lifecycle-manager/pull/2260 so it may need to be rebased once that merges.

PR #2262, opened 2021-07-16
feat(sub): Include IP failure condition message in sub status condition

anik120:

Currently when an InstallPlan fails, due to bundle unpacking error
is pending due to errors like invalid operatorgroup, or pending bundle unpacking job
the reason is propagated to the Subscription that owns it. The message for the failure is
however missing from the Subscription condition.

eg

```
kind: Subscription
status:
conditions:
    - lastTransitionTime: "2021-07-07T17:55:20Z"
      reason: Installing
      status: "True"
      type: InstallPlanPending
```

This PR propagates the message assosiated with the reason in the InstallPlan
condition when an InstallPlan is either pending, or has failed permanantly.

eg

```
kind: Subscription
status:
conditions:
    - lastTransitionTime: "2021-07-07T17:55:20Z"
      message: no operator group found that is managing this namespace
      reason: Installing
      status: "True"
      type: InstallPlanPending
```

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/catalog/subscription/state.go
+++ b/pkg/controller/operators/catalog/subscription/state.go
@@ -1,6 +1,7 @@
 package subscription
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 
@@ -452,7 +453,7 @@ func (i *installPlanReferencedState) CheckInstallPlanStatus(now *metav1.Time, cl
 		if cond.Reason == "" {
 			cond.Reason = string(phase)
 		}
-
+		cond.Message = extractMessage(status)
 		cond.Type = v1alpha1.SubscriptionInstallPlanPending
 		cond.Status = corev1.ConditionTrue
 		out.Status.SetCondition(cond)
@@ -472,6 +473,7 @@ func (i *installPlanReferencedState) CheckInstallPlanStatus(now *metav1.Time, cl
 		}
 
 		cond.Type = v1alpha1.SubscriptionInstallPlanFailed
+		cond.Message = extractMessage(status)
 		cond.Status = corev1.ConditionTrue
 		out.Status.SetCondition(cond)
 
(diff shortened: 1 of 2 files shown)


ankitathomas (approved):
/lgtm

anik120:
/hold 
have to review the test case.

ankitathomas (commented):
/lgtm

pkg/controller/operators/catalog/subscription/state.go
```diff
@@ -508,6 +510,26 @@ func (i *installPlanReferencedState) CheckInstallPlanStatus(now *metav1.Time, cl
...
+			if cond := lookup.GetCondition(v1alpha1.BundleLookupPending); cond.Status != corev1.ConditionUnknown {
+				b.WriteString(cond.Message)
+				b.WriteString(".")
+			}
+		}
+		str = b.String()
+	}
+	if cond := status.GetCondition(v1alpha1.InstallPlanInstalled); cond.Status != corev1.ConditionUnknown {
+		if cond.Message != "" {
+			str = cond.Message
```
awgreene, line 527: Is the assumption here that if an IP is installed no bundleLookupPending messages matter anymore? I assume the bundleLookupPending condition is removed after a successful install?

pkg/controller/operators/catalog/subscription/state.go
```diff
@@ -508,6 +510,26 @@ func (i *installPlanReferencedState) CheckInstallPlanStatus(now *metav1.Time, cl
...
 }
 
+func extractMessage(status *v1alpha1.InstallPlanStatus) string {
+	str := ""
+	if len(status.BundleLookups) > 0 {
+		var b bytes.Buffer
+		for _, lookup := range status.BundleLookups {
+			if cond := lookup.GetCondition(v1alpha1.BundleLookupPending); cond.Status != corev1.ConditionUnknown {
+				b.WriteString(cond.Message)
+				b.WriteString(".")
```
awgreene, line 520: Why add the `.` here rather than update the message itself?

awgreene (approved):
Nice work @anik120

PR #2263, opened 2021-07-16
Fix ClusterOperator selector for excluding copied CSVs.

benluddy:
The function that generated the selector had been returning a selector
that matches everything. There was no functional impact because
there's an additional client-side filter that discards objects based
on CSV.IsCopied().


dinhxuanvu (approved):
/lgtm

PR #2264, opened 2021-07-16
Don't cache copied CSVs in the controller-runtime-based controllers.

benluddy:
Only the main controller (i.e., the controller responsible for
reconciling ClusterServiceVersions) needs to watch copied
ClusterServiceVersions. The OperatorCondition and Operator controllers
-- which coincidentally maintain separate caches -- are not interested
in copied CSVs, so all copied CSVs can be excluded via from their
cache using a label selector.

### DIFF
--- a/cmd/olm/manager.go
+++ b/cmd/olm/manager.go
@@ -2,30 +2,57 @@ package main
 
 import (
 	"context"
-	"k8s.io/apimachinery/pkg/labels"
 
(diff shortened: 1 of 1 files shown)


dinhxuanvu:
Will lgtm but add a hold so the PR doesn't get merged prematurely.
/hold

dinhxuanvu (approved):
/lgtm

PR #2268, opened 2021-07-19
Update extractMessage

awgreene:
When building the subscription condition message,
check if the status of the InstallPlan is available
before iterating through bundle lookup conditions.

### DIFF
--- a/pkg/controller/operators/catalog/subscription/state.go
+++ b/pkg/controller/operators/catalog/subscription/state.go
@@ -511,23 +511,22 @@ func (i *installPlanReferencedState) CheckInstallPlanStatus(now *metav1.Time, cl
 }
 
 func extractMessage(status *v1alpha1.InstallPlanStatus) string {
-	str := ""
-	if len(status.BundleLookups) > 0 {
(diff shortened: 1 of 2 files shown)


anik120 (approved):
/lgtm

pkg/controller/operators/catalog/subscription/state.go
```diff
@@ -511,23 +511,24 @@ func (i *installPlanReferencedState) CheckInstallPlanStatus(now *metav1.Time, cl
...
-			}
-		}
-		str = b.String()
-	}
 	if cond := status.GetCondition(v1alpha1.InstallPlanInstalled); cond.Status != corev1.ConditionUnknown {
 		if cond.Message != "" {
-			str = cond.Message
+			return cond.Message
+		}
+	}
```
anik120, line 518: why not contract this block too?  
```
if cond := status.GetCondition(v1alpha1.InstallPlanInstalled); cond.Status != corev1.ConditionUnknown && cond.Message != "" {
	return cond.Message
}
```

PR #2269, opened 2021-07-19
feat(sub): Indicate resolution conflicts on Subscription statuses

anik120:

Resolution considers everything in a given namespace as a unit, so conflicts aren't
generated on a per-Subscription basis. However, as the main entry point to resolution,
users first look at the Subscription to understand resolution failures. The set of conflicts
today is written to Events with type ResolutionFailed. This PR projects resolution errors onto
all subscriptions on the namespace for any resolution failure.

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -909,9 +909,22 @@ func (o *Operator) syncResolvingNamespace(obj interface{}) error {
 		// not-satisfiable error
 		if _, ok := err.(solver.NotSatisfiable); ok {
 			logger.WithError(err).Debug("resolution failed")
+			updateErr := o.setSubsCond(subs, v1alpha1.SubscriptionResolutionFailed, "ConstraintsNotSatisfiable", err.Error(), true)
+			if updateErr != nil {
+				logger.WithError(updateErr).Debug("failed to update subs conditions")
+			}
 			return nil
 		}
+		updateErr := o.setSubsCond(subs, v1alpha1.SubscriptionResolutionFailed, "ErrorPreventedResolution", err.Error(), true)
+		if updateErr != nil {
+			logger.WithError(updateErr).Debug("failed to update subs conditions")
+		}
 		return err
+	} else {
+		updateErr := o.setSubsCond(subs, v1alpha1.SubscriptionResolutionFailed, "", "", false)
+		if updateErr != nil {
+			logger.WithError(updateErr).Debug("failed to update subs conditions")
+		}
 	}
 
 	// create installplan if anything updated
@@ -1187,6 +1200,55 @@ func (o *Operator) createInstallPlan(namespace string, gen int, subs []*v1alpha1
 	}
 
 	return reference.GetReference(res)
+}
+
+func (o *Operator) setSubsCond(subs []*v1alpha1.Subscription, condType v1alpha1.SubscriptionConditionType, reason, message string, setTrue bool) error {
+	var (
+		errs        []error
+		mu          sync.Mutex
+		wg          sync.WaitGroup
+		getOpts     = metav1.GetOptions{}
+		updateOpts  = metav1.UpdateOptions{}
+		lastUpdated = o.now()
+	)
+	for _, sub := range subs {
+		sub.Status.LastUpdated = lastUpdated
+		cond := sub.Status.GetCondition(condType)
+		cond.Reason = reason
+		cond.Message = message
+		if setTrue {
+			cond.Status = corev1.ConditionTrue
+		} else {
+			cond.Status = corev1.ConditionFalse
+		}
+		sub.Status.SetCondition(cond)
+
+		wg.Add(1)
+		go func(s v1alpha1.Subscription) {
+			defer wg.Done()
+
+			update := func() error {
+				// Update the status of the latest revision
+				latest, err := o.client.OperatorsV1alpha1().Subscriptions(s.GetNamespace()).Get(context.TODO(), s.GetName(), getOpts)
+				if err != nil {
+					return err
+				}
+
+				latest.Status = s.Status
+				_, err = o.client.OperatorsV1alpha1().Subscriptions(sub.Namespace).UpdateStatus(context.TODO(), latest, updateOpts)
+
+				return err
+			}
+			if err := retry.RetryOnConflict(retry.DefaultRetry, update); err != nil {
+				mu.Lock()
+				defer mu.Unlock()
+				errs = append(errs, err)
+			}
+		}(*sub)
+	}
+	wg.Wait()
+
+	return utilerrors.NewAggregate(errs)
 }
 
 type UnpackedBundleReference struct {
--- a/test/e2e/subscription_e2e_test.go
+++ b/test/e2e/subscription_e2e_test.go
@@ -2105,6 +2105,60 @@ var _ = Describe("Subscription", func() {
 		require.NoError(GinkgoT(), err)
 	})
 
+	When("A subscription is created for an operator that requires an API that is not available", func() {
+
+		var (
+			c          operatorclient.ClientInterface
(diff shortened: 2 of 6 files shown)


anik120:
/hold

Made changes to the subscription api inside the vendor folder for reviews. PR for api change is here: git push --force-with-lease origin sub-resolution-error

will unhold once this and the api repo PR is lgtmed and changes from api is vendored in here.

pkg/controller/operators/catalog/operator.go
```diff
@@ -909,9 +909,22 @@ func (o *Operator) syncResolvingNamespace(obj interface{}) error {
...
+				logger.WithError(updateErr).Debug("failed to update subs conditions")
+			}
 			return nil
 		}
+		updateErr := o.setSubsResFailedCond(subs, "Dependency resolution failed", err.Error(), true)
+		if updateErr != nil {
+			logger.WithError(updateErr).Debug("failed to update subs conditions")
+		}
 		return err
+	} else {
```
anik120, line 923: Question: This is under the assumption that if there is a terminal resolution error, user intervention can resolve the error and get the subscriptions in the objects back to a good state. Is that a correct assumption?

test/e2e/subscription_e2e_test.go
```diff
@@ -2105,6 +2105,55 @@ var _ = Describe("Subscription", func() {
...
+		})
+
+		AfterEach(func() {
+			cleanup()
+			teardown()
+		})
+
+		It("the subscription has a condition in it's status that indicates the resolution error", func() {
+			_, err := fetchSubscription(crc, testNamespace, subName, subscriptionHasResolutionFailedCondChecker)
+			require.NoError(GinkgoT(), err)
```
njhale, line 2152: If we can, we should directly use ginko/gomega for net-new tests -- something like:

```go
Eventually(func() (corev1.ConditionStatus, error) {
  sub, err = crc.OperatorsV1alpha1().Subscriptions(namespace).Get(context.Background(), name, metav1.GetOptions{})
  if err != nil {
    return corev1.ConditionUnknown, err
  }
  return sub.Status.GetCondition(operatorsv1alpha1.SubscriptionResolutionFailed), nil
}).Should(Equal(corev1.ConditionTrue))
```

this keeps us from building deeper into our old homespun test frameworks.

test/e2e/subscription_e2e_test.go
```diff
@@ -2488,6 +2537,20 @@ func subscriptionStateUpgradeAvailableChecker(subscription *operatorsv1alpha1.Su
...
 }
 
+func subscriptionHasResolutionFailedCondChecker(subscription *operatorsv1alpha1.Subscription) bool {
+	if cond := subscription.Status.GetCondition(operatorsv1alpha1.SubscriptionResolutionFailed); cond.Status == corev1.ConditionTrue {
+		return true
+	}
+	return false
+}
+
+func subscriptionDoesNotHaveResolutionFailedCondChecker(subscription *operatorsv1alpha1.Subscription) bool {
```
njhale, line 2547: looks like this function isn't being used anywhere

njhale (commented):
Looking great so far.

pkg/controller/operators/catalog/operator.go
```diff
@@ -1189,6 +1202,55 @@ func (o *Operator) createInstallPlan(namespace string, gen int, subs []*v1alpha1
 	return reference.GetReference(res)
 }
 
+func (o *Operator) setSubsResFailedCond(subs []*v1alpha1.Subscription, reason, message string, setTrue bool) error {
```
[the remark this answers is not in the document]
anik120: It's being set to `false` here https://github.com/operator-framework/operator-lifecycle-manager/pull/2269/commits/bc1d24328c4ceeca4901d00c94e3890bc0f278b4#diff-a1760d9b7ac1e93734eea675d8d8938c96a50e995434b163c6f77c91bace9990R924

test/e2e/subscription_e2e_test.go
anik120: Removing both of these functions while changing the test according to https://github.com/operator-framework/operator-lifecycle-manager/pull/2269#discussion_r674067049

pkg/controller/operators/catalog/operator.go
njhale: cluster admins can always nuke everything and recreate after the fact. there are also (rather gnarly) workarounds to pop an InstallPlan back into a non-terminal state for a retry (e.g. manually patch InstallPlan status).

njhale (commented):
Nice!

My only real concern is in the lack of unit tests.

anik120:
@njhale  there are tests [here](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/controller/operators/catalog/operator_test.go#L1071) that tests that the method returns the correct value/error when resolver error occurs.
Example the [NotSatisfiableError test](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/controller/operators/catalog/operator_test.go#L1117) checks that the sovler.NotSatisfiable error is returned when a constraint is not satisfyiable. And there's [another unit test](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/controller/operators/catalog/operator_test.go#L1181) to test any resolver error that is not a constraint not satisfyiable error. So we have unit tests that covers all scenarios for resolver error during namespace sync, this PR just bubbles the errors up to the subscriptions, which needed the e2e test to test

njhale:
> @njhale there are tests [here](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/controller/operators/catalog/operator_test.go#L1071) that tests that the method returns the correct value/error when resolver error occurs.
> Example the [NotSatisfiableError test](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/controller/operators/catalog/operator_test.go#L1117) checks that the sovler.NotSatisfiable error is returned when a constraint is not satisfyiable. And there's [another unit test](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/controller/operators/catalog/operator_test.go#L1181) to test any resolver error that is not a constraint not satisfyiable error. So we have unit tests that covers all scenarios for resolver error during namespace sync, this PR just bubbles the errors up to the subscriptions, which needed the e2e test to test

hmm, I was thinking of the class of test that uses a mock client and tests a wider range of I/O (i.e. not just checking the error) -- but since the e2e test is so similar, I don't think it's worth holding so close to freeze.

/approve

pkg/controller/operators/catalog/operator.go
```diff
@@ -909,9 +909,22 @@ func (o *Operator) syncResolvingNamespace(obj interface{}) error {
 		// not-satisfiable error
 		if _, ok := err.(solver.NotSatisfiable); ok {
 			logger.WithError(err).Debug("resolution failed")
+			updateErr := o.setSubsCond(subs, v1alpha1.SubscriptionResolutionFailed, "Constraint not satisfyable", err.Error(), true)
```
kevinrizza, line 912: nit: spelling error. this should say `Constraint not satisfiable`

kevinrizza:
minus that one nit, this looks good to me

pkg/controller/operators/catalog/operator.go
```diff
@@ -909,9 +909,22 @@ func (o *Operator) syncResolvingNamespace(obj interface{}) error {
 		// not-satisfiable error
 		if _, ok := err.(solver.NotSatisfiable); ok {
 			logger.WithError(err).Debug("resolution failed")
+			updateErr := o.setSubsCond(subs, v1alpha1.SubscriptionResolutionFailed, "Constraint not satisfiable", err.Error(), true)
```
benluddy, line 912: Isn't Reason usually intended to be machine-readable?

kevinrizza: nice catch

PR #2270, opened 2021-07-20
fix(openshift): handle unquoted short max versions

njhale:
## Description

Add special handling for version shorthand commonly used in pipelines for Red Hat operator catalogs. 

Allows unquoted short versions to be used when specifying an operator's OpenShift compatibility; e.g. olm.maxOpenShiftVersion: 4.8. Without this, OLM will only recognize quoted, fully qualified, maxOpenShiftVersion properties; e.g. olm.maxOpenShiftVersion: "4.8.0".

## Motivation

Under my direction, Red Hat catalog pipeline maintainers have already notified bundle authors of, built automation around, and updated catalog content using this shorthand. Given this, supporting the shorthand will save these clients the effort and confusion required to switch to a fully qualified version.

### DIFF
--- a/pkg/controller/operators/openshift/clusteroperator_controller_test.go
+++ b/pkg/controller/operators/openshift/clusteroperator_controller_test.go
@@ -1,6 +1,8 @@
 package openshift
 
 import (
+	"fmt"
+
 	semver "github.com/blang/semver/v4"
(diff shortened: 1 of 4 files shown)


dinhxuanvu (approved):
/lgtm

PR #2271, opened 2021-07-20
Add filter labels to catalog pod cache

dinhxuanvu:
Add catalogsource and update labels to pod cache to filter out
the pods that are not managed by catalog-operator in OLM. The
intention is to reduce the amount of cache for pod resources.

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -26,6 +26,7 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/selection"
 	utilclock "k8s.io/apimachinery/pkg/util/clock"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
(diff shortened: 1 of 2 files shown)


pkg/controller/operators/catalog/operator.go
```diff
@@ -311,7 +312,20 @@ func NewOperator(ctx context.Context, kubeconfigPath string, clock utilclock.Clo
...
+	if err != nil {
+		return nil, err
+	}
+	catsrcUpdateReq, err := labels.NewRequirement(reconciler.CatalogSourceUpdateKey, selection.Exists, nil)
+	if err != nil {
+		return nil, err
+	}
+	podLabels := labels.NewSelector()
+	podLabels.Add(*catsrcReq)
+	podLabels.Add(*catsrcUpdateReq)
```
benluddy, line 325: `Add` returns a new copy of the selector with the requirement added, so these aren't actually modifying `podLabels` (it will match everything).

https://github.com/kubernetes/apimachinery/blob/v0.21.3/pkg/labels/selector.go#L365-L376

benluddy:
/hold

:disappointed: https://github.com/openshift/operator-framework-olm/blob/master/staging/operator-lifecycle-manager/pkg/controller/bundle/bundle_unpacker.go#L509

ecordell:
/approve

worth thinking about a followup that changes the label from `job-name` - it really should be namespaced

PR #2272, opened 2021-07-20
Remove unused scripts

ecordell:
Removes some scripts that are no longer used.

kube_deps is not used for pinning kube dependencies anymore
the other scripts were used for older ways that olm integrated with
openshift and are no longer needed

**Reviewer Checklist**


timflannagan:
Manually overriding the DCO check - as an aside, we need to add a list of required status contexts for this repositories tide configuration.

PR #2275, opened 2021-07-21
Update catalog deployment template

awgreene:
The catalog deployment template only sets the
readiness and liveness probes to HTTPS Scheme if the
values.catalog.tlsKeyPath and values.catalog.tlsCertPath
configurations are provided. These values should be set
when the values.catalog.tlsSecret or
values.catalog.clientCASecret configurations are set as
it is done in the olm deployment template.

### DIFF
--- a/deploy/chart/templates/0000_50_olm_08-catalog-operator.deployment.yaml
+++ b/deploy/chart/templates/0000_50_olm_08-catalog-operator.deployment.yaml
@@ -81,12 +81,12 @@ spec:
             httpGet:
               path: /healthz
               port: {{ .Values.catalog.service.internalPort }}
-              scheme: {{ if and .Values.catalog.tlsKeyPath .Values.catalog.tlsCertPath }}HTTPS{{ else }}HTTP{{end}}
+              scheme: {{ if .Values.catalog.tlsSecret }}HTTPS{{ else }}HTTP{{end}}
           readinessProbe:
             httpGet:
               path: /healthz
               port: {{ .Values.catalog.service.internalPort }}
-              scheme: {{ if and .Values.catalog.tlsKeyPath .Values.catalog.tlsCertPath }}HTTPS{{ else }}HTTP{{end}}
+              scheme: {{ if .Values.catalog.tlsSecret }}HTTPS{{ else }}HTTP{{end}}
           terminationMessagePolicy: FallbackToLogsOnError
           {{- if .Values.catalog.resources }}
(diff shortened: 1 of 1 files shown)


deploy/chart/templates/0000_50_olm_08-catalog-operator.deployment.yaml
```diff
@@ -81,12 +81,12 @@ spec:
             httpGet:
               path: /healthz
               port: {{ .Values.catalog.service.internalPort }}
-              scheme: {{ if and .Values.catalog.tlsKeyPath .Values.catalog.tlsCertPath }}HTTPS{{ else }}HTTP{{end}}
+              scheme: {{ if or .Values.olm.tlsSecret .Values.catalog.clientCASecret }}HTTPS{{ else }}HTTP{{end}}
```
ecordell, line 84: shouldn't this be `.Values.catalog.tlsSecret`?

awgreene: Yes - sorry working late at night.

awgreene: Addressed in the latest version of the pr

benluddy (commented):
I don't think the presence of a client CA bundle should be sufficient to flip the template. It's the serving cert that makes it possible to run in HTTPS mode.

awgreene:
@benluddy addressed with the latest changes. Updated commit message.

deploy/chart/templates/0000_50_olm_08-catalog-operator.deployment.yaml
```diff
@@ -81,12 +81,12 @@ spec:
             httpGet:
               path: /healthz
               port: {{ .Values.catalog.service.internalPort }}
-              scheme: {{ if and .Values.catalog.tlsKeyPath .Values.catalog.tlsCertPath }}HTTPS{{ else }}HTTP{{end}}
+              scheme: {{ if or .Values.catalog.tlsSecret }}HTTPS{{ else }}HTTP{{end}}
```
benluddy, line 84:
```suggestion
              scheme: {{ if .Values.catalog.tlsSecret }}HTTPS{{ else }}HTTP{{end}}
```

awgreene: Addressed.

PR #2276, opened 2021-07-21
Update OLM Deployment template

awgreene:
The olm deployment template will only use HTTPS when values.olm.tlsSecret is set
in the values.yaml file.

### DIFF
--- a/deploy/chart/templates/0000_50_olm_07-olm-operator.deployment.yaml
+++ b/deploy/chart/templates/0000_50_olm_07-olm-operator.deployment.yaml
@@ -87,12 +87,12 @@ spec:
             httpGet:
               path: /healthz
               port: {{ .Values.olm.service.internalPort }}
-              scheme: {{ if or .Values.olm.tlsSecret .Values.olm.clientCASecret }}HTTPS{{ else }}HTTP{{end}}
+              scheme: {{ if .Values.olm.tlsSecret }}HTTPS{{ else }}HTTP{{end}}
           readinessProbe:
             httpGet:
               path: /healthz
               port: {{ .Values.olm.service.internalPort }}
-              scheme: {{ if or .Values.olm.tlsSecret .Values.olm.clientCASecret }}HTTPS{{ else }}HTTP{{end}}
+              scheme: {{ if .Values.olm.tlsSecret }}HTTPS{{ else }}HTTP{{end}}
           terminationMessagePolicy: FallbackToLogsOnError
           env:
           - name: OPERATOR_NAMESPACE


deploy/chart/templates/0000_50_olm_07-olm-operator.deployment.yaml
```diff
@@ -87,12 +87,12 @@ spec:
             httpGet:
               path: /healthz
               port: {{ .Values.olm.service.internalPort }}
-              scheme: {{ if or .Values.olm.tlsSecret .Values.olm.clientCASecret }}HTTPS{{ else }}HTTP{{end}}
+              scheme: {{ if Values.olm.tlsSecret }}HTTPS{{ else }}HTTP{{end}}
```
timflannagan, line 90: Is this missing a `.Values` prefix?

awgreene: Addressed.

deploy/chart/templates/0000_50_olm_07-olm-operator.deployment.yaml
```diff
@@ -87,12 +87,12 @@ spec:
             httpGet:
               path: /healthz
               port: {{ .Values.olm.service.internalPort }}
-              scheme: {{ if or .Values.olm.tlsSecret .Values.olm.clientCASecret }}HTTPS{{ else }}HTTP{{end}}
+              scheme: {{ if .Values.olm.tlsSecret }}HTTPS{{ else }}HTTP{{end}}
```
timflannagan, line 90: It wasn't immediately clear to me how helm handles undefined variables when evaluating conditionals. It looks like the `tlsSecret` variable is commented out in https://github.com/operator-framework/operator-lifecycle-manager/blob/master/deploy/chart/values.yaml#L18? Does this just always evaluate to HTTP in that case?

timflannagan: Ran this locally and that seems to be the behavior so we're likely in the clear.

One thought I had is that given the httpGet scheme defaults to HTTP maybe we could omit the need for a conditional statement entirely and only specify the scheme field when `tlsSecret: true` and override that field to HTTPS:

```template
...
httpGet:
{- if .Values.olm.tlsSecret }}
    scheme: HTTPS
{{- end }}
...
```

WYDT?

awgreene: Great question! Yes - if `.Values.olm.clientCASecret` does not exist it will default to `HTTP`.

If `.Values.olm` in the `values.yaml` file did not exist we'd get a nil ptr reference and the script would break.

PR #2277, opened 2021-07-21
Hash and prune copied CSVs in olm-operator.

benluddy:
OLM doesn't need to maintain a local cache of complete copied
ClusterServiceVersion objects because it also holds the source of
truth.

### DIFF
--- a/pkg/controller/operators/olm/operatorgroup.go
+++ b/pkg/controller/operators/olm/operatorgroup.go
@@ -3,6 +3,7 @@ package olm
 import (
 	"context"
 	"fmt"
+	"hash/fnv"
 	"reflect"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -641,13 +644,36 @@ func (a *Operator) ensureCSVsInNamespaces(csv *v1alpha1.ClusterServiceVersion, o
 	logger := a.logger.WithField("opgroup", operatorGroup.GetName()).WithField("csv", csv.GetName())
 
 	targetCSVs := make(map[string]*v1alpha1.ClusterServiceVersion)
+
+	copyPrototype := v1alpha1.ClusterServiceVersion{
```
ecordell, line 648: I don't have a strong preference here, but I think you could calculate the "prototype" in copyableCSVHash. that would make me more confident that we're hashing the same thing here as in the listwatcher

benluddy:
/hold

Will need a rebase once https://github.com/operator-framework/operator-lifecycle-manager/pull/2282 merges -- that dependency bump fixes a bug in client-go/testing that affects this change.

dinhxuanvu (approved):
/lgtm

PR #2281, opened 2021-07-22
Catalog switcher

jchunkins:

Implementation of [catalog switching enhancement proposal](https://github.com/operator-framework/enhancements/blob/master/enhancements/automatic-catalog-switching-annotations.md)

Quick check for e2e test:

```
make e2e-local TEST='adding catalog template adjusts image used'
```

See [summary](https://github.com/operator-framework/enhancements/blob/master/enhancements/automatic-catalog-switching-annotations.md#summary) and [motivation](https://github.com/operator-framework/enhancements/blob/master/enhancements/automatic-catalog-switching-annotations.md#motivation) of enhancement proposal.

**Reviewer Checklist**


go.mod
```diff
@@ -60,6 +60,7 @@ replace (
 	// controller runtime
 	github.com/openshift/api => github.com/openshift/api v0.0.0-20200331152225-585af27e34fd // release-4.5
 	github.com/openshift/client-go => github.com/openshift/client-go v0.0.0-20200326155132-2a6cd50aedd0 // release-4.5
+	github.com/operator-framework/api => /Users/jhunkins/go/src/github.com/operator-framework/api
```
timflannagan, line 63: Note to other reviewers - we're waiting on https://github.com/operator-framework/api/pull/139 to merge before this pin can be removed.

jchunkins: api reference has been updated

pkg/controller/operators/catalogtempate/operator.go
```diff
@@ -0,0 +1,311 @@
...
+		catalogsource.RemoveStatusConditions(logger, o.client, outputCatalogSource, StatusTypeTemplatesHaveResolved, StatusTypeResolvedImage)
+		// no further action is needed
+		return nil
+	}
+
+	processedCatalogImageTemplate, unresolvedTemplates := catalogsource.ReplaceTemplates(catalogImageTemplate)
+
+	templatesAreResolved := len(unresolvedTemplates) == 0
+	// curly braces are not allowed in a real image reference (see https://github.com/distribution/distribution/blob/2461543d988979529609e8cb6fca9ca190dc48da/reference/reference.go#L4-L24)
+	invalidCharsFound := strings.ContainsAny(processedCatalogImageTemplate, "{}")
```
kevinrizza, line 182: can we use an image distribution lib to validate the string? I assume there are other potentially invalid values here, or at the very least we want to separate the templating bits out and potentially validate afterwards?

jchunkins: I am familiar with that code and have used it elsewhere. Should be an easy thing to add... frankly I should have thought of this.

jchunkins: added the distribution lib and used its parser

pkg/lib/catalogsource/image_template.go
```diff
@@ -0,0 +1,356 @@
...
+
+// ReplaceTemplates takes a catalog image reference containing templates (i.e. catalogImageTemplate)
+// and attempts to replace the templates with actual values (if available).
+// The return value processedCatalogImageTemplate represents the catalog image reference after processing.
+// Callers of this function should check the unresolvedTemplates return value to determine
+// if all values were properly resolved (i.e. empty slice means all items were resolved, and presence
+// of a value in the slice means that the template was either not found in the cache or its value has not been
+// fetched yet). Providing an empty catalogImageTemplate results in empty processedCatalogImageTemplate and
+// zero length unresolvedTemplates
+func ReplaceTemplates(catalogImageTemplate string) (processedCatalogImageTemplate string, unresolvedTemplates []string) {
```
ecordell, line 147: without the dynamic gvk templating, this could probably be simplified to a call to the stdlib [template package](https://pkg.go.dev/text/template)

ecordell: also: since any unresolved components should mean the image reference is invalid, I'm not sure I get the value in returning a partially-resolved reference and a list of the unresolved ones: why not just return an error with the missing values?

scripts/copy_crds.sh
```diff
@@ -10,5 +10,5 @@ CRD_PATH="${SCRIPT_ROOT}/vendor/github.com/operator-framework/api/crds"
 rm "${SCRIPT_ROOT}"/deploy/chart/crds/*.yaml
 for f in "${CRD_PATH}"/*.yaml ; do
     echo "copying ${f}"
-    cp "${f}" "${SCRIPT_ROOT}/deploy/chart/crds/0000_50_olm_00-$(basename "$f" | sed 's/^.*_\([^.]\+\)\.yaml/\1.crd.yaml/')"
+    cp "${f}" "${SCRIPT_ROOT}/deploy/chart/crds/0000_50_olm_00-$(basename "$f" | sed 's/^.*_\([^.]\{1,\}\)\.yaml/\1.crd.yaml/')"
```
ecordell, line 13: why is this needed?

test/e2e/catalog_e2e_test.go
```diff
@@ -22,10 +23,13 @@ import (
 	"github.com/operator-framework/api/pkg/lib/version"
 	"github.com/operator-framework/api/pkg/operators/v1alpha1"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/clientset/versioned"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/operators/catalogtempate"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/registry"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/catalogsource"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/operatorclient"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/ownerutil"
 	"github.com/operator-framework/operator-lifecycle-manager/test/e2e/ctx"
+	"k8s.io/apimachinery/pkg/api/meta"
```
ecordell, line 32: nit: this one should go in the second import block

scripts/copy_crds.sh
jchunkins: This is needed to make the code portable so it can be run on linux/Mac versions of sed. Here's what I wrote in the commit comment:

```
The command + in sed is non-portable. The posix compliant version of
this command is {1,}

This change allows this script to run on a mac too (which uses BSD sed)

See https://riptutorial.com/sed/topic/9436/bsd-macos-sed-vs--gnu-sed-vs--the-posix-sed-specification
```

pkg/lib/catalogsource/image_template.go
jchunkins: The unresolvedTemplates are used to inform the user in a message about exactly what template that was not resolvable. It is used in messages exposed in both the log and the status conditions.

pkg/controller/operators/catalogtempate/operator.go
```diff
@@ -0,0 +1,319 @@
...
+	// code needs to be re-enabled once we've come up with a valid approach
+	_ = foundGVKs
+	// for _, gvk := range foundGVKs {
+	// 	o.processGVK(o.ctx, logger, gvk.GroupVersionKind)
+	// }
+	///////
+
+	catalogImageTemplate := catalogsource.GetCatalogTemplateAnnotation(outputCatalogSource)
+	if catalogImageTemplate == "" {
+		logger.Info("this catalog source is not participating in template replacement")
```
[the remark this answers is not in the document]
jchunkins: I assume you're looking for debug?

pkg/controller/operators/catalogtempate/operator.go
```diff
@@ -0,0 +1,319 @@
...
+	"sync"
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/operator-framework/api/pkg/operators/v1alpha1"
+
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/clientset/versioned"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/informers/externalversions"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/catalogsource"
```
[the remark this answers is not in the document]
jchunkins: Thanks... I learned something today. I always let the auto format handle imports thinking it would take care of the ordering, but I see now that I can use returns to keep blocks organized.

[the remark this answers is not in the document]
jchunkins: Also I assume that "everything else" includes the `api` project since its not part of the `operator-lifecycle-manager`

pkg/lib/catalogsource/image_template.go
jchunkins: Also the processedCatalogImageTemplate is used to inform the user of what the template looks like in the status conditions. It helps the user see what's wrong by looking at the message AND what the templating code computed.

pkg/controller/operators/catalogtempate/operator.go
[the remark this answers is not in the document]
jchunkins: fixed assuming I understand the organization, this has been fixed

test/e2e/catalog_e2e_test.go
jchunkins: fixed assuming I understand the organization, this has been fixed

pkg/controller/operators/catalogtempate/operator.go
```diff
@@ -0,0 +1,319 @@
...
+	logger.Info("syncing catalog source for annotation templates")
+
+	// this is our opportunity to discover GVK templates and setup watchers (if possible)
+	foundGVKs := catalogsource.InitializeCatalogSourceTemplates(outputCatalogSource)
+
+	///////
+	// FIXME: There have been concerns about security of allowing selection of arbitrary
+	// GVK during code reviews. For now we're disabling setting up dynamic watchers. This
+	// code needs to be re-enabled once we've come up with a valid approach
+	_ = foundGVKs
```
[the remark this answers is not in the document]
jchunkins: The section of code that starts the dynamic informers should be disabled. That’s the part of the code that is commented out (see where I pointed to that below)
```
	///////
	// FIXME: There have been concerns about security of allowing selection of arbitrary
	// GVK during code reviews. For now we're disabling setting up dynamic watchers. This
	// code needs to be re-enabled once we've come up with a valid approach
	_ = foundGVKs
	// for _, gvk := range foundGVKs {
	// 	o.processGVK(o.ctx, logger, gvk.GroupVersionKind) <----- this creates the informers
	// }
	///////
```

pkg/controller/operators/catalogtempate/operator.go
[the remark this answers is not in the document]
jchunkins: Changed to debug

pkg/lib/catalogsource/image_template.go
jchunkins: I'd like to address the security shortcomings of the GVK feature in the future, so I'd like to keep things as close to "working" as possible.

pkg/controller/operators/catalogtempate/operator.go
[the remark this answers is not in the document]
jchunkins: I think you're asking to remove the import statements? I'd like to address the security shortcomings of the GVK feature in the future, so I'd like to keep things as close to "working" as possible.

jchunkins:
It looks like `unit / unit (pull_request)` GitHub action failed, but when I run the error test case `TestSyncOperatorGroups` on my machine it does not have an error. Not sure how to restart the GitHub action to prove that its a flake of some kind.

scripts/copy_crds.sh
jchunkins: Sorry.. git comment is slightly misleading since its missing some of the slashes:

The command `\+` in sed is non-portable. The posix compliant version of
this command is `\{1,\}`

This change allows this script to run on a mac too (which uses BSD sed)

See https://riptutorial.com/sed/topic/9436/bsd-macos-sed-vs--gnu-sed-vs--the-posix-sed-specification

pkg/controller/operators/catalogtempate/operator.go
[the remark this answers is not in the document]
ecordell: sorry that was unclear - debug is fine, but I was just saying I don't think we need a log for this at all

pkg/controller/operators/catalogtempate/operator.go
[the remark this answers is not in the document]
jchunkins: removed all of the dynamic machinery code entirely

exdx:
> It looks like `unit / unit (pull_request)` GitHub action failed, but when I run the error test case `TestSyncOperatorGroups` on my machine it does not have an error. Not sure how to restart the GitHub action to prove that its a flake of some kind.

It's a well known flake, I restarted it

pkg/lib/catalogsource/image_template.go
```diff
@@ -0,0 +1,356 @@
...
+	"k8s.io/apimachinery/pkg/version"
+	"k8s.io/client-go/util/jsonpath"
+)
+
+const (
+
+	// regex capture group names
+
+	// cap_grp_kubemajorv is a capture group name for a kube major version
+	cap_grp_kubemajorv = "kubemajorv"
```
njhale, line 27: nit: `snake_case` is not conventional in go

pkg/lib/catalogsource/image_template.go
```diff
@@ -0,0 +1,356 @@
...
+			// only add named groups to the map
+			if names[i] != "" {
+				results[names[i]] = match
+			}
+		}
+		return results
+	}
+
+	catalogImageTemplate := GetCatalogTemplateAnnotation(catalogSource)
+	if catalogImageTemplate != "" {
```
njhale, line 221: you could return early `if catalogImageTemplate == "  {...` instead of nesting this large block

test/e2e/catalog_e2e_test.go
```diff
@@ -7,22 +7,26 @@ import (
...
 	"net"
 	"os"
+	"strconv"
 	"strings"
 	"time"
 
 	"github.com/blang/semver/v4"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	"github.com/operator-framework/api/pkg/lib/version"
```
njhale, line 17: nit: import order (looks like it was corrected in a few other places)

test/e2e/catalog_e2e_test.go
```diff
@@ -1028,6 +1032,165 @@ var _ = Describe("Catalog represents a store of bundles which OLM can use to ins
...
+				}
+				// condition should flip to true indicating that the catalog image has been updated
+				if meta.IsStatusConditionTrue(source.Status.Conditions, catalogtempate.StatusTypeTemplatesHaveResolved) {
+					return true, nil
+				}
+
+				return false, nil
+			}, 5*time.Minute, 1*time.Second).Should(BeTrue())
+			///////
+		*/
```
njhale, line 1166: Can we remove this in a cleanup commit? That way, we can retain the code in git without impacting readability/understandability.

test/e2e/catalog_e2e_test.go
```diff
@@ -1028,6 +1032,165 @@ var _ = Describe("Catalog represents a store of bundles which OLM can use to ins
...
+		Expect(err).ShouldNot(HaveOccurred(), "error updating catalog source with template annotation")
+
+		// wait for status condition to show up
+		Eventually(func() (bool, error) {
+			source, err = crc.OperatorsV1alpha1().CatalogSources(source.GetNamespace()).Get(context.TODO(), sourceName, metav1.GetOptions{})
+			if err != nil {
+				return false, err
+			}
+			// if the conditions array has the entry we know things got updated
+			condition := meta.FindStatusCondition(source.Status.Conditions, catalogtempate.StatusTypeTemplatesHaveResolved)
```
njhale, line 1115: `meta.FindStatusCondition(...)` <- nice find!

pkg/lib/catalogsource/image_template.go
```diff
@@ -0,0 +1,356 @@
...
+	}
+}
+
+func UpdateKubeVersion(serverVersion *version.Info, logger *logrus.Logger) {
+	if serverVersion == nil {
+		logger.Warn("no server version provided")
+		return
+	}
+
+	templateNameToReplacementValuesMap[TEMPL_KUBEMAJORV] = serverVersion.Major
```
njhale, line 346: Not threadsafe -- I suggest using a mutex here.

pkg/lib/catalogsource/image_template.go
```diff
@@ -0,0 +1,356 @@
...
+	}
+	result := strings.Join(regexEntryAsString, "|")
+	return result
+}
+
+var regexList = regexEntries{
+	{cap_grp_kubemajorv, TEMPL_KUBEMAJORV},
+	{cap_grp_kubeminorv, TEMPL_KUBEMINORV},
+	{cap_grp_kubepatchv, TEMPL_KUBEPATCHV},
+	{cap_grp_gvk, templ_gvk},
```
njhale, line 132: do we need to define or compile these if they're not going to be used in this PR? I'd like to cut code for features that are in question.

pkg/controller/operators/catalogtempate/operator.go
```diff
@@ -0,0 +1,241 @@
...
+	queueinformer.Operator
+	logger                        *logrus.Logger                  // common logger
+	namespace                     string                          // operator namespace
+	client                        versioned.Interface             // client used for OLM CRs
+	dynamicClient                 dynamic.Interface               // client used to dynamically discover resources
+	discoveryClient               *discovery.DiscoveryClient      // queries the API server to discover resources
+	lister                        operatorlister.OperatorLister   // union of versioned informer listers
+	catalogSourceTemplateQueueSet *queueinformer.ResourceQueueSet // work queues for a catalog source update
+	resyncPeriod                  func() time.Duration            // period of time between resync
+	ctx                           context.Context                 // context used for shutting down
```
njhale, line 44: nit: From the [context package](https://pkg.go.dev/context#pkg-overview):

> Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it. The Context should be the first parameter, typically named ctx

(although I know there are a few violations in OLM already)

pkg/controller/operators/catalogtempate/operator.go
```diff
@@ -0,0 +1,241 @@
...
+}
+
+func (o *Operator) syncCatalogSources(obj interface{}) error {
+	// this is an opportunity to update the server version (regardless of any other actions for processing a catalog source)
+	o.updateServerVersion()
+
+	inputCatalogSource, ok := obj.(*v1alpha1.CatalogSource)
+	if !ok {
+		o.logger.Debugf("wrong type: %#v", obj)
+		return fmt.Errorf("casting CatalogSource failed")
```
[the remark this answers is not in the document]
jchunkins: If this is the case, then `pkg/controller/operators/catalog/operator.go` has the same problem since this was copied from that code.

cmd/catalog/main.go
```diff
@@ -16,6 +16,7 @@ import (
 
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/client"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/operators/catalog"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/operators/catalogtempate"
```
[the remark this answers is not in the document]
jchunkins: :grimacing: oops

pkg/lib/catalogsource/catalogsource_update.go
```diff
@@ -0,0 +1,173 @@
...
+
+• logger: used to log errors only
+
+• client: used to fetch / update catalog source status
+
+• catsrc: the CatalogSource to use as a source for status updates. Callers are
+responsible for updating the catalog source status values as necessary.
+*/
+func UpdateStatus(logger *logrus.Entry, client versioned.Interface, catsrc *v1alpha1.CatalogSource) error {
+	mu.Lock()
```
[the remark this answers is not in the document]
jchunkins: TLDR; Using the locks protects the CatalogSource "shared resource" from two threads updating status at same time.

We've got two separate controllers attempting to update the same CatalogSource objects. If it was the same controller you could guarantee how it was being updated. Right now the status updates don't interfere with each other... the status conditions are used by the new controller, and the old controller updates its existing fields. However in the future if the existing controller updates the status conditions array, updates to this array need to be coordinated to ensure that the content and order of the array are maintained properly. Ultimately this is trying to minimize the chances of the two controllers interfering with each other.

test/e2e/catalog_e2e_test.go
jchunkins: thanks!

pkg/controller/operators/catalogtempate/operator.go
jchunkins: I realize storing the context is not ideal. This was necessary when I had the GVK code in place... the context was required in order to setup the watches, which would happen any time `syncCatalogSources` would fire. Now that the GVK code has been removed its not necessary, but will be necessary when adding the GVK code back someday.

pkg/lib/catalogsource/catalogsource_update.go
```diff
@@ -0,0 +1,173 @@
...
+
+	"github.com/operator-framework/api/pkg/operators/v1alpha1"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/clientset/versioned"
+	"github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// mu is a package scoped mutex for synchronizing catalog source updates
+var mu sync.Mutex
```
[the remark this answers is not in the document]
jchunkins: See related comment https://github.com/operator-framework/operator-lifecycle-manager/pull/2281#discussion_r677511023

pkg/lib/catalogsource/image_template.go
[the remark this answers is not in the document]
jchunkins: totally fair. I know this is supposed to be the Golang convention and somehow did the wrong thing here.

pkg/lib/catalogsource/catalogsource_update.go
```diff
@@ -0,0 +1,173 @@
...
+
+• logger: used to log errors only
+
+• client: used to fetch / update catalog source
+
+• catsrc: the CatalogSource to use as a source for image and status condition updates.
+
+• conditions: condition values to be updated
+*/
+func UpdateImageReferenceAndStatusCondition(logger *logrus.Entry, client versioned.Interface, catsrc *v1alpha1.CatalogSource, conditions ...metav1.Condition) error {
```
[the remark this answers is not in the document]
jchunkins: Its mostly about shared resource updating and locking... see https://github.com/operator-framework/operator-lifecycle-manager/pull/2281#discussion_r677511023

pkg/lib/catalogsource/image_template.go
jchunkins: agreed... see https://github.com/operator-framework/operator-lifecycle-manager/pull/2281#discussion_r677535154

test/e2e/catalog_e2e_test.go
jchunkins: yes it was will fix

cmd/catalog/main.go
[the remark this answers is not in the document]
jchunkins: fixed

pkg/controller/operators/catalogtempate/operator.go
```diff
@@ -0,0 +1,241 @@
+package catalogtempate
```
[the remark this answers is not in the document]
jchunkins: fixed

pkg/controller/operators/catalogtempate/operator.go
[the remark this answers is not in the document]
jchunkins: fixed

pkg/lib/catalogsource/image_template.go
jchunkins: fixed

test/e2e/catalog_e2e_test.go
jchunkins: fixed

pkg/lib/catalogsource/image_template.go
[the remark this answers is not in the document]
jchunkins: fixed

pkg/controller/operators/catalogtempate/operator.go
```diff
@@ -0,0 +1,241 @@
...
+const (
+	StatusTypeTemplatesHaveResolved = "TemplatesHaveResolved"
+	StatusTypeResolvedImage         = "ResolvedImage"
+
+	ReasonUnableToResolve      = "UnableToResolve"
+	ReasonAllTemplatesResolved = "AllTemplatesResolved"
+)
+
+type Operator struct {
+	queueinformer.Operator
```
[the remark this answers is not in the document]
jchunkins: This was brought to my attention in the dev call, but I got the impression that using this mechanism was OK too. The catalog operator is still using the old stuff and not sure if mixing old/new in same operator makes things any clearer.

pkg/controller/operators/catalogtempate/operator.go
jchunkins: fixed

pkg/lib/catalogsource/image_template.go
jchunkins: fixed

pkg/lib/catalogsource/image_template.go
jchunkins: fixed

test/e2e/catalog_e2e_test.go
jchunkins: fixed

pkg/lib/catalogsource/image_template.go
jchunkins: fixed

pkg/controller/operators/catalogtempate/operator.go
```diff
@@ -0,0 +1,241 @@
...
+			},
+			metav1.Condition{
+				Type:    StatusTypeResolvedImage,
+				Status:  metav1.ConditionFalse,
+				Reason:  ReasonUnableToResolve,
+				Message: processedCatalogImageTemplate,
+			},
+		)
+		logger.Infof(message)
+	}
```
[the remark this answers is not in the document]
jchunkins: fixed

pkg/controller/operators/catalogtemplate/operator.go
```diff
@@ -0,0 +1,235 @@
...
+		"catSrcName": outputCatalogSource.GetName(),
+		"id":         queueinformer.NewLoopID(),
+	})
+	logger.Info("syncing catalog source for annotation templates")
+
+	catalogImageTemplate := catalogsource.GetCatalogTemplateAnnotation(outputCatalogSource)
+	if catalogImageTemplate == "" {
+		logger.Debug("this catalog source is not participating in template replacement")
+		// make sure the conditions are removed
+		catalogsource.RemoveStatusConditions(logger, o.client, outputCatalogSource, StatusTypeTemplatesHaveResolved, StatusTypeResolvedImage)
```
kevinrizza, line 145: Shouldn't this handle and return the error from this function?

jchunkins: I suppose we can requeue these... same issue exists for other status functions as well

kevinrizza: We don't necessarily have to handle retry, but that function is currently swallowing the error

jchunkins: Just to be clear, I am making the change to return the error, which as I understand it, will requeue the catalog source.

kevinrizza: Sure, because otherwise we are failing to update. The controller itself should retry with a default backoff on failures. I'm just saying we don't necessarily need to write explicit retry logic wrapping the client call (so that the reconcile function doesn't error out) -- I don't think we want to have code that just swallows kube client errors. otherwise why are we trying to do an update?

jchunkins: makes sense. I've fix the issue and pushed.... also made some minor adjustments to logging.

pkg/lib/catalogsource/catalogsource_update.go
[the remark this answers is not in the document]
ecordell: > Ultimately this is trying to minimize the chances of the two controllers interfering with each other.

It does that, but it also bottlenecks the workers that are processing queues of catalogsources: they're working in parallel but because this mutex is in the package scope, updates to catalogsource status have to happen serially.

The way this is generally dealt with in the other controllers is:
 - use `patch` instead of `update` whenever possible (in this case, patch should always succeed since no other controller owns that condition)
  - if the patch/update fails for any reason, the key for the catalog is put back on the queue and the change is recalculated / retried
  
 This model has the additional benefit of decoupling the controllers so that they can run in different pods if needed

pkg/lib/catalogsource/catalogsource_update.go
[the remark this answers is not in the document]
ecordell: responded in the other thread

pkg/lib/catalogsource/catalogsource_update.go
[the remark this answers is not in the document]
jchunkins: Bottom line... the way things are written they do return errors if `UpdateStatus(...)` or `Update(...)` fail, and my last change ensure that the errors are propagated to the "sync" return value, so the retry should kick in as you've described. 

So it seems like I just need to remove the mutex since you're saying its not really needed.

pkg/lib/catalogsource/catalogsource_update.go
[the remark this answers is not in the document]
jchunkins: Ok... so the `Update*` functions I added are almost all dealing with status updates only (i.e. `client.OperatorsV1alpha1().CatalogSources(...).UpdateStatus(...)`). The only one that updates anything in the `spec` is this new controller (to update the image reference) so it uses `client.OperatorsV1alpha1().CatalogSources(...).Update(...)`. 

I think you're saying that patch (i.e. `client.OperatorsV1alpha1().CatalogSources(...).Patch(...)`) is preferable over `client.OperatorsV1alpha1().CatalogSources(...).Update(...)`. 

It does not seem like the patch function should be used for status updates,  might make sense for the `UpdateImageReferenceAndStatusCondition` function since it updates the image reference. 

Honestly it sounds like you're saying I don't need to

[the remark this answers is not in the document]
jchunkins: Resolving this conversation... can continue in other thread

pkg/lib/catalogsource/catalogsource_update.go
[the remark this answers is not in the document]
jchunkins: By the way, I added the mutex since someone was concerned about two controllers updating status.

njhale (commented):
looking really good. thank you for addressing all the feedback so far! I think we're in the home stretch.

pkg/lib/catalogsource/catalogsource_update.go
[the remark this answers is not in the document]
jchunkins: I did put in place a comment on the `UpdateStatus` function that says the following (only showing relevant portions here):

> ... Note that the status values are updated to the values from catsrc in their entirety when using this function.
...
• catsrc: the CatalogSource to use as a source for status updates. Callers are
responsible for updating the catalog source status values as necessary

The purpose of the documentation was to make it super clear that callers are responsible for making valid status updates when using this function. To that end I had analyzed all of the places where catalog source status is updated and found that the only offender was a section of code that blew up the entire status section:

```go
// The sourceType is valid, clear (all) status if there's existing invalid spec reason
if out.Status.Reason == v1alpha1.CatalogSourceSpecInvalidError {
    out.Status = v1alpha1.CatalogSourceStatus{}
}
```

I changed this code to this to preserve the conditions prior to resetting the other status:

```go
// The sourceType is valid, clear all status (other than status conditions array) if there's existing invalid spec reason
if out.Status.Reason == v1alpha1.CatalogSourceSpecInvalidError {
    out.Status = v1alpha1.CatalogSourceStatus{
        Conditions: out.Status.Conditions,
    }
}
```
This is the link to the diff shown above: https://github.com/operator-framework/operator-lifecycle-manager/pull/2281/files#diff-a1760d9b7ac1e93734eea675d8d8938c96a50e995434b163c6f77c91bace9990R606-R608

This is the other reason I created separate functions for updating the status... to make status updates as restricted in scope as possible.

pkg/lib/catalogsource/catalogsource_update.go
```diff
@@ -0,0 +1,173 @@
...
+• catsrc: the CatalogSource to use as a source for status updates.
+
+• conditions: condition values to be updated
+*/
+func UpdateStatusCondition(logger *logrus.Entry, client versioned.Interface, catsrc *v1alpha1.CatalogSource, conditions ...metav1.Condition) error {
+	mu.Lock()
+	defer mu.Unlock()
+
+	// get the absolute latest update of this catalog source in case it changed
+	latest, err := client.OperatorsV1alpha1().CatalogSources(catsrc.GetNamespace()).Get(context.TODO(), catsrc.GetName(), metav1.GetOptions{})
```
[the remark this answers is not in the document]
jchunkins: I found that without the `Get` followed `UpdateStatus` there were a lot more instances of update failures in the logs, so I assumed that's why the original code (prior to my PR) did this two step approach. If you think it is better to "fail" this way, I can make the change. I assume this applies to all of the `UpdateXXX` functions I've created in this PR.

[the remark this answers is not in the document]
njhale: > If you think it is better to "fail" this way, I can make the change

In the abstract, I think the controller needs to be given a chance to review any changes that occurred to the cluster since it formulated its changes. In this specific case, I think the current code will work because there are no other controllers that own the `status.conditions` field. In the long run, if we add independent controllers that touch independent conditions -- e.g. a controller deployed to a separate process -- then we can start to run into problems with data consistency; one controller sets the condition it cares about, the other inadvertently removes it.

You can probably avoid the

pkg/lib/catalogsource/catalogsource_update.go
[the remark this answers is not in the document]
njhale: I think https://github.com/operator-framework/operator-lifecycle-manager/pull/2281/files#r679474664 responds, in part, to this as well.

further, it seems like this helper is still stomping the entire status, and not just a specific condition.

pkg/lib/catalogsource/catalogsource_update.go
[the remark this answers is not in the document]
jchunkins: OK... I'll remove the mutex as well as the necessary CatalogSource `Get` calls in all of the `UpdateXXX` and `RemoveStatusConditions` functions

pkg/controller/operators/catalogtempate/operator.go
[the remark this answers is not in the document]
njhale: AFAIK, we're planning on phasing out `QueueInformer` as we introduce OLM's next gen APIs.

We'll want to follow up later and switch this over to controller-runtime if we want to support this long-term, but I don't think it's something to block the PR on at this point.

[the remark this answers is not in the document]
jchunkins: OK. Nothing to do at this time. Will close this conversation then. Thanks.

cmd/catalog/main.go
```diff
@@ -140,9 +141,17 @@ func main() {
 		log.Panicf("error configuring operator: %s", err.Error())
 	}
 
+	opCatalogTemplate, err := catalogtemplate.NewOperator(ctx, *kubeConfigPath, logger, *wakeupInterval, *catalogNamespace)
+	if err != nil {
+		log.Panicf("error configuring operator: %s", err.Error())
```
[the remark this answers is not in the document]
jchunkins: fixed in next commit

pkg/controller/operators/catalogtemplate/operator.go
```diff
@@ -0,0 +1,245 @@
...
+	catalogImageTemplate := catalogsource.GetCatalogTemplateAnnotation(outputCatalogSource)
+	if catalogImageTemplate == "" {
+		logger.Debug("this catalog source is not participating in template replacement")
+		// make sure the conditions are removed
+
+		if err := catalogsource.RemoveStatusConditions(logger, o.client, outputCatalogSource, StatusTypeTemplatesHaveResolved, StatusTypeResolvedImage); err != nil {
+			return err
+		}
+		// no further action is needed
+		return nil
```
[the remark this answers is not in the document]
jchunkins: fixed in next commit

pkg/lib/catalogsource/catalogsource_update.go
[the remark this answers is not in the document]
jchunkins: fixed

pkg/lib/catalogsource/catalogsource_update.go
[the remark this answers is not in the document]
jchunkins: fixed

njhale (approved):
/hold cancel
/lgtm

@jchunkins thanks for sticking with this through all the feedback!

I'd be interested in any thoughts you have about your experience contributing; e.g. good, bad, suggestions. Maybe we can continue in [slack](https://kubernetes.slack.com/archives/C0181L6JYQ2).

PR #2285, opened 2021-07-23
Prune unused ClusterServiceVersion fields from catalog's informer.

benluddy:
The catalog operator watches both copied and original
ClusterServiceVersions in order to perform dependency resolution and
to check for API ownership conflicts during installation, but it only
reads a small subset of fields. Pruning those fields as CSVs are read
from the wire can substantially reduce heap size on clusters with a
significant total number of CSVs across all namespaces (for example,
when operators target all namespaces on a cluster with thousands of
namespaces).


dinhxuanvu (approved):
/lgtm

dinhxuanvu (approved):
Additional change to keep the `version` field.
/lgtm

PR #2290, opened 2021-07-26
Expose errors generated while retrieving catalog content.

benluddy:
Failing to fetch catalog content should not silently return an empty
cache. Instead, it should fail outright with an error that indicates
which catalog(s) could not be fetched.

### DIFF
--- a/pkg/controller/registry/resolver/cache.go
+++ b/pkg/controller/registry/resolver/cache.go
@@ -8,6 +8,8 @@ import (
 	"sync"
 	"time"
 
+	"k8s.io/apimachinery/pkg/util/errors"
+
 	"github.com/sirupsen/logrus"
 
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/listers/operators/v1alpha1"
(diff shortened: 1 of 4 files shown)


pkg/controller/registry/resolver/cache.go
```diff
@@ -8,6 +8,8 @@ import (
 	"sync"
 	"time"
 
+	"k8s.io/apimachinery/pkg/util/errors"
+
```
ecordell, line 12: supernit: space

pkg/controller/registry/resolver/cache.go
```diff
@@ -183,6 +198,12 @@ func (c *OperatorCache) Namespaced(namespaces ...string) MultiCatalogOperatorFin
 
 func (c *OperatorCache) populate(ctx context.Context, snapshot *CatalogSnapshot, registry client.Interface) {
 	defer snapshot.m.Unlock()
+	defer func() {
+		// Don't cache an errorred snapshot.
+		if snapshot.err != nil {
+			snapshot.expiry = time.Time{}
```
ecordell, line 204: does this trigger a cache-refresh hotloop?

anik120 (approved):
/lgtm

pkg/controller/registry/resolver/cache.go
benluddy: At worst I'd expect syncResolvingNamespace to be requeued a handful of times, but that's a flat number of retries, not a feedback cycle. Do you have a particular cycle in mind that I missed?

PR #2293, opened 2021-07-27
Retry CRD updates on conflicts

joelanford:
Always retry CRD updates when conflicts occur.

Conflicts are common in eventually consistent controllers. Reconciliation should be tolerant to conflicts and retry updates with a backoff in the face of conflicts without requiring a full re-reconciliation of the object.

Currently, when a CRD update fails while reconciling an install plan due to a conflict, the install plan is permanantly marked as failed, which forces the user to delete the install plan and try again.

This PR treats conflicts as transient errors and performs retries with a default backoff.

Note that if a CRD update consistently results in conflict errors, it will still eventually be marked as failed, but this should happen extremely rarely (if ever) in practice.

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/install/webhook.go
+++ b/pkg/controller/install/webhook.go
@@ -5,16 +5,17 @@ import (
 	"fmt"
 	"hash/fnv"
 
-	"github.com/operator-framework/api/pkg/operators/v1alpha1"
-	hashutil "github.com/operator-framework/operator-lifecycle-manager/pkg/lib/kubernetes/pkg/util/hash"
-	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/ownerutil"
-
 	log "github.com/sirupsen/logrus"
 	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/rand"
+	"k8s.io/client-go/util/retry"
+
+	"github.com/operator-framework/api/pkg/operators/v1alpha1"
+	hashutil "github.com/operator-framework/operator-lifecycle-manager/pkg/lib/kubernetes/pkg/util/hash"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/ownerutil"
 )
 
 func ValidWebhookRules(rules []admissionregistrationv1.RuleWithOperations) error {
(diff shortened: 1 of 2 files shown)


pkg/controller/install/webhook.go
```diff
@@ -200,69 +201,74 @@ func (i *StrategyDeploymentInstaller) createOrUpdateConversionWebhook(caPEM []by
...
-			},
-		}
+			}
 
-		// update CRD conversion Specs
-		if _, err = i.strategyClient.GetOpClient().ApiextensionsInterface().ApiextensionsV1().CustomResourceDefinitions().Update(context.TODO(), crd, metav1.UpdateOptions{}); err != nil {
-			return fmt.Errorf("Error updating CRD with Conversion info: %v", err)
+			// update CRD conversion Specs
+			if _, err = i.strategyClient.GetOpClient().ApiextensionsInterface().ApiextensionsV1().CustomResourceDefinitions().Update(context.TODO(), crd, metav1.UpdateOptions{}); err != nil {
+				return fmt.Errorf("Error updating CRD with Conversion info: %w", err)
```
timflannagan, line 267: Re-reading https://pkg.go.dev/k8s.io/client-go/util/retry#RetryOnConflict seems to imply that wrapping the error here might be problematic when identifying whether `IsConflict(err)`?

pkg/controller/operators/catalog/step.go
```diff
@@ -181,7 +187,7 @@ func (b *builder) NewCRDV1Beta1Step(client apiextensionsv1beta1client.Apiextensi
 				if k8serrors.IsNotFound(err) {
 					return v1alpha1.StepStatusNotPresent, nil
 				} else {
-					return v1alpha1.StepStatusNotPresent, errors.Wrapf(err, "error finding the %s CRD", crd.Name)
+					return v1alpha1.StepStatusNotPresent, fmt.Errorf("error finding the %q CRD: %w", crd.Name, err)
```
timflannagan, line 190: minor nit, not blocking: the else statement here can be dropped as both branches are returning.

pkg/controller/operators/catalog/step.go
```diff
@@ -132,27 +132,33 @@ func (b *builder) NewCRDV1Step(client apiextensionsv1client.ApiextensionsV1Inter
...
-				safe, err := crdlib.SafeStorageVersionUpgrade(currentCRD, crd)
-				if !safe {
-					b.logger.Errorf("risk of data loss updating %s: %s", step.Resource.Name, err)
-					return v1alpha1.StepStatusUnknown, errors.Wrapf(err, "risk of data loss updating %s", step.Resource.Name)
-				}
-				if err != nil {
-					return v1alpha1.StepStatusUnknown, errors.Wrapf(err, "checking CRD for potential data loss updating %s", step.Resource.Name)
-				}
+					// check to see if stored versions changed and whether the upgrade could cause potential data loss
+					safe, err := crdlib.SafeStorageVersionUpgrade(currentCRD, crd)
```
exdx, line 143: Does it make sense to retry only the portion of the codepath that actually gets/updates CRDs and not the business logic around validation that we provide as well? I know right now it's all tangled together. But from what I gather from this change, if there is an conflict error updating the CRD, we retry the get and validation again as well, for no real benefit

pkg/controller/install/webhook.go
joelanford: `IsConflict` ends up calling `ReasonForError()`, which uses `errors.As` [here](https://github.com/kubernetes/apimachinery/blob/a644435e2c133d990b624858d9c01985d7f59adf/pkg/api/errors/errors.go#L650) to recursively unwrap and find the expected error type.

pkg/controller/operators/catalog/step.go
joelanford: I thought about that, but since `currentCRD` can change between retries and its an input into the validation, I figured we need to re-run the validation on each retry in case something relevant to the validation also changed.

exdx: Right, I suppose that is possible

pkg/controller/operators/catalog/step.go
joelanford: If any non-nit changes come up, I can fix this while I'm at it. Otherwise, I'm inclined to leave this to avoid another CI cycle.

Relatedly, I'm pretty sure this kind of thing would be caught with go vet or golangci-lint. Perhaps in a follow-up, we should introduce one or both of those into our Makefile/CI process and get everything like this fixed at once.

pkg/controller/install/webhook.go
njhale: ah

> If the format specifier includes a %w verb with an error operand, the returned error will implement an Unwrap method returning the operand. It is invalid to include more than one %w verb or to supply it with an operand that does not implement the error interface. 

interesting

njhale (approved):
/approve

PR #2296, opened 2021-07-28
fix(sub): Reset ResolutionFailed cond when error is resolved

anik120:

In #2269, a new condition was introduced for Subscription to indicate any
dependency resolution error in it's message. However, when the case of
the error was resolved, the condition status (true) and the message
was sticking around. This PR fixes the issue, and makes sure that the
condition status is set to false, and the message and reason of the
condition are cleared off.

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -1246,14 +1246,8 @@ func (o *Operator) setSubsCond(subs []*v1alpha1.Subscription, condType v1alpha1.
 		cond.Reason = reason
 		cond.Message = message
 		if setTrue {
-			if cond.Status == corev1.ConditionTrue {
-				continue
-			}
 			cond.Status = corev1.ConditionTrue
 		} else {
-			if cond.Status == corev1.ConditionFalse {
-				continue
-			}
 			cond.Status = corev1.ConditionFalse
 		}
 		sub.Status.SetCondition(cond)
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -961,34 +961,28 @@ func (o *Operator) syncResolvingNamespace(obj interface{}) error {
 		// not-satisfiable error
 		if _, ok := err.(solver.NotSatisfiable); ok {
 			logger.WithError(err).Debug("resolution failed")
-			_, updateErr := o.setSubsCond(subs, v1alpha1.SubscriptionResolutionFailed, "ConstraintsNotSatisfiable", err.Error(), true)
+			subs = o.setSubsCond(subs, v1alpha1.SubscriptionResolutionFailed, "ConstraintsNotSatisfiable", err.Error(), true)
+			_, updateErr := o.updateSubscriptionStatuses(subs)
 			if updateErr != nil {
 				logger.WithError(updateErr).Debug("failed to update subs conditions")
+				return updateErr
 			}
 			return nil
 		}
-		_, updateErr := o.setSubsCond(subs, v1alpha1.SubscriptionResolutionFailed, "ErrorPreventedResolution", err.Error(), true)
+		subs = o.setSubsCond(subs, v1alpha1.SubscriptionResolutionFailed, "ErrorPreventedResolution", err.Error(), true)
+		_, updateErr := o.updateSubscriptionStatuses(subs)
 		if updateErr != nil {
 			logger.WithError(updateErr).Debug("failed to update subs conditions")
+			return updateErr
 		}
 		return err
 	} else {
-		subs, updateErr := o.setSubsCond(updatedSubs, v1alpha1.SubscriptionResolutionFailed, "", "", false)
+		subs = o.setSubsCond(subs, v1alpha1.SubscriptionResolutionFailed, "", "", false)
+		_, updateErr := o.updateSubscriptionStatuses(subs)
 		if updateErr != nil {
 			logger.WithError(updateErr).Debug("failed to update subs conditions")
 		}
-		updatedSubs = subs
-
-		// Also clear the condition from the subscriptions there were not included in the list of updatedSubs
-		allSubs, err := o.listSubscriptions(namespace)
-		if err != nil {
-			logger.WithError(err).Debug("couldn't list subscriptions")
-			return err
-		}
-		_, updateErr = o.setSubsCond(allSubs, v1alpha1.SubscriptionResolutionFailed, "", "", false)
-		if updateErr != nil {
-			logger.WithError(updateErr).Debug("failed to update subs conditions")
-		}
+		updatedSubs = o.setSubsCond(updatedSubs, v1alpha1.SubscriptionResolutionFailed, "", "", false)
 	}
 
 	// create installplan if anything updated
@@ -1020,7 +1014,8 @@ func (o *Operator) syncResolvingNamespace(obj interface{}) error {
 			logger.WithError(err).Debug("error ensuring installplan")
 			return err
 		}
-		if err := o.updateSubscriptionStatus(namespace, maxGeneration+1, updatedSubs, installPlanReference); err != nil {
+		updatedSubs = o.setIPReference(updatedSubs, maxGeneration+1, installPlanReference)
+		if _, err := o.updateSubscriptionStatuses(updatedSubs); err != nil {
 			logger.WithError(err).Debug("error ensuring subscription installplan state")
 			return err
 		}
@@ -1140,12 +1135,8 @@ func (o *Operator) ensureSubscriptionCSVState(logger *logrus.Entry, sub *v1alpha
 	return updatedSub, true, nil
 }
 
-func (o *Operator) updateSubscriptionStatus(namespace string, gen int, subs []*v1alpha1.Subscription, installPlanRef *corev1.ObjectReference) error {
+func (o *Operator) setIPReference(subs []*v1alpha1.Subscription, gen int, installPlanRef *corev1.ObjectReference) []*v1alpha1.Subscription {
 	var (
-		errs        []error
-		mu          sync.Mutex
-		wg          sync.WaitGroup
-		getOpts     = metav1.GetOptions{}
 		lastUpdated = o.now()
 	)
 	for _, sub := range subs {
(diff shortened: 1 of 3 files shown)


pkg/controller/operators/catalog/operator.go
```diff
@@ -976,22 +976,23 @@ func (o *Operator) syncResolvingNamespace(obj interface{}) error {
...
 		}
-		updateErr := o.setSubsCond(subs, v1alpha1.SubscriptionResolutionFailed, "ErrorPreventedResolution", err.Error(), true)
+		_, updateErr := o.setSubsCond(subs, v1alpha1.SubscriptionResolutionFailed, "ErrorPreventedResolution", err.Error(), true)
 		if updateErr != nil {
 			logger.WithError(updateErr).Debug("failed to update subs conditions")
 		}
 		return err
 	} else {
-		updateErr := o.setSubsCond(subs, v1alpha1.SubscriptionResolutionFailed, "", "", false)
+		subs, updateErr := o.setSubsCond(updatedSubs, v1alpha1.SubscriptionResolutionFailed, "", "", false)
```
benluddy, line 991: Why does this only reset `updatedSubs`?

pkg/controller/operators/catalog/operator.go
```diff
@@ -1284,26 +1285,30 @@ func (o *Operator) setSubsCond(subs []*v1alpha1.Subscription, condType v1alpha1.
...
+				latest, err := o.client.OperatorsV1alpha1().Subscriptions(sub.GetNamespace()).Get(context.TODO(), sub.GetName(), getOpts)
 				if err != nil {
 					return err
 				}
-
-				latest.Status = s.Status
-				_, err = o.client.OperatorsV1alpha1().Subscriptions(s.Namespace).UpdateStatus(context.TODO(), latest, updateOpts)
-
+				latest.Status = sub.Status
+				_, err = o.client.OperatorsV1alpha1().Subscriptions(sub.Namespace).UpdateStatus(context.TODO(), latest, updateOpts)
```
benluddy, line 1311: Should the entry in `subs` be updated with the updated Subscription from here? Since the ResourceVersion will change, I imagine that this guarantees a 409 on any other updates that happen during this sync.

pkg/controller/operators/catalog/operator.go
anik120: @benluddy do you mean why are the other calls to `setSubsCond` also not using `updatedSubs`?

benluddy: No -- if only some of the Subscriptions in the namespace are present in `updatedSubs`, what is going to clear this status condition for the others?

pkg/controller/operators/catalog/operator.go
anik120: Updated the entry with the latest fetched.

pkg/controller/operators/catalog/operator.go
anik120: I see. This is probably highlighting a design choice we made, i.e the decision to project this to all subscriptions in the namespace. 

The point of this fix is that inside this `else` block, we work with the updated subs so that we can set the updated subs with ResolutionFailed condition reset to false, and that is the list of subs that the rest of the method continues with. 

If we pass in the full list of `subs` instead  of just the `updatedSubs`, we get back the list of all the subs, that is then overriden [here](https://github.com/operator-framework/operator-lifecycle-manager/pull/2296/files#diff-a1760d9b7ac1e93734eea675d8d8938c96a50e995434b163c6f77c91bace9990R995), i.e we have the wrong set of `updatedSubs` at that point.

anik120: Possible solutions include: 
1. We change the decision to project to all subscriptions in the namespace, and instead project the condition to only those subscriptions that are involved in the resolution step. 
2. We add extra logic to list all the subscriptions in the namespace after the `updatedSubs` are updated, and call the `setSubCond` function again. With the introduction of the check inside `setSubCond` to skip subscriptions that already have the desired state, we can be sure we're not doing the same work twice.

anik120: Updated PR with option 2.

pkg/controller/operators/catalog/operator.go
```diff
@@ -1284,26 +1296,31 @@ func (o *Operator) setSubsCond(subs []*v1alpha1.Subscription, condType v1alpha1.
...
+				continue
+			}
 			cond.Status = corev1.ConditionTrue
 		} else {
+			if cond.Status == corev1.ConditionFalse {
+				continue
+			}
 			cond.Status = corev1.ConditionFalse
 		}
 		sub.Status.SetCondition(cond)
```
ecordell, line 1310: What do you think about splitting this function here: something like `setConditions(subs)` and `updateStatuses(subs)`

Then you can use `setConditions` over `updatedSubs` and `setConditions` + `updateStatus` over `allSubs` (right now this is hitting the api twice for all `updatedSubs`)

(I think there's bonus points if you find a nice way to align this with the existing `updateSubscriptionStatus` method)

anik120: @ecordell I liked that idea. Please review the latest commit.

pkg/controller/operators/catalog/operator.go
```diff
@@ -976,22 +976,28 @@ func (o *Operator) syncResolvingNamespace(obj interface{}) error {
...
+			return updateErr
 		}
 		return err
 	} else {
-		updateErr := o.setSubsCond(subs, v1alpha1.SubscriptionResolutionFailed, "", "", false)
+		subs = o.setSubsCond(subs, v1alpha1.SubscriptionResolutionFailed, "", "", false)
+		_, updateErr := o.updateSubscriptionStatuses(subs)
 		if updateErr != nil {
 			logger.WithError(updateErr).Debug("failed to update subs conditions")
 		}
```
ecordell, line 999:
```suggestion

```

Can't this be deleted, since in the success case the status will be updated anyway?

anik120: 1. The rest of the lines in this method works with `updatedSubs`, so there's nothing else updating the conditions of all the subs after this
2. The lines that do work with `updateSubs`, they only set the install plan references for the updated subs. They do not clear the conditions that might have already been set.

pkg/controller/operators/catalog/operator.go
```diff
@@ -976,22 +976,28 @@ func (o *Operator) syncResolvingNamespace(obj interface{}) error {
...
-		updateErr := o.setSubsCond(subs, v1alpha1.SubscriptionResolutionFailed, "ErrorPreventedResolution", err.Error(), true)
+		subs = o.setSubsCond(subs, v1alpha1.SubscriptionResolutionFailed, "ErrorPreventedResolution", err.Error(), true)
+		_, updateErr := o.updateSubscriptionStatuses(subs)
 		if updateErr != nil {
 			logger.WithError(updateErr).Debug("failed to update subs conditions")
+			return updateErr
 		}
 		return err
 	} else {
```
benluddy, line 994: The "else" is unnecessary if the function is guaranteed to return on error, isn't it?

pkg/controller/operators/catalog/operator.go
```diff
@@ -976,22 +976,28 @@ func (o *Operator) syncResolvingNamespace(obj interface{}) error {
...
-		updateErr := o.setSubsCond(subs, v1alpha1.SubscriptionResolutionFailed, "", "", false)
+		subs = o.setSubsCond(subs, v1alpha1.SubscriptionResolutionFailed, "", "", false)
+		_, updateErr := o.updateSubscriptionStatuses(subs)
```
benluddy, line 996: Does it make sense to exclude from `subs` any Subscriptions in `updatedSubs` that are going to be updated again on L1033?

pkg/controller/operators/catalog/operator.go
```diff
@@ -976,22 +976,28 @@ func (o *Operator) syncResolvingNamespace(obj interface{}) error {
...
 		if updateErr != nil {
 			logger.WithError(updateErr).Debug("failed to update subs conditions")
```
benluddy, line 998: This one might not be visible as a sync error, and it might mean that there is a Subscription with a misleading status, so logging it at warning seems useful to me:

```suggestion
			logger.WithError(updateErr).Warn("failed to update subscription conditions")
```

pkg/controller/operators/catalog/operator.go
anik120: @benluddy here we call `setSubsCond` and then `updateSubscriptionStatuses`, which results in the subscriptions being updated on cluster.
The call to `setSubsCond` for `updateSubs` again on L1033 is just to make sure `updatedSubs` are in sync with all `subs`, otherwise
1. the subscriptions get updated in L996
2. `updateSubs` however do not have the recent picture of the subscriptions (i.e the condition struct for `updatedSubs` is still the old condition)
3. The subscriptions are updated again when `updateSubs` are updated, i.e the work done in step 1 is overriden for `updatedSubs`

Given that, if the suggestion is to only update the subscriptions in L996 that are not in `updatedSubs`, the update to the `updateSubs` only happen after this in L1033, before which the function could return in multiple places if it encounters an error. That'd leave us in a state where only some subscriptions are updated, while others are not. 

It makes sense to update all the subscriptions here.

anik120: On second thought, reducing the number of times we update the subscription statues sounds like a good idea. Added a new commit to do that.

pkg/controller/operators/catalog/operator.go
```diff
@@ -1269,26 +1246,44 @@ func (o *Operator) setSubsCond(subs []*v1alpha1.Subscription, condType v1alpha1.
 		cond.Reason = reason
 		cond.Message = message
 		if setTrue {
+			if cond.Status == corev1.ConditionTrue {
+				continue
+			}
```
joelanford, line 1251: This change means that if the existing sub has:
```yaml
- type: Foo
  status: True
  reason: Hello
  message: World
```

And we call this function with `setSubsCond([]{sub}, "Foo", "True",  "Goodnight", "Moon")`, then the resulting condition will not have it's reason and message updated to "Goodnight" and "Moon", right?

That seems a little bit counter-intuitive to me.

pkg/controller/operators/catalog/operator.go
```diff
@@ -1269,26 +1246,44 @@ func (o *Operator) setSubsCond(subs []*v1alpha1.Subscription, condType v1alpha1.
...
+				latest, err := o.client.OperatorsV1alpha1().Subscriptions(sub.GetNamespace()).Get(context.TODO(), sub.GetName(), getOpts)
 				if err != nil {
 					return err
 				}
-
-				latest.Status = s.Status
-				_, err = o.client.OperatorsV1alpha1().Subscriptions(s.Namespace).UpdateStatus(context.TODO(), latest, updateOpts)
-
+				latest.Status = sub.Status
+				sub = *latest
```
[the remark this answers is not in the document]
anik120: @joelanford you're right that the intent is to make sure `subs` has the latest version of the particular `sub` in it. And also, the `sub` inside the goroutine is what's being changed here. But if you notice in L1294, it's the individual `sub` from `subs` that's being passed to the goroutines, so it's effectively achieving the same objective as what the options you laid out would.

pkg/controller/operators/catalog/operator.go
anik120: @joelanford that's a good catch. However, in this particular case, our objective for now is to optimize the number of times we hit the api server to update the subscriptions. And the only call site where `setSubsCond` is being called for do not have a use case for changing the `reason` and `message` of a condition. 

Which means getting rid of the check and letting the update go through regardless would be sacrificing current optimization for a potential future use case. We can always get rid of the check if a similar use case to the one you mentioned arises later on.

benluddy: What if the reason changes from ErrorPreventedResolution to ConstraintsNotSatisfiable?

anik120: So I went and rechecked, and just above this if else block, the `Reason` and `Message` is being set regardless of whether the status is already in the expected state. So the answer to Joe's original questions is, no, it will set "Goodnight" and "Moon". 

However, I just realized that that if else block is a hangover from when `updateSubscriptionStatuses` was not separated out, so right now when we call that we're reaching out to the api server for all subscriptions regardless. So the `continue` here is unnecessary and looks like it's just creating more confusion. Adding a commit to remove that.

anik120: Done.

PR #2299, opened 2021-07-28
.github/workflows: Avoid running the e2e-tests workflow on doc-only PRs

timflannagan:
Update the e2e-tests.yml github workflow and avoid running that action
on PRs that only change the doc directory.

**Reviewer Checklist**

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -4,6 +4,9 @@ on:
     branches:
       - master
   pull_request:
+    paths:
+      - '**'
+      - '!doc/**'
 jobs:
   run-e2e-tests:
     runs-on: ubuntu-latest


.github/workflows/e2e-tests.yml
```diff
@@ -4,6 +4,9 @@ on:
     branches:
       - master
   pull_request:
+    paths:
+      - '**'
+      - '!doc/**'
```
joelanford, line 9: Nit: also ignore markdown files? 
```suggestion
      - '!doc/**'
      - '!**/*.md'
```

joelanford (commented):
Just a non-blocking nit.

/lgtm

.github/workflows/e2e-tests.yml
timflannagan: Yeah that's not a bad idea - any scenario where we would only want to restrict that to _root_ markdown files? Or extend this to changes made to similar files, like `.gitignore`? I'm mainly following suit w.r.t to the other workflows that disable that doc/ directory path, so maybe that's worth tackling down the line.

PR #2302, opened 2021-07-29
fix(openshift): block upgrades on invalid max properties

njhale:
~~Fail open when olm.maxOpenShiftVersion properties have invalid
values. This means that all invalid values are logged but are ultimately
ignored when determining cluster upgradeability.~~

Block OpenShift upgrades while:

- olm.maxOpenShiftVersion properties have invalid values
- cluster information is unavailable; e.g. the desired version of the cluster is undefined
- an installed operator has declared more than one olm.MaxOpenShiftVersion
  property

See https://bugzilla.redhat.com/show_bug.cgi?id=1986753 for motivation.

### DIFF
--- a/pkg/controller/operators/openshift/clusteroperator_controller.go
+++ b/pkg/controller/operators/openshift/clusteroperator_controller.go
@@ -209,7 +209,8 @@ func (r *ClusterOperatorReconciler) setDegraded(_ context.Context, co *ClusterOp
 }
 
 const (
-	IncompatibleOperatorsInstalled = "IncompatibleOperatorsInstalled"
+	IncompatibleOperatorsInstalled     = "IncompatibleOperatorsInstalled"
+	ErrorCheckingOperatorCompatibility = "ErrorCheckingOperatorCompatibility"
 )
 
 func (r *ClusterOperatorReconciler) setUpgradeable(ctx context.Context, co *ClusterOperator) error {
@@ -221,31 +222,39 @@ func (r *ClusterOperatorReconciler) setUpgradeable(ctx context.Context, co *Clus
 
 	// Set upgradeable=false if (either/or):
 	// 1. OLM currently upgrading (takes priorty in the status message)
-	// 2. Operators currently installed that are incompatible with the next OCP minor version
+	// 2. Operators currently installed that are incompatible with the next minor version of OpenShift
+	// 3. An error occurs while determining 2
+	var err error
 	if r.syncTracker.SuccessfulSyncs() < 1 || !versionsMatch(co.Status.Versions, r.TargetVersions) {
 		// OLM is still upgrading
 		desired.Status = configv1.ConditionFalse
 		desired.Message = "Waiting for updates to take effect"
 	} else {
-		incompatible, err := incompatibleOperators(ctx, r.Client)
+		var incompatible skews
+		incompatible, err = incompatibleOperators(ctx, r.Client)
 		if err != nil {
-			return err
-		}
-
-		if len(incompatible) > 0 {
-			// Some incompatible operator is installed
+			// "Fail closed" when we can't determine compatibility
+			// Note: Unspecified compatibility = universal compatibility; i.e. operators that don't specify a "maxOpenShiftVersion" property are compatible with everything.
+			desired.Status = configv1.ConditionFalse
+			desired.Reason = ErrorCheckingOperatorCompatibility
+			desired.Message = fmt.Sprintf("Encountered errors while checking compatibility with the next minor version of OpenShift: %s", err)
+		} else if len(incompatible) > 0 {
+			// Operators are installed that have incompatible and/or invalid max versions
 			desired.Status = configv1.ConditionFalse
 			desired.Reason = IncompatibleOperatorsInstalled
-			desired.Message = incompatible.String() // TODO: Truncate message to field length
+			desired.Message = incompatible.String()
 		}
 	}
 
+	// Only return transient errors likely resolved by retrying immediately
+	err = transientErrors(err)
+
 	current := co.GetCondition(configv1.OperatorUpgradeable)
 	if conditionsEqual(current, desired) { // Comparison ignores lastUpdated
-		return nil
+		return err
 	}
 
 	co.SetCondition(desired)
 
-	return nil
+	return err
 }
--- a/pkg/controller/operators/openshift/clusteroperator_controller_test.go
+++ b/pkg/controller/operators/openshift/clusteroperator_controller_test.go
@@ -142,8 +142,35 @@ var _ = Describe("ClusterOperator controller", func() {
 			LastTransitionTime: fixedNow(),
 		}))
 
-		By("setting upgradeable=false when incompatible operators exist")
+		By("setting upgradeable=false when there's an error determining compatibility")
+		cv.Status = configv1.ClusterVersionStatus{}
 
+		Eventually(func() error {
+			return k8sClient.Status().Update(ctx, cv)
+		}).Should(Succeed())
+
+		Eventually(func() ([]configv1.ClusterOperatorStatusCondition, error) {
+			err := k8sClient.Get(ctx, clusterOperatorName, co)
+			return co.Status.Conditions, err
+		}, timeout).Should(ContainElement(configv1.ClusterOperatorStatusCondition{
+			Type:               configv1.OperatorUpgradeable,
+			Status:             configv1.ConditionFalse,
+			Reason:             ErrorCheckingOperatorCompatibility,
+			Message:            "Encountered errors while checking compatibility with the next minor version of OpenShift: Desired release version missing from ClusterVersion",
+			LastTransitionTime: fixedNow(),
+		}))
+
+		cv.Status = configv1.ClusterVersionStatus{
+			Desired: configv1.Update{
+				Version: clusterVersion,
+			},
+		}
+
+		Eventually(func() error {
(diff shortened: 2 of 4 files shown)


joelanford:
Is ignoring invalid max versions the right thing to do here? It seems like it would be fairly easy for someone to accidentally fat-finger a version number and end up getting an upgrade when one isn't actually supported, which could then lead to an even worse situation.

Is it possible to error out and notify the user in some way that they have an invalid property value?

exdx:
> Is ignoring invalid max versions the right thing to do here? It seems like it would be fairly easy for someone to accidentally fat-finger a version number and end up getting an upgrade when one isn't actually supported, which could then lead to an even worse situation.
> 
> Is it possible to error out and notify the user in some way that they have an invalid property value?

I agree, maybe it safer to fail closed versus fail open when its a (delicate) update process?

njhale:
@joelanford @exdx 

Thanks for the quick review.

> Is it possible to error out and notify the user in some way that they have an invalid property value?

I had the same thought as I was implementing this. I'll update the logic to _fail closed_ and bubble an error to the ClusterVersion when an unparsable max property is found.

(part of the reason I did it like this in the first place was the expectation of QE to fail open in the original BZ)

njhale:
/hold

until I update this

njhale:
/hold cancel

updated behavior to address feedback.

pkg/controller/operators/openshift/helpers.go
```diff
@@ -178,57 +205,37 @@ func maxOpenShiftVersion(csv *operatorsv1alpha1.ClusterServiceVersion) (*semver.
 		return nil, err
 	}
 
-	// Take the highest semver if there's more than one max version specified
+	// Take the highest semver if there's more than one max version specified.
+	// Note: duplicate versions are not a problem because the intent to block a specific X.Y-stream is still clear.
```
joelanford, line 209: I feel like an argument could be made for accepting the lowest semver. If we accept highest, would that mean we'd technically be violating all the lower semver constraints?

njhale: I can see that argument.

I was thinking that it would be more common for authors to add higher versions, as they test compatibility with new OpenShift releases, than it would be to add lower versions over time.

Another route could be to require the max version property to be unique; i.e. be less tolerant of sloppy property maintenance.

joelanford: Ideally, I think enforcing uniqueness is the right answer, mainly because that removes all ambiguity.

I guess the thing to think about is where the properties come from and if we think there's a use case for them being additive like you suggest.

IMO, with an eye towards declarative config, and additive style flow sounds more imperative and probably not something we want to encourage or reinforce, so I think I'd vote for a uniqueness requirement on this property.

njhale: okay, you've convinced me that enforcing uniqueness is the best choice here. Will update.

/hold

njhale:
/hold cancel

added behavior

pkg/controller/operators/openshift/helpers.go
```diff
@@ -105,58 +131,59 @@ func incompatibleOperators(ctx context.Context, cli client.Client) (skews, error
 	}
 
 	if next == nil {
-		return nil, nil
+		// Note: This shouldn't happen
+		return nil, fmt.Errorf("Failed to determine next OpenShift Y-stream release")
 	}
 	next.Minor++
 
 	csvList := &operatorsv1alpha1.ClusterServiceVersionList{}
```
exdx, line 139: nit: is it possible to build this list excluding copied CSVs via the annotation, so we don't potentially make a huge cache call? Not sure how much that matters though.

pkg/controller/operators/openshift/helpers.go
```diff
@@ -105,58 +131,59 @@ func incompatibleOperators(ctx context.Context, cli client.Client) (skews, error
...
-		return nil, err
+		return nil, &transientError{fmt.Errorf("Failed to list ClusterServiceVersions: %w", err)}
 	}
 
-	var (
-		s    skews
-		errs []error
-	)
+	var incompatible skews
 	for _, csv := range csvList.Items {
```
exdx, line 145: One case I'm thinking about is where Operator A has a defined maxOpenshiftVersion property, but depends on Operator B which does not have one defined. In that case, on upgrade Operator A will report ready, so will Operator B, but B may break, breaking A. 

Catching this case would require doing introspection on the CSV, and checking any dependencies for their properties, which may be overly restrictive. But it's a scenario you could imagine happening on the cluster somewhat frequently if a popular dependency doesn't specify its maxOpenshiftVersion. Not sure how important this is to address up-front though.

exdx (commented):
Just some thoughts, we can let @joelanford review as well. 
/lgtm
/hold

pkg/controller/operators/openshift/helpers.go
njhale: Maybe we could do something with the newly introduced metadata client?

My preference would be to optimize in a follow up, since the topic of the PR is potentially release blocking.

pkg/controller/operators/openshift/helpers.go
njhale: > In that case, on upgrade Operator A will report ready, so will Operator B, but B may break, breaking A.

could you expand on how this relates to blocking cluster upgrades?

exdx: I guess it would not block cluster upgrades per se but this scenario would break the Operator that was using the maxOpenShiftVersion property, since its dependencies don't specify max version

pkg/controller/operators/openshift/clusteroperator_controller_test.go
```diff
@@ -245,5 +275,54 @@ var _ = Describe("ClusterOperator controller", func() {
...
+			return co.Status.Conditions, err
+		}, timeout).Should(ContainElement(configv1.ClusterOperatorStatusCondition{
+			Type:   configv1.OperatorUpgradeable,
+			Status: configv1.ConditionFalse,
+			Reason: IncompatibleOperatorsInstalled,
+			Message: skews{
+				{
+					namespace: ns.GetName(),
+					name:      incompatible.GetName(),
+					err:       fmt.Errorf(`Defining more than one "%s" property is not allowed`, MaxOpenShiftVersionProperty),
```
joelanford, line 322: For future reference, you can also use `%q` instead of "%s" to get a quoted string, which helps when you'd rather use double quotes to define the string literal.

pkg/controller/operators/openshift/helpers.go
```diff
@@ -105,58 +131,59 @@ func incompatibleOperators(ctx context.Context, cli client.Client) (skews, error
...
 
-	return s, utilerrors.NewAggregate(errs)
+	return incompatible, nil
 }
 
 func desiredRelease(ctx context.Context, cli client.Client) (*semver.Version, error) {
 	cv := configv1.ClusterVersion{}
 	if err := cli.Get(ctx, client.ObjectKey{Name: "version"}, &cv); err != nil { // "version" is the name of OpenShift's ClusterVersion singleton
-		return nil, err
+		return nil, &transientError{fmt.Errorf("Failed to get ClusterVersion: %w", err)}
```
joelanford, line 175: One thing that just crossed my mind: Is this section of code disabled in non-OpenShift clusters? I assume `maxOpenshiftVersion` is just ignored unless we're in Openshift?

timflannagan: It looks like the only call site for this function is in `incompatibleOperators`, and that function is only used in the context of the ClusterOperator controller that handles the maxOpenshiftVersion logic.

I guess we gate the creation of this controller on the [`--writeStatusName` CLI flag](https://github.com/operator-framework/operator-lifecycle-manager/blob/b044fdcccb00bc7859956ca3f3500d85b3af862f/cmd/olm/main.go#L226) that we set downstream, which seems like a relatively loose check (couldn't we discovery this API dynamically instead?), so tl;dr should be ignored on non-OCP clusters.

njhale: @joelanford, @timflannagan hit the nail on the head: the entire controller is disabled unless `--writeStatusName` is given when running `olm-operator`.

pkg/controller/operators/openshift/clusteroperator_controller.go
```diff
@@ -221,31 +222,39 @@ func (r *ClusterOperatorReconciler) setUpgradeable(ctx context.Context, co *Clus
...
-		if len(incompatible) > 0 {
-			// Some incompatible operator is installed
+			// "Fail closed" when we can't determine compatibility
+			// Note: Unspecified compatibility = universal compatibility; i.e. operators that don't specify a "maxOpenShiftVersion" property are compatible with everything.
+			desired.Status = configv1.ConditionFalse
+			desired.Reason = ErrorCheckingOperatorCompatibility
+			desired.Message = fmt.Sprintf("Encountered errors while checking compatibility with the next minor version of OpenShift: %s", err)
+		} else if len(incompatible) > 0 {
+			// Operators are installed that have incompatible and/or invalid max versions
 			desired.Status = configv1.ConditionFalse
```
timflannagan, line 243: Just trying to wrap my head around whether it makes sense to specify configv1.ConditionTrue if we're setting the reason to `IncompatibleOperatorsInstalled` when encountering incompatibility?

timflannagan: Disregard - I realize I was missing the full context and needed to look at the full function control flow.

PR #2303, opened 2021-07-29
cmd/catalog: Default to the $KUBECONFIG environment variable when building client configs

timflannagan:
Update the catalog operator and default to the $KUBECONFIG environment
variable value when the --kubeconfig flag hasn't been provided. This
mirrors the same way that the olm operator (e.g. controller-runtime is
handling this check under-the-hood) behaves.

**Reviewer Checklist**

### DIFF
--- a/cmd/catalog/main.go
+++ b/cmd/catalog/main.go
@@ -38,7 +38,7 @@ const (
 // config flags defined globally so that they appear on the test binary as well
 var (
 	kubeConfigPath = flag.String(
-		"kubeconfig", "", "absolute path to the kubeconfig file")
+		"kubeconfig", os.Getenv("KUBECONFIG"), "absolute path to the kubeconfig file")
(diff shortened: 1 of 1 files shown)


cmd/catalog/main.go
```diff
@@ -37,7 +37,7 @@ const (
 // config flags defined globally so that they appear on the test binary as well
 var (
 	kubeConfigPath = flag.String(
-		"kubeconfig", "", "absolute path to the kubeconfig file")
+		"kubeconfig", os.Getenv("KUBECONFIG"), "absolute path to the kubeconfig file")
```
timflannagan, line 40: When providing the value through the CLI flag, that value still takes precedence over the local $KUBECONFIG environment variable, which seems like the correct behavior. Might be worth calling out that behavior in this flags description?

PR #2304, opened 2021-07-30
Add PriorityClass setting to registry pods for CatalogSource via annotations

dinhxuanvu:
The registry pods may need to have necessary priorityclass settings.
OLM will set the priorityclass setting for registry pods by using
the priorityclass annotations in the catalogsources.

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/registry/reconciler/grpc.go
+++ b/pkg/controller/registry/reconciler/grpc.go
@@ -171,15 +171,16 @@ func (c *GrpcRegistryReconciler) currentUpdatePods(source grpcCatalogSourceDecor
 	return pods
 }
 
-func (c *GrpcRegistryReconciler) currentPodsWithCorrectImage(source grpcCatalogSourceDecorator) []*corev1.Pod {
+func (c *GrpcRegistryReconciler) currentPodsWithCorrectImageAndSpec(source grpcCatalogSourceDecorator, saName string) []*corev1.Pod {
 	pods, err := c.Lister.CoreV1().PodLister().Pods(source.GetNamespace()).List(labels.SelectorFromValidatedSet(source.Labels()))
 	if err != nil {
 		logrus.WithError(err).Warn("couldn't find pod in cache")
 		return nil
 	}
 	found := []*corev1.Pod{}
+	newPod := source.Pod(saName)
 	for _, p := range pods {
-		if p.Spec.Containers[0].Image == source.Spec.Image {
+		if p.Spec.Containers[0].Image == source.Spec.Image && podHashMatch(p, newPod) {
 			found = append(found, p)
 		}
 	}
@@ -192,11 +193,12 @@ func (c *GrpcRegistryReconciler) EnsureRegistryServer(catalogSource *v1alpha1.Ca
 
 	// if service status is nil, we force create every object to ensure they're created the first time
 	overwrite := source.Status.RegistryServiceStatus == nil
-	// recreate the pod if no existing pod is serving the latest image
-	overwritePod := overwrite || len(c.currentPodsWithCorrectImage(source)) == 0
 
 	//TODO: if any of these error out, we should write a status back (possibly set RegistryServiceStatus to nil so they get recreated)
 	sa, err := c.ensureSA(source)
+	// recreate the pod if no existing pod is serving the latest image or correct spec
+	overwritePod := overwrite || len(c.currentPodsWithCorrectImageAndSpec(source, sa.GetName())) == 0
+
 	if err != nil && !k8serror.IsAlreadyExists(err) {
 		return errors.Wrapf(err, "error ensuring service account: %s", source.GetName())
 	}
@@ -421,10 +423,9 @@ func (c *GrpcRegistryReconciler) removePods(pods []*corev1.Pod, namespace string
 // CheckRegistryServer returns true if the given CatalogSource is considered healthy; false otherwise.
 func (c *GrpcRegistryReconciler) CheckRegistryServer(catalogSource *v1alpha1.CatalogSource) (healthy bool, err error) {
 	source := grpcCatalogSourceDecorator{catalogSource}
-
 	// Check on registry resources
 	// TODO: add gRPC health check
-	if len(c.currentPodsWithCorrectImage(source)) < 1 ||
+	if len(c.currentPodsWithCorrectImageAndSpec(source, source.ServiceAccount().GetName())) < 1 ||
 		c.currentService(source) == nil {
 		healthy = false
 		return
@@ -477,4 +478,31 @@ func (c *GrpcRegistryReconciler) podFailed(pod *corev1.Pod) (bool, error) {
 		return true, nil
 	}
 	return false, nil
+}
+
+// podHashMatch will check the hash info in existing pod to ensure its
+// hash info matches the desired Service's hash.
+func podHashMatch(existing, new *corev1.Pod) bool {
+	labels := existing.GetLabels()
+	newLabels := new.GetLabels()
+	// If both new & existing pods don't have labels, consider it not matched
+	if len(labels) == 0 || len(newLabels) == 0 {
+		return false
+	}
+
+	existingPodSpecHash, ok := labels[PodHashLabelKey]
+	if !ok {
+		return false
+	}
+
+	newPodSpecHash, ok := newLabels[PodHashLabelKey]
+	if !ok {
+		return false
+	}
+
+	if existingPodSpecHash != newPodSpecHash {
+		return false
+	}
+
+	return true
 }
--- a/pkg/controller/registry/reconciler/grpc_test.go
+++ b/pkg/controller/registry/reconciler/grpc_test.go
@@ -339,6 +339,74 @@ func TestGrpcRegistryReconciler(t *testing.T) {
 	}
 }
 
+func TestRegistryPodPriorityClass(t *testing.T) {
+	now := func() metav1.Time { return metav1.Date(2018, time.January, 26, 20, 40, 0, 0, time.UTC) }
+
+	type cluster struct {
+		k8sObjs []runtime.Object
+	}
+	type in struct {
+		cluster cluster
+		catsrc  *v1alpha1.CatalogSource
+	}
+	tests := []struct {
+		testName      string
+		in            in
+		priorityclass string
+	}{
+		{
+			testName: "Grpc/WithValidPriorityClassAnnotation",
+			in: in{
+				catsrc: grpcCatalogSourceWithAnnotations(map[string]string{
+					"operatorframework.io/priorityclass": "system-cluster-critical",
+				}),
+			},
+			priorityclass: "system-cluster-critical",
+		},
+		{
+			testName: "Grpc/WithInvalidPriorityClassAnnotation",
+			in: in{
+				catsrc: grpcCatalogSourceWithAnnotations(map[string]string{
+					"operatorframework.io/priorityclass": "",
+				}),
+			},
+			priorityclass: "",
+		},
+		{
+			testName: "Grpc/WithNoPriorityClassAnnotation",
+			in: in{
+				catsrc: grpcCatalogSourceWithAnnotations(map[string]string{
+					"annotationkey": "annotationvalue",
+				}),
+			},
+			priorityclass: "",
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.testName, func(t *testing.T) {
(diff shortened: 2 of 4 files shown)


pkg/controller/registry/reconciler/grpc.go
```diff
@@ -478,3 +495,29 @@ func (c *GrpcRegistryReconciler) podFailed(pod *corev1.Pod) (bool, error) {
 	}
 	return false, nil
 }
+
+// PodHashMatch will check the hash info in existing pod to ensure its
+// hash info matches the desired Service's hash.
+func PodHashMatch(existing, new *corev1.Pod) bool {
```
joelanford, line 501: Any reason this needs to be exported?

pkg/controller/registry/reconciler/reconciler.go
```diff
@@ -160,5 +173,28 @@ func Pod(source *v1alpha1.CatalogSource, name string, image string, saName strin
...
+	if labels == nil {
+		labels = make(map[string]string)
+	}
+	labels[PodHashLabelKey] = HashPodSpec(pod.Spec)
+	pod.SetLabels(labels)
 	return pod
 }
+
+// HashPodSpec calculates a hash given a copy of the pod spec
+func HashPodSpec(spec corev1.PodSpec) string {
```
joelanford, line 196: Any reason this needs to be exported?

pkg/controller/registry/reconciler/grpc.go
```diff
@@ -478,3 +495,29 @@ func (c *GrpcRegistryReconciler) podFailed(pod *corev1.Pod) (bool, error) {
...
+	labels := existing.GetLabels()
+	newLabels := new.GetLabels()
+	if len(labels) == 0 || len(newLabels) == 0 {
+		return false
+	}
```
joelanford, line 506: Just double checking: this means that if neither `existing` nor `new` have the expected label, we consider that to mean they don't match? If so, is that worth a comment?

pkg/controller/registry/reconciler/reconciler.go
```diff
@@ -19,6 +24,14 @@ type nowFunc func() metav1.Time
 const (
 	// CatalogSourceLabelKey is the key for a label containing a CatalogSource name.
 	CatalogSourceLabelKey string = "olm.catalogSource"
+	// RedHatCatalogAnnotationKey is the key of an annotation in default catalogsources
+	RedHatCatalogAnnotationKey string = "operatorframework.io/managed-by"
+	// RedHatCatalogAnnotationValue is the value of an annotation in default catalogsources
+	RedHatCatalogAnnotationValue string = "marketplace-operator"
+	// RedHatCatalogPriorityClassKey is the key of an annotation in default catalogsources
+	RedHatCatalogPriorityClassKey string = "operatorframework.io/priorityclass"
```
joelanford, line 32: Do these constant names need "RedHat" in the name? Can we choose more upstream-friendly names?

pkg/controller/registry/reconciler/reconciler.go
```diff
@@ -160,5 +173,28 @@ func Pod(source *v1alpha1.CatalogSource, name string, image string, saName strin
 			ServiceAccountName: saName,
 		},
 	}
+
+	// Check if this catalogsource is a default one via annotation and set
+	// priorityclass if its annotation exists
+	if annotations[RedHatCatalogAnnotationKey] == RedHatCatalogAnnotationValue {
```
joelanford, line 179: Is there something special about the marketplace-operator managed catalog sources that says we should only apply this priority class label to its catalog sources? Why not just honor this annotation on all catalog sources?

pkg/controller/registry/reconciler/reconciler.go
timflannagan: Yeah agreed - I think prefixing variables with ocp/redhat/etc. makes sense if it's a downstream-specific API, but priority classes are an upstream concept: https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/#pod-priority.

pkg/controller/registry/reconciler/reconciler.go
```diff
@@ -2,14 +2,19 @@
 package reconciler
 
 import (
+	"fmt"
+	"hash/fnv"
 	"strings"
 
+	corev1 "k8s.io/api/core/v1"
```
timflannagan, line 9: nit, not blocking: there's already an existing k8s.io/api/core/v1 import declared below this.

timflannagan:
(I removed the bug from the PR title so automation doesn't pick this up once it merges)

timflannagan:
I think this implementation looks reasonable to me - I'd like to remove any downstream-specific handling here and just treat propagating priorityclass specifications into the catalogsource pods like Joe had eluded to in another comment.

pkg/controller/registry/reconciler/grpc.go
dinhxuanvu: It doesn't need to so I can lowercase it.

pkg/controller/registry/reconciler/reconciler.go
dinhxuanvu: Ditto

pkg/controller/registry/reconciler/reconciler.go
dinhxuanvu: The original request was to only add priorityclass to default catsrc only. However, it seems we can open to anyone so I will remove this gating.

dinhxuanvu:
Resolved and updated.

pkg/controller/registry/reconciler/grpc.go
```diff
@@ -186,17 +186,34 @@ func (c *GrpcRegistryReconciler) currentPodsWithCorrectImage(source grpcCatalogS
 	return found
 }
 
+func (c *GrpcRegistryReconciler) currentPodsWithCorrectHash(source grpcCatalogSourceDecorator, saName string) []*corev1.Pod {
+	pods, err := c.Lister.CoreV1().PodLister().Pods(source.GetNamespace()).List(labels.SelectorFromValidatedSet(source.Labels()))
```
timflannagan, line 190: Is it worth the effort on refactoring this and the above method to avoid making the same pod listing call, and instead attempt to return a filtered set of pods given a pod list parameter being provided at the call site? You would then make the list command at the EnsureRegistryServer call site, pass that podList variable to both of the currentPodWith* methods. This can likely be tackled in a follow-up though as I think you'd want a more functional refactoring than what I had suggested.

pkg/controller/registry/reconciler/reconciler.go
```diff
@@ -160,5 +168,25 @@ func Pod(source *v1alpha1.CatalogSource, name string, image string, saName strin
...
+	if prio, ok := annotations[CatalogPriorityClassKey]; ok && prio != "" {
+		pod.Spec.PriorityClassName = prio
+	}
+
+	// Add PodSpec hash
+	// This hash info will be used to detect PodSpec changes
+	if labels == nil {
+		labels = make(map[string]string)
+	}
+	labels[PodHashLabelKey] = hashPodSpec(pod.Spec)
```
timflannagan, line 182: I'm trying to wrap my head around whether this check should live further up in this functions control flow as the `labels` variable is provided as a parameter to the function. Is there any reason we'd want to perform these checks _before_ we build up the Pod type, as it looks like we blindly assign the parameter to the corresponding field in the metadata struct? Just thinking aloud and trying to determine if there's some edge case I'm missing here, besides we're potentially making duplicate calls.

pkg/controller/registry/reconciler/grpc.go
dinhxuanvu: I was thinking the same as well but I decided not to combine due to the fact that func is for something else that is not necessarily relevant to this fix. However, on a second thought, I will give it a shot.

pkg/controller/registry/reconciler/reconciler.go
dinhxuanvu: I'm not sure what you mean by commenting on this line specifically given there is no check here, just a label addition. This `Pod()` is where we create the Pod template for every registry pod so it is definite the place to add labels/annotations if needed.

timflannagan: Sorry, I meant we might want to perform this `if labels == nil` check at the top of the function, as the `labels` variable is provided as a parameter to the Pod function, not that we would move that check outside of this function. We blindly assign the metadata.Labels to that variable when building up the Pod type definition earlier in the function.

dinhxuanvu: The default value of a map is a nil, so even if `labels` is nil, assigning to a nil to the pod metadata is the same as omitting the labels altogether. So I don't see why we need to check it earlier. I have to check it later because I will assign something to the map and you cannot assign to nil map which will lead to nil panic. Hence, the check is there.

timflannagan: Yeah, I'm not super concerned about that anymore as I don't think there are any edge cases to worry about. I think this is more my preference at this point that we perform any upfront manipulations of the function parameters before we assign them to the Pod resource for readability.

pkg/controller/registry/reconciler/grpc_test.go
```diff
@@ -339,6 +339,74 @@ func TestGrpcRegistryReconciler(t *testing.T) {
...
+			testName: "Grpc/WithValidPriorityClassAnnotation",
+			in: in{
+				catsrc: grpcCatalogSourceWithAnnotations(map[string]string{
+					"operatorframework.io/priorityclass": "system-cluster-critical",
+				}),
+			},
+			priorityclass: "system-cluster-critical",
+		},
+		{
+			testName: "Grpc/WithInvalidPriorityClassAnnotation",
```
timflannagan, line 367: (not blocking) Technically not _invalid_ but we won't propagate any empty key value to the priorityclass default.

pkg/controller/registry/reconciler/grpc_test.go
```diff
@@ -339,6 +339,74 @@ func TestGrpcRegistryReconciler(t *testing.T) {
...
+			// Check for resource existence
+			decorated := grpcCatalogSourceDecorator{tt.in.catsrc}
+			pod := decorated.Pod(tt.in.catsrc.GetName())
+			listOptions := metav1.ListOptions{LabelSelector: labels.SelectorFromSet(labels.Set{CatalogSourceLabelKey: tt.in.catsrc.GetName()}).String()}
+			outPods, podErr := client.KubernetesInterface().CoreV1().Pods(pod.GetNamespace()).List(context.TODO(), listOptions)
+			require.NoError(t, podErr)
+			require.Len(t, outPods.Items, 1)
+			outPod := outPods.Items[0]
+			require.Equal(t, tt.priorityclass, outPod.Spec.PriorityClassName)
+			require.Equal(t, pod.GetLabels()[PodHashLabelKey], outPod.GetLabels()[PodHashLabelKey])
```
timflannagan, line 405: Not super concerned with blindingly indexing like this, but it might produce test flakes down the line.

pkg/controller/registry/reconciler/grpc.go
```diff
@@ -171,15 +171,16 @@ func (c *GrpcRegistryReconciler) currentUpdatePods(source grpcCatalogSourceDecor
 	return pods
 }
 
-func (c *GrpcRegistryReconciler) currentPodsWithCorrectImage(source grpcCatalogSourceDecorator) []*corev1.Pod {
+func (c *GrpcRegistryReconciler) currentPodsWithCorrectImageAndSpec(source grpcCatalogSourceDecorator, saName string) []*corev1.Pod {
```
timflannagan, line 174: Not technically what I had in mind but works for now :laughing:

pkg/controller/registry/reconciler/grpc_test.go
```diff
@@ -339,6 +339,74 @@ func TestGrpcRegistryReconciler(t *testing.T) {
...
+					"annotationkey": "annotationvalue",
+				}),
+			},
+			priorityclass: "",
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.testName, func(t *testing.T) {
+			stopc := make(chan struct{})
+			defer close(stopc)
```
timflannagan, line 388: Is this defer call respected inside of a t.Run? What happens if there's an uncaught panic here?

timflannagan: (Disregard - I realize I'm thinking of https://github.com/golang/go/issues/31651).

PR #2305, opened 2021-07-30
fix(og): Fix missing MultiOperatorGroups condition in some cases

dinhxuanvu:
In some special cases, the MultiOperatorGroups condition is missing
even though there are multiple OGs in the same namespace. The process
of adding this condition happens during syncNamespace which sometimes
doesn't happen if syncOperatorGroups fails prematurely due to other
errors. Moving the MultiOperatorGroups condition to syncOperatorGroups
to ensure it will be run everytime.

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/olm/operatorgroup.go
+++ b/pkg/controller/operators/olm/operatorgroup.go
@@ -13,6 +13,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	meta "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/errors"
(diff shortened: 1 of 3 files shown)


pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -64,8 +65,51 @@ func (a *Operator) syncOperatorGroups(obj interface{}) error {
...
+	if err != nil {
+		logger.WithError(err).Warnf("failed to list OperatorGroups in the namespace")
+	}
+
+	// Check if there is a stale multiple OG condition and clear it if existed.
+	if len(groups) == 1 {
+		og := groups[0]
+		if c := meta.FindStatusCondition(og.Status.Conditions, v1.MutlipleOperatorGroupCondition); c != nil {
+			meta.RemoveStatusCondition(&og.Status.Conditions, v1.MutlipleOperatorGroupCondition)
+			if og.GetName() == op.GetName() {
```
awgreene, line 79: It seems like we may want to exit early if the provided operatorGroup does not share a name with the found operatorGroup. Currently this code will have hit an error attempting to update an operatorGroup that does not exist.

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -64,8 +65,51 @@ func (a *Operator) syncOperatorGroups(obj interface{}) error {
...
+				meta.RemoveStatusCondition(&op.Status.Conditions, v1.MutlipleOperatorGroupCondition)
+			}
+			_, err = a.client.OperatorsV1().OperatorGroups(op.GetNamespace()).UpdateStatus(context.TODO(), og, metav1.UpdateOptions{})
```
awgreene, line 82: This is the update I think we can skip if the provided operatorGroup's name does not match the name of the one found on cluster.

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -64,8 +65,51 @@ func (a *Operator) syncOperatorGroups(obj interface{}) error {
...
+			Status:  metav1.ConditionTrue,
+			Reason:  v1.MultipleOperatorGroupsReason,
+			Message: "Multiple OperatorGroup found in the same namespace",
+		}
+		for _, og := range groups {
+			if c := meta.FindStatusCondition(og.Status.Conditions, v1.MutlipleOperatorGroupCondition); c != nil {
+				continue
+			}
+			meta.SetStatusCondition(&og.Status.Conditions, cond)
+			if og.GetName() == op.GetName() {
```
awgreene, line 101: Sane as above.

awgreene (changes requested):
Nice work @dinhxuanvu - a few questions.

pkg/controller/operators/olm/operatorgroup.go
awgreene: Actually - this may not apply as there could be other changes that we want to apply.

pkg/controller/operators/olm/operatorgroup.go
awgreene: NVM - this seems to be fine.

awgreene (approved):
Took a second glance at this and it looks good.

PR #2309, opened 2021-07-30
don't return an error if the singleton rbac doesn't exist

ecordell:
Doesn't return an error from `ensureSingletonRBAC` if we can detect that the cache is just a bit behind

Fixes https://github.com/operator-framework/operator-lifecycle-manager/issues/2091

prior to this change, I could reproduce with:
```
$ go test ./pkg/controller/operators/olm/... -v -run ^TestSyncOperatorGroups$ -count 10
```
after this change, I couldn't reproduce with:
```
$ go test ./pkg/controller/operators/olm/... -v -run ^TestSyncOperatorGroups$ -count 100
```

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/olm/operatorgroup.go
+++ b/pkg/controller/operators/olm/operatorgroup.go
@@ -485,8 +485,14 @@ func (a *Operator) ensureSingletonRBAC(operatorNamespace string, csv *v1alpha1.C
 					Resources: []string{"namespaces"},
 				}),
 			}
-			if _, err := a.opClient.CreateClusterRole(clusterRole); err != nil {
-				return err
+			// TODO: this should do something smarter if the cluster role already exists
+			if cr, err := a.opClient.CreateClusterRole(clusterRole); err != nil {
+				// if the CR already exists, but the label is correct, the cache is just behind
+				if k8serrors.IsAlreadyExists(err) && ownerutil.IsOwnedByLabel(cr, csv) {
+					continue
+				} else {
+					return err
+				}
 			}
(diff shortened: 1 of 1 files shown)


pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -485,8 +485,14 @@ func (a *Operator) ensureSingletonRBAC(operatorNamespace string, csv *v1alpha1.C
...
-			if _, err := a.opClient.CreateClusterRole(clusterRole); err != nil {
-				return err
+			// TODO: this should do something smarter if the cluster role already exists
+			if cr, err := a.opClient.CreateClusterRole(clusterRole); err != nil {
+				// if the CR already exists, but the label is correct, the cache is just behind
+				if k8serrors.IsAlreadyExists(err) && ownerutil.IsOwnedByLabel(cr, csv) {
+					continue
+				} else {
+					return err
+				}
```
timflannagan, line 495: nit:

```suggestion
				if k8serrors.IsAlreadyExists(err) && ownerutil.IsOwnedByLabel(cr, csv) {
					continue
				}
				return err
```

timflannagan:
Nice :sunglasses: 
Holding for another opinion (if needed) but this looks reasonable.

/lgtm
/hold

dinhxuanvu (approved):
/lgtm

dinhxuanvu:
This PR failed one e2e test `creation with nodeSelector config`.

njhale (approved):
I think it's also time to resurrect the flake analyzer

PR #2310, opened 2021-07-30
Let net.Listen pick a port for grpc servers generated in tests

ecordell:
Let net.Listen pick the port for test grpc servers. This seems to avoid issues where the ports are not freed in time for subsequent tests.

This maybe fixes https://github.com/operator-framework/operator-lifecycle-manager/issues/2092, though I was unable to reproduce it locally (prior to the change in this PR, I had trouble running this test at all locally)

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/registry/grpc/source_test.go
+++ b/pkg/controller/registry/grpc/source_test.go
@@ -22,8 +22,8 @@ import (
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/registry"
 )
 
-func server(store opregistry.Query, port int) (func(), func()) {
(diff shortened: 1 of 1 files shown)


dinhxuanvu (approved):
/lgtm

dinhxuanvu:
Interesting. We have the `csv-role already exists` flake and now it seems with this PR we have `csv-rolebinding already exists` flake.

timflannagan:
That should've been fixed in the previous PR - I wonder if we're simply testing out-of-date content on a per-PR basis? Was this PR rebased against the previous one, or was that other flake PR merged in parallel while this one was opened?

PR #2312, opened 2021-08-03
Use GitVersion to parse version from discovery

kevinrizza:
Fix an issue with the new catalog templating controller where it was incorrectly parsing the version field from discovery. Instead, directly use the gitVersion value to match the behavior for minKubeVersion requirement parsing.

### DIFF
--- a/pkg/lib/catalogsource/image_template.go
+++ b/pkg/lib/catalogsource/image_template.go
@@ -7,8 +7,8 @@ import (
 	"strings"
 	"sync"
 
-	"github.com/blang/semver/v4"
 	"github.com/sirupsen/logrus"
+	versionutil "k8s.io/apimachinery/pkg/util/version"
 	"k8s.io/apimachinery/pkg/version"
 
 	"github.com/operator-framework/api/pkg/operators/v1alpha1"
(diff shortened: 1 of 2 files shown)


pkg/lib/catalogsource/image_template.go
```diff
@@ -171,14 +171,15 @@ func UpdateKubeVersion(serverVersion *version.Info, logger *logrus.Logger) {
...
 	}
 
-	templateNameToReplacementValuesMap[TemplKubeMajorV] = serverVersion.Major
-	templateNameToReplacementValuesMap[TemplKubeMinorV] = serverVersion.Minor
-
-	// api server does not explicitly give patch value, so we have to resort to parsing the git version
-	serverGitVersion, err := semver.Parse(serverVersion.GitVersion)
+	// need to use the gitversion from version.info.String() because minor version is not always Uint value
+	// and patch version is not returned as a first class field
+	semver, err := semver.Parse(strings.Split(strings.TrimPrefix(serverVersion.String(), "v"), "-")[0])
```
timflannagan, line 176: Just a quick thought - the apimachinery util package has [parsing functions available](https://pkg.go.dev/k8s.io/apimachinery@v0.21.3/pkg/util/version#ParseSemantic) that output a struct, and methods on that struct for retrieving major/minor/patch fields as uints, so you wouldn't need to also convert the string representation to a uint later.

pkg/lib/catalogsource/image_template.go
```diff
@@ -7,8 +7,8 @@ import (
 	"strings"
 	"sync"
 
-	"github.com/blang/semver/v4"
 	"github.com/sirupsen/logrus"
+	versionUtil "k8s.io/apimachinery/pkg/util/version"
```
ecordell, line 11: supernit: usually lowercase module names?

ecordell (approved):
/approve

we should consider using this for the minkubeversion as well, though I'm less concerned about updating that if it's going to be replaced with runtime constraints soon

PR #2313, opened 2021-08-03
Bug 1989712: Fix for resolution error on channels with deprecated inner entries.

benluddy:
Resolution would fail with an error in the presence of channels
containing more than two entries with a deprecated inner
entry (e.g. V1 -> V2 -> V3, with deprecated V2). This was due to
special handling that filtered out deprecated bundles at the cache
query layer. A channel's complete replaces-chain must be returned from
cache queries in order to correctly determine the relative order
between entries in the channel.

### DIFF
--- a/pkg/controller/registry/resolver/cache.go
+++ b/pkg/controller/registry/resolver/cache.go
@@ -392,9 +392,6 @@ type OperatorPredicate func(*Operator) bool
 func (s *CatalogSnapshot) Find(p ...OperatorPredicate) []*Operator {
 	s.m.RLock()
 	defer s.m.RUnlock()
-	if len(p) > 0 {
-		p = append(p, WithoutDeprecatedProperty())
-	}
 	return Filter(s.operators, p...)
 }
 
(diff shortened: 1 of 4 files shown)


pkg/controller/registry/resolver/cache.go
```diff
@@ -453,17 +450,6 @@ func WithPackage(pkg string) OperatorPredicate {
...
-	return func(o *Operator) bool {
-		for _, p := range o.bundle.GetProperties() {
-			if p.GetType() == opregistry.DeprecatedType {
-				return false
-			}
-		}
-		return true
-	}
-}
-
```
timflannagan, line 466: Why are we removing this code block? I might be missing something obvious, but I didn't see anything in the 4.8 PR for removing fields from this cache package. Just want to rule out whether this is an artifact of the cherry-pick failing and rebase being needed.

timflannagan: Link to that block of code in the 4.8 PR: https://github.com/operator-framework/operator-lifecycle-manager/pull/2104/files#diff-5311e3f964b627fa6618423cf9e1ce3f7c84530d3045b93365f1102cea3280b8R464.

ecordell: deprecated properties are handled by the solver instead: https://github.com/operator-framework/operator-lifecycle-manager/pull/2313/files/40187e79db7ab1b5fb4b66bcecce198a03f8313f#diff-2c070869d0e7b6bf96c69fbdcde88e430d5d8781e047c1d96cb22ee0967c0111R64-R72

ecordell:
[patch-manager] :rocket: Approved for z-stream by score: 0.50

picked

PR #2324, opened 2021-08-11
fix(openshift): drop z from next calculated y-stream

njhale:
When determining operator compatibility, drop z and build
versions in the calculation of the next Y-stream (minor)
release of OpenShift.

e.g. If the current version is v4.9.5+build, the next Y-stream is
calculated as v4.10.0.

If a pre-release is included, drop it as well but don't increment
the minor version.

e.g. If the current version is v4.10.0-rc, the next Y-stream is
calculated as v4.10.0.

Before this change, the next Y-stream was the result of simply
iterating the cluster's minor version. When the cluster was at
a patch version greater than that specified by an operator, upgrades
would be erroneously blocked.

e.g. If the current version was v4.9.5, the next version would be
calculated as v4.10.5, which would block upgrades on operators with
max versions set to v4.10 -- or more explicitly [v4.10.0, v4.10.5) (')'
is read "exclusive").

For further motivation, see https://bugzilla.redhat.com/show_bug.cgi?id=1993286

### DIFF
--- a/pkg/controller/operators/openshift/helpers.go
+++ b/pkg/controller/operators/openshift/helpers.go
@@ -125,16 +125,20 @@ func transientErrors(err error) error {
 }
 
 func incompatibleOperators(ctx context.Context, cli client.Client) (skews, error) {
-	next, err := desiredRelease(ctx, cli)
+	desired, err := desiredRelease(ctx, cli)
 	if err != nil {
 		return nil, err
 	}
 
-	if next == nil {
+	if desired == nil {
 		// Note: This shouldn't happen
-		return nil, fmt.Errorf("Failed to determine next OpenShift Y-stream release")
+		return nil, fmt.Errorf("Failed to determine current OpenShift Y-stream release")
 	}
-	next.Minor++
+
+	next, err := nextY(*desired)
+	if err != nil {
+		return nil, err
+	}
 
 	csvList := &operatorsv1alpha1.ClusterServiceVersionList{}
 	if err := cli.List(ctx, csvList); err != nil {
@@ -158,7 +162,7 @@ func incompatibleOperators(ctx context.Context, cli client.Client) (skews, error
 			continue
 		}
 
-		if max == nil || max.GTE(*next) {
+		if max == nil || max.GTE(next) {
 			continue
 		}
 		s.maxOpenShiftVersion = max.String()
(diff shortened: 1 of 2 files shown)


pkg/controller/operators/openshift/helpers.go
```diff
@@ -189,6 +193,22 @@ func desiredRelease(ctx context.Context, cli client.Client) (*semver.Version, er
...
+		// Dropping pre-releases is equivalent to incrementing Y
+		v.Pre = nil
+		v.Patch = 0
+
+		return v, nil
+	}
+
+	err := v.IncrementMinor() // Sets Y=Y+1 and Z=0
+
+	return v, err
```
timflannagan, line 209: nit/not blocking: any reason we need to capture the error as a variable here instead of just returning the value of v.IncrementMinor() explicitly here?

pkg/controller/operators/openshift/helpers_test.go
```diff
@@ -441,6 +441,70 @@ func TestIncompatibleOperators(t *testing.T) {
...
+		},
+		{
+			description: "ClusterPre",
+			cv: configv1.ClusterVersion{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "version",
+				},
+				Status: configv1.ClusterVersionStatus{
+					Desired: configv1.Update{
+						Version: "1.1.0-pre", // Next Y-stream is 1.1.0, NOT 1.2.0
```
timflannagan, line 481: Do we need a test for `<major>.<minor>.<patch>-<pre-release>+<build>` Would we ever need to specify that kind of format downstream?

timflannagan:
This looks reasonable to me. I'd like to get another opinion with someone more familiar with semver syntax but feel free to remove the hold if this is time sensitive.

/lgtm
/hold

pkg/controller/operators/openshift/helpers.go
```diff
@@ -189,6 +193,22 @@ func desiredRelease(ctx context.Context, cli client.Client) (*semver.Version, er
 	return &desired, nil
 }
 
+func nextY(v semver.Version) (semver.Version, error) {
+	v.Build = nil // Builds are irrelevant
+
+	if len(v.Pre) > 0 {
+		// Dropping pre-releases is equivalent to incrementing Y
+		v.Pre = nil
+		v.Patch = 0
```
estroz, line 202: Is it impossible to have `1.2.3-alpha.0`?

pkg/controller/operators/openshift/helpers.go
njhale: was just unsure about copy ordering in a return statement -- looks like it doesn't happen until all the expressions are evaluated: https://play.golang.org/p/tKT2N-UCMXz

pkg/controller/operators/openshift/helpers_test.go
njhale: I omitted it because build versions have no partial order in semver, but it probably best to not make any assumptions when testing -- I'll add another test case

PR #2326, opened 2021-08-12
clarify maxOpenShiftVersion upgrade error message

ankitathomas:
Signed-off-by: Ankita Thomas <ankithom@redhat.com>

### DIFF
--- a/pkg/controller/operators/openshift/clusteroperator_controller_test.go
+++ b/pkg/controller/operators/openshift/clusteroperator_controller_test.go
@@ -2,7 +2,6 @@ package openshift
 
 import (
 	"fmt"
-
 	semver "github.com/blang/semver/v4"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
@@ -213,6 +212,7 @@ var _ = Describe("ClusterOperator controller", func() {
 			}).Should(Succeed())
 		}()
 
+		parsedVersion := semver.MustParse(clusterVersion)
 		Eventually(func() ([]configv1.ClusterOperatorStatusCondition, error) {
 			err := k8sClient.Get(ctx, clusterOperatorName, co)
 			return co.Status.Conditions, err
@@ -224,7 +224,7 @@ var _ = Describe("ClusterOperator controller", func() {
 				{
 					namespace:           ns.GetName(),
 					name:                incompatible.GetName(),
-					maxOpenShiftVersion: clusterVersion,
+					maxOpenShiftVersion: fmt.Sprintf("%d.%d", parsedVersion.Major, parsedVersion.Minor),
 				},
 			}.String(),
 			LastTransitionTime: fixedNow(),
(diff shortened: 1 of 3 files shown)


pkg/controller/operators/openshift/helpers.go
```diff
@@ -110,7 +110,7 @@ func (s skew) String() string {
 		return fmt.Sprintf("%s/%s has invalid %s properties: %s", s.namespace, s.name, MaxOpenShiftVersionProperty, s.err)
 	}
 
-	return fmt.Sprintf("%s/%s is incompatible with OpenShift versions greater than %s", s.namespace, s.name, s.maxOpenShiftVersion)
+	return fmt.Sprintf("%s/%s is incompatible with OpenShift minor versions greater than %s", s.namespace, s.name, s.maxOpenShiftVersion)
```
njhale, line 113: I believe `skew.maxOpenShiftVersion` is the result of directly stringifying the value of the property, which **can** specify patch versions. I would take care to format the message so that only `major.minor` is displayed.

njhale (changes requested):
Thanks for putting this together!

I am expecting CI to fail without test changes though.

pkg/controller/operators/openshift/helpers.go
awgreene: Could the test include a maxOpenShift version that includes a patch and could we ensure that the patch version is omitted from the message?

awgreene:
@ankitathomas thanks for the quick turnaround on this, one suggestion on the test, looks good otherwise.

pkg/controller/operators/openshift/helpers.go
kevinrizza: Shouldn't we just throw a validation error on a patch version? It doesn't seem to me to make a lot of sense to allow users to even set it like that if we can't actually block those upgrades anyway.

dankenigsberg: How about requiring the operators to declare `maxOpenShiftVersion: 4.8.9999` if they support all future 4.8.* versions?

kevinrizza: I just think it's confusing and erroneous to do that. OLM functionally can't do anything special with the patch version here, so we should limit it to the minor version with validation.

ankitathomas: Modified it to drop the build version and complain on a non-zero patch/pre-release version. I've added a warning on the api validation https://github.com/operator-framework/api/pull/146

pkg/controller/operators/openshift/clusteroperator_controller_test.go
```diff
@@ -2,7 +2,6 @@ package openshift
 
 import (
 	"fmt"
-
```
njhale, line 5: nit: this newline should exist to delineate standard library imports.

njhale (approved):
/approve

dinhxuanvu (approved):
/lgtm

sdodson:
/retitle Bug 1992677: clarify maxOpenShiftVersion upgrade error message

PR #2330, opened 2021-08-18
Prune API of the resolver package.

benluddy:
The resolver package exports a number of identifiers that have not been used for resolution in over a year. As part of the ongoing effort to decouple purely-resolution-related concerns from the rest of OLM, some of these have been rehomed into the packages that still consume them, and entirely unused code has been removed.


test/e2e/subscription_e2e_test.go
```diff
@@ -219,7 +218,7 @@ var _ = Describe("Subscription", func() {
 		stableChannel := "stable"
 		mainCSV := newCSV(mainPackageStable, testNamespace, "", semver.MustParse("0.1.0-1556661347"), []apiextensions.CustomResourceDefinition{crd}, nil, nil)
 		updatedCSV := newCSV(updatedPackageStable, testNamespace, "", semver.MustParse("0.1.0-1556661832"), []apiextensions.CustomResourceDefinition{crd}, nil, nil)
-		updatedCSV.SetAnnotations(map[string]string{resolver.SkipPackageAnnotationKey: ">=0.1.0-1556661347 <0.1.0-1556661832"})
+		updatedCSV.SetAnnotations(map[string]string{"olm.skipRange": ">=0.1.0-1556661347 <0.1.0-1556661832"})
```
dinhxuanvu, line 221: I know this is a test so it doesn't matter but this const `SkipPackageAnnotationKey` still exists. May as well keeping using it instead of hardcoding.

pkg/controller/operators/catalog/operator.go
```diff
@@ -890,7 +890,7 @@ func (o *Operator) syncResolvingNamespace(obj interface{}) error {
 	// get the set of sources that should be used for resolution and best-effort get their connections working
 	logger.Debug("resolving sources")
 
-	querier := resolver.NewNamespaceSourceQuerier(o.sources.AsClients(o.namespace, namespace))
```
njhale, line 893: TIL: this wasn't (never?) actually dependency injected

even if it was, the DI should have taken a factory anyway

pkg/controller/operators/catalog/querier.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+func (q *NamespaceSourceQuerier) Queryable() error {
+	if len(q.sources) == 0 {
+		return fmt.Errorf("no catalog sources available")
+	}
+	return nil
+}
+
+func (q *NamespaceSourceQuerier) FindReplacement(currentVersion *semver.Version, bundleName, pkgName, channelName string, initialSource registry.CatalogKey) (*api.Bundle, *registry.CatalogKey, error) {
```
dinhxuanvu, line 52: I kinda wish we get rid of this as well. This func call is just to detect if there is a new version by making multiple API calls to registry to query the info. It doesn't really matter to the resolution which is run later if there is a change. Sometimes, `FindReplacement` will find something that will replace existing version and change the Subscriptions status to `UpgradeAvailable`. However, when the resolver runs, the replacement may not meet the constraints so no upgrade happens. That's a bit confusing. Plus, may as well let the resolver run because maybe it is less expensive (API-call-wise) that way. That part is unknown so it is fine to keep it as it is.

dinhxuanvu (commented):
Looking good. Just a tiny nit.
I can lgtm if it is good to go. Certainly don't want to hold this on a tiny nit.

njhale (approved):
At times like this I'm glad we haven't bumped the module to v1 yet :)

/lgtm

test/e2e/subscription_e2e_test.go
benluddy: I decided that the end-to-end test shouldn't depend on this constant from the implementation. If its value is changed, the test should fail and need to be updated since it represents a change in behavior.

pkg/controller/operators/catalog/querier.go
benluddy: I feel the same way, but I didn't want to take on that change right now.

dinhxuanvu (approved):
/lgtm

Issue #2322, opened 2021-08-10
`irifrance/gini` moved to `go-air/gini`


zregvart:
## Bug Report

**What did you do?**
I tried to build operator depending on `operator-lifecycle-manager`

**What did you expect to see?**
Successful build.

**What did you see instead? Under which circumstances?**
Build failed with missing dependency

```
go: github.com/operator-framework/operator-lifecycle-manager@v0.18.2 requires
	github.com/irifrance/gini@v1.0.1: reading github.com/irifrance/gini/go.mod at revision v1.0.1: unknown revision v1.0.1
ERROR: Last command exited with 1
```

**Environment**
* operator-lifecycle-manager version: 0.18.2

* Kubernetes version information: n/a

* Kubernetes cluster kind: n/a

**Possible Solution**
Change the dependency to `https://github.com/go-air/gini` in `pkg/controller/registry/resolver/solver/*`

zregvart:
Seems that with cached dependencies locally this is not reproducible so adding a Dockerfile to help reproduce:

```Dockerfile
FROM fedora:34

RUN dnf -y install golang
WORKDIR /tmp/test
RUN go mod init test.com/test
RUN go get github.com/operator-framework/operator-lifecycle-manager@v0.18.3
```

Fails with:
```
...
STEP 5: RUN go get github.com/operator-framework/operator-lifecycle-manager@v0.18.3
go: downloading github.com/operator-framework/operator-lifecycle-manager v0.18.3
go get: github.com/operator-framework/operator-lifecycle-manager@none updating to
	github.com/operator-framework/operator-lifecycle-manager@v0.18.3 requires
	github.com/irifrance/gini@v1.0.1: reading github.com/irifrance/gini/go.mod at revision v1.0.1: unknown revision v1.0.1
Error: error building at STEP "RUN go get github.com/operator-framework/operator-lifecycle-manager@v0.18.3": error while running runtime: exit status 1
```

PR #2331, opened 2021-08-18
Use of/gini instead of inifrance/gini

kevinrizza:
The github.com/irifrance/gini package has been purged from github. An existing version that preserved that history has been forked to the operator-framework org, and this commit updates that dependency to use the forked version.

In the future, we should investigate https://github.com/go-air/gini if and when it creates consumable releases, as it appears that the maintainer of the missing package is now maintaining the package there.

Resolves https://github.com/operator-framework/operator-lifecycle-manager/issues/2322


timflannagan:
Interesting:

```
--- FAIL: TestSyncOperatorGroups (1.45s)
    --- FAIL: TestSyncOperatorGroups/AllNamespaces/CSVPresent/InstallModeNotSupported (0.16s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x11577b9]
```

timflannagan:
It looks like there's some ordering that `go mod tidy` is complaining about that's failing the verification check. Likely just need to re-run that command and re-commit.

kevinrizza:
@timflannagan should be fixed now

PR #2337, opened 2021-08-26
Reduce resolver dependencies on operator-registry and operator-lifecycle-manager.

benluddy:
This is a continuation of the ongoing work to allow the runtime dependency resolution features of catalog-operator to be reusable outside of OLM.


dinhxuanvu (approved):
One tiny nit but it doesn't really matter.
/lgtm

PR #2340, opened 2021-09-03
Remove hardcoded references to the registry gRPC API from the resolver packages.

benluddy:
The resolver component does not care whether a given cache entry was
populated by use of the registry gRPC API, so it should ultimately be
possible to drop the "Bundle" field from cache entries altogether. For
now, catalog-operator's "step resolver" still depends on a Bundle
proto field in order to unpack inline (as opposed to image-based)
packages.

NewOperatorForBundle and NewOperatorFromV1Alpha1CSV couple the
resolver cache package to the registry gRPC API and the
operators/v1alpha1 CSV API, respectively. They're effectively used
only by the resolver package, so they can be moved there and
unexported. The remaining calls to NewOperatorFromV1Alpha1CSV can be
extracted from core resolver logic into a Source implementation that
is responsible for the conversion from ClusterServiceVersion to
cache.Snapshot entry, and the remaining call to NewOperatorFromBundle
is used only by the Source implementation based on the registry gRPC
API.


benluddy:

dinhxuanvu (approved):
/lgtm

PR #2346, opened 2021-09-08
Introduce a sanity workflow for checking static linting violations in CI

timflannagan:
Update the Makefile and introduce a `lint` target. This target is responsible for running both `gofmt` and `goimports` against non-vendor packages in the codebase.

Introduce the ci/sanity GH workflow:
- Runs `make vendor`
- Runs `make lint`
- Ensures that both of those targets don't produce a diff when running
  `make diff`

Fix any linting gofmt/goimports linting violations that came up when running the `make lint` target locally.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -184,6 +184,9 @@ clean:
 	@rm -rf test/e2e/log
 	@rm -rf e2e.namespace
 
+lint:
+	find . -name '*.go' -not -path "./vendor/*" -not -path "./pkg/lib/operatorclient/operatorclientmocks/*" | xargs gofmt -w
+	find . -name '*.go' -not -path "./vendor/*" -not -path "./pkg/lib/operatorclient/operatorclientmocks/*" | xargs goimports -w
 
 # Copy CRD manifests
 manifests: vendor
--- a/.github/workflows/sanity.yaml
+++ b/.github/workflows/sanity.yaml
@@ -1,0 +1,23 @@
+name: ci
+
+on:
+  push:
+    branches:
+      - '**'
+  pull_request:
+    paths:
+      - '**'
+      - '!doc/**'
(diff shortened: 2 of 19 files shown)


Makefile
```diff
@@ -184,6 +184,9 @@ clean:
 	@rm -rf test/e2e/log
 	@rm -rf e2e.namespace
 
+lint:
+	find . -name '*.go' -not -path "./vendor/*" | xargs gofmt -w
+	find . -name '*.go' -not -path "./vendor/*" | xargs goimports -w
```
timflannagan, line 189: This likely isn't the greatest structure but I didn't have an immediate answer on how to clean it up.

It looks like we do some combination of the following:
- Not run both of those checks
- Make the list of files spit out by that `find` command a prerequisite and update the xargs command to run both gofmt/goimports references that same set of files.

timflannagan:

Makefile
```diff
@@ -184,6 +184,9 @@ clean:
 	@rm -rf test/e2e/log
 	@rm -rf e2e.namespace
 
+lint:
+	find . -name '*.go' -not -path "./vendor/*" -not -path "./pkg/lib/operatorclient/operatorclientmocks/*" | xargs gofmt -w
```
timflannagan, line 188: Note: mentioned this in the commit, but that operatorclientmocks package doesn't have the `//go:generate ...` comment that it looks like gofmt/goimports are configured to ignore, so we'd need to explicitly ignore this path from the find command output.

dinhxuanvu (approved):
/lgtm

timflannagan:
Manually merging - that unit test is permafailing (https://github.com/operator-framework/operator-lifecycle-manager/issues/2091).

PR #2347, opened 2021-09-08
.github/workflows: Don't run CI checks for markdown-only updates

timflannagan:
Update the ci-prefixed workflows and ensure that CI isn't running when
only markdown files are updated.

**Reviewer Checklist**

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,7 +2,6 @@ name: ci
 on:
   pull_request:
     paths-ignore:
-    - 'doc/**/*.md'
     - '**/*.md'
 jobs:
   image:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,10 +1,8 @@
 name: ci
 on:
   pull_request:
-    paths:
-    - '**'
-    - '!doc/contributors/**'
-    - '!doc/design/**'
+    paths-ignore:
+    - 'doc/**/*.md'
     - '**/*.md'
 jobs:
   image:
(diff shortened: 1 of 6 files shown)


.github/workflows/build.yml
```diff
@@ -5,6 +5,7 @@ on:
     - '**'
     - '!doc/contributors/**'
     - '!doc/design/**'
+    - '**/*.md'
```
timflannagan, line 8: TODO: we should move this "!*" type paths as [path_ignore](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-ignoring-paths) items instead.

.github/workflows/build.yml
```diff
@@ -1,10 +1,9 @@
 name: ci
 on:
   pull_request:
-    paths:
-    - '**'
-    - '!doc/contributors/**'
-    - '!doc/design/**'
+    paths-ignore:
+    - 'doc/**'
```
njhale, line 5: `./doc/install/local-values.yaml` used to be used in manifest generation, not sure if that's still true though.

.github/workflows/e2e-tests.yml
```diff
@@ -4,9 +4,9 @@ on:
     branches:
       - master
   pull_request:
-    paths:
-      - '**'
-      - '!doc/**'
+    paths-ignore:
+    - 'doc/**'
+    - '**/*.md'
```
njhale, line 9: `./doc/install/local-values.yaml` used to be used in manifest generation, not sure if that's still true though.

timflannagan: Updated to use `'doc/**/*.md` after looking at the examples outlined in https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#patterns-to-match-file-paths.

njhale (approved):
/approve

timflannagan:

estroz (approved):
/lgtm

PR #2349, opened 2021-09-08
pkg/controller: Fix panic when creating cluster-scoped RBAC in OG controller

timflannagan:
Fixes [#2091](https://github.com/operator-framework/operator-lifecycle-manager/issues/2091).
    
This is a follow-up to [operator-framework#2309](operator-framework#2309) that attempted to fix the original issue. When checking whether the ClusterRole/ClusterRoleBinding resources already exist, we're also checking whether the existing labels are owned by the CSV we're currently handling. When accessing the "cr" or "crb" variables that the Create calls output, a panic is produced as we're attempting to access the meta.Labels key from those resources, except those resources themselves are nil. Update the check to verify that the cr/crb variables are not nil before attempting to access those object's labels. The testing fake client may need to be updated in the future to handle returning these resources properly.

I ran `go test ./pkg/controller/operators/olm/... -v -run ^TestSyncOperatorGroups$ -count 250` a couple of times and wasn't able to reproduce. I was able to consistently reproduce without these changes using a much smaller count sizing before.

Reduce testing flakes.

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/olm/operator_test.go
+++ b/pkg/controller/operators/olm/operator_test.go
@@ -34,7 +34,6 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	utilclock "k8s.io/apimachinery/pkg/util/clock"
-	"k8s.io/apimachinery/pkg/util/diff"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/apimachinery/pkg/util/wait"
@@ -3667,9 +3666,13 @@ func TestUpdates(t *testing.T) {
 }
 
 func TestSyncOperatorGroups(t *testing.T) {
-	logrus.SetLevel(logrus.DebugLevel)
+	logrus.SetLevel(logrus.WarnLevel)
 	clockFake := utilclock.NewFakeClock(time.Date(2006, time.January, 2, 15, 4, 5, 0, time.FixedZone("MST", -7*3600)))
 	now := metav1.NewTime(clockFake.Now().UTC())
+	const (
+		timeout = 5 * time.Second
+		tick    = 50 * time.Millisecond
+	)
 
 	operatorNamespace := "operator-ns"
 	targetNamespace := "target-ns"
@@ -4305,6 +4308,10 @@ func TestSyncOperatorGroups(t *testing.T) {
 			expectedEqual: true,
 			initial: initial{
 				operatorGroup: &v1.OperatorGroup{
+					TypeMeta: metav1.TypeMeta{
+						Kind:       v1.OperatorGroupKind,
+						APIVersion: v1.GroupVersion.String(),
+					},
 					ObjectMeta: metav1.ObjectMeta{
 						Name:      "operator-group-1",
 						Namespace: operatorNamespace,
@@ -4334,6 +4341,10 @@ func TestSyncOperatorGroups(t *testing.T) {
 			final: final{objects: map[string][]runtime.Object{
 				operatorNamespace: {
 					&v1.OperatorGroup{
+						TypeMeta: metav1.TypeMeta{
+							Kind:       v1.OperatorGroupKind,
+							APIVersion: v1.GroupVersion.String(),
+						},
 						ObjectMeta: metav1.ObjectMeta{
 							Name:      "operator-group-1",
 							Namespace: operatorNamespace,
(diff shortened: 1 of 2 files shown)


timflannagan:
Evan had pointed out that this implementation may be another regression from fixing the original issue in slack.

/hold

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -572,12 +571,11 @@ func (a *Operator) ensureSingletonRBAC(operatorNamespace string, csv *v1alpha1.C
 			}
 			// TODO: this should do something smarter if the cluster role binding already exists
 			if crb, err := a.opClient.CreateClusterRoleBinding(clusterRoleBinding); err != nil {
-				// if the CR already exists, but the label is correct, the cache is just behind
-				if k8serrors.IsAlreadyExists(err) && ownerutil.IsOwnedByLabel(crb, csv) {
+				// if the CRB already exists, but the label is correct, the cache is just behind
+				if k8serrors.IsAlreadyExists(err) && crb != nil && ownerutil.IsOwnedByLabel(crb, csv) {
```
timflannagan, line 575: I realize this is the wrong logic - we need to handle an object already existing on top of that object being nil.

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -572,12 +573,14 @@ func (a *Operator) ensureSingletonRBAC(operatorNamespace string, csv *v1alpha1.C
...
-					continue
-				} else {
+				if !k8serrors.IsAlreadyExists(err) {
 					return err
 				}
+				// if the CRB already exists, but the label is correct, the cache is just behind
+				if crb != nil && ownerutil.IsOwnedByLabel(crb, csv) {
+					continue
+				}
+				return err
```
estroz, line 583: Should this error be returned?

njhale: I think this needs to return an error if we want `!IsOwnedByLabel` to cause a resync (i.e. to retry this logic).

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -532,11 +532,12 @@ func (a *Operator) ensureSingletonRBAC(operatorNamespace string, csv *v1alpha1.C
...
+				if !k8serrors.IsAlreadyExists(err) {
+					return err
+				}
 				// if the CR already exists, but the label is correct, the cache is just behind
-				if k8serrors.IsAlreadyExists(err) && ownerutil.IsOwnedByLabel(cr, csv) {
+				if cr != nil && ownerutil.IsOwnedByLabel(cr, csv) {
 					continue
-				} else {
-					return err
 				}
```
njhale, line 541: It looks like the @ecordell 's [original patch](https://github.com/operator-framework/operator-lifecycle-manager/commit/538b1315eace85e33df7b2893a85a000f8757349), effectively returns _any_ error if `!IsOwnedByLabel`, which this code does not. Assuming errors returned from this function trigger a resync -- i.e. a retry -- then we'll fail to retry the operation when we detect the cluster state isn't as expected.

I suspect something like the following will produce the desired results:

```go
if k8serrors.IsAlreadyExists(err) && cr != nil && ownerutil.IsOwnedByLabel(cr, csv) {
    continue
}
return err
```

njhale (changes requested):

pkg/controller/operators/olm/operatorgroup.go
timflannagan: When I had played around with this exact implementation, we had reverted back to the original testing flake behavior (e.g. "crb-role already exists") as there's likely a larger issue in the fake testing client where the resource returned from the Create call is always going to be nil. With that said, it's likely we don't need to nil check the variable returned by Create as this is a testing client problem. Any opinions?

timflannagan: Moving some of this from slack:

I had tried adding a client-go testing reactor for returning an object for these situations:

```diff
+// WithOperatorGroupReactors returns a fakeK8sClientOption that configures a Clientset to return the ClusterRole
+// or ClusterRoleBinding object during a create call.
+// Note(tflannag): Fix for https://github.com/operator-framework/operator-lifecycle-manager/issues/2091.
+func WithOperatorGroupReactors(tb testing.TB) Option {
+       return func(c ClientsetDecorator) {
+               c.PrependReactor("create", "clusterrolebindings", func(action clitesting.Action) (bool, runtime.Object, error) {
+                       createAction := action.(clitesting.CreateActionImpl)
+                       return false, createAction.Object, nil
+               })
+               c.PrependReactor("create", "clusterroles", func(action clitesting.Action) (bool, runtime.Object, error) {
+                       createAction := action.(clitesting.CreateActionImpl)
+                       return false, createAction.Object, nil
+               })
+       }
+}
```

But it wasn't immediately clear how to best integrate those reactors into the current fake client setup such that we're not always appending these reactors and increasing the time it takes to run unit tests, which likely indicates it's more useful as a one-off functional type setup:

```diff
-       k8sClientFake := k8sfake.NewSimpleClientset(config.k8sObjs...)
+       k8sClientFake := fake.NewClientSetDecoratorWithReactors(config.k8sObjs, config.fakeClientOptions...)
        k8sClientFake.Resources = apiResourcesForObjects(append(config.extObjs, config.regObjs...))
-       config.operatorClient = operatorclient.NewClient(k8sClientFake, apiextensionsfake.NewSimpleClientset(config.extObjs...), apiregistrationfake.NewSimpleClientset(config.regObjs...))
+       config.operatorClient = operatorclient.NewClient(k8sClientFake.Clientset, apiextensionsfake.NewSimpleClientset(config.extObjs...), apiregistrationfake.NewSimpleClientset(config.regObjs...))
```

I tried running down this route but got sidetracked with other work.

njhale: > But it wasn't immediately clear how to best integrate those reactors into the current fake client setup such that we're not always appending these reactors and increasing the time it takes to run unit tests, which likely indicates it's more useful as a one-off functional type setup

imho this reactor is generally useful for all unit tests, but I'm not sure the error from the create action would be carried forward properly by the existing decorator even if the example you gave was added.

> When I had played around with this exact implementation, we had reverted back to the original testing flake behavior (e.g. "crb-role already exists")

this is expected in the implementation with a bad client fake, but it should no longer panic at test time, and _should_ do the correct thing at runtime; i.e. return an error -- even if it's not the "right" error -- when the cluster has a different view of the cluster.

pkg/controller/operators/olm/operator_test.go
```diff
@@ -3667,9 +3666,13 @@ func TestUpdates(t *testing.T) {
 }
 
 func TestSyncOperatorGroups(t *testing.T) {
-	logrus.SetLevel(logrus.DebugLevel)
+	logrus.SetLevel(logrus.WarnLevel)
 	clockFake := utilclock.NewFakeClock(time.Date(2006, time.January, 2, 15, 4, 5, 0, time.FixedZone("MST", -7*3600)))
 	now := metav1.NewTime(clockFake.Now().UTC())
+	const (
```
dinhxuanvu, line 3672: I thought we had this const somewhere in the code base already.

dinhxuanvu (approved):
/lgtm

PR #2355, opened 2021-09-15
.github/workflows: Update the e2e-kind/e2e-minikube workflow(s)

timflannagan:
Rename the e2e-kind and e2e-minikube workflows to communicate that these
tests are used to ensure that OLM doesn't regress when running new OLM
local installations on these cluster types. These workflows were
accidentally renamed when the jobs in the e2e-tests workflow were split
out to their own workflows to allow individual retesting


benluddy:
/hold

I have concerns about removing this (it's a test of the local development scripts, which have different behaviors on minikube, not a test of the runtime components), discussing in kubernetes #olm-dev now.

PR #2356, opened 2021-09-15
.github/workflows: Don't run the sanity jobs for markdown-only updates

timflannagan:
Update the sanity workflow and avoid running the defined jobs when only
markdown changes are present in the PR. This is a follow-up to #2347
which enabled this behavior with the other actions, and this sanity
action was introduced after that PR had already merged.

No need to run CI checks on markdown-only changes.

**Reviewer Checklist**


dinhxuanvu (approved):
/lgtm

PR #2357, opened 2021-09-15
pkg/package-server: Add null check when attempting to initialize an API server config

timflannagan:
Update the pkg/package-server/server/server.go package and add a null
check for the config variable returned when running the `Config` method
for the `PackageServerOptions` structure. That `Config` method doesn't
ensure that the variable returned is populated when the error returned
is nil.

When attempting to initialize a delegating authentication configuration,
and the environment isn't properly setup (e.g. no cluster is running),
and SIGINT is sent to that process, an error will be returned from the
poll operation due to a timeout being reached, yet only the lastApplyErr
will be returned, which is nil in that scenario.

Note: this appears to only be reproducible when $KUBECONFIG is set locally, but no cluster is running, so this wouldn't be all that present in the wild.

Properly handle signals during the package-server executable so a panic isn't produced.

**Reviewer Checklist**


pkg/package-server/server/server.go
```diff
@@ -184,6 +185,9 @@ func (o *PackageServerOptions) Run(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
+	if config == nil {
+		return fmt.Errorf("failed to construct an API server config")
+	}
```
timflannagan, line 190: Adding more context to the PR/commit message:

- When attempting to initialize the delegating authentication options in https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/package-server/server/server.go#L130-L142 and no cluster is present, and SIGINT is sent to that package-server process, a timeout error will be returned during that poll operation as that signal is caught.
- The lastApplyErr is returned in https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/package-server/server/server.go#L140 despite that error being nil
- Line 184 initializes a `config, err := o.Config(ctx) ` variables
- That error will be evaluated as null as the lastApplyErr is always being returned and an empty config object is returned
- Line 191 attempts to access a nested field in an empty object, which produces a panic

Likely can also handle this in the o.Config method and instead double-check we're returning the correct error, so that call sites of this method can be ensured that the config structure is populated when an error is not nil.

perdasilva: I was about to suggest the `o.Config` solution. Seems to be the right place to put it and the right contract to give the developers for this method.

awgreene:
@timflannagan if you want to rebase this so the merge commit I added is removed, feel free.

PR #2358, opened 2021-09-15
Fail gracefully when client-ca file is not set or doesn't exist

sjenning:
There is not validation enforcing that the `--client-ca` flag be set.  Even if there was, we are currently eating the error on `NewCertPoolStore` if, for example, the file does not exist.  This results in a nil deref farther down the code path.

The PR returns on error so we gracefully handle the situation.

@benluddy @timflannagan


timflannagan:
#2091

timflannagan:
Manually merging - that unit test is permafailing (#2091)

PR #2359, opened 2021-09-15
Rename resolver/cache.Operator to Entry.

benluddy:
This type is a data object representing a resolver cache entry. It
doesn't have a particular connection to operators other than residing
where it resides. This name change should clarify what, exactly, the
type represents.

### DIFF
--- a/pkg/controller/registry/resolver/cache/predicates.go
+++ b/pkg/controller/registry/resolver/cache/predicates.go
@@ -10,18 +10,18 @@ import (
 	opregistry "github.com/operator-framework/operator-registry/pkg/registry"
 )
 
-type OperatorPredicate interface {
-	Test(*Operator) bool
+type Predicate interface {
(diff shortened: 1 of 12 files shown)


pkg/controller/registry/resolver/cache/predicates.go
```diff
@@ -11,7 +11,7 @@ import (
 )
 
 type OperatorPredicate interface {
```
njhale, line 13: Might as well rename this `Predicate` while we're at it.

njhale (approved):
/lgtm

timflannagan:
Known issue with the e2e tests that seems to be propping up more recently. Feel free to just manually squash merge this but holding so the bot doesn't go crazy.

/hold

benluddy:
@timflannagan Does this need to be rebased in order to get the E2E check passing?

benluddy:

PR #2360, opened 2021-09-16
Add system-wide solver constraint injection

njhale:

- Add injection point for system-wide solver constraints
- Factor the content of the catalog-operator's `main()` function into an exported function in a non-main package

Make it easier for platform-specific builds of OLM to perturb dependency resolution before "arbitrary runtime constraints" are available.


pkg/controller/registry/resolver/installabletypes.go
```diff
@@ -55,32 +56,110 @@ func bundleId(bundle, channel string, catalog cache.SourceKey) solver.Identifier
...
 		}
+
+		return nil, nil
 	}
+}
+
+// AddSystemConstraintProviders adds providers to the list of providers used system-wide, across all solvers.
+func AddSystemConstraintProviders(providers ...ConstraintProvider) {
+	systemConstraintProviders.add(providers...)
+}
```
joelanford, line 149: Is this the extension point were callers can add custom providers (e.g. from an `init` function)?

It might be good to add an example in the PR summary of how a caller would add a custom provider.

njhale:
we decided to carry a downstream-only patch that no longer requires this injection point.

PR #2364, opened 2021-09-17
Introduce GRPC_PROXY EnvVar Support

awgreene:
Introduce the ability to specify a dial context for GRPC connections.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -36,6 +36,7 @@ require (
 	github.com/spf13/cobra v1.1.3
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.7.0
+	golang.org/x/net v0.0.0-20210520170846-37e1c6afe023
 	golang.org/x/time v0.0.0-20210723032227-1f47c861a9ac
 	google.golang.org/grpc v1.38.0
 	gopkg.in/yaml.v2 v2.4.0
--- a/pkg/controller/registry/grpc/source.go
+++ b/pkg/controller/registry/grpc/source.go
@@ -2,12 +2,17 @@ package grpc
 
 import (
 	"context"
+	"net"
+	"net/url"
+	"os"
 	"sync"
 	"time"
 
 	"github.com/operator-framework/operator-registry/pkg/client"
 
 	"github.com/sirupsen/logrus"
+	"golang.org/x/net/http/httpproxy"
+	"golang.org/x/net/proxy"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/connectivity"
 
@@ -100,10 +105,71 @@ func (s *SourceStore) Get(key registry.CatalogKey) *SourceConn {
 	return &source
 }
 
+func grpcProxyURL(addr string) (*url.URL, error) {
+	// Handle ip addresses
+	host, _, err := net.SplitHostPort(addr)
+	if err != nil {
+		return nil, err
+	}
+
+	url, err := url.Parse(host)
+	if err != nil {
+		return nil, err
+	}
+
+	// Hardcode fields required for proxy resolution
+	url.Host = addr
+	url.Scheme = "http"
+
+	// Override HTTPS_PROXY and HTTP_PROXY with GRPC_PROXY
+	proxyConfig := &httpproxy.Config{
+		HTTPProxy:  getGRPCProxyEnv(),
+		HTTPSProxy: getGRPCProxyEnv(),
+		NoProxy:    getEnvAny("NO_PROXY", "no_proxy"),
+		CGI:        os.Getenv("REQUEST_METHOD") != "",
+	}
+
+	// Check if a proxy should be used based on environment variables
+	return proxyConfig.ProxyFunc()(url)
+}
+
+func getGRPCProxyEnv() string {
+	return getEnvAny("GRPC_PROXY", "grpc_proxy")
+}
+
+func getEnvAny(names ...string) string {
+	for _, n := range names {
+		if val := os.Getenv(n); val != "" {
+			return val
+		}
+	}
+	return ""
+}
+
+func grpcConnection(address string) (*grpc.ClientConn, error) {
+	dialOptions := []grpc.DialOption{grpc.WithInsecure()}
+	proxyURL, err := grpcProxyURL(address)
+	if err != nil {
+		return nil, err
+	}
+
+	if proxyURL != nil {
+		dialOptions = append(dialOptions, grpc.WithContextDialer(func(ctx context.Context, addr string) (net.Conn, error) {
+			dialer, err := proxy.FromURL(proxyURL, &net.Dialer{})
+			if err != nil {
+				return nil, err
+			}
+			return dialer.Dial("tcp", addr)
+		}))
+	}
+
+	return grpc.Dial(address, dialOptions...)
+}
+
 func (s *SourceStore) Add(key registry.CatalogKey, address string) (*SourceConn, error) {
 	_ = s.Remove(key)
 
-	conn, err := grpc.Dial(address, grpc.WithInsecure())
+	conn, err := grpcConnection(address)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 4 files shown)


awgreene:
Please take a look @kevinrizza @estroz

pkg/controller/registry/grpc/source.go
```diff
@@ -100,10 +105,69 @@ func (s *SourceStore) Get(key registry.CatalogKey) *SourceConn {
...
+
+	// Hardcode fields required for proxy resolution
+	url.Host = addr
+	url.Scheme = "http"
+
+	proxyConfig := httpproxy.FromEnvironment()
+
+	// Override HTTPS_PROXY and HTTP_PROXY with GRPC_PROXY
+	proxyConfig.HTTPProxy = getGRPCProxyEnv()
+	proxyConfig.HTTPSProxy = getGRPCProxyEnv()
```
tylerslaton, line 128: Would it be better to define an `httpproxy.Config` struct here instead? We could optimize cycles with less lines using a struct like so:

```suggestion
	proxyConfig := httpproxy.Config {
		HTTPProxy:  getGRPCProxyEnv()
		HTTPSProxy: getGRPCProxyEnv()
	}
```

pkg/controller/registry/grpc/source.go
```diff
@@ -100,10 +105,69 @@ func (s *SourceStore) Get(key registry.CatalogKey) *SourceConn {
...
+
+	// Check if a proxy should be used based on Environment variables
+	return proxyConfig.ProxyFunc()(url)
+}
+
+func getGRPCProxyEnv() string {
+	return getEnvAny("GRPC_PROXY", "grpc_proxy")
+}
+
+func getEnvAny(names ...string) string {
```
tylerslaton, line 138: Nit: Could this live in a utilities package somewhere? Seems like a useful function to be able to import in other places and reuse.

tylerslaton (commented):
Couple of questions! Looks good.

pkg/controller/registry/grpc/source.go
awgreene: We need the NO_PROXY variable as well and I thought it was easier to use a defined function than to recreate the code. But I'll go ahead and change it since it'll be easier to read.

awgreene: Addressed.

pkg/controller/registry/grpc/source.go
awgreene: Dave Cheney had a nice writeup on utility packages: https://dave.cheney.net/2019/01/08/avoid-package-names-like-base-util-or-common

To your point, this could be exported into a `environment` package, but I didn't see the value add since no other piece uses this yet.

tylerslaton: Ah, good point. I more meant making an `environment` package as opposed to a `util` or `common` package like you mentioned but I agree that the value add is low. Resolving.

pkg/controller/registry/grpc/source.go
```diff
@@ -100,10 +105,70 @@ func (s *SourceStore) Get(key registry.CatalogKey) *SourceConn {
...
+	url.Host = addr
+	url.Scheme = "http"
+
+	// Override HTTPS_PROXY and HTTP_PROXY with GRPC_PROXY
+	proxyConfig := &httpproxy.Config{
+		HTTPProxy:  getGRPCProxyEnv(),
+		HTTPSProxy: getGRPCProxyEnv(),
+		NoProxy:    getEnvAny("NO_PROXY", "no_proxy"),
+		CGI:        os.Getenv("REQUEST_METHOD") != "",
+	}
```
njhale, line 130: Is there a way for us to accomplish this _without_ exposing yet-another-knob ™️ to upstream OLM? e.g. downstream-only sidecar container.

Just seems like a lot of networking can be handled at a higher abstraction level (container/pod network).

njhale (commented):
design questions

pkg/controller/registry/grpc/source.go
awgreene: Ideally this would have been accomplished with the `HTTP_PROXY` environment variable, which grpc-go already supports. Unfortunately, that "knob" is already dedicated to our disconnected environment strategy in which all off cluster requests are routed through a proxy. Connecting to the registry pods through that proxy introduces changed behavior on those environments.

njhale: > Ideally this would have been accomplished with the HTTP_PROXY environment variable, which grpc-go already supports. 

to be clear, I don't think the way we've implemented HTTP proxy support is "at the right level" either.

njhale: So, my point here is that I don't think the clients here need to know what the proxies are if we have the ability to control networking at other scopes.

Maybe I'm misunderstanding the use case though.

awgreene: There are solutions that exist that could be used to route all traffic from the pod (istio and linkerd) but each of these solutions require the presence of a control plane to facilitate the routing. Short of introducing and supporting a new project that focuses on lightweight network routing, there does not seem to be a tool available that we could use to route GRPC connections.

go.mod
```diff
@@ -36,6 +36,7 @@ require (
 	github.com/spf13/cobra v1.1.3
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.7.0
+	golang.org/x/net v0.0.0-20210520170846-37e1c6afe023
```
timflannagan, line 39: IIRC, there's a medium CVE attached to x/net and we should double-check whether we're introducing an affected version.

awgreene: Is [this](https://www.cvedetails.com/cve/CVE-2021-33194/) the CVE you heard about?

timflannagan: Yeah, that's the one I was thinking about - it looks like we're in the clear.

awgreene: Seems fine:
```
golang.org/x/net before v0.0.0-20210520170846-37e1c6afe023 allows attackers to cause a denial of service (infinite loop) via crafted ParseFragment input.
```

pkg/controller/registry/grpc/source.go
njhale: fair enough.

njhale:
/hold

looks like some related unit tests are failing.

pkg/controller/registry/grpc/source.go
```diff
@@ -100,10 +105,71 @@ func (s *SourceStore) Get(key registry.CatalogKey) *SourceConn {
...
+
+	return grpc.Dial(address, dialOptions...)
+}
+
 func (s *SourceStore) Add(key registry.CatalogKey, address string) (*SourceConn, error) {
 	_ = s.Remove(key)
 
-	conn, err := grpc.Dial(address, grpc.WithInsecure())
+	conn, err := grpcConnection(address)
```
njhale, line 172: Can we abstract the dial logic away? maybe take a `Dailer` as a field of `SourceStore`?

pkg/controller/registry/grpc/source.go
```diff
@@ -100,10 +105,71 @@ func (s *SourceStore) Get(key registry.CatalogKey) *SourceConn {
...
+	// Hardcode fields required for proxy resolution
+	url.Host = addr
+	url.Scheme = "http"
+
+	// Override HTTPS_PROXY and HTTP_PROXY with GRPC_PROXY
+	proxyConfig := &httpproxy.Config{
+		HTTPProxy:  getGRPCProxyEnv(),
+		HTTPSProxy: getGRPCProxyEnv(),
+		NoProxy:    getEnvAny("NO_PROXY", "no_proxy"),
+		CGI:        os.Getenv("REQUEST_METHOD") != "",
```
njhale, line 129: We should probably memoize these so we don't fetch them for every dial.

njhale (commented):
sorry for the late review here, but spotted a few things of note:

njhale:
/lgtm
/hold cancel

looks like the unit test may have been on a stale commit

pkg/controller/registry/grpc/source.go
awgreene: If we pursue this option wouldn't the connection to the proxy need to remain open indefinitely?

pkg/controller/registry/grpc/source.go
awgreene: This seems reasonable but would make the values static, which likely isn't a problem given how deployments and pods work.

PR #2365, opened 2021-09-17
Integrate goreleaser to produce draft releases when new tags are pushed

### LINKED ISSUE #2118: streamline upstream release process
## Feature Request

A clear and concise description of what the problem is. Example: "I have an issue when (...)"
The current release process defined in https://github.com/operator-framework/operator-lifecycle-manager/blob/master/doc/design/release.md is manual and error-prone. It would be better to have a more consistent release cycle with less manual steps involved. 

Maybe using a workflow with goreleaser like in https://github.com/operator-framework/operator-sdk/blob/master/.github/workflows/deploy.yml could be the right step forward. 

We can cut more upstream releases faster and in a more automated way

timflannagan:
Integrate goreleaser to the OLM repository for building multi-arch
binaries and manifestlist container images, and help automate the
release process for OLM.

Add a GHA release workflow that's triggered on tags that's responsible
for building and pushing multi-arch (i.e. manifestlist) OLM container
images using goreleaser.

Goreleaser will also create a draft release, and generate a changelog
since the previous tag.

Rendering the release quickstart manifests is done after goreleaser has
created this draft release as there's no easy way to hook this
functionality into goreleaser after the docker images/manifestlists have
been pushed but before release artifacts are generated.

Use the 'softprops/action-gh-release' to update the newly created draft
release with these quickstart manifests as assets.

Update the release target and avoid hardcoding the
quay.io/operator-framework/olm quay repository and instead use the
existing IMAGE_REPO variable defined further up in the Makefile.

Update how the $(ver) variable is handled and avoid hardcoding the 'v'
prefix when overriding that 'ver' variable in the environment.

Alleviate the need for a multi-stage release process.

**Reviewer Checklist**

### DIFF
--- a/.github/workflows/goreleaser.yaml
+++ b/.github/workflows/goreleaser.yaml
@@ -49,7 +49,7 @@ jobs:
         uses: goreleaser/goreleaser-action@v2
         if: startsWith(github.ref, 'refs/tags')
         with:
-          version: latest
+          version: 0.177.0
           args: release --rm-dist
         env:
           GITHUB_TOKEN: ${{ github.token }}
--- a/.github/workflows/goreleaser.yaml
+++ b/.github/workflows/goreleaser.yaml
@@ -1,0 +1,73 @@
+name: release
+on:
+  pull_request:
+  push:
+    tags:
+      - 'v*'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Set up Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.16
+
+      - name: Get the image tag
+        if: startsWith(github.ref, 'refs/tags')
+        run: |
+          # Source: https://github.community/t/how-to-get-just-the-tag-name/16241/32
+          if [[ $GITHUB_REF == refs/tags/* ]]; then
+            echo IMAGE_TAG="${GITHUB_REF#refs/tags/}" >> $GITHUB_ENV
+          fi
+
+      - name: Create a draft release
+        uses: actions/create-release@v1
+        id: release
+        if: startsWith(github.ref, 'refs/tags')
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
+        with:
+          draft: true
+          tag_name: ${{ github.ref }}
+          release_name: ${{ github.ref }}
+
+      - name: Docker Login
+        uses: docker/login-action@v1
+        if: startsWith(github.ref, 'refs/tags')
+        with:
+          registry: quay.io
+          username: ${{ secrets.QUAY_USERNAME }}
+          password: ${{ secrets.QUAY_PASSWORD }}
+
+      - name: Run GoReleaser
+        uses: goreleaser/goreleaser-action@v2
+        if: startsWith(github.ref, 'refs/tags')
+        with:
+          version: latest
+          args: release --rm-dist
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
+          IMAGE_REPO: ${{ secrets.QUAY_USERNAME }}/olm
+          PKG: github.com/operator-framework/operator-lifecycle-manager
+
+      - name: Generate quickstart release manifests
+        if: startsWith(github.ref, 'refs/tags')
+        run: make release ver=${{ env.IMAGE_TAG }} IMAGE_REPO=quay.io/${{ secrets.QUAY_USERNAME }}/olm
+
+      - name: Update release artifacts with rendered Kubernetes manifests
+        uses: softprops/action-gh-release@v1
+        if: startsWith(github.ref, 'refs/tags')
+        with:
+          name: ${{ env.IMAGE_TAG }}
+          files: |
+            deploy/upstream/quickstart/crds.yaml
+            deploy/upstream/quickstart/olm.yaml
+            deploy/upstream/quickstart/install.sh
+          draft: true
+          token: ${{ github.token }}
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -1,0 +1,106 @@
+env:
+- GO111MODULE=on
+- CGO_ENABLED=0
+before:
+  hooks:
+    - go mod tidy
+    - go mod vendor
+builds:
+  - id: olm
+    main: ./cmd/olm
+    binary: olm
+    goos:
+    - linux
+    goarch:
+    - amd64
+    - arm64
+    tags:
+    - json1
+    flags:
+    - -mod=vendor
+    ldflags:
+    - -X {{ .Env.PKG }}/pkg/version.GitCommit={{ .FullCommit }}
+    - -X {{ .Env.PKG }}/pkg/version.OLMVersion={{ .Tag }}
+  - id: catalog
+    main: ./cmd/catalog
+    binary: catalog
+    goos:
+    - linux
+    goarch:
+    - amd64
+    - arm64
+    tags:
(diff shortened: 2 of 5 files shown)


timflannagan:
This still needs a couple of prerequisites added to the repository.

/hold

.goreleaser.yml
```diff
@@ -0,0 +1,113 @@
...
+  image_templates:
+  - quay.io/{{ .Env.IMAGE_REPO }}:{{ .Tag }}-amd64
+  - quay.io/{{ .Env.IMAGE_REPO }}:{{ .Tag }}-arm64
+
+- name_template: quay.io/{{ .Env.IMAGE_REPO }}:latest
+  image_templates:
+  - quay.io/{{ .Env.IMAGE_REPO }}:{{ .Tag }}-amd64
+  - quay.io/{{ .Env.IMAGE_REPO }}:{{ .Tag }}-arm64
+  create_flags:
+  - --amend
```
timflannagan, line 101: This produces manifestlist images for v.<minor>.<patch>, v.<minor>, and latest which seems like the right behavior since OLM isn't v1 and thus a major-only tag seems strange.

Something to note, is that docker images are pushed with amd64/arm64 suffixes using a v.<minor>.<patch> format (e.g. whatever tag triggered this action) so I'm not sure whether we're comfortable with that format.

.github/workflows/goreleaser.yaml
```diff
@@ -0,0 +1,56 @@
...
+        uses: docker/login-action@v1
+        with:
+          registry: quay.io
+          username: ${{ secrets.QUAY_USERNAME }}
+          password: ${{ secrets.QUAY_PASSWORD }}
+
+      - name: Run GoReleaser
+        uses: goreleaser/goreleaser-action@v2
+        with:
+          version: latest
```
joelanford, line 38: I'd suggest pinning this. Its always a bummer when you push a tag and then the release fails to build/push artifacts because the tool version broke your config.

Also, more nitpicky: it can be nice to have this setup as a make target so that it can be executed locally in case something needs to be reproduced or debugged.

.github/workflows/goreleaser.yaml
```diff
@@ -0,0 +1,56 @@
...
+
+      - name: Update release artifacts with rendered Kubernetes manifests
+        uses: softprops/action-gh-release@v1
+        with:
+          name: ${{ env.IMAGE_TAG }}
+          files: |
+            deploy/upstream/quickstart/crds.yaml
+            deploy/upstream/quickstart/olm.yaml
+          draft: true
+          token: ${{ secrets.GORELEASER_GITHUB_TOKEN }}
```
joelanford, line 56: GoReleaser has some ability to create GH releases and upload artifacts there. You may be able to move the release manifest generation above (or maybe included in) the goreleaser configuration and then have goreleaser handle creating the release and uploading artifacts.

Side note: only reason we're not doing this in operator-registry is because GoReleaser can't build a macOS binary alongside linux binaries without a cross-compiler running, and Apple doesn't permit that in general.

.goreleaser.yml
joelanford: We have to be careful with the way `latest` works here. If we push tag v0.21.0 and then later push tag v0.20.1, `latest` will point to v0.20.1 when we probably want it to stay on v0.21.0.

operator-registry solves this with some clever envvar value empty string shenanigans.

.github/workflows/goreleaser.yaml
timflannagan: Unfortunately, I've already tried rendering these release manifests before running the goreleaser action, but it dirtied the local environment (which I assume maps to `git diff` produces output) and failed the release process. I might be able to instead migrate towards generating these manifests, uploading using the archiving action, restore the git environment, and someone attach these files using the `extra_files` stanza, but it's not immediately clear how that'd work if those files are achived and not present locally. Maybe a .gitignore could work?

Another implementation that I played around with trying to global after hooks but that's a pro feature and there wasn't a step between building and pushing the manifestlist images and building out the release.

timflannagan: Maybe the cleaner option would be to create a draft release first, then run goreleaser, then upload any post-goreleaser artifacts that need to be attached so it's clear we're creating the release outside of goreleaser (and thus increase readability)? That's assuming goreleaser has support for working with existing draft releases, which playing around in my own fork seems to be a safe assumption.

.github/workflows/goreleaser.yaml
```diff
@@ -0,0 +1,56 @@
...
+          if [[ $GITHUB_REF == refs/tags/* ]]; then
+            echo IMAGE_TAG="${GITHUB_REF#refs/tags/}" >> $GITHUB_ENV
+          fi
+
+      - name: Docker Login
+        uses: docker/login-action@v1
+        with:
+          registry: quay.io
+          username: ${{ secrets.QUAY_USERNAME }}
+          password: ${{ secrets.QUAY_PASSWORD }}
```
joelanford, line 33: This will likely fail on PRs since PRs do not have access to repo secrets.
Probably need to add an `if` expression to only run on tag pushes.

.github/workflows/goreleaser.yaml
joelanford: Yeah perhaps separating them is cleaner in this case. It seems like getting things _just so_ for goreleaser would just add complexity and obfuscation for no real gain.

.github/workflows/goreleaser.yaml
timflannagan: The reason I strayed away from housing a Makefile target for this is because I wanted to avoid also setting up qemu-static and buildx actions but I realize that this goreleaser-action likely isn't handling that under-the-covers either, or this ubuntu VM already has all of that configured?

.github/workflows/goreleaser.yaml
```diff
@@ -0,0 +1,56 @@
...
+
+      - name: Run GoReleaser
+        uses: goreleaser/goreleaser-action@v2
+        with:
+          version: latest
+          args: release --rm-dist
```
joelanford, line 39: We probably need to add the `--snapshot` flag anytime we're not pushing images.

FYI - opm also pushes a `master` image on every push to master for folks that want to live on the edge :) It can be nice to have that, but definitely non-blocking.

.github/workflows/goreleaser.yaml
timflannagan: TIL about `if: startsWith(github.ref, 'refs/tags')` which seems like it should do the trick.

.goreleaser.yml
timflannagan: I think I'm leaning towards omitting the handlement around latest images entirely for the time being and trying to tackle that in a follow-up while we still work out what we'd like to see for tag structure in this repository. WDYT?

timflannagan:
I think this largely is read to go.

/hold cancel

.github/workflows/goreleaser.yaml
timflannagan: Ended up adding an action that's responsible for creating a draft release and it looks like goreleaser plays well with updating an existing draft release. So the overall workflow would be create an empty draft releaser -> run goreleaser to produce multi-arch container images and manifestlist images, and a generated changelog -> add post-goreleaser releaser artifacts by generating the quickstart manifests and using an action to attach those artifacts to this existing draft release.

The first step obviously isn't needed, but I think it improves the readability of the workflow file quite a bit as goreleaser abstracts quite a bit, and removes the multi-stage release process and removes the need for housing rendered YAML release manifests entirely. This means the new release process is maintainer creates a new tag locally and pushes that upstream -> the release action runs and creates a draft release -> maintainer publishes that draft release.

PR #2369, opened 2021-09-21
Remove oudated subscription update logic to improve resolution delay

dinhxuanvu:
Currently, olm logic checks for upgrade in subscription via another
obsolete API that is no longer in use for dependency solution. As a
result, sometimes, subscriptions display `UpgradeAvailable` status but
there will be no upgrades as the upgrade is not valid in the resolver.
Also, the `UpgradeAvailable` status is used to trigger the new resolution
even though that status is no longer a valid indicator of having a pending
upgrade. This leads to unwanted upgrade delay when the obsolete API works
properly.

This commit will remove the code that is using this obsolete API and
allow the resolution to happen when there is a subscription change.

**Reviewer Checklist**


dinhxuanvu:
@kevinrizza Need a re-lgtm if you don't mind. Thanks.

pkg/controller/operators/catalog/operator.go
```diff
@@ -1112,14 +1107,7 @@ func (o *Operator) ensureSubscriptionCSVState(logger *logrus.Entry, sub *v1alpha
 		if err := querier.Queryable(); err != nil {
 			return nil, false, err
 		}
```
benluddy, line 1109: This can be removed, too, right?

nunnatsa:
@dinhxuanvu - HCO upgrade tests start failing in prow after merging this PR. The test subscribes to a previous version and then performs an upgrade to the PR version (the version under test). Could you please check?

dinhxuanvu:
@nunnatsa Can you create a thread on Slack and tag me on it? I need some further logging and info to identify what the issue may be.

nunnatsa:
> @nunnatsa Can you create a thread on Slack and tag me on it? I need some further logging and info to identify what the issue may be.

There are no relevant logs from OLM, because OLM just ignores the change in the subscription.

PR #2371, opened 2021-09-23
Update default catalog namespace

kevinrizza:
default global catalog namespace shouldn't be vendor specific


timflannagan:
holding so the bot doesn't go crazy - hopefully have a fix for that problematic unit test.

/hold

timflannagan:
Closing/re-opening as the re-testing the unit test was checking out the wrong master branch commit.

/hold cancel

PR #2374, opened 2021-09-24
test/e2e: Avoid cleaning up kubeconfig tempdir until deprovision

timflannagan:
Update the kind provisioner and avoid cleaning up the tempdir containing
the generated kubeconfig until the process of deprovisioning the cluster
is encountered. This is largely to help increase the quality of life
trying to debug local installations while e2e is still running. Before,
the kubeconfig would be deleted before e2e tests were running.

**Reviewer Checklist**

### DIFF
--- a/test/e2e/ctx/provisioner_kind.go
+++ b/test/e2e/ctx/provisioner_kind.go
@@ -80,7 +80,6 @@ func Provision(ctx *TestContext) (func(), error) {
 	if err != nil {
 		return nil, fmt.Errorf("failed to create temporary directory: %s", err.Error())
 	}
-	defer os.RemoveAll(dir)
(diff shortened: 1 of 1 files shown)


benluddy:
I don't have a strong opinion on this, but my initial thought is that this temp directory is an implementation detail of the provisioner and shouldn't matter. Since these are kind clusters, have you tried `kind get clusters` and `kind export kubeconfig --name ...`? I believe the kind CLI identifies cluster node containers by label and extracts their kubeconfig directly, so it shouldn't be necessary to keep anything extra around on the host.

timflannagan:
Nope, you're right, that's just my unfamiliarity with kind leaking out. Exporting the kubeconfig by name works.

PR #2380, opened 2021-09-29
Remove NamespaceSourceQuerier API that is no longer in use

dinhxuanvu:
NamespaceSourceQuerier API is no longer in use for new resolver so it
should be removed to reduce tech debt and unusable code.

**Reviewer Checklist**


pkg/controller/operators/catalogtemplate/operator.go
```diff
@@ -130,6 +130,11 @@ func (o *Operator) syncCatalogSources(obj interface{}) error {
 		return nil
 	}
 
+	// This catalog template feature is not usable in configmap/internal catalogsource
+	if inputCatalogSource.Spec.SourceType == v1alpha1.SourceTypeInternal || inputCatalogSource.Spec.SourceType == v1alpha1.SourceTypeConfigmap {
+		return nil
+	}
```
timflannagan, line 136: How do you feel about splitting these changes into their own PR - namely in the context of record keeping downstream?

dinhxuanvu: I'm fine with that. Will do so after I see the result from CI e2e-tests. I want to take this opportunity to see how it is doing.

timflannagan:
@dinhxuanvu Can you rebase?

PR #2387, opened 2021-10-01
fix: wrap csv e2e test creation artifacts in Eventually statements 

exdx:

Noticed a flake locally 
```
• Failure in Spec Setup (BeforeEach) [60.056 seconds]
ClusterServiceVersion
/Users/dsover/code/src/operator-lifecycle-manager/test/e2e/csv_e2e_test.go:39
  when a CustomResourceDefinition was installed alongside a ClusterServiceVersion
  /Users/dsover/code/src/operator-lifecycle-manager/test/e2e/csv_e2e_test.go:60
    can satisfy an unassociated ClusterServiceVersion's ownership requirement if replaced by an associated ClusterServiceVersion [BeforeEach]
    /Users/dsover/code/src/operator-lifecycle-manager/test/e2e/csv_e2e_test.go:180

        <*errors.StatusError | 0xc0028c57c0>: {
            ErrStatus: {
                TypeMeta: {Kind: "", APIVersion: ""},
                ListMeta: {
                    SelfLink: "",
                    ResourceVersion: "",
                    Continue: "",
                    RemainingItemCount: nil,
                },
                Status: "Failure",
                Message: "object is being deleted: namespaces \"test-namespace-1\" already exists",
                Reason: "AlreadyExists",
                Details: {
                    Name: "test-namespace-1",
                    Group: "",
                    Kind: "namespaces",
                    UID: "",
                    Causes: nil,
                    RetryAfterSeconds: 0,
                },
                Code: 409,
            },
        }
        object is being deleted: namespaces "test-namespace-1" already exists

    /Users/dsover/code/src/operator-lifecycle-manager/test/e2e/csv_e2e_test.go:72
```

Reduce flakiness 

**Reviewer Checklist**


test/e2e/csv_e2e_test.go
```diff
@@ -69,7 +69,7 @@ var _ = Describe("ClusterServiceVersion", func() {
 					Name: "test-namespace-1",
 				},
 			}
-			Expect(ctx.Ctx().Client().Create(context.Background(), &ns)).To(Succeed())
+			Eventually(Expect(ctx.Ctx().Client().Create(context.Background(), &ns)).To(Succeed()))
```
njhale, line 72: I'm not sure how nested test specs like this behave.

You can avoid nesting by doing something like this:

```go
bctx, cancel := context.WithCancel(context.Background()) // Use a shared context to reduce boilerplate
defer cancel() // Be sure to kill any long-lived goroutines after we finish

cli := ctx.Ctx().Client()
Eventually(func() error {
     return ctx.Ctx().Client().Create(bctx, &ns)
}).Should(Succeed())
```

test/e2e/csv_e2e_test.go
```diff
@@ -80,7 +80,7 @@ var _ = Describe("ClusterServiceVersion", func() {
 					TargetNamespaces: []string{ns.GetName()},
 				},
 			}
-			Expect(ctx.Ctx().Client().Create(context.TODO(), &og)).To(Succeed())
+			Eventually(Expect(ctx.Ctx().Client().Create(context.TODO(), &og)).To(Succeed()))
```
njhale, line 83: we should probably switch to `context.Background()` in these tests (check out the godocs).

njhale (changes requested):

PR #2392, opened 2021-10-04
test/e2e: wrap subscription e2e test logic in Eventually statement

exdx:

Fixes up the "with starting CSV" test by wrapping a flake-prone path with an Eventually()

Flaky test

**Reviewer Checklist**

### DIFF
--- a/test/e2e/subscription_e2e_test.go
+++ b/test/e2e/subscription_e2e_test.go
@@ -467,9 +467,11 @@ var _ = Describe("Subscription", func() {
 		require.NoError(GinkgoT(), err)
 
 		// Wait for the subscription to begin upgrading to csvB
-		subscription, err = fetchSubscription(crc, testNamespace, subscriptionName, subscriptionStateUpgradePendingChecker)
-		require.NoError(GinkgoT(), err)
-		require.NotEqual(GinkgoT(), fetchedInstallPlan.GetName(), subscription.Status.InstallPlanRef.Name, "expected new installplan for upgraded csv")
+		// The upgrade changes the installplanref on the subscription
+		Eventually(func() (bool, error) {
+			subscription, err = crc.OperatorsV1alpha1().Subscriptions(testNamespace).Get(context.Background(), subscriptionName, metav1.GetOptions{})
+			return subscription != nil && subscription.Status.InstallPlanRef.Name != fetchedInstallPlan.GetName() && subscription.Status.State == operatorsv1alpha1.SubscriptionStateUpgradePending, err
+		}, 5*time.Minute, 1*time.Second).Should(BeTrue(), "expected new installplan for upgraded csv")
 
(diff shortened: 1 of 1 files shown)


test/e2e/subscription_e2e_test.go
```diff
@@ -467,9 +467,14 @@ var _ = Describe("Subscription", func() {
 		require.NoError(GinkgoT(), err)
 
 		// Wait for the subscription to begin upgrading to csvB
-		subscription, err = fetchSubscription(crc, testNamespace, subscriptionName, subscriptionStateUpgradePendingChecker)
```
timflannagan, line 470: I wonder if this should be handled at the fetchSubscription layer, and instead, create a new checker type.

/lgtm

test/e2e/subscription_e2e_test.go
```diff
@@ -468,8 +468,18 @@ var _ = Describe("Subscription", func() {
...
+			fetchedSubscription, err := crc.OperatorsV1alpha1().Subscriptions(testNamespace).Get(context.Background(), subscriptionName, metav1.GetOptions{})
+			if err != nil || fetchedSubscription == nil {
+				return false, err
+			}
+			if fetchedInstallPlan.GetName() == fetchedSubscription.Status.InstallPlanRef.Name {
+				return false, err
+			}
+			subscription = fetchedSubscription
+			return true, nil
+		}).Should(BeTrue(), "expected new installplan for upgraded csv")
```
njhale, line 482: Can we condense this and the previous expectation into a single `Eventually`? e.g.

```go
Eventually(func() (*v1alpha1.Subscription, error) {
    subscription, err = crc.OperatorsV1alpha1().Subscriptions(testNamespace).Get(context.Background(), subscriptionName, metav1.GetOptions{})
    return subscription != nil && subscription.Status.InstallPlanRef.Name == fetchedInstallPlan.GetName() && subscription.Status.State == operatorsv1alpha1.SubscriptionStateUpgradePending, err
}).Should(BeTrue(), "expected new installplan for upgraded csv")
```

njhale (changes requested):
Looks good, one thought:

test/e2e/subscription_e2e_test.go
njhale: I think we probably want to start getting rid of a lot of the custom test framework if we can write test specs simply at the call sites.

it seems like maintaining custom test frameworks is one place where the cost of being DRY outweighs its value

test/e2e/subscription_e2e_test.go
exdx: That sounds good. I added back in the timeouts from the `fetchSubscription` call, in case it takes a while on CI for the subscription to get into the `UpgradePending` state.

njhale (approved):

Issue #2390, opened 2021-10-04
Flake: Metrics are generated for OLM managed resources/a CSV is created/the OLM pod restarts


timflannagan:
https://github.com/operator-framework/operator-lifecycle-manager/pull/2216 had fixed a bug where the csv_succeeded was lost between deployment pod restarts. In those changes, a new metric e2e test was created that restarts (e.g. scales down/scales back up) to test whether the metric was retained after pod restarts:

```go
			When("the OLM pod restarts", func() {
				BeforeEach(func() {
					restartDeploymentWithLabel(c, "app=olm-operator")
				})
				It("CSV metric is preserved", func() {
					Expect(getMetricsFromPod(c, getPodWithLabel(c, "app=olm-operator"))).To(
						ContainElement(LikeMetric(WithFamily("csv_succeeded"), WithName(csv.Name), WithValue(1))),
					)
				})
			})
```

It looks like the `restartDeploymentWithLabel(...)` doesn't have enough safeguards to verify that the restarted deployment is ready and available, leading to issues running this e2e test on more bloated clusters, as we're attempting to grab the metric before the metric endpoint has been setup and ready to serve traffic.

timflannagan:
Closing this in favor of the duplicate I had just opened that now has more context than this issue.

PR #2393, opened 2021-10-04
chore(make): add e2e test options

njhale:
- Add more config options to the e2e and e2e-local test options
- Refactor e2e-local target to extend the e2e target

Signed-off-by: Nick Hale <njohnhale@gmail.com>


.github/workflows/e2e-kind.yml
```diff
@@ -22,5 +22,5 @@ jobs:
       run: |
         kind create cluster
         kind export kubeconfig
-    - name: Run e2e tests
```
njhale, line 25: These aren't really running e2e tests so much as they're ensuring we can deploy OLM to a specific target; how valuable is this?

timflannagan: I had previously attempted to remove these checks entirely but there was some pushback in #olm-dev as they're mainly in place to ensure there's no regression when running OLM on various cluster environments.

Makefile
```diff
@@ -119,29 +123,23 @@ setup-bare: clean e2e.namespace
 	. ./scripts/package_release.sh 1.0.0 test/e2e/resources test/e2e/e2e-bare-values.yaml
 	. ./scripts/install_bare.sh $(shell cat ./e2e.namespace) test/e2e/resources
 
-GINKGO_OPTS := -flakeAttempts 3 -randomizeAllSpecs -v --timeout 120m
+E2E_NODES ?= 1
+E2E_FLAKE_ATTEMPTS ?= 1
+E2E_TIMEOUT ?= 90m
+E2E_OPTS ?= $(if $(E2E_SEED),-seed '$(E2E_SEED)') $(if $(TEST),-focus '$(TEST)') -flakeAttempts $(E2E_FLAKE_ATTEMPTS) -nodes $(E2E_NODES) -timeout $(E2E_TIMEOUT) -randomizeAllSpecs -v
+E2E_INSTALL_NS ?= operator-lifecycle-manager
+E2E_TEST_NS ?= operators
```
njhale, line 131: Reminder: we need to set the downstream-specific namespaces in the root `Makefile` downstream before invoking this target; e.g. `E2E_INSTALL_NS=openshift-operator-lifecycle-manager E2E_TEST_NS=openshift-operators  $(MAKE) staging/operator-lifecycle-manager/Makefile`

.github/workflows/e2e-kind.yml
timflannagan: But for context, I had renamed this check and simply misjudged what these checks were trying to accomplish when splitting up that e2e-tests GHA workflow into their own workflows.

timflannagan:
`CSV metric is preserved` - #2390

timflannagan:
Hold so the bot doesn't go crazy.

/hold

PR #2395, opened 2021-10-04
fix(e2e): Use Eventually for the last CSV GC check

dinhxuanvu:

**Reviewer Checklist**


test/e2e/operator_groups_e2e_test.go
```diff
@@ -2023,7 +2023,7 @@ var _ = Describe("Operator Group", func() {
 		require.NoError(GinkgoT(), retry.RetryOnConflict(retry.DefaultBackoff, updateCSV))
 
 		// wait for CSV to be gc'd
-		err = wait.Poll(pollInterval, 2*pollDuration, func() (bool, error) {
+		Eventually(func() (bool, error) {
```
timflannagan, line 2026: Is there any functional difference swapping from wait.Poll to Eventually here, except we get the default eventually handler configurations for free?

dinhxuanvu: As far as I know, there are no fundamental differences. I use Eventually here to take advantage of Gingko for consistency. However, I notice on other tests that have a mix of `fetchSomething` (which uses wait.Poll) and Eventually, the `fetch` function tends to flake out more often. Not sure why.

PR #2399, opened 2021-10-05
fix(sub): Only update subscriptions that have changes in status

dinhxuanvu:
Currently, when resolution happens, every subscription will get
updated regardless if there are any changes applied or not. This
resolves into unnecessary API update calls.

This commit will filter out subscriptions that don't get changed
and only changed ones get updated.

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -962,30 +962,33 @@ func (o *Operator) syncResolvingNamespace(obj interface{}) error {
 		// not-satisfiable error
 		if _, ok := err.(solver.NotSatisfiable); ok {
 			logger.WithError(err).Debug("resolution failed")
-			subs = o.setSubsCond(subs, v1alpha1.SubscriptionResolutionFailed, "ConstraintsNotSatisfiable", err.Error(), true)
-			_, updateErr := o.updateSubscriptionStatuses(subs)
+			_, updateErr := o.updateSubscriptionStatuses(
+				o.setSubsCond(subs, v1alpha1.SubscriptionCondition{
+					Type:    v1alpha1.SubscriptionResolutionFailed,
+					Reason:  "ConstraintsNotSatisfiable",
+					Message: err.Error(),
+					Status:  corev1.ConditionTrue,
+				}))
 			if updateErr != nil {
 				logger.WithError(updateErr).Debug("failed to update subs conditions")
 				return updateErr
 			}
 			return nil
 		}
-		subs = o.setSubsCond(subs, v1alpha1.SubscriptionResolutionFailed, "ErrorPreventedResolution", err.Error(), true)
-		_, updateErr := o.updateSubscriptionStatuses(subs)
+
+		_, updateErr := o.updateSubscriptionStatuses(
+			o.setSubsCond(subs, v1alpha1.SubscriptionCondition{
+				Type:    v1alpha1.SubscriptionResolutionFailed,
+				Reason:  "ErrorPreventedResolution",
+				Message: err.Error(),
+				Status:  corev1.ConditionTrue,
+			}))
 		if updateErr != nil {
 			logger.WithError(updateErr).Debug("failed to update subs conditions")
 			return updateErr
 		}
 		return err
 	}
-
-	defer func() {
-		subs = o.setSubsCond(subs, v1alpha1.SubscriptionResolutionFailed, "", "", false)
-		_, updateErr := o.updateSubscriptionStatuses(subs)
-		if updateErr != nil {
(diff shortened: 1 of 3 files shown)


pkg/controller/operators/catalog/operator.go
```diff
@@ -1225,23 +1242,45 @@ func (o *Operator) createInstallPlan(namespace string, gen int, subs []*v1alpha1
 	return reference.GetReference(res)
 }
 
-func (o *Operator) setSubsCond(subs []*v1alpha1.Subscription, condType v1alpha1.SubscriptionConditionType, reason, message string, setTrue bool) []*v1alpha1.Subscription {
+// setSubsCond will set the condition to the subscription if it doesn't already
+// exist or if it is different
+// Only return the list of updated subscriptions
+func (o *Operator) setSubsCond(subs []*v1alpha1.Subscription, cond v1alpha1.SubscriptionCondition) []*v1alpha1.Subscription {
```
timflannagan, line 1248: This is more of a general question/comment than anything blocking: at a glance, this method currently resembles a Setter but we're passing the reference to the list of subscriptions, and outputting a list of subscriptions. Any idea on whether we're actually double dipping here - modifying the input and outputting a new subscription list? I guess that might not matter a whole ton as it's contingent on how the call site is handling this logic, and it's possible my brain power is limited going into EOD.

pkg/controller/operators/catalog/operator.go
```diff
@@ -962,31 +962,34 @@ func (o *Operator) syncResolvingNamespace(obj interface{}) error {
...
+				Status:  corev1.ConditionTrue,
+			}))
 		if updateErr != nil {
 			logger.WithError(updateErr).Debug("failed to update subs conditions")
 			return updateErr
 		}
 		return err
 	}
 
-	defer func() {
```
timflannagan, line 982: +1 to removing this defer case. I think this reads a lot more clearer, and being more explicit in this code path likely helps limit side effects.

pkg/controller/operators/catalog/operator.go
```diff
@@ -1017,17 +1020,36 @@ func (o *Operator) syncResolvingNamespace(obj interface{}) error {
...
-					subs[i] = updatedSub
-				}
-			}
-		}
 	} else {
 		logger.Debugf("no subscriptions were updated")
 	}
 
+	// Remove resolutionfailed condition from subscriptions
+	subs = o.removeSubsCond(subs, v1alpha1.SubscriptionResolutionFailed)
```
timflannagan, line 1028: Just a quick note: it feels like this function may be a candidate for refactoring in the future. I don't think we need to tackle  that kind of change in this PR but I wanted to get any thoughts on whether you had that same though when poking at this locally?

pkg/controller/operators/catalog/operator.go
```diff
@@ -1225,23 +1242,45 @@ func (o *Operator) createInstallPlan(namespace string, gen int, subs []*v1alpha1
...
 	var (
 		lastUpdated = o.now()
+		subList     []*v1alpha1.Subscription
 	)
+
 	for _, sub := range subs {
+		subCond := sub.Status.GetCondition(cond.Type)
+		if subCond.Equals(cond) {
+			continue
+		}
```
timflannagan, line 1258: It looks like this Equals method does account for the edge case where the existing and new status conditions are equal, and avoids bumping the subCond.LastTransitionTime unnecessarily, causing another hotlooping scenario. Any value in adding a debug logging here, or is that just overkill?

dinhxuanvu: We can add a log here to indicate the condition already exists. I think it can be helpful for debugging reason.

pkg/controller/operators/catalog/operator.go
dinhxuanvu: This is just a "lazy" solution in term of not wanting to deal with removing element(s) from existing list by slicing the array. I can change this to a solution that will do the slicing so that the existing modified list will be returned instead of a new one.

dinhxuanvu (commented):
Responded.

pkg/controller/operators/catalog/operator.go
timflannagan: Just copy-pasting the comment chain on the other PR: https://github.com/operator-framework/operator-lifecycle-manager/pull/2429#discussion_r758548598.

timflannagan: I'd like to hold off on refactoring and keeping the scope of these changes small as they're affecting upstream/downstream e2e runs, and combing down hot looping behavior in some edge cases.

timflannagan:
Let's tackling any refactoring outside of this PR as this is blocked e2e runs.

/hold cancel

PR #2418, opened 2021-10-11
feat(resolver): `olm.constraint` property and compound constraints

estroz:
- Add [`olm.constraint`](https://github.com/operator-framework/enhancements/pull/91) dependency property to resolver
- Add [`all`, `any`, `none`](https://github.com/operator-framework/enhancements/pull/97) constraint types

**Motivation for the change:** compound dependency constraints (see https://github.com/operator-framework/enhancements/pull/97).

**Reviewer Checklist**

/kind feature


estroz:
I'm in the process of adding resolver unit tests for compound constraints.

pkg/controller/registry/resolver/constraints/constraint.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+// maxConstraintSize defines the maximum raw size in bytes of an olm.constraint.
+// 64Kb seems reasonable, since this number allows for long description strings
+// and either few deep nestings or shallow nestings and long constraints lists,
+// but not both.
+const maxConstraintSize = 2 << 16
+
+// ErrMaxConstraintSizeExceeded is returned when a constraint's size > maxConstraintSize.
+var ErrMaxConstraintSizeExceeded = fmt.Errorf("olm.constraint value is greater than max constraint size %d", maxConstraintSize)
```
perdasilva, line 51: Might be worth adding the unit

perdasilva: For extra points you could add the actual size and the first 10-20 characters of it?

pkg/controller/registry/resolver/resolver.go
```diff
@@ -748,18 +753,73 @@ func DependencyPredicates(properties []*api.Property) ([]cache.Predicate, error)
...
 	if property == nil {
 		return nil, nil
 	}
-	p, ok := predicates[property.Type]
+
+	if property.Type == constraints.OLMConstraintType {
+		return pc.predicateForConstraintProperty(property.Value)
+	}
+
+	// Legacy behavior dictates that unknown properties are ignored.
```
perdasilva, line 765: Could it be worth adding more context here? (Maybe point to the discussion in the EP). Just so the next dev understands what's happening and why.

pkg/controller/registry/resolver/resolver.go
```diff
@@ -27,12 +28,14 @@ type OperatorResolver interface {
 type SatResolver struct {
 	cache cache.OperatorCacheProvider
```
dinhxuanvu, line 29: Just a preliminary review: given this is a new type of constraint, we should have a test case for it in resolver_test.go to ensure the compound constraint will get resolved into what we expect it to be.

dinhxuanvu (changes requested):
Just took a quick glance. Will do further review later.

pkg/controller/registry/resolver/resolver.go
estroz: I'll be pushing up a bunch of tests today

pkg/controller/registry/resolver/constraints/constraint.go
```diff
@@ -0,0 +1,69 @@
...
+	"fmt"
+)
+
+const OLMConstraintType = "olm.constraint"
+
+// Constraint holds parsed, potentially nested dependency constraints.
+type Constraint struct {
+	Message string `json:"message"`
+
+	Package *PackageConstraint `json:"package,omitempty"`
```
estroz, line 15: I can add this stuff to the api repo if that's the direction we decide on.

pkg/controller/registry/resolver/resolver.go
```diff
@@ -733,10 +736,14 @@ func sortChannel(bundles []*cache.Entry) ([]*cache.Entry, error) {
 	return chains[0], nil
 }
 
-func DependencyPredicates(properties []*api.Property) ([]cache.Predicate, error) {
+// predicateConverter configures olm.constraint value -> predicate conversion for the resolver.
+type predicateConverter struct{}
```
estroz, line 740: This will contain fields when CEL is added.

pkg/controller/registry/resolver/cache/predicates.go
```diff
@@ -282,6 +281,32 @@ func (p orPredicate) String() string {
 	return b.String()
 }
 
+func Not(p ...Predicate) Predicate {
```
gallettilance, line 284: this is a bit confusing to me. `Not` is usually an operator on a predicate not a list of predicates. Looking at the `Test()` it seems to imply that the list of predicate is implied to be a conjunction. Lmk if I'm missing something but coming into this cold I would have expected to use `Not` on a single predicate. If not, could we name it `None` or `Neither` or something.

estroz: I think the same logic (no pun intended) applies to "and" vs. "all", and "or" vs. "any", but that's how these resolver primitives were named. I'm just following the current naming scheme.

dinhxuanvu (approved):
The PR looks good.
/approve

PR #2422, opened 2021-10-13
Expose pprof endpoint if tls is not configured

awgreene:
Problem: Currently, if OLM is not configured to use TLS it is not
possible to reach the pprof endpoint. This limitation makes it
difficult to debug complex performance problems on clusters where
OLM is not configured to use tls.

Solution: If OLM is not configured to use tls, do not require a
tls certificate to access the pprof endpoint.


awgreene:
CC @benluddy for his thoughts on this.

PR #2423, opened 2021-10-13
chore(deps): bump grpc dependency to latest

exdx:

Bumps the version of the gRPC library to latest

Alongside bumping the gRPC version server-side in https://github.com/operator-framework/operator-registry/pull/810 this should improve the gRPC connection performance between OLM and registry pods on-cluster. 

**Reviewer Checklist**


timflannagan:
I haven't dived too deep into those e2e failures, but it looks like the majority of them are due to the test CatalogSource(s) being stuck in IDLE-ing states. Let's try and dive into those failures more, with the intention of determining whether these changes are applicable (e.g. influencing) those grpc-related failures.

/hold

timflannagan (commented):
There's already a hold placed on this PR but I ran into an issue locally debugging the registry repository bumping this dependency and it seems like there are some behavior changes between the 1.38.x release and the 1.41.x release we're pulling in here. 

This results in test catalogsources stuck in the IDLE last observed state, leading to the full 5m test period timeout being reached. Poking around the grpc client releases, I'd guess that the 1.41 release is the problematic one for us:

> balancer: SubConns no longer automatically reconnect after READY; instead they transition to IDLE on connection loss

So likely need to land changes to our own codebase to reflect those behavior changes in the past couple of minor releases.

timflannagan:
I'm going to close this out as it has merge conflicts - we'll need to open an issue for upstream OLM that explains that bumping to the latest grpc version is problematic (and we'll need to investigate why).

PR #2424, opened 2021-10-13
fix(e2e): Run each test in test suite in different namespace

dinhxuanvu:
All test cases in subscription suite shares the same test namespace.
In some cases, the resource cleanup which doesn't work properly from
previous test causes failure on subsequent test.

This commit will create new namespace for each test so it can be run
in insolation.

**Reviewer Checklist**

### DIFF
--- a/test/e2e/catalog_e2e_test.go
+++ b/test/e2e/catalog_e2e_test.go
@@ -34,17 +34,23 @@ import (
 
 var _ = Describe("Catalog represents a store of bundles which OLM can use to install Operators", func() {
 	var (
-		c   operatorclient.ClientInterface
-		crc versioned.Interface
+		c                  operatorclient.ClientInterface
+		crc                versioned.Interface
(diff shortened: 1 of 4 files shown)


timflannagan (commented):
Looking at the existing e2e codebase, it looks like introducing knobs to capture individual test failures is more nuanced than simply adding a top-level BeforeEach/AfterEach clause that generates a testing namespace for us. Given there's a merge conflict right now, do we want to close this PR out and incrementally introducing support on a per-package-basis or did you have a better approach in mind. Any general thoughts?

timflannagan:
It looks like this PR is fairly out-of-date at this point, and has a couple of conflicts that need to be resolved, so I'm going to close this out in the meantime. The overall test pollution is still an issue, but it's been watered down recently by breaking down the overall e2e suite into individual clusters that are responsible for running a set of jobs, and filtering out common flaky e2e tests into their own workflow.

PR #2427, opened 2021-10-18
pkg/controller: Prevent unnecessary status update's while resolving subscriptions in a namespace

timflannagan:
Update the catalog operator's operator.go package and expand the
safeguards in the o.setSubsCond(...) method to avoid updating the status
sub-resource of a subscription's conditions array when the new status
condition type and the type's status are already present in the
conditions array.

Previously, we were blindly firing off status sub-resource update calls,
which led to hotlooping in some scenarios.

**Reviewer Checklist**


timflannagan:
Closing in favor of #2429

PR #2429, opened 2021-10-19
fix(sub): Prevent unnecessary subscription update calls on condition changes

dinhxuanvu:
Currently, all subscriptions will get updated during resolution to either add or remove ResolutionFailed condition. This is unnecessary given not all subscriptions have/not have these conditions. All impacted subscriptions should get updated.

Update subscription e2e suite to run individual test on separate namespace to avoid inter-test-case failure.

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -962,30 +962,33 @@ func (o *Operator) syncResolvingNamespace(obj interface{}) error {
 		// not-satisfiable error
 		if _, ok := err.(solver.NotSatisfiable); ok {
 			logger.WithError(err).Debug("resolution failed")
-			subs = o.setSubsCond(subs, v1alpha1.SubscriptionResolutionFailed, "ConstraintsNotSatisfiable", err.Error(), true)
-			_, updateErr := o.updateSubscriptionStatuses(subs)
+			_, updateErr := o.updateSubscriptionStatuses(
+				o.setSubsCond(subs, v1alpha1.SubscriptionCondition{
+					Type:    v1alpha1.SubscriptionResolutionFailed,
+					Reason:  "ConstraintsNotSatisfiable",
+					Message: err.Error(),
+					Status:  corev1.ConditionTrue,
+				}))
 			if updateErr != nil {
 				logger.WithError(updateErr).Debug("failed to update subs conditions")
 				return updateErr
 			}
 			return nil
 		}
-		subs = o.setSubsCond(subs, v1alpha1.SubscriptionResolutionFailed, "ErrorPreventedResolution", err.Error(), true)
-		_, updateErr := o.updateSubscriptionStatuses(subs)
+
+		_, updateErr := o.updateSubscriptionStatuses(
+			o.setSubsCond(subs, v1alpha1.SubscriptionCondition{
+				Type:    v1alpha1.SubscriptionResolutionFailed,
+				Reason:  "ErrorPreventedResolution",
+				Message: err.Error(),
+				Status:  corev1.ConditionTrue,
+			}))
 		if updateErr != nil {
 			logger.WithError(updateErr).Debug("failed to update subs conditions")
 			return updateErr
 		}
 		return err
 	}
-
-	defer func() {
-		subs = o.setSubsCond(subs, v1alpha1.SubscriptionResolutionFailed, "", "", false)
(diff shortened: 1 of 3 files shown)


timflannagan:
@dinhxuanvu It looks like this needs to be rebased when you get a chance now that the other testing changes have landed in master.

pkg/controller/operators/catalog/operator.go
```diff
@@ -1017,17 +1020,36 @@ func (o *Operator) syncResolvingNamespace(obj interface{}) error {
...
-				}
-			}
-		}
 	} else {
 		logger.Debugf("no subscriptions were updated")
 	}
 
+	// Remove resolutionfailed condition from subscriptions
+	subs = o.removeSubsCond(subs, v1alpha1.SubscriptionResolutionFailed)
+	newSub := true
```
perdasilva, line 1029: adding the 'is' prefix to boolean variables can improve the readability of the code a little. i.e. `isNewSub`.

pkg/controller/operators/catalog/operator.go
```diff
@@ -962,31 +962,34 @@ func (o *Operator) syncResolvingNamespace(obj interface{}) error {
 		// not-satisfiable error
 		if _, ok := err.(solver.NotSatisfiable); ok {
 			logger.WithError(err).Debug("resolution failed")
-			subs = o.setSubsCond(subs, v1alpha1.SubscriptionResolutionFailed, "ConstraintsNotSatisfiable", err.Error(), true)
-			_, updateErr := o.updateSubscriptionStatuses(subs)
+			_, updateErr := o.updateSubscriptionStatuses(
+				o.setSubsCond(subs, v1alpha1.SubscriptionCondition{
```
perdasilva, line 966: suggestion: Maybe factor the reason strings out as constants? Another suggestion might be to create a couple of factory methods for creating the `SubscriptionCondition`s. Maybe it just takes the message as input? e.g.
```
func NewResolutionFailedCondition(msg string) *v1alpha1.SubscriptionCondition
```

pkg/controller/operators/catalog/operator.go
```diff
@@ -1017,17 +1020,36 @@ func (o *Operator) syncResolvingNamespace(obj interface{}) error {
...
-			}
-		}
 	} else {
 		logger.Debugf("no subscriptions were updated")
 	}
 
+	// Remove resolutionfailed condition from subscriptions
+	subs = o.removeSubsCond(subs, v1alpha1.SubscriptionResolutionFailed)
+	newSub := true
+	for _, updatedSub := range updatedSubs {
```
perdasilva, line 1030: suggestion: maybe it's worth moving this to its own function? Then rather than mutating the subs slice, just return a new slice?

timflannagan: +1 to this suggestion, although it feels like this method as a whole may need to be refactored at some point as there's a good chunk of complexity now. maybe we could tackle the refactoring as a follow-up?

pkg/controller/operators/catalog/operator.go
```diff
@@ -1017,17 +1020,36 @@ func (o *Operator) syncResolvingNamespace(obj interface{}) error {
...
+				newSub = false
+				break
+			}
+		}
+		if newSub {
+			subs = append(subs, updatedSub)
+			continue
+		}
+		newSub = true
+	}
```
timflannagan, line 1044: Just for my own understanding: it looks like we're essentially trying to find the intersection between the `subs` and any subscriptions that are present in the `updatedSubs` that require creation and/or updates? The goal being any element missing in the subs array that's present in the updatedSubs array is merged into the former, and all subscription's have the `v1alpha1.SubscriptionResolutionFailed` set to corev1.ConditionFalse?

timflannagan (commented):
This looks reasonable to me barring dropping the e2e-test-related commit(s) and rebasing to get rid of the merge conflict. I'd like to get back to refactoring these method eventually as the cognitive load + complexity is getting fairly high, but I think we can tackle this as a follow-up as these fixes will lead to reducing the current hot looping behavior we see in certain edge cases.

pkg/controller/operators/catalog/operator.go
timflannagan: I think this is also a reasonable suggestion - tackle this in a follow-up for a general refactoring of this method?

PR #2433, opened 2021-11-03
fix(e2e): Run each test in subscription in different namespace

dinhxuanvu:
All test cases in subscription suite shares the same test namespace.
In some cases, the resource cleanup which doesn't work properly from
previous test causes failure on subsequent test.

This commit will create new namespace for each test so it can be run
in insolation.

**Reviewer Checklist**

### DIFF
--- a/test/e2e/util_test.go
+++ b/test/e2e/util_test.go
@@ -36,6 +36,7 @@ import (
 	k8scontrollerclient "sigs.k8s.io/controller-runtime/pkg/client"
 
 	gtypes "github.com/onsi/gomega/types"
+	v1 "github.com/operator-framework/api/pkg/operators/v1"
 	"github.com/operator-framework/api/pkg/operators/v1alpha1"
 	operatorsv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/clientset/versioned"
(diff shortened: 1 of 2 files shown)


test/e2e/util_test.go
```diff
@@ -36,6 +36,7 @@ import (
 	k8scontrollerclient "sigs.k8s.io/controller-runtime/pkg/client"
 
 	gtypes "github.com/onsi/gomega/types"
+	v1 "github.com/operator-framework/api/pkg/operators/v1"
 	"github.com/operator-framework/api/pkg/operators/v1alpha1"
 	operatorsv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
```
timflannagan, line 41: Feel free to complete ignore this: v1 -> operatorsv1? It also looks like there's duplicate v1alpha1 imports here too.

test/e2e/util_test.go
```diff
@@ -918,3 +919,30 @@ func HaveMessage(goal string) gtypes.GomegaMatcher {
 		return plan.Status.Message
 	}, ContainSubstring(goal))
 }
+
+func SetUpGeneratedTestNamespace(name string) (string, corev1.Namespace) {
+	generatedNamespace := genName(name)
```
timflannagan, line 924: I'd expect this to happen at the call site, but this is fine for now.

test/e2e/util_test.go
```diff
@@ -918,3 +919,30 @@ func HaveMessage(goal string) gtypes.GomegaMatcher {
...
+		},
+		Spec: v1.OperatorGroupSpec{
+			TargetNamespaces: []string{ns.GetName()},
+		},
+	}
+	Eventually(func() error {
+		return ctx.Ctx().Client().Create(context.Background(), &og)
+	}).Should(Succeed())
+
+	return generatedNamespace, ns
```
timflannagan, line 947: Hmm, not opposed to this function signature, but you can access the generated namespace's name through `ns.GetName()` type operations (or indexing into the metadata.Name field directly).

test/e2e/util_test.go
```diff
@@ -918,3 +919,30 @@ func HaveMessage(goal string) gtypes.GomegaMatcher {
...
+	og := v1.OperatorGroup{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      fmt.Sprintf("%s-operatorgroup", ns.GetName()),
+			Namespace: ns.GetName(),
+		},
+		Spec: v1.OperatorGroupSpec{
+			TargetNamespaces: []string{ns.GetName()},
+		},
+	}
+	Eventually(func() error {
```
timflannagan, line 943: :+1:

test/e2e/util_test.go
```diff
@@ -918,3 +919,30 @@ func HaveMessage(goal string) gtypes.GomegaMatcher {
 		return plan.Status.Message
 	}, ContainSubstring(goal))
 }
+
+func SetUpGeneratedTestNamespace(name string) (string, corev1.Namespace) {
+	generatedNamespace := genName(name)
+	ns := corev1.Namespace{
+		ObjectMeta: metav1.ObjectMeta{
```
timflannagan, line 926: Mainly curious on when this is necessary: do we need to specify TypeMeta here as well?

timflannagan (commented):
Thanks, this should be help reduce many of the existing subscription-related e2e failures that are due to test pollution affect unrelated tests, and making debugging those failures less difficult. I only had one real comment: any reason we need to supply both a generatedNamespace and a corev1.Namespace as output variables from the SetUpGeneratedTestNamespace function? We should be able to infer the generated namespace by indexing into the corev1.Namespace's fields.

dinhxuanvu:
The commit is included in #2438 which is merged.
/close

PR #2435, opened 2021-11-08
Add experimental metrics to e2e runs

awgreene:
Adds counter metric for each sync performed by controller. Example:

```
$ curl -s localhost:8080/metrics | grep controller
# HELP controller_reconcile_adoption Number of reconciliations by the adoption controller
# TYPE controller_reconcile_adoption counter
controller_reconcile_adoption 1
# HELP controller_reconcile_operator Number of reconciliations by the operator controller
# TYPE controller_reconcile_operator counter
controller_reconcile_operator 0
# HELP controller_reconcile_operator_condition Number of reconciliations by the operator condition controller
# TYPE controller_reconcile_operator_condition counter
controller_reconcile_operator_condition 16
# HELP controller_reconcile_operator_condition_generator Number of reconciliations by the operator condition generator controller
# TYPE controller_reconcile_operator_condition_generator counter
controller_reconcile_operator_condition_generator 8
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,7 @@ SHELL := /bin/bash
 ORG := github.com/operator-framework
 PKG   := $(ORG)/operator-lifecycle-manager
 MOD_FLAGS := $(shell (go version | grep -q -E "1\.1[1-9]") && echo -mod=vendor)
+BUILD_TAGS := "json1"
 CMDS  := $(shell go list $(MOD_FLAGS) ./cmd/...)
 TCMDS := $(shell go list $(MOD_FLAGS) ./test/e2e/...)
 MOCKGEN := ./scripts/update_mockgen.sh
@@ -101,7 +102,7 @@ bin/cpb: FORCE
 
 $(CMDS): version_flags=-ldflags "-X $(PKG)/pkg/version.GitCommit=$(GIT_COMMIT) -X $(PKG)/pkg/version.OLMVersion=`cat OLM_VERSION`"
 $(CMDS):
-	$(arch_flags) go $(build_cmd) $(MOD_FLAGS) $(version_flags) -tags "json1" -o bin/$(shell basename $@) $@
+	$(arch_flags) go $(build_cmd) $(MOD_FLAGS) $(version_flags) -tags $(BUILD_TAGS) -o bin/$(shell basename $@) $@
 
 build: clean $(CMDS)
 
@@ -143,6 +144,7 @@ e2e:
 
 # See workflows/e2e-tests.yml See test/e2e/README.md for details.
 .PHONY: e2e-local
+e2e-local: BUILD_TAGS="json1 experimental_metrics"
 e2e-local: extra_args=-kind.images=../test/e2e-local.image.tar
 e2e-local: run=bin/e2e-local.test
 e2e-local: bin/e2e-local.test test/e2e-local.image.tar
--- a/pkg/metrics/experimental_register.go
+++ b/pkg/metrics/experimental_register.go
@@ -1,0 +1,41 @@
+//go:build experimental_metrics
+// +build experimental_metrics
+
+package metrics
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+func init() {
+	// Register experimental metrics
+	reconcileMetrics = reconcileCounters(operatorController, adoptionCSVController, adoptionSubscriptionController, operatorConditionController, operatorConditionGeneratorController)
+	registerReconcileMetrics()
+}
+
+func reconcileCounters(reconcilerNames ...string) map[string]*prometheus.CounterVec {
+	result := map[string]*prometheus.CounterVec{}
+	for _, s := range reconcilerNames {
+		result[s] = createReconcileCounterVec(s)
+	}
+	return result
+}
+
+func createReconcileCounterVec(name string) *prometheus.CounterVec {
+	return prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "controller_reconcile_" + name,
+			Help: fmt.Sprintf("Count of reconcile events by the %s controller", strings.Replace(name, "_", " ", -1)),
+		},
+		[]string{NamespaceLabel, NameLabel},
+	)
+}
+
+func registerReconcileMetrics() {
+	for _, v := range reconcileMetrics {
+		prometheus.MustRegister(v)
+	}
+}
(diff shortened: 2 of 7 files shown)


awgreene:
Wanted to get a basic implementation up quickly, will be working on the names / structure of the code.

Makefile
```diff
@@ -84,6 +85,7 @@ build-coverage: clean $(CMDS)
 
 build-linux: build_cmd=build
 build-linux: arch_flags=GOOS=linux GOARCH=386
+build-linux: BUILD_TAGS="json1 experimental_metrics"
```
timflannagan, line 88: Any idea on why we need to update the build-linux target as well? At a glance locally, it doesn't look like this is used anywhere to drive CI.

pkg/metrics/experimental.go
```diff
@@ -0,0 +1,40 @@
...
+		result[s] = reconcileCounter(s)
+	}
+	return result
+}
+
+func reconcileCounter(name string) prometheus.Counter {
+	return prometheus.NewCounter(
+		prometheus.CounterOpts{
+			Name: "controller_reconcile_" + name,
+			Help: fmt.Sprintf("Count of reconcile events by the %s controller", strings.Replace(name, "_", " ", -1)),
```
timflannagan, line 31: Would you be able to provide some context as to why these types of reconciliation counter gauges are useful? Anything in particular you had in mind that the introduction of these metrics help improve?

timflannagan (commented):
Just a quick review + played around with this locally and was able to extract the metrics during an e2e run from the olm-operator deployment, and I think the build tag implementation is the right approach.

I had a couple of quick thoughts and comments:
- Do we need to update the tes/e2e/README.md for how to interact with these metrics?
- Should we scope down the introduction of these metrics to only CI until we have a need for them outside of ephemeral environments?

Makefile
awgreene: I believe the e2e test relies on the [e2e-kind GitHub Workflow](https://github.com/operator-framework/operator-lifecycle-manager/blob/6dc902654d4f2089da9da56cc6b1d574be995d20/.github/workflows/e2e-kind.yml), which calls the`run-local` target [here](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/.github/workflows/e2e-kind.yml#L26), which calls the build-linux target [here](https://github.com/operator-framework/operator-lifecycle-manager/blob/41d0002d659f43eb07d45609dc9b56478fbc3c50/Makefile#L284).

This change might be better placed in the `run-local` target, thoughts?

pkg/metrics/experimental.go
awgreene: Primary value add is the ability to identify when a controller hoploops / if it is capable of doing so. I need to change this to a counterVec so we can capture which resource is triggering the hotloop.

timflannagan: Awesome, yeah that makes a lot more sense to me. This should be really helpful for identifying issues long term through CI.

pkg/metrics/experimental_register.go
```diff
@@ -0,0 +1,40 @@
...
+)
+
+func init() {
+	// Register experimental metrics
+	reconcileMetrics = reconcileCounters(operatorController, adoptionCSVController, adoptionSubscriptionController, operatorConditionController, operatorConditionGeneratorController)
+	registerReconcileMetrics()
+}
+
+func reconcileCounters(reconcilerNames ...string) map[string]prometheus.CounterVec {
+	result := map[string]prometheus.Counter{}
```
timflannagan, line 19: Should this be prometheus.CounterVec{} now that we've updated them from counters to counter vectors?

pkg/metrics/experimental_register.go
```diff
@@ -0,0 +1,40 @@
...
+	for _, s := range reconcilerNames {
+		result[s] = reconcileCounter(s)
+	}
+	return result
+}
+
+func reconcileCounter(name string) prometheus.CounterVec {
+	return prometheus.NewCounterVec(
```
timflannagan, line 27: Poking at the godoc for the Prometheus library, it looks like the signature returns a pointer to a prometheus.CounterVec, but we're expecting to return the value of that variable, vs. the reference in reconcileCounter.

pkg/metrics/experimental_register.go
timflannagan: Ah, yeah looking at the e2e-test output it wasn't able to compile:

```
go build -mod=vendor -ldflags "-X github.com/operator-framework/operator-lifecycle-manager/pkg/version.GitCommit=22d1a703fea54d1b562dfe99f1edc110e06d17f5 -X github.com/operator-framework/operator-lifecycle-manager/pkg/version.OLMVersion=`cat OLM_VERSION`" -tags "json1 experimental_metrics" -o bin/catalog github.com/operator-framework/operator-lifecycle-manager/cmd/catalog
# github.com/operator-framework/operator-lifecycle-manager/pkg/metrics
pkg/metrics/experimental_register.go:21:13: cannot use reconcileCounter(s) (type prometheus.CounterVec) as type prometheus.Counter in assignment:
	prometheus.CounterVec does not implement prometheus.Counter (missing Add method)
pkg/metrics/experimental_register.go:23:2: cannot use result (type map[string]prometheus.Counter) as type map[string]prometheus.CounterVec in return argument
pkg/metrics/experimental_register.go:27:33: cannot use prometheus.NewCounterVec(prometheus.CounterOpts{...}, []string{...}) (type *prometheus.CounterVec) as type prometheus.CounterVec in return argument
make: *** [Makefile:105: github.com/operator-framework/operator-lifecycle-manager/cmd/catalog] Error 2
```

awgreene: My mistake, fixed this.

timflannagan:
@awgreene It looks like the the DCO check still needs to be addressed.

pkg/metrics/experimental_register.go
awgreene: Thanks for your patience @timflannagan, I am working on getting this over the finish line now. I am not sure that I am capturing what you're trying to say here.

pkg/metrics/experimental_register.go
```diff
@@ -0,0 +1,40 @@
...
+
+func reconcileCounters(reconcilerNames ...string) map[string]prometheus.CounterVec {
+	result := map[string]prometheus.Counter{}
+	for _, s := range reconcilerNames {
+		result[s] = reconcileCounter(s)
+	}
+	return result
+}
+
+func reconcileCounter(name string) prometheus.CounterVec {
```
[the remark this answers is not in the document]
awgreene: Changed this to `createReconcileCounterVec` to better reflect the function.

pkg/metrics/experimental_register.go
timflannagan: It looks like you've addressed this issue since the last rebase + updates. This was an issue where we were using the wrong type definition, so we had seen that compilation was failing throughout the various CI checks.

PR #2436, opened 2021-11-08
test/e2e: Parallelize the process of waiting for CatalogSource creation and waiting for ready states

timflannagan:
Update test/e2e/subscription_e2e_test.go and parallelize the process of
creating duplicate CatalogSource resources and waiting for those
resources to report a ready status in the "creation with dependencies"
e2e test. That e2e test typically runs ~5m dependending on the cluster
environment, and most of that run time is waiting on the prerequisite
resources to become ready before running the core testing logic.

Previously, this test would serially create a duplicate CatalogSource,
wait until it's reporting a healthy state, then creating another, etc.
Because we have no need for resource ordering, we can offload that
creation/health check process to a goroutine, and wait until all the
resources are reporting healthy state before we create the Subscription
resource. This should lead to an average test time of ~1.5m when playing
around with these changes locally.

**Reviewer Checklist**

### DIFF
--- a/test/e2e/subscription_e2e_test.go
+++ b/test/e2e/subscription_e2e_test.go
@@ -1410,14 +1410,31 @@ var _ = Describe("Subscription", func() {
 		_, err := fetchCatalogSourceOnStatus(crClient, catsrc.GetName(), testNamespace, catalogSourceRegistryPodSynced)
 		require.NoError(GinkgoT(), err)
 
-		// Create duplicates of the CatalogSource
-		for i := 0; i < 10; i++ {
-			duplicateCatsrc, _, duplicateCatSrcCleanup := newCatalogSourceWithDependencies(GinkgoT(), kubeClient, crClient, "podconfig", testNamespace, permissions)
-			defer duplicateCatSrcCleanup()
+		duplicateCatalogSources := 10
+		var (
+			wg                  sync.WaitGroup
+			cleanupDuplicatesFn []cleanupFunc
+		)
+		wg.Add(duplicateCatalogSources)
 
-			// Ensure that the catalog source is resolved before we create a subscription.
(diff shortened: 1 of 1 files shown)


test/e2e/subscription_e2e_test.go
```diff
@@ -1410,14 +1410,31 @@ var _ = Describe("Subscription", func() {
...
+			wg                  sync.WaitGroup
+			cleanupDuplicatesFn []cleanupFunc
+		)
+		wg.Add(duplicateCatalogSources)
+
 		// Create duplicates of the CatalogSource
-		for i := 0; i < 10; i++ {
-			duplicateCatsrc, _, duplicateCatSrcCleanup := newCatalogSourceWithDependencies(GinkgoT(), kubeClient, crClient, "podconfig", testNamespace, permissions)
-			defer duplicateCatSrcCleanup()
+		for i := 0; i < duplicateCatalogSources; i++ {
```
timflannagan, line 1421: It's not immediately clear to me why 10 was chosen so I wanted to hold off on simply decrementing the counter without realizing the full context. I think a simpler solution to cut down on this long running e2e test is to simply decrement the number of duplicate CatalogSource resources that are created.

test/e2e/subscription_e2e_test.go
```diff
@@ -1410,14 +1410,31 @@ var _ = Describe("Subscription", func() {
 		_, err := fetchCatalogSourceOnStatus(crClient, catsrc.GetName(), testNamespace, catalogSourceRegistryPodSynced)
 		require.NoError(GinkgoT(), err)
```
timflannagan, line 1411: I realize I can also deflate this logic into the for loop and likely cut this test cases' run time even further.

timflannagan:
Ben had mentioned that this may be more applicable as a unit test.

/hold

PR #2442, opened 2021-11-10
Introduce a mechanism for gathering test artifacts during individual test failures

timflannagan:
Update the testing e2e suite and add a mechanism for gathering test
artifacts during individual test failures. Currently, container logs are
gathered when deprovisioning upstream kind clusters, yet we lack
fine-grain ability to diagnose test failures further.

Note: test failures use the `CurrentGinkgoTestDescription.Failed` field
to determine failures. Testing artifacts are only gathered when the base
$ARTIFACTS_DIR environment variable has been specified.

Add a collect-ci-artifacts.sh bash script, responsible for gathering OLM
native resources for an individual testing namespace. This bash script
will be called when tearing down the generated testing namespace for
relevant e2e packages. Currently, the artifact gathering process is
restricted to only a single namespace - longer term, it might be
possible to instead migrate towards collecting resources that all share
a similar label selector, and utilizing that label selector to handle
multi-namespace testing scenarios.

Introduce another helper function in test/e2e/util_test.go that's
responsible for gathering test artifacts (i.e. calling this newly
introduced script) when the test case had failed, and in either case,
delete the namespace.

**Reviewer Checklist**

### DIFF
--- a/test/e2e/collect-ci-artifacts.sh
+++ b/test/e2e/collect-ci-artifacts.sh
@@ -1,0 +1,26 @@
+#! /bin/bash
+
+set -o pipefail
+set -o nounset
+set -o errexit
+
+: "${KUBECONFIG:?}"
+: "${TEST_NAMESPACE:?}"
+: "${TEST_ARTIFACTS_DIR:?}"
+
+mkdir -p "${TEST_ARTIFACTS_DIR}"
+
+commands=()
+commands+=("get subscriptions -o yaml")
+commands+=("get operatorgroups -o yaml")
+commands+=("get clusterserviceversions -o yaml")
+commands+=("get installplans -o yaml")
+commands+=("get pods -o wide")
+commands+=("get events --sort-by .lastTimestamp")
+
+echo "Storing the test artifact output in the ${TEST_ARTIFACTS_DIR} directory"
+for command in "${commands[@]}"; do
+    echo "Collecting ${command} output..."
+    COMMAND_OUTPUT_FILE=${TEST_ARTIFACTS_DIR}/${command// /_}
+    kubectl -n ${TEST_NAMESPACE} ${command} >> "${COMMAND_OUTPUT_FILE}"
+done
--- a/test/e2e/ctx/ctx.go
+++ b/test/e2e/ctx/ctx.go
@@ -2,6 +2,9 @@ package ctx
 
 import (
 	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
 	"strings"
 
 	. "github.com/onsi/ginkgo"
@@ -32,6 +35,7 @@ type TestContext struct {
 	dynamicClient  dynamic.Interface
 	packageClient  pversioned.Interface
 	ssaClient      *controllerclient.ServerSideApplier
+	artifactsDir   string
 
 	scheme *runtime.Scheme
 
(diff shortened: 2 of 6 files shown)


test/e2e/ctx/ctx.go
```diff
@@ -86,6 +90,35 @@ func (ctx TestContext) SSAClient() *controllerclient.ServerSideApplier {
...
+		return err
+	}
+	envvars := []string{
+		"TEST_NAMESPACE=" + namespace,
+		"TEST_ARTIFACTS_DIR=" + logDir,
+		"KUBECONFIG=" + os.Getenv("KUBECONFIG"),
+	}
+
+	// compiled test binary running e2e tests is run from the root ./bin directory
+	cmd := exec.Command("../test/e2e/collect-ci-artifacts.sh")
```
timflannagan, line 111: Note: likely need to stat this file vs. relying on a local file reference.

test/e2e/util_test.go
```diff
@@ -946,3 +946,20 @@ func SetupGeneratedTestNamespace(name string) corev1.Namespace {
...
 }
+
+func TeardownNamespace(ns *corev1.Namespace) {
+	log := ctx.Ctx().Logf
+
+	currentTest := CurrentGinkgoTestDescription()
+	if currentTest.Failed {
+		log("collecting the %s namespace artifacts as the '%s' test case failed", ns.GetName(), currentTest.TestText)
+		if err := ctx.Ctx().DumpNamespaceArtifacts(ns.GetName()); err != nil {
+			log("failed to collect namespace artifacts: %v", err)
```
timflannagan, line 957: We avoid performing operations like `Expect(err).ToBe(Nil())` here as we want the namespace to be deleted in any case, regardless of whether this method fails to gather artifacts.

test/e2e/collect-ci-artifacts.sh
```diff
@@ -0,0 +1,26 @@
...
+commands+=("get operatorgroups -o yaml")
+commands+=("get clusterserviceversions -o yaml")
+commands+=("get installplans -o yaml")
+commands+=("get pods -o wide")
+commands+=("get events --sort-by .lastTimestamp")
+
+echo "Storing the test artifact output in the ${TEST_ARTIFACTS_DIR} directory"
+for command in "${commands[@]}"; do
+    echo "Collecting ${command} output..."
+    COMMAND_OUTPUT_FILE=${TEST_ARTIFACTS_DIR}/${command// /_}
```
timflannagan, line 24: This output file likely needs some work - here's an example of an e2e run locally:

```bash
$ make e2e-local ARTIFACTS_DIR=/tmp/artifacts
...
$ tree /tmp/artifacts/catalog-e2e-2jq84/
/tmp/artifacts/catalog-e2e-2jq84/
├── get_clusterserviceversions_-o_yaml
├── get_events_--sort-by_.lastTimestamp
├── get_installplans_-o_yaml
├── get_operatorgroups_-o_yaml
├── get_pods_-o_wide
└── get_subscriptions_-o_yaml
```

test/e2e/ctx/ctx.go
```diff
@@ -86,6 +90,35 @@ func (ctx TestContext) SSAClient() *controllerclient.ServerSideApplier {
...
 }
 
+func (ctx TestContext) DumpNamespaceArtifacts(namespace string) error {
+	if ctx.artifactsDir == "" {
+		ctx.Logf("$ARTIFACTS_DIR is unset -- not collecting failed test case logs")
+		return nil
+	}
+	ctx.Logf("collecting logs in the %s artifacts directory", ctx.artifactsDir)
+
+	logDir := filepath.Join(ctx.artifactsDir, namespace)
```
timflannagan, line 100: I'm open to suggestions on how to best dump individual test failure artifacts. In the current implementation, this creates a directory named after whatever namespace was generated for that test. It wouldn't be immediately clear how to map this to an individual test failure, outside of looking at the overall CI logs that are produced during a run (as the namespace generated is logged).

test/e2e/collect-ci-artifacts.sh
```diff
@@ -0,0 +1,26 @@
...
+    kubectl -n ${TEST_NAMESPACE} ${command} >> "${COMMAND_OUTPUT_FILE}"
```
timflannagan, line 25: Note: this script would also neglect deleting any "empty" files here - any file that contains an empty List will still be created and housed in this directory:

```
apiVersion: v1
items: []
kind: List
metadata:
  resourceVersion: ""
  selfLink: ""
```

anik120 (approved):
/lgtm

PR #2444, opened 2021-11-12
wait for catalogsource pod deletion

### LINKED ISSUE #2440: "Catalog represents a store of bundles which OLM can use to install Operators config map update triggers registry pod rollout"

akihikokuroda:

Add code that wait for deletion of the catalogsource pod in the CatalogSourceCleanupFunc.

Possible fix for issue https://github.com/operator-framework/operator-lifecycle-manager/issues/2440

**Reviewer Checklist**

### DIFF
--- a/test/e2e/util.go
+++ b/test/e2e/util.go
@@ -505,6 +505,18 @@ func buildCatalogSourceCleanupFunc(crc versioned.Interface, namespace string, ca
 		ctx.Ctx().Logf("Deleting catalog source %s...", catalogSource.GetName())
 		err := crc.OperatorsV1alpha1().CatalogSources(namespace).Delete(context.Background(), catalogSource.GetName(), metav1.DeleteOptions{})
 		Expect(err).ToNot(HaveOccurred())
+
+		Eventually(func() (bool, error) {
+			fetched, err := newKubeClient().KubernetesInterface().CoreV1().Pods(catalogSource.GetNamespace()).List(context.Background(), metav1.ListOptions{LabelSelector: "olm.catalogSource=" + catalogSource.GetName()})
+			if err != nil {
+				return false, err
+			}
+			if len(fetched.Items) == 0 {
+				return true, nil
+			}
+			ctx.Ctx().Logf("waiting for the catalog source %s pod to be deleted...", fetched.Items[0].GetName())
+			return false, nil
+		}).Should(BeTrue())
 	}
 }
 


test/e2e/util_test.go
```diff
@@ -504,11 +504,25 @@ func TearDown(namespace string) {
...
 }
 
-func buildCatalogSourceCleanupFunc(crc versioned.Interface, namespace string, catalogSource *operatorsv1alpha1.CatalogSource) cleanupFunc {
+func buildCatalogSourceCleanupFunc(c operatorclient.ClientInterface, crc versioned.Interface, namespace string, catalogSource *operatorsv1alpha1.CatalogSource) cleanupFunc {
 	return func() {
 		ctx.Ctx().Logf("Deleting catalog source %s...", catalogSource.GetName())
 		err := crc.OperatorsV1alpha1().CatalogSources(namespace).Delete(context.Background(), catalogSource.GetName(), metav1.DeleteOptions{})
 		Expect(err).ToNot(HaveOccurred())
+
+		if c != nil {
```
timflannagan, line 513: We shouldn't need to nil check the client here, and removing the conditional should improve readability. WDYT?

test/e2e/util_test.go
```diff
@@ -504,11 +504,25 @@ func TearDown(namespace string) {
...
+			err = wait.Poll(pollInterval, pollDuration, func() (bool, error) {
+				fetched, err := c.KubernetesInterface().CoreV1().Pods(catalogSource.GetNamespace()).List(context.TODO(), metav1.ListOptions{LabelSelector: "olm.catalogSource=" + catalogSource.GetName()})
```
timflannagan, line 515: nit: Substitute context.TODO() with context.Background()?

test/e2e/util_test.go
```diff
@@ -504,11 +504,25 @@ func TearDown(namespace string) {
...
+				if err != nil {
+					fmt.Println(err)
```
timflannagan, line 517: We have a dedicated global logger that can be accessed through ctx.Ctx().Logf(...) for any logging purposes.

test/e2e/util_test.go
```diff
@@ -504,11 +504,25 @@ func TearDown(namespace string) {
...
 
-func buildCatalogSourceCleanupFunc(crc versioned.Interface, namespace string, catalogSource *operatorsv1alpha1.CatalogSource) cleanupFunc {
+func buildCatalogSourceCleanupFunc(c operatorclient.ClientInterface, crc versioned.Interface, namespace string, catalogSource *operatorsv1alpha1.CatalogSource) cleanupFunc {
 	return func() {
 		ctx.Ctx().Logf("Deleting catalog source %s...", catalogSource.GetName())
 		err := crc.OperatorsV1alpha1().CatalogSources(namespace).Delete(context.Background(), catalogSource.GetName(), metav1.DeleteOptions{})
 		Expect(err).ToNot(HaveOccurred())
+
+		if c != nil {
+			err = wait.Poll(pollInterval, pollDuration, func() (bool, error) {
```
timflannagan, line 514: The e2e suite is a bit of a weird mixture of ginkgo/gomega and testify style assertions. Recently, we've been moving to the former and utilizing the `Eventually(...)` closure to handle polling. The equivalent to wait.Poll here would be something like the following:

```go
Eventually(func() (bool, error) {
    fetched, err := ...
    if err != nil {
        return false, err
    }
    if len(fetched.Items) == 0 {
        return true, nil
    }
    return false, nil
}
```

Any thoughts?

edit: I thought more about this, and refactoring this to use the Eventually-style function shouldn't improve too much, so feel free to disregard.

test/e2e/util_test.go
```diff
@@ -504,11 +504,25 @@ func TearDown(namespace string) {
...
+				fetched, err := c.KubernetesInterface().CoreV1().Pods(catalogSource.GetNamespace()).List(context.TODO(), metav1.ListOptions{LabelSelector: "olm.catalogSource=" + catalogSource.GetName()})
+				if err != nil {
+					fmt.Println(err)
+					return true, err
```
timflannagan, line 518: nit, feel free to disregard: `return false, err` here as we're technically not ready to exit the poll if we encounter any error while performing list operations.

test/e2e/util_test.go
```diff
@@ -548,7 +562,7 @@ func createInvalidGRPCCatalogSource(crc versioned.Interface, name, namespace str
 	catalogSource, err := crc.OperatorsV1alpha1().CatalogSources(namespace).Create(context.Background(), catalogSource, metav1.CreateOptions{})
 	Expect(err).ToNot(HaveOccurred())
 	ctx.Ctx().Logf("Catalog source %s created", name)
-	return catalogSource, buildCatalogSourceCleanupFunc(crc, namespace, catalogSource)
+	return catalogSource, buildCatalogSourceCleanupFunc(nil, crc, namespace, catalogSource)
```
timflannagan, line 565: This is an unfortunate side effect of a good chunk of testing helper functions that should be refactored at some point, but we could also just update this overall function to take a `c operatorclient.ClientInterface` parameter, and at the call sites, you can get access to an already initialized client through `newKubeClient()`.

timflannagan (commented):
Thanks for trying to tackling this and help cut down on some of the e2e tests that have been failing recently. I had a couple of quick comments/suggestions/etc. inline. I haven't poked around the testing packages to see whether this is an issue with test pollution, e.g. another, unrelated catalogsource test pod is influencing this failing assertion, but introducing additional safeguards should be sufficient. I'm hoping to take a better look later today too.

/ok-to-test

test/e2e/util_test.go
akihikokuroda: Thanks! I agree. I take out the check.

test/e2e/util_test.go
akihikokuroda: OK.  I change it.

test/e2e/util_test.go
akihikokuroda: I leave it for now.

test/e2e/util_test.go
akihikokuroda: OK.  In case of error, it probably times out.

test/e2e/util_test.go
akihikokuroda: Thanks.  I change to use newKubeClient() in the helper function instead of taking the client as an argument.

test/e2e/util_test.go
akihikokuroda: Thanks! I change it.

test/e2e/util_test.go
timflannagan: Sounds good - we'll eventually need to refactor those helper function signatures, but this should (hopefully) cut down on e2e flakes in the interim.

timflannagan:
I think these changes look reasonable to me, so I'll preemptively tag it while we wait for e2e run - we'll have to chase down an approver next week before it can merge.

/lgtm

akihikokuroda:
How can I rerun the failing test?

timflannagan:
@akihikokuroda I'm not entirely sure what kind of repository permissions you would need to restart checks unfortunately - I'm going manually recycle them as the workflows in this PR have been renamed.

test/e2e/util_test.go
```diff
@@ -509,6 +509,18 @@ func buildCatalogSourceCleanupFunc(crc versioned.Interface, namespace string, ca
...
+			fetched, err := newKubeClient().KubernetesInterface().CoreV1().Pods(catalogSource.GetNamespace()).List(context.Background(), metav1.ListOptions{LabelSelector: "olm.catalogSource=" + catalogSource.GetName()})
+			if err != nil {
+				ctx.Ctx().Logf("Error: getting catalog source pod: %s", err)
+				return false, err
+			}
+			if len(fetched.Items) == 0 {
+				return true, nil
+			}
+			return false, nil
+		})
```
timflannagan, line 523: Just coming back to this:

- It doesn't look like we're handling the wait.Poll error anywhere.
- We should try and aim for a better logging as well

WDYT about the following:

```go
Eventually(func() (bool, error) {
	fetched, err := newKubeClient().KubernetesInterface().CoreV1().Pods(catalogSource.GetNamespace()).List(context.Background(), metav1.ListOptions{LabelSelector: "olm.catalogSource=" + catalogSource.GetName()})
	if err != nil {
		return false, err
	}
	if len(fetched.Items) == 0 {
		return true, nil
	}
	ctx.Ctx().Logf("waiting for the catalog source %s pod to be deleted...", fetched.GetName())
	return false, nil
}).Should(BeTrue())
```

akihikokuroda: Thanks!  I'll take your changes.

test/e2e/util.go
```diff
@@ -505,6 +505,18 @@ func buildCatalogSourceCleanupFunc(crc versioned.Interface, namespace string, ca
 		ctx.Ctx().Logf("Deleting catalog source %s...", catalogSource.GetName())
 		err := crc.OperatorsV1alpha1().CatalogSources(namespace).Delete(context.Background(), catalogSource.GetName(), metav1.DeleteOptions{})
 		Expect(err).ToNot(HaveOccurred())
+
+		Eventually(func() (bool, error) {
+			fetched, err := newKubeClient().KubernetesInterface().CoreV1().Pods(catalogSource.GetNamespace()).List(context.Background(), metav1.ListOptions{LabelSelector: "olm.catalogSource=" + catalogSource.GetName()})
```
timflannagan, line 510: Any thoughts on whether we should update the parent function signature and pass the kube client as a parameter vs. utilizing the helper function to get it at runtime? I'm not super I have a strong opinion, but wanted to get another opinion if we think passing it as a parameter is cleaner (or easier to maintain).

akihikokuroda: I don't have a strong opinion, either.  I like this way because it makes this PR changes small.  I can implement passing the kube client as an argument.  It needs some additional changes in this file and one more file.  What do you think?

timflannagan: Sounds reasonable to me - lets hold off.

/lgtm

awgreene: I think we should fix this holistically across the test suite, I've created [this issue](https://github.com/operator-framework/operator-lifecycle-manager/issues/2570) to track the work.

awgreene (approved):
Great work.
/approve

timflannagan:
@akihikokuroda Can you rebase this when you get a chance? I think this is ready to go but I don't want to tag this PR with a lgtm and have the openshift-bot spam retest commands in the meantime.

timflannagan (approved):
Thanks!

/lgtm

PR #2447, opened 2021-11-12
(catsrc) set status reason/message on incorrect polling interval

anik120:
This PR sets the status reason as InvalidIntervalError and a status message
if updateStrategy.RegistryPoll.ParsingError is set for the catsrc.

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -727,10 +727,19 @@ func (o *Operator) syncRegistryServer(logger *logrus.Entry, in *v1alpha1.Catalog
 
 	// requeue the catalog sync based on the polling interval, for accurate syncs of catalogs with polling enabled
 	if out.Spec.UpdateStrategy != nil {
-		logger.Debugf("requeuing registry server sync based on polling interval %s", out.Spec.UpdateStrategy.Interval.Duration.String())
-		resyncPeriod := reconciler.SyncRegistryUpdateInterval(out, time.Now())
-		o.catsrcQueueSet.RequeueAfter(out.GetNamespace(), out.GetName(), queueinformer.ResyncWithJitter(resyncPeriod, 0.1)())
-		return
+		if out.Spec.UpdateStrategy.RegistryPoll != nil {
+			if out.Spec.UpdateStrategy.RegistryPoll.ParsingError != "" && out.Status.Reason != v1alpha1.CatalogSourceIntervalInvalidError {
+				out.SetError(v1alpha1.CatalogSourceIntervalInvalidError, fmt.Errorf(out.Spec.UpdateStrategy.RegistryPoll.ParsingError))
+				if _, err := o.client.OperatorsV1alpha1().CatalogSources(out.GetNamespace()).UpdateStatus(context.TODO(), out, metav1.UpdateOptions{}); err != nil {
+					logger.Errorf("error while updating catalogsource status for an invalid interval - %v", err)
+					return
+				}
+			}
+			logger.Debugf("requeuing registry server sync based on polling interval %s", out.Spec.UpdateStrategy.Interval.Duration.String())
+			resyncPeriod := reconciler.SyncRegistryUpdateInterval(out, time.Now())
+			o.catsrcQueueSet.RequeueAfter(out.GetNamespace(), out.GetName(), queueinformer.ResyncWithJitter(resyncPeriod, 0.1)())
+			return
+		}
 	}
 
 	if err := o.sources.Remove(sourceKey); err != nil {
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -730,10 +730,6 @@ func (o *Operator) syncRegistryServer(logger *logrus.Entry, in *v1alpha1.Catalog
 		if out.Spec.UpdateStrategy.RegistryPoll != nil {
 			if out.Spec.UpdateStrategy.RegistryPoll.ParsingError != "" && out.Status.Reason != v1alpha1.CatalogSourceIntervalInvalidError {
 				out.SetError(v1alpha1.CatalogSourceIntervalInvalidError, fmt.Errorf(out.Spec.UpdateStrategy.RegistryPoll.ParsingError))
-				if _, err := o.client.OperatorsV1alpha1().CatalogSources(out.GetNamespace()).UpdateStatus(context.TODO(), out, metav1.UpdateOptions{}); err != nil {
-					logger.Errorf("error while updating catalogsource status for an invalid interval - %v", err)
-					return
-				}
 			}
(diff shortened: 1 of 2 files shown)


pkg/controller/operators/catalog/operator.go
```diff
@@ -727,10 +727,19 @@ func (o *Operator) syncRegistryServer(logger *logrus.Entry, in *v1alpha1.Catalog
 
 	// requeue the catalog sync based on the polling interval, for accurate syncs of catalogs with polling enabled
 	if out.Spec.UpdateStrategy != nil {
-		logger.Debugf("requeuing registry server sync based on polling interval %s", out.Spec.UpdateStrategy.Interval.Duration.String())
-		resyncPeriod := reconciler.SyncRegistryUpdateInterval(out, time.Now())
-		o.catsrcQueueSet.RequeueAfter(out.GetNamespace(), out.GetName(), queueinformer.ResyncWithJitter(resyncPeriod, 0.1)())
-		return
+		if out.Spec.UpdateStrategy.RegistryPoll != nil {
+			if len(out.Spec.UpdateStrategy.RegistryPoll.ParsingError) != 0 {
```
exdx, line 731: since `ParsingError` is a string type is there a benefit to checking `len(string) != 0` versus checking the string directly via `string != ""`?

pkg/controller/operators/catalog/operator.go
```diff
@@ -727,10 +727,19 @@ func (o *Operator) syncRegistryServer(logger *logrus.Entry, in *v1alpha1.Catalog
...
+				out.SetError(v1alpha1.CatalogSourceIntervalInvalidError, fmt.Errorf(out.Spec.UpdateStrategy.RegistryPoll.ParsingError))
+				if _, err := o.client.OperatorsV1alpha1().CatalogSources(out.GetNamespace()).UpdateStatus(context.TODO(), out, metav1.UpdateOptions{}); err != nil {
+					logger.Errorf("error while setting catalogsource interval - %v", err)
```
exdx, line 734:
```suggestion
					logger.Errorf("error while updating catalogsource status for an invalid interval - %v", err)
```

pkg/controller/operators/catalog/operator.go
anik120: That's a good point @exdx . I can't remember why I'd checked for the length instead of checking for the zero value, possibly an oversight. I've updated the code to check for the zero value.

pkg/controller/operators/catalog/operator.go
```diff
@@ -727,10 +727,19 @@ func (o *Operator) syncRegistryServer(logger *logrus.Entry, in *v1alpha1.Catalog
...
 	// requeue the catalog sync based on the polling interval, for accurate syncs of catalogs with polling enabled
 	if out.Spec.UpdateStrategy != nil {
-		logger.Debugf("requeuing registry server sync based on polling interval %s", out.Spec.UpdateStrategy.Interval.Duration.String())
-		resyncPeriod := reconciler.SyncRegistryUpdateInterval(out, time.Now())
-		o.catsrcQueueSet.RequeueAfter(out.GetNamespace(), out.GetName(), queueinformer.ResyncWithJitter(resyncPeriod, 0.1)())
-		return
+		if out.Spec.UpdateStrategy.RegistryPoll != nil {
+			if out.Spec.UpdateStrategy.RegistryPoll.ParsingError != "" {
+				out.SetError(v1alpha1.CatalogSourceIntervalInvalidError, fmt.Errorf(out.Spec.UpdateStrategy.RegistryPoll.ParsingError))
+				if _, err := o.client.OperatorsV1alpha1().CatalogSources(out.GetNamespace()).UpdateStatus(context.TODO(), out, metav1.UpdateOptions{}); err != nil {
```
exdx, line 733: is this `UpdateStatus` call required every catalogsource reconciliation loop if the ParsingError is not blank? Maybe we should only do the status update once? Is there any way to check if that status is set before attempting the update call?

exdx: Maybe someone can clarify, but I'm not sure that we want to always flip the status.reason to `CatalogSourceIntervalInvalidError` from another status in the case where a parsing error was found on the catalog. This would cause us to stomp our other catalog status updates and potentially constantly update the status.

I think what we want is to provide some kind of long-lived indication that the catalog did have a invalid interval at some point (when it was created), which was changed to the default, with a message indicating what happened. WDYT about using a condition on the status instead of updating the status.Reason directly?

exdx: I see what you're saying, since we are checking the polling error in this final sequence of events, it may not necessarily be overwritten. But if some other condition earlier in the chain is met, we update the status, and then when it's resolved we will update here. This particular status (invalid polling interval) doesn't feel like a level-based error, if that makes sense. I'm still a little worried about the potential for hotlooping

Do we ever clear the `spec.UpdateStrategy.RegistryPoll.ParsingError` field? Because it feels like strange behavior to always update the status to point to an invalid catalog error, even if the polling error has been resolved (by defaulting) and the catalog is no longer invalid.  

Also, I believe we can simply set 
```
syncError = fmt.Errorf(out.Spec.UpdateStrategy.RegistryPoll.ParsingError)
out.SetError(v1alpha1.CatalogSourceIntervalInvalidError, syncError)
return
```
in the code and not have to call `UpdateStatus()` directly, the control loop will update the catalog for us

anik120: @exdx so the `spec.UpdateStrategy.RegistryPoll.ParsingError` does clear when we unmarshall the incoming object [here](https://github.com/operator-framework/api/blob/master/pkg/operators/v1alpha1/catalogsource_types.go#L144-L148). I added a test to verify that https://github.com/operator-framework/operator-lifecycle-manager/pull/2447/files#diff-65a3cf2eeb27e70487e6159bd540e1d057e3bc79bea13e4c874a147d6f147469R1195

anik120: I'm guessing if this test passes that'll take care of the hotlooping concerns too.

test/e2e/catalog_e2e_test.go
```diff
@@ -1074,76 +1074,150 @@ var _ = Describe("Catalog represents a store of bundles which OLM can use to ins
...
+			// No update to image: update pod should be deleted quickly
+			noPod := podCount(0)
+			updatePods, err = awaitPodsWithInterval(GinkgoT(), c, source.GetNamespace(), updateSelector.String(), 1*time.Second, 30*time.Second, noPod)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(updatePods.Items).To(HaveLen(0))
+		})
+
+	})
+
+	FWhen("A catalogSource is created with incorrect interval", func() {
```
exdx, line 1149: nit: focus set with F

dinhxuanvu (approved):
/approve

Issue #2439, opened 2021-11-09
Panic in produced in the "Install Plan creation with pre existing CRD owners [It]  OnePreExistingCRDOwner" e2e test


timflannagan:
## Bug Report

A panic was produced during an e2e test workflow run in https://github.com/operator-framework/operator-lifecycle-manager/pull/2438 when running the installplan-related e2e tests. That PR doesn't touch anything in the installplan testing suite.

The full panic trace logs:

```log
14:38:42.55: UpgradePending (nginx-dm7xg-stable): &ObjectReference{Kind:InstallPlan,Namespace:operators,Name:install-4cpdn,UID:1e03773e-75ac-4a05-a3aa-ce1989a03de7,APIVersion:operators.coreos.com/v1alpha1,ResourceVersion:6142,FieldPath:,}
waiting for sub-nginx-8fzfm to have installplan ref
installplan install-4cpdn is in phase Installing
installplan install-4cpdn is in phase Installing
installplan install-4cpdn is in phase Complete
Install plan install-4cpdn fetched with status Complete
E1109 14:38:46.687870   20587 runtime.go:76] Observed a panic: 
Your test failed.
Ginkgo panics to prevent subsequent assertions from running.
Normally Ginkgo rescues this panic so you shouldn't see it.

But, if you make an assertion in a goroutine, Ginkgo can't capture the panic.
To circumvent this, you should call

	defer GinkgoRecover()

at the top of the goroutine that caused this panic.

goroutine 14 [running]:
k8s.io/apimachinery/pkg/util/runtime.logPanic(0x2eb25a0, 0x368ab00)
	/home/runner/work/operator-lifecycle-manager/operator-lifecycle-manager/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:74 +0x95
k8s.io/apimachinery/pkg/util/runtime.HandleCrash(0x0, 0x0, 0x0)
	/home/runner/work/operator-lifecycle-manager/operator-lifecycle-manager/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:48 +0x86
panic(0x2eb25a0, 0x368ab00)
	/opt/hostedtoolcache/go/1.16.10/x64/src/runtime/panic.go:965 +0x1b9
github.com/onsi/ginkgo.Fail(0xc0007d7900, 0xc08, 0xc001f651e0, 0x1, 0x1)
[...]

PR #2448, opened 2021-11-15
remove panic

akihikokuroda:

There are places that cause 'panic' when the test fails.  This PR changes to report the test failure instead of panic when the test fails.

This fixes the panic of https://github.com/operator-framework/operator-lifecycle-manager/issues/2439.  It doesn't fix the test failure.

**Reviewer Checklist**

### DIFF
--- a/test/e2e/csv_e2e_test.go
+++ b/test/e2e/csv_e2e_test.go
@@ -16,10 +16,12 @@ import (
 	rbacv1 "k8s.io/api/rbac/v1"
 	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	k8slabels "k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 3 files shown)


test/e2e/csv_e2e_test.go
```diff
@@ -1551,7 +1553,7 @@ var _ = Describe("ClusterServiceVersion", func() {
 		// Fetch cluster service version again to check for unnecessary control loops
 		sameCSV, err := fetchCSV(crc, csv.Name, testNamespace, csvSucceededChecker)
 		Expect(err).ShouldNot(HaveOccurred())
-		compareResources(GinkgoT(), fetchedCSV, sameCSV)
+		Expect(equality.Semantic.DeepEqual(fetchedCSV, sameCSV)).Should(BeTrue(), "CSV is same: "  + diff.ObjectDiff(fetchedCSV, sameCSV))
```
timflannagan, line 1556: I think I'm largely fine with deflating the compareResources logic at the individual call sites but I can't wrap my head around what that `"CSV is same: "  + diff.ObjectDiff(fetchedCSV, sameCSV)` would look like when the assertion has failed. Do we need that prefixed "CSV is same: ..." or can we just use `diff.ObjectDiff(...)` to provide debuggability when the assertion has failed?

timflannagan: It looks like this is also failing the sanity check too.

akihikokuroda: OK.  I change to just the diff there.  Thanks!

perdasilva (approved):
lgtm

dinhxuanvu (approved):
/approve

PR #2460, opened 2021-11-18
test/e2e: Collect CatalogSource testing artifacts

timflannagan:
Update the test/e2e/collect-ci-artifacts.sh bash script and extend the
list of resources gathered to also create a snapshot of CatalogSource
resources in the test namespace.

**Reviewer Checklist**

### DIFF
--- a/test/e2e/collect-ci-artifacts.sh
+++ b/test/e2e/collect-ci-artifacts.sh
@@ -11,6 +11,7 @@ set -o errexit
 mkdir -p "${TEST_ARTIFACTS_DIR}"
 
 commands=()
+commands+=("get catalogsources -o yaml")
 commands+=("get subscriptions -o yaml")
 commands+=("get operatorgroups -o yaml")
 commands+=("get clusterserviceversions -o yaml")


test/e2e/collect-ci-artifacts.sh
```diff
@@ -24,3 +25,8 @@ for command in "${commands[@]}"; do
     COMMAND_OUTPUT_FILE=${TEST_ARTIFACTS_DIR}/${command// /_}
     kubectl -n ${TEST_NAMESPACE} ${command} >> "${COMMAND_OUTPUT_FILE}"
 done
+
+echo "Deleting any empty files in the ${TEST_ARTIFACTS_DIR} directory"
+# Note: delete any files less than 83 kilobytes as this should account for
+# any empty files, and files that contain zero length arrays of `kind: List`.
+find "${TEST_ARTIFACTS_DIR}" -type f -size -83k -delete
```
timflannagan, line 32: Hmm I realize we should probably torch empty directories as well. It looks like find also has support for identifying empty directories so it should just be another command.

test/e2e/collect-ci-artifacts.sh
```diff
@@ -11,6 +11,7 @@ set -o errexit
 mkdir -p "${TEST_ARTIFACTS_DIR}"
 
 commands=()
+commands+=("get catalogsources -o yaml")
```
njhale, line 14: nit: if the string doesn't need to be interpolated, use single quotes.

njhale (approved):
/approve

PR #2466, opened 2021-11-19
Allow users to toggle off csv copying for AllNamespace mode operators

awgreene:


pkg/controller/operators/olm/config.go
```diff
@@ -112,6 +113,12 @@ func WithLogger(logger *logrus.Logger) OperatorOption {
 	}
 }
 
+func WithCopiedCSVs(enabled bool) OperatorOption {
```
perdasilva, line 116: suggestion: wdyt if we change it to just:

```go
func WithCopiedCSVs() {...}
```

meaning it is disabled by default, but you can enable it with this option?

pkg/controller/operators/olm/config.go
```diff
@@ -33,6 +33,7 @@ type operatorConfig struct {
 	apiLabeler        labeler.Labeler
 	restConfig        *rest.Config
 	configClient      configv1client.Interface
+	copiedCSVs        bool
```
perdasilva, line 36: suggestion: its sometimes helpful to make boolean variables clear that they are boolean in their naming. E.g. by adding the `is` prefix. Wdyt of, instead of `copiedCSVs` we call it `isCopiedCSVsEnabled` or something like that? `copiedCSVs` reads a bit like it should be a slice.

pkg/controller/operators/olm/config.go
perdasilva: Alternatively, you could call it `WithCopiedCSVsEnabled(enabled bool)` or some such, tying in with the previous comment on naming

perdasilva: OR, also consider 

```go
func WithoutCopiedCSVs() {}
```

Depending on your defaults...

pkg/controller/operators/olm/operator.go
```diff
@@ -1241,14 +1249,112 @@ func (a *Operator) syncCopyCSV(obj interface{}) (syncError error) {
 	}).Debug("copying csv to targets")
 
 	// Check if we need to do any copying / annotation for the operatorgroup
-	if err := a.ensureCSVsInNamespaces(clusterServiceVersion, operatorGroup, NewNamespaceSet(operatorGroup.Status.Namespaces)); err != nil {
-		logger.WithError(err).Info("couldn't copy CSV to target namespaces")
-		syncError = err
+	namespaceSet := NewNamespaceSet(operatorGroup.Status.Namespaces)
+	if a.copiedCSVs || !namespaceSet.IsAllNamespaces() {
```
tylerslaton, line 1253: nit: +1 to Per's comment about renaming `copiedCSVs`. This line would read more clearly given that change.

pkg/controller/operators/olm/operator.go
```diff
@@ -1241,14 +1249,112 @@ func (a *Operator) syncCopyCSV(obj interface{}) (syncError error) {
...
+				return err
+			}
+		}
+
+		if err := a.createCSVCopyingDisabledEvent(clusterServiceVersion); err != nil {
+			return err
+		}
 	}
 
 	return
```
tylerslaton, line 1289: nit: Is the `else` needed here? We could just return early in the `if` statement. Thoughts?

pkg/controller/operators/olm/operator.go
```diff
@@ -1241,14 +1249,112 @@ func (a *Operator) syncCopyCSV(obj interface{}) (syncError error) {
...
+		return result, nil
+	}
+
+	events, err := a.opClient.KubernetesInterface().CoreV1().Events(csv.GetNamespace()).List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	for _, event := range events.Items {
+		a.logger.Infof("For event %v\n", event)
```
tylerslaton, line 1304: nit: Is there extra information for this log message we could include here? It may be difficult understanding what it conveys in a production system.

pkg/controller/operators/olm/operator.go
```diff
@@ -1241,14 +1249,112 @@ func (a *Operator) syncCopyCSV(obj interface{}) (syncError error) {
...
+	}
+
+	// Remove existing events.
+	if len(events) > 1 {
+		if err := a.deleteEvents(events); err != nil {
+			return err
+		}
+	}
+
+	go a.recorder.Eventf(csv, corev1.EventTypeWarning, copiedCSVsDisabledReason, "CSV copying disabled for %s/%s", csv.GetNamespace(), csv.GetName())
```
tylerslaton, line 1353: nit: Since this is an async call, it may make sense to add some comment around what it is trying to accomplish for  future review of this code.

tylerslaton (commented):
Looks great! Just a couple of nits.

pkg/controller/operators/olm/operator.go
awgreene: oh haha this is from some debugging I had done, I'll be removing it.

pkg/controller/operators/olm/operator.go
awgreene: SGTM!

pkg/controller/operators/olm/operator.go
awgreene: I've decided to remove the routine in a more recent copy of the code. I'll push it shortly.

pkg/controller/operators/olm/config.go
awgreene: I like the look of the signature, but it wouldn't play as nice with [this code](https://github.com/operator-framework/operator-lifecycle-manager/blob/eabd986fefe848d15c877988a7c4dea5924ae8eb/cmd/olm/main.go#L155-L165) a bit less readable. We would need to create the list of config options before the NewOperator call which isn't as clean IMO. Thoughts?

perdasilva: Yeah, that's fair enough.

cmd/olm/main.go
```diff
@@ -69,6 +69,9 @@ var (
 
 	namespace = pflag.String(
 		"namespace", "", "namespace where cleanup runs")
+
+	disableCopiedCSVs = pflag.Bool(
+		"disableCopiedCSVs", false, "disables copied CSVs in AllNamespace mode when enabled")
```
timflannagan, line 74: It looks like there's already precedence for camel-cased flag names poking around the flag definitions listed above - my preference is certainty towards [unix-style](http://www.catb.org/~esr/writings/taoup/html/ch10s05.html), e.g. --disable-coped-csvs. Any thoughts on that style vs. continuing to use this existing style?

pkg/controller/operators/olm/operator.go
```diff
@@ -30,6 +31,7 @@ import (
 	kagg "k8s.io/kube-aggregator/pkg/client/informers/externalversions"
 
 	"github.com/operator-framework/api/pkg/operators/v1alpha1"
+	operatorsv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
```
timflannagan, line 34: nit: duplicate package import

cmd/olm/main.go
timflannagan: A couple of more questions I had:

- How would this CLI flag interact with the OLMConfig API?
- Any thoughts on a feature gating implementation vs. a CLI flag?

pkg/controller/operators/olm/operator.go
```diff
@@ -86,6 +92,7 @@ type Operator struct {
 	clientAttenuator      *scoped.ClientAttenuator
 	serviceAccountQuerier *scoped.UserDefinedServiceAccountQuerier
 	clientFactory         clients.Factory
+	isCopiedCSVsEnabled   bool
```
timflannagan, line 95: nit: would areCopiedCSVsEnabled read better?

pkg/controller/operators/olm/operator.go
```diff
@@ -141,6 +148,7 @@ func newOperatorWithConfig(ctx context.Context, config *operatorConfig) (*Operat
 		clientAttenuator:      scoped.NewClientAttenuator(config.logger, config.restConfig, config.operatorClient),
 		serviceAccountQuerier: scoped.NewUserDefinedServiceAccountQuerier(config.logger, config.externalClient),
 		clientFactory:         clients.NewFactory(config.restConfig),
+		isCopiedCSVsEnabled:   config.isCopiedCSVsEnabled,
```
timflannagan, line 151: One other consideration for feature flagging vs. a CLI toggle is that we might be able to forgo expanding the config structure further, and instead just check whether that feature gate has been enabled at runtime.

pkg/controller/operators/olm/operator.go
tylerslaton: What about hasCopiedCSVsEnabled? Then it would read:

`if a.hasCopiedCSVsEnabled {`

timflannagan: `if a.copiedCSVsAreEnabled {...}`?

awgreene:
Relies on https://github.com/operator-framework/api/pull/198

cmd/olm/main.go
awgreene: > How would this CLI flag interact with the OLMConfig API?
This behavior is defined in the enhancement and warrented [some discussion](https://github.com/operator-framework/enhancements/pull/92/files#r750593846). Personally, I would like to remove the flag. Thoughts @joelanford @njhale @joelanford @timflannagan?

awgreene: >Any thoughts on a feature gating implementation vs. a CLI flag?

As of the latests version of this PR, this can be configured via the [olmConfig CRD](https://github.com/operator-framework/api/pull/198).

pkg/controller/operators/olm/operator.go
awgreene: Good eye.

cmd/olm/main.go
awgreene: >It looks like there's already precedence for camel-cased flag names poking around the flag definitions listed above - my preference is certainty towards unix-style, e.g. --disable-coped-csvs. Any thoughts on that style vs. continuing to use this existing style?

I prefer the unix style as well, but wanted to remain consistent within the project. If we drop support for the flag this is no longer an issue.

pkg/controller/operators/olm/operator.go
```diff
@@ -1240,15 +1248,147 @@ func (a *Operator) syncCopyCSV(obj interface{}) (syncError error) {
...
+// will be returned.
+//
+// If the "cluster" olmConfig resource is found, the value of
+// olmConfig.spec.features.disableCopiedCSVs will be returned along with a
+// nil error.
+//
+// If the "cluster" olmConfig resource is not found, the value will match
+// the value of the disableCopiedCSVs flag provided to OLM on startup will
+// be returned along with a nil error.
+func (a *Operator) isCSVCopyingEnabled() (bool, error) {
```
awgreene, line 1311: If we remove the flag I'll be changing the name of this func.

cmd/olm/main.go
timflannagan: I'm largely a fan of feature flagging but if we have another API that can drive those decisions, it's just extra complexity at the end of the day.

joelanford: I'd be okay with either of these:
1. change the name to `defaultDisableCopiedCSVs`, which implies it can/will be overridden by the OLMConfig object if set there.
2. just remove the flag

The only argument I have against (2) is that if you're a user deploying OLM with the intention of enabling the flag, it's less work to do it as a CLI flag. This probably really only applies to OLM developers, so maybe the value proposition isn't there.

Also, fwiw, I prefer unix-style, but I prefer consistency more. If we want to change to unix-style, we should deprecate all of the camelCase flags and add the unix-style variants, IMO.

awgreene: > The only argument I have against (2) is that if you're a user deploying OLM with the intention of enabling the flag, it's less work to do it as a CLI flag. This probably really only applies to OLM developers, so maybe the value proposition isn't there.

Although removing the flag adds a bit of a burden on the developer, there are benefits to having a single point of entry for toggling this feature:
1. It ensures that developers and users rely on the same workflow.
2. The flag can always be added later if there is demand for it. Removing the flag once released is much more difficult.
3. It isn't that hard to enable the feature via the API:
```
$ kubectl create -f -<<EOF
kind: OLMConfig
metadata:
  name: cluster
spec:
  features:
    disableCopiedCSVs: true
EOF
```

pkg/controller/operators/olm/operator.go
awgreene: I like `copiedCSVsAreEnabled`

cmd/olm/main.go
timflannagan: @awgreene If we're going to run down this direction of having the API drive all runtime configurations, do we need to also update the EP? How would users enable this configuration in downstream environments?

pkg/controller/operators/olm/operator.go
```diff
@@ -257,6 +257,27 @@ func newOperatorWithConfig(ctx context.Context, config *operatorConfig) (*Operat
...
+			queueinformer.WithQueue(olmConfigQueue),
+			queueinformer.WithIndexer(olmConfigInformer.GetIndexer()),
+			queueinformer.WithSyncer(queueinformer.LegacySyncHandler(op.syncOLMConfig).ToSyncer()),
+		)
+		if err != nil {
+			return nil, err
+		}
+		if err := op.RegisterQueueInformer(olmConfigQueueInformer); err != nil {
+			return nil, err
+		}
```
awgreene: The OLMConfig controller needs to requeue csvs, which is done by adding an object to [this queue](https://github.com/operator-framework/operator-lifecycle-manager/blob/e6cc305d49e865020b0d3481b621bf7412fb871a/pkg/controller/operators/olm/operator.go#L71). We could build and pass this queue into OLM if we wanted the olmConfig controller to be built with controller-runtime. Thoughts?

pkg/controller/operators/olm/operator.go
```diff
@@ -1212,6 +1240,99 @@ func (a *Operator) syncClusterServiceVersion(obj interface{}) (syncError error)
...
+		Type:               "ready",
+		LastTransitionTime: metav1.Now(),
+	}
+
+	// If a CSV was requeued, reflect that the cluster is not yet in the expect state.
+	if csvIsRequeued {
+		condition.Message = "At least one CSV had an unexpected number of copied CSVs"
+		condition.Status = metav1.ConditionFalse
+		defer func() {
+			a.olmConfigQueue.AddAfter(olmConfig, time.Second*5)
```
awgreene, line 1320: The Delay was not correctly incrementing the number of failures and would spam the API server. Planning to look at this tomorrow.

pkg/controller/operators/olm/operator.go
```diff
@@ -252,6 +260,26 @@ func newOperatorWithConfig(ctx context.Context, config *operatorConfig) (*Operat
...
+			queueinformer.WithIndexer(olmConfigInformer.GetIndexer()),
+			queueinformer.WithSyncer(queueinformer.LegacySyncHandler(op.syncOLMConfig).ToSyncer()),
+		)
+		if err != nil {
+			return nil, err
+		}
+		if err := op.RegisterQueueInformer(olmConfigQueueInformer); err != nil {
+			return nil, err
+		}
+
```
awgreene, line 282: The OLMConfig controller needs to requeue csvs, which is done by adding an object to this queue. We could build and pass this queue into OLM if we wanted the olmConfig controller to be built with controller-runtime. Thoughts?

awgreene: One benefit to the existing implementation is that the [logic](https://github.com/operator-framework/operator-lifecycle-manager/pull/2466/files#diff-1c45679686ec7da9a4e3c02d009d84f465130fec2278b9512d27d53e23224f3aR1364) responsible for deleting the CSVs is in the same file.

One negative to the existing implementation is that the olmConfig controller may be extended to support other features which might be easier to implement / understand if written with controller runtime.

pkg/controller/operators/olm/operator.go
awgreene: Could also be done by triggering a watch on copied csvs.

pkg/controller/operators/olm/operator.go
```diff
@@ -252,6 +260,26 @@ func newOperatorWithConfig(ctx context.Context, config *operatorConfig) (*Operat
 			return nil, err
 		}
 
+		// Register QueueInformer for olmConfig
+		olmConfigInformer := externalversions.NewSharedInformerFactoryWithOptions(
+			op.client,
+			config.resyncPeriod(),
+		).Operators().V1().OLMConfigs().Informer()
```
timflannagan, line 267: Is the intention to have the OLMConfig be immediately registered as a stable v1 API?

pkg/controller/operators/olm/operator.go
```diff
@@ -1240,15 +1361,146 @@ func (a *Operator) syncCopyCSV(obj interface{}) (syncError error) {
...
+// nil error.
+//
+// If the "cluster" olmConfig resource is not found, true will be returned
+// without an error.
+func (a *Operator) copiedCSVsAreEnabled() (bool, error) {
+	olmConfig, err := a.client.OperatorsV1().OLMConfigs().Get(context.TODO(), "cluster", metav1.GetOptions{})
+	// If there was an error that wasn't an IsNotFound, return the error
+	if err != nil && !k8serrors.IsNotFound(err) {
+		return false, err
+	}
```
timflannagan, line 1428: How are we handling the edge case where the GET request for this cluster singleton returns an apierrors.IsNotFound(err)? Default to true?

timflannagan: Disregard, commented too quick: `// Default to true if olmConfig cannot be found` - my only suggestion would be to move up this conditional check further up:

```go
	if err != nil && !k8serrors.IsNotFound(err) {
		return false, err
	}
        if k8serrors.IsNotFound(err) {
                return true, nil
        }
        return olmConfig.CopiedCSVsAreEnabled()
}
```

I think this is more readable personally, but this is likely a minor implementation so feel free to ignore this entirely.

go.mod
```diff
@@ -33,35 +33,36 @@ require (
...
+	sigs.k8s.io/controller-runtime v0.10.0
+	sigs.k8s.io/controller-tools v0.6.2
 	sigs.k8s.io/kind v0.11.1
 )
 
 replace (
 	// controller runtime
 	github.com/openshift/api => github.com/openshift/api v0.0.0-20200331152225-585af27e34fd // release-4.5
 	github.com/openshift/client-go => github.com/openshift/client-go v0.0.0-20200326155132-2a6cd50aedd0 // release-4.5
+	github.com/operator-framework/api => ../api //TODO: Remove
```
timflannagan, line 65: You could also just cut a release in your API fork and point to that release here if you don't think there will be a ton of iteration left on the OLMConfig API implementation.

cmd/olm/main.go
awgreene: >@awgreene If we're going to run down this direction of having the API drive all runtime configurations, do we need to also update the EP? 

Yes @timflannagan , I have a [WIP PR](https://github.com/operator-framework/enhancements/pull/103) for the enhancement and will capture these changes there

>How would users enable this configuration in downstream environments?

We will add the cluster olmConfig to the `/manifests` dir and add the `release.openshift.io/create-only: "true"` annotation, allowing users to enable/disable the feature.

pkg/controller/operators/olm/operator.go
awgreene: Flagging is not a viable solution to customers that want to disable olm csv copying as users will need to mark OLM as unmanged by CVO, thereby placing their cluster in an unmanaged/unupgradeable state.

cmd/olm/main.go
timflannagan: > We will add the cluster olmConfig to the /manifests dir with the release.openshift.io/create-only: "true" annotation, allowing users to enable/disable the feature.

+1 that's exactly what I had in mind as well.

pkg/controller/operators/olm/operator.go
awgreene:
```
func (a *Operator) copiedCSVsAreEnabled() (bool, error) {
	olmConfig, err := a.client.OperatorsV1().OLMConfigs().Get(context.TODO(), "cluster", metav1.GetOptions{})
	// If there was an error that wasn't an IsNotFound, return the error
	if err != nil {
		if !k8serrors.IsNotFound(err) {
			return false, err
		}
		// Default to true if olmConfig singleton cannot be found
		return true, nil
	}

	// If there was no error, return value based on olmConfig singleton
	return olmConfig.CopiedCSVsAreEnabled(), nil
}
```

awgreene: I like the suggestion, modifying the provided code slightly:
```
	olmConfig, err := a.client.OperatorsV1().OLMConfigs().Get(context.TODO(), "cluster", metav1.GetOptions{})
	if err != nil {
		// If there was an error that wasn't an IsNotFound, return the error
		if !k8serrors.IsNotFound(err) {
			return false, err
		}
		// Default to true if olmConfig singleton cannot be found
		return true, nil
	}

	// If there was no error, return value based on olmConfig singleton
	return olmConfig.CopiedCSVsAreEnabled(), nil
```

awgreene: Actually, this reads better:
```go
olmConfig, err := a.client.OperatorsV1().OLMConfigs().Get(context.TODO(), "cluster", metav1.GetOptions{})
	if err != nil {
		// Default to true if olmConfig singleton cannot be found
		if k8serrors.IsNotFound(err) {
			return true, nil
		}
		// If there was an error that wasn't an IsNotFound, return the error
		return false, err
	}
```

awgreene: Resolved

go.mod
awgreene: That's a great idea, done.

awgreene: Will remove this once the [api PR](https://github.com/operator-framework/api/pull/198) merges.

pkg/controller/operators/olm/operator.go
```diff
@@ -1211,6 +1239,99 @@ func (a *Operator) syncClusterServiceVersion(obj interface{}) (syncError error)
...
+			return err
+		}
+		csvIsRequeued = true
+	}
+	condition := metav1.Condition{
+		Reason:             "CopiedCSVCountCorrect",
+		Status:             metav1.ConditionTrue,
+		Message:            "Correct Number of copiedCSVs created",
+		ObservedGeneration: olmConfig.GetGeneration(),
+		Type:               "ready",
```
njhale, line 1310: I believe this is supposed to be the PascalCase ID of the condition; unique and descriptive.

pkg/controller/operators/olm/operator.go
```diff
@@ -1211,6 +1239,99 @@ func (a *Operator) syncClusterServiceVersion(obj interface{}) (syncError error)
...
+			return err
+		}
+
+		// If the correct number of copied CSVs were found, continue
+		if len(copiedCSVs) == 0 == olmConfig.Spec.Features.DisableCopiedCSVs {
+			continue
+		}
+
+		// There were an incorrect number of copied CSVs, requeue the original CSV.
+		if err := a.csvQueueSet.Requeue(csv.GetNamespace(), csv.GetName()); err != nil {
```
njhale, line 1300: Would it be easier if this sync function directly deletes copied CSVs when the feature is disabled?

njhale (changes requested):

go.mod
njhale: FYI, you can also reference `@digest` which will be converted to a pseudoversion automatically

pkg/controller/operators/olm/operator.go
```diff
@@ -1211,6 +1239,99 @@ func (a *Operator) syncClusterServiceVersion(obj interface{}) (syncError error)
...
+			return err
+		}
+
+		copiedCSVs, err := a.copiedCSVLister.List(labels.NewSelector().Add(*requirement))
+		if err != nil {
+			return err
+		}
+
+		// If the correct number of copied CSVs were found, continue
+		if len(copiedCSVs) == 0 == olmConfig.Spec.Features.DisableCopiedCSVs {
```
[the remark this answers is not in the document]
awgreene: That doesn't handle the case where len(copiedCSVs) is not zero and disabled copied CSVs is enabled.

deploy/chart/crds/0000_50_olm_00-olmconfigs.crd.yaml
```diff
@@ -0,0 +1,97 @@
...
+                      type: boolean
+                      default: false
+            status:
+              description: OLMConfigStatus is the status for an OLMConfig resource.
+              type: object
+              properties:
+                conditions:
+                  type: array
+                  items:
+                    description: "Condition contains details for one aspect of the current state of this API Resource. --- This struct is intended for direct use as an array at the field path .status.conditions.  For example, type FooStatus struct{     // Represents the observations of a foo's current state.     // Known .status.conditions.type are: \"Available\", \"Progressing\", and \"Degraded\"     // +patchMergeKey=type     // +patchStrategy=merge     // +listType=map     // +listMapKey=type     Conditions []metav1.Condition `json:\"conditions,omitempty\" patchStrategy:\"merge\" patchMergeKey:\"type\" protobuf:\"bytes,1,rep,name=conditions\"` \n     // other fields }"
```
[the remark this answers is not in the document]
awgreene: Is there a way to override the description of a structure not owned by your package?

pkg/controller/operators/olm/operator.go
```diff
@@ -51,6 +53,10 @@ import (
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/metrics"
 )
 
+const (
+	copiedCSVsDisabledReason = "CopiedCSVsDisabled"
```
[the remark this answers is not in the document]
awgreene: Declared it in the api repo, PR has been updated to reflect these changes.

pkg/controller/operators/olm/operator.go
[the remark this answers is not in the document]
njhale: Ah I see. okay

pkg/controller/operators/olm/operator.go
njhale: spoke about this offline -- was convinced that letting the CSV controller handle this is more clean

pkg/controller/operators/olm/operator.go
```diff
@@ -124,6 +127,7 @@ func newOperatorWithConfig(ctx context.Context, config *operatorConfig) (*Operat
 		client:                config.externalClient,
 		ogQueueSet:            queueinformer.NewEmptyResourceQueueSet(),
 		csvQueueSet:           queueinformer.NewEmptyResourceQueueSet(),
+		olmConfigQueue:        workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "olmConfig"),
```
timflannagan, line 130: What was the rationale again for this queue implementation vs. using the queueinformer?

test/e2e/csv_e2e_test.go
```diff
@@ -18,9 +18,13 @@ import (
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	k8slabels "k8s.io/apimachinery/pkg/labels"
```
timflannagan, line 24: nit: duplicate package import

test/e2e/csv_e2e_test.go
```diff
@@ -29,6 +33,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	operatorsv1 "github.com/operator-framework/api/pkg/operators/v1"
+	v1 "github.com/operator-framework/api/pkg/operators/v1"
+	"github.com/operator-framework/api/pkg/operators/v1alpha1"
 	operatorsv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
```
timflannagan, line 38: nit: duplicate package import(s)

test/e2e/csv_e2e_test.go
```diff
@@ -4169,6 +4175,210 @@ var _ = Describe("ClusterServiceVersion", func() {
...
+			},
+		}
+
+		// Create a new operator group for the new namespace
+		operatorGroup := v1.OperatorGroup{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      genName("csv-toggle-test-"),
+				Namespace: ns.GetName(),
+			},
+		}
```
timflannagan, line 4192: We could also use the `SetupGeneratedTestNamespace` function to create a testing namespace and an OperatorGroup resource that targets that namespace:

```go
var (
		generatedNamespace corev1.Namespace
)
BeforeEach(func() {
		generatedNamespace = SetupGeneratedTestNamespace(genName("csv-toggle-test-"))
}
AfterEach(func() {
		TeardownNamespace(generatedNamespace.GetName())
}
```

test/e2e/csv_e2e_test.go
```diff
@@ -4169,6 +4175,210 @@ var _ = Describe("ClusterServiceVersion", func() {
 
 		Expect(apiService2.Spec.CABundle).Should(Equal(apiService1.Spec.CABundle))
 	})
+
+	It("Disabling copied csvs", func() {
```
timflannagan, line 4179: nit: We may want to utilize something like a When/Context/Describe type container so we can utilize BeforeEach/AfterEach stanzas vs. nesting Context blocks.

test/e2e/csv_e2e_test.go
```diff
@@ -4169,6 +4175,210 @@ var _ = Describe("ClusterServiceVersion", func() {
...
+				InstallStrategy: newNginxInstallStrategy(genName("csv-toggle-test-"), nil, nil),
+				InstallModes: []v1alpha1.InstallMode{
+					{
+						Type:      v1alpha1.InstallModeTypeAllNamespaces,
+						Supported: true,
+					},
+				},
+			},
+		}
+		Context("When an operator is installed in all namespace mode", func() {
```
timflannagan, line 4209: Any idea on whether we could (or should) deflate this Context to the top-level, e.g. `Describe("CSV copying is disabled")` or `Context("when CSV copying is disabled")` and update this to be a `When("an operator is installed in all namespace mode")`?

pkg/controller/operators/olm/operator.go
```diff
@@ -1211,6 +1235,114 @@ func (a *Operator) syncClusterServiceVersion(obj interface{}) (syncError error)
...
+		LastTransitionTime: metav1.Now(),
+		Status:             metav1.ConditionFalse,
+	}
+	if !isDisabled {
+		condition.Reason = "CopiedCSVsEnabled"
+		condition.Message = "Copied CSVs are enabled"
+		if csvIsRequeued {
+			condition.Message = "No copied CSVs were founds for an operator installed in AllNamespace mode"
+		}
+	} else {
```
timflannagan, line 1332: nit: it looks like we could avoid the if/else conditional structure by returning the `condition` variable during the if clause and exiting early.

pkg/controller/operators/olm/operator.go
timflannagan: Reminder: we may want to create a follow-up documentation ticket to documenting how to disable copied CSVs for https://olm.operatorframework.io/ that can be tackled after feature freeze.

pkg/controller/operators/olm/operator.go
awgreene: >  or due to a bad actor (e.g. me).

I don't know why your stage performance should influence this PR, but I agree that this controller should watch/sync on copied csvs.

go.mod
```diff
@@ -33,35 +33,36 @@ require (
...
+	sigs.k8s.io/controller-tools v0.6.2
 	sigs.k8s.io/kind v0.11.1
 )
 
 replace (
 	// controller runtime
-	github.com/openshift/api => github.com/openshift/api v0.0.0-20200331152225-585af27e34fd // release-4.5
+	github.com/openshift/api => github.com/openshift/api v0.0.0-20211014063134-be2a7fb8aa44
 	github.com/openshift/client-go => github.com/openshift/client-go v0.0.0-20200326155132-2a6cd50aedd0 // release-4.5
+	github.com/operator-framework/api => github.com/awgreene/api v0.0.1 //TODO: Remove //TODO: Remove
```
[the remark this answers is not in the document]
awgreene: Removed and dumped.

pkg/controller/operators/olm/operator.go
awgreene: Agreed.

test/e2e/csv_e2e_test.go
awgreene: Removed. Thanks for catching this!

pkg/controller/operators/olm/operator.go
awgreene: Addressed in the latest version of the PR.

pkg/controller/operators/olm/operator.go
awgreene: I believe that you're asking why this field is not set the same way as the others using the `queueinformer.NewEmptyResourceQueueSet()` function.

The QueueSets eventually use the same `workqueue.NewNamedRateLimitingQueue` func to add the queue to the set ([example](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/controller/operators/olm/operator.go#L158-L159). OLM only ever cares about the clusterScoped OLMConfig queue so I didn't see the need to include a set.

go.mod
awgreene: Removed now that https://github.com/operator-framework/operator-lifecycle-manager/pull/2515 has merged.

pkg/controller/operators/olm/operator.go
```diff
@@ -1211,6 +1239,99 @@ func (a *Operator) syncClusterServiceVersion(obj interface{}) (syncError error)
...
+		if len(copiedCSVs) == 0 == olmConfig.Spec.Features.DisableCopiedCSVs {
+			continue
+		}
+
+		// There were an incorrect number of copied CSVs, requeue the original CSV.
+		if err := a.csvQueueSet.Requeue(csv.GetNamespace(), csv.GetName()); err != nil {
+			return err
+		}
+		csvIsRequeued = true
+	}
```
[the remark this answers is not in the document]
awgreene: Good idea. Process is now:

1. Get a list of AllNamespaced OperatorGroups
2. For each AllNamespaced OperatorGroup, get a list of copied CSVs from that operator group
3. If the expected number of copied CSVs are found, continue to the next operatorgroup. If not, requeue each of the CSVs that have an unexpected number of copied CSVs.

test/e2e/csv_e2e_test.go
awgreene: Yep, we'd just need to update that method to allow users to provide the set of targetNamespaces.

test/e2e/csv_e2e_test.go
awgreene: I'm going to give this an attempt now.

pkg/controller/operators/olm/operator.go
awgreene: I perused  moving this reconcile logic into controller-runtime, but all of OLM's controller-runtime operators are built with [a client that ignores copied csvs](https://github.com/operator-framework/operator-lifecycle-manager/blob/82fbb1aa0c391a2197e9a665759c3f94764850ab/cmd/olm/manager.go#L58-L60). In my opinion, the performance benefit of not having these controllers caching copied csvs outweighs those of moving towards controller-runtime. As an added benefit, we avoid introducing new controllers that really on CSVs and keep said logic within the olm controller.

pkg/controller/operators/olm/operator.go
awgreene: Following up here, I attempted to use a set because I am not a fan of getting the olmConfig resource to requeue it within the [SyncCopyCSV](https://github.com/operator-framework/operator-lifecycle-manager/pull/2466/files#diff-1c45679686ec7da9a4e3c02d009d84f465130fec2278b9512d27d53e23224f3aR1366) function, but the set didn't seem to actually requeue the olmConfig resource. We should probably look into this in the future.

pkg/controller/operators/olm/operator.go
```diff
@@ -1211,6 +1239,99 @@ func (a *Operator) syncClusterServiceVersion(obj interface{}) (syncError error)
...
+}
+
+func (a *Operator) syncOLMConfig(obj interface{}) (syncError error) {
+	a.logger.Info("Processing olmConfig")
+	olmConfig, ok := obj.(*v1.OLMConfig)
+	if !ok {
+		return fmt.Errorf("casting OLMConfig failed")
+	}
+
+	// Generate a map to track namespaces that are cluster scoped
```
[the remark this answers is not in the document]
awgreene: Done.

test/e2e/csv_e2e_test.go
awgreene: e2e has been updated to better align with Behavior-Driven Development  style.

test/e2e/csv_e2e_test.go
awgreene: e2e has been updated to better align with Behavior-Driven Development style.

test/e2e/csv_e2e_test.go
```diff
@@ -4171,6 +4174,252 @@ var _ = Describe("ClusterServiceVersion", func() {
...
+				}
+
+				// Exit early if copied CSVs are disabled.
+				if !olmConfig.CopiedCSVsAreEnabled() {
+					return nil
+				}
+
+				olmConfig.Spec = operatorsv1.OLMConfigSpec{
+					Features: &operatorsv1.Features{
+						DisableCopiedCSVs: getPointer(true),
```
njhale, line 4275: If the number of occurrences is few, you could also invoke an anonymous function:

```go
DisableCopiedCSVs: func() *bool {
    b := true
    return &b
}(),
```

njhale (approved):
/lgtm

pkg/controller/operators/olm/operator.go
```diff
@@ -1211,13 +1235,143 @@ func (a *Operator) syncClusterServiceVersion(obj interface{}) (syncError error)
...
 		return fmt.Errorf("casting ClusterServiceVersion failed")
 	}
 
+	olmConfig, err := a.client.OperatorsV1().OLMConfigs().Get(context.TODO(), "cluster", metav1.GetOptions{})
+	if err != nil && !k8serrors.IsNotFound(err) {
+		return err
+	}
+
+	if err == nil {
+		go a.olmConfigQueue.AddAfter(olmConfig, time.Second*5)
```
tylerslaton, line 1372: nit: should this time.Second*5 be a constant defined at the top of the file?

test/e2e/csv_e2e_test.go
```diff
@@ -4171,6 +4174,253 @@ var _ = Describe("ClusterServiceVersion", func() {
...
+				if err != nil {
+					return err
+				}
+
+				var namespaces corev1.NamespaceList
+				if err := ctx.Ctx().Client().List(context.TODO(), &namespaces, &client.ListOptions{}); err != nil {
+					return err
+				}
+
+				if len(namespaces.Items)-1 != len(copiedCSVs.Items) {
```
tylerslaton, line 4250: nb: Should we check that the namespaces themselves contain the same elements? I believe there is a `gomega` target to do this if it is valuable.

test/e2e/csv_e2e_test.go
```diff
@@ -4171,6 +4174,253 @@ var _ = Describe("ClusterServiceVersion", func() {
...
+				if err := ctx.Ctx().Client().Create(context.TODO(), &csv); err != nil && !k8serrors.IsAlreadyExists(err) {
+					ctx.Ctx().Logf("Unable to create csv: %v", err)
+					return err
+				}
+
+				return nil
+			}).Should(Succeed())
+		})
+
+		It("should have Copied CSVs in all other namespaces", func() {
```
tylerslaton, line 4230: nb: I'm probably confused, but if an operator is installed in `AllNamespaces` mode shouldn't it not have copied CSV's by default?

test/e2e/csv_e2e_test.go
```diff
@@ -4171,6 +4174,253 @@ var _ = Describe("ClusterServiceVersion", func() {
...
+					return err
+				}
+
+				if len(namespaces.Items)-1 != len(copiedCSVs.Items) {
+					return fmt.Errorf("%d copied CSVs found, expected %d", len(copiedCSVs.Items), len(namespaces.Items)-1)
+				}
+
+				return nil
+			}).Should(Succeed())
+		})
```
tylerslaton, line 4256: nb: Should there be a check for if the operator is in AllNamespace mode and had copied CSV enabled/disabled (whichever is the opposite of the default)?

test/e2e/csv_e2e_test.go
awgreene: Copied CSVs are enabled by default in OLM today, users will need to disable copied CSVs via the `olmConfig/cluster` CR.

test/e2e/csv_e2e_test.go
awgreene: I create the operator earlier in the test in AllNamespace mode so I don't think we get any value from this check.

PR #2498, opened 2021-11-25
[WIP] Add cluster runtime constraint handling

perdasilva:
Signed-off-by: Per G. da Silva <perdasilva@redhat.com>

This PR introduces a backdoor way adding cluster runtime constraints to the resolver. The mechanism will look for a yaml file containing `olm.constraint`s that will be parsed into the `cache.Predicate`s and considered during the resolution.

**Current Resolution Integration Approach**
I've come up with a couple of strategies:
 1. Apply a global filter to the cache (I've added code for this but it's commented out)
 2. Set any installable violating _any_ runtime constraint to `prohibited` in the `addInvariants` method.

Option 2. seems the most reasonable since 1. adds a bit of hidden behaviour to the `cache` and it mightn't always be clear what is being applied and when.

OLM will needs to support cluster runtime contraints. We are currently working on a fully fledged solution but it won't be ready in time to unblock our users. This approach gives downstreams a way to side-load runtime constraints in the mean time.

**Reviewer Checklist**


pkg/controller/registry/resolver/resolver.go
```diff
@@ -568,6 +595,16 @@ func (r *SatResolver) addInvariants(namespacedCache cache.MultiCatalogOperatorFi
 			}
 			packageConflictToInstallable[prop.PackageName] = append(packageConflictToInstallable[prop.PackageName], installable.Identifier())
 		}
+
+		// apply runtime constraints to packages that aren't already installed
+		if !catalog.Virtual() && r.runtimeConstraintsProvider != nil {
```
dinhxuanvu, line 600: I'm actually fine with this AddInvariants solution. I was expecting to be messier but it looks reasonable to me.

pkg/controller/registry/resolver/runtime_constraints/provider.go
```diff
@@ -0,0 +1,80 @@
...
+
+func readRuntimeConstraintsYaml(yamlPath string) (*registry.PropertiesFile, error) {
+	// Read file
+	yamlFile, err := ioutil.ReadFile(yamlPath)
+	if err != nil {
+		return nil, err
+	}
+
+	// Parse yaml
+	var propertiesFile = &registry.PropertiesFile{}
```
dinhxuanvu, line 73: We need to decide if this `property` or a `dependency` so we can use either `PropertiesFile` or `DependenciesFile`. Honestly, it is just wording but it is a decision to be made for syntax sake.

pkg/controller/registry/resolver/runtime_constraints/provider.go
```diff
@@ -0,0 +1,80 @@
...
+func NewFromFile(runtimeConstraintsFilePath string) (*RuntimeConstraintsProvider, error) {
+	propertiesFile, err := readRuntimeConstraintsYaml(runtimeConstraintsFilePath)
+	if err != nil {
+		return nil, err
+	}
+
+	// Using package type to test with
+	// We may only want to allow the generic constraint types once they are readym
+	var constraints = make([]cache.Predicate, 0)
+	for _, property := range propertiesFile.Properties {
```
dinhxuanvu, line 39: I understand this will be substituted with `olm.constraint` type later which sounds good. If you are interested in testing this on a live cluster with CatalogSources that contain bundles with `MaxOCPVersion` properties, you can add some hardcoded stuff to handle `MaxOCPVersion` dependency which is similar to package dependency. Just a suggestion.

pkg/controller/registry/resolver/runtime_constraints/provider.go
```diff
@@ -0,0 +1,80 @@
...
+			return nil, errors.Errorf("Too many runtime constraints defined (%d/%d)", len(constraints), maxRuntimeConstraints)
+		}
+	}
+
+	return New(constraints), nil
+}
+
+func readRuntimeConstraintsYaml(yamlPath string) (*registry.PropertiesFile, error) {
+	// Read file
+	yamlFile, err := ioutil.ReadFile(yamlPath)
```
dinhxuanvu, line 67: Did you test how this file-reading code will work in real life? How is this file gonna be "injected" into the process?

pkg/controller/registry/resolver/cache/predicates.go
```diff
@@ -282,6 +282,24 @@ func (p orPredicate) String() string {
 	return b.String()
 }
 
+type notPredicate struct {
```
dinhxuanvu, line 285: I understand this section of code is added in order to write those test cases below. However, I'm pretty sure similar code is introduced in Eric's compound constraint PR. Make sure they are identical so there won't be any conflicting code later.

pkg/controller/registry/resolver/runtime_constraints/provider.go
```diff
@@ -0,0 +1,80 @@
...
+	"encoding/json"
+	"io/ioutil"
+
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/registry/resolver/cache"
+	"github.com/operator-framework/operator-registry/pkg/registry"
+	"github.com/pkg/errors"
+)
+
+const (
+	maxRuntimeConstraints = 10
```
dinhxuanvu, line 13: Reasonable restriction to avoid abusing. Make sure this number is known and noticed upon the completion of this work.

dinhxuanvu (commented):
Looking promising.
I have a few comments. Plus, there is a bundles.json file and not sure what is it for?

pkg/controller/registry/reconciler/reconciler.go
```diff
@@ -121,6 +121,8 @@ func Pod(source *v1alpha1.CatalogSource, name string, image string, saName strin
 			Annotations:  annotations,
 		},
 		Spec: v1.PodSpec{
+			// TODO: Remove this before merging
```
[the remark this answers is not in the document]
perdasilva: This is just so that I can easily run the catalog sources on kind and still talk to them via running/debugging the operator in GoLand. It will go away.

pkg/controller/registry/resolver/runtime_constraints/provider.go
perdasilva: That's not a bad idea! Can i use CRC for this? Or should I use the slack bot?

pkg/controller/registry/resolver/cache/predicates.go
perdasilva: Yeah, this can be thought of as provisional. If it adds value, we can keep it, if it will come from Eric, or don't need, punt it away

pkg/controller/registry/resolver/runtime_constraints/provider.go
perdasilva: I've tested it manually. I still need to add unit tests for this. I'm thinking the way to side-load it should be we have an environment variable that points to the path of the constraints file. Downstream, the file will be backed in to the olm operator image and we'll update the deployment to include the env var. This way, if people want to turn it off, it's doable. I'm not sure how this strategy fits with support, etc. E.g. if they turn it off would that mean the customer is in an unmanaged state?

perdasilva:
> Looking promising. I have a few comments. Plus, there is a bundles.json file and not sure what is it for?

Thank you ^^ I've removed the bundles.json. That was a mistake.

njhale (commented):
So this is very similar to an injection method we rejected a while back: https://github.com/operator-framework/operator-lifecycle-manager/pull/2360/files

> OLM will needs to support cluster runtime contraints. We are currently working on a fully fledged solution but it won't be ready in time to unblock our users. This approach gives downstreams a way to side-load runtime constraints in the mean time.

Which users are blocked without a method to side load global constraints? If we're talking about OpenShift users and the MaxOpenShiftVersion property, then I'd like to avoid adding new upstream APIs for them to be immediately superseded. We can carry a patch downstream that injects such specific constraints. In fact, I have some prior art there: https://github.com/openshift/operator-framework-olm/pull/194

pkg/controller/registry/resolver/resolver.go
```diff
@@ -568,6 +590,16 @@ func (r *SatResolver) addInvariants(namespacedCache cache.MultiCatalogOperatorFi
...
+			for _, predicate := range r.runtimeConstraintsProvider.Constraints() {
```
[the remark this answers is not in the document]
perdasilva: @njhale I'm happy to shift this to whatever implementation is best. But, I think in this case I need a little help. I don't have a handle on why a virtual bundle would be better, or how to exactly implement it. Would it look something like this?
 - Create a new virtual bundle
 - Mark is as mandatory
 - Add conflicts against any package that doesn't meet the bundle's properties

Does the resolver already handle virual bundles? What happens if it gets chosen as part of the resolver set? Does it know to throw it out? What do errors look like in this case?

pkg/controller/registry/resolver/resolver.go
```diff
@@ -20,19 +22,39 @@ import (
 	opregistry "github.com/operator-framework/operator-registry/pkg/registry"
 )
 
+const (
+	runtimeConstraintsFilePath = "runtime_constraints.json"
```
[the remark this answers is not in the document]
perdasilva: I think this could be much nicer. I just don't know what it means for the longer term vision of deppy and olm, etc. If this is still the way we'll want to handle these types of constraints. My thinking right now is to keep the "under-the-covers" way of doing it (baking this file in the downstream operator images) because it's super easy to roll-back and it buys us time to be more deliberate about how we want to handle these constraints in a first-class way. @joelanford @kevinrizza wdyt?

pkg/controller/registry/resolver/resolver_test.go
```diff
@@ -82,6 +102,101 @@ func TestDisjointChannelGraph(t *testing.T) {
...
+			expectedOperators:  cache.OperatorSet{"packageA.v1": packageA, "packageB.v1": packageB},
+			csvs:               nil,
+			subs:               []*v1alpha1.Subscription{packageASub},
+			err:                "",
+		},
+		{
+			title:           "Runtime constraints only accept packages A and C",
+			snapshotEntries: []*cache.Entry{packageA, packageB, packageC, packageD},
+			runtimeConstraints: []cache.Predicate{
+				cache.Or(cache.PkgPredicate("packageA"), cache.PkgPredicate("packageC")),
```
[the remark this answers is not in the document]
perdasilva: The test case here is a bit contrived atm. But once we have compound constraints, it ought to be doable. If not, then definitely with the generic constraints. The idea is to have properties/dependencies defined in the exact same way they are in the bundle.

perdasilva:
> > OLM will needs to support cluster runtime contraints. We are currently working on a fully fledged solution but it won't be ready in time to unblock our users. This approach gives downstreams a way to side-load runtime constraints in the mean time.
> 
> Which users are blocked without a method to side load global constraints? If we're talking about OpenShift users and the MaxOpenShiftVersion property, then I'd like to avoid adding new upstream APIs for them to be immediately superseded. We can carry a patch downstream that injects such specific constraints. In fact, I have some prior art there: [openshift/operator-framework-olm#194](https://github.com/openshift/operator-framework-olm/pull/194)

Yeah, the main targets at the moment are OpenShift and IBM. I don't think we will be targetting anything other than OCP version. But, personally, I don't think we are making too many throw away changes. I think being able to inject these constraints (somehow) and have them understood by the resolver already adds value to the upstream, even if we don't offer a nice interface (yet) to define those. It also puts it in the hand of the community to test, evaluate, use our way of side-loading, come up with their own, etc. So, I think I'd push back a bit on the idea of a downstream (code) patch atm.

pkg/controller/registry/resolver/resolver.go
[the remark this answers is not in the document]
njhale: we spoke about this synchronously

perdasilva:
closing in favor of #2523

PR #2456, opened 2021-11-17
Introduce Maintainers File

awgreene:
This commit introduces the MAINTAINERS.md file in an effort
to decouple project maintainer status from approval and lgtm
permissions relied on by the project automated continuous
integration tooling.

PR #2500, opened 2021-11-29
.github/workflows: Avoid running CI on markdown-only pull requests 

timflannagan:
Update the relevant set of workflows defined in the .github/workflows
directory and avoid running the CI-related jobs when PRs only contain
markdown-only changes.

Previously, this functionality was achieved through the paths_ignore
workflow syntax, but that keyword doesn't play well with repositories
that have setup explicit required action context's through branch
protection. This results in documentation-only PRs with the requisite
labels present being gated as those skipped jobs are reported as unknown
checks and automatic merging tooling cannot perform the merge.

See [1] for more information on this behavior.

[1] https://github.community/t/path-filtering-on-required-pull-request-checks/18402/3

Markdown-only PRs labeled with lgtm/approved cannot be successfully merged by tide due to the combination of how the paths-ignore keyword interacts with branch protection rules.

**Reviewer Checklist**

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,9 +1,24 @@
 name: build
 on:
+  push:
   pull_request:
 jobs:
+  skip_check:
+    runs-on: ubuntu-latest
+    outputs:
+      should_skip: ${{ steps.skip_check.outputs.should_skip }}
+    steps:
+      - id: skip_check
+        uses: fkirc/skip-duplicate-actions@v3.4.0
+        with:
+          paths_ignore: '["**/*.md"]'
+          cancel_others: 'true'
+          concurrent_skipping: 'outdated_runs'
+          skip_after_successful_duplicate: 'true'
   image:
     runs-on: ubuntu-latest
+    needs: skip_check
+    if: ${{ needs.skip_check.outputs.should_skip != 'true' }}
     steps:
     - name: Check out the repo
       uses: actions/checkout@v2
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,8 +1,6 @@
 name: build
 on:
   pull_request:
-    paths-ignore:
-    - '**/*.md'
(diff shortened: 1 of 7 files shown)


.github/workflows/build.yml
```diff
@@ -1,11 +1,24 @@
...
-    paths-ignore:
-    - '**/*.md'
 jobs:
+  skip_check:
+    runs-on: ubuntu-latest
+    outputs:
+      should_skip: ${{ steps.skip_check.outputs.should_skip }}
+    steps:
+      - id: skip_check
+        uses: fkirc/skip-duplicate-actions@v3.4.0
```
timflannagan, line 12: There's alternative actions we can use but I wasn't a huge fan of their ignore filtering behaviors when playing around with them locally. Another viable alternative would be to perform this filtering ourselves through the REST API using the /pulls/<PR #>/files endpoint.

timflannagan:
A concrete example of the current behavior between paths-ignore + branch protection policies conflicting: https://github.com/operator-framework/operator-lifecycle-manager/pull/2456.

.github/workflows/build.yml
timflannagan: There's likely a couple of more alternatives the more I think about:

- Run CI on every PR, regardless of content. This is the most costly option but most of the CI runs we like to skip are markdown-only PRs and most of the OLM documentation is housed in another repository.
- Dynamically determine required checks; this would require removing required status checks entirely and updating our tide configuration to accommodate that requirement
- Dynamically determine when to skip certain required checks (which is what this PR attempts to do)
  - Use the rest API and parse through the list of files that were changed in the PR
  - Use a custom action to perform this filtering as a job that runs before any CI-related jobs (e.g. a prerequisite job)
  - Could also play around with a workflow_dispatch singleton that triggers CI-related jobs but we'd likely run into the same issue of filtering + existing required status checks blocking PRs

PR #2501, opened 2021-12-01
pkg,test: Remove duplicate package imports

timflannagan:
Update various pkg,test packages are removed any obvious duplicate
package imports that violate local static linters.

Remove any obvious duplicate package imports that are identified by local, static linters.

Reduce duplicate package imports to help increase maintainability over time (and help alleviate a huge diff when migrating to enforcing this eventually in CI).

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/olm/labeler.go
+++ b/pkg/controller/operators/olm/labeler.go
@@ -6,7 +6,6 @@ import (
 
 	"github.com/operator-framework/api/pkg/operators/v1alpha1"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/registry/resolver/cache"
-	"github.com/operator-framework/operator-registry/pkg/registry"
 	opregistry "github.com/operator-framework/operator-registry/pkg/registry"
 	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	extv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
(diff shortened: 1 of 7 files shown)


pkg/controller/operators/olm/labeler.go
```diff
@@ -6,7 +6,6 @@ import (
 
 	"github.com/operator-framework/api/pkg/operators/v1alpha1"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/registry/resolver/cache"
-	"github.com/operator-framework/operator-registry/pkg/registry"
```
tylerslaton, line 9: `non-blocking question`: Would it make sense to keep the default name and remove the alternatively named one (opregistry) instead?

pkg/controller/registry/resolver/resolver_test.go
```diff
@@ -14,7 +14,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 
-	"github.com/operator-framework/api/pkg/lib/version"
```
tylerslaton, line 17: `non-blocking question`: Same as above

tylerslaton (commented):
Looks good to go. Have two small non-blocking questions.

njhale (approved):
/lgtm

pkg/controller/operators/olm/labeler.go
timflannagan: Answering both comments: totally fair question - I mainly went with my gut locally and tried to avoid removing the package import alias that was used more frequently. The one caveat is more common/standardized package imports that are used and I value consistency over standardization generally speaking.

PR #2503, opened 2021-12-01
pkg/controllers: Avoid logging transient errors to reduce log spam

timflannagan:
Update the operatorcondition and operatorconditiongenerator controllers
and avoid logging any apierrors.IsNotFound(err) to reduce the log spam.

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/operatorcondition_controller.go
+++ b/pkg/controller/operators/operatorcondition_controller.go
@@ -100,8 +100,7 @@ func (r *OperatorConditionReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	operatorCondition := &operatorsv2.OperatorCondition{}
 	err := r.Client.Get(context.TODO(), req.NamespacedName, operatorCondition)
 	if err != nil {
-		log.V(1).Error(err, "Unable to find operatorcondition")
-		return ctrl.Result{}, err
+		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/controller/operators/operatorcondition_controller.go
```diff
@@ -98,8 +98,7 @@ func (r *OperatorConditionReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	operatorCondition := &operatorsv2.OperatorCondition{}
 	err := r.Client.Get(context.TODO(), req.NamespacedName, operatorCondition)
 	if err != nil {
-		log.V(1).Error(err, "Unable to find operatorcondition")
-		return ctrl.Result{}, err
```
timflannagan, line 102: Just for further context, this is the log spamming I was seeing locally:

```
$ cat olm.log | grep "Unable to find operatorcondition" | wc -l
699
```

That olm operator container log had only been running for ~10m and the logging level was at the default info level.

timflannagan: Also in this case, it seems sufficient to ignore an error when apierrors.IsNotFound(err) evaluates to true if the operatorconditiongenerator controller is responsible for generating that resource.

exdx:
needs a rebase but LGTM
/lgtm

timflannagan:
I just want to double-check this is the correct implementation again.

/hold

PR #2504, opened 2021-12-02
Add instructions for setting up envtest

fgiloux:

Added instructions for installing envtest, mainly a link to controller-runtime documentation.

The first instruction `$ make test` fails when envtest is not available. 

**Reviewer Checklist**

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -27,6 +27,8 @@
 
 This project uses the built-in testing support for golang.
 
+Envtest is also used and needs to be set up. Follow [controller-runtime instructions] and set `KUBEBUILDER_ASSETS` environment variable to point to the installation directory, for instance: `/usr/local/kubebuilder/bin`.
+
 To run the tests for all go packages outside of the vendor directory, run:
 ```sh
 $ make test
@@ -46,11 +48,14 @@ To run a specific e2e test locally:
 $ make e2e-local TEST=TestCreateInstallPlanManualApproval
 ```
 
+[controller-runtime instructions]: https://pkg.go.dev/sigs.k8s.io/controller-runtime/tools/setup-envtest#section-readme
+
 #### Building
 
-Ensure your version of go is up to date; check that you're running v1.9 with the
-command:
+Ensure your version of go is up to date; check that you're running the same version as in go.mod with the
+commands:
 ```sh
+$ head go.mod
 $ go version
 ```
 


fgiloux:
Also changed the instruction for go version. Recommending looking at go.mod instead of using v1.9

njhale:
So the `make unit` target has a prereq that ensures kubebuilder is installed. `make test` seems to be vestigial/unused. I'd be interested in merging the two and incorporating some of the `envtest` changes here as part of the Makefile directly.

(see https://github.com/operator-framework/operator-lifecycle-manager/blob/40b8b7104e3ce577c1c0f0f08e6854d62978169c/Makefile#L50)

njhale (approved):
/approve

fgiloux:
> So the `make unit` target has a prereq that ensures kubebuilder is installed. `make test` seems to be vestigial/unused. I'd be interested in merging the two and incorporating some of the `envtest` changes here as part of the Makefile directly.
> 
> (see
> 
> https://github.com/operator-framework/operator-lifecycle-manager/blob/40b8b7104e3ce577c1c0f0f08e6854d62978169c/Makefile#L50
> )

It seems that there are more things than what I expected. If I understand it properly now:
1. document the use of `make unit` instead of `make test`
2. amend Makefile and merge `make unit` and `make test`
3.  document the installation of envtest (the purpose of this PR). In this regard you have "kubebuilder" in the doc and in the Makefile as prerequisite but I only see the use of envtest. Am I missing something? If not the Makefile should point to the instructions for envtest installation and the kubebuilder check could be removed (keeping only api and etcd):
~~~
# ifeq (, $(wildcard $(KUBEBUILDER_ASSETS)/kubebuilder))
#       $(error kubebuilder $(KUBEBUILDER_ASSETS_ERR))
# endif
~~~

In regards to 2. I am not sure exactly of the options you need to keep so that I would rather let it to you. Let me know how you want to proceed.

timflannagan:
Just a quick note: our current workflow setup of skipping markdown-only PRs doesn't work well with required branch protection checks so despite the requisite labels being present, tide can't merge this due to those defined protection rules. I took a stab at fixing this in https://github.com/operator-framework/operator-lifecycle-manager/pull/2500.

timflannagan:
(just recycling the checks as we should've merged the update to the workflows to they don't block tide from merging markdown-only PRs)

PR #2506, opened 2021-12-03
feat(resolver): support generic constraint using CEL

dinhxuanvu:

Introduce the new type of constraint that uses CEL (Common Expression Language)
as an expression language to define the constraint in a generic way.
This type of constraint will allow operator authors to specify a dependency
on any arbitrary properties in the bundle. This constraint also supports
logical operator such as AND and OR in the CEL expression.

https://github.com/operator-framework/enhancements/blob/master/enhancements/generic-constraints.md

**Reviewer Checklist**


pkg/controller/registry/resolver/cache/predicates.go
```diff
@@ -329,3 +330,41 @@ func (c countingPredicate) String() string {
...
+	rule      string
+	message   string
+}
+
+func (ep *evaluatorPredicate) Test(entry *Entry) bool {
+	props := make([]map[string]interface{}, len(entry.Properties))
+	for i, p := range entry.Properties {
+		var v interface{}
+		if err := json.Unmarshal([]byte(p.Value), &v); err != nil {
+			continue
```
perdasilva, line 345: should we at least log that there was an unmarshalling error? Or, maybe panic?

pkg/controller/registry/resolver/cache/predicates.go
```diff
@@ -329,3 +330,41 @@ func (c countingPredicate) String() string {
...
+		}
+		props[i] = map[string]interface{}{
+			"type":  p.Type,
+			"value": v,
+		}
+	}
+
+	ok, err := ep.evaluator.Evaluate(map[string]interface{}{"properties": props})
+	if err != nil {
+		return false
```
perdasilva, line 355: same here: should we log or panic?

pkg/controller/registry/resolver/constraints/cel.go
```diff
@@ -0,0 +1,113 @@
...
+		cel.Declarations(
+			decls.NewFunction("semver_compare",
+				decls.NewOverload("semver_compare",
+					[]*exprpb.Type{decls.Any, decls.Any},
+					decls.Int))),
+	}
+}
+
+func (semverLib) ProgramOptions() []cel.ProgramOption {
+	return []cel.ProgramOption{
```
perdasilva, line 62: it would be good to get some contextual comments to undrestand which options you are a setting and why (maybe with links to external CEL documentation?) (same for the `CompileOptions` method above)

pkg/controller/registry/resolver/constraints/cel.go
```diff
@@ -0,0 +1,113 @@
...
+		env: env,
+	}
+}
+
+type celEvaluatorProvider struct {
+	env *cel.Env
+}
+
+type celEvaluator struct {
+	p cel.Program
```
perdasilva, line 46: should we call it `celProgram` or `program` instead of `p`? Single letter names make the code harder to read.

dinhxuanvu:
Due to the automatic approved label added by the bot, I place a hold on here to avoid accident merging.
/hold

pkg/controller/registry/resolver/constraints/cel.go
dinhxuanvu: Will add some comments to explain some context here.

pkg/controller/registry/resolver/cache/predicates.go
dinhxuanvu: In the original code, it was a panic. However, as you can see other predicates don't panic on json unmarshal error as we simply ignore it. So that's why I do it this way. We currently don't have logger for predicates so there is no logging for that either.

perdasilva: How hard would it be to add logging? If we don't want to handle this in this PR, can we track this somewhere? No logging could lead to some hard to find bugs. Especially with a generic constraints evaluator, being in the dark here could be a bit painful.

perdasilva: I'd even suggest adding the panic back until we have logging

pkg/controller/registry/resolver/cache/predicates.go
```diff
@@ -329,3 +330,41 @@ func (c countingPredicate) String() string {
 func CountingPredicate(p Predicate, n *int) Predicate {
 	return countingPredicate{p: p, n: n}
 }
+
+type evaluatorPredicate struct {
+	evaluator constraints.Evaluator
+	rule      string
+	message   string
```
joelanford, line 337: Maybe I'm being naive, but I'm a little surprised to see constraint-specific logic on the OLM side of this. I was imagining that we'd define a generic `Constraint` interface in `o-f/api` that `Cel`, `All`, `Any`, `None`, etc. would all implement.

That way, all OLM would need to do would be unmarshal the constraints and then use that generic interface to evaluate them.  What am I missing?

pkg/controller/registry/resolver/constraints/cel.go
```diff
@@ -0,0 +1,127 @@
+package constraints
```
joelanford, line 1: Why is this code duplicated in both API and OLM? Shouldn't OLM just import from API?

pkg/controller/registry/resolver/resolver.go
```diff
@@ -343,6 +346,30 @@ func (r *SatResolver) getBundleInstallables(preferredNamespace string, bundleSta
...
 
+		for _, prop := range bundle.Properties {
+			if prop.Type != "olm.constraint" {
+				continue
+			}
+
+			var val struct {
+				Message string           `json:"message"`
+				Cel     *constraints.Cel `json:"cel"`
+			}
```
joelanford, line 357: Use the `Constraint` struct from API here?

pkg/controller/registry/resolver/constraints/cel.go
```diff
@@ -0,0 +1,127 @@
...
+	return false, fmt.Errorf("cel expression evalutated to %T, not bool", result.Value())
+}
+
+func (e *celEvaluatorProvider) Evaluator(rule string) (Evaluator, error) {
+	ast, issues := e.env.Compile(rule)
+	if err := issues.Err(); err != nil {
+		return nil, err
+	}
+
+	if ast.ResultType() != decls.Bool {
```
perdasilva, line 118: Should we also add a test for this code path?

pkg/controller/registry/resolver/cache/predicates.go
dinhxuanvu: This is predicate information. The `rule` and `message` are there to surface those information to the resolver during failure with `.String()` function. It is not the same as `Constraint` struct in `api` repo.

pkg/controller/registry/resolver/cache/predicates.go
dinhxuanvu: This is `continue` behavior is consistent with other predicates with the intention to ignore bad properties. Usually, bad properties shouldn't even make it to the cluster as bundle validation should catch this. A panic would cause the entire resolver action to fail which essentially means one bad property in one bundle will cause OLM to stop working entirely. I can go with `panic` for now given there is no logging here.

pkg/controller/registry/resolver/constraints/cel.go
dinhxuanvu: The PR on api repo was created later after I was adding some validation on registry repo and realized there were some shared codes between OLM and registry that can be consolidated in api repo.
I'm keeping the code here for now so that all of tests can be run properly while waiting for the api PR to merge. Even if we merge this PR first, I can do a subsequent PR to vendor new codes from api. That's merely a refactoring PR.

pkg/controller/registry/resolver/resolver.go
dinhxuanvu: Yes. That's the intention after the api PR merged.

pkg/controller/registry/resolver/constraints/cel.go
dinhxuanvu: I saw your comment on this in api PR. Will address this over there given this code is moving over to api repo anyway.

pkg/controller/registry/resolver/cache/predicates.go
perdasilva: Yeah, that makes sense. Silent errors aren't the best tho. Is it very difficult to have attach a logger to the class? Maybe just instantiate a global one for the package? I'm not sure how logging works in Go and our projects yet. If it's too much effort or too outside the scope of the PR and panic can be disruptive to the user, I'd suggest creating a tracking issue and link it in a comment.

dinhxuanvu: I think logging the error is the better solution here but as you said, it is not in-scope of this PR to rework the predicate to include logging. The `panic` solution is more disruptive unfortunately. It is kind of lose-lose situation. At best, we can keep it status quo with the `continue` hence my original code. Anyway, let me see what I can do here.

pkg/controller/registry/resolver/cache/predicates.go
```diff
@@ -329,3 +330,41 @@ func (c countingPredicate) String() string {
...
+	}
+
+	ok, err := cp.program.Evaluate(map[string]interface{}{"properties": props})
+	if err != nil {
+		return false
+	}
+	return ok
+}
+
+func CreateCelPredicate(env *constraints.CelEnvironment, rule string, message string) (Predicate, error) {
```
anik120, line 360: Same comment here, `CreateGenericPredicate`

anik120 (changes requested):

pkg/controller/registry/resolver/cache/predicates.go
```diff
@@ -329,3 +330,41 @@ func (c countingPredicate) String() string {
 func CountingPredicate(p Predicate, n *int) Predicate {
 	return countingPredicate{p: p, n: n}
 }
+
+type celPredicate struct {
```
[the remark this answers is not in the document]
dinhxuanvu: As we did discuss a bit on api [PR](https://github.com/operator-framework/api/pull/202#issuecomment-989401668), we defer any further interface design to a later time given right now given there isn't a need yet for a generic interface that can support multiple types of constraints. The generic constraint only supports CEL so it is a CEL constraint. We may or may not expand the support to other languages anytime soon.

estroz:
It would be nice to refactor some of this into a more generic setup like in #2418 but I can always rebase my PR if it goes in after this one, or do a follow-up if this goes in last.

/lgtm

estroz (commented):
Nice work @dinhxuanvu 

/lgtm

PR #2512, opened 2021-12-08
feat: catalog source pod scheduling config overrides

perdasilva:
The CatalogSource CRD was recently [updated](https://github.com/operator-framework/api/pull/173) to carry an additional and optional attribute `grpcPodConfig` as an envelope for pod spec overrides that we may want to give to the users, starting with: `tolerations`, `nodeSelector`, and `priorityClassName`. For instance:

```yaml
kind: CatalogSource
...
# optional
grpcPodConfig:
  # optional
  priorityClassName: https://github.com/operator-framework/api/pull/173
  #optional
  tolerations:
    - key: some
      operator: Equal
      value: value
      effect: NoSchedule
  # optional
  nodeSelector: 
     kubernetes.io/os: linux
     some: tag
```

We want to give cluster admins a way to force catalog source pods to be scheduled on specific infrastructure (i.e. nodes)

**Questions**

**Does it make sense to expose `priorityClassName` and not `priority`?**

`priorityClassName` seems to be used with special (custom) priority classes and `system-node-critical` and `system-cluster-critical` (which I think are reserved and very high priority). I think it should be fine to **not** expose the `priority` knob as well. I think `priorityClassName` covers most cases where an admin would want to affect priority. But, I'd be keen to have this assumption checked =)

**Should we update the api and use `*string` for `PriorityClassName`**

Since we want to override settings, it could make sense to differentiate between not set and empty. A priorityClassName of `""` says to use the default priority. We don't set our own priorityClassName in our catalog source pod definition. So, the only issue here is that in case we decide to do that (which I don't think is likely), the user will not be able to unset it, i.e. override it to be `""`. But, even though unlikely, if we have the chance and it is quick, why not "do it right"? (if you guys also agree that seems the most correctest)

**NOTE:** This PR will eventually also introduce `vendor` changes. For reviewer sanity, I'll try to keep those off the PR for as long as possible

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/registry/reconciler/reconciler.go
+++ b/pkg/controller/registry/reconciler/reconciler.go
@@ -169,6 +169,32 @@ func Pod(source *v1alpha1.CatalogSource, name string, image string, saName strin
 		},
 	}
 
+	// Override scheduling options if specified
+	if source.Spec.GrpcPodConfig != nil {
+		grpcPodConfig := source.Spec.GrpcPodConfig
+
+		// Override node selector
+		if grpcPodConfig.NodeSelector != nil {
+			pod.Spec.NodeSelector = make(map[string]string, len(grpcPodConfig.NodeSelector))
+			for key, value := range grpcPodConfig.NodeSelector {
+				pod.Spec.NodeSelector[key] = value
+			}
+		}
+
+		// Override priority class name
+		if grpcPodConfig.PriorityClassName != nil {
+			pod.Spec.PriorityClassName = *grpcPodConfig.PriorityClassName
+		}
+
+		// Override tolerations
+		if grpcPodConfig.Tolerations != nil {
+			pod.Spec.Tolerations = make([]v1.Toleration, len(grpcPodConfig.Tolerations))
+			for index, toleration := range grpcPodConfig.Tolerations {
+				pod.Spec.Tolerations[index] = *toleration.DeepCopy()
+			}
+		}
+	}
+
 	// Set priorityclass if its annotation exists
 	if prio, ok := annotations[CatalogPriorityClassKey]; ok && prio != "" {
 		pod.Spec.PriorityClassName = prio
--- a/pkg/controller/registry/reconciler/reconciler_test.go
+++ b/pkg/controller/registry/reconciler/reconciler_test.go
@@ -94,3 +94,179 @@ func TestPodContainerSecurityContext(t *testing.T) {
 	gotContainerSecCtx := gotPod.Spec.Containers[0].SecurityContext
 	require.Equal(t, expectedContainerSecCtx, gotContainerSecCtx)
 }
+
+func TestPodSchedulingOverrides(t *testing.T) {
+	// This test ensures that any overriding pod scheduling configuration elements
+	// defined in spec.grpcPodConfig are applied to the catalog source pod created
+	// when spec.sourceType = 'grpc' and spec.image is set.
+	var tolerationSeconds int64 = 120
+	var overriddenPriorityClassName = "some-prio-class"
+	var overriddenNodeSelectors = map[string]string{
+		"label":  "value",
+		"label2": "value2",
+	}
+	var defaultNodeSelectors = map[string]string{
+		"kubernetes.io/os": "linux",
+	}
+	var defaultPriorityClassName = ""
+
+	var overriddenTolerations = []corev1.Toleration{
+		{
+			Key:               "some/key",
+			Operator:          corev1.TolerationOpExists,
+			Effect:            corev1.TaintEffectNoExecute,
+			TolerationSeconds: &tolerationSeconds,
+		},
+		{
+			Key:      "someother/key",
+			Operator: corev1.TolerationOpEqual,
+			Effect:   corev1.TaintEffectNoSchedule,
+		},
+	}
+
+	testCases := []struct {
+		title                     string
+		catalogSource             *v1alpha1.CatalogSource
+		expectedNodeSelectors     map[string]string
+		expectedTolerations       []corev1.Toleration
+		expectedPriorityClassName string
+		annotations               map[string]string
+	}{
+		{
+			title: "no overrides",
+			catalogSource: &v1alpha1.CatalogSource{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "test",
+					Namespace: "testns",
+				},
+				Spec: v1alpha1.CatalogSourceSpec{
+					SourceType: v1alpha1.SourceTypeGrpc,
+					Image:      "repo/image:tag",
+				},
+			},
+			expectedTolerations:       nil,
+			expectedPriorityClassName: defaultPriorityClassName,
+			expectedNodeSelectors:     defaultNodeSelectors,
+		}, {
+			title: "override node selectors",
+			catalogSource: &v1alpha1.CatalogSource{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "test",
+					Namespace: "testns",
+				},
+				Spec: v1alpha1.CatalogSourceSpec{
+					SourceType: v1alpha1.SourceTypeGrpc,
+					Image:      "repo/image:tag",
+					GrpcPodConfig: &v1alpha1.GrpcPodConfig{
+						NodeSelector: overriddenNodeSelectors,
+					},
+				},
+			},
+			expectedTolerations:       nil,
+			expectedPriorityClassName: defaultPriorityClassName,
+			expectedNodeSelectors:     overriddenNodeSelectors,
+		}, {
+			title: "override priority class name",
+			catalogSource: &v1alpha1.CatalogSource{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "test",
+					Namespace: "testns",
+				},
+				Spec: v1alpha1.CatalogSourceSpec{
+					SourceType: v1alpha1.SourceTypeGrpc,
+					Image:      "repo/image:tag",
+					GrpcPodConfig: &v1alpha1.GrpcPodConfig{
+						PriorityClassName: &overriddenPriorityClassName,
+					},
+				},
+			},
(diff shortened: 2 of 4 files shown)


perdasilva:
holding until api changes are finalized and merged

tylerslaton (commented):
Looks good to me.

non-block question: Do we need to write an E2E test that validates that the changes to the pod spec are actually being utilized as expected? Or is that something that is owned by some other k8s API?

perdasilva:
> non-block question: Do we need to write an E2E test that validates that the changes to the pod spec are actually being utilized as expected? Or is that something that is owned by some other k8s API?

Yeah, I'll def add some e2e tests!

pkg/controller/registry/reconciler/reconciler.go
```diff
@@ -134,8 +134,7 @@ func Pod(source *v1alpha1.CatalogSource, name string, image string, saName strin
 					ReadinessProbe: &v1.Probe{
 						Handler: v1.Handler{
 							Exec: &v1.ExecAction{
-								Command: []string{"grpc_health_probe", "-addr=:50051"},
-							},
+								Command: []string{"grpc_health_probe", "-addr=:50051"}},
```
njhale, line 137: nit: the last right-bracket on this line matches `ExecAction{` -- is this change a typo?

pkg/controller/registry/reconciler/reconciler.go
```diff
@@ -169,6 +168,32 @@ func Pod(source *v1alpha1.CatalogSource, name string, image string, saName strin
...
+		// Override node selector
+		if grpcPodConfig.NodeSelector != nil {
+			pod.Spec.NodeSelector = make(map[string]string, len(grpcPodConfig.NodeSelector))
+			for key, value := range grpcPodConfig.NodeSelector {
+				pod.Spec.NodeSelector[key] = value
+			}
+		}
+
+		// Override priority class name
+		if grpcPodConfig.PriorityClassName != "" {
```
njhale, line 184: @perdasilva Now that I'm reading it, I think you're right to suggest we make this a pointer -- it leaves the option open for OLM to have some default that's not `""` and for users to opt into the original.

njhale (commented):
> NOTE: This PR will eventually also introduce vendor changes. For reviewer sanity, I'll try to keep those off the PR for as long as possible

I think it's sufficiently clean to include vendor changes in a separate commit, that way we can still view the meaningful diffs separately and don't need to re-lgtm later.

Other than that, this is looking good! I think we just need an e2e test to make sure OLM respects these settings and updates existing catalog pods when changes are detected.

deploy/chart/crds/0000_50_olm_00-catalogsources.crd.yaml
```diff
@@ -210,4 +245,4 @@ spec:
       served: true
       storage: true
       subresources:
-        status: {}
+        status: {}
```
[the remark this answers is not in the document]
perdasilva: it may have been me - I think this will be fixed in the next iteration when I include all the vendored changes properly

pkg/controller/registry/reconciler/reconciler.go
perdasilva: might have come from goland format or make format. I'll :eyes:

pkg/controller/registry/reconciler/reconciler.go
timflannagan: Any thoughts on how we should derive configuration when the `operatorframework.io/priorityclass` annotation has been specified on a CatalogSource: https://github.com/operator-framework/operator-lifecycle-manager/blob/4fc476fc520d4d595368ffc8e4c032c4d34dd796/pkg/controller/registry/reconciler/reconciler.go#L27? I'm not entirely sure if that hacky way of propagating configuration to the underlying Pod violates any API contract? It might also be entirely reasonable to point users towards these new configuration fields in the CatalogSource entirely, but it's unknown how many users were relying on that behavior.

gallettilance: this change seems to be backward compatible. If I had a priorityclass set on the catalogsource and now I add a config that contains a priority class, it's still [the former that is propagated through](https://github.com/operator-framework/operator-lifecycle-manager/pull/2512/files#diff-fffdeef1fc140a5dc5dc92dda323f567a6e46fc2ecbb0b91ba907acd02bde50dL173-L175).

gallettilance (approved):
lgtm

pkg/controller/registry/reconciler/reconciler.go
perdasilva: @timflannagan I've added a test to ensure that the annotation takes precedence. We'll handle this in documentation as well. We should probably deprecate the annotation tho.

pkg/controller/registry/reconciler/reconciler_test.go
```diff
@@ -94,3 +94,179 @@ func TestPodContainerSecurityContext(t *testing.T) {
...
+						PriorityClassName: &overriddenPriorityClassName,
+						Tolerations:       overriddenTolerations,
+					},
+				},
+			},
+			expectedTolerations:       overriddenTolerations,
+			expectedPriorityClassName: overriddenPriorityClassName,
+			expectedNodeSelectors:     overriddenNodeSelectors,
+		}, {
+			title: "priorityClassName annotation takes precedence",
```
timflannagan, line 243: :+1:

test/e2e/catsrc_pod_config_e2e_test.go
```diff
@@ -0,0 +1,201 @@
...
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/operator-framework/api/pkg/operators/v1alpha1"
+	"github.com/operator-framework/operator-lifecycle-manager/test/e2e/ctx"
+	k8serror "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	k8scontrollerclient "sigs.k8s.io/controller-runtime/pkg/client"
+
+	//. "github.com/onsi/gomega"
```
timflannagan, line 16: nit: reminder to remove this

perdasilva:
> non-block question: Do we need to write an E2E test that validates that the changes to the pod spec are actually being utilized as expected? Or is that something that is owned by some other k8s API?

@tylerslaton I thought about it, but yeah. That's the conclusion I reached. As long as its setting the right things in the right places the rest is up to kube

test/e2e/ctx/provisioner_kind.go
```diff
@@ -83,6 +83,7 @@ func Provision(ctx *TestContext) (func(), error) {
 	}
 	kubeconfigPath := filepath.Join(dir, "kubeconfig")
 
+	fmt.Printf("kubeconfig: %s\n", kubeconfigPath)
```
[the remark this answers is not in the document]
perdasilva: I'll update to use the context logger! Nice one! I think this is useful output (even though you can just export the kubeconfig - also didn't know about _that_ - so thanks for that too!

pkg/controller/registry/reconciler/reconciler.go
perdasilva: I think that's correct - it may have been just a format change.

pkg/controller/registry/reconciler/reconciler.go
```diff
@@ -169,6 +169,32 @@ func Pod(source *v1alpha1.CatalogSource, name string, image string, saName strin
...
+	// Override scheduling options if specified
+	if source.Spec.GrpcPodConfig != nil {
+		grpcPodConfig := source.Spec.GrpcPodConfig
+
+		// Override node selector
+		if grpcPodConfig.NodeSelector != nil {
+			pod.Spec.NodeSelector = make(map[string]string, len(grpcPodConfig.NodeSelector))
+			for key, value := range grpcPodConfig.NodeSelector {
+				pod.Spec.NodeSelector[key] = value
+			}
```
anik120, line 181: non blocking nit: why is copying the map necessary here? Why not make `grpcPodConfig.NodeSelector` and `pod.Spec.NodeSelector` point to the same thing?

test/e2e/catsrc_pod_config_e2e_test.go
```diff
@@ -0,0 +1,200 @@
...
+			_ = client.Delete(context.TODO(), catalogSource)
+
+			// wait for it to go away
+			Eventually(func() bool {
+				err := client.Get(context.TODO(), k8scontrollerclient.ObjectKey{
+					Name:      catalogSource.GetName(),
+					Namespace: catalogSource.GetNamespace(),
+				}, &v1alpha1.CatalogSource{})
+				return k8serror.IsNotFound(err)
+			}).Should(BeTrue())
```
anik120, line 75: Another non-blocking nit: I'll be surprised if this isn't already there in a `utils`, but this looks like code that'll be repeated fairly often. Something along the line of `deleteObj(client, name, namespace)` func could help reduce the redundancy.

anik120 (changes requested):
Looks good mostly @perdasilva , had a couple of non-blocking questions otherwise this looks good to go.

pkg/controller/registry/reconciler/reconciler.go
perdasilva: I thought I'd be defensive here because the nodeselector map is passed by reference. If possible (thought probably unlikely) that you could have out of band changes make it into the pod spec (e.g. if something else changes the map). So, I thought it safer to copy it over and avoid possible mysterious bugs.

test/e2e/catsrc_pod_config_e2e_test.go
perdasilva: This is a great point - I will check if I can find anything

pkg/controller/registry/reconciler/reconciler.go
anik120: okay that's fair

PR #2368, opened 2021-09-21
go.*,pkg,vendor: Bump controller-runtime to v0.10.1

timflannagan:
**Description of the change:**
Bumps the controller runtime dependency to v0.10.1 which contains the race condition fix around metadata-only watchers.

**Motivation for the change:**
Removes the replace pin in the root go.mod.

**Reviewer Checklist**
- [x] Implementation matches the proposed design, or proposal is updated to match implementation
- [x] Sufficient unit test coverage
- [x] Sufficient end-to-end test coverage
- [x] Docs updated or added to `/doc`
- [x] Commit messages sensible and descriptive

Fixes #2353

PR #2513, opened 2021-12-08
WIP: Introduce olm config (Revert CR Bump in API Repo)

awgreene:
This PR is similar to https://github.com/operator-framework/operator-lifecycle-manager/pull/2466 but revert the CR bump in the api repo.


go.mod
```diff
@@ -62,6 +62,7 @@ replace (
 	// controller runtime
 	github.com/openshift/api => github.com/openshift/api v0.0.0-20200331152225-585af27e34fd // release-4.5
 	github.com/openshift/client-go => github.com/openshift/client-go v0.0.0-20200326155132-2a6cd50aedd0 // release-4.5
+	github.com/operator-framework/api => github.com/awgreene/api v0.0.4
```
awgreene, line 65: TODO: If we are not able to unblock https://github.com/operator-framework/operator-lifecycle-manager/pull/2368 then revert the CR bump in https://github.com/operator-framework/api/pull/154

deploy/chart/templates/0000_50_olm_03-olmConfig.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: operators.coreos.com/v1
+kind: OLMConfig
+metadata:
+  name: cluster
```
njhale, line 4: not sure this needs to be included by default; OLM should probably create a default if it can't find one on the cluster.

nit: we need to make sure this isn't included downstream by default b/c CVO will stomp user changes.

awgreene: We'd need to add the create-only annotation downstream.

awgreene:
Closing in favor of #2466

PR #2518, opened 2021-12-10
test/e2e: Refactor the Operator e2e tests to clean up testing resources

timflannagan:
Update test/e2e/operator_test.go and attempt to refactor the current e2e
specs to reduce the number of testing artifacts that are left behind
when individual tests pass/fail.

**Reviewer Checklist**

### DIFF
--- a/test/e2e/operator_test.go
+++ b/test/e2e/operator_test.go
@@ -37,9 +37,7 @@ var _ = Describe("Operator API", func() {
 		client          controllerclient.Client
 		operatorFactory decorators.OperatorFactory
 	)
-
 	BeforeEach(func() {
-		// Setup common utilities
 		clientCtx = context.Background()
 		scheme = ctx.Ctx().Scheme()
 		listOpts = metav1.ListOptions{}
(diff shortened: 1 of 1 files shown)


test/e2e/operator_test.go
```diff
@@ -70,172 +76,175 @@ var _ = Describe("Operator API", func() {
...
-	It("should surface components in its status", func() {
-		o := &operatorsv1.Operator{}
-		o.SetName(genName("o-"))
-
-		Consistently(o).ShouldNot(ContainCopiedCSVReferences())
-
-		Eventually(func() error {
-			return client.Create(clientCtx, o)
-		}).Should(Succeed())
+	Context("when an Operator resource can select its components by label", func() {
```
timflannagan, line 79: Note: moved this from an `It` to a `Context` so we gain access to BeforeEach/AfterEach/etc. blocks. Before, this It spec would leak testing Operator resources when it failed the test case (but cleaned up properly when it passed).

test/e2e/operator_test.go
```diff
@@ -70,172 +76,175 @@ var _ = Describe("Operator API", func() {
...
 			}).Should(Succeed())
-		}()
+		})
 
-		By("eventually having a status that contains its component label selector")
-		w, err := operatorClient.Watch(clientCtx, listOpts)
-		Expect(err).ToNot(HaveOccurred())
-		defer w.Stop()
+		It("should not contain copied csv status references", func() {
+			Consistently(o).ShouldNot(ContainCopiedCSVReferences())
```
timflannagan, line 98: I moved this check to it's own It spec as before we were checking whether the operator object had any CSV references _before_ creating that resource, leading to what appeared to be a wasted check.

test/e2e/operator_test.go
```diff
@@ -70,172 +76,175 @@ var _ = Describe("Operator API", func() {
...
+				err := client.Get(clientCtx, types.NamespacedName{Name: o.GetName()}, o)
+				if apierrors.IsNotFound(err) {
+					return nil
+				}
+				return err
+			}).Should(Succeed())
+		})
 	})
 
 	Context("when a subscription to a package exists", func() {
```
timflannagan, line 250: Note: we were previously leaking the kiali installation, it's CRDs, and the corresponding Operator resource that was generated despite these test cases passing or failing.

test/e2e/operator_test.go
```diff
@@ -323,12 +332,21 @@ var _ = Describe("Operator API", func() {
...
+			}).Should(Succeed())
+
+			By("Deleting the kiali-related CRDs")
+			Eventually(func() error {
+				return client.DeleteAllOf(clientCtx, &apiextensionsv1.CustomResourceDefinition{}, controllerclient.MatchingLabels{
+					fmt.Sprintf("operators.coreos.com/%s", operatorName.Name): "",
+				})
+			}).Should(Succeed())
+
+			By("Deleting the test Operator resource")
```
timflannagan, line 347: It's highly likely that the generated Operator resource will be left behind, despite this check being in place, as there's some caching issues in the current operator controller. I added a top-level DeleteCollection call in the AfterEach block to help alleviate the frequency this happens.

test/e2e/operator_test.go
anik120: A question and a nit: 

Question: Why not start with a `When`? 

nit: We should probably clean up the long list in the comment too. I believe they were written when old tests were being ported over to ginkgo, but time constraint forced to putting these in `It` blocks and keep those comments for safe measures.

test/e2e/operator_test.go
```diff
@@ -70,172 +76,175 @@ var _ = Describe("Operator API", func() {
...
+			Consistently(o).ShouldNot(ContainCopiedCSVReferences())
+		})
 
-		deadline, cancel := context.WithTimeout(clientCtx, 1*time.Minute)
-		defer cancel()
+		It("should surface referenced components in its status", func() {
+			By("eventually having a status that contains its component label selector")
+			w, err := operatorClient.Watch(clientCtx, listOpts)
+			Expect(err).To(BeNil())
+			defer w.Stop()
```
anik120, line 105:
```suggestion
			defer w.Stop()
			Expect(err).To(BeNil())
```
```suggestion
			Expect(err).To(BeNil())
			defer w.Stop()
```

test/e2e/operator_test.go
```diff
@@ -70,172 +76,175 @@ var _ = Describe("Operator API", func() {
...
-		nsB.SetName(genName("ns-b-"))
+				return false
+			}))
+			defer w.Stop()
 
-		for _, ns := range []*corev1.Namespace{nsA, nsB} {
-			Eventually(func() error {
-				return client.Create(clientCtx, ns)
-			}).Should(Succeed())
+			// Create namespaces ns-a and ns-b
```
anik120, line 126: nit: these comments are just so unnecessary

anik120 (commented):
looks great Tim, have a few very small nits other lgtm

test/e2e/operator_test.go
tylerslaton: @anik120 +1 for both points.

How easily accessible is the same information in those comments if we were to remove them? Only thing I'd want to make sure of is that we're not losing context in remove them.

test/e2e/operator_test.go
```diff
@@ -70,172 +76,175 @@ var _ = Describe("Operator API", func() {
...
+				}
 
-		// Create namespaces ns-a and ns-b
-		nsA := &corev1.Namespace{}
-		nsA.SetName(genName("ns-a-"))
-		nsB := &corev1.Namespace{}
-		nsB.SetName(genName("ns-b-"))
+				return false
+			}))
+			defer w.Stop()
```
tylerslaton, line 124: Is this second `defer w.Stop()` necassary?

tylerslaton (commented):
/lgtm

Small question, but overall the approach looks great. I like the defer pattern you have for cleaning up the resources.

test/e2e/operator_test.go
timflannagan: That diff is misleading - I was simply indenting that existing codepaths, but agreed, I don't see a ton of value in this style of comment either.

test/e2e/operator_test.go
timflannagan: I don't think so - I mainly just indented the existing test cases vs. making any heavy modifications.

test/e2e/operator_test.go
timflannagan: I ended up removing them entirely, or replacing them with `By(...)` statements in the latest force push.

test/e2e/operator_test.go
timflannagan: I updated the Context -> When given the comment structure was already there. Do you have any preference when to use a Context vs. When containers? I wasn't able to find any best practices around structuring ginkgo e2e tests so there might be something I'm missing.

PR #2520, opened 2021-12-10
chore(e2e): naively parallelize CI jobs by chunking alphabetically

estroz:
**Description of the change:** parallelized e2e tests (for upstream CI only)

**Motivation for the change:** current sequential runs are ~1h10m. the longest parallel run is ~24 min, and could be lower if we increase the matrix size. you will hit Amdahl's law quickly from test setup. it's a [little hacky](https://github.com/operator-framework/operator-lifecycle-manager/pull/2520/files#diff-081212da21b2878374614577832dc62c197e75e2b441903900e4d40b4f32fa29R16) but adjusts automatically when new specs are added/removed.

I'll add some tests next week

**Reviewer Checklist**

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -10,6 +10,7 @@ on:
 jobs:
   e2e-tests:
     strategy:
+      fail-fast: false
       matrix:
         parallel-id: [0, 1, 2, 3]
     runs-on: ubuntu-latest
@@ -25,3 +26,4 @@ jobs:
         with:
           name: e2e-test-output-${{ (github.event.pull_request.head.sha || github.sha) }}-${{ github.run_id }}-${{ matrix.parallel-id }}
           path: ${{ github.workspace }}/bin/artifacts/*
+# TODO: create job to combine test artifacts using code in https://github.com/operator-framework/operator-lifecycle-manager/pull/1476
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -8,7 +8,7 @@ on:
   pull_request:
   workflow_dispatch:
 jobs:
-  e2e-tests:
+  e2e:
     strategy:
       fail-fast: false
       matrix:
@@ -26,4 +26,16 @@ jobs:
         with:
           name: e2e-test-output-${{ (github.event.pull_request.head.sha || github.sha) }}-${{ github.run_id }}-${{ matrix.parallel-id }}
           path: ${{ github.workspace }}/bin/artifacts/*
-# TODO: create job to combine test artifacts using code in https://github.com/operator-framework/operator-lifecycle-manager/pull/1476
+        # TODO: create job to combine test artifacts using code in https://github.com/operator-framework/operator-lifecycle-manager/pull/1476
+  e2e-tests:
+    if: ${{ always() }}
+    runs-on: ubuntu-latest
+    needs: e2e
+    steps:
+    - run: |
+        echo "Matrix result: ${{ needs.e2e.result }}"
+    - name: check individual matrix results
+      if: ${{ needs.e2e.result == 'failure' }}
+      run: |
+        echo 'Failure: at least one e2e matrix job has failed'
+        exit 1
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -9,16 +9,19 @@ on:
   workflow_dispatch:
 jobs:
   e2e-tests:
+    strategy:
+      matrix:
+        parallel-id: [0, 1, 2, 3]
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v1
       - uses: actions/setup-go@v2
         with:
           go-version: '~1.17'
-      - run: make e2e-local E2E_NODES=2 ARTIFACTS_DIR=./artifacts/
+      - run: make e2e-local E2E_TEST_CHUNK=${{ matrix.parallel-id }} E2E_TEST_NUM_CHUNKS=${{ strategy.job-total }} E2E_NODES=2 ARTIFACTS_DIR=./artifacts/
       - name: Archive Test Artifacts # test results, failed or not, are always uploaded.
         if: ${{ always() }}
         uses: actions/upload-artifact@v2
         with:
-          name: e2e-test-output-${{(github.event.pull_request.head.sha||github.sha)}}-${{ github.run_id }}
+          name: e2e-test-output-${{ (github.event.pull_request.head.sha || github.sha) }}-${{ github.run_id }}-${{ matrix.parallel-id }}
           path: ${{ github.workspace }}/bin/artifacts/*
--- a/Makefile
+++ b/Makefile
@@ -126,7 +126,17 @@ setup-bare: clean e2e.namespace
 E2E_NODES ?= 1
 E2E_FLAKE_ATTEMPTS ?= 1
 E2E_TIMEOUT ?= 90m
-E2E_OPTS ?= $(if $(E2E_SEED),-seed '$(E2E_SEED)') $(if $(TEST),-focus '$(TEST)') -flakeAttempts $(E2E_FLAKE_ATTEMPTS) -nodes $(E2E_NODES) -timeout $(E2E_TIMEOUT) -v -randomizeSuites -race -trace -progress
+E2E_COND_OPTS := $(if $(E2E_SEED),-seed '$(E2E_SEED)')
+# Optionally run an individual chunk of e2e test specs.
+# Do not use this from the CLI; this is intended to be used by CI only.
+E2E_TEST_CHUNK ?= all
+E2E_TEST_NUM_CHUNKS ?= 4
+ifeq (all,$(E2E_TEST_CHUNK))
+E2E_COND_OPTS := $(E2E_COND_OPTS) $(if $(TEST),-focus '$(TEST)')
+else
+E2E_COND_OPTS := $(E2E_COND_OPTS) -focus "$(shell go run ./test/e2e/split/... -chunks $(E2E_TEST_NUM_CHUNKS) -print-chunk $(E2E_TEST_CHUNK) ./test/e2e)"
+endif
+E2E_OPTS ?= $(E2E_COND_OPTS) -flakeAttempts $(E2E_FLAKE_ATTEMPTS) -nodes $(E2E_NODES) -timeout $(E2E_TIMEOUT) -v -randomizeSuites -race -trace -progress
 E2E_INSTALL_NS ?= operator-lifecycle-manager
 E2E_TEST_NS ?= operators
 
--- a/Makefile
+++ b/Makefile
@@ -43,7 +43,7 @@ all: test build
 test: clean cover.out
 
 unit: kubebuilder
-	KUBEBUILDER_ASSETS=$(KUBEBUILDER_ASSETS) go test $(MOD_FLAGS) $(SPECIFIC_UNIT_TEST) -tags "json1" -race -count=1 ./pkg/...
+	KUBEBUILDER_ASSETS=$(KUBEBUILDER_ASSETS) go test $(MOD_FLAGS) $(SPECIFIC_UNIT_TEST) -tags "json1" -race -count=1 ./pkg/... ./test/e2e/split/...
 
 # Ensure kubebuilder is installed before continuing
 KUBEBUILDER_ASSETS_ERR := not detected in $(KUBEBUILDER_ASSETS), to override the assets path set the KUBEBUILDER_ASSETS environment variable, for install instructions see https://book.kubebuilder.io/quick-start.html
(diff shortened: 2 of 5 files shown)


estroz:
This was tested on my private fork, see https://github.com/estroz/operator-lifecycle-manager/actions/runs/1565786458

.github/workflows/e2e-tests.yml
```diff
@@ -6,16 +6,19 @@ on:
...
         with:
           go-version: '~1.16'
-      - run: make e2e-local E2E_NODES=2 ARTIFACTS_DIR=./artifacts/
+      - run: make e2e-local E2E_TEST_CHUNK=${{ matrix.parallel-id }} E2E_TEST_NUM_CHUNKS=${{ strategy.job-total }} E2E_NODES=2 ARTIFACTS_DIR=./artifacts/
       - name: Archive Test Artifacts # test results, failed or not, are always uploaded.
         if: ${{ always() }}
         uses: actions/upload-artifact@v2
         with:
-          name: e2e-test-output-${{(github.event.pull_request.head.sha||github.sha)}}-${{ github.run_id }}
+          name: e2e-test-output-${{ (github.event.pull_request.head.sha || github.sha) }}-${{ github.run_id }}-${{ matrix.parallel-id }}
```
timflannagan, line 23: I haven't poked around the actual testing artifacts, but I saw these warning messages being produced during one of e2e runs for this PR:

```
No files were found with the provided path: /home/runner/work/operator-lifecycle-manager/operator-lifecycle-manager/bin/artifacts-1565786458/*
```

estroz: Fix that, was using the wrong context prior. It should have used the matrix context, and does now.

.github/workflows/e2e-tests.yml
```diff
@@ -6,16 +6,19 @@ on:
   pull_request:
 jobs:
   e2e-tests:
+    strategy:
+      matrix:
+        parallel-id: [0, 1, 2, 3]
```
timflannagan, line 11: Ah TIL that build matrix strategies don't play well with existing required checks defined through branch protection policies. We may need to work towards a model where we remove required checks entirely (which may implicitly imply we need to remove tide entirely for this repository).

test/e2e/split/main.go
```diff
@@ -0,0 +1,100 @@
...
+	currIdx := 0
+	for chunkIdx := 0; chunkIdx < numChunks; chunkIdx++ {
+		nextIdx := int(math.Min(float64(currIdx+interval), float64(len(describes))))
+		chunks[chunkIdx] = describes[currIdx:nextIdx]
+		currIdx = nextIdx
+	}
+
+	sb := strings.Builder{}
+	sb.WriteString("(")
+	sb.WriteString(chunks[printChunk][0])
```
timflannagan, line 92: nit: this looks like it's prone to an index out of range panic?

test/e2e/split/main.go
```diff
@@ -0,0 +1,100 @@
...
+	var numChunks, printChunk int
+	flag.IntVar(&numChunks, "chunks", 1, "Number of chunks to create focus regexps for")
+	flag.IntVar(&printChunk, "print-chunk", 0, "Chunk to print a regexp for")
+	flag.Parse()
+
+	if printChunk >= numChunks {
+		log.Fatalf("the chunk to print (%d) must be a smaller number than the number of chunks (%d)", printChunk, numChunks)
+	}
+
+	dir := flag.Arg(0)
```
timflannagan, line 28: nit: we may need validation to ensure this directory has been supplied as a flag?

.github/workflows/e2e-tests.yml
```diff
@@ -6,16 +6,19 @@ on:
...
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-go@v2
         with:
           go-version: '~1.16'
-      - run: make e2e-local E2E_NODES=2 ARTIFACTS_DIR=./artifacts/
+      - run: make e2e-local E2E_TEST_CHUNK=${{ matrix.parallel-id }} E2E_TEST_NUM_CHUNKS=${{ strategy.job-total }} E2E_NODES=2 ARTIFACTS_DIR=./artifacts/
```
timflannagan, line 18: Any idea on where I can find some documentation around the `strategy.job-total` syntax?

Makefile
```diff
@@ -126,7 +126,17 @@ setup-bare: clean e2e.namespace
...
-E2E_OPTS ?= $(if $(E2E_SEED),-seed '$(E2E_SEED)') $(if $(TEST),-focus '$(TEST)') -flakeAttempts $(E2E_FLAKE_ATTEMPTS) -nodes $(E2E_NODES) -timeout $(E2E_TIMEOUT) -v -randomizeSuites -race -trace -progress
+E2E_COND_OPTS := $(if $(E2E_SEED),-seed '$(E2E_SEED)')
+# Optionally run an individual chunk of e2e test specs.
+# Do not use this from the CLI; this is intended to be used by CI only.
+E2E_TEST_CHUNK ?= all
+E2E_TEST_NUM_CHUNKS ?= 4
+ifeq (all,$(E2E_TEST_CHUNK))
+E2E_COND_OPTS := $(E2E_COND_OPTS) $(if $(TEST),-focus '$(TEST)')
+else
+E2E_COND_OPTS := $(E2E_COND_OPTS) -focus "$(shell go run ./test/e2e/split/... -chunks $(E2E_TEST_NUM_CHUNKS) -print-chunk $(E2E_TEST_CHUNK) ./test/e2e)"
```
timflannagan, line 137: Just for my own edification: we're essentially firing off several e2e instances using the job matrix and printing out individual chunks (e.g. give me the chunk for the first index, second index, etc.), we create a kind cluster, ginkgo runs that chunk of focused tests, etc.? Any idea on whether this chunking is deterministic (it reads like it is when poking around the splitting package), and whether this scales out when adding new tests?

timflannagan, line 137: Also TIL ginkgo has support for regex focus grouping :+1:

timflannagan:
This is really, really (albeit a bit hacky) cool work man - this will infinitely improve the current state of long, frustrating e2e feedback. I just had a couple of questions, and I'd like to land anything that substantially improves the e2e feedback, but we'll need to update the branch protection for our tide configuration (or remove tide entirely which I think is my preference).

.github/workflows/e2e-tests.yml
estroz: Yeah unfortunately actions' matrix capabilities, like required checks and rerunning failed matrix cells only, are lacking.

test/e2e/split/main.go
estroz: Indeed, I need to add tests for this and other stuff.

Makefile
estroz: Correct interpretation. Chunking is deterministic because specs are sorted, and scales with added or removed specs because test files are re-searched each run for new specs.

.github/workflows/e2e-tests.yml
estroz: I couldn't find much other than this [mention](https://docs.github.com/en/actions/learn-github-actions/contexts#about-contexts) and this [post](https://github.community/t/get-length-of-strategy-matrix-or-get-all-matrix-options/18342/3). I suspect you'd have to read the actions code to get more info.

estroz:
@timflannagan the main things I want to add before this merges is tests for the script, and some deduplication by prefix (if both `Subscriptions` and `Subscripts blah blah` exist, the latter will be run twice if they are in different chunks). I'll be working on that this week so this can get in asap.

test/e2e/split/main.go
estroz: Tests and fixes done.

.github/workflows/e2e-tests.yml
timflannagan: Nope that's fine - I think it's a bit obvious once you wrap you head around the new e2e system, and was mainly curious.

.github/workflows/e2e-tests.yml
```diff
@@ -6,16 +6,21 @@ on:
...
       - name: Archive Test Artifacts # test results, failed or not, are always uploaded.
         if: ${{ always() }}
         uses: actions/upload-artifact@v2
         with:
-          name: e2e-test-output-${{(github.event.pull_request.head.sha||github.sha)}}-${{ github.run_id }}
+          name: e2e-test-output-${{ (github.event.pull_request.head.sha || github.sha) }}-${{ github.run_id }}-${{ matrix.parallel-id }}
           path: ${{ github.workspace }}/bin/artifacts/*
+# TODO: create job to combine test artifacts using code in https://github.com/operator-framework/operator-lifecycle-manager/pull/1476
```
timflannagan, line 26: :+1:

timflannagan:
I think the overall changes are reasonable to me looking back at that this but my brain is pretty much mush at this point. I'd like to land these changes ASAP but we'd also need to land a tide configuration update as well. I should be back to full capacity by tomorrow.

.github/workflows/e2e-tests.yml
timflannagan: I think my preference would be to temporarily remove this as required checks vs. purging tide entirely (which I'd still like to do long term). I can work towards allowlisting the e2e-tests as a optional check and we'll just have reviewers ensure they're passing before labeling any PR.

test/e2e/split/main.go
```diff
@@ -0,0 +1,226 @@
...
+	for _, sp := range split {
+		if sp = strings.TrimSpace(sp); sp == "" {
+			continue
+		}
+		next, hasNext := curr.children[sp]
+		if !hasNext {
+			next = &wordTrieNode{word: sp, children: make(map[string]*wordTrieNode)}
+			curr.children[sp] = next
+		} else {
+		}
```
njhale, line 221: unused `else`

.github/workflows/e2e-tests.yml
```diff
@@ -6,16 +6,21 @@ on:
   pull_request:
 jobs:
   e2e-tests:
+    strategy:
+      fail-fast: false
```
timflannagan, line 10: :+1:

.github/workflows/e2e-tests.yml
timflannagan: @njhale This might be a problem with some of the artifact parsing we've been talking about.

test/e2e/split/main.go
```diff
@@ -0,0 +1,226 @@
...
+	return prefixes
+}
+
+// wordTrie is a trie of word nodes, instead of individual characters.
+type wordTrie map[string]*wordTrieNode
+
+type wordTrieNode struct {
+	word     string
+	children map[string]*wordTrieNode
+}
```
timflannagan, line 205: Any particular reason you settled on a trie implementation vs. a more primitive implementation of sorting test specs and chunking arbitrarily?

estroz: Primarily to prevent one spec being run by two different chunks. Imagine running two specs defined by phrases "Foo" and "Foo bar" naively in two chunks: the first chunk (`^Foo.+`) would run both specs, while the second chunk (`^Foo bar.+`) would run only the latter spec. By using a word trie, the phrases are reduced down to just the word "Foo", which will run both specs in one chunk. At first the downside appears to be a less even distribution of chunk runtimes, since "Foo" and "Foo bar" will always be run together; however this downside also affects naive chunking because spec phrase prefixes are still shared.

test/e2e/split/main.go
```diff
@@ -0,0 +1,225 @@
...
+	flag.IntVar(&opts.numChunks, "chunks", 1, "Number of chunks to create focus regexps for")
+	flag.IntVar(&opts.printChunk, "print-chunk", 0, "Chunk to print a regexp for")
+	flag.BoolVar(&opts.printDebug, "print-debug", false, "Print all spec prefixes in non-regexp format. Use for debugging")
+	flag.Parse()
+
+	dir := flag.Arg(0)
+	if dir == "" {
+		exitIfErr(fmt.Errorf("test directory required as the argument"))
+	}
+
```
timflannagan, line 37: Do we need validation around the opts.printChunk >= opts.numChunks?

timflannagan: Disregard: I see that we have validation further down in the call stack. Any reason we could've bubble that validation call  to the driver function?

test/e2e/split/main.go
```diff
@@ -0,0 +1,225 @@
...
+		return nil, err
+	}
+	for _, match := range matches {
+		b, err := ioutil.ReadFile(match)
+		if err != nil {
+			return nil, err
+		}
+		specNames := topDescribeRE.FindAllSubmatch(b, -1)
+		if len(specNames) == 0 {
+			log.Printf("%s: found no top level describes, skipping", match)
```
timflannagan, line 100: It would be nice to control the log verbosity of this but I'm not too concerned if we're namely running this package in CI environments.

test/e2e/split/main.go
perdasilva: This is probably a really naive question, but, why not just use Ginkgo's parallelization: e.g. `ginkgo -procs=N`

test/e2e/split/main.go
estroz: Moved to main.

test/e2e/split/main.go
estroz: That's what I was thinking too. I'll leave a note for a follow up.

test/e2e/split/main.go
estroz: `-procs` gives you core parallelism, the actions matrix + chunks gives you core and runner parallelism. Runner parallelism is important here because we cannot choose machine size.

.github/workflows/e2e-tests.yml
timflannagan: Actually can we create a follow-up issue that tackles this? I don't think we need to block this PR merging, but I don't want to lose touch of this work.

test/e2e/split/main.go
timflannagan: It's also important to note that the way our CI environment is setup, we create a kind cluster everytime we spawn a new e2e process, so this PR simply builds on top of that existing setup and chunks test specs and feeds those test specs as focus regex groups that an individual kind cluster will then run. There's likely a cleaner solution going forward, but we'd have to shift around some of our current setup to get it working correctly.

timflannagan:
I think I'm comfortable enough with the implementation to avoid holding up this PR with any further commits given it helps improve QoL quite a bit. I would like to see us open an issue around aggregating the debug/junit/etc. artifacts into a single source, but that can be done asynchronously anyways.

/lgtm

timflannagan:
Interesting, the extra info-level logging is showing up as errors in the run output:

```
Run make e2e-local E2E_TEST_CHUNK=1 E2E_TEST_NUM_CHUNKS=4 E2E_NODES=2 ARTIFACTS_DIR=./artifacts/
Error: 2022/01/05 20:31:52 test/e2e/e2e_test.go: found no top level describes, skipping
Error: 2022/01/05 20:31:52 test/e2e/like_metric_matcher_test.go: found no top level describes, skipping
Error: 2022/01/05 20:31:52 test/e2e/setup_bare_test.go: found no top level describes, skipping
...
```

.github/workflows/e2e-tests.yml
```diff
@@ -9,16 +9,21 @@ on:
   workflow_dispatch:
 jobs:
   e2e-tests:
+    strategy:
```
timflannagan, line 12: Another thing I realize that might be a show stopper is the fact we'd have to remove the e2e-tests job as a required check to allow for the individual matrix strategy sub-jobs to run, increasing reviewer burden. Any ideas on how we'd solve this? It doesn't look like github has support for aggregating matrix results into a single result, so a composite action may/may not be usable. Maybe utilizing an action that labels PRs with a hold if any of the matrix runs fail? Thoughts?

timflannagan: cc @joelanford @perdasilva in case any of you have thoughts on how to solve this too.

timflannagan: https://github.community/t/get-status-of-parallel-jobs-into-a-single-webhook-payload/18077/2 looks fairly promising so maybe another job can be a workaround.

timflannagan: I ended up just tacking on another job to this existing workflow, and renaming the jobs defined in this workflow, such that the aggregate status job is named `e2e-tests` so we don't have to change any branch protection rules that are already defined for this repository.

.github/workflows/e2e-tests.yml
timflannagan: I'll create a follow-up ticket that tracks this work after this merges. We can add this to the #2571 tracking issue.

PR #2521, opened 2021-12-12
fix registry grpc unit test intermittent failure

akihikokuroda:

The `TestConnectionEvents` test checks all the grpc connection state transitions but some intermidiate states are not captured correctly.  This change checks the connection state moving to the `ready` state and check if there is any unexpected intermidiate states are observed in the transition.  

This error was seen in several CI build unit tests.
https://github.com/operator-framework/operator-lifecycle-manager/runs/4490427895?check_suite_focus=true
Closes #https://github.com/operator-framework/operator-lifecycle-manager/issues/2522

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/registry/grpc/source_test.go
+++ b/pkg/controller/registry/grpc/source_test.go
@@ -50,8 +50,8 @@ type FakeSourceSyncer struct {
 	History map[registry.CatalogKey][]connectivity.State
 
 	sync.Mutex
-	expectedEvents int
-	done           chan struct{}
+	expectedReadies int
+	done            chan struct{}
 }
 
 func (f *FakeSourceSyncer) sync(state SourceState) {
@@ -60,18 +60,20 @@ func (f *FakeSourceSyncer) sync(state SourceState) {
 		f.History[state.Key] = []connectivity.State{}
 	}
 	f.History[state.Key] = append(f.History[state.Key], state.State)
-	f.expectedEvents--
-	if f.expectedEvents == 0 {
+	if state.State == connectivity.Ready {
+		f.expectedReadies--
+	}
+	if f.expectedReadies == 0 {
 		f.done <- struct{}{}
 	}
 	f.Unlock()
 }
 
-func NewFakeSourceSyncer(expectedEvents int) *FakeSourceSyncer {
+func NewFakeSourceSyncer(expectedReadies int) *FakeSourceSyncer {
 	return &FakeSourceSyncer{
-		History:        map[registry.CatalogKey][]connectivity.State{},
-		expectedEvents: expectedEvents,
-		done:           make(chan struct{}),
+		History:         map[registry.CatalogKey][]connectivity.State{},
+		expectedReadies: expectedReadies,
+		done:            make(chan struct{}),
 	}
 }
 
@@ -84,11 +86,9 @@ func TestConnectionEvents(t *testing.T) {
 	test := func(tt testcase) func(t *testing.T) {
 		return func(t *testing.T) {
 			// start server for each catalog
-			totalEvents := 0
 			addresses := map[registry.CatalogKey]string{}
 
-			for catalog, events := range tt.expectedHistory {
-				totalEvents += len(events)
+			for catalog := range tt.expectedHistory {
 				serve, address, stop := server(&fakes.FakeQuery{})
 				addresses[catalog] = address
 				go serve()
@@ -96,9 +96,9 @@ func TestConnectionEvents(t *testing.T) {
 			}
 
 			// start source manager
-			syncer := NewFakeSourceSyncer(totalEvents)
+			syncer := NewFakeSourceSyncer(len(tt.expectedHistory))
 			sources := NewSourceStore(logrus.New(), 1*time.Second, 5*time.Second, syncer.sync)
-			ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+			ctx, cancel := context.WithTimeout(context.Background(), 20*time.Second)
 			defer cancel()
 			sources.Start(ctx)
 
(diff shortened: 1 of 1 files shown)


pkg/controller/registry/grpc/source_test.go
```diff
@@ -61,18 +61,20 @@ func (f *FakeSourceSyncer) sync(state SourceState) {
 		f.History[state.Key] = []connectivity.State{}
 	}
 	f.History[state.Key] = append(f.History[state.Key], state.State)
-	f.expectedEvents -= 1
-	if f.expectedEvents == 0 {
+	if state.State == connectivity.Ready {
+		f.expectedReadies -= 1
+	}
+	if f.expectedReadies == 0 {
```
timflannagan, line 67: I think I might need more coffee in me this morning but I'm having a hard time groking why this check is needed: it seems like we're transitioning from the test that ensures that we're encountering a specific set of connection state event histories, to a test that ensures we're eventually catching the Ready state and making some inferences about any missing events along the way. Is this the right read, or am I missing something entirely?

akihikokuroda: The test assumed that the sync function is called at every state changes of the grpc connection but it can miss some.  The GRPC connection is started at `conn, err := grpcConnection(address)` in `Add` function and the state change monitoring is started at `if source.Conn.WaitForStateChange(timer, state) {` in `watch` function.  Some state changes can happen between these 2 points so the sync function may not be called for every state changes. 
 The code changes are to make sure that all connection become the ready state and the connection doesn't have any unexpected states in the transition to the ready state.

awgreene: Thanks again for another PR @akihikokuroda!

I am a bit concerned with the proposed changes as there is no guarantee that the catalog is moving through phases as we expect. For example, I may have written the test expecting that the catalog end in the Ready phase but the test could pass if the catalog entered the "ready" phase prior to failing. I see two possible solutions:

1. Update the existing solution to ensure that the expected events occur sequentially and finish in the final expected state.
2. Fix the race condition in which history can be lost (I am unfamiliar with [this code](https://github.com/operator-framework/operator-lifecycle-manager/blob/c6cc1d7f706d355e34b1fc6523693bc13ceb4650/pkg/controller/registry/grpc/source.go#L169-L195) and not sure if this is possible).

awgreene: I would like to point out that I believe that the proposed changes are overall reasonable.

awgreene: After looking at the proposed changes a bit more - its clear that every catalog reaches the "ready" state and that the "captured" history doesn't contain any unexpected states. This seems fine for now, but it'd be great if we could remove the race condition.

awgreene: I am comfortable merging this if we create an issue capturing the race condition.

timflannagan: I think my preference is towards fixing the unit test vs. fixing the race condition, but I'd like to track fixing that potential race condition somewhere (e.g. a github issue) so it doesn't get swept under the rug. Any thoughts?

akihikokuroda: OK.  I agree to open an git issue for it.

akihikokuroda: I opened a git issue https://github.com/operator-framework/operator-lifecycle-manager/issues/2537 for the race condition investigation.

awgreene:
/approved

Great work on this @akihikokuroda

akihikokuroda:
I had to resolve conflict.  Then sanity errored.

pkg/controller/registry/grpc/source_test.go
```diff
@@ -60,18 +60,20 @@ func (f *FakeSourceSyncer) sync(state SourceState) {
 		f.History[state.Key] = []connectivity.State{}
 	}
 	f.History[state.Key] = append(f.History[state.Key], state.State)
-	f.expectedEvents--
-	if f.expectedEvents == 0 {
+	if state.State == connectivity.Ready {
+		f.expectedReadies -= 1
```
timflannagan, line 64: Nit: 
```suggestion
		f.expectedReadies--
```

pkg/controller/registry/grpc/source_test.go
```diff
@@ -84,21 +86,19 @@ func TestConnectionEvents(t *testing.T) {
 	test := func(tt testcase) func(t *testing.T) {
 		return func(t *testing.T) {
 			// start server for each catalog
-			totalEvents := 0
 			addresses := map[registry.CatalogKey]string{}
 
-			for catalog, events := range tt.expectedHistory {
-				totalEvents += len(events)
+			for catalog, _ := range tt.expectedHistory {
```
timflannagan, line 91: It looks like the sanity check is complaining that omitting the keys value variable is unnecessary: 
```suggestion
			for catalog := range tt.expectedHistory {
```

akihikokuroda: Thanks!  I update it.

pkg/controller/registry/grpc/source_test.go
akihikokuroda: Thanks! I change it.

timflannagan:
It looks like Alex misspelt the approve command so manually adding it :laughing:

akihikokuroda:
The e2e test failure is addressed in https://github.com/operator-framework/operator-lifecycle-manager/issues/2542

perdasilva:
@akihikokuroda could you please rebase

perdasilva:
@akihikokuroda I think you may need to run make format or make lint

PR #2523, opened 2021-12-13
Add system constraint provider logic to resolver

perdasilva:
Signed-off-by: Per G. da Silva <perdasilva@redhat.com>

This PR adds a `systemConstraintsProvider` to the `resolver`. It also modifies the resolver to apply any constraints comming from the `systemContraintsProvider`. Secondly, it add `initHooks` to the `OperatorStepResolver`. This allows the downstream to modify resolver behaviour.

For 4.10 we'd like to have a way for the resolver to respect the `maxOpenShiftVersion` property that can be provided by bundles to ensure that operator supports the cluster onto which it is being installed. This is done by injecting an OpenShift `systemConstraintsProvider` into the `satResolver` using an `initHook` on the downstream.

The init hooks provide a lightweight way of doing this from downstream without affecting the upstream or providing (in the upstream) a way for users to reach the `systemConstraintsProvider`. At least until we want to open this more generally.

*Note:* This PR must be merged sync'ed with the downstream before the downstream PR can be cut

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/registry/resolver/hooks.go
+++ b/pkg/controller/registry/resolver/hooks.go
@@ -1,0 +1,9 @@
+package resolver
+
+// stepResolverInitHook provides a way for the downstream
+// to modify the step resolver at creation time.
+// This is a bit of a hack to enable system constraints downstream
+// without affecting the upstream. We may want to clean this up when
+// either we have a more pluggable architecture; or system constraints
+// come to the upstream
+type stepResolverInitHook func(*OperatorStepResolver) error
--- a/pkg/controller/registry/resolver/resolver.go
+++ b/pkg/controller/registry/resolver/resolver.go
@@ -26,9 +26,10 @@ type OperatorResolver interface {
 }
 
 type SatResolver struct {
-	cache cache.OperatorCacheProvider
-	log   logrus.FieldLogger
-	pc    *predicateConverter
+	cache                     cache.OperatorCacheProvider
+	log                       logrus.FieldLogger
+	pc                        *predicateConverter
+	systemConstraintsProvider solver.ConstraintProvider
 }
 
 func NewDefaultSatResolver(rcp cache.SourceProvider, catsrcLister v1alpha1listers.CatalogSourceLister, logger logrus.FieldLogger) *SatResolver {
@@ -175,6 +176,25 @@ func (r *SatResolver) SolveOperators(namespaces []string, csvs []*v1alpha1.Clust
 	}
 
 	return operators, nil
+}
+
+// newBundleInstallableFromEntry converts an entry into a bundle installable with
+// system constraints applied, if they are defined for the entry
+func (r *SatResolver) newBundleInstallableFromEntry(entry *cache.Entry) (*BundleInstallable, error) {
+	bundleInstalleble, err := NewBundleInstallableFromOperator(entry)
+	if err != nil {
+		return nil, err
+	}
+
+	// apply system constraints if necessary
+	if r.systemConstraintsProvider != nil && !(entry.SourceInfo.Catalog.Virtual()) {
+		systemConstraints, err := r.systemConstraintsProvider.Constraints(entry)
+		if err != nil {
+			return nil, err
+		}
+		bundleInstalleble.constraints = append(bundleInstalleble.constraints, systemConstraints...)
+	}
+	return &bundleInstalleble, nil
 }
 
 func (r *SatResolver) getSubscriptionInstallables(sub *v1alpha1.Subscription, current *cache.Entry, namespacedCache cache.MultiCatalogOperatorFinder, visited map[*cache.Entry]*BundleInstallable) (map[solver.Identifier]solver.Installable, error) {
@@ -334,13 +354,13 @@ func (r *SatResolver) getBundleInstallables(preferredNamespace string, bundleSta
 			continue
 		}
 
-		bundleInstallable, err := NewBundleInstallableFromOperator(bundle)
+		bundleInstallable, err := r.newBundleInstallableFromEntry(bundle)
 		if err != nil {
 			errs = append(errs, err)
 			continue
 		}
 
-		visited[bundle] = &bundleInstallable
+		visited[bundle] = bundleInstallable
 
 		dependencyPredicates, err := r.pc.convertDependencyProperties(bundle.Properties)
 		if err != nil {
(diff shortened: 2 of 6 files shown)


pkg/controller/registry/resolver/resolver.go
```diff
@@ -172,6 +173,23 @@ func (r *SatResolver) SolveOperators(namespaces []string, csvs []*v1alpha1.Clust
 	return operators, nil
 }
 
+func (r *SatResolver) toBundleInstallable(entry *cache.Entry) (*BundleInstallable, error) {
```
anik120, line 176: I'm a little confused about why we're adding a `toBundleInstallable`, when we have a `getBundleInstallable` [here](https://github.com/operator-framework/operator-lifecycle-manager/pull/2523/files#R327)

Could we not have applied the system constraints in `getBundleInstallable`? Am I mis-reading this? Or is this to help keep it clean for the downstream code to call upon?

anik120 (changes requested):
@perdasilva I can see the commit here that modifies the kubeconfig flag. Should that be a separate PR if it's not related to what the PR title/summary talks about, unless it's absolutely necessary for that commit to be in this PR? 

Also, is the downstream PR ready? It's hard to grasp how the `init` hook will be used or what code changes here are motivated by the eventual downstream code changes without looking at that code beside this.

pkg/lib/operatorclient/client.go
```diff
@@ -146,23 +144,8 @@ type Client struct {
...
-		logger.Infof("Using in-cluster kube client config")
-		config, err = rest.InClusterConfig()
-	}
-
-	if err != nil {
-		logger.Fatalf("Cannot load config for REST client: %v", err)
-	}
-
+// NewClientFromConfig creates a kubernetes client
+func NewClientFromConfig(config *rest.Config) ClientInterface {
```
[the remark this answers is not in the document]
perdasilva: I'll fix these up - I'm moving these changes to a different PR though. Just to keep things tidy. I should have done that in the first place =S

pkg/controller/registry/resolver/resolver.go
perdasilva: I think this is ok. It fells like `getBundleInstallables` is very busy doing lots of different things - it also has two sources of entries (the stack and the filter queries it launches against the cache). Plus there's a difference in scope as well - `getBundleInstallables` gets a collection of BundleInstallables given the namespace, etc. `toBundleInstallable` converts an entry into a bundle installable.

perdasilva:
> @perdasilva I can see the commit here that modifies the kubeconfig flag. Should that be a separate PR if it's not related to what the PR title/summary talks about, unless it's absolutely necessary for that commit to be in this PR?
> 
> Also, is the downstream PR ready? It's hard to grasp how the `init` hook will be used or what code changes here are motivated by the eventual downstream code changes without looking at that code beside this.

@anik120 
I've removed the commit with the flag fixes. Here's a PR: https://github.com/perdasilva/operator-framework-olm/pull/1 that contains both up and downs stream changes. (the upstream part may slightly differ due to changes in response to reviewer comments - but should be fine).

pkg/controller/registry/resolver/resolver.go
```diff
@@ -26,9 +26,10 @@ type OperatorResolver interface {
 }
 
 type SatResolver struct {
-	cache cache.OperatorCacheProvider
-	log   logrus.FieldLogger
-	pc    *predicateConverter
+	cache                     cache.OperatorCacheProvider
+	log                       logrus.FieldLogger
+	pc                        *predicateConverter
+	systemConstraintsProvider solver.ConstraintProvider
```
dinhxuanvu, line 32: Now that Eric introduced the `predicateConverter`, would like to see if this `systemConstraintsProvider` can be nested under that but it doesn't seem to fit. Another field in SatResolver is a bit messy but it is not too bad if we want to get rid of it later on. Just a thought here. No blocker.

pkg/controller/registry/resolver/resolver.go
```diff
@@ -175,6 +176,23 @@ func (r *SatResolver) SolveOperators(namespaces []string, csvs []*v1alpha1.Clust
 	return operators, nil
 }
 
+func (r *SatResolver) toBundleInstallable(entry *cache.Entry) (*BundleInstallable, error) {
```
dinhxuanvu, line 179: I would like to see a comment here to explain what this func does and perhaps a better name for it as it is too vague. This func essentially adds cluster/system constraints to the installable bundles which is fine.

pkg/controller/registry/resolver/solver/constraint_provider.go
```diff
@@ -0,0 +1,15 @@
...
+
+// ConstraintProvider knows how to provide solver constraints for a given cache entry.
+type ConstraintProvider interface {
+	// Constraints returns a set of solver constraints for a cache entry.
+	Constraints(e *cache.Entry) ([]Constraint, error)
+}
+
+type ConstraintProviderFunc func(e *cache.Entry) ([]Constraint, error)
+
+func (c ConstraintProviderFunc) Constraints(e *cache.Entry) ([]Constraint, error) {
```
dinhxuanvu, line 13: Comment here would be nice to indicate this func extracts constraints from catalog entry (bundle).

dinhxuanvu (changes requested):
Looking good. A few comments.

pkg/controller/registry/resolver/resolver.go
perdasilva: I've had a crack at it. See if it's better.

pkg/controller/registry/resolver/solver/constraint_provider.go
perdasilva: It's not quite for extracting constraints from the bundle, but providing additional constraints. I've added some comments here and to the ConstraintProvider interface as well

dinhxuanvu: This interface is a bit more confusing than I thought but I got what you meant.

pkg/controller/registry/resolver/resolver.go
dinhxuanvu: I think the new name is better.

pkg/controller/registry/resolver/hooks.go
```diff
@@ -0,0 +1,9 @@
+package resolver
+
+// StepResolverInitHook provides a way for the downstream
+// to modify the step resolver at creation time.
+// This is a bit of a hack to enable system constraints downstream
+// without affecting the upstream. We may want to clean this up when
+// either we have a more pluggable architecture; or system constraints
+// come to the upstream
+type StepResolverInitHook func(*OperatorStepResolver) error
```
njhale, line 9: If the injection only occurs in-package, then this interface doesn't need to be public, right?

njhale (approved):
/approve

pkg/controller/registry/resolver/hooks.go
perdasilva: +1 - great catch - ty!

PR #2527, opened 2021-12-16
Magic catalog

perdasilva:
This PR instroduces `MagicCatalog`, an easy(er) way to deploy custom/arbitrary catalogs for e2e tests. I've also refactored the `utils_test.go` to `utils.go` and started taking the first steps to cut the utilities out into its own package, if possible.

At the moment it only includes a single implementation of the `FileBasedCatalogProvider`, which serves up the FBC given by a filesystem path (the `FileBasedFileBasedCatalogProvider`). Other providers could be imagined that give a more programmatic approach to catalog building...

This PR could probably be improved by bringing in registry classes (maybe the FileBasedCatalogProvider could provide some FileBasedCatalog obj, etc.). Curious to know what you guys think.

It's really hard to bring arbitrary 

**Reviewer Checklist**

### DIFF
--- a/test/e2e/magic_catalog.go
+++ b/test/e2e/magic_catalog.go
@@ -1,0 +1,280 @@
+package e2e
+
+import (
+	"context"
+	"fmt"
+
+	operatorsv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
+	k8serror "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	k8scontrollerclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	olmCatalogLabel    string = "olm.catalogSource"
+	catalogMountPath   string = "/opt/olm"
+	catalogServicePort int32  = 50051
+	catalogReadyState  string = "READY"
+)
+
+type MagicCatalog interface {
+	DeployCatalog(ctx context.Context) error
+	UndeployCatalog(ctx context.Context) []error
+}
+
+type magicCatalog struct {
+	fileBasedCatalog FileBasedCatalogProvider
+	kubeClient       k8scontrollerclient.Client
+	namespace        string
+	name             string
+	configMapName    string
+	serviceName      string
+	podName          string
+}
+
+// NewMagicCatalog creates an object that can deploy an arbitrary file-based catalog given by the FileBasedCatalogProvider
+// Keep in mind that there are limits to the configMaps. So, the catalogs need to be relatively simple
+func NewMagicCatalog(kubeClient k8scontrollerclient.Client, namespace string, catalogName string, provider FileBasedCatalogProvider) MagicCatalog {
+	return &magicCatalog{
+		fileBasedCatalog: provider,
+		kubeClient:       kubeClient,
+		namespace:        namespace,
+		name:             catalogName,
+		configMapName:    catalogName + "-configmap",
+		serviceName:      catalogName + "-svc",
+		podName:          catalogName + "-pod",
+	}
+}
+
+func (c *magicCatalog) DeployCatalog(ctx context.Context) error {
+
+	catalogSource := c.makeCatalogSource()
+	resourcesInOrderOfDeployment := []k8scontrollerclient.Object{
+		c.makeConfigMap(),
+		c.makeCatalogSourcePod(),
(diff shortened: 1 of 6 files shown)


dinhxuanvu (commented):
/lgtm

test/e2e/magic_catalog_test.go
```diff
@@ -0,0 +1,40 @@
...
+
+	BeforeEach(func() {
+		generatedNamespace = SetupGeneratedTestNamespace(genName("magic-catalog-e2e-"))
+	})
+
+	AfterEach(func() {
+		TeardownNamespace(generatedNamespace.GetName())
+	})
+
+	It("Deploys and Undeploys a File-based Catalog", func() {
```
njhale, line 25: Tests for the test harness 🙂

test/e2e/magic_catalog_test.go
```diff
@@ -0,0 +1,40 @@
...
+		// create dependencies
+		const catalogName = "test"
+		namespace := generatedNamespace.GetName()
+		kubeClient := ctx.Ctx().Client()
+		provider, err := NewFileBasedFiledBasedCatalogProvider("../test/e2e/testdata/fbc_catalog.json")
```
njhale, line 30: Personally, I'd rather see the FBC JSON literal in the test case itself than in a golden file, but this works for now.

njhale (approved):
Wow, thanks for putting this together! It should really help.

/approve

(reminder: we need to add @perdasilva to the approver set)

test/e2e/magic_catalog.go
```diff
@@ -0,0 +1,280 @@
...
+			SourceType: operatorsv1alpha1.SourceTypeGrpc,
+			Address:    fmt.Sprintf("%s.%s.svc:50051", c.serviceName, c.namespace),
+		},
+	}
+}
+
+func (c *magicCatalog) makeCatalogSourcePod() *corev1.Pod {
+
+	const (
+		image                  = "quay.io/operator-framework/upstream-opm-builder"
```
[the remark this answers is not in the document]
perdasilva: that's a very good point - we could think of using env vars or something like that

test/e2e/magic_catalog_test.go
perdasilva: tbh - I'm not a huge fan of big string literals, especially when we can't nicely format them due to code formatting or indentation, etc. For small stuff sure, for bigger files it's start to get awkward for _me_. But this is a taste thing. I've tried to create an abstraction through this provider interface where we can plug-in different fbc sources (file, string, etc.). When I have a little more time, I'd like to have a way to create the fbcs programatically. Might be even more readable that way...idk...

test/e2e/magic_catalog.go
```diff
@@ -0,0 +1,280 @@
...
+}
+
+func (c *magicCatalog) makeConfigMap() *corev1.ConfigMap {
+	isImmutable := true
+	return &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      c.configMapName,
+			Namespace: c.namespace,
+		},
+		Immutable: &isImmutable,
```
timflannagan, line 158: Super unimportant: we likely have a test helper function (or at least we should) that returns a pointer to a bool somewhere.

test/e2e/magic_catalog.go
```diff
@@ -0,0 +1,280 @@
...
+
+	if err != nil {
+		return c.cleanUpAfter(ctx, err)
+	}
+
+	return nil
+}
+
+func (c *magicCatalog) UndeployCatalog(ctx context.Context) []error {
+	var errs []error = nil
```
timflannagan, line 97: nit: this appears to be redundant as `var errs []error` should be equivalent to an empty slice.

test/e2e/util.go
```diff
@@ -77,6 +77,15 @@ func newPMClient() pmversioned.Interface {
...
 }
 
+type cleanupFunc func()
+
+// waitFor wraps wait.Pool with default polling parameters
+func waitFor(fn func() (bool, error)) error {
+	return wait.Poll(pollInterval, pollDuration, func() (bool, error) {
+		return fn()
+	})
+}
```
timflannagan, line 87: :+1:

timflannagan (approved):
I think there's a good chunk of potential linting errors (the golangci-lint work we just merged excludes the test/e2e package for now, but we'll remove that exclusion over time) that the sanity check isn't complaining about, so I'm not super concerned. I don't have any problem with the implementation, and we should even have an issue open for switching to FBC for test fixtures open somewhere. Thanks for renaming that test utilities package, I've had the urge to rename that for a while now.

/lgtm

test/e2e/magic_catalog.go
perdasilva: right! old habits hehehe

PR #2528, opened 2021-12-16
chore(github): run e2e workflow against master periodically

njhale:

Run the e2e GitHub workflow against master periodically, twice per day, at 00:30 and 12:30 ET.

In order to effectively judge if a failing test is due to changes introduced by a given PR, we need a baseline to compare against; i.e. was this test failing in master before this change? how often? is this actually a flake? By making the e2e tests periodic, we can start building a historical set of test artifacts to act as this baseline.

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -1,5 +1,7 @@
 name: e2e
 on:
+  schedule:
+    - cron: '30 5,17 * * *' # run this every day at 5:30 and 17:30 UTC (00:30 and 12:30 ET)
   push:
     branches:
       - master


.github/workflows/e2e-tests.yml
```diff
@@ -1,5 +1,7 @@
 name: e2e
 on:
+  schedule:
+    - cron: '30 5,17 * * *' # run this every day at 5:30 and 17:30 UTC (00:30 and 12:30 ET)
```
timflannagan, line 4: How do you feel about `0 */12 * * *` for readability?

njhale: I was hoping to keep things in ET. I like getting a run in over lunch.

timflannagan: Yeah that's reasonable to me. We may want to re-evaluate the twice-a-day frequency at some point if it's too costly over the long term, but likely fine in the short term to help reduce the overall CI noise and create attempt to create metrics over time.

PR #2536, opened 2021-12-22
Integrate golangci-lint to run aggregate linting checks for CI

timflannagan:

**Reviewer Checklist**


.golangci.yaml
```diff
@@ -0,0 +1,30 @@
+run:
+  timeout: 5m
+  skip-dirs:
+  - pkg/lib
+  - pkg/api
+  - pkg/fakes
+  - pkg/package-server/apis
+  - test/e2e
```
timflannagan, line 8: Removing this can be done in a follow-up PR.

.golangci.yaml
```diff
@@ -0,0 +1,30 @@
...
+  - gofmt
+  - goimports
+  - importas
+  - misspell
+  - stylecheck
+  - tparallel
+  - unconvert
+  - whitespace
+  disable:
+  - errcheck
```
timflannagan, line 22: Removing this can be done in a follow-up PR.

.golangci.yaml
```diff
@@ -0,0 +1,30 @@
...
+
+issues:
+  max-issues-per-linter: 0
+  max-same-issues: 0
```
timflannagan, line 26: Setting both of these configuration options limit the number of linting violations that are discovered at a time. Setting the values to `0` disables that behavior.

cmd/catalog/main.go
```diff
@@ -69,7 +69,7 @@ var (
 	tlsCertPath = flag.String(
 		"tls-cert", "", "Path to use for certificate key (requires tls-key)")
 
-	profiling = flag.Bool("profiling", false, "deprecated")
+	_ = flag.Bool("profiling", false, "deprecated")
```
timflannagan, line 72: This flag is currently unused right now since we merge the profiling/metrics/health/etc. endpoints last minor release. I thought this was the easiest solution, vs. updating the .golangci-lint.yaml configuration and explicitly allowlisting this flag.

pkg/controller/operators/olm/operator.go
```diff
@@ -1718,6 +1712,10 @@ func (a *Operator) transitionCSVState(in v1alpha1.ClusterServiceVersion) (out *v
 
 	// Check for intersecting provided APIs in intersecting OperatorGroups
 	allGroups, err := a.lister.OperatorsV1().OperatorGroupLister().List(labels.Everything())
+	if err != nil {
+		logger.WithError(err).Warn("failed to list operatorgroups")
+		return
+	}
```
timflannagan, line 1718: Note: need to double-check whether this is the correct behavior. It wasn't clear to me poking around whether I should be returning a syncError here.

.golangci.yaml
```diff
@@ -0,0 +1,30 @@
...
+  - pkg/fakes
+  - pkg/package-server/apis
+  - test/e2e
+
+linters:
+  enable:
+  - depguard
+  - gofmt
+  - goimports
+  - importas
```
timflannagan, line 15: I'd like to tackle adding a more fleshed out importas configuration ( e.g. enforcing consistency with corev1/metav1/operatorsv1alpha1/etc. package alias') as a follow-up.

dinhxuanvu (commented):
Looking good. Just a quick question.

.golangci.yaml
dinhxuanvu: Can you explain the exclusion for these paths? I can see why we don't care about fakes and test but the rest are good to have.

timflannagan: Yeah, definitely - I had a couple of reasons:

- Aim to reduce the reviewer burden with an already large changeset;
- Get the main infrastructure in place so we golangci-lint can identify linting issues ahead of time to reduce reviewer responsibilities going forward vs. fixing all lints all at once;
- I was already fairly burnt out identifying and manually fixing all the linting violations and didn't want to hold this up further, as I'm less concerned about test code

dinhxuanvu (approved):
/approve

pkg/controller/operators/catalog/subscription/state.go
```diff
@@ -243,7 +243,6 @@ func (c *catalogHealthState) UpdateHealth(now *metav1.Time, client clientv1alpha
 	var known CatalogHealthKnownState
 	switch {
 	case missingTargeted:
-		healthy = false
```
awgreene, line 246: Why was this removed?

awgreene: Ah, it isn't used anywhere below.

awgreene (approved):
/approve

Great work on this @timflannagan

Issue #2540, opened 2021-12-24
e2e "ClusterServiceVersion when a CustomResourceDefinition was installed alongside a ClusterServiceVersion" test failures 


akihikokuroda:
## Bug Report
e2e "ClusterServiceVersion when a CustomResourceDefinition was installed alongside a ClusterServiceVersion" failure 

https://github.com/operator-framework/operator-lifecycle-manager/runs/4622357457?check_suite_focus=true#step:4:9553

[Fail] ClusterServiceVersion when a CustomResourceDefinition was installed alongside a ClusterServiceVersion [AfterEach] can satisfy an associated ClusterServiceVersion's ownership requirement 
/home/runner/work/operator-lifecycle-manager/operator-lifecycle-manager/test/e2e/csv_e2e_test.go:129

[Fail] ClusterServiceVersion when a CustomResourceDefinition was installed alongside a ClusterServiceVersion [BeforeEach] when an unassociated ClusterServiceVersion in different namespace owns the same CRD can satisfy the unassociated ClusterServiceVersion's ownership requirement 
/home/runner/work/operator-lifecycle-manager/operator-lifecycle-manager/test/e2e/csv_e2e_test.go:123

**What did you do?**
e2e test in the CI build failed.

**What did you expect to see?**
No failure

**What did you see instead? Under which circumstances?**
Failure above.

The first error is a timeout removing the namespace during the test clean up.  The second error is caused by the uncleaned resources by the first error. 

**Environment**
* operator-lifecycle-manager version:

* Kubernetes version information:

* Kubernetes cluster kind: 

**Possible Solution**

**Additional context**

PR #2541, opened 2021-12-24
e2e CSV test - use unique name resources for each test

akihikokuroda:

The "ClusterServiceVersion when a CustomResourceDefinition was installed alongside a ClusterServiceVersion" tests uses the same namespace name for each test.  It creates and remove the namespace every time.  It doesn't delete the resources in the namespace before it delete the namespace.  It may be causing the longer delete time of the namespace and the timeout of the test.  The tests executed after the timeout fail because the uncleaned namespace and resources in it.
 
1. use the unique namespace, resource names for each test.
2. delete the resources created in the test before the deletion of the namespace 

Closes https://github.com/operator-framework/operator-lifecycle-manager/issues/2540
**Reviewer Checklist**

### DIFF
--- a/test/e2e/csv_e2e_test.go
+++ b/test/e2e/csv_e2e_test.go
@@ -65,21 +65,25 @@ var _ = Describe("ClusterServiceVersion", func() {
 		var (
 			ns  corev1.Namespace
 			crd apiextensionsv1.CustomResourceDefinition
+			og operatorsv1.OperatorGroup
+			apiname string
+			apifullname string
 		)
 
 		BeforeEach(func() {
 			ns = corev1.Namespace{
 				ObjectMeta: metav1.ObjectMeta{
-					Name: "test-namespace-1",
+					Name: genName("test-namespace-"),
 				},
 			}
+
 			Eventually(func() error {
 				return ctx.Ctx().Client().Create(context.Background(), &ns)
 			}).Should(Succeed())
 
-			og := operatorsv1.OperatorGroup{
+			og = operatorsv1.OperatorGroup{
 				ObjectMeta: metav1.ObjectMeta{
-					Name:      fmt.Sprintf("%s-operatorgroup", ns.GetName()),
+					Name:      genName(fmt.Sprintf("%s-operatorgroup-", ns.GetName())),
 					Namespace: ns.GetName(),
 				},
 				Spec: operatorsv1.OperatorGroupSpec{
@@ -90,9 +94,11 @@ var _ = Describe("ClusterServiceVersion", func() {
 				return ctx.Ctx().Client().Create(context.Background(), &og)
 			}).Should(Succeed())
 
+			apiname = genName("api")
+			apifullname = apiname + "s.example.com"
 			crd = apiextensionsv1.CustomResourceDefinition{
 				ObjectMeta: metav1.ObjectMeta{
-					Name: "tests.example.com",
+					Name: apifullname,
 					Annotations: map[string]string{
 						"operatorframework.io/installed-alongside-0": fmt.Sprintf("%s/associated-csv", ns.GetName()),
 					},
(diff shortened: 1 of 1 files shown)


test/e2e/csv_e2e_test.go
```diff
@@ -90,9 +93,11 @@ var _ = Describe("ClusterServiceVersion", func() {
 				return ctx.Ctx().Client().Create(context.Background(), &og)
 			}).Should(Succeed())
 
+			apiname := genName("api")
```
timflannagan, line 96: Does this need to be `apiname = genName("api")` as we declared apiname as a top-level variable for this when container?

test/e2e/csv_e2e_test.go
```diff
@@ -306,7 +320,7 @@ var _ = Describe("ClusterServiceVersion", func() {
 				}
 				Expect(ctx.Ctx().Client().Create(context.Background(), &ns)).To(Succeed())
 
-				og := operatorsv1.OperatorGroup{
+				og = operatorsv1.OperatorGroup{
```
timflannagan, line 323: Any idea on whether we may need to also auto-generate the "test-namespace-2" that gets created [above this](https://github.com/operator-framework/operator-lifecycle-manager/pull/2541/files#diff-e94bc9e5ebd2b6db20f851288184630432208a5a772cf0515935707ad3660799R318) variable declaration in the BeforeEach clause to avoid any potential flakes down the line around attempting to create a namespace/CRD/etc. that already exists (e.g. it wasn't properly cleaned up for whatever reason)? 

If I'm reading this code right, we'll have the top-level BeforeEach create a testing namespace (that gets autogenerated through the `genName(...)` call), and then we'll create a `test-namespace-2` namespace to test collision for each test spec defined in this nested when container.

timflannagan:
@akihikokuroda Can you also rebase when you get a chance? I saw the unit test failure, but we just merged those changes since this PR has been opened.

test/e2e/csv_e2e_test.go
akihikokuroda: Yes, I'll move the apiname to top and remove ':'.

test/e2e/csv_e2e_test.go
akihikokuroda: Yes, it's better to be with the generated name.  I'll change it.  Thanks!

test/e2e/csv_e2e_test.go
```diff
@@ -182,6 +192,8 @@ var _ = Describe("ClusterServiceVersion", func() {
 					Status:  operatorsv1alpha1.RequirementStatusReasonPresent,
 				},
 			))
+
+			Expect(ctx.Ctx().Client().Delete(context.Background(), &associated)).To(Succeed())
```
timflannagan, line 196: Wrap this in an Eventually?

test/e2e/csv_e2e_test.go
```diff
@@ -240,6 +252,8 @@ var _ = Describe("ClusterServiceVersion", func() {
 			Eventually(func() error {
 				return ctx.Ctx().Client().Get(context.Background(), client.ObjectKeyFromObject(&unassociated), &unassociated)
 			}).Should(WithTransform(k8serrors.IsNotFound, BeTrue()))
+
+			Expect(ctx.Ctx().Client().Delete(context.Background(), &associated)).To(Succeed())
```
timflannagan, line 256: Wrap in an Eventually?

akihikokuroda: I'll add it.  Thanks!

timflannagan:
@akihikokuroda You may need to also rebase against master too.

test/e2e/csv_e2e_test.go
akihikokuroda: I'll add it. Thanks!

awgreene:
@akihikokuroda could you rebase this test as well?

timflannagan:
@akihikokuroda Apologies - we _just_ merged some e2e improvements that parallelize the e2e suite into their own jobs, so the overall e2e execution should drop from ~1h to ~20m. Can you rebase your PR once more, and I'll tag it once it passes the checks?

timflannagan:
Longer term, we need to solve the situation where branches constantly need to be rebased when changes land in the master branch as it's not the greatest UX for contributors.

timflannagan (approved):
Thanks

/lgtm

PR #2547, opened 2022-01-05
e2e - add Eventually for config injection check

### LINKED ISSUE #2507: e2e Subscription - creation with pod config [It] fails intermittently
## Bug Report

I saw this error in the e2e test in the CI.  This error can be seen more often by adding "time.Sleep(10 * time.Second)" in the "case v1alpha1.CSVPhaseInstalling:" part in transitionCSVState function in the olm/operator.go.

Subscription 
/home/runner/work/operator-lifecycle-manager/operator-lifecycle-manager/test/e2e/subscription_e2e_test.go:52
  creation with pod config [It]
  /home/runner/work/operator-lifecycle-manager/operator-lifecycle-manager/test/e2e/subscription_e2e_test.go:1203

          Error Trace:        subscription_e2e_test.go:2874
                              subscription_e2e_test.go:1350
                              runner.go:113
                              runner.go:64
                              it_node.go:26
                              spec.go:215
                              spec.go:138
                              spec_runner.go:200
                              spec_runner.go:170
                              spec_runner.go:66
                              suite.go:79
                              ginkgo_dsl.go:238
                              ginkgo_dsl.go:221
                              e2e_test.go:59
          Error:              Should be true
          Test:               Subscription creation with pod config
          Messages:           Volume name=test-volume-7gxcg not injected

  /home/runner/work/operator-lifecycle-manager/operator-lifecycle-manager/vendor/github.com/stretchr/testify/assert/assertion_format.go:714
**What did you do?**
Error in CI e2e test

Should not fail

Failure in CI e2e test

**Environment**
* operator-lifecycle-manager version: Pull request branches 

* Kubernetes cluster kind: 

**Possible Solution**

**Additional context**

akihikokuroda:

There is a timing gap between the `operatorsv1alpha1.SubscriptionStateAtLatest` being set in the subscription state and the configs are injected into the CSV.

This PR adds an Eventually waiting for the config injection. 

Closes #2507 

**Reviewer Checklist**

### DIFF
--- a/test/e2e/subscription_e2e_test.go
+++ b/test/e2e/subscription_e2e_test.go
@@ -1313,14 +1313,18 @@ var _ = Describe("Subscription", func() {
 		require.NoError(GinkgoT(), err)
 		require.NotNil(GinkgoT(), subscription)
 
-		csv, err := fetchCSV(crClient, subscription.Status.CurrentCSV, generatedNamespace.GetName(), buildCSVConditionChecker(operatorsv1alpha1.CSVPhaseSucceeded))
-		require.NoError(GinkgoT(), err)
-
 		proxyEnv := proxyEnvVarFunc(GinkgoT(), config)
 		expected := podEnv
 		expected = append(expected, proxyEnv...)
 
-		checkDeploymentWithPodConfiguration(GinkgoT(), kubeClient, csv, podConfig.Env, podConfig.Volumes, podConfig.VolumeMounts, podConfig.Tolerations, podConfig.Resources)
+		Eventually(func() error {
+			csv, err := fetchCSV(crClient, subscription.Status.CurrentCSV, generatedNamespace.GetName(), buildCSVConditionChecker(operatorsv1alpha1.CSVPhaseSucceeded))
+			if err != nil {
+				return err
+			}
+
+			return checkDeploymentWithPodConfiguration(kubeClient, csv, podConfig.Env, podConfig.Volumes, podConfig.VolumeMounts, podConfig.Tolerations, podConfig.Resources)
+		}).Should(Succeed())
 	})
 
 	It("creation with nodeSelector config", func() {
@@ -2743,14 +2747,14 @@ func checkDeploymentHasPodConfigNodeSelector(t GinkgoTInterface, client operator
 	return nil
 }
 
-func checkDeploymentWithPodConfiguration(t GinkgoTInterface, client operatorclient.ClientInterface, csv *operatorsv1alpha1.ClusterServiceVersion, envVar []corev1.EnvVar, volumes []corev1.Volume, volumeMounts []corev1.VolumeMount, tolerations []corev1.Toleration, resources *corev1.ResourceRequirements) {
+func checkDeploymentWithPodConfiguration(client operatorclient.ClientInterface, csv *operatorsv1alpha1.ClusterServiceVersion, envVar []corev1.EnvVar, volumes []corev1.Volume, volumeMounts []corev1.VolumeMount, tolerations []corev1.Toleration, resources *corev1.ResourceRequirements) error {
 	resolver := install.StrategyResolver{}
 
 	strategy, err := resolver.UnmarshalStrategy(csv.Spec.InstallStrategy)
-	require.NoError(t, err)
+	Expect(err).NotTo(HaveOccurred())
 
 	strategyDetailsDeployment, ok := strategy.(*operatorsv1alpha1.StrategyDetailsDeployment)
-	require.Truef(t, ok, "could not cast install strategy as type %T", strategyDetailsDeployment)
+	Expect(ok).To(BeTrue(), "could not cast install strategy as type %T", strategyDetailsDeployment)
 
 	findEnvVar := func(envVar []corev1.EnvVar, name string) (foundEnvVar *corev1.EnvVar, found bool) {
 		for i := range envVar {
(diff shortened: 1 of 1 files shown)


test/e2e/subscription_e2e_test.go
```diff
@@ -2813,48 +2816,58 @@ func checkDeploymentWithPodConfiguration(t GinkgoTInterface, client operatorclie
 		return
 	}
 
-	check := func(container *corev1.Container) {
+	check := func(container *corev1.Container) error {
```
timflannagan, line 2819: :+1: I've had the similar urge to change this signature when tackling other e2e work.

test/e2e/subscription_e2e_test.go
```diff
@@ -1313,14 +1314,16 @@ var _ = Describe("Subscription", func() {
...
-		require.NoError(GinkgoT(), err)
-
 		proxyEnv := proxyEnvVarFunc(GinkgoT(), config)
 		expected := podEnv
 		expected = append(expected, proxyEnv...)
 
-		checkDeploymentWithPodConfiguration(GinkgoT(), kubeClient, csv, podConfig.Env, podConfig.Volumes, podConfig.VolumeMounts, podConfig.Tolerations, podConfig.Resources)
+		Eventually(func() error {
+			csv, err := fetchCSV(crClient, subscription.Status.CurrentCSV, generatedNamespace.GetName(), buildCSVConditionChecker(operatorsv1alpha1.CSVPhaseSucceeded))
+			require.NoError(GinkgoT(), err)
```
timflannagan, line 1323: small nit: can we return the error here instead of making assertions through the require package?

test/e2e/subscription_e2e_test.go
```diff
@@ -2743,7 +2746,7 @@ func checkDeploymentHasPodConfigNodeSelector(t GinkgoTInterface, client operator
 	return nil
 }
 
-func checkDeploymentWithPodConfiguration(t GinkgoTInterface, client operatorclient.ClientInterface, csv *operatorsv1alpha1.ClusterServiceVersion, envVar []corev1.EnvVar, volumes []corev1.Volume, volumeMounts []corev1.VolumeMount, tolerations []corev1.Toleration, resources *corev1.ResourceRequirements) {
+func checkDeploymentWithPodConfiguration(t GinkgoTInterface, client operatorclient.ClientInterface, csv *operatorsv1alpha1.ClusterServiceVersion, envVar []corev1.EnvVar, volumes []corev1.Volume, volumeMounts []corev1.VolumeMount, tolerations []corev1.Toleration, resources *corev1.ResourceRequirements) error {
```
timflannagan, line 2749: Quick note: it looks like we might be able to lightly refactor this helper function and remove the GinkgoTInterface parameter by removing any mentions of the require package throughout the function's flow, and given we already changed a couple of the require statements to be error checks throughout this PR, it shouldn't be all that painful. It would be nice to migrate any testify statements to gomega expressions throughout the e2e suite longer term.

I think making the required changes would be pretty quick, but it also increases the scope of this PR, so I'll leave that up to whether you want to tackle that here.

test/e2e/subscription_e2e_test.go
```diff
@@ -2813,48 +2816,58 @@ func checkDeploymentWithPodConfiguration(t GinkgoTInterface, client operatorclie
...
-	check := func(container *corev1.Container) {
+	check := func(container *corev1.Container) error {
 		for _, e := range envVar {
 			existing, found := findEnvVar(container.Env, e.Name)
-			require.Truef(t, found, "env variable name=%s not injected", e.Name)
-			require.NotNil(t, existing)
+			if !found || existing == nil {
+				return errors.New(fmt.Sprintf("env variable name=%s not injected", e.Name))
+			}
 			require.Equalf(t, e.Value, existing.Value, "env variable value does not match %s=%s", e.Name, e.Value)
```
timflannagan, line 2825: Mentioned above: if we're going to be working towards error handling in this inline check function, then I think we should avoid using any testify assertions, and instead returning an error if the conditional check had failed. WYDT?

akihikokuroda: I agree.  Let me try.

test/e2e/subscription_e2e_test.go
akihikokuroda: I'll look into it.  Thanks!

test/e2e/subscription_e2e_test.go
akihikokuroda: Yes, I'll change it.  Thanks!

test/e2e/subscription_e2e_test.go
```diff
@@ -2743,14 +2748,14 @@ func checkDeploymentHasPodConfigNodeSelector(t GinkgoTInterface, client operator
 	return nil
 }
 
-func checkDeploymentWithPodConfiguration(t GinkgoTInterface, client operatorclient.ClientInterface, csv *operatorsv1alpha1.ClusterServiceVersion, envVar []corev1.EnvVar, volumes []corev1.Volume, volumeMounts []corev1.VolumeMount, tolerations []corev1.Toleration, resources *corev1.ResourceRequirements) {
+func checkDeploymentWithPodConfiguration(client operatorclient.ClientInterface, csv *operatorsv1alpha1.ClusterServiceVersion, envVar []corev1.EnvVar, volumes []corev1.Volume, volumeMounts []corev1.VolumeMount, tolerations []corev1.Toleration, resources *corev1.ResourceRequirements) error {
```
timflannagan, line 2751: I'm surprised there's only a single call site for this function (maybe it's worth purging this function in the future?) but thanks for making the requisite refactoring to drop the extra parameter.

test/e2e/subscription_e2e_test.go
```diff
@@ -2813,48 +2818,58 @@ func checkDeploymentWithPodConfiguration(t GinkgoTInterface, client operatorclie
 		return
 	}
 
-	check := func(container *corev1.Container) {
+	check := func(container *corev1.Container) error {
 		for _, e := range envVar {
 			existing, found := findEnvVar(container.Env, e.Name)
-			require.Truef(t, found, "env variable name=%s not injected", e.Name)
-			require.NotNil(t, existing)
-			require.Equalf(t, e.Value, existing.Value, "env variable value does not match %s=%s", e.Name, e.Value)
```
timflannagan, line 2821: general comment (as it also applies to the code changes made further down): do we risk losing debug information by removing the debug line when an assertion fails?

akihikokuroda: I restore some of the error messages.  Thanks!

test/e2e/subscription_e2e_test.go
```diff
@@ -2813,48 +2818,58 @@ func checkDeploymentWithPodConfiguration(t GinkgoTInterface, client operatorclie
...
+			if !found || existing == nil {
+				return errors.New(fmt.Sprintf("env variable name=%s not injected", e.Name))
```
timflannagan, line 2825: we can simply return a formatted error string through `return fmt.Errorf("env variable name=%s not injected", e.Name)`.

timflannagan:
I had one more comment for moving any `errors.New(fmt.Sprintf(...))` to simply `fmt.Errorf(...)` which would remove the `errors` package import as well (it looks like there are a couple of instances here). Otherwise, this looks ready to go imo.

test/e2e/subscription_e2e_test.go
akihikokuroda: Thanks!  I'll update.

test/e2e/subscription_e2e_test.go
```diff
@@ -2813,48 +2817,58 @@ func checkDeploymentWithPodConfiguration(t GinkgoTInterface, client operatorclie
...
+			}
+			Expect(*existing).Should(Equal(toleration), "Toleration=%v does not match expected Toleration=%v", existing, toleration)
 		}
 
 		for i := range deployment.Spec.Template.Spec.Containers {
-			check(&deployment.Spec.Template.Spec.Containers[i])
+			err = check(&deployment.Spec.Template.Spec.Containers[i])
+			if err != nil {
+				return err
+			}
```
awgreene, line 2868: Non-blocking nit:
```suggestion
			if err = check(&deployment.Spec.Template.Spec.Containers[i]); err != nil {
				return err
			}
```

awgreene (approved):
Great work on this.

PR #2551, opened 2022-01-06
Update OWNERS file

awgreene:


timflannagan:
Force merging as this PR only affects the OWNERS file and we don't need to continuously run the e2e-tests workflow to get a successful check for this be automatically merged.

PR #2552, opened 2022-01-06
Extend Step to surface whether a manifest is optional

fgiloux:

This pull request adds some logic for avoiding failure of the InstallPlan when a manifest provided by a bundle cannot be installed and has been flagged as optional through a bundle property. [Here is the enhancement proposal](https://github.com/operator-framework/enhancements/blob/master/enhancements/optional-manifest.md)

This PR relies on two other PRs for the api and the operator-registry repositories.
https://github.com/operator-framework/api/pull/209
https://github.com/operator-framework/operator-registry/pull/893

I was not too sure how to proceed in such a case and have created the PRs at roughly the same time. It also means that for the code in this PR to work it needs to have the code of the other PRs available. For my tests I used "replace" pointing to a local repository in go.mod 

VerticalPodAutoscaler, ServiceMonitors for instances are custom resources and may not be available on all clusters. Operator authors may want to take benefit of these mechanisms for clusters having them but not having the installation fail on clusters without them as they may be seen as nice to have. This is what allows this PR.

See [the enhancement proposal](https://github.com/operator-framework/enhancements/blob/master/enhancements/optional-manifest.md) for details.

**Reviewer Checklist**


timflannagan:
>I was not too sure how to proceed in such a case and have created the PRs at roughly the same time. It also means that for the code in this PR to work it needs to have the code of the other PRs available. For my tests I used "replace" pointing to a local repository in go.mod

You can try and cut a release in your API fork and point the go.mod to your fork using the replace directive, and we can make a TODO comment that we remove that replace pin before merging this PR.

timflannagan (commented):
Can we separate the go.mod + vendor changes into their own commit, so viewing the full diff is a bit easier? One of these days we'll remove the vendor directory entirely from OLM, so this wouldn't be needed.

fgiloux:
> Can we separate the go.mod + vendor changes into their own commit, so viewing the full diff is a bit easier? One of these days we'll remove the vendor directory entirely from OLM, so this wouldn't be needed.

I have separated the commits and done a rebase

fgiloux:
Hi @timflannagan any further comment on this pull request?

fgiloux:
@timflannagan after the API rebase we now have: the API depending on k8s.io/api v0.23.0 whereas OLM was on k8s.io/api v0.22.2 and breaks when the dependency is bumped. This was introduced by: https://github.com/operator-framework/api/commit/12a70c0e4176b7ec7922e2ce3a6b3c94b4f3062d
How do we move forward?

go.mod
```diff
@@ -223,23 +222,28 @@ require (
...
+	sigs.k8s.io/yaml v1.3.0 // indirect
 )
 
 replace (
 	// controller runtime
 	github.com/openshift/api => github.com/openshift/api v0.0.0-20211014063134-be2a7fb8aa44
 	github.com/openshift/client-go => github.com/openshift/client-go v0.0.0-20200326155132-2a6cd50aedd0 // release-4.5
 
+	// Temporary for optional manifests before the changes get merged in the API and operator-registry repo
+	github.com/operator-framework/api v0.11.1 => github.com/fgiloux/api v0.10.8-0.20220118162926-a11336b43231
```
timflannagan, line 244: We can likely update this now to pick up the new commit through a psuedoversion tag, or reach out to the team-operator-fw slack channel for cutting a new o-f/api minor version release.

timflannagan:
@fgiloux It looks like the CI failures are complaining about various breaking changes that happen in the v0.23.x kube dependencies. We can either tackle bumping OLM's kube dependencies, and making the requisite changes to the codebase in another PR, or update the root go.mod and utilize the `replace` stanza to ensure we're using the v0.22.x kube dependencies. WYDT?

fgiloux:
> It looks like the CI failures are complaining about various breaking changes that happen in the v0.23.x kube dependencies.

Yes it is what I was saying earlier.

>  We can either tackle bumping OLM's kube dependencies, and making the requisite changes to the codebase in another PR, or update the root go.mod and utilize the replace stanza to ensure we're using the v0.22.x kube dependencies. WYDT?

I think that bumping OLM dependencies is required by any way, not just for this PR but for anything that needs to import a newer version of the API and for compatibility with newer Kubernetes versions. I will open an issue and start looking at it.

fgiloux:
Hi @timflannagan based on o-f api v0.13.0 all e2e tests passed except 1. I ran it locally and it was successful.
One thing is that I had to downgrade grpc version to v1.40.0 in operator-registry due to [the issue you identified](https://github.com/operator-framework/operator-lifecycle-manager/pull/2423#pullrequestreview-819616924).
Can you please take a look?

pkg/controller/registry/resolver/steps.go
```diff
@@ -230,3 +284,39 @@ func NewServiceAccountStepResources(csv *v1alpha1.ClusterServiceVersion, catalog
...
 	return rbacSteps, nil
 }
+
+type optionalManifests struct {
+	Files []string `json:"manifests"`
+}
+
+// isOptional returns a func giving whether a resource is in the list of optional manifests
+// or false if any issue occurs.
+func isOptional(properties []*api.Property) func(key string) bool {
```
dinhxuanvu, line 294: This func is doing operation that is not matching to what is described in the EP. In the EP, it is file-name-based identifier but for this implementation, this is a domain-based identifier. If this implementation is the intended choice, then the EP should be updated so that reviewers can look back to EP for references. Otherwise, this is very confusing.

pkg/controller/registry/resolver/steps.go
```diff
@@ -132,40 +137,50 @@ func NewStepResourceFromBundle(bundle *api.Bundle, namespace, replaces, catalogS
...
 	}
 
 	operatorServiceAccountSteps, err := NewServiceAccountStepResources(csv, catalogSourceName, catalogSourceNamespace)
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 	steps = append(steps, operatorServiceAccountSteps...)
-	return steps, nil
+	saNamespaces := []string{}
```
dinhxuanvu, line 172: I'm a bit unclear for why we need the namespace here. The catalogsource namespace is used because that where the bundle is unpacked but using it for identifier is really a safe choice? What happens if multiple operators have the same resources and all in the same catalogsource?

dinhxuanvu (commented):
Reviewed and added comments.

pkg/controller/registry/resolver/steps.go
dinhxuanvu: Looking back at a previous comment, it seems the domain-based identifier is a change that you intended to do. I think an update PR for the EP is in order.

pkg/controller/operators/catalog/operator.go
```diff
@@ -2312,7 +2319,26 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
 					// Ensure Unstructured Object
 					status, err := ensurer.EnsureUnstructuredObject(resourceInterface, unstructuredObject)
 					if err != nil {
-						return err
+						if step.Optional {
+							switch k8serrors.ReasonForError(err) {
+							case
+								metav1.StatusReasonUnauthorized,
```
dinhxuanvu, line 2325: Do we handle the case of `StatusReasonAlreadyExists` somewhere? Or do we need to?

pkg/controller/registry/resolver/steps.go
```diff
@@ -230,3 +284,39 @@ func NewServiceAccountStepResources(csv *v1alpha1.ClusterServiceVersion, catalog
 	}
 	return rbacSteps, nil
 }
+
+type optionalManifests struct {
+	Files []string `json:"manifests"`
```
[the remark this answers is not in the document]
fgiloux: As per previous review and comment below: this is implemented the change that was requested

pkg/controller/registry/resolver/steps.go
fgiloux: Hi @dinhxuanvu yes this is a change. It was requested by @joelanford during review. I will create an update PR for the EP but wanted to avoid too many round trips. We have already had a fair amount here.

pkg/controller/registry/resolver/steps.go
fgiloux: This is a bit of an edge case and I am open to suggestions. The original intent of the enhancement is to allow operator installation when APIs of "optional resources" are not available.
ServiceAccounts are core resources and as such their API will always be available.
Wanting to make things consistent I kept what was populated into namespace for them. Another option would be to keep the namespace empty. Do you think something else would make more sense?

pkg/controller/operators/catalog/operator.go
fgiloux: True, this deviates from the EP. There should not be an existing resource (I don't see a reason for it) with same namespace/name when the operator gets installed. Failing the InstallPlan would give the opportunity to hte cluster admin to remove it and retry the installation. This would guarantee that the available resource is the intended one. In that case I would update the EP and align it with the implementation. Let me know if you think that it should be done differently.

fgiloux:
pushed a rebase and field and var names in unit tests that were not reflecting the new approach

fgiloux:
Created a [PR](https://github.com/operator-framework/enhancements/pull/111) to update the enhancement proposal with the new approach

perdasilva:
closing as we we'll likely want to put this effort into v1 instead

PR #2562, opened 2022-01-10
Do not modify object from the lister cache 

awgreene:
Problem: Objects retrieved from the lister should
not be modified.

Solution: Return a DeepCopy of the object retrieved
from the lister which can then be manipulated as
needed.

### DIFF
--- a/pkg/controller/operators/olm/operator.go
+++ b/pkg/controller/operators/olm/operator.go
@@ -842,6 +842,7 @@ func (a *Operator) syncObject(obj interface{}) (syncError error) {
 	if related {
 		csvList := a.csvSet(metaObj.GetNamespace(), v1alpha1.CSVPhaseFailed)
 		for _, csv := range csvList {
+			csv = csv.DeepCopy()
 			if csv.Status.Reason != v1alpha1.CSVReasonComponentFailedNoRetry {
 				continue
 			}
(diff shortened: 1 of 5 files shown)


awgreene:
I am going to spend a bit of time looking for instances where objects returned from the lister are mutated.
/hold

pkg/controller/operators/olm/operator.go
```diff
@@ -1627,7 +1627,7 @@ func (a *Operator) operatorGroupForCSV(csv *v1alpha1.ClusterServiceVersion, logg
 			return nil, nil
 		}
 		logger.Debug("csv in operatorgroup")
-		return operatorGroup, nil
```
timflannagan, line 1630: Any idea on whether we should be handling this kind of operation here vs. any call sites to this function? If I'm reading the code right, we get a list of OperatorGroup resources in the CSV's namespace, and all of those OperatorGroup resources should be treated as read-only resources, so it might be reasonable to return a copy of that resource if the blast radius is high.

tylerslaton:
Approach looks good to me, just need to make sure that we update the branch with the latest changes.

tylerslaton:
Nice work!

/lgtm

timflannagan:
Holding so the bot doesn't go crazy.

/hold

Issue #2573, opened 2022-01-19
Bumping operator-framework API dependency to  v0.11.1 breaks OLM


fgiloux:
## Bug Report

**What did you do?**
Bumped OLM dependency to v0.11.1 (the current latest)

**What did you expect to see?**
OLM needs to work with the newer API version

**What did you see instead? Under which circumstances?**
Build failed due to the newer version of k8s.io/api to v0.23.0:

~~~
Error: pkg/controller/operators/suite_test.go:91:3: cannot use []"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1".CustomResourceDefinition{...} (type []"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1".CustomResourceDefinition) as type []*"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1".CustomResourceDefinition in field value
# github.com/operator-framework/operator-lifecycle-manager/pkg/controller/registry/reconciler
Error: pkg/controller/registry/reconciler/reconciler.go:135:7: unknown field 'Handler' in struct literal of type "k8s.io/api/core/v1".Probe
Error: pkg/controller/registry/reconciler/reconciler.go:144:7: unknown field 'Handler' in struct literal of type "k8s.io/api/core/v1".Probe
[...]
~~~

**Environment**
* operator-lifecycle-manager version: master branch

**Additional context**
I will have a closer look at it.
Dependencies were bumped in o-f API in this PR: https://github.com/operator-framework/api/pull/212

fgiloux:
Another side effect of the bump is that logr version was updated from 0.4.0 to 1.2.0. This is a major change with braking changes. Especially logr.Logger is an interface in v0.4.0, renamed to logr.LogSink, and a new type logr.Logger is defined as a struct.

~~~
pkg/controller/operators/openshift/options.go:59:11: invalid operation: c.Log == nil (mismatched types logr.Logger and nil)
~~~

Issue #2353, opened 2021-09-10
Remove controller-runtime replace pin


timflannagan:
## Feature Request

**Is your feature request related to a problem? Please describe.**
A clear and concise description of what the problem is. Example: "I have an issue when (...)"

Remove the [replace pin](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/go.mod#L66) for controller runtime once the race condition issue around metadata-only watchers has been resolved.

**Describe the solution you'd like**
A clear and concise description of what you want to happen. Add any considered drawbacks.

Bump the controller runtime dependency and remove the replace pin once https://github.com/kubernetes-sigs/controller-runtime/pull/1650 gets tagged in a release.

timflannagan:
It looks like c-r v0.10.1 contains that linked PR - @benluddy any idea on whether we can remove the replace pin in the root go.mod now that Joe's PR has been tagged?

benluddy:
> It looks like c-r v0.10.1 contains that linked PR - @benluddy any idea on whether we can remove the replace pin in the root go.mod now that Joe's PR has been tagged?

I don't see why not. The partial metadata race bug was the only reason for the temporary fork.

exdx:
@timflannagan this seems to be resolved in the latest OLM version? There is no longer a pinned version in the go.mod. Could we close this issue out?

PR #2574, opened 2022-01-19
Bump k8s dependencies to 1.23 to align with o-f/api

fgiloux:

Bump k8s dependencies to 1.23 to align with o-f/api
Also as a side effect: newer logr version: v0.4 => v1.2

Alignment with o-f/api, alignment with newer k8 versions.

**Reviewer Checklist**

Closes: #2573


fgiloux:
@timflannagan as discussed the PR for the dependency bump. I am not sure why the build fails. 
It seems to pick a different commit: 
~~~
go build -mod=vendor -ldflags "-X github.com/operator-framework/operator-lifecycle-manager/pkg/version.GitCommit=495a1d5513067fce1fb60d456dd3e9cfc718df4a -X github.com/operator-framework/operator-lifecycle-manager/pkg/version.OLMVersion=`cat OLM_VERSION`" -tags "json1" -o bin/catalog github.com/operator-framework/operator-lifecycle-manager/cmd/catalog
~~~

Building locally is successful, `make build` then uses `GitCommit=121c31f17f4ea121a7c1a7a3fda58bd03446aedf`

fgiloux:
> This is most likely the culprit (go 1.17 should be used):
> 
> ```
> golang                   x86_64   1.16.13-1.fc34               updates   608 k
> ```

I have amended the Dockerfile so that go 1.17 gets installed (not available in Fedora 34)

fgiloux:
I had to amend the github action for unit tests:
- the envtest binaries were very very old (K8 1.16 IIRC)
- the distribution of kubebuilder binaries and envtest changed in the meantime
- kubebuilder is not really needed for the tests and not part of the envtest installation so that I disabled its check in the Makefile

fgiloux:
@timflannagan 

Next issue:

~~~
The CustomResourceDefinition "clusterserviceversions.operators.coreos.com" is invalid: metadata.annotations: Too long: must have at most 262144 bytes
~~~

It seems that the main change is the addition of a grpc field, which increases the size of the CSV CRD. I am missing some background. I will have a look but any hint is welcome

dinhxuanvu:
> @timflannagan
> 
> Next issue:
> 
> ```
> The CustomResourceDefinition "clusterserviceversions.operators.coreos.com" is invalid: metadata.annotations: Too long: must have at most 262144 bytes
> ```
> 
> It seems that the main change is the addition of a grpc field, which increases the size of the CSV CRD. I am missing some background. I will have a look but any hint is welcome

I dont think this is a grpc problem. This seems to be a part of the install script where we use `kubectl apply` which will save the entire CRD yaml into annotation but that should only happen if you apply to an existing CRD already. At the startup time, that CRD shouldn't exist. Not sure why.

fgiloux:
>  we use kubectl apply which will save the entire CRD yaml into annotation but that should only happen if you apply to already-existing CRD.

If you don't want the annotation `kubectl create`, `kubectl replace` should be used. It may not have happen earlier due to the old version in use. I will see whether I can amend the script

timflannagan:
> kubebuilder is not really needed for the tests and not part of the envtest installation so that I disabled its check in the Makefile

Completely agreed that we should be ripping out kubebuilder, and there might be an open issue for this (unless I'm tracking it only on my local TODO list) and replace it with envtest, as that's the only thing we require for this repository.

@fgiloux I'm still largely tied up with non-OLM work, so I likely won't have much capacity to help push this along.

fgiloux:
@timflannagan @dinhxuanvu there were only 2 end to end tests failing. I ran them locally twice and they passed each time. 
It either be flaky in the CI environment or I may need more hint. What I could see was mainly a timeout

timflannagan:
@fgiloux OLM's e2e suite is notoriously flaky so that's the most likely culprit here, and we've been burning down a good chunk of those flakes but they still happen fairly frequently. https://github.com/operator-framework/operator-lifecycle-manager/issues?q=is%3Aissue+is%3Aopen+label%3Akind%2Ffailing-test is a list of some of the more frequent ones, but most of the recent issues that have been created are tracking flakes that are popping up in the e2e suite.

fgiloux:
> @fgiloux OLM's e2e suite is notoriously flaky so that's the most likely culprit here, and we've been burning down a good chunk of those flakes but they still happen fairly frequently.

the two tests that failed were:
"config map update triggers registry pod rollout"
"should create a Subscription for the latest entry providing the required GVK"

It may be worth a /retest

fgiloux:
@timflannagan rebased and all checks have passed. A sanity check would be welcome.

timflannagan:
@fgiloux Nice - I'll have some time in like an hour to look over the changes and review.

go.mod
```diff
@@ -223,23 +223,26 @@ require (
...
+	sigs.k8s.io/structured-merge-diff/v4 v4.2.0 // indirect
+	sigs.k8s.io/yaml v1.3.0 // indirect
 )
 
 replace (
 	// controller runtime
 	github.com/openshift/api => github.com/openshift/api v0.0.0-20211014063134-be2a7fb8aa44
 	github.com/openshift/client-go => github.com/openshift/client-go v0.0.0-20200326155132-2a6cd50aedd0 // release-4.5
 
+	github.com/operator-framework/api v0.11.1 => github.com/operator-framework/api v0.11.2-0.20220118211235-330a17e17dca
```
timflannagan, line 244: I think my preference would be to avoid coupling the 1.23 kube dependency bump with an o-f/api bump as well (e.g. those changes are handled in it's own PR, or as an additional commit in your optional manifest work). It would help reduce the scope of this PR, and cut down on the overall diff. Any thoughts? In any case, I think we still need to properly cut an o-f/api tag for those CRD changes you had merged.

Dockerfile
```diff
@@ -5,7 +5,9 @@ WORKDIR /build
 # install dependencies and go 1.16
 
 # copy just enough of the git repo to parse HEAD, used to record version in OLM binaries
-RUN dnf update -y && dnf install -y bash make git mercurial jq wget golang && dnf upgrade -y
+RUN dnf update -y && dnf install -y bash make git mercurial jq wget && dnf upgrade -y
+RUN curl -sSL https://go.dev/dl/go1.17.6.linux-amd64.tar.gz | tar -xzf - -C /usr/local
+ENV PATH=/usr/local/go/bin:$PATH
```
timflannagan, line 10: Hmm this feels fairly hacky - it might be time to reword the build image that we use for this and work towards using a proper Go image. It's unclear to me looking at this stage why we need a fedora base image, as we're just building the OLM binaries and any utility packages needed in the pipeline.

Makefile
```diff
@@ -49,9 +49,10 @@ unit: kubebuilder
 # Ensure kubebuilder is installed before continuing
 KUBEBUILDER_ASSETS_ERR := not detected in $(KUBEBUILDER_ASSETS), to override the assets path set the KUBEBUILDER_ASSETS environment variable, for install instructions see https://book.kubebuilder.io/quick-start.html
 kubebuilder:
-ifeq (, $(wildcard $(KUBEBUILDER_ASSETS)/kubebuilder))
-	$(error kubebuilder $(KUBEBUILDER_ASSETS_ERR))
-endif
+# not needed for envtest
+# ifeq (, $(wildcard $(KUBEBUILDER_ASSETS)/kubebuilder))
+# 	$(error kubebuilder $(KUBEBUILDER_ASSETS_ERR))
+# endif
```
timflannagan, line 55: Do we need to still provide a Makefile target that installs envtest? That target could then be a pre-requisite for the `unit` target. Thoughts?

cmd/package-server/main.go
```diff
@@ -19,6 +20,11 @@ func main() {
 	options := server.NewPackageServerOptions(os.Stdout, os.Stderr)
 	cmd := server.NewCommandStartPackageServer(ctx, options)
 	cmd.Flags().AddGoFlagSet(flag.CommandLine)
+	// klog flags used by dependencies need to be explicitly initialized now
+	klogFlags := flag.NewFlagSet("klog", flag.ExitOnError)
+	klog.InitFlags(klogFlags)
+	cmd.Flags().AddGoFlagSet(klogFlags)
```
timflannagan, line 26: You've been closer to the ground than I have on this - any idea on where I would be able to learn more about this further, or could provide some more color on the comment you added here? It's possible I'm missing something obvious and I need some more caffeine in me.

go.mod
```diff
@@ -30,29 +30,29 @@ require (
...
+	k8s.io/component-base v0.23.1
 	k8s.io/klog v1.0.0
 	k8s.io/kube-aggregator v0.22.0
-	k8s.io/kube-openapi v0.0.0-20210527164424-3c818078ee3d
-	k8s.io/utils v0.0.0-20210819203725-bdf08cb9a70a
-	sigs.k8s.io/controller-runtime v0.10.0
-	sigs.k8s.io/controller-tools v0.6.2
+	k8s.io/kube-openapi v0.0.0-20211115234752-e816edb12b65
+	k8s.io/utils v0.0.0-20210930125809-cb0fa318a74b
+	sigs.k8s.io/controller-runtime v0.11.0
```
timflannagan, line 54: FYI - we [replace pin](https://github.com/operator-framework/operator-lifecycle-manager/blob/e2f2c5ba3b37898427a35e2e3fb9000d4781b4ee/go.mod#L248) the controller-runtime dependency to my fork due to how partial metadata objects are handled upstream right now.

go.mod
fgiloux: I am not sure that what you are proposing is feasible. The kube 1.23 bump was driven by  the need to bump o-f/api. The latest o-f/api version comes with kube 1.23 dependencies and a newer logging stack.

Dockerfile
fgiloux: I would not disagree that it would make sense to have a better for purpose base image. As I was focusing on the version upgrade I kept the existing approach: download and unpack

Makefile
fgiloux: yes I would vote for it. This would be for a separate PR.

go.mod
timflannagan: Yeah I should've been clearer: I think we can bump the kube 1.21 dependencies + the logr dependency manually, without having to bump o-f/api at the same time. I'm still working through the set of required changes, and viewing the overall diff is fairly disruptive, so maybe it's easier to simply chunk out some of the overall set of changes into their own commits (without having to create another PR), or we can simply bite the bullet for now if this is urgent enough.

cmd/package-server/main.go
fgiloux: See: https://pkg.go.dev/k8s.io/klog#readme-how-to-use-klog
"Use klog.InitFlags(nil) explicitly for initializing global flags as we no longer use init() method to register the flags"

Dockerfile
timflannagan: Yeah, totally fair, that was more of a PSA than me requesting changes. Maybe we should be tracking this as an upstream issue then?

Just a quick note to other reviewers, it looks like these changes are needed as [builds were failing](https://github.com/operator-framework/operator-lifecycle-manager/pull/2574#issuecomment-1016368075) as this Dockerfile uses a fedora image that doesn't contain an up-to-date Go minor version.

go.mod
fgiloux: ok. Have you got an issue, something I could read to better understand the limitations?

pkg/controller/operators/suite_test.go
```diff
@@ -88,14 +88,14 @@ var _ = BeforeSuite(func() {
...
 		UseExistingCluster: &useExisting,
-		CRDs: []apiextensionsv1.CustomResourceDefinition{
-			*crds.CatalogSource(),
-			*crds.ClusterServiceVersion(),
-			*crds.InstallPlan(),
-			*crds.Subscription(),
-			*crds.OperatorGroup(),
-			*crds.Operator(),
-			*crds.OperatorCondition(),
+		CRDs: []*apiextensionsv1.CustomResourceDefinition{
```
timflannagan, line 91: Interesting - it looks like this was changed in https://github.com/kubernetes-sigs/controller-runtime/commit/a14a68cdaba04ccbc2dd4501f162f8472f11d800 but if I'm reading that commit tagging right, it's only available in v0.11.0 versions of c-r but we still replace pin to my fork's v0.10.x version. Any idea on why we need to make this change? In order to make CI happy as we're pulling in the latest envtest binary now, so there's some potential skew between our c-r version and entest?

pkg/controller/operators/openshift/options.go
```diff
@@ -57,7 +57,7 @@ func (c *ReconcilerConfig) complete() error {
 		return fmt.Errorf("no ClusterOperator name specified")
 	}
 	if c.Log == nil {
-		c.Log = ctrl.Log.WithName(c.Name)
+		c.Log = ctrl.Log.WithName(c.Name).GetSink()
```
timflannagan, line 60: Just for my own edification: this change is needed since the logr.Logger type has changed, and we can no longer nil check it's value, or was there some other breaking change in the logr major release bump?

go.mod
timflannagan: https://github.com/operator-framework/operator-lifecycle-manager/issues/2353 is the tracking issue but it's lacks any real context.

pkg/controller/operators/openshift/options.go
fgiloux: "The breaking change in logr seems to be that logr.Logger is an interface in v0.4.0, renamed to logr.LogSink, and a new type logr.Logger is defined as a struct."
https://github.com/kubernetes-sigs/controller-runtime/issues/1607
This lead to " invalid operation: logging.logr == nil (mismatched types logr.Logger and nil"

pkg/controller/operators/suite_test.go
fgiloux: I may have missed something:
~~~
require (
[...]
sigs.k8s.io/controller-runtime v0.11.0
[...]
replace (
sigs.k8s.io/controller-runtime v0.10.0 => github.com/timflannagan/controller-runtime v0.10.1-0.20211210161403-6756a4203e70
~~~
The replace version v0.10.0 does not match the request version v0.11.0. This means that v0.11.0 was in use. With your version of controller-runtime there are incompatibilities with the logr version brought by o-f/api, for instance:
~~~
vendor/sigs.k8s.io/controller-runtime/pkg/log/deleg.go:79:12: undefined: logr.WithCallDepth
vendor/sigs.k8s.io/controller-runtime/pkg/log/deleg.go:163:2: cannot use res (type *DelegatingLogger) as type logr.Logger in return argument
~~~

I am looking at: #2353

fgiloux: I looked at #2368 but did not see why you decided to keep being on a fork. [The issue](https://github.com/kubernetes-sigs/controller-runtime/pull/1650) you mentioned in regards to the way metadata is handled in controller-runtime has been merged with commit 74ba5d77bf99d8968543f3536e4d49a2e3818fee is in [v0.11](https://github.com/kubernetes-sigs/controller-runtime/pull/1650/commits/74ba5d77bf99d8968543f3536e4d49a2e3818fee)
Am I missing something?

timflannagan: I'm somewhat blanking on the overall context as to why that fork was necessary when bumping the c-r dependency in OLM to v0.10.x at the time. I remember seeing e2e was constantly failing despite the changes being fairly innocuous, and we were under the impression that Joe's upstream fix solved the issue we originally had to hack around. Let me see if I can dig up any more context today, but given the current e2e checks are passing, it's possible we had misdiagnosed the root case at the time.

scripts/install_local.sh
```diff
@@ -22,7 +22,9 @@ do
     then
         kubectl replace --force -f "${f}"
     else
-        kubectl apply -f "${f}"
+	if ! kubectl create -f "${f}"; then
+		kubectl replace -f "${f}"
+	fi
```
timflannagan, line 27: Any value in adding a comment to explain why this needed?

pkg/controller/operators/openshift/options.go
```diff
@@ -28,7 +28,7 @@ type NowFunc func() metav1.Time
 type ReconcilerConfig struct {
 	Client       client.Client
 	Scheme       *runtime.Scheme
-	Log          logr.Logger
+	Log          logr.LogSink
```
timflannagan, line 31: I think if I'm understanding the changes correctly, we can avoid changing the type declaration for the Log field here, keeping the logr.Logger, and instead simply update the conditional check on line 59 to `if c.Log.GetSink() == nil {`, and the overall diff for this file would be a single LOC changed.

pkg/controller/operators/openshift/options.go
timflannagan: I think we can minimize the changes to this file but only updating the conditional check on line 59 without mucking with any updated type declarations. See https://github.com/operator-framework/operator-lifecycle-manager/pull/2574/files#r789780717 for what I had in mind.

Makefile
timflannagan: Yeah, that's sounds reasonable to me as long as we commit to tackling that in a follow-up PR.

timflannagan:
@fgiloux I think we're getting closer to the finish line with these changes. Apologies for the slow turn around over the past week, but it seems like the github UI still doesn't handle large changesets well, so overall review time is substantially increased due to pages becoming unresponsive.

I think we need to do the following to push this along:

- Any PR this size will need another reviewer to review these changes
- We need to be certain that we're not regressing with the controller-runtime bump w.r.t to metadata-only watchers
- Iron out the envtest installation in CI

Is there anything that I'm missing?

scripts/install_local.sh
fgiloux: sure. I can add a comment

pkg/controller/operators/openshift/options.go
fgiloux: I have done the change. It compiled and unit tests were fine locally. Let see how it goes with the rest of the checks.

fgiloux:
> Is there anything that I'm missing?

Besides Kube and go versions and there is a fair amount of tests for controlling them the biggest change is in regards to the logging stack. It may be good to make a sanity check to ensure that logs have not degraded.

fgiloux:
@timflannagan @perdasilva I have rebased the PR. It would be good to get it merged now.

.github/workflows/unit.yml
```diff
@@ -17,7 +17,9 @@ jobs:
         run: |
           os=$(go env GOOS)
           arch=$(go env GOARCH)
-          curl -L https://github.com/kubernetes-sigs/kubebuilder/releases/download/v2.3.1/kubebuilder_2.3.1_${os}_${arch}.tar.gz | tar -xz -C /tmp/
-          sudo mv /tmp/kubebuilder_2.3.1_${os}_${arch} /usr/local/kubebuilder
+          go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
+          setup-envtest use --use-env -p path 1.19.x!
+          sudo mkdir -p /usr/local/kubebuilder/bin
+          sudo cp /home/runner/.local/share/kubebuilder-envtest/k8s/1.19.2-linux-amd64/* /usr/local/kubebuilder/bin/
```
timflannagan, line 23: I just have a couple quick comments:

- It looks like we can remove the os/arch variable declarations on lines 18/19
- We can use the `env` field to provide the k8s path version
- We can explicitly configure the KUBEBUILDER_ASSETS variable to the `make unit` check on line 25

```yaml
# global envvar
name: unit
env:
  K8S_VERSION: 1.22.x
...
- name: Install Kubebuilder
  run: |
    go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
    setup-envtest use ${K8S_VERSION}
- name: Run unit tests
  run: KUBEBUILDER_ASSETS=$(setup-envtest use -i -p path ${K8S_VERSION}) make unit
```

You might need to also update the root Makefile and replace the logic that checks for the presence of the [`kubebuilder` binary](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/Makefile#L52) and instead check for the `kubectl` binary.

cmd/package-server/main.go
timflannagan: Interesting - looking at the 1.23 release notes, it looks like the majority of the klog-related flags were deprecated outside of the `-v` flag. Did you need to add this as compilations were failing, or because the packageserver was crashing at runtime because the new changes resulted in mis-configuration?

timflannagan:
I think this is largely read to go but I haven't verified the controller-runtime changes yet. If we're bumping the c-r version, we should also remove the replace pin from the root go.mod.

.github/workflows/unit.yml
timflannagan: In general, it looks like we can cleanup that KUBEBUILDER_ASSETS check(s) in the root Makefile, and remove the kubebuilder binary check in favor of checking the kubectl binary now.

fgiloux: @timflannagan Thanks for the recommendations they are good improvements. The only thing is that I kept 1.19.2 for envtest as there are known issues by tear-down with newer versions.
I also did the Makefile change you proposed.

fgiloux:
>  If we're bumping the c-r version, we should also remove the replace pin from the root go.mod.
I have done that as well. Let see how the tests go.

.github/workflows/unit.yml
timflannagan: Alright I think that's fine for now, although I'm not sure how much we care about proper tear-down as we aren't hosting our own Github CI runners. It's probably best to stick with this version for now, and we can re-evaluate later to avoid any skew between our CI clusters and the envtest-setup.

cmd/package-server/main.go
timflannagan: Just pulled these changes down locally and deployed w/o these klog changes and it results in runtime errors over time (vs. static compilation failures).

timflannagan:
IIRC, we ran into these two specific test case failures last time we tried to bump the c-r version (hence why the replace pin to my fork was needed): https://github.com/operator-framework/operator-lifecycle-manager/runs/4942485265?check_suite_focus=true

fgiloux:
@timflannagan I have been investigating locally. "OperatorCondition Upgradeable type and overrides " passed but "Operator API when a subscription to a package exists [It] should automatically adopt components" consistently failed.
As far as I have seen the reference to the service account is missing in the operator and prevents:

~~~
Eventually(func() (*operatorsv1.Operator, error) {
                                o := &operatorsv1.Operator{}
                                err := client.Get(clientCtx, operatorName, o)
                                return o, err
                        }).Should(ReferenceComponents([]*corev1.ObjectReference{
                                getReference(scheme, sub),
                                getReference(scheme, ip),
                                getReference(scheme, testobj.WithNamespacedName(
                                        &types.NamespacedName{Namespace: sub.GetNamespace(), Name: "kiali-operator.v1.4.2"},
                                        &operatorsv1alpha1.ClusterServiceVersion{},
                                )),
                                getReference(scheme, testobj.WithNamespacedName(
                                        &types.NamespacedName{Namespace: sub.GetNamespace(), Name: "kiali-operator"},
                                        &corev1.ServiceAccount{},
                                )),
                                getReference(scheme, testobj.WithName("kialis.kiali.io", &apiextensionsv1.CustomResourceDefinition{})),
                                getReference(scheme, testobj.WithName("monitoringdashboards.monitoring.kiali.io", &apiextensionsv1.CustomResourceDefinition{})),
                        }))

~~~

Looking at the logs after failure I can see:

~~~
E0126 13:54:10.097690       1 queueinformer_operator.go:290] sync {"update" "ns-wh857/cs-62lnj"} failed: couldn't ensure registry server - error ensuring service account: cs-62lnj: serviceaccounts "cs-62lnj" is forbidden: unable to create new content in namespace ns-wh857 because it is being terminated
~~~

fgiloux:
@timflannagan I spent more time debugging the e2e test mentioned in my previous comment. I got it to be successful locally but there are quite a few things that don't look right and sorry if I am missing something. Here are the points that are not what I would expect:
- The resources get added to the references in the operator only if they are created as part of the installation and have the right label, something like: operators.coreos.com/kiali.ns-67rrb: "". That means if the SA or the CRDs were created beforehand the test will fail.
- The AfterEach of the e2e test does not fully clean the resources created. It only deletes the namespace. Note that the operator resource created here is used in the next context. This means that CRDs, operator are left and they prevent due to the first point any second run to be successful in the same cluster. As per the doc I run the specific e2e test with:
~~~
GO111MODULE=on GOFLAGS="-mod=vendor" go run github.com/onsi/ginkgo/ginkgo -focus "should automatically adopt components" -v --progress ./test/e2e -- -namespace=operators -olmNamespace=olm -dummyImage=bitnami/nginx:latest
~~~
- When an operator resource gets deleted, the reconciliation does not rebuild it based on the cluster state but based on an old version of the operator resource stored in memory. I had to restart the olm pods to get rid of an operator resource after I had deleted all the resources listed in the references part of the operator. Otherwise it would get created referencing again resources that did not exist anymore.

The e2e test could be tweaked to pass by ensuring a virgin environment before it runs but it seems to highlight something that it is genuinely suspicious. I doubt  that this is directly due to the dependency bump. It may only exacerbate an existing behavior.

timflannagan:
@fgiloux Nice, I've seen similar issues locally when debugging these tests further. I had opened https://github.com/operator-framework/operator-lifecycle-manager/pull/2518 which aimed to fixed that AfterEach issue but I haven't had the time to revisit it recently. That Operator controller has some known caching issues due to the way we track the last resourceVersion that we processed, so it's possible that you can delete all the requisite underlying resources (e.g. SA, CRDs, etc.) and the Operator resource will still be created. I tried to fix that behavior (without fixing the root case) by firing off a DeleteCollection call at the top-level e2e spec.

If I remember correctly with the last c-r bump, we were running into constant e2e test failures in the adoption and operator controllers, where both of those controllers weren't able to adopt resources (note: it was the ServiceAccount resource nearly 100% of the time) and propagate them to the operator resource's status sub-resource. That ServiceAccount resource was registered as a metadata-only watch when instantiating both controllers, and I was under the impression we might've been missing cache events at the informer level, and the new upstream fix in the v0.11.x release didn't work for us.

We don't have to fix the test setup issues in this PR, but I'd like to avoid a situation where OLM's existing suite is already extremely flaky, and becomes more flaky due to a dependency bump.

fgiloux:
@timflannagan I forgot one point:
- The kiali CSV is full of RBAC policies referencing resources that don't exist on a KinD cluster, an extract:
~~~
I0127 05:54:06.485179       1 rbac.go:118] RBAC DENY: user "system:serviceaccount:ns-4vxkx:kiali-operator" groups [] cannot "create" resource "fluentds.config.istio.io" cluster-wide
I0127 05:54:06.485232       1 rbac.go:118] RBAC DENY: user "system:serviceaccount:ns-4vxkx:kiali-operator" groups [] cannot "create" resource "sidecars.networking.istio.io" cluster-wide
I0127 05:54:06.485268       1 rbac.go:118] RBAC DENY: user "system:serviceaccount:ns-4vxkx:kiali-operator" groups [] cannot "watch" resource "meshpolicies.authentication.istio.io" cluster-wide
~~~

There are many points in  #2518. Too bad that it has not been merged yet. 

For my own sake what is the point of tracking the last resourceVersion and having different logic based on it? It seems not aligned with general practices. What am I missing?

> That ServiceAccount resource was registered as a metadata-only watch when instantiating both controllers, and I was under the impression we might've been missing cache events at the informer level,

It could be but when I checked the SA did not have the label. If the controller logic for the operator was working it should add the reference to the operator when the label is added to the SA but it did not in my test.

awgreene:
The following test is failing locally for me, looking into it now.
>[Fail] Catalog represents a store of bundles which OLM can use to install Operators [It] registry polls on the correct interval

awgreene:
Closed in favor of https://github.com/operator-framework/operator-lifecycle-manager/pull/2617

PR #2579, opened 2022-01-21
ci: Support configuring the logging level for the splitter package

timflannagan:
Update the split package and add support for configuring the logging
level to avoid CI interpretting info-level logging as error messages.

Add support for configuring the logging level for the e2e splitter package, and default to the error level.

This is to workaround CI interpreting the previous log.Print* calls as error-level messages. A concrete example can be found in this action run: https://github.com/operator-framework/operator-lifecycle-manager/actions/runs/1724759061 where "found no top level describes, skipping" has a red market above it.

**Reviewer Checklist**

### DIFF
--- a/test/e2e/split/main.go
+++ b/test/e2e/split/main.go
@@ -12,15 +12,16 @@ import (
 	"regexp"
 	"sort"
 	"strings"
-)
 
-// TODO: configurable log verbosity.
+	"github.com/sirupsen/logrus"
+)
 
 type options struct {
 	numChunks  int
 	printChunk int
 	printDebug bool
 	writer     io.Writer
+	logLevel   string
 }
 
 func main() {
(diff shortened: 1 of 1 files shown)


test/e2e/split/main.go
```diff
@@ -103,14 +112,14 @@ func findDescribes(dir string) ([]string, error) {
 		}
 		specNames := topDescribeRE.FindAllSubmatch(b, -1)
 		if len(specNames) == 0 {
-			log.Printf("%s: found no top level describes, skipping", match)
+			logger.Warnf("%s: found no top level describes, skipping", match)
```
timflannagan, line 115: Another implementation: we could also just remove logging from this package entirely. It's nice for debugging this package while developing locally, but provides little value when running in CI environments.

dinhxuanvu (commented):
/lgtm

test/e2e/split/main.go
perdasilva: If it's not getting in the way, I'd say leave it. Logging is helpful during development. Though a debugger is also nice...

PR #2580, opened 2022-01-21
Remove grace period from csv deletions

perdasilva:

This PR removes the 0 grace period from the CSV deletion calls

This causes all csv owned objects to also be deleted with a 0 grace period. This is impacting a customer that relies on grace period to clean up their pods.

Seems like there are [other](https://github.com/operator-framework/operator-lifecycle-manager/pull/2047) good reasons as well:

```
Force deleting a pod is not allowed for automated processes within Kube or OpenShift unless done so by a human. That action is reserved because it is effectively bypassing the safety mechanisms of a cluster which ensure that only one pod is on any one node at a time, and leaves the state of the system inconsistency between apiserver and node (the node may still run that old process indefinitely).
```

**Reviewer Checklist**


dinhxuanvu:
@perdasilva I removed the BZ reference on the title. You only include the BZ info on downstream PR.

perdasilva:
@dinhxuanvu whoopsie - thank you, Vu!

PR #2606, opened 2022-02-01
Fix a bug in deletion of webhook service for replacement

orenc1:
The condition that checks for an error with the webhook service deletion is based on the wrong variable, and the return statement is never reached, which leads to a continuous error when OLM tries to replace the webhook service on certain conditions

Signed-off-by: orenc1 <ocohen@redhat.com>

https://bugzilla.redhat.com/show_bug.cgi?id=2048441

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/install/certresources.go
+++ b/pkg/controller/install/certresources.go
@@ -246,7 +246,7 @@ func (i *StrategyDeploymentInstaller) installCertRequirementsForDeployment(deplo
 
 		// Delete the Service to replace
 		deleteErr := i.strategyClient.GetOpClient().DeleteService(service.GetNamespace(), service.GetName(), &metav1.DeleteOptions{})
-		if err != nil && !k8serrors.IsNotFound(deleteErr) {
+		if deleteErr != nil && !k8serrors.IsNotFound(deleteErr) {
 			return nil, nil, fmt.Errorf("could not delete existing service %s", service.GetName())
 		}
 	}


timflannagan:
(Removing the BZ from the title as it messes up downstream automation)

/ok-to-test

orenc1:
looks like the e2e failures were transient.
/retest

dinhxuanvu (approved):
/lgtm

awgreene (approved):
Thanks for submitting this fix @orenc1

orenc1:
hi @awgreene do you think we are ready to rerun the CI?
thanks

pkg/controller/install/certresources.go
```diff
@@ -246,7 +246,7 @@ func (i *StrategyDeploymentInstaller) installCertRequirementsForDeployment(deplo
 
 		// Delete the Service to replace
 		deleteErr := i.strategyClient.GetOpClient().DeleteService(service.GetNamespace(), service.GetName(), &metav1.DeleteOptions{})
-		if err != nil && !k8serrors.IsNotFound(deleteErr) {
```
perdasilva, line 249: This looks like something that should have been picked up by unit testing. Would it be a lot of effort to add a test?

perdasilva: Since e2e is green, let's merge and put the unit test in a follow up PR. Would that work for you?

orenc1: Yes, although I'm not sure how can we check this particular case in a unit test, in which the service could not be deleted and therefore the `installCertRequirementsForDeployment` method returns an error.
I assume we'll need to use:
```
mockOpClient.EXPECT().DeleteService(namespace, "test-service", &metav1.DeleteOptions{}).Return(fmt.Errorf("could not delete existing service test-service"))
```
But not quite sure how to proceed.

timflannagan (approved):

PR #2607, opened 2022-02-01
fix: remove hardcoded minikube version from build_local.sh 

exdx:
This change supports OLM minikube users locally since the currently hardcoded kubernetes version v1.16.4 is severely outdated and doesn't work on recent versions of minikube. The script will instead use the version of kubernetes associated with the existing version of minikube. 

OLM has largely moved to kind for CI and local testing, but users running minikube locally should also be able to run OLM. 

**Reviewer Checklist**

### DIFF
--- a/scripts/build_local.sh
+++ b/scripts/build_local.sh
@@ -8,7 +8,7 @@ set -e
 [ -x "$(command -v kind)" ] && [[ "$(kubectl config current-context)" =~ ^kind-? ]] && KIND=1 NO_MINIKUBE=1
 
 if [ -z "$NO_MINIKUBE" ]; then
-  pgrep -f "[m]inikube" >/dev/null || minikube start --kubernetes-version="v1.16.4" --extra-config=apiserver.v=4 || { echo 'Cannot start minikube.'; exit 1; }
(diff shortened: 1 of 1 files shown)


timflannagan (approved):

scripts/build_local.sh
```diff
@@ -8,7 +8,7 @@ set -e
 [ -x "$(command -v kind)" ] && [[ "$(kubectl config current-context)" =~ ^kind-? ]] && KIND=1 NO_MINIKUBE=1
 
 if [ -z "$NO_MINIKUBE" ]; then
-  pgrep -f "[m]inikube" >/dev/null || minikube start --kubernetes-version="v1.16.4" --extra-config=apiserver.v=4 || { echo 'Cannot start minikube.'; exit 1; }
+  pgrep -f "[m]inikube" >/dev/null || minikube start --extra-config=apiserver.v=4 || { echo 'Cannot start minikube.'; exit 1; }
```
timflannagan, line 11: Actually the more I think about this: do we risk skew between CI and local environments by defaulting to the latest k8s version? How would this change affect a repository that's using 1.22 k8s APIs, when the server version is 1.23?

timflannagan: Okay I can buy that. It looks like this needs a rebase again due to branch protection rules, but once rebased against master, I can tag this again.

PR #2609, opened 2022-02-02
Infer/Set Blank APIVersion - Bug 1982737

nsapse:
This commit modifies creation of the CSV passed to ownerReferences by setting it to v1alpha1 (inferred from the fact that it is being used for all other purposes) in the case that it has been left blank in the original bundle.

In previous versions of OLM is a user passed in a bundle missing an APIVersion the install plan would fail with a difficult to decipher message stating it had failed to create a service account for the  `metadata.ownerReferences.apiVersion: Invalid value: "": version must not be empty'. `  The original bug report asked for the following in response:

1.  For OPM index add to fail if an APIVersion had not been set in the bundle.

To enforce separation of concerns the proper place for this to fail is with Operator SDK Validate. This error is it is in fact caught properly by the SDK and fails at that stage as expected.

2.  For the InstallPlan to fail with a message pointing the user to an error in the bundle's CSV.

Rather than update the error message this change fixes the underlying issue preventing the InstallPlan from failing.

**Reviewer Checklist**


benluddy:
What happens if the manifest was intended to represent a v2beta3 CSV? Say there were some new fields added -- wouldn't they have been lost during unmarshaling into the v1beta1 struct? The error is desirable here.

njhale:
/hold

Just had a long conversation with @benluddy about this and I agree that we shouldn't infer the version is v1alpha1.

I originally suggested inference on the __false assumption__ that a bundle could be installed today w/o the CSV version field. As it turns out, most -- I really think all -- bundles that don't specify the CSV version would cause the InstallPlan to fail on the same ownerreference issue. This means backwards compatibility is no longer a problem for solutions that don't support unspecified versions.

That being said, instead of inferring the version, I think we should fail InstallPlans earlier in their execution and output a more helpful message to their status when we notice an unspecified version. In reality, we should be doing this already for resource that use the dynamic (untyped) client (so switching to that for POSTing CSV could be helpful too).

anik120:
>I think we should fail InstallPlans earlier in their execution and output a more helpful message to their status when we notice an unspecified version

@njhale @benluddy isn't that under the assumption that the apiversion is the only field that'll be absent from the CSV? For example, since the apiserver infers the kind too if that's missing, do we need another check for kind? Are there other fields that are optional in the mainfest but will be inferred by the apiserver? Do we need checks for all of those fields?

nsapse:
Per conversation above PR has been closed in favor of working on an alternate solution in PR#2673[](https://github.com/operator-framework/operator-lifecycle-manager/pull/2673)

PR #2616, opened 2022-02-04
change Lister to DynamicClient for subscription clean up in Teardown

### LINKED ISSUE #2615: e2e - subscription is left after tests intermittently
## Bug Report
The instance of subscription is left after the test intermittently. It sometime causes the following test failures because the other resources such as the catalogsource are deleted.

**What did you do?**
I ran the following test with minikube or Rancher desktop.

```
GO111MODULE=on GOFLAGS="-mod=vendor" go run github.com/onsi/ginkgo/ginkgo -focus "config map update triggers registry pod rollout" -v --progress ./test/e2e -- -namespace=operators -olmNamespace=olm -dummyImage=bitnami/nginx:latest
```
No test subscription instance after the test. 

I sometimes saw one subscription instance there after the test. 

**Environment**
Latest from the master branch

```
Client Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.3", GitCommit:"816c97ab8cff8a1c72eccca1026f7820e93e0d25", GitTreeState:"clean", BuildDate:"2022-01-25T21:17:57Z", GoVersion:"go1.17.6", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.3+k3s1", GitCommit:"5fb370e53e0014dc96183b8ecb2c25a61e891e76", GitTreeState:"clean", BuildDate:"2022-01-27T02:12:21Z", GoVersion:"go1.17.5", Compiler:"gc", Platform:"linux/amd64"}

```
* Kubernetes cluster kind: 

**Possible Solution**

**Additional context**

akihikokuroda:

This line https://github.com/operator-framework/operator-lifecycle-manager/blob/ac4de75fd3972ca12bdc12639bd8f79ff75e64fd/test/e2e/util.go#L403 in the Teardown should delete all subscriptions in the namespace but it sometime doesn't (I don't know why so far, I 'll investigate, later).  

This line https://github.com/operator-framework/operator-lifecycle-manager/blob/ac4de75fd3972ca12bdc12639bd8f79ff75e64fd/test/e2e/util.go#L408 is used to wait until the all subscriptions are deleted but it always gets no instance because the `DeleteAllOf` function deletes the subscriptions in the cache and the `LIst` function gets the instance from the cache.

This PR changes the way to get the subscriptions from the API server directly instead of the cache.

Closes #2615 

**Reviewer Checklist**

### DIFF
--- a/test/e2e/util.go
+++ b/test/e2e/util.go
@@ -391,6 +391,7 @@ func TearDown(namespace string) {
 	var (
 		clientCtx   = context.Background()
 		client      = ctx.Ctx().Client()
+		dynamic     = ctx.Ctx().DynamicClient()
 		inNamespace = k8scontrollerclient.InNamespace(namespace)
 		logf        = ctx.Ctx().Logf
 	)
@@ -403,14 +404,13 @@ func TearDown(namespace string) {
 		return client.DeleteAllOf(clientCtx, &operatorsv1alpha1.Subscription{}, inNamespace)
 	}).Should(Succeed(), "failed to delete test subscriptions")
 
-	Eventually(func() (remaining []operatorsv1alpha1.Subscription, err error) {
-		list := &operatorsv1alpha1.SubscriptionList{}
-		err = client.List(clientCtx, list, inNamespace)
-		if list != nil {
-			remaining = list.Items
+	var subscriptiongvr = schema.GroupVersionResource{Group: "operators.coreos.com", Version: "v1alpha1", Resource: "subscriptions"}
+	Eventually(func() ([]unstructured.Unstructured, error) {
+		list, err := dynamic.Resource(subscriptiongvr).Namespace(namespace).List(context.Background(), metav1.ListOptions{})
+		if err != nil {
+			return nil, err
 		}
-
-		return
+		return list.Items, nil
 	}).Should(BeEmpty(), "failed to await deletion of test subscriptions")
 
 	logf("deleting test installplans...")


test/e2e/util.go
```diff
@@ -403,14 +404,13 @@ func TearDown(namespace string) {
 		return client.DeleteAllOf(clientCtx, &operatorsv1alpha1.Subscription{}, inNamespace)
 	}).Should(Succeed(), "failed to delete test subscriptions")
 
-	Eventually(func() (remaining []operatorsv1alpha1.Subscription, err error) {
-		list := &operatorsv1alpha1.SubscriptionList{}
-		err = client.List(clientCtx, list, inNamespace)
-		if list != nil {
-			remaining = list.Items
+	Eventually(func() ([]unstructured.Unstructured, error) {
+		var subscriptiongvr = schema.GroupVersionResource{Group: "operators.coreos.com", Version: "v1alpha1", Resource: "subscriptions"}
```
exdx, line 408: nit: could we move this assignment out of the `Eventually()` into the var block above since it only needs to be done once (and is not part of the core polling logic)?

test/e2e/util.go
```diff
@@ -403,14 +404,13 @@ func TearDown(namespace string) {
...
+		list, err := dynamic.Resource(subscriptiongvr).Namespace(namespace).List(context.Background(), metav1.ListOptions{})
```
exdx, line 409: I think this logic makes sense, moving the call from the cache to the api-server. Maybe if this works well we can expand the change to the other types in `TearDown()` as well.

test/e2e/util.go
akihikokuroda: Sure.  I'll move it.  Thanks!

test/e2e/util.go
akihikokuroda: Sounds good to me.  Thanks!

timflannagan (approved):

PR #2624, opened 2022-02-14
Skip flaky tests

perdasilva:

This PR updates the e2e test mechanism to skip tests marked with a [FLAKY] tag.

#### Note
With Ginkgo v2.0 we can actually label tests, which might make this system less brittle. I just don't know what the upfront work required would be to make the upgrade. So, let's use this as a stop gap solution until we migrate and it should be easy enough to then migrate the flaky tests to use labels instead.

Merging PRs is a PITA atm because of flakes. This gives us a way to mark flakes and remove them from the critical path while they get fixed

**Reviewer Checklist**

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -19,7 +19,7 @@ jobs:
       - uses: actions/setup-go@v2
         with:
           go-version: '~1.17'
-      - run: make e2e-local E2E_TEST_CHUNK=${{ matrix.parallel-id }} E2E_TEST_NUM_CHUNKS=${{ strategy.job-total }} E2E_NODES=2 ARTIFACTS_DIR=./artifacts/
+      - run: make e2e-local E2E_TEST_CHUNK=${{ matrix.parallel-id }} E2E_TEST_NUM_CHUNKS=${{ strategy.job-total }} E2E_NODES=2 ARTIFACTS_DIR=./artifacts/ SKIP='\[FLAKY\]'
       - name: Archive Test Artifacts # test results, failed or not, are always uploaded.
         if: ${{ always() }}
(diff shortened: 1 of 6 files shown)


timflannagan (commented):
Any value in adding another workflow that's responsible for running the "[FLAKY]" regex so we don't skip the coverage of these tests? The overall CI signal is fairly watered down right now, so offloading them to their own workflow (instead of directly skipping them all together) wouldn't add a ton of value, but at least we wouldn't have to re-enable them, or manually promote a flaky test to a stable one, etc. over time.

That can be done as a follow-up as well.

perdasilva:
> Any value in adding another workflow that's responsible for running the "[FLAKY]" regex so we don't skip the coverage of these tests? The overall CI signal is fairly watered down right now, so offloading them to their own workflow (instead of directly skipping them all together) wouldn't add a ton of value, but at least we wouldn't have to re-enable them, or manually promote a flaky test to a stable one, etc. over time.
> 
> That can be done as a follow-up as well.

Yup - that's something I've had in mind to do as well. I want to find a way to execute these in isolation one at a time. To further reduce the failure noise while we fix them. I'm just re-running the e2e jobs over and over a few times until I feel the PR is "stable enough" and we have accounted for all (as many as possible) flaky tests and have those with a corresponding ticket.

timflannagan:
Yeah agreed - my main worry with simply skipping these tests as there's no tracking vehicle for fixing them right now, outside of creating individual issues and burning down those over time. Having a dedicated workflow that still gives CI signal, while not blocking a PR from merging as that flaky workflow isn't a required check, might be decent short term balance. I don't have a strong opinion on this approach as long as we commit to tracking these flakes $somewhere so they don't inevitably get swept under the rug.

PR #2631, opened 2022-02-15
fix: lower logging level to reduce noisiness 

exdx:
This change moves following logs to the debug level, to lower the amount of logs generated by OLM by default:
```
* syncing catalog source for annotation templates
* catalog polling result: no update
* catalog update required
* state.Key.Namespace={NS} state.Key.Name={NAME} state.State={STATE}
* multiple pods found for selector
* checking {CSVNAME}
```

This PR does not address the logs generated by conflicts as those are emitted by the queue_informer and it is not readily apparent how to turn the logging level down for those particular events. 

Large numbers of uneventful logs generated by OLM causes unnecessarily high memory usage by log aggregation tools that parse OLM logs. 

**Reviewer Checklist**

### DIFF
--- a/pkg/lib/operatorstatus/status.go
+++ b/pkg/lib/operatorstatus/status.go
@@ -10,7 +10,7 @@ import (
 	configv1 "github.com/openshift/api/config/v1"
 	configv1client "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/clientset/versioned"
-	log "github.com/sirupsen/logrus"
+	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -33,7 +33,8 @@ const (
 	installPlanResource           = "installplans"
 )
 
-func MonitorClusterStatus(name string, syncCh <-chan error, stopCh <-chan struct{}, opClient operatorclient.ClientInterface, configClient configv1client.ConfigV1Interface, crClient versioned.Interface) {
+func MonitorClusterStatus(name string, syncCh <-chan error, stopCh <-chan struct{}, opClient operatorclient.ClientInterface,
+	configClient configv1client.ConfigV1Interface, crClient versioned.Interface, log *logrus.Logger) {
 	var (
 		syncs              int
 		successfulSyncs    int
@@ -123,7 +124,7 @@ func MonitorClusterStatus(name string, syncCh <-chan error, stopCh <-chan struct
 				log.Errorf("Failed to create cluster operator: %v\n", createErr)
 				return
 			}
-			created.Status.RelatedObjects, err = relatedObjects(name, opClient, crClient)
+			created.Status.RelatedObjects, err = relatedObjects(name, opClient, crClient, log)
 			if err != nil {
 				log.Errorf("Failed to get related objects: %v", err)
 			}
@@ -208,7 +209,7 @@ func MonitorClusterStatus(name string, syncCh <-chan error, stopCh <-chan struct
 		}
 
 		// always update the related objects in case changes have occurred
-		existing.Status.RelatedObjects, err = relatedObjects(name, opClient, crClient)
+		existing.Status.RelatedObjects, err = relatedObjects(name, opClient, crClient, log)
 		if err != nil {
 			log.Errorf("Failed to get related objects: %v", err)
 		}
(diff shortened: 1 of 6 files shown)


pkg/controller/operators/olm/operator.go
```diff
@@ -1246,7 +1246,7 @@ func (a *Operator) allNamespaceOperatorGroups() ([]*v1.OperatorGroup, error) {
 }
 
 func (a *Operator) syncOLMConfig(obj interface{}) (syncError error) {
-	a.logger.Debug("Processing olmConfig")
+	a.logger.Debugf("Processing olmConfig")
```
exdx, line 1249: I'm not sure why, but this log comes in at the info level on an openshift cluster, even though in the code it's at a debug level. Maybe this change would cause it to remain at the debug level and not appear in the logs, not sure.

timflannagan (commented):
There's probably a good chunk of places throughout OLM we can make a better job with logging: some obvious that come to mind is the operatorcondition and operatorconditiongenerator controllers, as they produce to the overall noisiness there as well. I tried to tackle that portion in https://github.com/operator-framework/operator-lifecycle-manager/pull/2503 but never followed up with that PR.

pkg/controller/operators/olm/operator.go
exdx: Yes, but looking at the logs in the olm operator, the logging level is set to info, so it seems unexpected
```
time="2022-02-15T16:17:33Z" level=info msg="log level info"
```

exdx:
> There's probably a good chunk of places throughout OLM we can make a better job with logging: some obvious that come to mind is the operatorcondition and operatorconditiongenerator controllers, as they produce to the overall noisiness there as well. I tried to tackle that portion in #2503 but never followed up with that PR.

That change looks helpful too. I think we should merge that small PR separately, WDYT?

pkg/controller/operators/olm/operator.go
njhale: Some possibilities:
- the user has set the debug flag to true on the olm-operator deployment in OpenShift (if the logs you're citing come from there, this would be unlikely)
- there is a version of OLM we released that has it set true by default (unlikely for the same reason)
- the logger instantiated here isn't the same one OLM's controller ultimately uses (we fail to pass the logger in somewhere)

njhale, line 1249: Switching from `Debug` to `Debugf` shouldn't change the level at which it's output.

pkg/lib/operatorstatus/status.go
```diff
@@ -267,7 +267,7 @@ func findOperatorStatusCondition(conditions []configv1.ClusterOperatorStatusCond
 // RelatedObjects are consumed by https://github.com/openshift/must-gather
 func relatedObjects(name string, opClient operatorclient.ClientInterface, crClient versioned.Interface) ([]configv1.ObjectReference, error) {
 	var objectReferences []configv1.ObjectReference
-	log.Infof("Adding related objects for %v", name)
+	log.Debugf("Adding related objects for %v", name)
```
njhale, line 270: hmm, it looks like everything in here is using the global logger instead of one that's passed in. We should make sure the level is set correctly for this by the init logic [mentioned in an earlier thread](https://github.com/operator-framework/operator-lifecycle-manager/pull/2631/files#r807155143).

njhale (changes requested):
Thanks for putting this together! I have one or two small items to be addressed before we merge:

pkg/controller/operators/olm/operator.go
exdx: Yes, that idea didn't really make sense. Still trying to track down why this log comes through though.

timflannagan:
@exdx can you update the branch when you get a chance? I think this looks ready to go imo.

exdx:
> @exdx can you update the branch when you get a chance? I think this looks ready to go imo.

Pushed, thanks. I think this PR with it's small scope is good to go as-is and is backportable.  It basically just takes logs at the info level that should be at the debug level and moves them down to debug. 

Furthermore, we should also explore ways to lower the sync error logs coming from the "object has been modified" type errors that we see in the logs, for example: 

```
error="Operation cannot be fulfilled on catalogsources.operators.coreos.com \"operatorhubio-catalog\": the object has been modified; please apply your changes to the latest version and try again" id=OTXzH source=operatorhubio-catalog
```

However, changing those logs is slightly different and produces a more complex diff. Unless we find a way to suppress the logs and keep the error being propagated down to the sync handler, it would change the behavior of OLM. I will open a tracking issue to investigate making that change -- I don't expect it to be backportable necessarily.

exdx:
Related: #2672

timflannagan (approved):
/lgtm

PR #2632, opened 2022-02-15
Replace hardcoded resolver cache mutation with CSV cache.Source.

benluddy:
The resolver depends on getting a cache of snapshots of all available catalogs, then synthesizing an extra snapshot containing the CSVs in the target namespace and inserting it into the cache. The `cache.Source` and `cache.SourceProvider` interfaces were introduced to support injecting cache entries from various non-registry sources. Implementing these interfaces reduces the coupling between the core resolution component and the CSV API, which should be a benefit if/when it comes time to support new on-cluster dependency resolution APIs. As an added benefit, the exported surface of the cache package shrinks.

### DIFF
--- a/pkg/controller/registry/resolver/step_resolver.go
+++ b/pkg/controller/registry/resolver/step_resolver.go
@@ -9,8 +9,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/labels"
-	"k8s.io/client-go/kubernetes"
 
 	"github.com/operator-framework/api/pkg/operators/v1alpha1"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/clientset/versioned"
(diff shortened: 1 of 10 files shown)


pkg/controller/registry/resolver/step_resolver.go
```diff
@@ -37,23 +35,30 @@ type OperatorStepResolver struct {
...
 
 var _ StepResolver = &OperatorStepResolver{}
 
-func NewOperatorStepResolver(lister operatorlister.OperatorLister, client versioned.Interface, kubeclient kubernetes.Interface,
-	globalCatalogNamespace string, provider RegistryClientProvider, log logrus.FieldLogger) *OperatorStepResolver {
+func NewOperatorStepResolver(lister operatorlister.OperatorLister, client versioned.Interface, globalCatalogNamespace string, provider RegistryClientProvider, log logrus.FieldLogger) *OperatorStepResolver {
+	cacheSourceProvider := &mergedSourceProvider{
+		sps: []cache.SourceProvider{
+			SourceProviderFromRegistryClientProvider(provider, log),
+			&csvSourceProvider{
```
timflannagan, line 49: Do we need to make an exported constructor for this so we can instantiate a CSVSourceProvider outside of this resolution package?

benluddy: No, the CSV would also leave this package when that happens.

PR #2647, opened 2022-02-16
Rename solver.Installable to solver.Variable.

benluddy:
As part of preparing the resolution component (and subcomponents) for
use as a library, this clarifies the purpose of the Variable type: a
variable in a constraint satisfaction problem. The original name,
Installable, unnecessarily ties it to the resolver package's operator
installability use case. The solver package finds solutions to
constraint satisfaction problems in general.


pkg/controller/registry/resolver/solver/lit_mapping.go
```diff
@@ -107,8 +107,8 @@ func (d *litMapping) ConstraintOf(m z.Lit) AppliedConstraint {
 	}
 	d.errs = append(d.errs, fmt.Errorf("no constraint corresponding to %s", m))
 	return AppliedConstraint{
-		Installable: zeroInstallable{},
-		Constraint:  zeroConstraint{},
+		Variable:   zeroVariable{},
```
timflannagan, line 110: :+1:

pkg/controller/registry/resolver/solver/variable.go
```diff
@@ -14,27 +14,27 @@ func IdentifierFromString(s string) Identifier {
 	return Identifier(s)
 }
 
-// Installable values are the basic unit of problems and solutions
+// Variable values are the basic unit of problems and solutions
 // understood by this package.
```
timflannagan, line 18: Any value in beefing up this godoc comment to tie it back to the CSP?

timflannagan:
Thanks - I think this is a step in the right direction, as variables like `BundleInstallable` didn't make much intuitive sense to me while reading these packages throughout the codebase. The only other thing I'd like to see is better godoc comments any user-facing type/interface/etc. that would make traversing this package a bit easier, and could help us work towards making the responsibilities of each level of abstraction here a bit more clear.

PR #2648, opened 2022-02-16
Remove CatalogSourceLister usage from resolver/cache.

benluddy:
The cache hardcodes the use of a CatalogSourceLister (imported from a
non-resolver package in the OLM module) in order to assign source
priorities based on the spec.priority field of CatalogSource. Not all
cache sources map to a CatalogSource, so this can instead accept a
priority provider interface and the resolver -> non-resolver imports
can be removed.


timflannagan:
Ah, it looks like approval-acts-as-lgtm doesn't work on this repository.

/lgtm

njhale (approved):
/lgtm

PR #2649, opened 2022-02-16
Remove generated and unused resolver/fakes subpackage.

benluddy:
The go:generate comment that produces this was removed in 9b537bf1079d55244c0cd733fc55d12d02c31574, and nothing appears to be using it.


benluddy:
Is this linter broken for commits that have fewer packages than their parent did?

benluddy:
This has to be a bug in the GH action, especially since it's happening during "package loading" and this PR does nothing besides delete a package.

PR #2652, opened 2022-02-16
chore(gh): use golang-ci lint directly in sanity workflow

njhale:
Use golang-ci lint directly in the "sanity" GitHub Workflow. This
avoids the use of a third-party golangci-lint action which can
sporadically timeout, causing the workflow to fail.


pkg/controller/registry/grpc/source.go
```diff
@@ -147,7 +147,8 @@ func getEnvAny(names ...string) string {
 }
 
 func grpcConnection(address string) (*grpc.ClientConn, error) {
-	dialOptions := []grpc.DialOption{grpc.WithInsecure()}
+	creds := insecure.NewCredentials()
```
dinhxuanvu, line 150: Are these changes required because of what the lint took found?

tylerslaton:
Looks like it needs a rebase and then I'll go ahead and LGTM it, we had to do something similar for Combo.

dinhxuanvu:
The sanity test still fails due to timeout.

exdx:
Surprising that this change causes such a large diff in our dependencies, but I guess it's fine 🤷

pkg/controller/registry/grpc/source.go
njhale: yup

njhale:
/hold

This is still timing out. Going to do some more testing.

pkg/controller/registry/grpc/source.go
dinhxuanvu: 👍

njhale:
we've fixed the caching with the first-class github action

PR #2656, opened 2022-02-17
Remove vestigal cache.OperatorSet type.

benluddy:
Nothing of consequence has used OperatorSet in ages, but it's been
lingering in tests and as the result type of the main resolution
method.


timflannagan (commented):
Should we split out the operatorset changes from the renaming of the satresolver type?

benluddy:
I don't think it's worth the extra PR churn, especially since there aren't any Go API stability guarantees for this package.

PR #2658, opened 2022-02-17
Remove unused ttl field from cache.Cache.

benluddy:
The cache itself isn't responsible for snapshot invalidation anymore,
but this field has been left behind.


joelanford:
🎉 Thanks!

I guess this was a red herring when Ben and I we're debugging the CPU spikes of catalog pods. I thought this may have had something to do with it.

benluddy:
@joelanford Still possible, I moved the hardcoded TTL out of the cache just this week, but there is a still an at-most five minute TTL on snapshots of catalog content from registry pods (https://github.com/operator-framework/operator-lifecycle-manager/pull/2651).

PR #2659, opened 2022-02-17
[WIP] Create catalog pods as pullIFNotPresent, use a pullalways probe pod to check for new SHAs

bparees:
Instead of just creating catalog pods as pull always (becomes a problem in temporarily disconnected clusters), create themn as pull if not present.

when creating a new pod to check for updates, create it as pullalways, see if it gets a new SHA, if so, delete it and recreate it as a pullifnotpresent pod but using the new SHA.

Clusters w/ only intermittent registry connection experience pod startup failures when
catalog pods are restarted while the registry is inaccessible.  This change
reduces the period of time during which registry access is required.

The downside is that when there is an update to the catalog image for a given catalogsource, we will do 2 pod creations/startups instead of just one. (And, if the 2 pods land on different nodes, we'll also have to pull the image twice).

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/registry/reconciler/configmap_test.go
+++ b/pkg/controller/registry/reconciler/configmap_test.go
@@ -198,7 +198,7 @@ func objectsForCatalogSource(catsrc *v1alpha1.CatalogSource) []runtime.Object {
 		if catsrc.Spec.Image != "" {
 			decorated := grpcCatalogSourceDecorator{catsrc}
 			objs = clientfake.AddSimpleGeneratedNames(
-				decorated.Pod(catsrc.GetName()),
+				decorated.Pod(catsrc.GetName(), ""),
 				decorated.Service(),
 				decorated.ServiceAccount(),
 			)
--- a/pkg/controller/registry/reconciler/grpc.go
+++ b/pkg/controller/registry/reconciler/grpc.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"hash/fnv"
+	"runtime/debug"
 	"time"
 
 	"github.com/pkg/errors"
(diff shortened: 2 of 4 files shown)


bparees:
notes to self:  there's some logic in EnsureRegistryServer that tries really hard to ensure the existing catalog pod has an image field value that exactly matches the image field value from the catalogsource object, which makes it hard to create a pod based on the SHA instead.

Everything else about this PR seems to work as expected, but then that reconcile logic comes in and stomps the SHA based pod back to being whatever value is in the catalogSource.

bparees:
well this "works" w/ the caveat that for an initial catalog creation:

1) a pod is created as pullalways, referencing the catalogsource image pullspec
2) the update pod gets created as pullalways referencing the pullspec
3) if the update pod does not pull a new SHA, the update pod is thrown away and the pod from (1) is kept, meaning we still have a pod using pullalways and a tag

if the update pod *does* pull a new SHA, this code correctly deletes the update pod and creates a new "update" pod using pullifnotpresent with the newly determined SHA, which then gets promoted and the original catalog pod is deleted.

So until the first SHA update happens, the catalog pod is still using pullalways and a tag.

But the more important thing i realized while doing this is it doesn't really help at all because these pods are not deployment managed, so if the catalog pod gets deleted/drained and recreated on a new node, we're back to having to do an initial pullalways of a tag.

So to really solve this problem, we need to:

1) create the pullalways updated pod using the pullspec from the CatalogSource spec
2) determine if the update pod gets a new SHA
3) place that SHA on the CatalogSource status for future reference
4) whenever creating a catalog pod, create it based on the SHA that is on the CatalogSource status, using pullifnotpresent.

bparees:
i'm going to close this since per https://github.com/operator-framework/operator-lifecycle-manager/pull/2659#issuecomment-1072655734 i think it's a dead end, but i think the overall direction is worth considering for how catalog pods are created/managed, to shift them to a SHA-based, pullifnotpresent configuration.

PR #2661, opened 2022-02-17
Remove import of operators.coreos.com lister package from resolver.

benluddy:
A test double was missed when cache.SourcePriorityProvider was
introduced. These resolver-to-nonresolver package imports will make it
difficult for the resolution component to stand on its own as a
library module.

### DIFF
--- a/pkg/controller/registry/resolver/resolver_test.go
+++ b/pkg/controller/registry/resolver/resolver_test.go
@@ -13,12 +13,8 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/labels"
-
 	"github.com/operator-framework/api/pkg/constraints"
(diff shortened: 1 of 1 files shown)


pkg/controller/registry/resolver/resolver_test.go
```diff
@@ -409,38 +405,10 @@ func TestSolveOperators_FindLatestVersionWithNestedDependencies(t *testing.T) {
...
-	return nil, errors.New("stub not found")
-}
-
-func (l *stubCatalogSourceLister) CatalogSources(namespace string) listersv1alpha1.CatalogSourceNamespaceLister {
-	return &stubCatalogSourceLister{
-		catsrcs:   l.catsrcs,
-		namespace: namespace,
-	}
+func (spp stubSourcePriorityProvider) Priority(k cache.SourceKey) int {
+	return spp[k]
```
timflannagan, line 411: Do we need to null check this $somewhere to avoid a panic?

timflannagan: Oh, disregard - I think this is fine as it should just return 0 when indexing on a key that doesn't exist.

/lgtm

PR #2673, opened 2022-02-23
Fixes difficult to understand error message when CSV is missing APIVersion

nsapse:

Modifies CSV unbundle process so that after unbundling APIVersion or Kind are missing the installPlan is failed and the user receives a helpful and descriptive failure message letting them know that data was missing.

Per BZ#1982737[](https://bugzilla.redhat.com/show_bug.cgi?id=1982737) currently when a CSV is missing APIVersion the associated installPlan fails with a cryptic message involving failure to create a service account. I initially tried to solve this by inferring the APIVersion if left blank per PR#2609[](https://github.com/operator-framework/operator-lifecycle-manager/pull/2609#issue-1122164908) however the conversation following the PR made it clear that approach was not an acceptable fix. 

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/errors/errors.go
+++ b/pkg/controller/errors/errors.go
@@ -47,6 +47,21 @@ func IsMultipleExistingCRDOwnersError(err error) bool {
 	return false
 }
 
+type FatalError struct {
+	error
+}
+
+func NewFatalError(err error) FatalError {
+	return FatalError{err}
+}
+func IsFatal(err error) bool {
+	switch err.(type) {
+	case FatalError:
+		return true
+	}
+	return false
+}
+
 // GroupVersionKindNotFoundError occurs when we can't find an API via discovery
 type GroupVersionKindNotFoundError struct {
 	Group   string
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -1376,7 +1376,13 @@ type UnpackedBundleReference struct {
 	Properties             string `json:"properties"`
 }
 
-// unpackBundles makes one walk through the bundlelookups and attempts to progress them
+/* unpackBundles makes one walk through the bundlelookups and attempts to progress them
+Returns:
+    unpacked: bool - If the bundle was successfully unpacked
+    out:      *v1alpha1.InstallPlan - the resulting installPlan
+	error:    error
+*/
+
 func (o *Operator) unpackBundles(plan *v1alpha1.InstallPlan) (bool, *v1alpha1.InstallPlan, error) {
 	out := plan.DeepCopy()
 	unpacked := true
@@ -1427,6 +1433,10 @@ func (o *Operator) unpackBundles(plan *v1alpha1.InstallPlan) (bool, *v1alpha1.In
 		// Ensure that bundle can be applied by the current version of OLM by converting to steps
 		steps, err := resolver.NewStepsFromBundle(res.Bundle(), out.GetNamespace(), res.Replaces, res.CatalogSourceRef.Name, res.CatalogSourceRef.Namespace)
 		if err != nil {
+			if fatal := olmerrors.IsFatal(err); fatal {
+				return false, nil, err
+			}
+
 			errs = append(errs, fmt.Errorf("failed to turn bundle into steps: %v", err))
 			unpacked = false
 			continue
@@ -1615,6 +1625,14 @@ func (o *Operator) syncInstallPlans(obj interface{}) (syncError error) {
 	if len(plan.Status.BundleLookups) > 0 {
 		unpacked, out, err := o.unpackBundles(plan)
 		if err != nil {
+			// If the error was fatal capture and fail
+			if fatal := olmerrors.IsFatal(err); fatal {
+				if err := o.transitionInstallPlanToFailed(plan, logger, v1alpha1.InstallPlanReasonInstallCheckFailed, err.Error()); err != nil {
+					// retry for failure to update status
+					syncError = err
+					return
+				}
+			}
 			// Retry sync if non-fatal error
 			syncError = fmt.Errorf("bundle unpacking failed: %v", err)
 			return
--- a/pkg/controller/registry/resolver/steps.go
+++ b/pkg/controller/registry/resolver/steps.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"strings"
 
+	olmerrors "github.com/operator-framework/operator-lifecycle-manager/pkg/controller/errors"
 	"github.com/operator-framework/operator-registry/pkg/api"
 	extScheme "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/scheme"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -116,6 +117,15 @@ func NewStepResourceFromBundle(bundle *api.Bundle, namespace, replaces, catalogS
 	csv, err := V1alpha1CSVFromBundle(bundle)
 	if err != nil {
 		return nil, err
+	}
+
+	// Check unpacked bundled for for missing APIVersion or Kind
+	if csv.APIVersion == "" {
+		return nil, olmerrors.NewFatalError(fmt.Errorf("bundle CSV %s missing APIVersion", csv.Name))
+	}
+
+	if csv.Kind == "" {
+		return nil, olmerrors.NewFatalError(fmt.Errorf("bundle CSV %s missing Kind", csv.Name))
 	}
 
 	csv.SetNamespace(namespace)
--- a/test/e2e/catalog_e2e_test.go
+++ b/test/e2e/catalog_e2e_test.go
@@ -7,11 +7,13 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"path/filepath"
 	"strconv"
 	"strings"
 	"time"
 
 	operatorsv1 "github.com/operator-framework/api/pkg/operators/v1"
+	operatorsv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 	k8serror "k8s.io/apimachinery/pkg/api/errors"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
@@ -40,6 +42,7 @@ import (
 const (
 	openshiftregistryFQDN = "image-registry.openshift-image-registry.svc:5000"
 	catsrcImage           = "docker://quay.io/olmtest/catsrc-update-test:"
+	badCSVDir             = "bad-csv"
 )
 
 var _ = Describe("Starting CatalogSource e2e tests", func() {
@@ -1340,6 +1343,68 @@ var _ = Describe("Starting CatalogSource e2e tests", func() {
 			}
 		}
 	})
+
+	When("A CatalogSource is created with an operator that has a CSV with missing metadata.ApiVersion", func() {
+
+		var (
+			magicCatalog      MagicCatalog
+			catalogSourceName string
+			subscription      *operatorsv1alpha1.Subscription
+			c                 client.Client
+		)
+
+		BeforeEach(func() {
+			c = ctx.Ctx().Client()
+
+			provider, err := NewFileBasedFiledBasedCatalogProvider(filepath.Join(testdataDir, badCSVDir, "bad-csv.yaml"))
+			Expect(err).To(BeNil())
+
+			catalogSourceName = genName("cat-bad-csv")
+			magicCatalog = NewMagicCatalog(c, ns.GetName(), catalogSourceName, provider)
+			Expect(magicCatalog.DeployCatalog(context.Background())).To(BeNil())
+
+		})
+
+		AfterEach(func() {
+			TeardownNamespace(ns.GetName())
+		})
+
+		When("A Subscription is created catalogSource built with the malformed CSV", func() {
+			BeforeEach(func ()  {
+				subscription = &operatorsv1alpha1.Subscription{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      fmt.Sprintf("%s-sub", catalogSourceName),
+						Namespace: ns.GetName(),
+					},
+					Spec: &operatorsv1alpha1.SubscriptionSpec{
+						CatalogSource:          catalogSourceName,
+						CatalogSourceNamespace: ns.GetName(),
+						Channel:                "stable",
+						Package:                "packageA",
+					},
+				}
+				Expect(c.Create(context.Background(), subscription)).To(BeNil())
+			})
+
+			It("fails with a ResolutionFailed error condition, and a message that highlights the missing field in the CSV", func() {
+
+				subscription, err := fetchSubscription(crc, subscription.GetNamespace(), subscription.GetName(), subscriptionHasInstallPlanChecker)
+				Expect(err).Should(BeNil())
+				installPlanName := subscription.Status.Install.Name
+
+				// ensure we wait for the installPlan to fail before moving forward then fetch the subscription again
+				_, err = fetchInstallPlan(GinkgoT(), crc, installPlanName, subscription.GetNamespace(), buildInstallPlanPhaseCheckFunc(operatorsv1alpha1.InstallPlanPhaseFailed))
+				Expect(err).To(BeNil())
+				subscription, err = fetchSubscription(crc, subscription.GetNamespace(), subscription.GetName(), subscriptionHasInstallPlanChecker)
+				Expect(err).To(BeNil())
+
+				// expect the message that API missing
+				failingCondition := subscription.Status.GetCondition(operatorsv1alpha1.SubscriptionInstallPlanFailed)
+				Expect(failingCondition.Message).To(ContainSubstring("missing APIVersion"))
+			})
+		})
+	})
+
 })
 
 func getOperatorDeployment(c operatorclient.ClientInterface, namespace string, operatorLabels labels.Set) (*appsv1.Deployment, error) {
--- a/test/e2e/testdata/bad-csv/bad-csv.yaml
+++ b/test/e2e/testdata/bad-csv/bad-csv.yaml
@@ -1,0 +1,25 @@
+---
+schema: olm.package
+name: packageA
+defaultChannel: stable
+---
+schema: olm.channel
+package: packageA
+name: stable
+entries:
+  - name: bad-csv
+---
+schema: olm.bundle
+name: bad-csv
+package: packageA
+image: quay.io/olmtest/missing_api_version:latest
+properties:
+  - type: olm.gvk
+    value:
+      group: example.com
+      kind: TestA
+      version: v1alpha1
+  - type: olm.package
+    value:
+      packageName: packageA
+      version: 1.0.0


pkg/controller/registry/resolver/steps.go
```diff
@@ -118,6 +118,15 @@ func NewStepResourceFromBundle(bundle *api.Bundle, namespace, replaces, catalogS
 		return nil, err
 	}
 
+	// Check unpacked bundled for for missing APIVersion or Kind
+	if csv.APIVersion == "" {
+		return nil, fmt.Errorf("bundle CSV %s missing metadata.APIVersion", csv.Name)
```
benluddy, line 123: Nit: `apiVersion` and `kind` aren't subfields of `metadata`.

pkg/controller/operators/catalog/operator.go
```diff
@@ -1347,11 +1347,22 @@ type UnpackedBundleReference struct {
...
-// unpackBundles makes one walk through the bundlelookups and attempts to progress them
-func (o *Operator) unpackBundles(plan *v1alpha1.InstallPlan) (bool, *v1alpha1.InstallPlan, error) {
+/* unpackBundles makes one walk through the bundlelookups and attempts to progress them
+Returns:
+    unpacked: bool - If the bundle was successfully unpacked
+    out: *v1alpha1.InstallPlan - the resulting installPlan
+    errorWasFatal: bool - If a returned error is fatal and should result in IP failure
+	error: error
+*/
+
```
dinhxuanvu, line 1357: No need to have an empty line here.

pkg/controller/operators/catalog/operator.go
```diff
@@ -1347,11 +1347,22 @@ type UnpackedBundleReference struct {
...
+func (o *Operator) unpackBundles(plan *v1alpha1.InstallPlan) (bool, *v1alpha1.InstallPlan, bool, error) {
 	out := plan.DeepCopy()
 	unpacked := true
 
+	// We should return whether the error which caused failure was transient (install should retry) or
+	// fatal (it should not)
+	errorWasFatal := false
```
dinhxuanvu, line 1364: `isFatal` would be sufficient for variable name. Also `var isFatal bool` would be good as well.

nsapse: var isFatal bool definitely does a better job of communicating what is going on in a readable way. Will implement.

dinhxuanvu:
Just a friendly reminder here is that you need to identify if an e2e test case is needed. The unit test is preferred if e2e is unnecessary.

test/e2e/installplan_e2e_test.go
```diff
@@ -4028,8 +4029,182 @@ var _ = Describe("Install Plan", func() {
 		}).Should(HavePhase(operatorsv1alpha1.InstallPlanPhaseComplete))
 
 	})
+
+
+FDescribe("When an operator is attempted to be installed from a catalog", func ()  {
```
tylerslaton, line 4034: Make sure to remove this `FDescribe` before we get this through, please.

anik120: 👍🏽 fyi the test will fail if the focus is on so no focused test can get in anyway.

exdx (commented):
Nice job -- I just had a thought on the overall purpose of the implementation. Code-wise everything looks good.

pkg/controller/registry/resolver/steps.go
```diff
@@ -118,6 +118,15 @@ func NewStepResourceFromBundle(bundle *api.Bundle, namespace, replaces, catalogS
 		return nil, err
 	}
 
+	// Check unpacked bundled for for missing APIVersion or Kind
+	if csv.APIVersion == "" {
+		return nil, fmt.Errorf("bundle CSV %s missing APIVersion", csv.Name)
```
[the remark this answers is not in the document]
exdx: I see https://github.com/operator-framework/operator-lifecycle-manager/pull/2609 has discussion around this specific issue, based on that I agree with what was stated and we can keep this as is.

pkg/controller/registry/resolver/steps.go
```diff
@@ -118,6 +118,15 @@ func NewStepResourceFromBundle(bundle *api.Bundle, namespace, replaces, catalogS
...
+	}
+
+	if csv.Kind == "" {
```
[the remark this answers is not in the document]
nsapse: Following up on this comment that I missed. Question before I implement your suggested change - does it maybe make sense from a UX perspective to return an error here anyway so that there is consistent behavior when a field is missing from a CSV?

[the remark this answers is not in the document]
exdx: I think it's reasonable for the sake of consistency to return an error -- it's doesn't make total sense, since we are handling a concrete CSV type in the code (which we inferred somehow from the underlying JSON representation of the CSV in the bundle) to then turn around and say "this CSV Kind is undefined", but I guess it's simply a matter of preference: silently accepting the manifest (which I argue a smart package manager should be doing) versus erroring. But I'm ok with staying consistent. 

/lgtm

pkg/controller/operators/catalog/operator.go
```diff
@@ -1430,10 +1442,13 @@ func (o *Operator) unpackBundles(plan *v1alpha1.InstallPlan) (bool, *v1alpha1.In
 
 	if err := utilerrors.NewAggregate(errs); err != nil {
 		o.logger.Debugf("failed to unpack bundles: %v", err)
-		return false, nil, err
+		if errorWasFatal {
```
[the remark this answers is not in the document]
nsapse: Would this be your/team style preference? I originally had something similar but refactored because so many returns looked a bit unsightly to me.

test/e2e/catalog_e2e_test.go
```diff
@@ -1229,6 +1229,100 @@ var _ = Describe("Catalog represents a store of bundles which OLM can use to ins
 			}
 		}
 	})
+
+	When("A CatalogSource is created with an operator that has a CSV with missing metadata.ApiVersion", func() {
```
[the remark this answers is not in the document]
nsapse: Unsure, @anik120 suggested this formatting.

test/e2e/catalog_e2e_test.go
```diff
@@ -1229,6 +1229,100 @@ var _ = Describe("Catalog represents a store of bundles which OLM can use to ins
...
+					Channel:                stableChannel,
+					StartingCSV:            csv.GetName(),
+					InstallPlanApproval:    v1alpha1.ApprovalAutomatic,
+				}
+
+				subscriptionName = genName("sub-")
+				createSubscriptionForCatalogWithSpec(GinkgoT(), crc, ns.GetName(), subscriptionName, subscriptionSpec)
+			})
+
+			It("The Subscription fails with a ResolutionFailed error condition, and a message that highlights the missing field in the CSV", func() {
```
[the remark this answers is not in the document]
nsapse: Will make this change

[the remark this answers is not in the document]
nsapse: That seems fair. Will implement.

pkg/controller/operators/catalog/operator.go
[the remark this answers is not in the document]
nsapse: Also re: isRetryable() - I will try hacking on the code and see if I can make that work.

pkg/controller/operators/catalog/operator.go
nsapse: To begin I think I may need guidance on which errors should be considered "retryable" and which should not.

nsapse: I have implemented a basic version of this. Let me know how it looks.

test/e2e/catalog_e2e_test.go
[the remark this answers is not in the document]
anik120: It's quite normal to have nested `When` blocks actually. Here's an example: https://github.com/operator-framework/operator-lifecycle-manager/blob/master/test/e2e/metrics_e2e_test.go#L57-L126
There are more nested `When` blocks in that test suite

pkg/controller/operators/catalog/operator.go
```diff
@@ -1352,8 +1352,15 @@ type UnpackedBundleReference struct {
...
-func (o *Operator) unpackBundles(plan *v1alpha1.InstallPlan) (bool, *v1alpha1.InstallPlan, error) {
+/* unpackBundles makes one walk through the bundlelookups and attempts to progress them
+Returns:
+    unpacked: bool - If the bundle was successfully unpacked
+    out:      *v1alpha1.InstallPlan - the resulting installPlan
+	isFatal:  bool - if the resulting error was fatal (not retryable)
+	error:    error
+*/
+
+func (o *Operator) unpackBundles(plan *v1alpha1.InstallPlan) (bool, *v1alpha1.InstallPlan, bool, error) {
```
perdasilva, line 1363: Great start! I would now remove the errorIsFatal boolean from the return. Then the callers of unpackBundles can just us olmerrors.IsRetryable

pkg/controller/errors/errors.go
```diff
@@ -47,6 +47,31 @@ func IsMultipleExistingCRDOwnersError(err error) bool {
...
+
+func IsRetryable(err error) bool {
+	switch err.(type) {
+	case InstallPlanError:
+		return false
+	}
+
+	return true
+}
+
```
perdasilva, line 74:
```suggestion
type RetryableError struct {
	error
}

func NewRetryableError(err error) RetryableError {
	return RetryableError{err}
}

func IsRetryable(err error) bool {
	switch err.(type) {
	case RetryableError:
		return true
	}

	return false
}
```

What do you think?

pkg/controller/operators/catalog/operator.go
```diff
@@ -1589,8 +1600,16 @@ func (o *Operator) syncInstallPlans(obj interface{}) (syncError error) {
 	// Attempt to unpack bundles before installing
 	// Note: This should probably use the attenuated client to prevent users from resolving resources they otherwise don't have access to.
 	if len(plan.Status.BundleLookups) > 0 {
-		unpacked, out, err := o.unpackBundles(plan)
+		unpacked, out, errorIsFatal, err := o.unpackBundles(plan)
 		if err != nil {
+			// If the error was fatal capture and fail
+			if errorIsFatal {
```
perdasilva, line 1606: Remove errorIsFalar and use error.IsRetryable here

pkg/controller/errors/errors.go
nsapse: Works for me. I had conceived of a solution where already existing error classes were divided into "retryable" and "unretryable" by a big switch statement in the sky but this is much cleaner. Will implement.

test/e2e/catalog_e2e_test.go
```diff
@@ -1309,6 +1309,181 @@ var _ = Describe("Starting CatalogSource e2e tests", func() {
...
+			og = operatorsv1.OperatorGroup{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      fmt.Sprintf("%s-operatorgroup", namespaceName),
+					Namespace: namespaceName,
+				},
+			}
+			ns = SetupGeneratedTestNamespaceWithOperatorGroup(namespaceName, og)
+
+			// create a simple catalogsource
+			packageName = genName("good-csv")
```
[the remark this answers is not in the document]
perdasilva: just a nit: this way the package name will be good-csv-XXXX

test/e2e/catalog_e2e_test.go
```diff
@@ -1309,6 +1309,181 @@ var _ = Describe("Starting CatalogSource e2e tests", func() {
 			}
 		}
 	})
+
+	When("A CatalogSource is created with a properly formed CSV", func() {
+		var (
+			og               operatorsv1.OperatorGroup
+			namespaceName    string
+			ns               corev1.Namespace
```
perdasilva, line 1317: You don't need another ns. There's a global one that gets created and torndown before every spec: https://github.com/operator-framework/operator-lifecycle-manager/blob/master/test/e2e/catalog_e2e_test.go#L61

test/e2e/catalog_e2e_test.go
```diff
@@ -1309,6 +1309,181 @@ var _ = Describe("Starting CatalogSource e2e tests", func() {
...
+		})
+
+	})
+
+	When("A CatalogSource is created with an operator that has a CSV with missing metadata.ApiVersion", func() {
+
+		var (
+			og               operatorsv1.OperatorGroup
+			namespaceName    string
+			ns               corev1.Namespace
```
perdasilva, line 1399: same comment as above

test/e2e/catalog_e2e_test.go
```diff
@@ -1309,6 +1309,179 @@ var _ = Describe("Starting CatalogSource e2e tests", func() {
...
+
+		BeforeEach(func() {
+			namespaceName = genName("good-csv-")
+			og = operatorsv1.OperatorGroup{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      fmt.Sprintf("%s-operatorgroup", namespaceName),
+					Namespace: namespaceName,
+				},
+			}
+			ns = SetupGeneratedTestNamespaceWithOperatorGroup(namespaceName, og)
```
timflannagan, line 1334: Can we piggy back off the top-level namespace that's been already created for us before this test case runs: https://github.com/operator-framework/operator-lifecycle-manager/blob/4be8b0c5ad2fd8cebe041ce9d7992019b8f02fbb/test/e2e/catalog_e2e_test.go#L54-L61?

test/e2e/catalog_e2e_test.go
```diff
@@ -1309,6 +1309,179 @@ var _ = Describe("Starting CatalogSource e2e tests", func() {
...
+			csv = newCSV(packageStable, ns.GetName(), "", semver.MustParse("0.1.0"), []apiextensions.CustomResourceDefinition{crd}, nil, nil)
+
+			catalogSourceName := genName("mock-ocs-")
+			cs, _ = createInternalCatalogSource(c, crc, catalogSourceName, ns.GetName(), manifests, []apiextensions.CustomResourceDefinition{crd}, []v1alpha1.ClusterServiceVersion{csv})
+			_, err := fetchCatalogSourceOnStatus(crc, catalogSourceName, ns.GetName(), catalogSourceRegistryPodSynced)
+			Expect(err).ShouldNot(HaveOccurred())
+		})
+
+		AfterEach(func() {
+			TearDown(namespaceName)
```
timflannagan, line 1360:
```suggestion
			TeardownNamespace(namespaceName)
```

test/e2e/installplan_e2e_test.go
```diff
@@ -45,6 +45,7 @@ import (
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/bundle"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/operators/catalog"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/registry"
+
```
timflannagan, line 48: nit: newline.

test/e2e/installplan_e2e_test.go
```diff
@@ -4220,6 +4221,7 @@ var _ = Describe("Install Plan", func() {
 			return newPlan, ctx.Ctx().Client().Get(context.Background(), newKey, newPlan)
 		}).Should(HavePhase(operatorsv1alpha1.InstallPlanPhaseComplete))
 	})
+
 })
```
timflannagan, line 4225: nit: newline.

pkg/controller/errors/errors.go
```diff
@@ -47,6 +47,21 @@ func IsMultipleExistingCRDOwnersError(err error) bool {
...
+func NewRetryableError(err error) RetryableError {
+	return RetryableError{err}
+}
+func IsRetryable(err error) bool {
+	switch err.(type) {
+	case RetryableError:
+		return true
+	}
+	return false
+}
```
timflannagan, line 63: Should we investigate an [`errors.Is(...)]`(https://pkg.go.dev/errors#example-Is) implementation? I'm not sure we need this kind of wrapper logic since that method was introduced in Go 1.13.

pkg/controller/registry/resolver/steps.go
```diff
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"strings"
 
+	olmerrors "github.com/operator-framework/operator-lifecycle-manager/pkg/controller/errors"
```
timflannagan, line 9: nit: package import should live with other OLM/API imports that start on line 19.

test/e2e/catalog_e2e_test.go
```diff
@@ -1309,6 +1309,179 @@ var _ = Describe("Starting CatalogSource e2e tests", func() {
...
+				},
+			}
+
+			crd = newCRD(genName("ins"))
+			csv = newCSV(packageStable, ns.GetName(), "", semver.MustParse("0.1.0"), []apiextensions.CustomResourceDefinition{crd}, nil, nil)
+
+			catalogSourceName := genName("mock-ocs-")
+			cs, _ = createInternalCatalogSource(c, crc, catalogSourceName, ns.GetName(), manifests, []apiextensions.CustomResourceDefinition{crd}, []v1alpha1.ClusterServiceVersion{csv})
+			_, err := fetchCatalogSourceOnStatus(crc, catalogSourceName, ns.GetName(), catalogSourceRegistryPodSynced)
+			Expect(err).ShouldNot(HaveOccurred())
```
timflannagan, line 1356: @perdasilva Should we start introducing the magic catalog implementation here? We don't have to refactor everything to use that library, but we could enforce all net new e2e tests use that package to bootstrap catalogs for testing purposes.

timflannagan:
Just took a quick look through these changes. My main takeaways from that PR:

- I don't think we need to use a custom errors logic here and can just use errors.Is directly
- We should use the TeardownNamespace function as it dumps out test artifacts when the test case fails (vs. TearDown which is flake prone and doesn't have the extra logic)
- I think we can remove the comment on the unpackBundles and just punt to another PR that adds the godoc for that method
- It wasn't super clear to me whether we need to be calling SetupGeneratedTestNamespaceWithOperatorGroup inside the When containers as the top-level Describe creates them for us
- We should aim towards using this magic catalog library that Per introduced and I've been tinkering with recently for our testing catalog needs

test/e2e/catalog_e2e_test.go
nsapse: Will do.

test/e2e/catalog_e2e_test.go
nsapse: Turns out this is totally doable. Will delete redundant work.

test/e2e/catalog_e2e_test.go
```diff
@@ -1309,6 +1309,179 @@ var _ = Describe("Starting CatalogSource e2e tests", func() {
...
+					InstallPlanApproval:    v1alpha1.ApprovalAutomatic,
+				}
+
+				subscriptionName = genName("sub-")
+				createSubscriptionForCatalogWithSpec(GinkgoT(), crc, ns.GetName(), subscriptionName, subscriptionSpec)
+			})
+			It("succeeds in installing", func() {
+				sub, err = fetchSubscription(crc, ns.GetName(), subscriptionName, subscriptionHasInstallPlanChecker)
+				// plan, err = fetchInstallPlan(GinkgoT(), crc, installPlanName, )
+				fmt.Printf("Sub: %+v\n", sub)
```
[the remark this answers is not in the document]
nsapse: Embarrassing catch. Deleted my vestigial print statements.

test/e2e/catalog_e2e_test.go
perdasilva: I'm totally good with that! I think it makes things much easier and leaner on the code side.

pkg/controller/errors/errors.go
perdasilva: I think the problem is we don't really return any typed errors in the code. Just plain old Error. So, it becomes difficult to discern the retryability of errors. To use straight up `errors.Is(...)`, I think we'd need to create some sort of `CSVError` and add some `retryable` boolean to that. The nice thing about the wrapper is that it can wrap any error as a retryable.

nsapse: So leave this as is?

test/e2e/catalog_e2e_test.go
nsapse: When rewriting these tests with the magic catalog is the happy path (everything properly installs from a properly formed csv) need to re-written? I might be misunderstanding but it seems like that's already tested in the main magic-catalog tests [here](https://github.com/operator-framework/operator-lifecycle-manager/blob/604825027920cdc52a489669b22f51910cabfa69/test/e2e/magic_catalog_test.go#L29).

test/e2e/catalog_e2e_test.go
```diff
@@ -1340,6 +1343,61 @@ var _ = Describe("Starting CatalogSource e2e tests", func() {
...
+			Expect(err).To(BeNil())
+			fmt.Printf("Provider: %v\n", provider)
+
+			catalogSourceName = genName("cat-bad-csv")
+			magicCatalog = NewMagicCatalog(c, ns.GetName(), catalogSourceName, provider)
+			fmt.Printf("magicCatalog: %v\n", magicCatalog)
+			Expect(magicCatalog.DeployCatalog(context.Background())).To(BeNil())
+			fmt.Printf("namespace: %s\n\n", ns.GetName())
+			
+		})
```
timflannagan, line 1376: I mentioned this in slack but this looks like a potential scoping issue:

- We're building up and deploying a catalog that contains a bundle with a bad CSV in the test generated namespace
- Because we're wrapping this logic in an It spec vs. a BeforeEach container, when the `"A Subscription is created catalogSource built with the malformed CSV"` test case is run, the namespace that contains this catalog has already been deleted, and a new namespace has been created.

I think we only need to perform some light refactoring of this current setup to get everything working as intended:

> Note: I jotted this down inline so YMMV w.r.t compilation errors

```go
		BeforeEach(func() {
			c = ctx.Ctx().Client()

			provider, err := NewFileBasedFiledBasedCatalogProvider(filepath.Join(testdataDir, badCSVDir, "bad-csv.yaml"))
			Expect(err).To(BeNil())

			catalogSourceName = genName("cat-bad-csv")
			magicCatalog = NewMagicCatalog(c, ns.GetName(), catalogSourceName, provider)
			Expect(magicCatalog.DeployCatalog(context.Background())).To(BeNil())
		})

		When("A Subscription is created catalogSource built with the malformed CSV", func() {
			It("fails with a ResolutionFailed error condition, and a message that highlights the missing field in the CSV", func() {
				subscription = &operatorsv1alpha1.Subscription{
						Name:      fmt.Sprintf("%s-sub", catalogSourceName),
						Namespace: ns.GetName(),
					},
					Spec: &operatorsv1alpha1.SubscriptionSpec{
						CatalogSource:          catalogSourceName,
						CatalogSourceNamespace: ns.GetName(),
						Channel:                "stable",
						Package:                "packageA",
					},
				}
				Expect(c.Create(context.Background(), subscription)).To(BeNil())

				_, err := fetchSubscription(crc, subscription.GetNamespace(), subscription.GetName(), subscriptionHasInstallPlanChecker)
				Expect(err).Should(BeNil())
			})
		})
```

nsapse: That fixed it. I am forever in your debt.

pkg/controller/registry/resolver/steps.go
```diff
@@ -118,6 +119,15 @@ func NewStepResourceFromBundle(bundle *api.Bundle, namespace, replaces, catalogS
 		return nil, err
 	}
 
+	// Check unpacked bundled for for missing APIVersion or Kind
+	if csv.APIVersion == "" {
+		return nil, olmerrors.NewRetryableError(fmt.Errorf("bundle CSV %s missing APIVersion", csv.Name))
```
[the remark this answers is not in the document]
nsapse: Yeah, these errors should be terminal. The issue is that regular errors were doing just what you said - ending up in an infinite reconciliation loop. Per suggested the `retryableError` class to differentiate from a regular error but you're right. Would be make sense if tis error type were  called `fatalError` or something more descriptive?

test/e2e/catalog_e2e_test.go
```diff
@@ -1340,6 +1343,59 @@ var _ = Describe("Starting CatalogSource e2e tests", func() {
...
+
+		AfterEach(func() {
+			TeardownNamespace(ns.GetName())
+		})
+
+		When("A Subscription is created catalogSource built with the malformed CSV", func() {
+			It("fails with a ResolutionFailed error condition, and a message that highlights the missing field in the CSV", func() {
+
+				subscription = &operatorsv1alpha1.Subscription{
```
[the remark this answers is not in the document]
nsapse: will do

test/e2e/catalog_e2e_test.go
```diff
@@ -1340,6 +1343,59 @@ var _ = Describe("Starting CatalogSource e2e tests", func() {
...
+					},
+				}
+				Expect(c.Create(context.Background(), subscription)).To(BeNil())
+				subscription, err := fetchSubscription(crc, subscription.GetNamespace(), subscription.GetName(), subscriptionHasInstallPlanChecker)
+				Expect(err).Should(BeNil())
+				installPlanName := subscription.Status.Install.Name
+				installPlan, err := fetchInstallPlan(GinkgoT(), crc, installPlanName, subscription.GetNamespace(), buildInstallPlanPhaseCheckFunc(operatorsv1alpha1.InstallPlanPhaseFailed))
+
+				// expect the message that API missing
+				Expect(installPlan.Status.Conditions[0].Message).To(ContainSubstring("missing APIVersion"))
```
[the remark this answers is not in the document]
nsapse: fair point - the status should propagate up to the subscription. I'll try and catch it there

test/e2e/testdata/bad-csv/fbc_catalog.json
```diff
@@ -0,0 +1,37 @@
...
+    },
+    {
+      "type": "olm.package",
+      "value": {
+        "packageName": "cockroach",
+        "version": "1.0.0"
+      }
+    }
+  ]
+}
```
timflannagan, line 37: Any idea on whether we need this FBC manifest, or can we drop it from the set of changes? I didn't see it used anywhere going through these changes.

test/e2e/testdata/bad-csv/example-operator.v0.1.0.yaml
```diff
@@ -0,0 +1,25 @@
...
+properties:
+  - type: olm.gvk
+    value:
+      group: example.com
+      kind: TestA
+      version: v1alpha1
+  - type: olm.package
+    value:
+      packageName: packageA
+      version: 1.0.0
```
timflannagan, line 25: Same comment as [below](https://github.com/operator-framework/operator-lifecycle-manager/pull/2673/files#r861210613) - are we using this manifest anywhere, or can we drop it from the set of changes?

test/e2e/catalog_e2e_test.go
```diff
@@ -1340,6 +1343,59 @@ var _ = Describe("Starting CatalogSource e2e tests", func() {
...
+						CatalogSourceNamespace: ns.GetName(),
+						Channel:                "stable",
+						Package:                "packageA",
+					},
+				}
+				Expect(c.Create(context.Background(), subscription)).To(BeNil())
+				subscription, err := fetchSubscription(crc, subscription.GetNamespace(), subscription.GetName(), subscriptionHasInstallPlanChecker)
+				Expect(err).Should(BeNil())
+				installPlanName := subscription.Status.Install.Name
+				installPlan, err := fetchInstallPlan(GinkgoT(), crc, installPlanName, subscription.GetNamespace(), buildInstallPlanPhaseCheckFunc(operatorsv1alpha1.InstallPlanPhaseFailed))
```
timflannagan, line 1391: Do we need to check whether the err here is nil?

test/e2e/testdata/bad-csv/fbc_catalog.json
nsapse: This is vestigial and I thought I had removed it. Thanks for catching that.

test/e2e/testdata/bad-csv/example-operator.v0.1.0.yaml
nsapse: same comment as below.

test/e2e/catalog_e2e_test.go
nsapse: Not sure? I was just following what looked like a general pattern of checking that things succeeded (i.e. didn't fail) before moving forward. Can remove it if that's better practice.

test/e2e/installplan_e2e_test.go
nsapse: fixed

test/e2e/catalog_e2e_test.go
nsapse: Added.

pkg/controller/registry/resolver/steps.go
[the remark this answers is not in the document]
nsapse: @exdx just following up - how does that proposed solution sound to you?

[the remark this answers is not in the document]
exdx: I think switching the implementation around and instead calling it a `FatalError` makes sense, that way you can use the new error type in the reconciliation loop. We don't requeue on a `FatalError` which is logical. 

The alternative using this existing implementation would be to change this error to be a plain error, and then leave the retriable error as a new type that's not being used. I think changing the implementation to fatal makes more sense.

pkg/controller/operators/catalog/operator.go
```diff
@@ -1439,6 +1445,10 @@ func (o *Operator) unpackBundles(plan *v1alpha1.InstallPlan) (bool, *v1alpha1.In
 		// Ensure that bundle can be applied by the current version of OLM by converting to steps
 		steps, err := resolver.NewStepsFromBundle(res.Bundle(), out.GetNamespace(), res.Replaces, res.CatalogSourceRef.Name, res.CatalogSourceRef.Namespace)
 		if err != nil {
+			if retryable := olmerrors.IsFatal(err); !retryable {
```
exdx, line 1448: nit: should this be 
```
if fatal := olmerrors.IsFatal(err); fatal {
  return false, nil, err
}
```

pkg/controller/operators/catalog/operator.go
```diff
@@ -1627,6 +1637,14 @@ func (o *Operator) syncInstallPlans(obj interface{}) (syncError error) {
 	if len(plan.Status.BundleLookups) > 0 {
 		unpacked, out, err := o.unpackBundles(plan)
 		if err != nil {
+			// If the error was fatal capture and fail
+			if retryable := olmerrors.IsFatal(err); !retryable {
```
exdx, line 1641: same here, I think we want
```
if fatal := olmerrors.IsFatal(err); fatal {}
```

nsapse: Oof, absolutely, yeah. Completely my bad. I just searched and replaced for instances of the actual error. Somehow didn't think about references pointing to it. Friday brain.

pkg/controller/operators/catalog/operator.go
```diff
@@ -1627,6 +1637,14 @@ func (o *Operator) syncInstallPlans(obj interface{}) (syncError error) {
 	if len(plan.Status.BundleLookups) > 0 {
 		unpacked, out, err := o.unpackBundles(plan)
 		if err != nil {
+			// If the error was fatal capture and fail
+			if fatal := olmerrors.IsFatal(err); !fatal {
```
exdx, line 1641:
```suggestion
			if fatal := olmerrors.IsFatal(err); fatal {
```
Since in this case, if the error is fatal, we want to transition the installplan to failed

exdx (commented):
small nit: the commit message should be updated to reflect the new fatal error type.

After that, this LGTM

pkg/controller/operators/catalog/operator.go
```diff
@@ -1439,6 +1445,10 @@ func (o *Operator) unpackBundles(plan *v1alpha1.InstallPlan) (bool, *v1alpha1.In
 		// Ensure that bundle can be applied by the current version of OLM by converting to steps
 		steps, err := resolver.NewStepsFromBundle(res.Bundle(), out.GetNamespace(), res.Replaces, res.CatalogSourceRef.Name, res.CatalogSourceRef.Namespace)
 		if err != nil {
+			if fatal := olmerrors.IsFatal(err); !fatal {
```
exdx, line 1448:
```suggestion
			if fatal := olmerrors.IsFatal(err); fatal {
```
Looking at this again, I think here we need to look for the fatal error here and return early as well

nsapse:
Realizing that the test currently references a quay repo in my own repo rather than the `olmtest` repo which I don't have write access for. That seems like a bad idea from a link rot perspective. What's the most efficient way for me to push my image to the `olmtest` repo?

exdx (approved):

PR #2675, opened 2022-02-24
Remove ConstraintProvider from the exported solver package API.

benluddy:
This interface is a resolver package concern, so its definition can
live beside the code that accepts it. Removing it from solver also
removes the only solver->cache package dependency.

As I understand it, the only way to inject an implementation of this
interface today is by having access to unexported fields of
resolver.Resolver, so for now I've also made the interface
unexported. In the context of the ongoing desire to support resolution
as a separate library module, this will allow more time to evaluate
the shape of a stable library API.


pkg/controller/registry/resolver/resolver_test.go
```diff
@@ -80,6 +80,13 @@ func TestSolveOperators(t *testing.T) {
 	require.ElementsMatch(t, expected, operators)
 }
 
+// ConstraintProviderFunc is a simple implementation of ConstraintProvider
+type ConstraintProviderFunc func(e *cache.Entry) ([]solver.Constraint, error)
+
+func (c ConstraintProviderFunc) Constraints(e *cache.Entry) ([]solver.Constraint, error) {
+	return c(e)
+}
```
njhale, line 88: Should this be unexported as well?

njhale: nvm, didn't notice this was in a test; carry on.

PR #2679, opened 2022-03-02
fix(operatorconditions): don't retry on 404

### LINKED ISSUE #2678: OperatorCondition/Generator controllers churn on missing CSVs
## Bug Report

The controllers managing OperatorConditions can churn indefinitely, periodically producing error logs, when a CSV is deleted.

**What did you do?**

1. `kind create cluster`
2. `make run-local`
3. `kubectl -n olm edit deployment packageserver`, add a finalizer (example: `foo.io/bar`), save
4. `kubectl -n olm delete csv packageserver`

A single log entry -- if any -- from the OperatorConditions controllers indicating that the `packageserver` CSV is missing.

Periodic log entries on the subject:

```

```

**Environment**
* operator-lifecycle-manager version: `5cf7f1f4e`

```
Client Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.1", GitCommit:"5e58841cce77d4bc13713ad2b91fa0d961e69192", GitTreeState:"clean", BuildDate:"2021-05-12T14:11:29Z", GoVersion:"go1.16.3", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.1", GitCommit:"86ec240af8cbd1b60bcc4c03c20da9b98005b92e", GitTreeState:"clean", BuildDate:"2022-01-06T22:13:50Z", GoVersion:"go1.17.5", Compiler:"gc", Platform:"linux/amd64"}
```

* Kubernetes cluster kind: kind

**Possible Solution**

Don't retry OperatorCondition reconciliation when expected CSVs are not found; explicitly, when the response for a GET request is 404 NOT_FOUND.

njhale:
The controllers managing OperatorConditions can churn indefinitely, periodically producing error logs, when a CSV is deleted. To prevent this, don't retry when presented with select NOT_FOUND errors during reconciliation.

To verify this works, I followed the reproduction steps in #2678 both before and after this patch is applied. I then confirmed occurrences of `Unable to find` in the logs of the olm-operator pod were significantly reduced.

*Before*

```sh
$ kubectl -n olm logs olm-operator-785d845d6b-5plbq | rg 'Unable to find' | wc -l
31
```

This number seems to increase over time.

*After*

```sh
$ kubectl -n olm logs olm-operator-785d845d6b-vzsgm | rg 'Unable to find' | wc -l
 8
```

This number seems to be the steady state.

Note: There's probably a better way to test this fix (for example, compare queue prometheus metrics or pprof cpu profiles)

fixes #2678

### DIFF
--- a/pkg/controller/operators/operatorcondition_controller.go
+++ b/pkg/controller/operators/operatorcondition_controller.go
@@ -93,39 +93,34 @@ var _ reconcile.Reconciler = &OperatorConditionReconciler{}
 
 func (r *OperatorConditionReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	// Set up a convenient log object so we don't have to type request over and over again
-	log := r.log.WithValues("request", req)
-	log.V(2).Info("reconciling operatorcondition")
+	log := r.log.WithValues("request", req).V(1)
+	log.Info("reconciling")
 	metrics.EmitOperatorConditionReconcile(req.Namespace, req.Name)
 
 	operatorCondition := &operatorsv2.OperatorCondition{}
-	err := r.Client.Get(context.TODO(), req.NamespacedName, operatorCondition)
-	if err != nil {
-		log.V(1).Error(err, "Unable to find operatorcondition")
+	if err := r.Client.Get(ctx, req.NamespacedName, operatorCondition); err != nil {
+		log.Info("Unable to find OperatorCondition")
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+
+	if err := r.ensureOperatorConditionRole(operatorCondition); err != nil {
+		log.Info("Error ensuring OperatorCondition Role")
 		return ctrl.Result{}, err
 	}
 
-	err = r.ensureOperatorConditionRole(operatorCondition)
-	if err != nil {
-		log.V(1).Error(err, "Error ensuring OperatorCondition Role")
-		return ctrl.Result{Requeue: true}, err
+	if err := r.ensureOperatorConditionRoleBinding(operatorCondition); err != nil {
+		log.Info("Error ensuring OperatorCondition RoleBinding")
+		return ctrl.Result{}, err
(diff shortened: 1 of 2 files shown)


pkg/controller/operators/operatorcondition_controller.go
```diff
@@ -93,39 +93,34 @@ var _ reconcile.Reconciler = &OperatorConditionReconciler{}
 
 func (r *OperatorConditionReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	// Set up a convenient log object so we don't have to type request over and over again
-	log := r.log.WithValues("request", req)
-	log.V(2).Info("reconciling operatorcondition")
+	log := r.log.WithValues("request", req).V(1)
```
njhale, line 96: `V(1)` is debug level for the underlying logr implementation, `zapr`.

perdasilva: lol - of course it is. Why would you call it...`Debug`...?

joelanford: `logr` doesn't have named levels, really. It has error logs and info logs. info logs can have varying degrees of importance, so the number reflects the verbosity. Higher number = higher verbosity.

However logr is just an interface. There's an implementation of this interface for all the major logging libraries, so if the underlying implementation has named levels, they end up being mapped to a verbosity number in logr.

joelanford: Also... in logr, the verbosity is additive. So `log.V(1).V(1)` is the same as `log.V(2)`, It's somewhat misleading to say `log.V(1)` is equivalent to `Debug` since it depends on what the base verbosity of `log` is.

njhale: Yeah, I'm not sure I'm a huge fan of the `logr` api from a user perspective.

PR #2707, opened 2022-03-24
Move deprecated e2e to one namespace per spec

perdasilva:

This PR moves the deprecated e2e tests to one namespace per spec

e2e stability

**Reviewer Checklist**

### DIFF
--- a/test/e2e/deprecated_e2e_test.go
+++ b/test/e2e/deprecated_e2e_test.go
@@ -2,89 +2,92 @@ package e2e
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	"github.com/blang/semver/v4"
 	. "github.com/onsi/ginkgo"
-	"github.com/onsi/ginkgo/extensions/table"
 	. "github.com/onsi/gomega"
+	operatorsv1 "github.com/operator-framework/api/pkg/operators/v1"
 	operatorsv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
+	"github.com/operator-framework/operator-lifecycle-manager/test/e2e/ctx"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-
-	"github.com/operator-framework/operator-lifecycle-manager/test/e2e/ctx"
 )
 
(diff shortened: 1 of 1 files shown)


test/e2e/deprecated_e2e_test.go
```diff
@@ -18,12 +21,25 @@ import (
 var missingAPI = `{"apiVersion":"verticalpodautoscalers.autoscaling.k8s.io/v1","kind":"VerticalPodAutoscaler","metadata":{"name":"my.thing","namespace":"foo"}}`
 
 var _ = Describe("Not found APIs", func() {
+
+	var ns corev1.Namespace
+
 	BeforeEach(func() {
-		csv := newCSV("test-csv", testNamespace, "", semver.Version{}, nil, nil, nil)
+		namespaceName := genName("catsrc-e2e-")
```
akihikokuroda, line 28: nit: strange prefix.  It may be better some string that can be used to identify this test case like "dep-e2e"

timflannagan: Yup +1

test/e2e/deprecated_e2e_test.go
```diff
@@ -39,8 +55,8 @@ var _ = Describe("Not found APIs", func() {
 				name: "installplan contains a missing API",
 				IP: &operatorsv1alpha1.InstallPlan{
 					ObjectMeta: metav1.ObjectMeta{
-						Namespace: *namespace, // this is necessary due to ginkgo table semantics, see https://github.com/onsi/ginkgo/issues/378
-						Name:      "test-plan-api",
+						// Namespace: ns.GetName(), // this is necessary due to ginkgo table semantics, see https://github.com/onsi/ginkgo/issues/378 - setting namespace in describe table
```
timflannagan, line 58: side note: it would be nice to refactor this test spec to avoid using a table-driven implementation to make it consistent with the rest of the e2e suite.

akihikokuroda, line 58: There is just one table entry.  It may better not use the tableEntries to avoid this.

test/e2e/deprecated_e2e_test.go
timflannagan: Looks like this was addressed - resolving the conversation.

perdasilva:
@akihikokuroda @timflannagan I've addressed the comments =D

test/e2e/deprecated_e2e_test.go
perdasilva: copy/pasta bug =D great catch! thank you =D

PR #2432, opened 2021-11-02
.github,test: Upload container logs during e2e runs as artifacts

timflannagan:
**Description of the change:**
Upload the container logs from the various kind clusters that were provisioned during individual e2e runs as artifacts using the upload-artifact action.

**Motivation for the change:**
Currently, there's little visibility into why individual e2e runs failed. A quick, short-term win is to collect the individual container logs that were running on the provisioned kind cluster to the configured artifacts directory for GHA.

**Reviewer Checklist**
- [x] Implementation matches the proposed design, or proposal is updated to match implementation
- [x] Sufficient unit test coverage
- [x] Sufficient end-to-end test coverage
- [x] Docs updated or added to `/doc`
- [x] Commit messages sensible and descriptive

PR #2721, opened 2022-03-29
Ensure the $ARTIFACTS_DIR variable is respected

timflannagan:
#2432 

**Reviewer Checklist**


test/e2e/ctx/ctx.go
```diff
@@ -116,8 +117,7 @@ func (ctx TestContext) DumpNamespaceArtifacts(namespace string) error {
 		"KUBECONFIG=" + kubeconfigPath,
 	}
 
-	// compiled test binary running e2e tests is run from the root ./bin directory
-	cmd := exec.Command("../test/e2e/collect-ci-artifacts.sh")
```
timflannagan, line 120: I ran into a sourcing issue when running the `make e2e` command locally as it's working directory points to ./test/e2e, so there's some skew between the working directory when using the kind or kubeconfig provisioners.

test/e2e/e2e_test.go
```diff
@@ -79,6 +85,9 @@ var _ = BeforeSuite(func() {
 		// This flag can be deprecated in favor of the kubeconfig provisioner:
 		os.Setenv("KUBECONFIG", *kubeConfigPath)
 	}
+	if collectArtifactsScriptPath != nil && *collectArtifactsScriptPath != "" {
+		os.Setenv("E2E_ARTIFACTS_SCRIPT", *collectArtifactsScriptPath)
+	}
```
timflannagan, line 90: Note: there wasn't any clear way of injecting global state to the ctx package, and there weren't any setter methods so I wanted to avoid introducing that pattern, so we're using an internal environment variable to basically act as a cross-package variable. This is admittedly a bit hacky of an implementation so I'm open to suggestions.

test/e2e/ctx/ctx.go
```diff
@@ -137,6 +137,17 @@ func setDerivedFields(ctx *TestContext) error {
 		return fmt.Errorf("nil RESTClient")
 	}
 
+	if ctx.artifactsDir == "" {
```
perdasilva, line 140: Could this poorly interact with some downstream only code: https://github.com/openshift/operator-framework-olm/blob/master/staging/operator-lifecycle-manager/test/e2e/e2e_test.go#L69?

timflannagan: We should be in the clear there as we're building up the junit reporter for ginkgo vs. telling either of the provisioner implementation where to dump testing artifacts when configured.

timflannagan: Also: we could bubble up that downstream-only change to make upstream backwards-compatible to eliminate any skew between implementations (which _might_ be relevant for any effort around ginkgo v1 -> v2)?

timflannagan:
Ran these changes locally after rebasing against master:

```console
$ make e2e ARTIFACT_DIR=./tmp E2E_INSTALL_NS=olm TEST="Not found APIs"
...
$ tree test/e2e/tmp
test/e2e/tmp
├── deprecated-e2e-pvt9s
│   ├── get_catalogsources_-o_yaml
│   ├── get_clusterserviceversions_-o_yaml
│   ├── get_events_--sort-by_.lastTimestamp
│   ├── get_installplans_-o_yaml
│   ├── get_operatorgroups_-o_yaml
│   ├── get_pods_-o_wide
│   └── get_subscriptions_-o_yaml
└── junit
    └── junit_e2e_01.xml
```

And I had updated one of the test cases in the deprecated e2e suite that's being focused on in that example to purposely fail:

```diff
-                               Expect(ctx.Ctx().Client().Status().Update(context.Background(), ip)).To(Succeed(), "failed to update the resource")
+                               Expect(ctx.Ctx().Client().Status().Update(context.Background(), ip)).ToNot(Succeed(), "failed to update the resource")
```

test/e2e/e2e_test.go
```diff
@@ -79,6 +85,9 @@ var _ = BeforeSuite(func() {
 		// This flag can be deprecated in favor of the kubeconfig provisioner:
 		os.Setenv("KUBECONFIG", *kubeConfigPath)
 	}
+	if collectArtifactsScriptPath != nil && *collectArtifactsScriptPath != "" {
+		os.Setenv("E2E_ARTIFACT_SCRIPT", *collectArtifactsScriptPath)
+	}
```
timflannagan, line 91: Note: just a hack to inject state so the ctx package can derive where to find the artifacts bash script path location.

dinhxuanvu:
Can we correct the title, commit message and all references to `ARTIFACT_DIR` given we just merged a PR to correct that?

dinhxuanvu (commented):
/lgtm

PR #2715, opened 2022-03-26
Add importas linting configuration to enforce common package import naming

timflannagan:
**Description of the change:**
Introduce a configuration for the `importas` configuration for commonly used package imports, e.g. metav1/corev1/appsv1, when running golangci-lint runner. This is currently used in the operator-framework/rukpak repository, and helps lessen reviewer burden as we're focusing less on styling differences.

**Motivation for the change:**
Enforce package import naming for commonly imported packages to lessen the reviewer burden.

**Reviewer Checklist**
- [ ] Implementation matches the proposed design, or proposal is updated to match implementation
- [ ] Sufficient unit test coverage
- [ ] Sufficient end-to-end test coverage
- [ ] Docs updated or added to `/doc`
- [ ] Commit messages sensible and descriptive
- [ ] Tests marked as `[FLAKE]` are truly flaky
- [ ] Tests that remove the `[FLAKE]` tag are no longer flaky

PR #2728, opened 2022-04-06
Introduce support for Fail Forward Upgrades

awgreene:
Introduce support for Fail Forward Upgrades

**Description:**
This commit introduces OLM support for fail forward upgrades. Existing upgrade behavior is now defined as "Default" upgrade behavior. A new upgrade strategy referred to as UnsafeFailForward has been introduced in a tech preview state. When fail forward upgrades are enabled, CSVs will now be able to move from the failed state to the replacing state, effectively allowing operators in a namespace to fail forward from failed installs or upgrades. Additionally, it will be possible to recover from failed installPlans by updating the upgrade graph defined in the catalog. Fail forward upgrades are generally unsafe and  unsupported as they could result in unexpected behavior or lost data due to the fact that critical version of an operator may not be visited.

**Enhancement Proposal:** https://github.com/operator-framework/enhancements/blob/87e6ac32a0caac86fdfc733740892da612d00ee2/enhancements/fail-forward-upgrades.md
**JIRA Ticket:** https://issues.redhat.com/browse/OLM-2475


pkg/controller/operators/catalog/operator.go
```diff
@@ -1105,6 +1117,10 @@ func (o *Operator) ensureSubscriptionInstallPlanState(logger *logrus.Entry, sub
 	out.Status.InstallPlanRef = ref
 	out.Status.Install = v1alpha1.NewInstallPlanReference(ref)
 	out.Status.State = v1alpha1.SubscriptionStateUpgradePending
+	// TODO(fail-forward): Decide if this should be default behavior.
```
awgreene, line 1120: Thoughts?

pkg/controller/operators/catalog/operator.go
```diff
@@ -1105,6 +1117,10 @@ func (o *Operator) ensureSubscriptionInstallPlanState(logger *logrus.Entry, sub
...
+	if failForwardEnabled && ip.Status.Phase == v1alpha1.InstallPlanPhaseFailed {
```
awgreene, line 1121: Thoughts?
```suggestion
	// TODO(fail-forward): Decide if this should be default behavior.
	if ip.Status.Phase == v1alpha1.InstallPlanPhaseFailed {
```

pkg/controller/registry/resolver/step_resolver.go
```diff
@@ -84,14 +86,54 @@ func NewOperatorStepResolver(lister operatorlister.OperatorLister, client versio
...
 
+// replacementChainFailed returns true if the latest entry in a chain of operators is in the failed state.
+func replacementChainFailed(csv *v1alpha1.ClusterServiceVersion, csvToReplacement map[string]*v1alpha1.ClusterServiceVersion) bool {
+	if csv == nil {
+		return false
+	}
+	if csv.Status.Phase == v1alpha1.CSVPhaseFailed {
+		return true
+	}
+	return replacementChainFailed(csvToReplacement[csv.GetName()], csvToReplacement)
```
njhale, line 97: The last CSV in a replacement chain shouldn't have a replacement, so doesn't the base-case `csv == nil` always occur, even if the csv isn't failed (order of base-case issue)?

pkg/controller/registry/resolver/step_resolver.go
```diff
@@ -84,14 +86,54 @@ func NewOperatorStepResolver(lister operatorlister.OperatorLister, client versio
...
+		}
+
+		for i := range csvs {
+			if !csvs[i].IsCopied() {
+				if csvs[i].Status.Phase == v1alpha1.CSVPhaseReplacing && replacementChainFailed(csvs[i], replacementMapping) {
+					omittedCSVs = append(omittedCSVs, cache.Not(cache.CSVNamePredicate(csvs[i].GetName())))
+				}
+			}
+		}
+	}
```
njhale, line 134: This can be encapsulated in a function to clean things up a bit.

pkg/controller/registry/resolver/step_resolver.go
```diff
@@ -84,14 +86,54 @@ func NewOperatorStepResolver(lister operatorlister.OperatorLister, client versio
...
+	return replacementChainFailed(csvToReplacement[csv.GetName()], csvToReplacement)
+}
+
 func (r *OperatorStepResolver) ResolveSteps(namespace string) ([]*v1alpha1.Step, []v1alpha1.BundleLookup, []*v1alpha1.Subscription, error) {
 	subs, err := r.listSubscriptions(namespace)
 	if err != nil {
 		return nil, nil, nil, err
 	}
 
+	var omittedCSVs []cache.Predicate
```
njhale, line 106: A little comment explaining what the intent of these predicates are would be helpful when reading through in the future.

njhale (dismissed):

pkg/controller/operators/catalog/operator.go
```diff
@@ -42,6 +42,7 @@ import (
 	"k8s.io/client-go/util/workqueue"
 
 	"github.com/operator-framework/api/pkg/operators/reference"
+	v1 "github.com/operator-framework/api/pkg/operators/v1"
```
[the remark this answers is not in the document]
awgreene: That's fine with me.

pkg/controller/registry/resolver/step_resolver.go
```diff
@@ -84,14 +86,54 @@ func NewOperatorStepResolver(lister operatorlister.OperatorLister, client versio
 	return stepResolver
 }
 
+// replacementChainFailed returns true if the latest entry in a chain of operators is in the failed state.
```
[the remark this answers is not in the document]
awgreene: Great point, will fix shortly.

pkg/controller/operators/catalog/operator.go
```diff
@@ -1121,6 +1137,12 @@ func (o *Operator) ensureSubscriptionCSVState(logger *logrus.Entry, sub *v1alpha
 	if err != nil {
 		logger.WithError(err).WithField("currentCSV", sub.Status.CurrentCSV).Debug("error fetching csv listed in subscription status")
 		out.Status.State = v1alpha1.SubscriptionStateUpgradePending
+		ip, err := o.client.OperatorsV1alpha1().InstallPlans(sub.GetNamespace()).Get(context.TODO(), sub.Status.InstallPlanRef.Name, metav1.GetOptions{})
+		if err != nil {
+			logger.WithError(err).WithField("currentCSV", sub.Status.CurrentCSV).Debug("error fetching installplan listed in subscription status")
+		} else if ip.Status.Phase == v1alpha1.InstallPlanPhaseFailed {
+			out.Status.State = v1alpha1.SubscriptionStateFailed
+		}
```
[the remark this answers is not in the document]
awgreene: This covers the case where a CSV was not applied and the current installPlan failed. Adding a fail-forward flag.

pkg/controller/operators/catalog/operator.go
```diff
@@ -903,6 +904,17 @@ func (o *Operator) syncResolvingNamespace(obj interface{}) error {
 		return err
 	}
 
+	failForwardEnabled := false
+	ogs, err := o.lister.OperatorsV1().OperatorGroupLister().OperatorGroups(namespace).List(labels.Everything())
```
perdasilva, line 908: suggestion: might be worth putting this in a function just to lean this one out. Not a blocker tho.

pkg/controller/operators/catalog/operator.go
```diff
@@ -903,6 +904,17 @@ func (o *Operator) syncResolvingNamespace(obj interface{}) error {
...
+	if err != nil {
+		logger.WithError(err).Debug("couldn't list operatorGroups, assuming default upgradeStrategy")
+	}
+	if len(ogs) != 1 {
+		logger.WithError(err).Debugf("Found %d operatorGroups, expected 1, assuming default upgradeStrategy", len(ogs))
+	} else {
+		failForwardEnabled = ogs[0].UpgradeStrategy() == operatorsv1.UnsafeFailForwardUpgradeStrategy
+	}
+
 	// TODO: parallel
```
perdasilva, line 918: is this TODO still a thing? XD

pkg/controller/operators/olm/operator.go
```diff
@@ -1996,6 +1996,16 @@ func (a *Operator) transitionCSVState(in v1alpha1.ClusterServiceVersion) (out *v
 		}
 
 	case v1alpha1.CSVPhaseFailed:
+		// TODO(fail-forward): If the csv allows fail forward
+		// Transition to replacing if FailForward is enabled, a CSV exists that replaces the operator, and the operator being replaced supports failing forward.
```
perdasilva, line 2000:
```suggestion
		// Transition to replacing if FailForward is enabled and a CSV exists that replaces the operator
```

perdasilva: This is no longer a requirement right?

pkg/controller/operators/olm/operator.go
```diff
@@ -2087,7 +2098,15 @@ func (a *Operator) transitionCSVState(in v1alpha1.ClusterServiceVersion) (out *v
 		}
 
 		// If there is a succeeded replacement, mark this for deletion
-		if next := a.isBeingReplaced(out, a.csvSet(out.GetNamespace(), v1alpha1.CSVPhaseAny)); next != nil {
+		next := a.isBeingReplaced(out, a.csvSet(out.GetNamespace(), v1alpha1.CSVPhaseAny))
+		if operatorGroup.UpgradeStrategy() == operatorsv1.UnsafeFailForwardUpgradeStrategy {
```
perdasilva, line 2102: Might be worth segregating this to its own method `IsFailForwardEnabled` or some such. Just to make it easy to refactor later in case something changes =P

awgreene:
Gonna hold this until unit/e2e tests are added.
/hold

pkg/controller/operators/catalog/operator.go
```diff
@@ -1105,13 +1117,16 @@ func (o *Operator) ensureSubscriptionInstallPlanState(logger *logrus.Entry, sub
 	out.Status.InstallPlanRef = ref
 	out.Status.Install = v1alpha1.NewInstallPlanReference(ref)
 	out.Status.State = v1alpha1.SubscriptionStateUpgradePending
+	if failForwardEnabled && ip.Status.Phase == v1alpha1.InstallPlanPhaseFailed {
+		out.Status.State = v1alpha1.SubscriptionStateFailed
```
dinhxuanvu, line 1121: I'm slightly concerned that `UpgradeFailed` may be misleading given the initial installation is not really an upgrade. However, given there is no other status, this is the only choice.

dinhxuanvu: On another note, the subscription status doesn't really matter given the CSV's status is what we care about.

pkg/controller/operators/olm/operator.go
```diff
@@ -1996,6 +1996,16 @@ func (a *Operator) transitionCSVState(in v1alpha1.ClusterServiceVersion) (out *v
 		}
 
 	case v1alpha1.CSVPhaseFailed:
+		// TODO(fail-forward): If the csv allows fail forward
+		// Transition to replacing if FailForward is enabled, a CSV exists that replaces the operator, and the operator being replaced supports failing forward.
+		if operatorGroup.UpgradeStrategy() == operatorsv1.UnsafeFailForwardUpgradeStrategy {
+			if replacement := a.isBeingReplaced(out, a.csvSet(out.GetNamespace(), v1alpha1.CSVPhaseAny)); replacement != nil {
+				msg := fmt.Sprintf("%s csv being replaced by csv: %s", out.Status.Phase, replacement.GetName())
+				out.SetPhaseWithEvent(v1alpha1.CSVPhaseReplacing, v1alpha1.CSVReasonBeingReplaced, msg, a.now(), a.recorder)
```
dinhxuanvu, line 2004: Given we set an event here, I would like to see the msg to reflect this is unsafeupgrade mode, not just standard transition phase. WDYT?

pkg/controller/operators/olm/operator.go
```diff
@@ -2087,7 +2097,15 @@ func (a *Operator) transitionCSVState(in v1alpha1.ClusterServiceVersion) (out *v
 		}
 
 		// If there is a succeeded replacement, mark this for deletion
-		if next := a.isBeingReplaced(out, a.csvSet(out.GetNamespace(), v1alpha1.CSVPhaseAny)); next != nil {
+		next := a.isBeingReplaced(out, a.csvSet(out.GetNamespace(), v1alpha1.CSVPhaseAny))
```
dinhxuanvu, line 2100: This block of code is to determine the oldest CSV in the CSV replacing chain so it can be deleted first. Please put a comment here to reflect that. It helps with the review process and future changes.

pkg/controller/registry/resolver/step_resolver.go
```diff
@@ -84,14 +86,83 @@ func NewOperatorStepResolver(lister operatorlister.OperatorLister, client versio
...
+		// Move iterator along replacement chain
+		itr = next
+	}
+	return itr.Status.Phase == v1alpha1.CSVPhaseFailed
+}
+
+func (r *OperatorStepResolver) CachePredicates(namespace string) ([]cache.Predicate, error) {
+	predicates := []cache.Predicate{}
+
+	operatorGroupList, err := r.client.OperatorsV1().OperatorGroups(namespace).List(context.TODO(), metav1.ListOptions{})
```
dinhxuanvu, line 120: Looking at all of these OG's query in here and other places, it makes me wonder if `UpgradeStrategy` should be in Subscription instead of OG given resolver is working with Subscription so we don't need to query OG at all. Too late for it now so it's just me stating my thought here.

dinhxuanvu: May as well add a OGlister to OperatorStepResolver for the sake of efficiency.

pkg/controller/registry/resolver/resolver.go
```diff
@@ -55,7 +55,7 @@ func (w *debugWriter) Write(b []byte) (int, error) {
 	return n, nil
 }
 
-func (r *Resolver) Resolve(namespaces []string, subs []*v1alpha1.Subscription) ([]*cache.Entry, error) {
+func (r *Resolver) Resolve(namespaces []string, subs []*v1alpha1.Subscription, existingEntryPredicates ...cache.Predicate) ([]*cache.Entry, error) {
```
dinhxuanvu, line 58: As we talked about this in the meeting already but I wanted to point it out here for another folks to be aware of: this change is basically an undo of what Ben Luddy did. Not a big deal but something to be aware of.

pkg/controller/registry/resolver/step_resolver.go
```diff
@@ -84,14 +86,83 @@ func NewOperatorStepResolver(lister operatorlister.OperatorLister, client versio
...
+		for _, csv := range csvs {
+			if csv.Spec.Replaces != "" {
+				replacementMapping[csv.Spec.Replaces] = csv
+			}
+		}
+
+		for i := range csvs {
+			if !csvs[i].IsCopied() {
+				if csvs[i].Status.Phase == v1alpha1.CSVPhaseReplacing && replacementChainEndsInFailure(csvs[i], replacementMapping) {
+					predicates = append(predicates, cache.Not(cache.CSVNamePredicate(csvs[i].GetName())))
```
dinhxuanvu, line 141: I see what you are trying to do here but I'm a bit iffy with using `Not` predicate given it bit us in the butt once with Eric's work. Frankly, I wonder if we simply exclude all `Replacing` CSV from resolution. Would that be better? I personally want to have Ben's opinion here but he is unavailable.
I think it will be OK with this solution if we have a solid e2e test case to ensure this will work.

dinhxuanvu (commented):
Reviewed with comments.

pkg/controller/operators/olm/operator.go
awgreene: That sounds reasonable to me!

pkg/controller/operators/catalog/operator.go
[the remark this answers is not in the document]
timflannagan: I'm somewhat surprised the linting check didn't catch this locally. I had added a goimportas linting configuration in #2715 that enforces the operatorsv1 package import. Maybe we're filtering out this package path in the golangci-lint configuration?

pkg/controller/operators/olm/operator.go
timflannagan: This is what we do with the downstream package server manager component that's solely responsible for reading the topology mode from this cluster-wide API and returning a bool for whether single node functionality should be enabled. Seems relevant here but not blocking imo.

pkg/controller/operators/catalog/operator.go
```diff
@@ -1121,6 +1136,14 @@ func (o *Operator) ensureSubscriptionCSVState(logger *logrus.Entry, sub *v1alpha
 	if err != nil {
 		logger.WithError(err).WithField("currentCSV", sub.Status.CurrentCSV).Debug("error fetching csv listed in subscription status")
 		out.Status.State = v1alpha1.SubscriptionStateUpgradePending
+		if failForwardEnabled {
+			ip, err := o.client.OperatorsV1alpha1().InstallPlans(sub.GetNamespace()).Get(context.TODO(), sub.Status.InstallPlanRef.Name, metav1.GetOptions{})
+			if err != nil {
+				logger.WithError(err).WithField("currentCSV", sub.Status.CurrentCSV).Debug("error fetching installplan listed in subscription status")
+			} else if ip.Status.Phase == v1alpha1.InstallPlanPhaseFailed {
+				out.Status.State = v1alpha1.SubscriptionStateFailed
+			}
```
timflannagan, line 1145: I'll preface this with I could be reading these changes wrong: let's say I have an initial catalog that contains a v1.0.0 that produces a successful install plan, then update that catalog with an olm.bundle v1.0.1 that replaces v1.0.0 that contains a bad image, and eventually patch that catalog to replace that bad version. What ends up happening here? 

It looks like we ensure the Subscription has an install plan reference between calling this ensureSubscriptionCSVState function, which has the following terminal state:

```go
if sub.Status.InstallPlanRef != nil || sub.Status.Install != nil {
		return sub, false, nil
}
```

I'd gather in this poor v1.0.0 -> v1.0.1 rollout scenarios, I'd have a Subscription that's pointing to the new, yet failed CSV that was unpacked, yet that CSV doesn't exist (e.g. the InstallPlan had a bad olm.bundle image). This would imply that despite the UpgradeFailed CSV state, we'd still have an InstallPlan reference that's present in the Subscription's status, which wouldn't result in a new resolution being produced. Is this the right read? Should we be hacking around the populated InstallPlan reference, and reseting that status field? Am I missing something obvious?

I can pull down these changes tomorrow and double check this potential edge case (although this feels like the happy for automatically rolling forward).

pkg/controller/operators/catalog/operator.go
```diff
@@ -1121,6 +1141,14 @@ func (o *Operator) ensureSubscriptionCSVState(logger *logrus.Entry, sub *v1alpha
 	if err != nil {
 		logger.WithError(err).WithField("currentCSV", sub.Status.CurrentCSV).Debug("error fetching csv listed in subscription status")
 		out.Status.State = v1alpha1.SubscriptionStateUpgradePending
+		if failForwardEnabled {
+			ip, err := o.client.OperatorsV1alpha1().InstallPlans(sub.GetNamespace()).Get(context.TODO(), sub.Status.InstallPlanRef.Name, metav1.GetOptions{})
```
timflannagan, line 1145: Do we need to nil check the Subscription has an installplan reference in it's status?

go.mod
```diff
@@ -240,4 +240,6 @@ replace (
 	// controller runtime
 	github.com/openshift/api => github.com/openshift/api v0.0.0-20211014063134-be2a7fb8aa44
 	github.com/openshift/client-go => github.com/openshift/client-go v0.0.0-20200326155132-2a6cd50aedd0 // release-4.5
+	// TODO(fail-forward): Remove when api changes are released
+	github.com/operator-framework/api => ../api
```
timflannagan, line 244: Alternatively: `github.com/operator-framework/api => github.com/awgreene/api v0.0.6-0.20220406202109-98b495eac8a7`?

pkg/controller/registry/resolver/step_resolver.go
```diff
@@ -84,14 +86,54 @@ func NewOperatorStepResolver(lister operatorlister.OperatorLister, client versio
...
 
+	var omittedCSVs []cache.Predicate
+	og, err := r.client.OperatorsV1().OperatorGroups(namespace).List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		return nil, nil, nil, err
+	}
+	if len(og.Items) != 1 {
+		return nil, nil, nil, fmt.Errorf("Too many operatorGroups in namespace %s", namespace)
+	}
+	if og.Items[0].UpgradeStrategy() == v1.UnsafeFailForwardUpgradeStrategy {
```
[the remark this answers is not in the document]
awgreene: I've added a flag, I wasn't sure what you meant by inject predicates directly.

pkg/controller/operators/catalog/operator.go
awgreene: >I'm slightly concerned that UpgradeFailed may be misleading given the initial installation is not really an upgrade. However, given there is no other status, this is the only choice.

In the context of "Failing Forward", I believe that placing the subscription in the failed state is fitting as the subscription kicked off an upgrade and the InstallPlan failed.

pkg/controller/registry/resolver/step_resolver.go
awgreene: Refactored to remove the need to list existing operatorGroups.

pkg/controller/operators/catalog/operator.go
awgreene: Yes, great catch!

pkg/controller/registry/resolver/step_resolver_test.go
```diff
@@ -46,6 +47,240 @@ var (
...
+					},
+				},
+			},
+			expected: out{
+				b:   false,
+				err: fmt.Errorf("clusterServiceVersion bar/foo-v2 in phase Failed instead of Replacing"),
+			},
+		},
+		{
+			name: "InfiniteLoopReplacementChain",
```
perdasilva, line 232: Nice!

pkg/controller/operators/catalog/operator.go
```diff
@@ -903,6 +915,11 @@ func (o *Operator) syncResolvingNamespace(obj interface{}) error {
 		return err
 	}
 
+	failForwardEnabled, err := o.isFailForwardEnabled(namespace)
+	if err != nil {
+		o.logger.WithError(err).Debug("err")
```
perdasilva, line 920: Are there any side effect letting the reconciliation go through even through we cannot tell whether we should failForward? Or will it just keep retrying until we can finally query the OG? If so, would it make sense to just log the error in the `isFailForwardEnabled` function and not return it?

njhale:
@awgreene Please add a summary on the PR with a description of the patch and relevant links to the EP, jira, and elsewhere. I find that a good description really helps future us maintain the project.

pkg/controller/operators/catalog/operator.go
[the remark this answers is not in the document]
njhale: not sure if the diff changed here, but listers query the informer cache; they don't even contain an http client

go.mod
```diff
@@ -240,4 +240,6 @@ replace (
 	// controller runtime
 	github.com/openshift/api => github.com/openshift/api v0.0.0-20211014063134-be2a7fb8aa44
 	github.com/openshift/client-go => github.com/openshift/client-go v0.0.0-20200326155132-2a6cd50aedd0 // release-4.5
+	// TODO(fail-forward): Remove when api changes are released
+	github.com/operator-framework/api => github.com/awgreene/api v0.0.6-0.20220412182759-c4956d3f11bd
```
njhale, line 244: Why don't we merge the api change first, then point this at the pseudo-version for `master`? I'd rather not depend on a personal fork if we can help it.

pkg/controller/operators/catalog/operator.go
timflannagan: The only side effect that I ran into locally was that we were using a nil/unregistered OG lister which I believe resulted in an error being produced, and zero OG being able to be listed in a namespace. This seems like a potential rough edge, as there's potential to silently ignore the default behavior, or silently regress when attempting to list OGs in a namespace that contain this fail forward upgrade strategy as we'll treat an error there as a namespace using the default behavior.

awgreene: Great points Nick, I'm happy pursuing that implementation.

pkg/controller/operators/catalog/operator.go
```diff
@@ -886,10 +886,12 @@ func (o *Operator) syncCatalogSources(obj interface{}) (syncError error) {
...
 	ogs, err := o.lister.OperatorsV1().OperatorGroupLister().OperatorGroups(namespace).List(labels.Everything())
 	if err != nil {
-		return false, fmt.Errorf("couldn't list operatorGroups, assuming default upgradeStrategy")
+		// Couldn't list operatorGroups, assuming default upgradeStrategy
+		// so existing behavior is observed for failed CSVs.
+		return false, nil
 	}
 	if len(ogs) != 1 {
-		return false, fmt.Errorf("found %d operatorGroups, expected 1, assuming default upgradeStrategy", len(ogs))
+		return false, fmt.Errorf("found %d operatorGroups in namespace %s, expected 1", len(ogs), namespace)
```
timflannagan, line 894: Does error-ing out here get reflected back to the Subscription's status?

awgreene: We could do something like [this](https://github.com/operator-framework/operator-lifecycle-manager/blob/8abe078086bba81e2765a244b5a16fe1196f0f7c/pkg/controller/operators/catalog/operator.go#L972)

timflannagan: Yeah that seems reasonable to me? Maybe handle that at the call site though (e.g. o.syncResolvingNamespaces)?

awgreene: o.syncResolvingNamespaces is definitely the correct place to do this, but:
- We'll need to introduce a new [SubscriptionConditionType](https://github.com/operator-framework/api/blob/master/pkg/operators/v1alpha1/subscription_types.go#L91)
- Other errors in the function are not reflected in the status.
I feel like we ought to address this holistically, possibly in a follow up PR, thoughts?

pkg/controller/operators/catalog/operator.go
```diff
@@ -882,6 +883,19 @@ func (o *Operator) syncCatalogSources(obj interface{}) (syncError error) {
 	return
 }
 
+func (o *Operator) isFailForwardEnabled(namespace string) (bool, error) {
```
exdx, line 886: nit: does it make sense to pass the lister in as an argument and not have this function be a method? It doesn't really feel like checking if fail forward is enabled for a namespace should necessarily be part of the catalog operator versus a standalone function. This is more of a preference though, either way is fine.

pkg/controller/operators/catalog/operator.go
```diff
@@ -1110,13 +1129,16 @@ func (o *Operator) ensureSubscriptionInstallPlanState(logger *logrus.Entry, sub
 	out.Status.InstallPlanRef = ref
 	out.Status.Install = v1alpha1.NewInstallPlanReference(ref)
 	out.Status.State = v1alpha1.SubscriptionStateUpgradePending
+	if failForwardEnabled && ip.Status.Phase == v1alpha1.InstallPlanPhaseFailed {
+		out.Status.State = v1alpha1.SubscriptionStateFailed
```
exdx, line 1133: should we add another conditional check for `&& out.Status.State != v1lpha1.SubscriptionStateFailed`? so we don't continuously attempt to update the status if it's already set? Or does the underlying update mechanism handle that case for us?

pkg/controller/registry/resolver/step_resolver.go
```diff
@@ -84,14 +85,122 @@ func NewOperatorStepResolver(lister operatorlister.OperatorLister, client versio
 	return stepResolver
 }
 
-func (r *OperatorStepResolver) ResolveSteps(namespace string) ([]*v1alpha1.Step, []v1alpha1.BundleLookup, []*v1alpha1.Subscription, error) {
+type walkOption func(csv *v1alpha1.ClusterServiceVersion) error
+
+func WithCSVPhase(phase v1alpha1.ClusterServiceVersionPhase) walkOption {
+	return func(csv *v1alpha1.ClusterServiceVersion) error {
+		if csv == nil || csv.Status.Phase != phase {
+			return fmt.Errorf("clusterServiceVersion %s/%s in phase %s instead of %s", csv.GetNamespace(), csv.GetName(), csv.Status.Phase, phase)
```
exdx, line 93: nit: `ClusterServiceVersion` to have consistent capitalization? or maybe simply `csv`?

pkg/controller/registry/resolver/step_resolver.go
```diff
@@ -84,14 +85,122 @@ func NewOperatorStepResolver(lister operatorlister.OperatorLister, client versio
...
+		if csv.Spec.Replaces != "" {
+			replacementMapping[csv.Spec.Replaces] = csv
+		}
+	}
+	return replacementMapping
+}
+
+func (r *OperatorStepResolver) cachePredicates(namespace string) ([]cache.Predicate, error) {
+	predicates := []cache.Predicate{}
+	csvs, err := r.csvLister.ClusterServiceVersions(namespace).List(labels.Everything())
```
exdx, line 164: does it make more sense to filter the list by the copied-csv label here and only get back non-copied ones? Instead of checking below in the `for` loop whether or not it's copied.

awgreene: That seems reasonable to me from a readability standpoint.

pkg/controller/registry/resolver/step_resolver.go
awgreene: I'll change it to `csv`.

pkg/controller/operators/catalog/operator.go
awgreene: In existing behavior the `out.Status.State` is always set to `v1alpha1.SubscriptionStateUpgradePending` and I haven't seen any indication of hotlooping with my tests where the conditional is met, so I think we're safe here.

pkg/controller/operators/catalog/operator.go
timflannagan: Talked in slack: this is fine with me as a follow-up.

timflannagan:
I ran this locally and ran into the following error (after added a debug level log and re-deploying the catalog operator):

```
time="2022-04-14T13:54:20Z" level=debug msg="failed to list the operatorgroups in the ff-e2e-fcprn namespace: cannot list OperatorGroups with a NullOperatorGroupNamespaceLister"
```

/hold

pkg/controller/operators/catalog/operator.go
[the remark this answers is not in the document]
dinhxuanvu: > I ran this locally and ran into the following error (after added a debug level log and re-deploying the catalog operator):
> 
> ```
> time="2022-04-14T13:54:20Z" level=debug msg="failed to list the operatorgroups in the ff-e2e-fcprn namespace: cannot list OperatorGroups with a NullOperatorGroupNamespaceLister"
> ```
> 
> /hold

We don't have OG lister on catalog operator. That's why I said in one of my comments about adding the informer and lister for operatorgroup in order to use the OperatorGroupLister here. (See like 887 on catalog/operator.go).

dinhxuanvu:
> I ran this locally and ran into the following error (after added a debug level log and re-deploying the catalog operator):
> 
> ```
> time="2022-04-14T13:54:20Z" level=debug msg="failed to list the operatorgroups in the ff-e2e-fcprn namespace: cannot list OperatorGroups with a NullOperatorGroupNamespaceLister"
> ```
> 
> /hold

We don't have OG lister on catalog operator. That's why I said in one of my comments in my review about adding the informer and lister for operatorgroup in order to use the OperatorGroupLister here. (See like 887 on catalog/operator.go).

pkg/controller/operators/catalog/operator.go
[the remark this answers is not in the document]
dinhxuanvu: I meant there is currently no informer or lister for OG in catalog operator. You will need to registry it. Otherwise, you will need to use k8s client to query OG directly. My original comment was a bit confusing.

pkg/controller/operators/catalog/operator.go
```diff
@@ -255,6 +258,24 @@ func NewOperator(ctx context.Context, kubeconfigPath string, clock utilclock.Clo
 		return nil, err
 	}
 
+	operatorGroupInformer := crInformerFactory.Operators().V1().OperatorGroups()
```
perdasilva, line 261: should we add a comment about why this is needed? (non-blocking)

pkg/controller/operators/olm/operator_test.go
```diff
@@ -3091,6 +3091,289 @@ func TestTransitionCSV(t *testing.T) {
 	}
 }
 
+// TODO: Merge the following set of tests with those defined in TestTransitionCSV
```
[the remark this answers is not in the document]
awgreene: Yes.

pkg/controller/operators/catalog/operator.go
awgreene: I think we're okay, its a pretty common standard to define informers when you want to list objects in a controller.

pkg/controller/registry/resolver/resolver.go
awgreene: Addressed in https://github.com/operator-framework/operator-lifecycle-manager/pull/2743

PR #2732, opened 2022-04-09
Update ginkgo to v2

perdasilva:
Signed-off-by: Per Goncalves da Silva <pegoncal@redhat.com>

Updates ginkgo to v2

It's about time

**Reviewer Checklist**


.github/workflows/e2e-tests.yml
```diff
@@ -19,6 +19,8 @@ jobs:
       - uses: actions/setup-go@v2
         with:
           go-version: '~1.17'
+      - run: mkdir -p artifacts
```
perdasilva, line 22: Had to create the artifacts dir. Calling ginkgo (v2) with --junit-report ./artifacts/junit-report.xml fails with file not found

.github/workflows/e2e-tests.yml
```diff
@@ -19,6 +19,8 @@ jobs:
...
+      - run: go install github.com/onsi/ginkgo/v2/ginkgo@latest
```
perdasilva, line 23: Had to install ginkgo v2 cli. Calling via go run github.com/onsi/ginkgo/ginkgo still runs v1.

Makefile
```diff
@@ -26,7 +26,7 @@ YQ_INTERNAL := go run $(MOD_FLAGS) ./vendor/github.com/mikefarah/yq/v3/
 KUBEBUILDER_ASSETS := $(or $(or $(KUBEBUILDER_ASSETS),$(dir $(shell command -v kubebuilder))),/usr/local/kubebuilder/bin)
 export KUBEBUILDER_ASSETS
 GO := GO111MODULE=on GOFLAGS="$(MOD_FLAGS)" go
-GINKGO := $(GO) run github.com/onsi/ginkgo/ginkgo
+GINKGO := ginkgo
```
perdasilva, line 29: go run github.com/onsi/ginkgo/ginkgo runs v1. Had to install the v2 cli locally and use that =(

perdasilva: I think I found a way to execute it similar to before. I had to remove -mod=vendor though

.github/workflows/e2e-tests.yml
perdasilva: I think I found a way to execute it similar to before. I had to remove -mod=vendor though

.github/workflows/e2e-tests.yml
timflannagan: I don't think this is blocking but I wonder if we should be handling any stat or directory creation process in the e2e suite?

Makefile
```diff
@@ -11,7 +11,7 @@ endif
 SHELL := /bin/bash
 ORG := github.com/operator-framework
 PKG   := $(ORG)/operator-lifecycle-manager
-MOD_FLAGS := $(shell (go version | grep -q -E "1\.1[1-9]") && echo -mod=vendor)
+MOD_FLAGS := $(shell (go version | grep -q -E "1\.1[1-4]") && echo -mod=vendor)
```
timflannagan, line 14: Any idea on why this was needed?

Makefile
```diff
@@ -132,7 +133,7 @@ E2E_TEST_NUM_CHUNKS ?= 4
 ifneq (all,$(E2E_TEST_CHUNK))
 TEST := $(shell go run ./test/e2e/split/... -chunks $(E2E_TEST_NUM_CHUNKS) -print-chunk $(E2E_TEST_CHUNK) ./test/e2e)
 endif
-E2E_OPTS ?= $(if $(E2E_SEED),-seed '$(E2E_SEED)') $(if $(SKIP), -skip '$(SKIP)') $(if $(TEST),-focus '$(TEST)') -flakeAttempts $(E2E_FLAKE_ATTEMPTS) -nodes $(E2E_NODES) -timeout $(E2E_TIMEOUT) -v -randomizeSuites -race -trace -progress
+E2E_OPTS ?= $(if $(E2E_SEED),-seed '$(E2E_SEED)') $(if $(SKIP), -skip '$(SKIP)') $(if $(TEST),-focus '$(TEST)') $(if $(ARTIFACT_DIR), -junit-report '$(ARTIFACT_DIR)junit_e2e.xml') -flake-attempts $(E2E_FLAKE_ATTEMPTS) -nodes $(E2E_NODES) -timeout $(E2E_TIMEOUT) -v -randomize-suites -race -trace -progress
```
timflannagan, line 136: Looking at https://onsi.github.io/ginkgo/MIGRATING_TO_V2#generating-machine-readable-reports: do we need to handle the --output-dir flag as well? What happens if the $ARTIFACT_DIR variable is set, and an individual test case fails?

timflannagan: Actually disregard: I think we're in the clear here.

Makefile
perdasilva: I've since reverted it. But from the notes, you don't need -mod=vendor from 1.14 I think.

.github/workflows/e2e-tests.yml
timflannagan: I dived into this some more - it looks like ginkgo v2 just fires off an os.Create call so if the artifacts directory doesn't already exist, it will produce an error so creating it beforehand is fine with me for now.

PR #2733, opened 2022-04-09
e2e: Add support for updating MagicCatalogs

timflannagan:

Update the MagicCatalog library and add support for updating existing catalogs.

Enable integrating FBC into OLM's e2e testing suite a bit more practical.

**Reviewer Checklist**


test/e2e/magic_catalog.go
```diff
@@ -78,46 +75,93 @@ func (c *magicCatalog) DeployCatalog(ctx context.Context) error {
...
+		if k8serror.IsNotFound(err) {
+			return true, nil
+		}
+		return false, err
+	})
+	if err != nil {
+		return fmt.Errorf("failed to successfully update the catalog deployment: %v", err)
+	}
 
+	c.fileBasedCatalog = provider
```
timflannagan, line 127: Note: we only update the underlying provider until after we delete the registry pod and configmap resources attached to the previous provider. This fbc_provider.go package might need a setter method, or use the options pattern so NewMagicCatalog().WithProvider(...) is slightly more readable?

test/e2e/magic_catalog_test.go
```diff
@@ -27,15 +28,43 @@ var _ = Describe("MagicCatalog", func() {
...
+		AfterEach(func() {
+			Expect(mc.UndeployCatalog(context.Background())).To(BeNil())
+		})
+
+		It("should succeed when the magic catalog is updated", func() {
+			updatedProvider, err := NewFileBasedFiledBasedCatalogProvider(filepath.Join(testdataDir, "magiccatalog/fbc_updated.yaml"))
+			Expect(err).To(BeNil())
+
+			updatedMC := NewMagicCatalog(c, generatedNamespace.GetName(), catalogName, updatedProvider)
+			Expect(updatedMC.UpdateCatalog(context.Background(), updatedProvider)).To(BeNil())
```
perdasilva, line 64: is it worth checking that the contents of the catalog got updated somehow:?

timflannagan: Yep - I had briefly thought about this yesterday when poking at the update logic. I was tempted to verify the grpc contents directly but that didn't feel very representative of an e2e test. I think validating that the configmap contents have changed might be sufficient? Thoughts?

test/e2e/testdata/magiccatalog/fbc_updated.yaml
```diff
@@ -0,0 +1,42 @@
...
+properties:
+  - type: olm.gvk
+    value:
+      group: example.com
+      kind: TestA
+      version: v1alpha1
+  - type: olm.package
+    value:
+      packageName: packageA
+      version: 1.0.0
```
timflannagan, line 27: Note: I realize I can probably reduce this file size by removing this olm.bundle channel entry entirely and this would still be a valid FBC.

test/e2e/magic_catalog_test.go
perdasilva: Yeah - I'm good with that. Just a sanity check. You could also check against a check sum or something.

timflannagan: What did you have in mind for a check sum implementation? That seems like an interesting angle, but I ended up settling with comparing the configmap catalog.json directly with our expectation of those contents.

test/e2e/magic_catalog.go
```diff
@@ -50,74 +52,119 @@ func NewMagicCatalog(kubeClient k8scontrollerclient.Client, namespace string, ca
...
+		}, cs)
+		if err != nil || cs.Status.GRPCConnectionState == nil {
+			return false, err
+		}
+		state := cs.Status.GRPCConnectionState.LastObservedState
+		if state != catalogReadyState {
+			return false, nil
+		}
+		return true, nil
+	})
```
awgreene, line 87:
```suggestion
return state == catalogReadyState, nil
```

timflannagan: Yep - I had noticed this locally too but forgot to push up a change.

awgreene (commented):
A late review but this looks great!

PR #2741, opened 2022-04-14
e2e: Add support for overriding the kubectl binary when gather artifacts

timflannagan:

Add support for overriding the kubectl binary that's responsible for gathering namespace-scoped test artifacts when individual test cases fail.

Avoid hardcoding the kubectl binary location when kubectl isn't present in a users/CI environment's $PATH.

**Reviewer Checklist**


test/e2e/collect-ci-artifacts.sh
```diff
@@ -7,7 +7,10 @@ set -o errexit
 : "${KUBECONFIG:?}"
 : "${TEST_NAMESPACE:?}"
 : "${TEST_ARTIFACTS_DIR:?}"
+: "${KUBECTL:=kubectl}"
```
timflannagan, line 10: Note: this KUBECTL variable defaults to kubectl when it hasn't been defined/overridden locally. This change now allows users to override this variable in their local/CI environments, e.g. `make e2e KUBECTL=oc`.

test/e2e/ctx/ctx.go
```diff
@@ -132,6 +132,10 @@ func (ctx TestContext) DumpNamespaceArtifacts(namespace string) error {
 		"TEST_ARTIFACTS_DIR=" + logDir,
 		"KUBECONFIG=" + kubeconfigPath,
 	}
+	kubectl := os.Getenv("KUBECTL")
+	if kubectl != "" {
+		envvars = append(envvars, fmt.Sprintf("KUBECTL=%s", kubectl))
+	}
```
njhale, line 138: nit: can we condense this into the literal?

```go
envvars := []string{
    "TEST_NAMESPACE=" + namespace,
    "TEST_ARTIFACTS_DIR=" + logDir,
    "KUBECONFIG=" + kubeconfigPath,
    "KUBECTL=" + os.Getenv("KUBECTL"),
}
```

Related: could also avoid a lot of dependency issues by porting `collect-ci-artifacts.sh` to go?

timflannagan: What happens if KUBECTL is empty (which should be the default case)?

njhale: I imagine the script would detect an empty/non-existent environment variable and would use a default

timflannagan: I'm not sure where I was going with my last comment - makes sense to me given we already provide a default value when the $KUBECTL variable is undefined or empty, so we should be good to go. Updated in the last force push.

PR #2743, opened 2022-04-14
Identify fail forward in csvSources

awgreene:

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/registry/resolver/fail_forward.go
+++ b/pkg/controller/registry/resolver/fail_forward.go
@@ -1,0 +1,103 @@
+package resolver
+
+import (
+	"fmt"
+
+	operatorsv1 "github.com/operator-framework/api/pkg/operators/v1"
+	operatorsv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
+	v1listers "github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/listers/operators/v1"
+	"k8s.io/apimachinery/pkg/labels"
+)
+
+// IsFailForwardEnabled takes a namespaced operatorGroup lister and returns
+// True if an operatorGroup exists in the namespace and its upgradeStrategy
+// is set to UnsafeFailForward and false otherwise. An error is returned if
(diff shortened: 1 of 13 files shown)


pkg/controller/registry/resolver/resolver_test.go
```diff
@@ -196,80 +196,6 @@ func WithInstalledCSV(sub *v1alpha1.Subscription, csvName string) *v1alpha1.Subs
 	return sub
 }
 
-func TestSolveOperators_WithFailForward(t *testing.T) {
```
awgreene, line 199: Removed since the changes are not done on the resolver level anymore.

pkg/controller/registry/resolver/fail_forward.go
```diff
@@ -0,0 +1,103 @@
...
+	v1listers "github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/listers/operators/v1"
+	"k8s.io/apimachinery/pkg/labels"
+)
+
+// IsFailForwardEnabled takes a namespaced operatorGroup lister and returns
+// True if an operatorGroup exists in the namespace and its upgradeStrategy
+// is set to UnsafeFailForward and false otherwise. An error is returned if
+// an more than one operatorGroup exists in the namespace.
+// No error is returned if no OperatorGroups are found to keep the resolver
+// backwards compatible.
```
tylerslaton, line 17: nit:
```suggestion
// IsFailForwardEnabled takes a namespaced operatorGroup lister and returns
// True if an operatorGroup exists in the namespace and its upgradeStrategy
// is set to UnsafeFailForward and false otherwise. An error is returned if
// more than one operatorGroup exists in the namespace. No error is returned 
// if no OperatorGroups are found to keep the resolver backwards compatible.
```

tylerslaton (commented):
Great quality of life improvement here for the resolver as we don't have to continuously add the "fail forward" boolean for resolution. Eventually we'll need to remove the concept of CSV's from the resolver package entirely as a result of the breakout but until then this look
good. One small non-blocking nit.

/lgtm

PR #2744, opened 2022-04-15
Remove the run-console-local Makefile target

timflannagan:

Remove the largely unused run-console-local Makefile target which is responsible for running the openshift console in a docker container. This kind of target doesn't feel appropriate for an upstream environment and it reduces overall maintenance. 

Cleanup

**Reviewer Checklist**


awgreene:
Agreed, this can live downstream if we need it.
/lgtm

PR #2749, opened 2022-04-19
fix: limit resolution scope to locally referenced catalogs in a subscription 

exdx:
Add logic to the resolver to exclude the global catalog namespace from
resolution in the case where all subscriptions in the namespace
reference the namespace directly. This prevents errors from catalogs in
the global namespace to affect resolution in other, non-related
namespaces scoped only to that namespace.

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/registry/resolver/step_resolver.go
+++ b/pkg/controller/registry/resolver/step_resolver.go
@@ -91,7 +91,22 @@ func (r *OperatorStepResolver) ResolveSteps(namespace string) ([]*v1alpha1.Step,
 		return nil, nil, nil, err
 	}
 
-	namespaces := []string{namespace, r.globalCatalogNamespace}
+	// if spec.sourceNamespace is provided on all the subscriptions in the namespace, only consider that namespace for resolution
+	// this prevents errors when unnecessarily searching in the global catalog namespace when those catalogs are unavailable
+	var namespaces = []string{namespace}
+	var searchGlobal bool
+	for _, sub := range subs {
+		if sub.Spec.CatalogSourceNamespace == namespace {
+			continue
+		} else {
+			searchGlobal = true
+			break
+		}
+	}
+	if searchGlobal {
+		namespaces = append(namespaces, r.globalCatalogNamespace)
+	}
+
 	operators, err := r.resolver.Resolve(namespaces, subs)
 	if err != nil {
 		return nil, nil, nil, err
--- a/test/e2e/catalog_e2e_test.go
+++ b/test/e2e/catalog_e2e_test.go
@@ -1340,6 +1340,97 @@ var _ = Describe("Starting CatalogSource e2e tests", func() {
 			}
 		}
 	})
+
+	When("a broken catalog is created in the global catalog namespace", func() {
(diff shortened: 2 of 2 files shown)


benluddy:
Is this approach safe? We don't know up-front whether or not one of the catalogs in the global namespace might provide an operator that satisfies a dependency from one of the local catalogs.

exdx:
> Is this approach safe? We don't know up-front whether or not one of the catalogs in the global namespace might provide an operator that satisfies a dependency from one of the local catalogs.

It's probably not entirely safe, no. I realize this approach would break existing users who would expect resolution of their dependencies to continue to behave the way they did before, which means this is not a safe change. 

I still think we need to do something with this overall problem though -- I'm going to write a doc outlining some different options. I'll leave this PR up for now as a reference but will close it soon.

dinhxuanvu:
> Is this approach safe? We don't know up-front whether or not one of the catalogs in the global namespace might provide an operator that satisfies a dependency from one of the local catalogs.

This approach certainly has its problems but I think the intention here is to resolve dependencies with what are available at that moment in time.

pkg/controller/registry/resolver/step_resolver.go
```diff
@@ -91,7 +91,22 @@ func (r *OperatorStepResolver) ResolveSteps(namespace string) ([]*v1alpha1.Step,
...
+	var searchGlobal bool
+	for _, sub := range subs {
+		if sub.Spec.CatalogSourceNamespace == namespace {
+			continue
+		} else {
+			searchGlobal = true
+			break
+		}
+	}
+	if searchGlobal {
```
dinhxuanvu, line 106: I feel this namespace exclusion approach is a bit too broad though I can see it is the simplest solution for this particular problem. This solution preemptively excludes global catalogsources from resolution regardless if there are any issues with them or not. I would like to see if we can have a more refined solution that will still consider global catalogsources and simply ignore errors from those catalogsources if occur and let the resolution continue.

dinhxuanvu (changes requested):

pkg/controller/registry/resolver/step_resolver.go
benluddy: Ignoring catalog errors breaks resolution determinism, since any transient error could cause a different result. How about making this explicitly configurable on a per-namespace basis via a new field on the OperatorGroup spec?

dinhxuanvu: Adding configurable settings in OperatorGroup is quite trending later so I don't see the harm of doing the same for this particular scenario. My point still stands though on the fact preemptive solution is too broad.

njhale: > Ignoring catalog errors breaks resolution determinism, since any transient error could cause a different result. How about making this explicitly configurable on a per-namespace basis via a new field on the OperatorGroup spec?

@benluddy I think we have other options too, like:
- lean on the SAT-solver for resolution determinism; e.g. operators that require deterministic resolution have constraints on unique label properties only available in the "correct catalog"

exdx:
Write-up doc: https://hackmd.io/BOA66e55QqG8XjBMIMFvIg

PR #2752, opened 2022-04-26
Fix csv.status.RotatedAt update

perdasilva:
Signed-off-by: Per Goncalves da Silva <pegoncal@redhat.com>

This PR updates the CSV InstallReady state to only update the`csv.status.CertsLastUpdatedAt` and `csv.status.CertsRotateAt`
status fields if they are not set (and the csv contains CAResources), or if expired.

NOTE: I couldn't find a clear way to surface the certificate expiry and rotation dates from the installer. So, I patched and cleaned the current implementation. Though, if anyone has good ideas, I'm all ears!

The complete context can be seen in the [bz](https://bugzilla.redhat.com/show_bug.cgi?id=2020484).
Essentially, if the package server pod get killed, the csv moves to a `Failed` states and progresses forward to `Succeeded`. During the `InstallReady` phase, the `installer` would not rotate the csv if the certificate has not expired yet. BUT, the csv status fields were still being updated, moving the rotatedAt out another 2 years. This creates skew between the certificate's rotation time and the what's in the csv status. However, the operator decides when to rotate the certificate based on the `status.RotateAt`. Which would in turn push out the rotation erroneously.

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/olm/operator.go
+++ b/pkg/controller/operators/olm/operator.go
@@ -1888,10 +1888,19 @@ func (a *Operator) transitionCSVState(in v1alpha1.ClusterServiceVersion) (out *v
 			return
 		}
 
-		if out.HasCAResources() {
+		// Only update certificate status if:
+		//  - the CSV has CAResources; and
+		//  - the certificate lastUpdated and rotateAt timestamps have not already been set, or the certificate should be rotated
+		// Note: the code here is a bit wonky and it wasn't clear how to clean it up without some major refactoring:
+		// the installer is in charge of generating and rotating the certificates. It detects whether a certificate should be rotated by
+		// looking at the csv.status.RotateAt value. But, it does not update this value or surface
+		// the certificate expiry information. So, the rotatedAt value is to be re-calculated here. This is bad because you have
+		// two different components doing the same thing (installer and operator are both calculating RotateAt). If we're not careful
+		// there could be skew
+		// See pkg/controller/install/certresources.go
+		if shouldUpdateCertificateDates(out) {
 			now := metav1.Now()
-			expiration := now.Add(install.DefaultCertValidFor)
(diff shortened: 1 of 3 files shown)


pkg/controller/operators/olm/operator.go
```diff
@@ -1888,10 +1888,17 @@ func (a *Operator) transitionCSVState(in v1alpha1.ClusterServiceVersion) (out *v
 			return
 		}
 
-		if out.HasCAResources() {
+		// Only update certificate status if:
+		//  - there the CSV has CAResources; and
+		//  - it has not already been set, or they should be rotated
+		// See bug: https://bugzilla.redhat.com/show_bug.cgi?id=2020484 for more information
```
awgreene, line 1894:
```suggestion
		// Only update certificate status if:
		//  - the CSV has CAResources; and
		//  - it has not already been set, or they should be rotated
```

awgreene: +1 IRT dropping the BZ reference.

awgreene: BZ is captured in the Pull Request description anyways.

awgreene (approved):
Looks good, giving this PR my approval.
/approve

pkg/controller/operators/olm/operator.go
perdasilva: Lol - just wrote that in the previous response. I'm good with removing that.

dinhxuanvu:
I can lgtm after test case(s) is added and rebased.

tylerslaton (approved):
Great work here Per, giving it a rubber stamp and echoing what everyone else has said. The implementation looks good to me and I'll give it the official label once the test cases come through.

dinhxuanvu (approved):
/lgtm

dinhxuanvu:
I think the PR looks good. Ready for merge whenever you remove the WIP tag on the title.

PR #2772, opened 2022-05-10
Multiple operatorGroup error in subscription status

awgreene:
Problem: When reconciling subscriptions in a namespace, OLM will error
out if multiple operatorGroups exist in the namespace when deciding if
the fail-forward feature is enabled. The failure is only loged and not
surfaced in the status of the subscription.

Solution: By assuming that fail forward is disabled when more than one
operatorGroups are found in a namespace, the syncResolvingNamespace
function will attempt to find upgrades using the resolver which will
fail due to multiple operatorGroups in the namespace anyways and
portray this information in the subscription status.

### DIFF
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -934,9 +934,16 @@ func (o *Operator) syncResolvingNamespace(obj interface{}) error {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


ankitathomas (approved):
/lgtm

perdasilva:
@awgreene could you please rebase? I can't do it from this side =(

PR #2779, opened 2022-05-19
Make CatalogSource the source of truth for available catalogs.

benluddy:
Internally, the catalog operator has always maintained a set of
registry clients for each CatalogSource. Although this set is
reconciled toward containing a client per CatalogSource object, there
is some latency before changes made to CatalogSources are reflected in
the client set, and differences between CatalogSources and client set
membership are a potential source of error.

Instead, the catalog operator should list CatalogSources from its
informer cache to determine which catalogs are reachable from a given
namespace.

### DIFF
--- a/pkg/controller/registry/resolver/source_registry.go
+++ b/pkg/controller/registry/resolver/source_registry.go
@@ -8,12 +8,14 @@ import (
 	"time"
 
 	"github.com/blang/semver/v4"
+	v1alpha1listers "github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/listers/operators/v1alpha1"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/registry"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/registry/resolver/cache"
 	"github.com/operator-framework/operator-registry/pkg/api"
 	"github.com/operator-framework/operator-registry/pkg/client"
 	opregistry "github.com/operator-framework/operator-registry/pkg/registry"
 	"github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/labels"
 )
 
 // todo: move to pkg/controller/operators/catalog
@@ -66,14 +68,16 @@ func (i *sourceInvalidator) GetValidChannel(key cache.SourceKey) <-chan struct{}
 
 type RegistrySourceProvider struct {
 	rcp         RegistryClientProvider
+	catsrcs     v1alpha1listers.CatalogSourceLister
 	logger      logrus.StdLogger
 	invalidator *sourceInvalidator
 }
 
-func SourceProviderFromRegistryClientProvider(rcp RegistryClientProvider, logger logrus.StdLogger) *RegistrySourceProvider {
+func SourceProviderFromRegistryClientProvider(rcp RegistryClientProvider, catsrcs v1alpha1listers.CatalogSourceLister, logger logrus.StdLogger) *RegistrySourceProvider {
 	return &RegistrySourceProvider{
-		rcp:    rcp,
-		logger: logger,
+		rcp:     rcp,
+		catsrcs: catsrcs,
+		logger:  logger,
 		invalidator: &sourceInvalidator{
 			validChans: make(map[cache.SourceKey]chan struct{}),
 			ttl:        5 * time.Minute,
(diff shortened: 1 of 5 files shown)


pkg/controller/registry/resolver/source_registry.go
```diff
@@ -66,22 +69,41 @@ func (i *sourceInvalidator) GetValidChannel(key cache.SourceKey) <-chan struct{}
...
+func (s errorSource) Snapshot(_ context.Context) (*cache.Snapshot, error) {
+	return nil, s.error
+}
+
 func (a *RegistrySourceProvider) Sources(namespaces ...string) map[cache.SourceKey]cache.Source {
+	var catalogs []*v1alpha1.CatalogSource
+	for _, namespace := range namespaces {
+		cats, err := a.catsrcs.CatalogSources(namespace).List(labels.Everything())
+		if err != nil {
+			// todo: this must be properly exposed
```
benluddy, line 102: This definitely needs to be addressed before merging.

test/e2e/subscription_e2e_test.go
```diff
@@ -62,6 +64,82 @@ var _ = Describe("Subscription", func() {
...
+				cond := sub.Status.GetCondition(v1alpha1.SubscriptionResolutionFailed)
+				ctx.Ctx().Logf("%#v\n", cond)
+				return v1alpha1.SubscriptionCondition{
+					Type:   cond.Type,
+					Reason: cond.Reason,
+					Status: cond.Status,
+				}, nil
+			}
+
+			// this doesn't seem very robust. basically, this subscription condition should arrive directly at True/ErrorPreventedResolution without passing through any other non-Unknown states
```
benluddy, line 123: Not happy with this test, but it consistently reproduced the bug for me on master.

benluddy:
Here's the progression of the ResolutionFailed condition with this patch. The second line is new:

```
v1alpha1.SubscriptionCondition{Type:"ResolutionFailed", Status:"Unknown", Reason:"", Message:"", LastHeartbeatTime:<nil>, LastTransitionTime:<nil>}
v1alpha1.SubscriptionCondition{Type:"ResolutionFailed", Status:"True", Reason:"ErrorPreventedResolution", Message:"error using catalog without-registry-server-q8l65 (in namespace subscription-e2e-q8q7t): registry server not reachable for catalogsource subscription-e2e-q8q7t/without-registry-server-q8l65", LastHeartbeatTime:<nil>, LastTransitionTime:<nil>}
v1alpha1.SubscriptionCondition{Type:"ResolutionFailed", Status:"True", Reason:"ErrorPreventedResolution", Message:"error using catalog without-registry-server-q8l65 (in namespace subscription-e2e-q8q7t): failed to list bundles: rpc error: code = Unavailable desc = connection error: desc = \"transport: Error while dialing dial tcp 10.96.106.125:50051: i/o timeout\"", LastHeartbeatTime:<nil>, LastTransitionTime:<nil>}
```

Without it, the unreachable catalog is treated as completely empty and resolution proceeds as normal. This can produce ResolutionFailed: true with reason ConstraintsNotSatsfiable, or it can actually choose to install something that is technically acceptable but lower priority than the candidate in the unreachable catalog. In other words, transient skew between the SourceStore and the informer cache can cause the catalog operator to behave nondeterministically.

anik120:
Thanks for the PR @benluddy 🎉 

Do you think you'll have some cycles to get this PR to a reviewable stage? Or would you prefer someone from the operator-framework team to take over this PR for you?

test/e2e/subscription_e2e_test.go
```diff
@@ -64,6 +66,82 @@ var _ = Describe("Subscription", func() {
...
+				},
+			}
+			Expect(ctx.Ctx().Client().Create(context.Background(), &sub)).To(Succeed())
+
+			getCondition := func() (v1alpha1.SubscriptionCondition, error) {
+				if err := ctx.Ctx().Client().Get(context.Background(), client.ObjectKeyFromObject(&sub), &sub); err != nil {
+					return v1alpha1.SubscriptionCondition{}, err
+				}
+				cond := sub.Status.GetCondition(v1alpha1.SubscriptionResolutionFailed)
+				ctx.Ctx().Logf("%#v\n", cond)
```
oceanc80, line 117: Clean up logging

benluddy:
I still think this is worthwhile, but I am several years out of the loop when it comes to operator framework issues.

PR #2781, opened 2022-05-21
[WIP] bug(2076323): Refactor resolution to bail if any referenced catsrc is unreachable

perdasilva:

Updates the resolved to bail if any of the catalog sources references by the subscriptions being resolved is unavailable

https://hackmd.io/BOA66e55QqG8XjBMIMFvIg?view

**Reviewer Checklist**


pkg/controller/registry/resolver/cache/cache.go
```diff
@@ -145,6 +145,29 @@ func (c *NamespacedOperatorCache) Error() error {
...
+	sourceKey := SourceKey{
+		Name:      name,
+		Namespace: namespace,
+	}
+
+	snapshot, ok := c.snapshots[sourceKey]
+
+	// Should this return an error?
+	// Or will the resolution fail because the catsrc hasn't been indexed yet
+	// and not bundle found?
```
benluddy, line 158: Maybe for packages references directly by a Subscription, but not necessarily for an operator that is installed to satisfy a dependency.

pkg/controller/registry/resolver/resolver.go
```diff
@@ -114,10 +123,6 @@ func (r *Resolver) Resolve(namespaces []string, subs []*v1alpha1.Subscription) (
 
 	r.addInvariants(namespacedCache, variables)
 
-	if err := namespacedCache.Error(); err != nil {
```
benluddy, line 117: There are cache sources that do not correspond to registry pods (like the CSVs that are already installed), so removing this ignores more errors than you're intending.

exdx, line 117: Would removing this error check here cause all errors to get overlooked, versus just those related to catalog connectivity? Are we risking some other kind of error case sneaking through and causing issues in the `Solve()` function below? 

Maybe to be safe, we can introduce a custom error type that is specific for catalog connectivity issues, and purposefully drop that error here?

benluddy (commented):
I'm not sure about intentionally making dependency selection nondeterministic, but putting that aside, the `cache.Source` interface (and SourceProvider) is intended to encompass this kind of cache population behavior without changing the resolver or resolver cache implementations. I opened a similar PR last week: https://github.com/operator-framework/operator-lifecycle-manager/pull/2779.

pkg/controller/registry/resolver/resolver.go
```diff
@@ -85,6 +85,15 @@ func (r *Resolver) Resolve(namespaces []string, subs []*v1alpha1.Subscription) (
 		// find the currently installed operator (if it exists)
 		var current *cache.Entry
 
+		// Bail resolution if one of the catalog sources in the subscriptions to resolve is in a bad state
+		// Mental model:
+		// The namespace and its subscriptions are resolved in unison. We ride together, we die together.
+		// We should respect the subscription spec and get the top level operator from the given catalog.
+		// We do our best to resolve dependencies from the available catalog sources in priority order.
+		if err := namespacedCache.GetSnapshotError(sub.Spec.CatalogSourceNamespace, sub.Spec.CatalogSource); err != nil {
```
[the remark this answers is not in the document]
perdasilva: This was an open question I had. I looked at the api definition and they don't source and sourceNamespace seem to be required: https://github.com/operator-framework/api/blob/master/pkg/operators/v1alpha1/subscription_types.go#L32

(at least there's no optional annotation - am I reading it right?)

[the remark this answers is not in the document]
benluddy: https://github.com/operator-framework/api/blob/master/crds/operators.coreos.com_subscriptions.yaml#L55-L58

But the empty string validates against this schema. IIRC this is supported and means "any CatalogSource accessible by this namespace".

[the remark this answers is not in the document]
exdx: Hmm, I thought that they were not required, but it looks to be the case that they are. I tried to apply the following Subscription:
```
k apply -f -<<EOF
kind: Subscription
metadata:
  name: busybox-sub
spec:
  channel: alpha
  name: busybox
EOF
```

and got back the following error:
```
error: error validating "STDIN": error validating data: [ValidationError(Subscription.spec): missing required field "source" in com.coreos.operators.v1alpha1.Subscription.spec, ValidationError(Subscription.spec): missing required field "sourceNamespace" in com.coreos.operators.v1alpha1.Subscription.spec]; if you choose to ignore these errors, turn validation off with --validate=false
```

As Ben pointed out, an empty string is considered valid though:
```
k apply -f -<<EOF
kind: Subscription
metadata:
  name: busybox-sub
spec:
  channel: alpha
  name: busybox
  source: ""
  sourceNamespace: ""
EOF

subscription.operators.coreos.com/busybox-sub created
```

so we may potentially need to handle this edge-case here.

pkg/controller/registry/resolver/step_resolver.go
```diff
@@ -31,16 +32,43 @@ type StepResolver interface {
...
+}
+
+func NewNamespaceAnnotationGlobalCatalogToggle(opClient operatorclient.ClientInterface) *NamespaceAnnotationGlobalCatalogToggle {
+	return &NamespaceAnnotationGlobalCatalogToggle{
+		opClient: opClient,
+	}
+}
+
+type NamespaceAnnotationGlobalCatalogToggle struct {
+	opClient operatorclient.ClientInterface
```
benluddy, line 57: Catalog operator should already have an informer for namespaces -- you likely want to query via the Lister interface instead of directly making an API request.

pkg/controller/registry/resolver/step_resolver.go
```diff
@@ -85,13 +113,40 @@ func NewOperatorStepResolver(lister operatorlister.OperatorLister, client versio
...
+	if err != nil {
+		return nil, nil, nil, err
+	}
+
+	namespaces := []string{namespace}
+
+	if includeGlobalCatalogs {
+		namespaces = append(namespaces, r.globalCatalogNamespace)
+	}
+
```
benluddy, line 149: This kind of behavior is exactly what cache.SourceProvider is meant to implement. It is effectively `f(namespaces) -> sources`, so you should be able to write an implementation that conditionally removes the global namespace and then delegates to another SourceProvider.

pkg/controller/registry/resolver/step_resolver.go
```diff
@@ -31,16 +32,43 @@ type StepResolver interface {
...
+	return &NamespaceAnnotationGlobalCatalogToggle{
+		opClient: opClient,
+	}
+}
+
+type NamespaceAnnotationGlobalCatalogToggle struct {
+	opClient operatorclient.ClientInterface
+}
+
+func (n *NamespaceAnnotationGlobalCatalogToggle) ShouldEnableGlobalNamespaceResolution(ctx context.Context, namespace string) (bool, error) {
```
exdx, line 60: Is it safer to have this be `ShouldDisableGlobalNamespaceResolution` and preserve the default resolution behavior by default? I think we should see this as users can opt-out of global resolution on a per-namespace basis, versus opting-in to global catalog resolution, which is the existing behavior

pkg/controller/registry/resolver/step_resolver.go
```diff
@@ -31,16 +32,43 @@ type StepResolver interface {
...
+	fmt.Println("[PING] Getting global catalog source toggle")
+	ns, err := n.opClient.KubernetesInterface().CoreV1().Namespaces().Get(ctx, namespace, metav1.GetOptions{})
+	fmt.Printf("Got namespace %s and error %s", ns, err)
+	if err != nil {
+		return false, err
```
exdx, line 65: In the case of an error fetching the namespace, we should preserve the existing global resolution behavior (which would mean returning true here). It may be more logical to invert the function and return false here instead (to not disable global resolution).

pkg/controller/registry/resolver/step_resolver.go
```diff
@@ -31,16 +32,43 @@ type StepResolver interface {
...
+	}
+	_, ok := ns.Annotations["olm.disableGlobalCatalogs"]
```
exdx, line 67: Maybe we should consider the value of the annotation as well? For example, if a user sets it to `true`, then we would disable resolution from the global catalogs. But if the same user were to later decide that they did want resolution from the global catalogs as well, they could set the annotation to anything else (?) and then can opt-in to global resolution again? 

But I'm not opposed to the existing implementation where we would just say "put this annotation key on the namespace, regardless of the value it will work as expected". It just seems like maybe if they set the value to false they may expect the behavior to revert, which it would not.

PR #2788, opened 2022-05-31
resolver: Add support for excluding global catalogs from resolution

exdx:
The resolver is currently configured to always consider resolving
from global catalogs. In some cases, this may not be
desirable, and the user would like to explicitly ignore global catalogs
for the purposes of resolution. This change enables per-namespace
exclusion from global catalog sources via an annotation on the
operatorgroup.

The operatorgroup annotation key is `olm.operatorframework.io/exclude-global-namespace-resolution` 
and setting the value to `"true"` will cause resolution to exclude global catalogs in that namespace. 

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/catalog/og_source_provider.go
+++ b/pkg/controller/operators/catalog/og_source_provider.go
@@ -1,0 +1,82 @@
+package catalog
+
+import (
+	"context"
+	"fmt"
+
+	v1listers "github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/listers/operators/v1"
+
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/registry/resolver/cache"
+	"github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/labels"
+)
+
+type OperatorGroupToggleSourceProvider struct {
+	sp       cache.SourceProvider
+	logger   *logrus.Logger
+	ogLister v1listers.OperatorGroupLister
+}
+
+func NewOperatorGroupToggleSourceProvider(sp cache.SourceProvider, logger *logrus.Logger,
+	ogLister v1listers.OperatorGroupLister) *OperatorGroupToggleSourceProvider {
+	return &OperatorGroupToggleSourceProvider{
+		sp:       sp,
+		logger:   logger,
+		ogLister: ogLister,
+	}
+}
+
+const exclusionAnnotation string = "olm.operatorframework.io/exclude-global-namespace-resolution"
+
+func (e *OperatorGroupToggleSourceProvider) Sources(namespaces ...string) map[cache.SourceKey]cache.Source {
+	// Check if annotation is set first
+	resolutionNamespaces, err := e.CheckForExclusion(namespaces...)
+	if err != nil {
+		e.logger.Errorf("error checking namespaces %#v for global resolution exlusion: %s", namespaces, err)
+		// Fail early with a dummy Source that returns an error
+		// TODO: Update the Sources interface to return an error
+		m := make(map[cache.SourceKey]cache.Source)
+		key := cache.SourceKey{Name: "operatorgroup-unavailable", Namespace: namespaces[0]}
+		source := errorSource{err}
+		m[key] = source
+		return m
+	}
+	return e.sp.Sources(resolutionNamespaces...)
+}
+
+type errorSource struct {
+	error
+}
+
+func (e errorSource) Snapshot(ctx context.Context) (*cache.Snapshot, error) {
+	return nil, e.error
+}
+
+func (e *OperatorGroupToggleSourceProvider) CheckForExclusion(namespaces ...string) ([]string, error) {
+	var defaultResult = namespaces
+	// The first namespace provided is always the current namespace being synced
+	var ownNamespace = namespaces[0]
+	var toggledResult = []string{ownNamespace}
+
+	// Check the OG on the NS provided for the exclusion annotation
+	ogs, err := e.ogLister.OperatorGroups(ownNamespace).List(labels.Everything())
+	if err != nil {
+		return nil, fmt.Errorf("listing operatorgroups in namespace %s: %s", ownNamespace, err)
+	}
+
+	if len(ogs) != 1 {
+		// Problem with the operatorgroup configuration in the namespace, or the operatorgroup has not yet been persisted
+		// Note: a resync will be triggered if/when the operatorgroup becomes available
+		return nil, fmt.Errorf("found %d operatorgroups in namespace %s: expected 1", len(ogs), ownNamespace)
+	}
+
+	var og = ogs[0]
+	if val, ok := og.Annotations[exclusionAnnotation]; ok && val == "true" {
+		// Exclusion specified
+		// Ignore the globalNamespace for the purposes of resolution in this namespace
+		e.logger.Printf("excluding global catalogs from resolution in namespace %s", ownNamespace)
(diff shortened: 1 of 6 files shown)


pkg/controller/registry/resolver/cache/cache.go
```diff
@@ -52,6 +52,11 @@ type SourceProvider interface {
 	Sources(namespaces ...string) map[SourceKey]Source
 }
 
+type SourceProviderWithInvalidate interface {
+	SourceProvider
+	Invalidate(key SourceKey)
+}
```
benluddy, line 58: Since Invalidate isn't used by the cache package, can we do without declaring this interface type here?

It might be reasonable to define an interface beside `OperatorGroupToggleSourceProvider` that looks like

```go
type SourceInvalidator interface {
 Invalidate(SourceKey)
}
```

then change `OperatorGroupToggleSourceProvider` to accept plain old `cache.SourceProvider` as its wrapped type. Then, in `(OperatorGroupToggleSourceProvider).Invalidate(SourceKey)`, do a type-assertion on the wrapped value to see if it implements `SourceInvalidator` to decide whether or not to delegate the invalidate call.

This would be slightly more effort, but idiomatic. More importantly, it avoids expanding the Go API surface of resolver and its subpackages. There's still interest in treating those as a shared library.

perdasilva: Could we just move it to operator.go and define it as a client side interface?

pkg/controller/registry/resolver/source_registry.go
```diff
@@ -302,3 +306,73 @@ func legacyDependenciesToProperties(dependencies []*api.Dependency) ([]*api.Prop
...
+func (e *OperatorGroupToggleSourceProvider) Invalidate(key cache.SourceKey) {
+	e.sp.Invalidate(key)
+}
+
+func (e *OperatorGroupToggleSourceProvider) CheckForExclusion(namespaces ...string) []string {
+	var defaultResult = namespaces
+	// Only two namespaces are provided during the course of resolution, in deterministic order
+	// The first is the namespace being synced and the second is the global namespace
+	var ownNamespace = namespaces[0]
+	var globalNamespace = namespaces[1]
```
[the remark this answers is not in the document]
exdx: Yes, that sounds like a better idea. I think I can thread that namespace variable through to the underlying `OperatorGroupToggleSourceProvider`

pkg/controller/registry/resolver/source_registry.go
```diff
@@ -302,3 +306,73 @@ func legacyDependenciesToProperties(dependencies []*api.Dependency) ([]*api.Prop
...
+
+	if ownNamespace == globalNamespace {
+		// Global namespace is being synced
+		// Return early with default
+		return defaultResult
+	}
+
+	// Check the OG on the NS provided for the exclusion annotation
+	ogs, err := e.ogLister.OperatorGroups(ownNamespace).List(labels.Everything())
+	if err != nil {
```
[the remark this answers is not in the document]
exdx: That sounds reasonable -- is it best to do so in this PR? I'm just a bit wary because I think we will have to backport this down and changing the interface would expand the diff. 

I think we can get away with simply going with the default behavior in the case of an error? That does lead to the undesirable case where exclusion was specified and because the List OperatorGroup call failed we end up doing global resolution instead. That might be enough of an issue to where we would have to add an error return type to the interface in this PR?

pkg/controller/operators/catalog/og_source_provider.go
```diff
@@ -0,0 +1,106 @@
...
+	if err != nil {
+		// Assume resolution was global in the case of an error (the default behavior)
+		return defaultResult, fmt.Errorf("listing operatorgroups in namespace %s: %s", ownNamespace, err)
+	}
+
+	if len(ogs) != 1 || ogs[0] == nil {
+		// Problem with the operatorgroup configuration in the namespace
+		// The namespace may not be configured as an operator installation target namespace
+		// Assume global resolution (default)
+		return defaultResult, nil
```
benluddy, line 90: This could give you flaky behavior in cases where users are simultaneously creating an OperatorGroup with Subscriptions -- if the catalog-operator sees one of the Subscriptions before it sees the OperatorGroup, it could end up doing the wrong thing. Better off returning an error here and letting it try again during the namespace resync triggered by seeing the OperatorGroup.

pkg/controller/operators/catalog/operator.go
```diff
@@ -1085,6 +1099,20 @@ func (o *Operator) syncSubscriptions(obj interface{}) error {
 	return nil
 }
 
+// syncOperatorGroups requeues the namespace resolution queue on changes to an operatorgroup
+// This is because the operatorgroup is now an input to resolution via the global catalog exclusion annotation
```
benluddy, line 1103: :+1:

pkg/controller/operators/catalog/og_source_provider.go
exdx: I agree there's a race condition, but would returning an error here change the default behavior of OLM? Where now creating a subscription in a namespace without 1 OG in it is a resolution-time error, instead of a runtime error. Is it safe to make that change?

pkg/controller/operators/catalog/og_source_provider.go
```diff
@@ -0,0 +1,106 @@
...
+	var defaultResult = namespaces
+	// The first namespace provided is always the current namespace being synced
+	var ownNamespace = namespaces[0]
+	var toggledResult = []string{ownNamespace}
+
+	if ownNamespace == e.globalNamespace {
+		// Global namespace is being synced
+		// Return early with default
+		return defaultResult, nil
+	}
```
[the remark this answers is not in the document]
exdx: That's true -- this is a special case but we may not even care about it. Will update.

pkg/controller/operators/catalog/og_source_provider.go
benluddy: I'm not sure exactly it fails today. I do know that there has occasionally been talk about respecting supported install modes in catalog-operator decisions, and I think you can make a good argument for preventing the creation of an InstallPlan if it's known to be invalid ahead of time. In this particular situation, the (hypothetical) InstallPlan may or may not be invalid, and the user may or may not have wanted to exclude operators from the global catalogs.

Is there a realistic use case for creating an InstallPlan in a namespace that doesn't have an OperatorGroup? It would have to be pretty convincing to accept that decisions may vary when creating OGs and Subscriptions at about the same time.

exdx:
I squashed everything down to one-commit (made Per the co-author on it as well). 

The e2e test needs #2801 to go in first (that was originally part of this branch, but was broken out separately since its unrelated), then this can be rebased again, and it should be good after that.

benluddy (commented):
Do you want to make [the error text in cache.go](https://github.com/operator-framework/operator-lifecycle-manager/blob/97b64e1677721faeb3d0e5cd4add2d3298733309/pkg/controller/registry/resolver/cache/cache.go#L142) unspecific to catalogs? Maybe something like "failed to populate resolver cache from source %v: %w"?

exdx:
> Do you want to make [the error text in cache.go](https://github.com/operator-framework/operator-lifecycle-manager/blob/97b64e1677721faeb3d0e5cd4add2d3298733309/pkg/controller/registry/resolver/cache/cache.go#L142) unspecific to catalogs? Maybe something like "failed to populate resolver cache from source %v: %w"?

That sounds good -- should make the error message more clear

PR #2789, opened 2022-06-01
Update for new grpc library

perdasilva:
Updates olm to use the lates operator-registry release

It's a been a while, and the latest version contains the latest version of the grpc library with security fixes needed by the downstream. See https://github.com/operator-framework/operator-registry/issues/959

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -488,6 +488,15 @@ func (o *Operator) syncSourceState(state grpc.SourceState) {
 	metrics.RegisterCatalogSourceState(state.Key.Name, state.Key.Namespace, state.State)
 
 	switch state.State {
+	case connectivity.Idle:
+		// If the connection is IDLE attempt to connect to attempt to bring it to a READY state
+		// A connection will transition to IDLE if:
+		// "No RPC activity on channel for IDLE_TIMEOUT (default: 5 mins) or upon receiving a GOAWAY while there are no pending RPCs."
+		// https://gitlab.uni-hannover.de/tci-gateway-module/grpc/-/blob/972e31165218b49d93e5e1f1a1e8bbcd3fa830d1/doc/connectivity-semantics-and-api.md
+		sourceConn := o.sources.Get(state.Key)
+		if sourceConn != nil && sourceConn.Conn != nil {
+			_, _ = registry.NewClientFromConn(sourceConn.Conn).HealthCheck(context.Background(), time.Second*5)
+		}
 	case connectivity.Ready:
 		o.sourceInvalidator.Invalidate(resolvercache.SourceKey(state.Key))
 		if o.namespace == state.Key.Namespace {
--- a/pkg/controller/registry/grpc/source.go
+++ b/pkg/controller/registry/grpc/source.go
@@ -9,7 +9,6 @@ import (
 	"time"
 
 	"github.com/operator-framework/operator-registry/pkg/client"
-
 	"github.com/sirupsen/logrus"
 	"golang.org/x/net/http/httpproxy"
 	"golang.org/x/net/proxy"
--- a/pkg/controller/registry/grpc/source_test.go
+++ b/pkg/controller/registry/grpc/source_test.go
@@ -134,6 +134,7 @@ func TestConnectionEvents(t *testing.T) {
 				{Name: "test", Namespace: "test"}: {
 					connectivity.Connecting,
 					connectivity.Ready,
+					connectivity.Idle,
 				},
 			},
 		},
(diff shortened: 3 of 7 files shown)


perdasilva:
@anik120 could you please review this? I'm a bit worried about the metrics. This update changes the way grpc connection states change. Basically they remain on idle until there's an rpc call before they transition to ready. Any idea how this could affect SRE?

pkg/controller/operators/catalog/operator.go
```diff
@@ -474,6 +474,8 @@ func (o *Operator) syncSourceState(state grpc.SourceState) {
 	metrics.RegisterCatalogSourceState(state.Key.Name, state.Key.Namespace, state.State)
 
 	switch state.State {
+	case connectivity.Idle:
+		fallthrough
```
anik120, line 478: Why are we falling through here exactly? 
Also, could you elaborate on what `Idle` connectivity state means?

anik120 (commented):
@perdasilva could you expand a bit more on 

> This update changes the way grpc connection states change. Basically they remain on idle until there's an rpc call before they transition to ready.

From the code changes it looks like idle and ready state are treated the same now, whereas before there was a distinction between idle and ready. If that's accurate, what's the motivation for that? (It isn't very clear to me how the `grpc_health_probe` related CVEs are related to this change in how we treat the different states).

pkg/controller/operators/catalog/operator.go
perdasilva: Good questions all around. So, in turn:
We need to bump the grpc_healthprobe version to satisfy the CVE -> this leads to a grpc library version bump.
The latest grpc library seems to have affected the connection state. Before the connection state would be READY once the CatalogSource pod was up. Now it sits on IDLE until there's an RPC call down the pipe. The transition from CONNECTION -> IDLE happens when there's "No RPC activity on channel for IDLE_TIMEOUT" [doc](https://gitlab.uni-hannover.de/tci-gateway-module/grpc/-/blob/972e31165218b49d93e5e1f1a1e8bbcd3fa830d1/doc/connectivity-semantics-and-api.md).

Let me refresh myself on the code base to give you a better explanation about the `passthrough`

anik120: I see. `fallthrough` here (and in another place below) just means "if connectivity.Idle, do whatever you were supposed to do for connectivity.Ready". We probably want to confirm that that's the right thing to do to. 

i.e with this change in the library, when connectivity is in Idle, it feels like the intuitive thing to do would be for the controller to "sit idle" too, instead with "fallthrough" it's going to follow through with 1. Invalidate cache 2. Add to resolve queue...etc etc. Is that what we want? 

cc: @benluddy @njhale

anik120:
Also, we probably want to set `catalogSourceReady` to 1 in both the `Idle` and `Ready` state [here](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/metrics/metrics.go#L238-L239) too.

pkg/controller/operators/catalog/operator.go
anik120: Also, this would be considered a breaking change for our consumers, so either we find a way to set the CatalogSource `.status.GRPCConnectionState.LastObservedState` to `Ready` for both `Idle` and `Ready`, or we need to send out communications to our community to alert them about the change.

perdasilva: I spent yesterday afternoon poking around grpc and playing with [keepalive](https://github.com/grpc/grpc-go/blob/master/Documentation/keepalive.md) and changing both server and client settings. I couldn't really get it to return to the previous behavior. I think leaving it on IDLE could have the adverse effect of only finding out the CatalogSource is down when you actually try to talk to it. I removed the IDLE handling code above and did a test with a customer CatalogSource created in the default namespaces. I observed that the connection state stays on IDLE even if you kill the backing pod. Presumably because the content hadn't been cached yet (i.e. no gRPC calls). From the cluster administration perspective, I'm not sure that this is desirable. While we reduce network use and CatalogSource server resources improving scalability. On the other hand, ops would only get notified of a bad CatalogSource at the time of use. 

Another alternative, but it's a bit long winded:
Decouple the connection state from the registry server health (service can be up, but broken) and have metrics for connection state and registry server health implement the grpc health checks in the `RegistryReconciler`s ([exp branch](https://github.com/operator-framework/operator-lifecycle-manager/pull/2797))

For now, I've updated the code to create a connection in case of the IDLE state. This brings the behavior back inline with what we had before.

dinhxuanvu (commented):
/lgtm

perdasilva:
closing this as it's stale

PR #2782, opened 2022-05-23
Update CatalogSource Pod security context

perdasilva:
Signed-off-by: perdasilva <perdasilva@redhat.com>

**Description of the change:**
Updates the security context stanzas for the backing CatalogSource pod and container to be more explicit and limited

This PR also includes:
 - A migration of all e2e test catalogs to FBC (needed to get the tests to pass)
 - Moves any bundle image stored in a personal quay.io account to the olmtest account
 - Updates `scripts/build_test_images.sh` to build and publish all e2e test catalogs
 - An update to the ci artifact collection script:
   - simplifies the `get` and `describe` calls to operate against all resources in the namespace
   - collects the logs of any pod left in the namespace

**Motivation for the change:**
https://bugzilla.redhat.com/show_bug.cgi?id=2088541

**Reviewer Checklist**
- [ ] Implementation matches the proposed design, or proposal is updated to match implementation
- [ ] Sufficient unit test coverage
- [ ] Sufficient end-to-end test coverage
- [ ] Docs updated or added to `/doc`
- [ ] Commit messages sensible and descriptive
- [ ] Tests marked as `[FLAKE]` are truly flaky
- [ ] Tests that remove the `[FLAKE]` tag are no longer flaky

PR #2793, opened 2022-06-08
Update unpack job pod security

perdasilva:
Updates the security context stanzas for the bundle unpacking job's pod and containers to be more explicit and limited

https://bugzilla.redhat.com/show_bug.cgi?id=2088541

**Architectural changes:**
I've added a `security` package with a function to update a pod spec with the `standard` security settings and refactored the CatalogSource pod creation function to use it as well

**Testing remarks:**
I've updated the unit tests for the bundle unpacking job creation to pass. This also ensures that the correct security settings are in the job pod spec. I've also run the e2e tests and they passed.

**NOTE**: This PR is ready for review, though it will stay blocked until #2782 is merged - as this is a continuation of that PR

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/bundle/bundle_unpacker.go
+++ b/pkg/controller/bundle/bundle_unpacker.go
@@ -7,7 +7,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/security"
 	"github.com/operator-framework/operator-registry/pkg/api"
 	"github.com/operator-framework/operator-registry/pkg/configmap"
 	"github.com/sirupsen/logrus"
@@ -29,6 +28,7 @@ import (
 	listersoperatorsv1alpha1 "github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/listers/operators/v1alpha1"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/install"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/registry/resolver/projection"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/security"
 )
 
 const (
--- a/pkg/controller/bundle/bundle_unpacker.go
+++ b/pkg/controller/bundle/bundle_unpacker.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/security"
 	"github.com/operator-framework/operator-registry/pkg/api"
 	"github.com/operator-framework/operator-registry/pkg/configmap"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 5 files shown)


pkg/controller/bundle/bundle_unpacker.go
```diff
@@ -7,6 +7,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/security"
```
timflannagan, line 10: nit: move this to the other OLM package imports that are grouped together.

pkg/controller/security/security.go
```diff
@@ -0,0 +1,37 @@
+package security
```
timflannagan, line 1: nit: Do these need to be defined as global package variables given they're unexported?

pkg/controller/registry/reconciler/reconciler.go
```diff
@@ -6,6 +6,7 @@ import (
 	"hash/fnv"
 	"strings"
 
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/security"
```
timflannagan, line 9: nit: same comment as above.

pkg/controller/security/security.go
```diff
@@ -0,0 +1,37 @@
...
+	ReadOnlyRootFilesystem:   &readOnlyRootFilesystem,
+	AllowPrivilegeEscalation: &allowPrivilegeEscalation,
+	Capabilities: &corev1.Capabilities{
+		Drop: []corev1.Capability{"ALL"},
+	},
+}
+
+var podSecurityContext = &corev1.PodSecurityContext{
+	RunAsNonRoot: &runAsNonRoot,
+	RunAsUser:    &runAsUser,
```
timflannagan, line 22: There's a k8s.io/utils/pointer package that can be used here too.

pkg/controller/security/security.go
```diff
@@ -0,0 +1,37 @@
...
+import corev1 "k8s.io/api/core/v1"
+
+var readOnlyRootFilesystem = false
+var allowPrivilegeEscalation = false
+var runAsNonRoot = true
+
+// See: https://github.com/operator-framework/operator-registry/blob/master/Dockerfile#L27
+var runAsUser int64 = 1001
+
+var containerSecurityContext = &corev1.SecurityContext{
```
timflannagan, line 12: Do we need to specify the "privileged" field here too, or is that the default value and we don't need to be explicit here?

timflannagan (approved):
I have a couple of questions/nits/comments but nothing that seemed blocking.

pkg/controller/bundle/bundle_unpacker.go
perdasilva: are there some gofmt settings I can use to to automagically sort them in the required way?

pkg/controller/security/security.go
perdasilva: privileged does indeed default to false. But I have made it explicit, just in case. If I'm reading the [docs](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-container) correctly, the container should inherit the seccomp profile from the pod security context. So, I think we should be good. All good call outs!

pkg/controller/bundle/bundle_unpacker.go
timflannagan: I think goimportas actually handles this, which is why I opened #2798 when I was initially reviewing this PR. That linter configuration is already present in rukpak/deppy, and has helped cut down on the back-and-forth's for style-related PR review comments.

timflannagan:
Holding in case we need more reviewers to take a look at these changes.

/lgtm
/hold

PR #2809, opened 2022-07-01
Only update namespaces when OperatorGroup labels need to change.

benluddy:
Even when the update request is a semantic no-op, it generates a
little apiserver and etcd load. Also, other controllers are watching
namespaces and will perform redundant work in response to the
generated watch event and resourceVersion bump.


pkg/controller/operators/olm/operator.go
```diff
@@ -902,36 +902,58 @@ func (a *Operator) syncNamespace(obj interface{}) error {
...
+			if ve, ok := namespace.Labels[kd]; !ok || ve != vd {
+				return true
+			}
+		}
+		return false
+	}(); !changed {
+		return nil
+	}
+
+	namespace = namespace.DeepCopy()
```
benluddy, line 944: This should also fix informer cache corruption.

pkg/controller/operators/olm/operator.go
```diff
@@ -902,36 +902,58 @@ func (a *Operator) syncNamespace(obj interface{}) error {
...
+		if operatorsv1.IsOperatorGroupLabel(k) {
+			delete(namespace.Labels, k)
+		}
+	}
+	if namespace.Labels == nil && len(desiredGroupLabels) > 0 {
+		namespace.Labels = make(map[string]string)
+	}
+	for k, v := range desiredGroupLabels {
+		namespace.Labels[k] = v
 	}
```
awgreene, line 955: Is there a reason you decided to separate:
- Checking if anything has changed
- Setting the correct labels

awgreene: Thanks!

awgreene (commented):
/lgtm

Thanks @benluddy!

PR #2811, opened 2022-07-05
Introduce protectedCopiedCSVNamespaces flag

awgreene:

Introduce a runtime flag that prevents OLM from deleting Copied CSVs in a given set of namespaces.

**Problem:** Users rely on Copied CSVs in order to understand which
operators are available in a given namespace. When installing All
Namespace operators, a Copied CSV is created in every namespace which
can place a huge performance strain on clusters with many namespaces.
OLM introduced the ability to disable Copied CSVs for All Namespace mode
operators  in an effort to resolve the performance issues on large clusters,
unfortunately removing the ability for users to identify which operators are
available in a given namespace.

**Solution:** The protectedCopiedCSVNamespaces runtime flag can be used to
prevent Copied CSVs from being deleted even when Copied CSVs are
disabled. An admin can then provide users with the proper RBAC to view
which operators are running in All Namespace mode.
**Architectural changes:**

**Testing remarks:**

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/olm/config.go
+++ b/pkg/controller/operators/olm/config.go
@@ -1,6 +1,7 @@
 package olm
 
 import (
+	"strings"
 	"time"
 
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/queueinformer"
@@ -21,18 +22,19 @@ import (
 type OperatorOption func(*operatorConfig)
 
 type operatorConfig struct {
-	resyncPeriod      func() time.Duration
-	operatorNamespace string
-	watchedNamespaces []string
-	clock             utilclock.Clock
-	logger            *logrus.Logger
-	operatorClient    operatorclient.ClientInterface
-	externalClient    versioned.Interface
-	strategyResolver  install.StrategyResolverInterface
-	apiReconciler     APIIntersectionReconciler
-	apiLabeler        labeler.Labeler
-	restConfig        *rest.Config
-	configClient      configv1client.Interface
+	protectedCopiedCSVNamespaces map[string]struct{}
+	resyncPeriod                 func() time.Duration
+	operatorNamespace            string
+	watchedNamespaces            []string
+	clock                        utilclock.Clock
+	logger                       *logrus.Logger
+	operatorClient               operatorclient.ClientInterface
+	externalClient               versioned.Interface
+	strategyResolver             install.StrategyResolverInterface
+	apiReconciler                APIIntersectionReconciler
+	apiLabeler                   labeler.Labeler
+	restConfig                   *rest.Config
+	configClient                 configv1client.Interface
 }
 
 func (o *operatorConfig) apply(options []OperatorOption) {
@@ -77,14 +79,15 @@ func (o *operatorConfig) validate() (err error) {
 
 func defaultOperatorConfig() *operatorConfig {
 	return &operatorConfig{
-		resyncPeriod:      queueinformer.ResyncWithJitter(30*time.Second, 0.2),
-		operatorNamespace: "default",
-		watchedNamespaces: []string{metav1.NamespaceAll},
-		clock:             utilclock.RealClock{},
-		logger:            logrus.New(),
-		strategyResolver:  &install.StrategyResolver{},
-		apiReconciler:     APIIntersectionReconcileFunc(ReconcileAPIIntersection),
-		apiLabeler:        labeler.Func(LabelSetsFor),
+		resyncPeriod:                 queueinformer.ResyncWithJitter(30*time.Second, 0.2),
+		operatorNamespace:            "default",
+		watchedNamespaces:            []string{metav1.NamespaceAll},
+		clock:                        utilclock.RealClock{},
+		logger:                       logrus.New(),
+		strategyResolver:             &install.StrategyResolver{},
+		apiReconciler:                APIIntersectionReconcileFunc(ReconcileAPIIntersection),
+		apiLabeler:                   labeler.Func(LabelSetsFor),
+		protectedCopiedCSVNamespaces: map[string]struct{}{},
 	}
 }
 
@@ -109,6 +112,18 @@ func WithWatchedNamespaces(namespaces ...string) OperatorOption {
 func WithLogger(logger *logrus.Logger) OperatorOption {
 	return func(config *operatorConfig) {
 		config.logger = logger
+	}
+}
+
+func WithProtectedCopiedCSVNamespaces(namespaces string) OperatorOption {
+	return func(config *operatorConfig) {
+		if namespaces == "" {
+			return
+		}
+
+		for _, ns := range strings.Split(namespaces, ",") {
+			config.protectedCopiedCSVNamespaces[ns] = struct{}{}
+		}
 	}
 }
 
--- a/pkg/controller/operators/olm/operator.go
+++ b/pkg/controller/operators/olm/operator.go
@@ -63,32 +63,33 @@ var (
 type Operator struct {
 	queueinformer.Operator
 
-	clock                 utilclock.Clock
-	logger                *logrus.Logger
-	opClient              operatorclient.ClientInterface
-	client                versioned.Interface
-	lister                operatorlister.OperatorLister
-	copiedCSVLister       operatorsv1alpha1listers.ClusterServiceVersionLister
-	ogQueueSet            *queueinformer.ResourceQueueSet
-	csvQueueSet           *queueinformer.ResourceQueueSet
-	olmConfigQueue        workqueue.RateLimitingInterface
-	csvCopyQueueSet       *queueinformer.ResourceQueueSet
-	copiedCSVGCQueueSet   *queueinformer.ResourceQueueSet
-	objGCQueueSet         *queueinformer.ResourceQueueSet
-	nsQueueSet            workqueue.RateLimitingInterface
-	apiServiceQueue       workqueue.RateLimitingInterface
-	csvIndexers           map[string]cache.Indexer
-	recorder              record.EventRecorder
-	resolver              install.StrategyResolverInterface
-	apiReconciler         APIIntersectionReconciler
-	apiLabeler            labeler.Labeler
-	csvSetGenerator       csvutility.SetGenerator
-	csvReplaceFinder      csvutility.ReplaceFinder
-	csvNotification       csvutility.WatchNotification
-	serviceAccountSyncer  *scoped.UserDefinedServiceAccountSyncer
-	clientAttenuator      *scoped.ClientAttenuator
-	serviceAccountQuerier *scoped.UserDefinedServiceAccountQuerier
-	clientFactory         clients.Factory
+	clock                        utilclock.Clock
+	logger                       *logrus.Logger
+	opClient                     operatorclient.ClientInterface
+	client                       versioned.Interface
+	lister                       operatorlister.OperatorLister
+	protectedCopiedCSVNamespaces map[string]struct{}
+	copiedCSVLister              operatorsv1alpha1listers.ClusterServiceVersionLister
+	ogQueueSet                   *queueinformer.ResourceQueueSet
+	csvQueueSet                  *queueinformer.ResourceQueueSet
+	olmConfigQueue               workqueue.RateLimitingInterface
+	csvCopyQueueSet              *queueinformer.ResourceQueueSet
+	copiedCSVGCQueueSet          *queueinformer.ResourceQueueSet
+	objGCQueueSet                *queueinformer.ResourceQueueSet
+	nsQueueSet                   workqueue.RateLimitingInterface
+	apiServiceQueue              workqueue.RateLimitingInterface
+	csvIndexers                  map[string]cache.Indexer
+	recorder                     record.EventRecorder
+	resolver                     install.StrategyResolverInterface
+	apiReconciler                APIIntersectionReconciler
+	apiLabeler                   labeler.Labeler
+	csvSetGenerator              csvutility.SetGenerator
+	csvReplaceFinder             csvutility.ReplaceFinder
+	csvNotification              csvutility.WatchNotification
+	serviceAccountSyncer         *scoped.UserDefinedServiceAccountSyncer
+	clientAttenuator             *scoped.ClientAttenuator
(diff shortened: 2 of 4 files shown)


timflannagan (commented):
I think I'm generally onboard with these changes, although I would prefer an API-based approach (e.g. OLMConfig?) but it's unclear whether that's overkill right now.

Do we need to accompany these changes with a test in some capacity? Do we need any upstream documentation changes here too?

awgreene:
>I think I'm generally onboard with these changes, although I would prefer an API-based approach (e.g. OLMConfig?) but it's unclear whether that's overkill right now.

I agree that it would have been nice to introduce this via the OLMConfig resource, but:
- This is a workaround to support a very specific use case.
- Ideally, most users will never touch this this feature.
- Introducing it through the olmConfig API would encourage customers to use this feature.
- There is no reason it can't be added to the olmConfig api in the future if requested.

>Do we need to accompany these changes with a test in some capacity? Do we need any upstream documentation changes here too?

I've updated existing tests to account for this feature, but it's only tested if the operator is ran with the protectedCopiedCSVNamespaces flag.

pkg/controller/operators/olm/config.go
```diff
@@ -112,6 +114,19 @@ func WithLogger(logger *logrus.Logger) OperatorOption {
 	}
 }
 
+func WithProtectedCopiedCSVNamespaces(namespaces *string) OperatorOption {
+	return func(config *operatorConfig) {
+		config.protectedCopiedCSVNamespaces = map[string]struct{}{}
+		if namespaces == nil {
+			return
+		}
```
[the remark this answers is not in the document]
awgreene: Sure, we can avoid the nil check by changing the namespace parameter from `*string` to `string`.

timflannagan:
> Begin Captured GinkgoWriter Output >>
    [BeforeEach] Disabling copied CSVs
      /home/runner/work/operator-lifecycle-manager/operator-lifecycle-manager/test/e2e/disabling_copied_csv_e2e_test.go:32
    created the csv-toggle-test-sldbr testing namespace
    [BeforeEach] Copied CSVs are disabled
      /home/runner/work/operator-lifecycle-manager/operator-lifecycle-manager/test/e2e/disabling_copied_csv_e2e_test.go:101
    [It] should be reflected in the olmConfig.Status.Condition array that the expected number of copied CSVs exist
      /home/runner/work/operator-lifecycle-manager/operator-lifecycle-manager/test/e2e/disabling_copied_csv_e2e_test.go:168
    [AfterEach] Disabling copied CSVs
      /home/runner/work/operator-lifecycle-manager/operator-lifecycle-manager/test/e2e/disabling_copied_csv_e2e_test.go:61
    collecting the csv-toggle-test-sldbr namespace artifacts as the 'should be reflected in the olmConfig.Status.Condition array that the expected number of copied CSVs exist' test case failed
    collecting logs in the ./artifacts/ artifacts directory
    tearing down the csv-toggle-test-sldbr namespace
    resetting e2e kube client
    deleting csv-toggle-test-sldbr/csv-toggle-test-sldbr-operatorgroup
    deleting <global>/csv-toggle-test-sldbr
    garbage collecting CRDs
<< End Captured GinkgoWriter Output
Timed out after 60.001s.
      <*errors.errorString | 0xc0003f9a20>: {
          s: "condition does not have expected reason, message, and status. Expected { True 0 0001-01-01 00:00:00 +0000 UTC CopiedCSVsDisabled Copied CSVs are disabled and no unexpected copied CSVs were found for operators installed in AllNamespace mode}, got &Condition{Type:DisabledCopiedCSVs,Status:False,ObservedGeneration:0,LastTransitionTime:2022-07-20 14:14:10 +0000 UTC,Reason:CopiedCSVsDisabled,Message:Copied CSVs are disabled and at least one unexpected copied CSV was found for an operator installed in AllNamespace mode,}",
      }
      condition does not have expected reason, message, and status. Expected { True 0 0001-01-01 00:00:00 +0000 UTC CopiedCSVsDisabled Copied CSVs are disabled and no unexpected copied CSVs were found for operators installed in AllNamespace mode}, got &Condition{Type:DisabledCopiedCSVs,Status:False,ObservedGeneration:0,LastTransitionTime:2022-07-20 14:14:10 +0000 UTC,Reason:CopiedCSVsDisabled,Message:Copied CSVs are disabled and at least one unexpected copied CSV was found for an operator installed in AllNamespace mode,}
In [It] at: /home/runner/work/operator-lifecycle-manager/operator-lifecycle-manager/test/e2e/disabling_copied_csv_e2e_test.go:193

@awgreene That seems like a legitimate e2e failure?

pkg/controller/operators/olm/operator.go
```diff
@@ -1444,7 +1454,25 @@ func (a *Operator) syncCopyCSV(obj interface{}) (syncError error) {
...
+	for ns := range a.protectedCopiedCSVNamespaces {
+		if ns == clusterServiceVersion.GetNamespace() {
+			continue
+		}
+		protectedNamespaces = append(protectedNamespaces, ns)
+	}
+
+	if err := a.ensureCSVsInNamespaces(clusterServiceVersion, operatorGroup, NewNamespaceSet(protectedNamespaces)); err != nil {
+		logger.WithError(err).Info("couldn't copy CSV to protected Copied CSV namespaces")
+		syncError = err
```
timflannagan, line 1468: Should we add a comment for why we're falling through here vs. returning early?

pkg/controller/operators/olm/operator.go
```diff
@@ -1361,15 +1372,14 @@ func getCopiedCSVsCondition(isDisabled, csvIsRequeued bool) metav1.Condition {
 		return condition
 	}
 
+	condition.Reason = "CopiedCSVsDisabled"
```
timflannagan, line 1375: Does this need to live as a constant? Where are the reasons typically defined, o-f/api?

pkg/controller/operators/olm/operator.go
```diff
@@ -1299,20 +1301,29 @@ func (a *Operator) syncOLMConfig(obj interface{}) (syncError error) {
...
+		if err != nil {
+			return err
+		}
+
 		for _, csv := range csvs {
-			// If the correct number of copied CSVs were found, continue
-			if _, ok := uniqueCopiedCSVs[csv.GetName()]; ok == olmConfig.CopiedCSVsAreEnabled() {
+			numberOfCopiedCSVs := copiedCSVCount[csv.GetName()]
+			// Ignore NS where actual CSV is installed
+			if olmConfig.CopiedCSVsAreEnabled() && numberOfCopiedCSVs == len(namespaces)-1 {
```
timflannagan, line 1322: Store the value of olmConfig.CopiedCSVsAreEnabled() before iterating over the list of csvs so we don't reinitialize per loop iteration?

timflannagan: Also: do we have a more robust way to identify whether the CSV we're processing actually lives in the namespace vs. making that assumption by comparing the number of copied CSVs we already calculated to the number of namespaces we expect?

test/e2e/disabling_copied_csv_e2e_test.go
```diff
@@ -139,8 +145,21 @@ var _ = Describe("Disabling copied CSVs", func() {
 					return err
 				}
 
-				if numCSVs := len(copiedCSVs.Items); numCSVs != 0 {
-					return fmt.Errorf("Found %d copied CSVs, should be 0", numCSVs)
+				if numCSVs := len(copiedCSVs.Items); numCSVs != len(protectedCopiedCSVNamespaces) {
```
tylerslaton, line 148: nit: This reads a bit better to me personally, but I could be convinced to the other format. That being said, I think this is a bit clearer while reading.

```suggestion
				if len(copiedCSVs.Items) != len(protectedCopiedCSVNamespaces) {
```

pkg/controller/operators/olm/operator.go
awgreene: >Also: do we have a more robust way to identify whether the CSV we're processing actually lives in the namespace vs. making that assumption by comparing the number of copied CSVs we already calculated to the number of namespaces we expect?

Yeah, that's probably a nicer implementation, updated.

awgreene: > Store the value of olmConfig.CopiedCSVsAreEnabled() before iterating over the list of csvs so we don't reinitialize per loop iteration?

Good idea

awgreene: This should all be resolved now.

test/e2e/disabling_copied_csv_e2e_test.go
awgreene: I'm fine, I wasn't a fan of multiple `len` calls but this isn't an issue with the revised implementation.

awgreene:
>@awgreene That seems like a legitimate e2e failure?

@timflannagan this was a legitimate error where namespaces in a terminated state were included in the expected copied CSV count, fixed.

pkg/controller/operators/olm/operator.go
awgreene: This should have been returned as an error, addressing.

pkg/controller/operators/olm/operator.go
awgreene: That's probably a good idea, any objections to doing this in a followup PR if I create an issue to track it?

pkg/controller/operators/olm/config.go
```diff
@@ -112,6 +115,18 @@ func WithLogger(logger *logrus.Logger) OperatorOption {
 	}
 }
 
+func WithProtectedCopiedCSVNamespaces(namespaces string) OperatorOption {
+	return func(config *operatorConfig) {
+		if namespaces == "" {
+			return
+		}
+
+		for _, ns := range strings.Split(namespaces, ",") {
```
tylerslaton, line 124: nit: Should this just take a slice of strings that it uses instead of needing to split a single string?

pkg/controller/operators/olm/operator.go
```diff
@@ -1299,20 +1301,56 @@ func (a *Operator) syncOLMConfig(obj interface{}) (syncError error) {
...
 
-		// Filter to unique copies
-		uniqueCopiedCSVs := map[string]struct{}{}
+		copiedCSVNamespaces := map[string]map[string]struct{}{}
 		for _, copiedCSV := range copiedCSVs {
-			uniqueCopiedCSVs[copiedCSV.GetName()] = struct{}{}
+			if _, ok := copiedCSVNamespaces[copiedCSV.GetName()]; !ok {
+				copiedCSVNamespaces[copiedCSV.GetName()] = map[string]struct{}{}
+			}
+			copiedCSVNamespaces[copiedCSV.GetName()][copiedCSV.GetNamespace()] = struct{}{}
```
tylerslaton, line 1309: nit: This gets a little hard to read due to the nature of the data-structures and looping we're doing here. A comment explaining why we're doing this would be a nice addition for future contributors.

tylerslaton (approved):
Couple of non-blocking comments that could be addressed later if it becomes essential. 

/lgtm

test/e2e/disabling_copied_csv_e2e_test.go
```diff
@@ -159,8 +178,8 @@ var _ = Describe("Disabling copied CSVs", func() {
 				}
 
 				expectedCondition := metav1.Condition{
-					Reason:  "NoCopiedCSVsFound",
-					Message: "Copied CSVs are disabled and none were found for operators installed in AllNamespace mode",
+					Reason:  "CopiedCSVsDisabled",
```
timflannagan, line 181: Yep, this is why I'd like to see a follow-up that moves this reason to a constant variable that can be shared around.

test/e2e/disabling_copied_csv_e2e_test.go
```diff
@@ -260,3 +279,24 @@ var _ = Describe("Disabling copied CSVs", func() {
 		})
 	})
 })
+
+func getProtectedCopiedCSVNamespaces(protectedCopiedCSVNamespaces map[string]struct{}) error {
+	var olmDeployment appsv1.Deployment
+	if err := ctx.Ctx().Client().Get(context.TODO(), apitypes.NamespacedName{Name: "olm-operator", Namespace: operatorNamespace}, &olmDeployment); err != nil {
```
timflannagan, line 285: nit/not blocking: avoid hardcoding the context and namespaced name fields and pass those as parameters to this function instead?

timflannagan (approved):

pkg/controller/operators/olm/operator.go
timflannagan: Just saw this: that's fine with me, but I'd like to see us tackle this sooner than later.

pkg/controller/operators/olm/config.go
awgreene: Good question.

The flag, which is used to configure OLM, is passed in as a string, so I opted to convert that string into an array within the function.

awgreene: I still believe that it makes sense to live here.

pkg/controller/operators/olm/operator.go
awgreene: Agreed.

pkg/controller/operators/olm/operator.go
```diff
@@ -1299,20 +1301,56 @@ func (a *Operator) syncOLMConfig(obj interface{}) (syncError error) {
 			return err
 		}
 
-		// Filter to unique copies
-		uniqueCopiedCSVs := map[string]struct{}{}
+		copiedCSVNamespaces := map[string]map[string]struct{}{}
```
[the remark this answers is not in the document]
awgreene: That's correct, I'll add a comment in the next update.

pkg/controller/operators/olm/operator.go
```diff
@@ -1299,20 +1301,56 @@ func (a *Operator) syncOLMConfig(obj interface{}) (syncError error) {
...
+
+				for protectedNS := range a.protectedCopiedCSVNamespaces {
+					if _, ok := m[protectedNS]; !ok {
+						return false
+					}
+				}
+
+				return true
+			}
+		}
```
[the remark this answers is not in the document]
awgreene: Will be in the next update.

timflannagan:
Nicely done. Feel free to remove the hold.

/lgtm

PR #2813, opened 2022-07-07
Return copy of annotations in grpcCatalogSourceDecorator

perdasilva:

Recent changes adding the grpcCatalogSourceDecorator introduced a concurrent write issue on the annotations. This PR updates the Annotations method to return a copy of the annotations to avoid this issue.

https://bugzilla.redhat.com/show_bug.cgi?id=2101357

**Architectural changes:**
None

**Testing remarks:**
Deployed on kind and not seeing the issue

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/registry/reconciler/reconciler_test.go
+++ b/pkg/controller/registry/reconciler/reconciler_test.go
@@ -112,7 +112,7 @@ func TestPodAvoidsConcurrentWrite(t *testing.T) {
 	}
 
 	annotations := map[string]string{
-		"annotation": "somethignelse",
+		"annotation": "somethingelse",
 	}
 
 	gotPod := Pod(catsrc, "hello", "busybox", "", labels, annotations, int32(0), int32(0))


awgreene (approved):
Alternatively, we could generate the list of annotations once when creating the grpcCatalogSourceDecorator, but I don't think this will introduce a huge performance impact.

timflannagan (approved):
+1 on Alex's suggestion as well. Looking through the rest of these methods, it looks like we wouldn't have the same issue as labels. Do we need a test here?

perdasilva:
> +1 on Alex's suggestion as well. Looking through the rest of these methods, it looks like we wouldn't have the same issue as labels. Do we need a test here?

I was trying to think how to add one, but wasn't sure the best way to test it. I'm also wondering if we should move the logic to the reconciler.

awgreene:
I think this is a small enough change that I'm comfortable merging as is.

timflannagan:
Yep, same here, and it's definitely helpful that you linked to the BZ so we're tracking some historical context over time.

pkg/controller/registry/reconciler/reconciler_test.go
```diff
@@ -95,6 +95,37 @@ func TestPodContainerSecurityContext(t *testing.T) {
...
+			Namespace: "testns",
+		},
+	}
+
+	labels := map[string]string{
+		"label": "something",
+	}
+
+	annotations := map[string]string{
+		"annotation": "somethignelse",
```
awgreene, line 115:
```suggestion
		"annotation": "somethingelse",
```

awgreene: I also prefer camelCase.

awgreene (changes requested):
Found one spelling mistake, otherwise looks good.

perdasilva:
> Found one spelling mistake, otherwise looks good.

Fixed!! good eyes =D

awgreene (commented):
Another option is to add a comment stating that the `reconcile.Pod function` will mutate the labels/annotations provided and updating all calls to provide copies, but making the function copy the arrays is probably safer for future developers that might rely on this method.

/lgtm

awgreene (approved):

PR #2817, opened 2022-07-15
fix(openshift): use env var instead of clusterversion status

tylerslaton:
Introducing some code that Nick wrote a fix for prior to him leaving Red Hat. Here are his comments:

Get the current OpenShift release version from the OPENSHIFT_RELEASE
environment variable since the behavior of the original source --
the ClusterVersion desired release status field -- has changed.

> **Note**: This is for a high priority bug fix downstream.

**Reviewer Checklist**
- [X] Implementation matches the proposed design, or proposal is updated to match implementation
- [X] Sufficient unit test coverage
- [X] Sufficient end-to-end test coverage
- [X] Bug fixes are accompanied by regression test(s)
- [X] e2e tests and flake fixes are accompanied evidence of flake testing, e.g. executing the test 100(0) times
- [X] tech debt/todo is accompanied by issue link(s) in comments in the surrounding code
- [X] Tests are comprehensible, e.g. Ginkgo DSL is being used appropriately
- [X] Docs updated or added to `/doc`
- [X] Commit messages sensible and descriptive
- [X] Tests marked as `[FLAKE]` are truly flaky and have an issue
- [X] Code is properly formatted

### DIFF
--- a/pkg/controller/operators/openshift/clusteroperator_controller_test.go
+++ b/pkg/controller/operators/openshift/clusteroperator_controller_test.go
@@ -2,6 +2,7 @@ package openshift
 
 import (
 	"fmt"
+	"os"
 
 	semver "github.com/blang/semver/v4"
 	. "github.com/onsi/ginkgo/v2"
@@ -19,7 +20,6 @@ import (
 var _ = Describe("ClusterOperator controller", func() {
 	var (
 		clusterOperatorName types.NamespacedName
-		cv                  *configv1.ClusterVersion
 		csv                 *operatorsv1alpha1.ClusterServiceVersion
 	)
 
@@ -55,33 +55,11 @@ var _ = Describe("ClusterOperator controller", func() {
 	})
 
 	BeforeEach(func() {
-		// "version" singleton is available in OpenShift by default
-		cv = &configv1.ClusterVersion{}
-		cv.SetName("version")
-
-		Eventually(func() error {
-			return k8sClient.Create(ctx, cv)
-		}).Should(Succeed())
-
-		cv.Status = configv1.ClusterVersionStatus{
-			Desired: configv1.Release{
-				Version: clusterVersion,
-			},
-		}
-
-		Eventually(func() error {
-			return k8sClient.Status().Update(ctx, cv)
-		}).Should(Succeed())
+		os.Setenv(releaseEnvVar, clusterVersion)
 	})
 
 	AfterEach(func() {
-		Eventually(func() error {
-			err := k8sClient.Delete(ctx, cv)
-			if err != nil && apierrors.IsNotFound(err) {
-				err = nil
-			}
-			return err
-		}).Should(Succeed())
+		resetCurrentReleaseTo(clusterVersion)
 	})
 
 	It("should ensure the ClusterOperator always exists", func() {
@@ -143,10 +121,14 @@ var _ = Describe("ClusterOperator controller", func() {
 		}))
 
 		By("setting upgradeable=false when there's an error determining compatibility")
-		cv.Status = configv1.ClusterVersionStatus{}
-
+		// Reset the ClusterOperator with an invalid version set
+		Expect(resetCurrentReleaseTo("")).To(Succeed())
 		Eventually(func() error {
-			return k8sClient.Status().Update(ctx, cv)
+			err := k8sClient.Delete(ctx, co)
+			if err != nil && !apierrors.IsNotFound(err) {
+				return err
+			}
+			return nil
 		}).Should(Succeed())
 
 		Eventually(func() ([]configv1.ClusterOperatorStatusCondition, error) {
@@ -156,18 +138,18 @@ var _ = Describe("ClusterOperator controller", func() {
 			Type:               configv1.OperatorUpgradeable,
 			Status:             configv1.ConditionFalse,
 			Reason:             ErrorCheckingOperatorCompatibility,
-			Message:            "Encountered errors while checking compatibility with the next minor version of OpenShift: desired release version missing from ClusterVersion",
+			Message:            fmt.Sprintf("Encountered errors while checking compatibility with the next minor version of OpenShift: desired release version missing from %v env variable", releaseEnvVar),
 			LastTransitionTime: fixedNow(),
 		}))
 
-		cv.Status = configv1.ClusterVersionStatus{
-			Desired: configv1.Release{
-				Version: clusterVersion,
-			},
-		}
-
+		// Reset the ClusterOperator with a valid version set
+		Expect(resetCurrentReleaseTo(clusterVersion)).To(Succeed())
 		Eventually(func() error {
-			return k8sClient.Status().Update(ctx, cv)
+			err := k8sClient.Delete(ctx, co)
+			if err != nil && !apierrors.IsNotFound(err) {
+				return err
+			}
+			return nil
 		}).Should(Succeed())
 
 		By("setting upgradeable=false when incompatible operators exist")
@@ -327,3 +309,14 @@ var _ = Describe("ClusterOperator controller", func() {
 		}))
 	})
 })
+
+// resetCurrentRelease thread safely updates the currentRelease.version and then sets the openshift release
+// env var to the desired version. WARNING: This function should only be used for testing purposes as it
+// goes around the desired logic of only setting the version of the cluster for this operator once.
+func resetCurrentReleaseTo(version string) error {
+	currentRelease.mu.Lock()
+	defer currentRelease.mu.Unlock()
+
+	currentRelease.version = nil
+	return os.Setenv(releaseEnvVar, version)
+}
--- a/pkg/controller/operators/openshift/helpers.go
+++ b/pkg/controller/operators/openshift/helpers.go
@@ -4,7 +4,9 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"os"
 	"strings"
+	"sync"
 
 	semver "github.com/blang/semver/v4"
 	configv1 "github.com/openshift/api/config/v1"
@@ -119,17 +121,17 @@ func transientErrors(err error) error {
 }
 
 func incompatibleOperators(ctx context.Context, cli client.Client) (skews, error) {
-	desired, err := desiredRelease(ctx, cli)
+	current, err := getCurrentRelease()
 	if err != nil {
 		return nil, err
 	}
 
-	if desired == nil {
+	if current == nil {
 		// Note: This shouldn't happen
 		return nil, fmt.Errorf("failed to determine current OpenShift Y-stream release")
 	}
 
-	next, err := nextY(*desired)
+	next, err := nextY(*current)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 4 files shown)


grokspawn (approved):
Looks good.  Nice work.

timflannagan:
@tylerslaton Can you investigate those unit test failures:

```log
••
------------------------------
• [FAILED] [20.053 seconds]
ClusterOperator controller
/home/runner/work/operator-lifecycle-manager/operator-lifecycle-manager/pkg/controller/operators/openshift/clusteroperator_controller_test.go:19
[It] should gate OpenShift upgrades
```

Those test failures seem relevant to these set of changes.

pkg/controller/operators/openshift/helpers.go
```diff
@@ -168,24 +170,39 @@ func incompatibleOperators(ctx context.Context, cli client.Client) (skews, error
 	return incompatible, nil
 }
 
-func desiredRelease(ctx context.Context, cli client.Client) (*semver.Version, error) {
-	cv := configv1.ClusterVersion{}
-	if err := cli.Get(ctx, client.ObjectKey{Name: "version"}, &cv); err != nil { // "version" is the name of OpenShift's ClusterVersion singleton
-		return nil, &transientError{fmt.Errorf("failed to get ClusterVersion: %w", err)}
+var (
+	openshiftRelease *semver.Version
+	releaseMutex     sync.Mutex
```
timflannagan, line 175: I think I have some reservations around a package-level mutex. What happens if multiple places are locking and unlocking this mutex?

pkg/controller/operators/openshift/helpers.go
```diff
@@ -168,24 +170,39 @@ func incompatibleOperators(ctx context.Context, cli client.Client) (skews, error
...
+)
+
+const (
+	releaseEnvVar = "OPENSHIFT_RELEASE"
```
timflannagan, line 179: Add a comment for why this constant variable is needed?

tylerslaton:
> @tylerslaton Can you investigate those unit test failures:

Yep, been doing that this morning and should have a fix out soon.

pkg/controller/operators/openshift/helpers.go
```diff
@@ -168,24 +170,39 @@ func incompatibleOperators(ctx context.Context, cli client.Client) (skews, error
 	return incompatible, nil
 }
 
-func desiredRelease(ctx context.Context, cli client.Client) (*semver.Version, error) {
-	cv := configv1.ClusterVersion{}
-	if err := cli.Get(ctx, client.ObjectKey{Name: "version"}, &cv); err != nil { // "version" is the name of OpenShift's ClusterVersion singleton
-		return nil, &transientError{fmt.Errorf("failed to get ClusterVersion: %w", err)}
+var (
+	openshiftRelease *semver.Version
```
timflannagan, line 174: Same comment as the mutex example: can we avoid using a package-level variable here, and instead determine this release semver version through function/method parameters?

timflannagan (commented):
Quick passthrough of the changes here. I'm a bit hesitant with using package-level variables here, and noted some comments inline.

IIRC, this controller watches the ClusterVersion resource: is that watch still applicable in this new implementation?

timflannagan (commented):
Forgot to publish my incomplete review before we starting talking in slack.

pkg/controller/operators/openshift/helpers.go
```diff
@@ -168,24 +170,43 @@ func incompatibleOperators(ctx context.Context, cli client.Client) (skews, error
...
+	version *semver.Version
+	mu      sync.Mutex
+}
+
+var (
+	currentRelease = &openshiftRelease{}
+)
+
+const (
+	releaseEnvVar = "OPENSHIFT_RELEASE" // OpenShift's env variable for defining the current release
```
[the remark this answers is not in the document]
tylerslaton: I haven't been able to find anything and neither did Nick before he left. I'll look around to see if I can find anything.

pkg/controller/operators/openshift/helpers_test.go
```diff
@@ -525,7 +464,7 @@ func TestIncompatibleOperators(t *testing.T) {
 			scheme := runtime.NewScheme()
 			require.NoError(t, AddToScheme(scheme))
 
-			fcli := fake.NewClientBuilder().WithScheme(scheme).WithObjects(objs...).Build()
+			fcli := fake.NewClientBuilder().WithObjects(objs...).WithScheme(scheme).Build()
```
awgreene, line 467: Why did we swap these two calls?

pkg/controller/operators/openshift/options.go
```diff
@@ -174,8 +173,7 @@ func WithOLMOperator() ReconcilerOption {
 			return !csv.IsCopied() // Keep original CSVs only
 		})
 		config.TweakBuilder = func(bldr *builder.Builder) *builder.Builder {
-			return bldr.Watches(&source.Kind{Type: &operatorsv1alpha1.ClusterServiceVersion{}}, enqueue, builder.WithPredicates(originalCSV)).
-				Watches(&source.Kind{Type: &configv1.ClusterVersion{}}, enqueue, builder.WithPredicates(watchName(&name)))
+			return bldr.Watches(&source.Kind{Type: &operatorsv1alpha1.ClusterServiceVersion{}}, enqueue, builder.WithPredicates(originalCSV))
```
awgreene, line 176: Could you explain this change?

awgreene (commented):
Overall this looks good, nice work @tylerslaton

pkg/controller/operators/openshift/helpers_test.go
tylerslaton: I must have done this on accident while working. I'll revert it to minimize the diff.

pkg/controller/operators/openshift/options.go
```diff
@@ -174,8 +173,7 @@ func WithOLMOperator() ReconcilerOption {
 			return !csv.IsCopied() // Keep original CSVs only
 		})
 		config.TweakBuilder = func(bldr *builder.Builder) *builder.Builder {
-			return bldr.Watches(&source.Kind{Type: &operatorsv1alpha1.ClusterServiceVersion{}}, enqueue, builder.WithPredicates(originalCSV)).
```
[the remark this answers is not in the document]
tylerslaton: Yep, exactly. What's more, we expect this controller to be restarted whenever a cluster upgrade is occurring which results in the `openshiftRelease` singleton being refreshed.

pkg/controller/operators/openshift/helpers.go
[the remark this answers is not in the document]
tylerslaton: @timflannagan So a couple of things here after some digging.

OpenShift's documentation around how it sets versions for ClusterOperators - https://github.com/openshift/enhancements/blame/f97876821d3bb506d28fee565271d3bebbbc682c/dev-guide/cluster-version-operator/dev/clusteroperator.md#L31-L59

Downstream's manifests setting this environment variable that should get substituted -
https://github.com/openshift/operator-framework-olm/blob/master/manifests/0000_50_olm_07-olm-operator.deployment.yaml#L79-L80

One thing to note - I found that the variable name is actually different than what Nick originally had (at least in our downstream manifests, I'm still investigating where OPENSHIFT_RELEASE came from or if it was just a placeholder)

pkg/controller/operators/openshift/clusteroperator_controller_test.go
```diff
@@ -143,10 +121,11 @@ var _ = Describe("ClusterOperator controller", func() {
 		}))
 
 		By("setting upgradeable=false when there's an error determining compatibility")
-		cv.Status = configv1.ClusterVersionStatus{}
 
+		By("changing the openshift environment variable to be invalid and restarting the ClusterOperator")
+		resetCurrentReleaseTo("")
 		Eventually(func() error {
-			return k8sClient.Status().Update(ctx, cv)
+			return k8sClient.Delete(ctx, co)
```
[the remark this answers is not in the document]
tylerslaton: Fixed 🙏🏻

pkg/controller/operators/openshift/helpers.go
```diff
@@ -168,24 +170,51 @@ func incompatibleOperators(ctx context.Context, cli client.Client) (skews, error
...
+)
+
+func getCurrentRelease() (*semver.Version, error) {
+	currentRelease.mu.Lock()
+	defer currentRelease.mu.Unlock()
+
+	if currentRelease.version != nil {
+		/*
+			If the version is already set, we don't want to set it again as the currentRelease
+			is designed to be a singleton. If a new version is set, we are making an assumption
```
perdasilva, line 193: Have we checked this assumption with the CVO team?

perdasilva:
Is it worth having an e2e test, somehow? Could we have a regression test that matches the bug and ensure this fixes it?

tylerslaton:
> Is it worth having an e2e test, somehow? Could we have a regression test that matches the bug and ensure this fixes it?

I updated the E2E suite in this PR to test using the env var instead of testing the same logic with the CV. Do you think that would be enough here?

pkg/controller/operators/openshift/helpers.go
tylerslaton: Yes! Just checked with that team.

timflannagan:
I pulled down these changes and didn't see any more "clusterversion" usage, outside of e2e test utilities, so I think we're covered on that front now.

pkg/controller/operators/openshift/helpers.go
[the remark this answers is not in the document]
timflannagan: Alright works for me. Thanks.

tylerslaton:
@awgreene @perdasilva @timflannagan How is this looking to you all? I was wondering if we could try to get this through today but let me know if anything is blocking.

pkg/controller/operators/openshift/helpers_test.go
```diff
@@ -501,7 +438,9 @@ func TestIncompatibleOperators(t *testing.T) {
 		},
 	} {
 		t.Run(tt.description, func(t *testing.T) {
-			objs := []client.Object{tt.cv.DeepCopy()}
+			objs := []client.Object{}
```
timflannagan, line 441: Any idea on why we're removing the deep copy call here?

pkg/controller/operators/openshift/helpers.go
```diff
@@ -168,24 +170,51 @@ func incompatibleOperators(ctx context.Context, cli client.Client) (skews, error
...
-	desired, err := semver.ParseTolerant(v)
+	release, err := semver.ParseTolerant(raw)
 	if err != nil {
 		return nil, fmt.Errorf("cluster version has invalid desired release version: %w", err)
 	}
 
-	return &desired, nil
+	currentRelease.version = &release
+
+	return currentRelease.version, nil
```
timflannagan, line 217: Should we be combining setter-esq functionality in the same function that's responsible for getting the current release version? Can we return the parsed release semver, and have the call site of this function set this value?

pkg/controller/operators/openshift/helpers.go
```diff
@@ -168,24 +170,51 @@ func incompatibleOperators(ctx context.Context, cli client.Client) (skews, error
...
+
+var (
+	currentRelease = &openshiftRelease{}
+)
+
+const (
+	releaseEnvVar = "RELEASE_VERSION" // OpenShift's env variable for defining the current release
+)
+
+func getCurrentRelease() (*semver.Version, error) {
```
timflannagan, line 186: nit: godoc

timflannagan (approved):

pkg/controller/operators/openshift/helpers_test.go
tylerslaton: Because we're not referencing the CV anymore and it doesn't exists in the test case (a.k.a `tt`)

pkg/controller/operators/openshift/helpers.go
tylerslaton: That's a tricky question. In my opinion, with the currentRelease being a singleton by design, attempting to get its value should return the same thing no matter when/where it is called throughout the lifecycle of the pod. As a result, the getter setting the value so that all subsequent calls to the get return the same value makes sense to me. What are your thoughts though? I could be swayed but it could be messy if we have a lot of places in code trying to set the singleton value.

timflannagan: Yeah, maybe the function just needs to be renamed but I won't block the PR from merging given we at least have a godoc that explains that function's responsibilities.

PR #2818, opened 2022-07-15
Fix default E2E_INSTALL_NS value

awgreene:
The e2e tests assume that OLM is installed in the
operator-lifecycle-manager namespace if E2E_INSTALL_NS is not set. OLM
is typically installed in the olm namespace.


timflannagan (commented):
It looks like we also have a `LOCAL_NAMESPACE` Makefile variable defined that's set to the olm namespace:

https://github.com/operator-framework/operator-lifecycle-manager/blob/83e3ebf96856643286e2e3e9438352c975923858/Makefile#L23

Do we see any skew here when running `make e2e` vs. `make e2e-local`?

awgreene:
@timflannagan my only complaint with the `e2e-local` target is that it builds the binaries and creates a new kind cluster, which slows things down when I'm trying to iterate when developing/updating a test.

awgreene:
>That said, I'm not opposed to investigating a new target or modifying e2e-local to allow users to avoid building the kind cluster and binaries.

I suppose I could just do this with the existing `e2e` target by passing in the `E2E_INSTALL_NS` variable, but it still doesn't address the question why we default to `operator-lifecycle-manager`.

timflannagan:
Yeah, I typically stay away from the e2e-local workflow in favor of using an existing kind cluster, and overriding thr E2E_INSTALL_NS variable. Maybe this workflow is something we can document? I'm not too sure why the operator-lifecycle-manager namespace was chosen.

perdasilva:
@awgreene could you please fix the conflicts - then we can get this on in =D

perdasilva:
closing PR as stale. Please re-open if it's still important.

PR #2820, opened 2022-07-22
(psa) make workloads compatible with psa:restricted profile

### LINKED ISSUE #2644: [FLAKY] cleanup csvs with bad owner operator groups
### Description

Unfortunately, I marked this as a flake without capturing the stack trace ._.
Once I get it I'll update the issue. Or if this test fails for you, please paste in the stack trace and log link if possible.

anik120:

With the introduction of [Pod Security Admission](https://kubernetes.io/docs/concepts/security/pod-security-admission/#pod-security-admission-labels-for-namespaces), the recommended best practice is to enforce the `Restricted` policy of admission (see [1] for more details).

[1] https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted

This PR
    *) Lables the olm namespace as `enforce:restricted`
    *) Labels the operators namespace as `enforce:baseline` (to allow existing CSV deployments
    without securityContext set to deploy in the namespace, which won't be possible with
    `enforce:resticted`)
    *) updates the securityContext of olm workload pods(olm-operator, catalog-operator,
    and CatalogSource registry pods) to adhere to the `Restricted` policy.
    *) updates the bundle unpacking job to create a pod that adheres to the `Restricted` policy,
    so that bundles can be unpacked in the `Restricted` namespace

Will also close #2644 as a byproduct of the need to fix test failures 

**Architectural changes:**

**Testing remarks:**

**Reviewer Checklist**

### DIFF
--- a/cmd/catalog/start.go
+++ b/cmd/catalog/start.go
@@ -25,6 +25,7 @@ type options struct {
 	tlsKeyPath           string
 	tlsCertPath          string
 	clientCAPath         string
+	setWorkloadUserID    bool
 
 	installPlanTimeout  time.Duration
 	bundleUnpackTimeout time.Duration
@@ -66,6 +67,7 @@ func newRootCmd() *cobra.Command {
 	cmd.Flags().StringVar(&o.opmImage, "opmImage", defaultOPMImage, "the image to use for unpacking bundle content with opm")
 	cmd.Flags().StringVar(&o.utilImage, "util-image", defaultUtilImage, "an image containing custom olm utilities")
 	cmd.Flags().StringVar(&o.writeStatusName, "writeStatusName", defaultOperatorName, "ClusterOperator name in which to write status, set to \"\" to disable.")
+	cmd.Flags().BoolVar(&o.setWorkloadUserID, "set-workload-user-id", false, "set user ID for all workloads (registry pods/bundle unpack jobs to default 1001")
 
 	cmd.Flags().BoolVar(&o.debug, "debug", false, "use debug log level")
 	cmd.Flags().BoolVar(&o.version, "version", false, "displays the olm version")
--- a/deploy/chart/templates/0000_50_olm_00-namespace.yaml
+++ b/deploy/chart/templates/0000_50_olm_00-namespace.yaml
@@ -2,9 +2,15 @@ apiVersion: v1
 kind: Namespace
 metadata:
   name: {{ .Values.namespace }}
+  labels: 
+    pod-security.kubernetes.io/enforce: restricted
+    pod-security.kubernetes.io/enforce-version: latest
 
 ---
 apiVersion: v1
 kind: Namespace
 metadata:
   name: {{ .Values.operator_namespace }}
+  labels:
+    pod-security.kubernetes.io/enforce: baseline
+    pod-security.kubernetes.io/enforce-version: latest
--- a/deploy/chart/templates/0000_50_olm_07-olm-operator.deployment.yaml
+++ b/deploy/chart/templates/0000_50_olm_07-olm-operator.deployment.yaml
@@ -17,6 +17,10 @@ spec:
       labels:
         app: olm-operator
     spec:
+      securityContext:
+        runAsNonRoot: true
+        seccompProfile:
+          type: RuntimeDefault
       serviceAccountName: olm-operator-serviceaccount
       {{- if or .Values.olm.tlsSecret .Values.olm.clientCASecret }}
       volumes: 
@@ -33,6 +37,10 @@ spec:
       {{- end }}
       containers:
         - name: olm-operator
+          securityContext:
+            allowPrivilegeEscalation: false
+            capabilities:
+              drop: [ "ALL" ]
           {{- if or .Values.olm.tlsSecret .Values.olm.clientCASecret }}
           volumeMounts:
           {{- end }}
--- a/deploy/chart/templates/0000_50_olm_08-catalog-operator.deployment.yaml
+++ b/deploy/chart/templates/0000_50_olm_08-catalog-operator.deployment.yaml
@@ -17,6 +17,10 @@ spec:
       labels:
         app: catalog-operator
     spec:
+      securityContext:
+        runAsNonRoot: true
+        seccompProfile:
+          type: RuntimeDefault
       serviceAccountName: olm-operator-serviceaccount
       {{- if or .Values.catalog.tlsSecret .Values.catalog.clientCASecret }}
       volumes:
@@ -33,6 +37,10 @@ spec:
       {{- end }}
       containers:
         - name: catalog-operator
+          securityContext:
+            allowPrivilegeEscalation: false
+            capabilities:
+              drop: [ "ALL" ]
           {{- if or .Values.catalog.tlsSecret .Values.catalog.clientCASecret }}
           volumeMounts:
           {{- end }}
@@ -76,6 +84,8 @@ spec:
           - --client-ca
           - /profile-collector-cert/tls.crt
           {{- end }}
+          - --set-workload-user-id
+          - "true"
           image: {{ .Values.catalog.image.ref }}
           imagePullPolicy: {{ .Values.catalog.image.pullPolicy }}
           ports:
--- a/deploy/chart/templates/_packageserver.deployment-spec.yaml
+++ b/deploy/chart/templates/_packageserver.deployment-spec.yaml
@@ -14,6 +14,10 @@ spec:
       labels:
         app: packageserver
     spec:
+      securityContext:
+        runAsNonRoot: true
+        seccompProfile:
+          type: RuntimeDefault
       serviceAccountName: olm-operator-serviceaccount
       {{- if .Values.package.nodeSelector }}
       nodeSelector:
@@ -25,6 +29,10 @@ spec:
       {{- end }}
       containers:
       - name: packageserver
+        securityContext:
+          allowPrivilegeEscalation: false
+          capabilities:
+            drop: [ "ALL" ]
         command:
         - /bin/package-server
         - -v=4
@@ -60,10 +68,6 @@ spec:
         {{- if .Values.package.resources }}
         resources:
 {{ toYaml .Values.package.resources | indent 10 }}
-        {{- end }}
-        {{- if .Values.package.securityContext }}
-        securityContext:
-          runAsUser: {{ .Values.package.securityContext.runAsUser }}
         {{- end }}
         volumeMounts:
         - name: tmpfs
--- a/pkg/controller/bundle/bundle_unpacker.go
+++ b/pkg/controller/bundle/bundle_unpacker.go
@@ -22,6 +22,7 @@ import (
 	listersbatchv1 "k8s.io/client-go/listers/batch/v1"
 	listerscorev1 "k8s.io/client-go/listers/core/v1"
 	listersrbacv1 "k8s.io/client-go/listers/rbac/v1"
+	"k8s.io/utils/pointer"
 
 	"github.com/operator-framework/api/pkg/operators/reference"
 	operatorsv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
@@ -101,6 +102,11 @@ func (c *ConfigMapUnpacker) job(cmRef *corev1.ObjectReference, bundlePath string
 					// See: https://kubernetes.io/docs/concepts/workloads/controllers/job/#pod-backoff-failure-policy
 					RestartPolicy:    corev1.RestartPolicyNever,
 					ImagePullSecrets: secrets,
+					SecurityContext: &corev1.PodSecurityContext{
+						SeccompProfile: &corev1.SeccompProfile{
+							Type: corev1.SeccompProfileTypeRuntimeDefault,
+						},
+					},
 					Containers: []corev1.Container{
 						{
 							Name:  "extract",
@@ -129,6 +135,12 @@ func (c *ConfigMapUnpacker) job(cmRef *corev1.ObjectReference, bundlePath string
 									corev1.ResourceMemory: resource.MustParse("50Mi"),
 								},
 							},
+							SecurityContext: &corev1.SecurityContext{
+								AllowPrivilegeEscalation: pointer.Bool(false),
+								Capabilities: &corev1.Capabilities{
+									Drop: []corev1.Capability{"ALL"},
+								},
+							},
 						},
 					},
 					InitContainers: []corev1.Container{
@@ -146,6 +158,12 @@ func (c *ConfigMapUnpacker) job(cmRef *corev1.ObjectReference, bundlePath string
 								Requests: corev1.ResourceList{
 									corev1.ResourceCPU:    resource.MustParse("10m"),
 									corev1.ResourceMemory: resource.MustParse("50Mi"),
+								},
+							},
+							SecurityContext: &corev1.SecurityContext{
+								AllowPrivilegeEscalation: pointer.Bool(false),
+								Capabilities: &corev1.Capabilities{
+									Drop: []corev1.Capability{"ALL"},
 								},
 							},
 						},
@@ -168,6 +186,12 @@ func (c *ConfigMapUnpacker) job(cmRef *corev1.ObjectReference, bundlePath string
 								Requests: corev1.ResourceList{
 									corev1.ResourceCPU:    resource.MustParse("10m"),
 									corev1.ResourceMemory: resource.MustParse("50Mi"),
+								},
+							},
+							SecurityContext: &corev1.SecurityContext{
+								AllowPrivilegeEscalation: pointer.Bool(false),
+								Capabilities: &corev1.Capabilities{
+									Drop: []corev1.Capability{"ALL"},
 								},
 							},
 						},
@@ -193,7 +217,10 @@ func (c *ConfigMapUnpacker) job(cmRef *corev1.ObjectReference, bundlePath string
 	job.SetNamespace(cmRef.Namespace)
 	job.SetName(cmRef.Name)
 	job.SetOwnerReferences([]metav1.OwnerReference{ownerRef(cmRef)})
-
+	if c.runAsUser > 0 {
+		job.Spec.Template.Spec.SecurityContext.RunAsUser = &c.runAsUser
+		job.Spec.Template.Spec.SecurityContext.RunAsNonRoot = pointer.Bool(true)
+	}
 	// By default the BackoffLimit is set to 6 which with exponential backoff 10s + 20s + 40s ...
 	// translates to ~10m of waiting time.
 	// We want to fail faster than that when we have repeated failures from the bundle unpack pod
@@ -246,6 +273,7 @@ type ConfigMapUnpacker struct {
 	loader        *configmap.BundleLoader
 	now           func() metav1.Time
 	unpackTimeout time.Duration
+	runAsUser     int64
 }
 
 type ConfigMapUnpackerOption func(*ConfigMapUnpacker)
@@ -332,6 +360,12 @@ func WithRoleBindingLister(rbLister listersrbacv1.RoleBindingLister) ConfigMapUn
 func WithNow(now func() metav1.Time) ConfigMapUnpackerOption {
 	return func(unpacker *ConfigMapUnpacker) {
 		unpacker.now = now
+	}
+}
+
+func WithUserID(id int64) ConfigMapUnpackerOption {
+	return func(unpacker *ConfigMapUnpacker) {
+		unpacker.runAsUser = id
 	}
 }
 
--- a/pkg/controller/bundle/bundle_unpacker_test.go
+++ b/pkg/controller/bundle/bundle_unpacker_test.go
@@ -15,6 +15,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/informers"
 	k8sfake "k8s.io/client-go/kubernetes/fake"
+	"k8s.io/utils/pointer"
 
 	operatorsv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 	crfake "github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/clientset/versioned/fake"
@@ -32,6 +33,7 @@ const (
 	opmImage    = "opm-image"
 	utilImage   = "util-image"
 	bundlePath  = "bundle-path"
+	runAsUser   = 1001
 )
 
 func TestConfigMapUnpacker(t *testing.T) {
@@ -220,6 +222,13 @@ func TestConfigMapUnpacker(t *testing.T) {
 								Spec: corev1.PodSpec{
 									RestartPolicy:    corev1.RestartPolicyNever,
 									ImagePullSecrets: []corev1.LocalObjectReference{{Name: "my-secret"}},
+									SecurityContext: &corev1.PodSecurityContext{
+										RunAsNonRoot: pointer.Bool(true),
+										RunAsUser:    pointer.Int64(runAsUser),
+										SeccompProfile: &corev1.SeccompProfile{
+											Type: corev1.SeccompProfileTypeRuntimeDefault,
+										},
+									},
 									Containers: []corev1.Container{
 										{
 											Name:    "extract",
@@ -243,6 +252,12 @@ func TestConfigMapUnpacker(t *testing.T) {
 													corev1.ResourceMemory: resource.MustParse("50Mi"),
 												},
 											},
+											SecurityContext: &corev1.SecurityContext{
+												AllowPrivilegeEscalation: pointer.Bool(false),
+												Capabilities: &corev1.Capabilities{
+													Drop: []corev1.Capability{"ALL"},
+												},
+											},
 										},
 									},
 									InitContainers: []corev1.Container{
@@ -260,6 +275,12 @@ func TestConfigMapUnpacker(t *testing.T) {
 												Requests: corev1.ResourceList{
 													corev1.ResourceCPU:    resource.MustParse("10m"),
 													corev1.ResourceMemory: resource.MustParse("50Mi"),
+												},
+											},
+											SecurityContext: &corev1.SecurityContext{
+												AllowPrivilegeEscalation: pointer.Bool(false),
+												Capabilities: &corev1.Capabilities{
+													Drop: []corev1.Capability{"ALL"},
 												},
 											},
 										},
@@ -282,6 +303,12 @@ func TestConfigMapUnpacker(t *testing.T) {
 												Requests: corev1.ResourceList{
 													corev1.ResourceCPU:    resource.MustParse("10m"),
 													corev1.ResourceMemory: resource.MustParse("50Mi"),
+												},
+											},
+											SecurityContext: &corev1.SecurityContext{
+												AllowPrivilegeEscalation: pointer.Bool(false),
+												Capabilities: &corev1.Capabilities{
+													Drop: []corev1.Capability{"ALL"},
 												},
 											},
 										},
@@ -397,6 +424,13 @@ func TestConfigMapUnpacker(t *testing.T) {
 								},
 								Spec: corev1.PodSpec{
 									RestartPolicy: corev1.RestartPolicyNever,
+									SecurityContext: &corev1.PodSecurityContext{
+										RunAsNonRoot: pointer.Bool(true),
+										RunAsUser:    pointer.Int64(runAsUser),
+										SeccompProfile: &corev1.SeccompProfile{
+											Type: corev1.SeccompProfileTypeRuntimeDefault,
+										},
+									},
 									Containers: []corev1.Container{
 										{
 											Name:    "extract",
@@ -420,6 +454,12 @@ func TestConfigMapUnpacker(t *testing.T) {
 													corev1.ResourceMemory: resource.MustParse("50Mi"),
 												},
 											},
+											SecurityContext: &corev1.SecurityContext{
+												AllowPrivilegeEscalation: pointer.Bool(false),
+												Capabilities: &corev1.Capabilities{
+													Drop: []corev1.Capability{"ALL"},
+												},
+											},
 										},
 									},
 									InitContainers: []corev1.Container{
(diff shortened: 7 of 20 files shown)


deploy/chart/templates/_packageserver.deployment-spec.yaml
```diff
@@ -14,6 +14,13 @@ spec:
       labels:
         app: packageserver
     spec:
+      securityContext:
+        runAsNonRoot: true
+        seccompProfile:
+          type: RuntimeDefault
+        {{- if eq .Values.installType "upstream" }}
```
perdasilva, line 21: nice!

deploy/chart/templates/0000_50_olm_00-namespace.yaml
```diff
@@ -2,9 +2,13 @@ apiVersion: v1
 kind: Namespace
 metadata:
   name: {{ .Values.namespace }}
+  labels: 
+    pod-security.kubernetes.io/enforce: restricted
```
perdasilva, line 6: we should also add
`pod-security.kubernetes.io/restricted-version: latest`

deploy/chart/templates/0000_50_olm_00-namespace.yaml
```diff
@@ -2,9 +2,13 @@ apiVersion: v1
...
 
 ---
 apiVersion: v1
 kind: Namespace
 metadata:
   name: {{ .Values.operator_namespace }}
+  labels:
+    pod-security.kubernetes.io/enforce: restricted
```
perdasilva, line 14: same as above

perdasilva:
In general it lgtm. Seems to hit all the right changes. Nicely done.
In case I missed it, it's probably worthwhile adding input validation to the runAsUser. I think at the moment 0 is allowed (as input), but won't end up stamping out a runAsUser (which is fine). But, without the validation it could be a bit confusing for the user.

I still suggest not giving the userId as a degree of freedom to the admin, at least not if we don't have a specific ask for it. The reasons being is that it could lead to backwards compatibility pain later on in case we need to fix the userId for some reason or another. If there's no clear customer need, we're also increasing the complexity of the code for no good reason. Even if we do needed it, I wonder if there's a nice library or pattern in go for storing this kind of configuration so it doesn't have to get passed all the way down the stack.

perdasilva:
Another couple of things that occured to me while I slept is that by default we _need_ to have the runAsUser in the securityContext of the jobs and catalog source pods in the majority of cases. The reason being that we cannot guarantee that the catalog source image or the bundle image have users defined in their containers. But, we kept the default userId value to -1 (i.e. no runAsUser). This means that the new flag is always necessary (except in OCP where SCC will inject the value). Even if we still want to keep the user id flag (rather than use a toggle), we should probably have a different default so that it doesn't have to be used/defined in the majority case.

anik120:
>Another couple of things that occured to me while I slept is that by default we need to have the runAsUser in the securityContext of the jobs and catalog source pods in the majority of cases. The reason being that we cannot guarantee that the catalog source image or the bundle image have users defined in their containers. But, we kept the default userId value to -1 (i.e. no runAsUser). This means that the new flag is always necessary (except in OCP where SCC will inject the value). Even if we still want to keep the user id flag (rather than use a toggle), we should probably have a different default so that it doesn't have to be used/defined in the majority case.

I see your point, I've changed the default value to 1001. 

> In case I missed it, it's probably worthwhile adding input validation to the runAsUser. I think at the moment 0 is allowed (as input), but won't end up stamping out a runAsUser (which is fine). But, without the validation it could be a bit confusing for the user.

With that ☝🏽 change in the context, the validation  `if runAsUser >= 0{ set UserID}` allows for any positive numerical value, including 0 (root), which is fine since if someone sets `runAsNonRoot` but sets `runAsUser` to 0, PSA controller will help us out by denying admission. But also, if for whatever reason, we do need to change back to `runAsRoot`, we can easily set UID to 0 without any code change. 

>I still suggest not giving the userId as a degree of freedom to the admin, at least not if we don't have a specific ask for it. The reasons being is that it could lead to backwards compatibility pain later on in case we need to fix the userId for some reason or another.

We do have other args that we pass to the containers, and the same backward compatibility argument can be extended to them too no? In fact, if we hardcode the value, and have to change the value, we'll have to actually backport code changes. Whereas if we don't hardcode the values, and we need to change the UID for previous releases, all we need is release manifest changes for previous releases 🎉 

> If there's no clear customer need, we're also increasing the complexity of the code for no good reason

On the contrary, if we don't leave this degree of freedom, besides the problem mentioned above, we'll possibly have to keep changing the hardcoded value with evolving PSA enablement effort org wide at present too. 

Hardcoding "run workload as UID" sounds like brings the exact problems you're trying to avoid in the first place @perdasilva , and the solution is to not harcode it.

anik120:
Oh and 
> Even if we do needed it, I wonder if there's a nice library or pattern in go for storing this kind of configuration so it doesn't have to get passed all the way down the stack.

Unfortunately, since the workloads are being created way down in the stack, the configuration needs to be passed down all the way to the bottom (almost). I am not a Go expert yet (😉) either, but this is the pattern I've always observed being used every where. And I believe it is one of the short comings of Go too, but don't quote me on that (says he after documenting it down in writing), although I'm sure an actual Go expert can articulate the shortcoming better than me.

perdasilva:
It's more a question of control. If we need to change it, we can, and it's not impact for the user. If we decide we need to pin to a user or another because of the container or other possible future changes, we can easily control that. Up and downstream. I'd just check that we aren't pulling from olm:latest for the controller. If that's not the case then the point is moot. But, just as a general guideline (and that's all it is), don't add more functionality than you need, because you'll end up having to support it.

perdasilva:
Oh, just came to me. Something we'd do a lot at Amazon for our services was to use env vars instead. Those can be easily picked up the code without having to traverse the whole stack. It also avoids all the unit testing issues of using a singleton.

camilamacedo86 (changes requested):
Hi @anik120 that is great but we can **not** use **RunAsUser** 
We need to remove ALL RunAsUser fields added.


#### Option A)  

Define the USER ID via the Dockerfile image. Therefore, we do NOT need to use the spec RunAsUser in the SecurityContext spec  at all and we do not face the issue "`container has runAsNonRoot and image will run as root` "

Example:
Docker Image
USER 1001

Then, SecurityContext:

```ymal
    spec:
      securityContext:
        runAsNonRoot: true
        # Please ensure that you can use SeccompProfile and do not use
        # if your project must work on old Kubernetes
        # versions < 1.19 or on vendors versions which
        # do NOT support this field by default (i.e. Openshift < 4.11 )
        seccompProfile:
          type: RuntimeDefault
      ...
      containers:
      - name: controller-manager
        securityContext:
          allowPrivilegeEscalation: false
          capabilities:
            drop:
            - ALL

```

Result: the containers will be qualified to restricted-v2 in OCP and will run as restricted on k8s

####  Option B: 

- We will need to ensure that the namespace is enforced with the Baseline
- We do not use the RunAsNonRoot must be empty or false

cmd/catalog/start.go
```diff
@@ -66,6 +67,7 @@ func newRootCmd() *cobra.Command {
 	cmd.Flags().StringVar(&o.opmImage, "opmImage", defaultOPMImage, "the image to use for unpacking bundle content with opm")
 	cmd.Flags().StringVar(&o.utilImage, "util-image", defaultUtilImage, "an image containing custom olm utilities")
 	cmd.Flags().StringVar(&o.writeStatusName, "writeStatusName", defaultOperatorName, "ClusterOperator name in which to write status, set to \"\" to disable.")
+	cmd.Flags().Int64Var(&o.workloadUserID, "workload-user-id", 1001, "The non-root user ID that registry pods will run as.")
```
camilamacedo86, line 70: Why are we adding this flag? 
What is the image used which does not define the USER ID?

deploy/chart/templates/0000_50_olm_00-namespace.yaml
```diff
@@ -2,9 +2,19 @@ apiVersion: v1
 kind: Namespace
 metadata:
   name: {{ .Values.namespace }}
+  labels: 
+    pod-security.kubernetes.io/enforce: restricted
+    pod-security.kubernetes.io/enforce-version: latest
+    pod-security.kubernetes.io/warn: restricted
```
camilamacedo86, line 8:
```suggestion
```
If we enforce: restricted then we do not need to warn it.
The containers that cannot run as restricted will not be admitted so has no point to warn

deploy/chart/templates/0000_50_olm_00-namespace.yaml
```diff
@@ -2,9 +2,19 @@ apiVersion: v1
...
 
 ---
 apiVersion: v1
 kind: Namespace
 metadata:
   name: {{ .Values.operator_namespace }}
+  labels:
+    pod-security.kubernetes.io/enforce: baseline
+    pod-security.kubernetes.io/enforce-version: latest
+    pod-security.kubernetes.io/warn: baseline
```
camilamacedo86, line 19:
```suggestion
    pod-security.kubernetes.io/warn: restricted
```

If the enforce is baseline then we can warn when any container on this namespace is not restricted OR not set up the warning at all. But the warning makes only sense if be less permissive than the enforcement.

deploy/chart/templates/0000_50_olm_07-olm-operator.deployment.yaml
```diff
@@ -17,6 +17,13 @@ spec:
       labels:
         app: olm-operator
     spec:
+      securityContext:
+        runAsNonRoot: true
+        seccompProfile:
+          type: RuntimeDefault
+        {{- if eq .Values.installType "upstream" }}
+        runAsUser: {{ .Values.package.securityContext.runAsUser }}
+        {{- end }}
```
camilamacedo86, line 26:
```suggestion
```
We cannot use RunAsUser at all,
That is ONLY required because the Dockerfile image is not defining the user when should be.
So, what is the image used here?

deploy/chart/templates/0000_50_olm_08-catalog-operator.deployment.yaml
```diff
@@ -17,6 +17,13 @@ spec:
       labels:
         app: catalog-operator
     spec:
+      securityContext:
+        runAsNonRoot: true
+        seccompProfile:
+          type: RuntimeDefault
+        {{- if eq .Values.installType "upstream" }}
+        runAsUser: {{ .Values.package.securityContext.runAsUser }}
+        {{- end }}
```
camilamacedo86, line 26:
```suggestion
```
Same. We cannot use RunAsUser at all,
That is ONLY required because the Dockerfile image is not defining the user when should be.
So, what is the image used here? 

Another option, in this case, is to ensure enforcement as a baseline or privileged and not set RunAsNonRoot either.

deploy/chart/templates/0000_50_olm_08-catalog-operator.deployment.yaml
```diff
@@ -76,6 +87,8 @@ spec:
           - --client-ca
           - /profile-collector-cert/tls.crt
           {{- end }}
+          - --workload-user-id
```
camilamacedo86, line 90:
```suggestion
```
The idea of passing the USER ID will not work.
We have no way to know a fixed value number in OCP.

deploy/chart/templates/0000_50_olm_08-catalog-operator.deployment.yaml
perdasilva: Nah we should keep security as nerfed as possible. I don't think the controllers need anything above restricted.

deploy/chart/templates/0000_50_olm_08-catalog-operator.deployment.yaml
perdasilva: The idea here for the downstream is to omit the flag and let SCC inject the user id in the range it wants.

deploy/chart/templates/0000_50_olm_07-olm-operator.deployment.yaml
perdasilva: That's a fair point actually. Since we are baking the USER into the olm container we don't need runAsUser for the controller deployments. Only for the catalog source and unpack jobs, since we have less control there.

deploy/chart/templates/0000_50_olm_00-namespace.yaml
perdasilva: Oh, we should enforce restricted on the olm namespace! Can remove the warn (and audit) label(s)

perdasilva:
Regarding validation it's better to fail fast, rather than let PSA catch it. Might cause confusion. Idk that I'd change the runAsNonRoot behavior. More stuff to test.

cmd/catalog/start.go
camilamacedo86, line 70:
```suggestion
```


The idea of passing the USER ID will not work.
We do have not a way to know a fixed value number valid for OCP. 
Also, that would mean ensuring that the ServiceAccount used would have permissions to the anyuid SCC.

Therefore, we have two options:
a) Ensure that the  USERID is define in the dockerfile image
OR 
b) Do not set ANY userID and do NOT set RunAsNonRoot 
Then, for downstream we need to ensure that the SA has access to the required SCC ( i.e. previleged )
Also, ensure that the namespace will be labeled


The valid range is based on the namespace config. 
Also, we cannot use RunAsUser. 

From OCP:

> 

If your S2I image does not include a USER declaration with a numeric user, your builds will fail by default. In order to allow images that use either named users or the root (0) user to build in OpenShift Container Platform, you can add the project’s builder service account (system:serviceaccount:<your-project>:builder) to the privileged security context constraint (SCC). Alternatively, you can allow all images to run as any user.
--

deploy/chart/templates/_packageserver.deployment-spec.yaml
```diff
@@ -14,6 +14,13 @@ spec:
...
+        runAsUser: {{ .Values.package.securityContext.runAsUser }}
+        {{- end }}
```
camilamacedo86, line 23:
```suggestion
```
Same

deploy/chart/values.yaml
```diff
@@ -53,6 +53,8 @@ package:
     internalPort: 5443
   nodeSelector:
     kubernetes.io/os: linux
+  securityContext: 
+    runAsUser: 1001
```
camilamacedo86, line 57:
```suggestion
```
Same

camilamacedo86: The image `quay.io/operator-framework/olm:master` must define the USER ID instead

pkg/controller/bundle/bundle_unpacker.go
```diff
@@ -101,6 +102,12 @@ func (c *ConfigMapUnpacker) job(cmRef *corev1.ObjectReference, bundlePath string
 					// See: https://kubernetes.io/docs/concepts/workloads/controllers/job/#pod-backoff-failure-policy
 					RestartPolicy:    corev1.RestartPolicyNever,
 					ImagePullSecrets: secrets,
+					SecurityContext: &corev1.PodSecurityContext{
+						RunAsNonRoot: pointer.Bool(true),
+						SeccompProfile: &corev1.SeccompProfile{
+							Type: corev1.SeccompProfileTypeRuntimeDefault,
```
camilamacedo86, line 108: will we run this version in OCP < 4.11 ? If so we cannot use it.

pkg/controller/bundle/bundle_unpacker.go
```diff
@@ -193,7 +218,9 @@ func (c *ConfigMapUnpacker) job(cmRef *corev1.ObjectReference, bundlePath string
 	job.SetNamespace(cmRef.Namespace)
 	job.SetName(cmRef.Name)
 	job.SetOwnerReferences([]metav1.OwnerReference{ownerRef(cmRef)})
-
+	if runAsUser >= 0 {
+		job.Spec.Template.Spec.SecurityContext.RunAsUser = &runAsUser
+	}
```
camilamacedo86, line 223:
```suggestion
```
Same

pkg/controller/bundle/bundle_unpacker.go
```diff
@@ -229,7 +256,7 @@ func (c *ConfigMapUnpacker) job(cmRef *corev1.ObjectReference, bundlePath string
 //go:generate go run github.com/maxbrunsfeld/counterfeiter/v6 . Unpacker
 
 type Unpacker interface {
-	UnpackBundle(lookup *operatorsv1alpha1.BundleLookup, timeout time.Duration) (result *BundleUnpackResult, err error)
+	UnpackBundle(lookup *operatorsv1alpha1.BundleLookup, timeout time.Duration, runAsUser int64) (result *BundleUnpackResult, err error)
```
camilamacedo86, line 259:
```suggestion
	UnpackBundle(lookup *operatorsv1alpha1.BundleLookup, timeout time.Duration) (result *BundleUnpackResult, err error)
```

pkg/controller/bundle/bundle_unpacker.go
```diff
@@ -383,7 +410,7 @@ const (
 	NotUnpackedMessage          = "bundle contents have not yet been persisted to installplan status"
 )
 
-func (c *ConfigMapUnpacker) UnpackBundle(lookup *operatorsv1alpha1.BundleLookup, timeout time.Duration) (result *BundleUnpackResult, err error) {
+func (c *ConfigMapUnpacker) UnpackBundle(lookup *operatorsv1alpha1.BundleLookup, timeout time.Duration, runAsUser int64) (result *BundleUnpackResult, err error) {
```
camilamacedo86, line 413:
```suggestion
func (c *ConfigMapUnpacker) UnpackBundle(lookup *operatorsv1alpha1.BundleLookup, timeout time.Duration) (result *BundleUnpackResult, err error) {
```

pkg/controller/bundle/bundle_unpacker.go
```diff
@@ -445,7 +472,7 @@ func (c *ConfigMapUnpacker) UnpackBundle(lookup *operatorsv1alpha1.BundleLookup,
 		secrets = append(secrets, corev1.LocalObjectReference{Name: secretName})
 	}
 	var job *batchv1.Job
-	job, err = c.ensureJob(cmRef, result.Path, secrets, timeout)
+	job, err = c.ensureJob(cmRef, result.Path, secrets, timeout, runAsUser)
```
camilamacedo86, line 475:
```suggestion
	job, err = c.ensureJob(cmRef, result.Path, secrets, timeout)
```

pkg/controller/bundle/bundle_unpacker.go
```diff
@@ -584,8 +611,8 @@ func (c *ConfigMapUnpacker) ensureConfigmap(csRef *corev1.ObjectReference, name
 	return
 }
 
-func (c *ConfigMapUnpacker) ensureJob(cmRef *corev1.ObjectReference, bundlePath string, secrets []corev1.LocalObjectReference, timeout time.Duration) (job *batchv1.Job, err error) {
-	fresh := c.job(cmRef, bundlePath, secrets, timeout)
+func (c *ConfigMapUnpacker) ensureJob(cmRef *corev1.ObjectReference, bundlePath string, secrets []corev1.LocalObjectReference, timeout time.Duration, runAsUser int64) (job *batchv1.Job, err error) {
+	fresh := c.job(cmRef, bundlePath, secrets, timeout, runAsUser)
```
camilamacedo86, line 615:
```suggestion
	fresh := c.job(cmRef, bundlePath, secrets, timeout)
```

pkg/controller/bundle/bundle_unpacker_test.go
```diff
@@ -220,6 +222,13 @@ func TestConfigMapUnpacker(t *testing.T) {
 								Spec: corev1.PodSpec{
 									RestartPolicy:    corev1.RestartPolicyNever,
 									ImagePullSecrets: []corev1.LocalObjectReference{{Name: "my-secret"}},
+									SecurityContext: &corev1.PodSecurityContext{
+										RunAsNonRoot: pointer.Bool(true),
+										RunAsUser:    pointer.Int64(runAsUser),
```
camilamacedo86, line 227:
```suggestion
```

pkg/controller/bundle/bundle_unpacker_test.go
```diff
@@ -32,6 +33,7 @@ const (
 	opmImage    = "opm-image"
 	utilImage   = "util-image"
 	bundlePath  = "bundle-path"
+	runAsUser   = 1001
```
camilamacedo86, line 36:
```suggestion
```

anik120:
> Hi @anik120 that is great but we can **not** use **RunAsUser** We need to remove ALL RunAsUser fields added.
> 
> #### Option A)  
> Define the USER ID via the Dockerfile image. Therefore, we do NOT need to use the spec RunAsUser in the SecurityContext spec at all and we do not face the issue "`container has runAsNonRoot and image will run as root` "
> 
> Example: Docker Image USER 1001
> 
> Then, SecurityContext:
> 
> ```
>     spec:
>       securityContext:
>         runAsNonRoot: true
> ```

@camilamacedo86 keep in mind that this is the upstream PR. Conversations about setting `runAsUser` to comply with OCP SCC are not valid for this PR. That'll be discussed while downstreaming this PR.

perdasilva:
> On the contrary, if we don't leave this degree of freedom, besides the problem mentioned above, we'll possibly have to keep changing the hardcoded value with evolving PSA enablement effort org wide at present too.

I'm more worried about upstream users here. Changing a hard-coded value is trivial. Deprecating behavior is hard. Removing the degree of freedom here gives us control to do what we want without affecting users or having to take support promises into account. I don't know what kinds of decisions admins will make around user id and how removing ability would affect them. But I'll leave it to you to decide. It's just a call out. Configuration certainly has its advantages. I just cannot predict how this space will change, which is why I tend to fall on the side of keeping control.

deploy/chart/values.yaml
camilamacedo86: @anik120 We need to define the user in : https://github.com/operator-framework/operator-lifecycle-manager/blob/master/Dockerfile#L36-L38

pkg/controller/bundle/bundle_unpacker.go
anik120: We will not, this is for upstream: latest. Please reserve all conversations about OCP for the downstream PR.

cmd/catalog/start.go
anik120: Again, this conversation is valid for downstream, so we'll discuss this in the downstream PR.

deploy/chart/templates/0000_50_olm_08-catalog-operator.deployment.yaml
anik120: +1 to what Per said.

camilamacedo86 (changes requested):
We need to change the Dockerfile images instead of pass the UserId as flag
/hold

deploy/chart/templates/0000_50_olm_08-catalog-operator.deployment.yaml
camilamacedo86: > The idea here for the downstream is to omit the flag and let SCC inject the user id in the range it wants\

I understood the idea and that will not work.
We have no way to know the value to inform. What value will you pass when it will run in OCP? That is not a field value or a fixed range at all. 

Also, the good practices for Docker are to have the userID set in the image ( which sort out the issue ), remove all these complexities, and make it eaiser for us to keep the project maintained.

deploy/chart/templates/0000_50_olm_07-olm-operator.deployment.yaml
camilamacedo86, line 26:
```suggestion
```

deploy/chart/templates/0000_50_olm_08-catalog-operator.deployment.yaml
camilamacedo86: @perdasilva @anik120 could you describe how this injection will work?

deploy/chart/templates/0000_50_olm_00-namespace.yaml
camilamacedo86: @anik120 ^

deploy/chart/templates/0000_50_olm_00-namespace.yaml
camilamacedo86: @anik120

deploy/chart/templates/0000_50_olm_07-olm-operator.deployment.yaml
camilamacedo86: @anik120

deploy/chart/templates/0000_50_olm_07-olm-operator.deployment.yaml
```diff
@@ -17,6 +17,13 @@ spec:
       labels:
         app: olm-operator
     spec:
+      securityContext:
+        runAsNonRoot: true
+        seccompProfile:
+          type: RuntimeDefault
+        {{- if eq .Values.installType "upstream" }}
+        runAsUser: {{ .Values.package.securityContext.runAsUser }}
```
camilamacedo86, line 25: @anik120 @perdasilva,
 
In "downstream" if we omit this value when the image (Dockerfile) does not set the USER ID that will fail with the error `container has runAsNonRoot and image has non-numeric user`.

Please, test this scenario with an image that requires the RunAsUserID example: 'memcached:1.4.36-alpine'
Example : https://gist.github.com/camilamacedo86/c7c06e51272f8c348655123d0f18eceb

deploy/chart/templates/0000_50_olm_00-namespace.yaml
anik120: For the `operators` namespace, I've updated it to be the following: 

```
pod-security.kubernetes.io/enforce: baseline
pod-security.kubernetes.io/enforce-version: latest
pod-security.kubernetes.io/warn: privileged
pod-security.kubernetes.io/warn-version: latest
```

From what I've read, I think this means "allow baseline pods, but also allow privileged pod with a warning". Reason for keeping this configuration is that baseline should allow most of the CSVs out there to be installed in that namespace. If there are any operators that require privileged access, we need to allow those, but throw a warning. PLMK if the above configuration doesn't achieve what I think it does.

anik120:
@perdasilva @camilamacedo86 updated the PR with the following updates: 

1. The olm and catalog operator deployment does not explicitly set `runAsUser`, instead the images are built with user 1001. 
2. The registry pods and the bundle unpacking job *still absolutely needs* their `runAsUser` set explicitly, since they're not being built from any image (check code of how they're created for more details) 
3. The flag for the `catalog-operator` container arg is not a boolean (that sets the `runAsUser` as mentioned in 2 to 1001 if set to `true`)
>I'm more worried about upstream users here. Changing a hard-coded value is trivial. Deprecating behavior is hard. Removing the degree of freedom here gives us control to do what we want without affecting users or having to take support promises into account

That makes sense to me. It'll behoove us to take this conversation to the wg meeting and let the community know that we are exposing this flag, and based on the feedback we receive, if we see that there's appetite for needing to set a custom user id for those workloads, we can introduce a new flag and get rid of the boolean one.

deploy/chart/templates/0000_50_olm_00-namespace.yaml
```diff
@@ -2,9 +2,17 @@ apiVersion: v1
...
+    pod-security.kubernetes.io/enforce: restricted
+    pod-security.kubernetes.io/enforce-version: latest
 
 ---
 apiVersion: v1
 kind: Namespace
 metadata:
   name: {{ .Values.operator_namespace }}
+  labels:
+    pod-security.kubernetes.io/enforce: baseline
```
camilamacedo86, line 15: @anik120 @perdasilva any pod installed on this namespace will NOT have ONLY restricted-v2 SCC. 
Therefore, if you do not inform the RunAsUser and you set RunAsNonRoot true the SCC will not inject the RunAsUser. 

It will fail in Openshift when the image does not define the UserID

camilamacedo86: /hold

deploy/chart/templates/0000_50_olm_00-namespace.yaml
```diff
@@ -2,9 +2,17 @@ apiVersion: v1
...
+    pod-security.kubernetes.io/enforce-version: latest
+    pod-security.kubernetes.io/warn: privileged
```
camilamacedo86, line 17: @anik120 @perdasilva 

This namespace has `pod-security.kubernetes.io/enforce: baseline` which means `privileged ` pods will not run. So, make no sense we add the ` pod-security.kubernetes.io/warn: privileged`

deploy/chart/templates/0000_50_olm_00-namespace.yaml
```diff
@@ -2,9 +2,17 @@ apiVersion: v1
 kind: Namespace
 metadata:
   name: {{ .Values.namespace }}
+  labels: 
+    pod-security.kubernetes.io/enforce: restricted
```
camilamacedo86, line 6: In Openshift the label syncer can change that. 
Therefore, that means the solution proposed here will fail when:

- The Dockerfile image does not define the USER ID
AND
- The Pod be created with RunAsNonRoot
AND
- The label syncer changes the enforcement of this namespace  

@perdasilva @anik120

camilamacedo86 (commented):
/lgtm

camilamacedo86 (commented):
/hold

deploy/chart/templates/0000_50_olm_00-namespace.yaml
```diff
@@ -2,9 +2,15 @@ apiVersion: v1
...
+    pod-security.kubernetes.io/enforce-version: latest
 
 ---
 apiVersion: v1
 kind: Namespace
 metadata:
   name: {{ .Values.operator_namespace }}
+  labels:
+    pod-security.kubernetes.io/enforce: baseline
+    pod-security.kubernetes.io/enforce-version: latest
```
camilamacedo86, line 16: @anik120 is just a detail, if we add here the latest that means it will get the latest criteria available from the policy. therefore, in the future, it might no longer work with the changes in this code/release.

camilamacedo86: we create an issue to address this scenario: https://github.com/operator-framework/operator-lifecycle-manager/issues/2827 and it is blocking the release and will get done as follow up so all fine. 

/hold cancel

PR #2825, opened 2022-07-29
ci: Add support for configuring the global catalog namespace

timflannagan:

Add support for configuring the global catalog namespace in the e2e testing suite. Introduces a CLI flag for configuring the global catalog namespace for OLM. This kind of configuration is already present in the catalog-operator's CLI flags, so this is essentially a no-op for upstream OLM given we deploy the OLM stack and the catalog sources in the same namespace. These changes should make it easier for different distributions of OLM, e.g. openshift, that deploy any default catalog source resources in a dedicated namespace.

Accommodate different distributions of OLM that have dedicated catalog namespaces. Reduce any testing flakes for OLM.

**Architectural changes:**
N/A.

**Testing remarks:**

**Reviewer Checklist**


Makefile
```diff
@@ -134,10 +134,11 @@ TEST := $(shell go run ./test/e2e/split/... -chunks $(E2E_TEST_NUM_CHUNKS) -prin
 endif
 E2E_OPTS ?= $(if $(E2E_SEED),-seed '$(E2E_SEED)') $(if $(SKIP), -skip '$(SKIP)') $(if $(TEST),-focus '$(TEST)') $(if $(ARTIFACT_DIR), -output-dir $(ARTIFACT_DIR) -junit-report junit_e2e.xml) -flake-attempts $(E2E_FLAKE_ATTEMPTS) -nodes $(E2E_NODES) -timeout $(E2E_TIMEOUT) -v -randomize-suites -race -trace -progress
 E2E_INSTALL_NS ?= operator-lifecycle-manager
+E2E_CATALOG_NS ?= $(E2E_INSTALL_NS)
```
timflannagan, line 137: Setting this to the same value as the OLM installation namespace is intention. This seemed like the best approach to reduce the amount of variables that needed to be overridden between the e2e and e2e-local targets. For upstream environments, you may only need to run `make e2e E2E_CATALOG_NS=olm` for clusters that were already created, and for the e2e-local target, it should be a no-op but there's no integration tests that can easily test all these configuration permutations.

timflannagan:
Note: this should fix a permafailing downstream OLM testing flake. See https://github.com/openshift/operator-framework-olm/pull/345 which was a sanity check before opening these changes.

When running the OLM testing suite on an OCP (or crc) cluster, you could run the following: `make e2e E2E_CATALOG_NS=openshift-marketplace E2E_INSTALL_NS=openshift-operator-lifecycle-manager E2E_TEST_NS=openshift-operators` now.

dinhxuanvu (approved):
/approve

Issue #2830, opened 2022-08-03
Attempting to reinstall an operator that uses conversion webhooks fails


amisevsk:
## Bug Report

**What did you do?**

When attempting to reinstall an operator that uses conversion webhooks by

1. Deleting the operator subscription and any CSVs associated with it
2. Recreating the operator subscription

The resulting InstallPlan enters a failed state with message similar to

> error validating existing CRs against new CRD's schema for "devworkspaces.workspace.devfile.io": error listing resources in GroupVersionResource schema.GroupVersionResource{Group:"workspace.devfile.io", Version:"v1alpha1", Resource:"devworkspaces"}: conversion webhook for workspace.devfile.io/v1alpha2, Kind=DevWorkspace failed: Post "[https://devworkspace-controller-manager-service.test-namespace.svc:443/convert?timeout=30s](https://devworkspace-controller-manager-service.test-namespace.svc/convert?timeout=30s)": service "devworkspace-controller-manager-service" not found

When the original CSVs are deleted, the operator's main deployment and service are removed, but CRDs are left in-cluster. However, since the service/CA bundle/deployment that serve the conversion webhook are removed, conversion webhooks are broken at that point. Eventually this impacts garbage collection on the cluster as well.

This can be reproduced by installing the DevWorkspace Operator from the Red Hat catalog. (I can provide yamls/upstream images that reproduce as well, if that's helpful). It may be necessary to create a DevWorkspace in the cluster before deletion, e.g. by `oc apply -f https://raw.githubusercontent.com/devfile/devworkspace-operator/main/samples/plain.yaml`

**What did you expect to see?**
Operator is able to be reinstalled without removing CRDs and all instances.

**What did you see instead? Under which circumstances?**
It's necessary to completely remove the operator including CRDs. For our operator (DevWorkspace), this also makes uninstall especially complicated as finalizers are used (so CRDs cannot be deleted if the controller is removed, and the controller cannot be restored by reinstalling)

**Environment**
* operator-lifecycle-manager version: 4.10.24

* Kubernetes version information: Kubernetes Version: v1.23.5+012e945 (OpenShift 4.10.24)

* Kubernetes cluster kind:  OpenShift

**Possible Solution**
The only details I noticed that may be relevant are
1. The order of objects in the InstallPlan varies per install, and sometimes CRDs are applied earlier
2. The installPlan fails at the point of trying to apply CRDs
[...]

PR #2832, opened 2022-08-05
Cleanup conversion webhooks when an operator is uninstalled

awgreene:
Addresses #2830

Problem: When uninstalling a CSV, OLM has always avoided deleting the
associated CRD as all CRs on cluster are subsequently deleted, possibly
resulting in user dataloss.

OLM supports defining conversion webhooks within the CSV. On cluster,
conversion webhooks are defined with a CRD and point to a service that
handles conversion.  If the service is unable to fulfill the request,
all requests against the CRs associated with the CRD will fail.

When uninstalling a CSV, OLM does not remove the conversion webhook from
the CRD, meaning that all requests against the CRs associated with the
CRD will fail, resulting in at least two concerns:
1. OLM is unable to subsequently reinstall the operator. When installing
   a CSV, if the CRD already exists and instances of CRs exist as well,
   OLM performs a series of checks which ensure that none of the CRs are
   invalidated against the new schema. The existing CRD's conversion
   webhooks points to a non-existant service, causing the check to fail
   and preventing installs.
2. Broken conversion webhooks causes kubernete's garbage collection to
   fail.

Solution: When a CSV is deleted, if no CSV exists that is replacing it,
set the CRD's conversion strategy to None.

### DIFF
--- a/pkg/controller/operators/olm/operator.go
+++ b/pkg/controller/operators/olm/operator.go
@@ -11,6 +11,7 @@ import (
 	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	extinf "k8s.io/apiextensions-apiserver/pkg/client/informers/externalversions"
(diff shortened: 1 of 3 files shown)


awgreene:
Feel free to take a look @amisevsk.

awgreene:
Needs an e2e test
/hold

awgreene:
I've added the e2e test.
/unhold

PR #2834, opened 2022-08-09
improve CA and certificate generation

joelanford:
Recently during an audit on a user's cluster, it was discovered that
OLM's certificate generation functionality has a few minor shortcomings.
  1) The generated CA and server cert do not include a common name,
     which causes some tooling to have trouble tracing the cert chain.
  2) The generated CA and server cert include unnecessary key usages,
     which means those certificates can be used for more than their
     intended purposes.

This commit resolves the above issues by ensuring the certificates
include common names and by using the minimal key usages necessary.

**Testing remarks:**
Thus far, I have only manually tested via `run-local` and verified that:
1) The generated packageserver secret contained the expected cerificates.
2) The packagemanifests API worked (thus proving the kube-apiserver can use OLM's generated CA to verify OLM's generated cert for the packageserver api service.

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/certs/certs.go
+++ b/pkg/controller/certs/certs.go
@@ -71,13 +71,13 @@ func GenerateCA(notAfter time.Time, organization string) (*KeyPair, error) {
 	caDetails := &x509.Certificate{
 		SerialNumber: serial,
 		Subject: pkix.Name{
+			CommonName:   fmt.Sprintf("olm-selfsigned-%x", serial),
 			Organization: []string{organization},
 		},
 		NotBefore:             notBefore,
 		NotAfter:              notAfter,
 		IsCA:                  true,
-		ExtKeyUsage:           []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth, x509.ExtKeyUsageServerAuth},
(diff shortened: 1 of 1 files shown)


dinhxuanvu (approved):
/approve

pkg/controller/certs/certs.go
```diff
@@ -120,12 +120,12 @@ func CreateSignedServingPair(notAfter time.Time, organization string, ca *KeyPai
 	certDetails := &x509.Certificate{
 		SerialNumber: serial,
 		Subject: pkix.Name{
+			CommonName:   hosts[0],
```
timflannagan, line 123: Do we do any length checking for the hosts array length anywhere?

timflannagan:
Just one quick question, but otherwise this lgtm.

pkg/controller/certs/certs.go
joelanford: Right now, it always has exactly two items:

0. `<svcName>.<svcNamespace>`
1. `<svcName>.<svcNamespace>.svc`

timflannagan: Alright, works for me.

/lgtm

PR #2845, opened 2022-08-24
(psa) allow legacy Catalogsources to run in non-restrcted namespaces

anik120:
This PR configures the Catalogsource reconciler to use the spec.GrpcPodConfig.SecurityContextConfig
field to determine if the pod.spec.securityContext and container[*].spec.SecurityContext for the registry
pod should be configured to be runnable in a PSA restrcited namespace or not, so that cluster admins can
indicate that they want to run legacy catalogsources in a non-resctricted (baseline/privileged)	namespace. 

This allows cluster admins to run catalogsources that are built with a version of opm that is less than 
v1.23.2 (i.e a version of opm that does not contain [this commit](https://github.com/operator-framework/operator-registry/pull/974)

Signed-off-by: Anik Bhattacharjee <anikbhattacharya93@gmail.com>

**Architectural changes:**

**Testing remarks:**

**Reviewer Checklist**


vendor/github.com/operator-framework/api/pkg/operators/v1alpha1/catalogsource_types.go
```diff
@@ -88,6 +89,10 @@ type CatalogSourceSpec struct {
 	// +optional
 	Secrets []string `json:"secrets,omitempty"`
 
+	// RunAsRoot allows admins to indicate that they wish to run the CatalogSource pod in a privileged
+	// more as root
+	RunAsRoot bool `json:"runAsRoot,omitempty"`
```
dinhxuanvu, line 94: I assume this is a temp solution for discussion before you make a permanent PR on api repo? Plus, this is considered an API change. Also, do we consider this field to be optional or not? If it ism an `+optional` marker would be nice.

anik120: Unfortunately the introduction of the new filed isn't a temp solution. It's the only way to get around the problem of "old catalogs cannot run in 4.12 with enforce:restricted enabled by default". I'll update the PR to mark this field as `+optional`

anik120: Oh right, that's for review. Just opened a PR in the api repo that'll be vendor-ed in https://github.com/operator-framework/api/pull/257

deploy/chart/crds/0000_50_olm_00-catalogsources.crd.yaml
```diff
@@ -120,6 +120,9 @@ spec:
                   type: integer
                 publisher:
                   type: string
+                runAsRoot:
+                  description: RunAsRoot allows admins to indicate that they wish to run the CatalogSource pod in a privileged more as root
```
bparees, line 124:
```suggestion
                  description: RunAsRoot allows admins to indicate that they wish to run the CatalogSource pod in a privileged pod as root.  This should only be enabled when running older catalog images which could not be run as non-root.
```

anik120: Has been incorporated in the API PR https://github.com/operator-framework/api/pull/258

pkg/controller/registry/reconciler/reconciler.go
```diff
@@ -181,32 +181,38 @@ func Pod(source *operatorsv1alpha1.CatalogSource, name string, image string, saN
...
 		},
 	}
-
-	if runAsUser > 0 {
-		pod.Spec.SecurityContext.RunAsUser = &runAsUser
-		pod.Spec.SecurityContext.RunAsNonRoot = pointer.Bool(true)
+	if source.Spec.RunAsRoot {
+		pod.Spec.SecurityContext = &corev1.PodSecurityContext{
+			RunAsUser:    pointer.Int64(0),
+			RunAsNonRoot: pointer.Bool(false),
```
camilamacedo86, line 197: We can leave both empty.
What we need is to set AllowPrivilegeEscalation = true 

c/c @anik120 @bparees

camilamacedo86: /hold

bparees: the container doesn't actually need to run as UID 0?

anik120: does it hurt to be explicit about it them anyway?

anik120: also /hold cancel since there's already a WIP progress tag on the PR

anik120: /hold cancel

bparees: > As far as I know, we never had so far any UID set (either we did not set before uid =0) and we were not setting any USERID in the Dockerfile.

if nothing is set in the dockerfile nor in the pod/container, then the container will try to run as uid 0 as long as the user's SCC permissions allow it.  Which means catalogs are probably running as uid 0 today (on older OCP versions).

So i agree with @anik120 that explicitly setting it to uid 0 is appropriate, unless someone has tested and confirmed that it works when not run as uid 0.

camilamacedo86: @bparees,

> if nothing is set in the dockerfile nor in the pod/container, then the container will try to run as uid 0 as long as the user's SCC permissions allow it

I understand that in this case will be any user id but it will be a member from the root group
Also,  I do not think that `RunAsUser:  0,` can work on OCP and we have an implementation to ommit the runAsUser in downstream

So, ihmo we just need to make clear that no matter the user used it must have access to the root file system and only set AllowPrivilegeEscalation = true shows enough for upstream and downstream. 

However, if you folks check that it works and prefer to keep this implementation I am ok with too

/hold cancel

bparees: > I understand that in this case will be any user id but it will be a member from the root group

it depends on the what SCC it is admitted under.  again, i'm pretty sure that historically these pods ran as uid 0, and had to.  But someone should certainly confirm.

anik120: Fyi we're building both our [upstream](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/Dockerfile#L36) and [downstream](https://github.com/openshift/operator-framework-olm/blob/master/operator-lifecycle-manager.Dockerfile#L31) images with `runAsUser: 1001`

> we have an implementation to ommit the runAsUser in downstream 

could you point us to that implementation code? I'll be extremely surprised if I as an admin want to run my worload as root by setting runAsUser to 0 but some controller in openshift says "nope, none of that business here, so what you're admin?"

anik120: Also, for the namespaces that these CatalogSources will run, all SCC labels have to be disabled like we discussed, and the only label that'll be there for the namespace is PSA enforce:privileged, so I don't think there's any question of "workload `runAsUser` will be modified by an openshift controller" in the namespaces we care about for this scenario.

camilamacedo86: @anik120 

> Also, for the namespaces that these CatalogSources will run, all SCC labels have to be disabled
What do you mean by disabling SCC labels? 

> so I don't think there's any question of "workload runAsUser will be modified by an openshift controller" in the namespaces we care about for this scenario.

we spoke about it before, we cannot set values for `runAsUser ` in OCP (in downstream we leave RunAsUser empty and let the OCP assign a value.)

anik120: So this is completely different from all of the things you've mentioned in your comment. This PR is only for old CatalogSources that cannot be run in any of the openshift-* namespaces (since they're already being managed by us and have restricted profile enforce). We're dealing with the situation here where admins encounter old catalogsources they care about, so they'll create a new namespace, edit the namespace to enforce:privileged, and create the catalogsource in that namespace with `runAsRoot:true`. As simple as that. No OCP controller acting on that namespace (since we're explicitly making sure of that)

bparees: > Fyi we're building both our [upstream](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/Dockerfile#L36) and [downstream](https://github.com/openshift/operator-framework-olm/blob/master/operator-lifecycle-manager.Dockerfile#L31) images with runAsUser: 1001

we are now.  but my understanding is legacy catalog images didn't specify this, and thus run as uid==0 by default (i.e. when the pod doesn't otherwise specify a uid)

Can someone please confirm on a 4.10 install what the pod yaml looks like for the running catalog pod?  (not just what the pod yaml that the catalog controller creates, because you also have to consider what SCC admission will do during mutation)

Ultimately that is how we want the pod to look when this new field(RunAsRoot) is set to true

anik120: > we are now. but my understanding is legacy catalog images didn't specify this, and thus run as uid==0 by default (i.e. when the pod doesn't otherwise specify a uid)

that is a good point. And I launched a 4.10 cluster and this is how the yaml looks like: 

```oc get pod community-operators-d2btv -n openshift-marketplace -o yaml 
apiVersion: v1
kind: Pod
metadata:
  annotations:
    .
    openshift.io/scc: anyuid
    operatorframework.io/managed-by: marketplace-operator
  creationTimestamp: "2022-08-29T21:24:48Z"
  generateName: community-operators-
  labels:
    catalogsource.operators.coreos.com/update: ""
    olm.catalogSource: community-operators
    olm.pod-spec-hash: 656f95f5d4
  name: community-operators-d2btv
  namespace: openshift-marketplace
  ownerReferences:
  - apiVersion: operators.coreos.com/v1alpha1
    blockOwnerDeletion: false
    controller: false
    kind: CatalogSource
    name: community-operators
    uid: f2cf0647-138d-4f72-b1e0-172ff4acb672
  resourceVersion: "31463"
  uid: acdd15f7-c191-4cf5-8174-6196bbcd000b
spec:
  containers:
  - image: registry.redhat.io/redhat/community-operator-index:v4.10
    imagePullPolicy: Always
    livenessProbe:
      exec:
        command:
        - grpc_health_probe
        - -addr=:50051
      failureThreshold: 3
      initialDelaySeconds: 10
      periodSeconds: 10
      successThreshold: 1
      timeoutSeconds: 5
    name: registry-server
    ports:
    - containerPort: 50051
      name: grpc
      protocol: TCP
    readinessProbe:
      exec:
        command:
        - grpc_health_probe
        - -addr=:50051
      failureThreshold: 3
      initialDelaySeconds: 5
      periodSeconds: 10
      successThreshold: 1
      timeoutSeconds: 5
    resources:
      requests:
        cpu: 10m
        memory: 50Mi
    securityContext:
      capabilities:
        drop:
        - MKNOD
      readOnlyRootFilesystem: false
    startupProbe:
      exec:
        command:
        - grpc_health_probe
        - -addr=:50051
      failureThreshold: 15
      periodSeconds: 10
      successThreshold: 1
      timeoutSeconds: 1
    terminationMessagePath: /dev/termination-log
    terminationMessagePolicy: FallbackToLogsOnError
    volumeMounts:
    - mountPath: /var/run/secrets/kubernetes.io/serviceaccount
      name: kube-api-access-2vtdl
      readOnly: true
  dnsPolicy: ClusterFirst
  enableServiceLinks: true
  imagePullSecrets:
  - name: community-operators-dockercfg-bjn7x
  nodeName: ci-ln-prs6mt2-72292-7w4cb-master-0
  nodeSelector:
    kubernetes.io/os: linux
    node-role.kubernetes.io/master: ""
  preemptionPolicy: PreemptLowerPriority
  priority: 2000000000
  priorityClassName: system-cluster-critical
  restartPolicy: Always
  schedulerName: default-scheduler
  securityContext:
    seLinuxOptions:
      level: s0:c17,c4
  serviceAccount: community-operators
  serviceAccountName: community-operators
  terminationGracePeriodSeconds: 30
  tolerations:
  - effect: NoSchedule
    key: node-role.kubernetes.io/master
    operator: Exists
  - effect: NoExecute
    key: node.kubernetes.io/unreachable
    operator: Exists
    tolerationSeconds: 120
  - effect: NoExecute
    key: node.kubernetes.io/not-ready
    operator: Exists
    tolerationSeconds: 120
  - effect: NoSchedule
    key: node.kubernetes.io/memory-pressure
    operator: Exists
  volumes:
  - name: kube-api-access-2vtdl
    projected:
      defaultMode: 420
      sources:
      - serviceAccountToken:
          expirationSeconds: 3607
          path: token
      - configMap:
          items:
          - key: ca.crt
            path: ca.crt
          name: kube-root-ca.crt
      - downwardAPI:
          items:
          - fieldRef:
              apiVersion: v1
              fieldPath: metadata.namespace
            path: namespace
      - configMap:
          items:
          - key: service-ca.crt
            path: service-ca.crt
          name: openshift-service-ca.crt
status:
  .
  .
  .
```

so I guess this means we leave out setting runAsUser to 0 and `runAsRoot: true`, and we didn't have `allowPrivilegeEscalation` either so I'm guessing we don't need that either.

bparees: so it lands in the "Anyuid" SCC which makes sense in that it's necessary so the container can run as UID 0, though i'm not entirely clear on *why* that SCC is selected given the securityContext the pod defines.

but i'll say this, it will quickly become apparent when you go to test this, whether or not you are defining the pod correctly such that a legacy catalog image can run successfully on a 4.12 cluster (in a PSA-privileged NS). 

I do still suspect you're going to have to set runAsUser: 0, in the end, to make it work.  Otherwise the pod will get assigned to the restricted-v2 SCC and get assigned a random uid, i think.  But do the experiment, if nothing else you'll learn some things.

as for whether you need to set "runAsNonRoot=false" on the pod (that is the official name of the securityContext for the field, for the record, not "runAsRoot"), you may not need to, but you probably should, if indeed the container is going to run as root.  No reason not to be explicit.

.github/workflows/verify.yml
```diff
@@ -12,7 +12,7 @@ jobs:
       - uses: actions/checkout@v1
       - uses: actions/setup-go@v2
         with:
-          go-version: '~1.18'
+          go-version: '~1.19'
```
perdasilva, line 15: should we put updating to go1.19 on another PR to keep the scope tight?

pkg/controller/registry/reconciler/reconciler.go
camilamacedo86: @bparees,

> I do still suspect you're going to have to set runAsUser: 0, in the end, to make it work. Otherwise the pod will get assigned to the restricted-v2 SCC and get assigned a random uid, i think. But do the experiment, if nothing else you'll learn some things.

If we set AllowPrivilegeEscalation = true then is all sorted out
The restricted-v2 SCC policy cannot be applied and we granted the right permissions
I do not think that we can set RunAsUser = 0 at all. 
Therefore, it will work as before ( we will have a anyuid that has access to the root filesystem )

c/c @anik120

benluddy:
What does this field mean for CatalogSources that don't have any managed pods?

pkg/controller/registry/reconciler/reconciler.go
bparees: I brought up a 4.10 cluster and see that the default marketplace catalogs are running as uid 1001, not 0 (as @camilamacedo86 alluded to).  So among other things, i think that means "runAsRoot" is the wrong name for the config field.  And yes, it means we should not be setting `runAsUser: 0` in the pod.  (I don't think it would fail, since the catalog controller has permission to use the "anyuid" SCC, so it can probably create pods that run as uid 0, but it doesn't seem like that's what we want it to do here).

perdasilva:
@operator-framework/team-fbc test

pkg/controller/registry/reconciler/reconciler.go
anik120: @camilamacedo86 @bparees updated the PR, not running as 0 anymore, renamed the field to `requirePrivilegedAcess`, and only setting AllowPrivilegeEscalation = true and readOnlyRootFilesystem: false when `requirePrivilegedAccess` is set to true. PTAL.

deploy/chart/crds/0000_50_olm_00-catalogsources.crd.yaml
```diff
@@ -76,6 +76,9 @@ spec:
                       type: object
                       additionalProperties:
                         type: string
+                    requirePrivilegedAccess:
+                      description: RequirePrivilegeAccess when set to true, allows the containers in the pod to run in privileged mode. This should only be set to true when using older catalog images which could be run in restricted mode, and requires privileged access.
```
bparees, line 80:
```suggestion
                      description: RequirePrivilegeAccess when set to true, allows the containers in the pod to run in privileged mode. This should only be set to true when using older catalog images which could not be run in restricted mode, and require privileged access.
```

pkg/controller/registry/reconciler/reconciler.go
```diff
@@ -182,11 +183,7 @@ func Pod(source *operatorsv1alpha1.CatalogSource, name string, image string, saN
 						},
 					},
 					SecurityContext: &corev1.SecurityContext{
-						ReadOnlyRootFilesystem:   pointer.Bool(false),
-						AllowPrivilegeEscalation: pointer.Bool(false),
-						Capabilities: &corev1.Capabilities{
-							Drop: []corev1.Capability{"ALL"},
-						},
```
bparees, line 189: you can still drop the caps in all cases, right?  or did legacy catalog images require some additional caps?

pkg/controller/registry/reconciler/reconciler.go
```diff
@@ -195,19 +192,29 @@ func Pod(source *operatorsv1alpha1.CatalogSource, name string, image string, saN
...
 		},
 	}
 
-	if runAsUser > 0 {
-		pod.Spec.SecurityContext.RunAsUser = &runAsUser
-		pod.Spec.SecurityContext.RunAsNonRoot = pointer.Bool(true)
+	if source.Spec.GrpcPodConfig != nil && source.Spec.GrpcPodConfig.RequirePrivilegedAccess {
+		pod.Spec.Containers[0].SecurityContext.AllowPrivilegeEscalation = pointer.Bool(true)
+	} else {
+		log.Printf("(anik120) setting everything")
```
bparees, line 202: leftover debug

pkg/controller/registry/reconciler/reconciler.go
```diff
@@ -195,19 +192,29 @@ func Pod(source *operatorsv1alpha1.CatalogSource, name string, image string, saN
...
-			},
 			ServiceAccountName: saName,
 		},
 	}
 
-	if runAsUser > 0 {
-		pod.Spec.SecurityContext.RunAsUser = &runAsUser
-		pod.Spec.SecurityContext.RunAsNonRoot = pointer.Bool(true)
+	if source.Spec.GrpcPodConfig != nil && source.Spec.GrpcPodConfig.RequirePrivilegedAccess {
+		pod.Spec.Containers[0].SecurityContext.AllowPrivilegeEscalation = pointer.Bool(true)
```
bparees, line 200: @perdasilva i think you had done the deepest investigation on what permissions our legacy catalogs/opm binary required to be able to run on a 4.12 cluster (in the process of you fixing the issues so we didn't require those permissions anymore), can you take a look at whether this seems right?  It's not what i would have expected.

pkg/controller/registry/reconciler/reconciler.go
```diff
@@ -195,19 +192,29 @@ func Pod(source *operatorsv1alpha1.CatalogSource, name string, image string, saN
...
+		pod.Spec.Containers[0].SecurityContext.AllowPrivilegeEscalation = pointer.Bool(false)
+		pod.Spec.Containers[0].SecurityContext.Capabilities = &corev1.Capabilities{
+			Drop: []corev1.Capability{"ALL"},
+		}
+		pod.Spec.SecurityContext = &corev1.PodSecurityContext{
+			SeccompProfile: &corev1.SeccompProfile{
+				Type: corev1.SeccompProfileTypeRuntimeDefault,
+			},
+		}
+		if runAsUser > 0 {
```
bparees, line 212: when running legacy catalogs you're not going to respect this value?  (who sets this? where does it get set? why would they set it?)

pkg/controller/registry/reconciler/reconciler.go
anik120: legacy catalogs did not require additional capabilities, but were the pods created by mentioning "drop all caps". So if our goal is to mimic the 4.10 configs, then what's there now is good. If that isn't our goal then I can test to make sure dropping all capabilities is fine before adding this back in.

bparees: > legacy catalogs did not require additional capabilities, but were the pods created by mentioning "drop all caps". So if our goal is to mimic the 4.10 configs, then what's there now is good.

i'm confused.  in 4.10 did the catalog pods specify "drop all caps" or not?  We should align with what 4.10 did, imho.

anik120: My apologies, I just realized I had a typo in my last comment, edited it so should make more sense now. 
Just to reiterate with more info: we DID NOT specify "drop all caps" for the pods: https://github.com/openshift/operator-framework-olm/blob/release-4.10/staging/operator-lifecycle-manager/pkg/controller/registry/reconciler/reconciler.go#L167-L169

pkg/controller/registry/reconciler/reconciler.go
anik120: The value for this is dependent on an [argument to the container](https://github.com/openshift/operator-framework-olm/pull/349/files#diff-3095edb22e78d950b5bb2d52942b13bad32f4b097451d907aa04cb69ad658985R66), that is determined [here](https://github.com/openshift/operator-framework-olm/pull/349/files#diff-3a59260edccecabbc7c818c855c7b35842572551f5039cf4e6590f939beb81bfR87-R90). We decided to introduce that argument so that we can set that to `true` for upstream deployment, and `false` for downstream deployment, and then set a default userID for when that flag's value is `true`, because we wanted to keep control of what userID the container's run with (in absence of a requirement for a configurable userID). 


For legacy catalogs, [we never set these fields](https://github.com/openshift/operator-framework-olm/blob/release-4.10/staging/operator-lifecycle-manager/pkg/controller/registry/reconciler/reconciler.go#L116-L180).  So even if the flag `set-workload-user-id` is set to `true`, when `catsrc.spec.grpcPodConfig.requirePrivilegedAccess` is set to true, that is an indication that we don't need to set the userID to the default value (instead, let whatever used to happen in 4.10, happen. Next comment has logs for an example)

anik120:
```
$ k get ns test-ns -o yaml | grep labels: -A 3
  labels:
    kubernetes.io/metadata.name: test-ns
    pod-security.kubernetes.io/enforce: privileged
    pod-security.kubernetes.io/enforce-version: v1.24

$ k get pods -n test-ns 
NAME                READY   STATUS    RESTARTS   AGE
ditto-index-4cnmf   1/1     Running   0          23h

$ k get pods -n test-ns -o yaml | grep securityContext -A 2
      securityContext:
        allowPrivilegeEscalation: true
        readOnlyRootFilesystem: false
--
    securityContext: {}
    serviceAccount: ditto-index
    serviceAccountName: ditto-index

$ k get pods -n test-ns -o yaml | grep containerStatuses -A 11
    containerStatuses:
    - containerID: containerd://00e416a3797d00b2b29a4059cec21511134b1e113afd0acfe629c75ac5c0cd19
      image: docker.io/anik120/ditto-index-opm:v1.21.0
      imageID: docker.io/anik120/ditto-index-opm@sha256:36a8daca4365b24bb8699e5f89678dad36a7b4a01127963bf289afd67ecde96a
      lastState: {}
      name: registry-server
      ready: true
      restartCount: 0
      started: true
      state:
        running:
          startedAt: "2022-09-06T19:30:32Z"

```

pkg/controller/registry/reconciler/reconciler.go
bparees: got it.

pkg/controller/registry/reconciler/reconciler.go
bparees: got it

deploy/chart/crds/0000_50_olm_00-catalogsources.crd.yaml
perdasilva: also you need a comma before 'which' (or you can switch to 'that')

pkg/controller/registry/reconciler/reconciler.go
perdasilva: Here's what it came down to:
```
securityContext:
      allowPrivilegeEscalation: false
      capabilities:
        drop:
        - MKNOD
      readOnlyRootFilesystem: false
```
and
```
securityContext:
    runAsNonRoot: false
    runAsUser: 0
    seccompProfile:
      type: RuntimeDefault
```

This ran on a `baseline` enforced namespace. Here's something curious though, @bparees, I only managed to change the PSA enforcement profile on a (new) `openshit-` namespace. I.e.

```
$ oc create ns test
$ oc describe ns test
...
Labels:  kubernetes.io/metadata.name=test
              pod-security.kubernetes.io/enforce=restricted
              pod-security.kubernetes.io/enforce-version=v1.24
...
$ oc label ns test pod-security.kubernetes.io/enforce=baseline --overwrite
$ oc describe ns test
...
Labels:  kubernetes.io/metadata.name=test
              pod-security.kubernetes.io/enforce=restricted
              pod-security.kubernetes.io/enforce-version=v1.24
...
```

So, I think we may still have problems even with the api override =S

anik120: Fyi @perdasilva not sure if you saw [this conversation](https://github.com/operator-framework/operator-lifecycle-manager/pull/2845#discussion_r957031708), but it might help fill in the context for why I switched from setting `runAsUser: 0` to not setting it at all. Given the fact that we're saying to run legacy catalogs, admins need to make their namespaces look like what it was pre PSA restricted world, it does make sense to make the registry pods look like what they were pre PSA restricted world too, and in that world we did not set `runAsUser:0`, unless there are specific reasons why we should be setting `runAsUser:0` now.

pkg/controller/registry/reconciler/reconciler.go
```diff
@@ -182,11 +182,7 @@ func Pod(source *operatorsv1alpha1.CatalogSource, name string, image string, saN
 						},
 					},
 					SecurityContext: &corev1.SecurityContext{
-						ReadOnlyRootFilesystem:   pointer.Bool(false),
-						AllowPrivilegeEscalation: pointer.Bool(false),
```
camilamacedo86, line 186: Also, because we need to read root file system, they away to grant permissions to that is setting true to AllowPrivilegeEscalation

And RunAsUser=0 is not required ( we need an user that have permissions to do what needs to be done as before, and if the bt setting true there the user will be part of the root group and have the same permissions.

Btw, RunAsUser=0 might not work on OCP (it will depends of the namespace and it probably will raise the error invalid value in the range)

c/c @perdasilva

bparees: what does "AllowPrivilegeEscalation" have to do w/ reading the root filesystem?

allow privilege escalation means you can run a setuid binary or otherwise allow a process to "become root", does the opm registry binary do so?  

I would simply expect the index.db to be readable (by any uid), or in general we should be ensuring that the container is run w/ the appropriate uid and fsgroup to allow us to read (and write, if necessary) to the appropriate directories.  None of that should require privilege escalation.


```
allowPrivilegeEscalation: Controls whether a process can gain more privileges than its parent process.
```

https://kubernetes.io/docs/tasks/configure-pod-container/security-context/

pkg/controller/registry/reconciler/reconciler.go
bparees: > So, I think we may still have problems even with the api override =S

@perdasilva i believe you need to explicitly label the NS as having the labelsyncer disabled, before you can set the PSA labels manually.  It should be the same label you are setting on openshift-* NSes to enable the labelsyncer there (except you'd set it to false, of course, here)

non payload openshift-* NSes default to labelsyncer on (so you need to label the NS appropriately to disable it)
openshift-* NSes default to labelsyncer off (so you need to label the NS appropriately to enable it)

pkg/controller/registry/reconciler/reconciler.go
anik120: That'd mean we don't need to specify `AllowPrivilegeEscalation` at all then, since opm doesn't require to be root to read or write to the sqllite db, and also in 4.10 we didn't specify `AllowPrivilegeEscalation` either (and opm not requiring to be root hasn't changed since 4.10, and is unlikely to change either).

deploy/chart/crds/0000_50_olm_00-catalogsources.crd.yaml
anik120: looks like we have agreement on the field name, opened a PR in api to incorporate that: https://github.com/operator-framework/api/pull/261

bparees: didn't we just establish that we don't need to run even legacy catalogs with privileged access?

anik120: `requirePrivilegedAccess` is different than `AllowPrivilegedEscalation` no? `requirePrivilegedAccess` is saying "don't add the usual securityContext fields that are being added to the registry pods, coz I need more privileges than that". Which is what this PR is about. If the name is still confusing, I'm all ears for suggestions.

anik120: open to suggestions for the description also to make it clear what we're trying to do here.

pkg/controller/registry/reconciler/reconciler.go
anik120: Fyi was updating the tests to use image hosted in the quay.io/olmtest registry instead of my personal registry, so took that opportunity to remove `AllowPrivilegeEscalation:true` too. cc: @perdasilva now I'm happy with the PR going in :)

deploy/chart/crds/0000_50_olm_00-catalogsources.crd.yaml
bparees: The problem is that "privileged=true" is another container setting, which we also don't need, but a name like this implies that's what we'd be setting on the pod (running the catalog pod as privileged).

Here's my current understanding of what used to happen:

We historically created the catalog pod w/o specifying anything about the desired user to run it as, see:
https://github.com/operator-framework/operator-lifecycle-manager/pull/2845#discussion_r957816689

The pods were historically admitted under the "anyuid" SCC, probably because the catalog controller has access to that SCC, and that SCC has priority over all other SCCs.

In 4.12, OLM must have dropped the catalog controller's permission to use the anyuid SCC(can someone confirm?).  That means now the pods get created under the restricted-v2 SCC, which means they get "runAsRange", among other things and no longer run as the default image user.

This breaks legacy catalogs because legacy catalogs need to run as the default image user (*not necessarily root/uid 0*).  So we are trying to give users a way to say "run this pod as the default image user".  That, as i understand it, should be the *only* securityContext setting that needs to be different between running a "legacy" catalog image, and a new one.

Unfortunately I think the only way we can make this happen is for the catalog controller to have 2 service accounts, one which has access to the anyuid SCC and one which does not.  We'd create legacy catalog pods using the one that has access to the anyuid SCC.

It's probably the best way to ensure the pod gets admitted under the correct SCC, and it's only by admitting the pod under the anyuid SCC that we can ensure the image runs as its default user (because we need to ensure the created pod has the runAsUser field left empty, which is what running under the anyuid SCC accomplishes).

cc @perdasilva @joelanford

bparees: And assuming i got all of the above correct, then an appropriate name for the field might be "runAsImageUser", but i'm open to other ideas.

anik120: Yea this is very confusing to me since I am not aware of any adding/dropping of "catalog controller's permission to use anyuid SCC". [This](https://github.com/openshift/operator-framework-olm/blob/master/manifests/0000_50_olm_01-olm-operator.serviceaccount.yaml) is the SA the catalog-controller uses, which hasn't changed at all for since at least 4.8, expect for a IBM related label. 

```
k get pod catalog-operator-78b957fb94-rptfn -n olm -o yaml | grep serviceaccount
    - mountPath: /var/run/secrets/kubernetes.io/serviceaccount
  serviceAccount: olm-operator-serviceaccount
  serviceAccountName: olm-operator-serviceaccount
```

And we've admitted the catalog controller (not the registry pods[1]) in a [namespace labled as SCC:anyuid](https://github.com/openshift/operator-framework-olm/blob/release-4.6/manifests/0000_50_olm_00-namespace.yaml#L20) since as far back as 4.6 at least, and [that hasn't changed either](https://github.com/openshift/operator-framework-olm/blob/master/manifests/0000_50_olm_00-namespace.yaml#L8). 

[1] [namespace registry pods are deployed to that is not labeled as SCC:anyuid](https://github.com/operator-framework/operator-marketplace/blob/master/manifests/01_namespace.yaml)

anik120: If our goal is to simply allow legacy catalogs to run in a namespace that is not `restricted`, all we need to do is not add the securityContexts being added to allow registry pods to run in `restricted` namespace, and run that catalog in a `privileged` namespace no? Why do we need a second service account, when what is being used now hasn't changed since 4.6?

bparees: Ok. I think the change that caused the pods to stop running in the "anyuid" SCC was the addition of `seccompProfile: runtime/default`.  To use the "anyuid" SCC, you need to leave that empty (which you are doing)

So overall i think the PR is correct, but the field name that allows users to opt into this behavior still doesn't reflect what it actually does.  It's not running anything as privileged.  It's allowing the image to run as whatever user it wants (which could include root, but still won't be a privileged container)

So i still think "runAsImageUser" is a reasonable name for the config field.  that is the meaningful delta.

That or just punting entirely and calling it "useLegacyCatalogPodSecurityContextConfiguration"

anik120: I'm not sure I can look the user in the eye when asking them to use "useLegacyCatalogPodSecurityContextConfiguration". And `runAsImageUser` also sounds like it expects users to fill in a userID. Are we opposed to `isLegacyCatalog`? Is that too general? (we could use the description to define what counts as Legacy for us: a catalog image built with opm < v1.23.2)

bparees: I think that is my main caution around calling it "legacyCatalog"  (it's legacy today, in the future what we consider a modern catalog may also be legacy). (admittedly this problem exists in my more longwinded name suggestion above also).

I'd also rather tie the name to the way it dictates the config of the pod that is constructed, rather than tying it to the images you should use it with.

how about `useLegacySecurityContext` or just `legacySecurityContext`?

anik120: I can get behind `useLegacySecurityContext`. Slightly longer than what'd be perfect, but it conveys the entire purpose of it, and it can be a bool too. I'll go ahead and update the api PR with that if we're all happy with that.

bparees: one other suggestion, since upstream k8s at least now rejects boolean api fields as a best practice:

make it an enum, named "securityContextConfig", with two values (for now):
`legacy` - opt-in, runs the catalog as the default image user
`restricted` - default, runs the catalog under PSA-restricted requirements which includes a randomly assigned uid on openshift.

That gives us some flexibility if we need to change behavior in the future by adding new enum values.

anik120: > since upstream k8s at least now rejects boolean api fields as a best practice

TIL. Okay I'll update here once I'm done updating the api PR with an enum.

pkg/controller/registry/reconciler/reconciler.go
perdasilva: doh! should have inferred that :facepalm:

perdasilva: > Fyi @perdasilva not sure if you saw [this conversation](https://github.com/operator-framework/operator-lifecycle-manager/pull/2845#discussion_r957031708), but it might help fill in the context for why I switched from setting `runAsUser: 0` to not setting it at all. Given the fact that we're saying to run legacy catalogs, admins need to make their namespaces look like what they were pre PSA restricted world, it does make sense to make the registry pods look like what they were pre PSA restricted world too, and in that world we did not set `runAsUser:0`, unless there are specific reasons why we should be setting `runAsUser:0` now.

So, not setting `runAsUser` seemed to push the restricted-v2 scc profile causing the perm issue. If we don't want to set it to 0, we'll probably need to set the `fsGroup` to 0 in lieu of that. I've also tested by removing runAsUser and runAsNonRoot and setting AllowPriviledgeEscalation to true and that worked as well. Maybe that's the best way forward, then.

bparees: > So, not setting runAsUser seemed to push the restricted-v2 scc profile causing the perm issue. If we don't want to set it to 0, we'll probably need to set the fsGroup to 0 in lieu of that. 

if you do not set the seccompprofile, and you have access to the anyuid SCC, i believe you'll get the anyuid SCC which allows you to leave the runAsUser unset, and the container will run w/ the image's default UID, which i believe is what you want.

deploy/chart/crds/0000_50_olm_00-catalogsources.crd.yaml
anik120: @bparees updated the PR: changed the field name to `SecurityContextConfig` and the accepted values are `legacy` or `restricted`, with `restricted` being default. PTAL.

deploy/chart/crds/0000_50_olm_00-catalogsources.crd.yaml
```diff
@@ -79,6 +79,9 @@ spec:
                     priorityClassName:
                       description: If specified, indicates the pod's priority. If not specified, the pod priority will be default or zero if there is no default.
                       type: string
+                    securityContextConfig:
+                      description: 'SecurityContextConfig can be one of `legacy` or `restricted`. The CatalogSource''s pod is either injected with the right pod.spec.securityContext and pod.spec.container[*].securityContext values to allow the pod to run in Pod Security Admission(PSA) controller''s `restricted` mode, or doesn''t set these values at all so that the pod can be run in PSA `privileged` namespace. By default, SecurityContextConfig is set to `restricted`. If a value other than `legacy` or `restricted` is used as input, the default value is set instead of using the incorrect value When using older catalog images, which could not be run in restricted mode, the SecurityContextConfig should be set to `legacy`. More info about PSA can be found here: https://kubernetes.io/docs/concepts/security/pod-security-admission/'
```
bparees, line 83: setting a value other than `restricited` or `legacy` should be an error, not fallback to the default.

leaving it empty should fallback to the default.

bparees:
```suggestion
                      description: 'SecurityContextConfig can be one of `legacy` or `restricted`. The CatalogSource''s pod is either injected with the right pod.spec.securityContext and pod.spec.container[*].securityContext values to allow the pod to run in Pod Security Admission(PSA) controller''s `restricted` mode, or doesn''t set these values at all in which case the pod can only run in PSA `privileged` namespaces. By default, SecurityContextConfig is set to `restricted`. If the value is unspecified, the default value of `restricted` is used.  Specifying any other value will result in a validation error.  When using older catalog images, which could not be run in restricted mode, the SecurityContextConfig should be set to `legacy`. More information about PSA can be found here: https://kubernetes.io/docs/concepts/security/pod-security-admission/'
```

anik120:
Looks like we've finally converged to a spot where we're all on the same page. Let's merge this so that we can start the downstreaming process, and if we discover we need to make changes to the configs in this PR for downstream, we can come back upstream and make those changes. 

In lieu of that, will need the labels for https://github.com/operator-framework/api/pull/261 cc: @perdasilva

pkg/controller/registry/reconciler/reconciler.go
```diff
@@ -195,19 +191,26 @@ func Pod(source *operatorsv1alpha1.CatalogSource, name string, image string, saN
...
-				},
-			},
 			ServiceAccountName: saName,
 		},
 	}
 
-	if runAsUser > 0 {
-		pod.Spec.SecurityContext.RunAsUser = &runAsUser
-		pod.Spec.SecurityContext.RunAsNonRoot = pointer.Bool(true)
+	if source.Spec.GrpcPodConfig == nil || source.Spec.GrpcPodConfig.SecurityContextConfig != operatorsv1alpha1.Legacy {
```
bparees, line 198: this should be "if config==nil || config==restricted"

you don't want to treat unknown values as if they were "restricted", you want to treat unknown values as an error.

pkg/controller/registry/reconciler/reconciler.go
```diff
@@ -182,11 +182,7 @@ func Pod(source *operatorsv1alpha1.CatalogSource, name string, image string, saN
 						},
 					},
 					SecurityContext: &corev1.SecurityContext{
-						ReadOnlyRootFilesystem:   pointer.Bool(false),
-						AllowPrivilegeEscalation: pointer.Bool(false),
-						Capabilities: &corev1.Capabilities{
-							Drop: []corev1.Capability{"ALL"},
-						},
+						ReadOnlyRootFilesystem: pointer.Bool(false),
```
bparees, line 185: Do "restricted" mode catalogs actually need a writeable root filesystem?

pkg/controller/registry/reconciler/reconciler.go
bparees: (so you also need an else here to handle unknown values.  Yes validation should prevent it, but if someone adds a new valid enum and doesn't fix up this logic, you want to catch it.)

pkg/controller/registry/reconciler/reconciler.go
anik120: I'm guessing "restricted" mode doesn't change the fact the [database might have to be copied to a different location in a file system ](https://github.com/operator-framework/operator-lifecycle-manager/pull/2167). Messing with this sounds like a potential to reintroduce a bug that PR was trying to solve for.

bparees: > I'm guessing "restricted" mode doesn't change the fact the https://github.com/operator-framework/operator-lifecycle-manager/pull/2167

I was thinking we copy to /tmp now and that /tmp might be mounted separately, but looking at a catalog pod it appears /tmp is part of the root filesystem, so i guess not (even if we are copying to /tmp)

So yeah, we can close this comment.

pkg/controller/registry/reconciler/reconciler.go
anik120: so spec.GrpcPodConfig.SecurityContextConfig is marked as `optional`, which means most of the CatalogSources created will not have the "GrpcPodConfig" struct populated, so config==restricted will actually not be true. 

For example, this is the yaml of the operatorhub catalog applied to the cluster on local installation in Kind: 

```
$ k get catsrc -n olm -o yaml 
apiVersion: v1
items:
- apiVersion: operators.coreos.com/v1alpha1
  kind: CatalogSource
  metadata:
    creationTimestamp: "2022-09-14T17:11:29Z"
    generation: 1
    name: operatorhubio-catalog
    namespace: olm
    resourceVersion: "757"
    uid: 54db9c12-be5e-4b0b-9e16-3d1d80acaf54
  spec:
    displayName: Community Operators
    image: quay.io/operatorhubio/catalog:latest
    publisher: OperatorHub.io
    sourceType: grpc
    updateStrategy:
      registryPoll:
        interval: 60m
  status:
    connectionState:
      address: operatorhubio-catalog.olm.svc:50051
      lastConnect: "2022-09-14T17:11:59Z"
      lastObservedState: READY
    registryService:
      createdAt: "2022-09-14T17:11:29Z"
      port: "50051"
      protocol: grpc
      serviceName: operatorhubio-catalog
      serviceNamespace: olm
kind: List
metadata:
  resourceVersion: ""
  selfLink: ""
```

I'm guessing that means when a third value is introduced, grpcpodConfig.SecurityContextConfig should be marked as `required` and then this can be changed to explicitly check config==restricted. 

Coz the other option is "config==restricted" || config==nil", which doesn't sound that explicit anyway. Looks like even with defining a default value for it, if the top level grpcPodConfig isn't defined at all, the entire struct remains nil. 

I've added a [comment](https://github.com/operator-framework/operator-lifecycle-manager/pull/2845/files#diff-fffdeef1fc140a5dc5dc92dda323f567a6e46fc2ecbb0b91ba907acd02bde50dR198-R204) to prevent someone accidentally introducing faulty logic when a third enum is added

pkg/controller/registry/reconciler/reconciler.go
perdasilva: For sqlite, yes. opm will still create the database copy (but in the /tmp directory now =D)

PR #2862, opened 2022-09-15
(bug) Fix service account token secret reference

awgreene:
Problem: The filterSecretsBySAName function attempts to identify all service account token secrets related to a serviceAccount. To do so, the filterSecretsBySAName function uses a range-for loop to iterate over entries in the secrets argument. If a valid service account token secret is found, a pointer to the range-for loop's value variable is added to a map of results. Unfortunately, if a valid entry is found in the middle of the list of secrets, the value returned by the range-for loop is updated, causes the entry in the map to change.

Solution: Add a pointer to the actual secret instead of the range-for loop's value variable.


pkg/lib/scoped/token_retriever.go
```diff
@@ -81,11 +81,11 @@ func getAPISecret(logger logrus.FieldLogger, kubeclient operatorclient.ClientInt
 // specific ServiceAccount via annotations kubernetes.io/service-account.name
 func filterSecretsBySAName(saName string, secrets *corev1.SecretList) map[string]*corev1.Secret {
 	secretMap := make(map[string]*corev1.Secret)
-	for _, ref := range secrets.Items {
+	for i, ref := range secrets.Items {
 		annotations := ref.GetAnnotations()
 		value := annotations[corev1.ServiceAccountNameKey]
 		if value == saName {
-			secretMap[ref.Name] = &ref
+			secretMap[ref.Name] = &secrets.Items[i]
```
awgreene, line 88: We _could_ use a deepCopy here:
* Pro: Object in the client cache isn't returned. As it is, users should never edit the secrets returned by this function for risk of concurrent map writes. Also prevents changes to these secrets if updated on cluster.
* Con: Memory footprint.

pkg/lib/scoped/token_retriever.go
```diff
@@ -81,11 +81,11 @@ func getAPISecret(logger logrus.FieldLogger, kubeclient operatorclient.ClientInt
...
 		}
 	}
```
awgreene, line 90: I created this [go playground code](https://go.dev/play/p/PosGriATixw) to showcase the issue.

pkg/lib/scoped/token_retriever.go
tylerslaton: I'd vote for DeepCopy here since we potentially leave ourselves open to race conditions in the future where the secrets list gets changed on cluster in the middle of this operation. In that case, we could output a reference that is invalid. How big of an extra memory footprint are we thinking here?

joelanford: The canonical solution I've seen (and typically use) for this is:
```go
for _, ref := range secrets.Items {
	ref := ref
	...
}

Comment (awgreene) [reply pkg/lib/scoped/token_retriever.go:88]: I'm for this.

Comment (awgreene) [reply pkg/lib/scoped/token_retriever.go]: Updated per @joelanford's suggestion, but moved it inside the if statement.

**Comment by tylerslaton** [comment]
Looks like we need to rebase. Then I'll LGTM.

**Review by timflannagan — COMMENTED** [review]
Is there an issue link tracking this bug or was this discovered naturally?

**Comment by tylerslaton** [comment]
@timflannagan Yes, it is 2094303 in BZ and that produced a new bug in Jira with an ID of 1428.

File: pkg/lib/scoped/token_retriever.go [hunk 3]
```diff
@@ -82,6 +82,12 @@ func getAPISecret(logger logrus.FieldLogger, kubeclient operatorclient.ClientInt
 func filterSecretsBySAName(saName string, secrets *corev1.SecretList) map[string]*corev1.Secret {
 	secretMap := make(map[string]*corev1.Secret)
 	for _, ref := range secrets.Items {
+		// Avoid referencing the "ref" created by the range-for loop as
+		// the secret stored in the map will change if there are more
+		// entries in the list of secrets that the range-for loop is
+		// iterating over.
+		ref := ref
```
grokspawn, line 89: I really prefer not forcing unintuitive inner/outer variable collisions to avoid this.  
I'll LGTM, but IMHO the better approach is to use a non-colliding loop-scoped variable with a note.  
I concede that this yields a better code delta, but if it requires a substantial comment like the one provided, is it really an improvement?

awgreene: This has already merged, but I 100% agree with the sentiment.

In this case, @joelanford pointed out that this is the canonical solution that he usually sees to address this intuitive problem. As you said, we could create our own way of avoiding the issue and forego the comment, but I would rather:
- Adhere to standards found in multiple go projects.
- Not "hide" the problem by coming up with our own solution.
- Document the reason for the approach with a comment so our team is aware of this pattern and can apply it when writing/reviewing code later.

Again, I don't know that I would have supported this approach had I been there when the pattern was created, but now that it is established I will respect it.

Issue #2872, opened 2022-10-10
catalog-operator in CrashLoopBackOff


rohasharma:
catalog-operator Pod is in CrashLoopBackOff state post deploying OLM using helm command. 
The reason shown in the logs are:
Error: unknown flag: --set-workload-user-id
unknown flag: --set-workload-user-id

Irrespective if we set the flag in values as True or False same behaviour is observed. The only way to proceed right now is to remove the else statement from 0000_50_olm_08-catalog-operator.deployment.yaml and then set setWorkloadUserID: true as false in values.yaml.

My kubernetes is running v1.23.10.

Thanks
Rohan Roy Sharma

dmesser:
I think this may be due to the fact that the chart default value for the catalog operator image points to quay.io/operator-framework/olm:master - this tag hasn't been updated since the last 6 months. Your workaround would be to use any of the newer tags, e.g. quay.io/operator-framework/olm:v0.22

@anik120 The build job for OLM that generates an image according to branch name on quay.io is disabled. I guess you may want to either re-enable or make the helm chart part of the upstream release so it gets a versioned image.

rohasharma:
Thank you so much @dmesser for the response. I was in-fact hardcoding the version from master to v0.22 in past but along with it I did remove the flag from values.yaml as a workaround. Since we are automating the OLM deployment using flux we want to make least possible change in the git repo so that we can import it as it is. 

Since we are on this topic of using default templates/values from this repo in our environment there is one more thing I want to bring in notice here: 
These template files has flag set for runAsNonRoot: true
`chart/templates/0000_50_olm_07-olm-operator.deployment.yaml
chart/templates/_packageserver.deployment-spec.yaml
chart/templates/0000_50_olm_08-catalog-operator.deployment.yaml`
However, if we run the helm install as it is, it fails the deployment and as a work around we need to add 
`runAsUser: 1000`
Should we add it in our template files so that we don't see any failure with pod deployments?

anik120:
[...]

PR #2873, opened 2022-10-12
bump OLM version to v0.22.0

anik120:

**Architectural changes:**

**Testing remarks:**

**Reviewer Checklist**


OLM_VERSION
```diff
@@ -1 +1 @@
-0.19.0
+0.22.0
```
timflannagan, line 1: IMO, this file is tech debt and doesn't affect anything upstream outside of the related-related Makefile targets. Do we have any context for why we're bumping this version vs. working towards removing it entirely from the repository and deriving any versioning information from git commands?

anik120: @timflannagan I just linked this PR in issue #2872 that should have the context. We need to have the v0.22.0 manifests to unblock our upstream consumers, till we figure out an alternative.

anik120: Or do you mean it's sufficient to keep hardcoding the version in the target instead of have it fetch from the OLM_VERSION file till we figure out the alternative?

I'm fine with anything, just as long as we release the v0.22.0 manifests (and don't keep missing out on releasing manifests for the future releases)

anik120: I've asked if the release artifacts can be used instead of using the charts

/hold 

till we hear back.

anik120: @timflannagan looks like there are consumers who have use cases for the helm charts https://github.com/operator-framework/operator-lifecycle-manager/issues/2872#issuecomment-1276426894

I don't know enough about FluxCD to pose any argument for/against that, so I'm inclined to just release the v0.22.0 manifests (and keep releasing future manifests) in tree. 

/hold cancel

perdasilva:
clsing due to inactivity - please re-open if necessary

PR #2878, opened 2022-10-24
bugfix: Update service deletion to retry with timeout

perdasilva:

Adds a retry with timeout when attempting to delete a service during installation/upgrade.
This is a small mitigation for the bug linked in the title. The real fix for this would be to make install plan retryable. But given the efforts towards v1, it's probably not worth it.

In the operator upgrade process, any services will be deleted and re-created. Sometimes these services can have finalizers attached that may delay the deletion of the resource and cause the upgrade to fail. 

**Architectural changes:**

**Testing remarks:**

**Reviewer Checklist**


Xia-Zhao-rh:
@perdasilva 
I can test the downstream PR, but for this upstream PR, I have no way to test it. I am sorry that.
For the jira, please link to the downstream PR :)

perdasilva:
> @perdasilva I can test the downstream PR, but for this upstream PR, I have no way to test it. I am sorry that. For the jira, please link to the downstream PR :)

@Xia-Zhao-rh I've added you guys more so you could see what will be coming down the pipeline - maybe it helps with devising the test strategies, etc.

Xia-Zhao-rh:
> 

Hi, @perdasilva Thanks a lot, it helps a lot :).

pkg/controller/install/certresources.go
```diff
@@ -260,9 +266,8 @@ func (i *StrategyDeploymentInstaller) installCertRequirementsForDeployment(deplo
 		service.SetOwnerReferences(existingService.GetOwnerReferences())
 
 		// Delete the Service to replace
-		deleteErr := i.strategyClient.GetOpClient().DeleteService(service.GetNamespace(), service.GetName(), &metav1.DeleteOptions{})
-		if deleteErr != nil && !apierrors.IsNotFound(deleteErr) {
-			return nil, nil, fmt.Errorf("could not delete existing service %s", service.GetName())
+		if err := i.deleteServiceWithTimeout(service, serviceDeletionTimeout); err != nil {
+			return nil, nil, err
```
timflannagan, line 270: I think we briefly talked about this yesterday, but did you end up finding any context for why we're creating/deletion these resources vs. creating/updating/etc.? Was it a technical limitation at the time? Could we use SSA in our controllers?

perdasilva: @timflannagan I didn't manage to get to the bottom of it =S @awgreene @kevinrizza do you guys happen to know?

awgreene: I believe its because a number of these resources don't support update commands on certain fields, IIRC it originally surfaced with a service update request?

awgreene:
Thanks for your patience @tiraboschi,

I’m not entirely convinced that this is the correct workaround. This approach will effectively force the CatalogOperator to introduce up to a 5 minute wait if a service has a finalizer and the controller hasn’t cleaned up the service yet.

This can dramatically degrade our multitenency experience. I believe that the catalog operator is configured to reconcile up to two installPlans at a time. If multiple users hit this issue at once, the catalog operator will not make any progress on other installPlans on the cluster. This can negatively affect the user experience of other tenants if on a multitenant cluster, as they:
- Won’t be able to install or upgrade any operators.
- Won’t be notified _why_ the catalog operator isn’t reconciling installPlans.
- Won’t be able to inspect the catalog operator logs for information (viewing the logs is a privileged role as it informs the user about other tenant’s activity).
- 
I can understand that this issue is frustrating, if an installPlan fails it almost certainly requires user interaction to resolve the issue. This is an inherent design flaw with OLM V0, in which installPlans we’re meant to be transactional records of what was going to be installed. In some instances, this is done purposely to force the cluster admin to resolve some critical issue. Redesigning this user experience is not trivial, and can be considered a new feature as opposed to a bug fix.

This is not an issue in the new OLM V1 project, as we are purposely pursuing a design in which the “set of installables” is determined by the state of the cluster and purposely avoid introducing a resource that effectively “halts” the cluster from recovering.

We will need to weigh the benefit of redesigning OLM V0 to accommodate this usecase versus continuing to make strides in delivering OLM V1. As we make this decision, the only option is not to place finalizers on any resources that OLM owns as this isn’t supported.

perdasilva:
closing as we've decided that this isn't a suitable workaround/fix

Issue #2874, opened 2022-10-18
High CPU usage by the olm operator


mykhailo-b:
Greetings. We faced the problem of high CPU usage by the olm operator in openshift 4.11

https://github.com/okd-project/okd/releases/download/4.11.0-0.okd-2022-08-20-022919/release.txt

We examined the source code and images of the operator 
( https://quay.io/repository/openshift/okd-content/manifest/sha256:6ad02f2e27937f4ec449718c27dbbb0870b55c910b21f4a22f202ce1cfb56d6f ) 
and found out that the operator is built from this repository https://github.com/openshift/operator-framework-olm

It seems that this repository uses an outdated version of operator-lifecycle-manager
https://github.com/openshift/operator-framework-olm/tree/master/staging/operator-lifecycle-manager
https://github.com/openshift/operator-framework-olm/tree/master/vendor/github.com/operator-framework/operator-lifecycle-manager

Although it is indicated that this is version 0.19.0
https://github.com/openshift/operator-framework-olm/blob/master/staging/operator-lifecycle-manager/OLM_VERSION
but in reality it's not.

We were especially interested in the absence of this fix
https://github.com/operator-framework/operator-lifecycle-manager/commit/b85df587ecdcc1adb0f89ad4c6e8bac0b7a75af2

Can you comment on our findings ?

kevinrizza:
Hey @mykhailo-b ,

That repo (https://github.com/openshift/operator-framework-olm) is where OCP and OKD releases live. Today, there isn't a relationship between specific versioned releases of this repository and that downstream OpenShift version -- the openshift branches are generally a snapshot in time + a set of curated commits that are pulled onto a given release branch.

The fix you referenced is quite old and actually predates the inception of that downstream openshift repository, so it's definitely included -- keep in mind that that repo is not a fork so there isn't commit matching, but you can search commit message history after initial inception (which happened around the end of the year in 2020) for specific commits if you are interested.

So, all that being said, I think it's unlikely that the commit you referenced is related to a performance problem you're having in OKD 4.11. Could you give us any more information about the specific cpu issue? Are you seeing it on the olm-operator? The catalog-operator? What's the topology of your cluster? Any specific data you have about the cpu profile would be helpful.

EugeneMospan:
Hi @kevinrizza  

Thank you for your quick reply.
Let me step in because we were working together with @mykhailo-b  on the issue.

Our context is the following:
1) We are using OKD 4.11
[...]

PR #2880, opened 2022-10-27
Fix RichReference Ordering

awgreene:
Problem: The operator CR's status includes a list of componenets owned by the operator. The list of components are ordered by GVK, but the order of objects with the same GVK may change. If an operator owns many components, there is a high likelyhood that OLM will continuously update the status of the operator CR because the order of its components have changed

Solution: Order an operators component references so OLM does not attempt to update the status of the operator CR.

### DIFF
--- a/pkg/controller/operators/operator_controller.go
+++ b/pkg/controller/operators/operator_controller.go
@@ -9,6 +9,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/labels"
@@ -152,9 +153,11 @@ func (r *OperatorReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 			return ctrl.Result{Requeue: true}, nil
 		}
 	} else {
-		if err := r.Status().Update(ctx, operator.Operator); err != nil {
-			log.Error(err, "Could not update Operator status")
-			return ctrl.Result{Requeue: true}, nil
+		if !equality.Semantic.DeepEqual(in.Status, operator.Operator.Status) {
+			if err := r.Status().Update(ctx, operator.Operator); err != nil {
+				log.Error(err, "Could not update Operator status")
+				return ctrl.Result{Requeue: true}, nil
+			}
 		}
 	}
 
--- a/pkg/controller/operators/operator_controller_test.go
+++ b/pkg/controller/operators/operator_controller_test.go
@@ -2,6 +2,7 @@ package operators
 
 import (
 	"context"
+	"fmt"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
@@ -201,6 +202,44 @@ var _ = Describe("Operator Controller", func() {
 						err := k8sClient.Get(ctx, name, operator)
 						return operator.Status.Components.Refs, err
 					}, timeout, interval).Should(ConsistOf(expectedRefs))
+				})
+			})
+
+			Context("when multiple types of a gvk are labeled", func() {
+				BeforeEach(func() {
+					newObjs := make([]runtime.Object, 9)
+
+					// Create objects in reverse order to ensure they are eventually ordered alphabetically in the status of the operator.
+					for i := 8; i >= 0; i-- {
+						newObjs[8-i] = testobj.WithLabels(map[string]string{expectedKey: ""}, testobj.WithNamespacedName(
+							&types.NamespacedName{Namespace: namespace, Name: fmt.Sprintf("sa-%d", i)}, &corev1.ServiceAccount{},
+						))
+					}
+
+					for _, obj := range newObjs {
+						Expect(k8sClient.Create(ctx, obj.(client.Object))).To(Succeed())
+					}
+
+					objs = append(objs, newObjs...)
+					expectedRefs = append(expectedRefs, toRefs(scheme, newObjs...)...)
+				})
+
+				It("should list each of the component references in alphabetical order by namespace and name", func() {
+					Eventually(func() ([]operatorsv1.RichReference, error) {
+						err := k8sClient.Get(ctx, name, operator)
+						return operator.Status.Components.Refs, err
+					}, timeout, interval).Should(ConsistOf(expectedRefs))
+
+					serviceAccountCount := 0
+					for _, ref := range operator.Status.Components.Refs {
+						if ref.Kind != rbacv1.ServiceAccountKind {
+							continue
+						}
+
+						Expect(ref.Name).Should(Equal(fmt.Sprintf("sa-%d", serviceAccountCount)))
+						serviceAccountCount++
+					}
+					Expect(serviceAccountCount).To(Equal(9))
 				})
 			})
 
--- a/pkg/controller/operators/decorators/operator.go
+++ b/pkg/controller/operators/decorators/operator.go
@@ -2,6 +2,7 @@ package decorators
 
 import (
 	"fmt"
+	"sort"
 	"strings"
 
 	"github.com/itchyny/gojq"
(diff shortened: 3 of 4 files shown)


awgreene:
Fixes #2874

vendor/github.com/operator-framework/api/pkg/operators/v1/operator_types.go
```diff
@@ -47,6 +47,29 @@ type Components struct {
 	Refs []RichReference `json:"refs,omitempty"`
 }
 
+func (c Components) Len() int {
```
awgreene, line 50: I need to cut a release of the https://github.com/operator-framework/api instead of modifying the vendor files directly.

awgreene:
I'll have to bump the k8s version to v0.25.0 to use the changes introduced in https://github.com/operator-framework/api/pull/271

go.mod
```diff
@@ -7,58 +7,59 @@ require (
...
+	k8s.io/utils v0.0.0-20220728103510-ee6ede2d64ed
+	sigs.k8s.io/controller-runtime v0.13.0
 	sigs.k8s.io/controller-tools v0.8.0
 	sigs.k8s.io/kind v0.14.0
 )
 
 replace google.golang.org/grpc => google.golang.org/grpc v1.40.0
 
 replace (
+	github.com/operator-framework/api => github.com/awgreene/api v0.13.0
```
awgreene, line 62: /hold until this changes in https://github.com/operator-framework/api/pull/271 are merged and we use an official release of the of/api

pkg/controller/operators/operator_controller.go
```diff
@@ -152,9 +153,11 @@ func (r *OperatorReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 			return ctrl.Result{Requeue: true}, nil
 		}
 	} else {
-		if err := r.Status().Update(ctx, operator.Operator); err != nil {
-			log.Error(err, "Could not update Operator status")
-			return ctrl.Result{Requeue: true}, nil
+		if !reflect.DeepEqual(in.Status, operator.Operator.Status) {
```
joelanford, line 156: Nit: it's probably better to use `equality.Semantic.DeepEqual` from `k8s.io/apimachinery/pkg/api/equality` since that correctly handles non-material differences in the comparison.

https://pkg.go.dev/k8s.io/apimachinery/pkg/api/equality#pkg-variables

awgreene: cc @asmacdo this would have been good to include in your operator's best practices doc!

asmacdo: I'll update that doc, TIL

go.mod
awgreene: /unhold
The current PR does not need to update the operator-frameowrk/api version.

vendor/github.com/operator-framework/api/pkg/operators/v1/operator_types.go
awgreene: No longer required.

pkg/controller/operators/operator_controller.go
```diff
@@ -152,9 +153,11 @@ func (r *OperatorReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 			return ctrl.Result{Requeue: true}, nil
 		}
 	} else {
-		if err := r.Status().Update(ctx, operator.Operator); err != nil {
-			log.Error(err, "Could not update Operator status")
-			return ctrl.Result{Requeue: true}, nil
+		if !equality.Semantic.DeepEqual(in.Status, operator.Operator.Status) {
+			if err := r.Status().Update(ctx, operator.Operator); err != nil {
+				log.Error(err, "Could not update Operator status")
+				return ctrl.Result{Requeue: true}, nil
```
timflannagan, line 159: nit: any reason we're requeuing here when we know we're in an error path?

```go
if err := r.Status().Update(ctx, operator.Operator); err != nil {
    return ctrl.Result{}, err
```

Do we need to log an error here if we're requeueing?

pkg/controller/operators/operator_controller_test.go
```diff
@@ -204,6 +204,44 @@ var _ = Describe("Operator Controller", func() {
...
+
+					objs = append(objs, newObjs...)
+					expectedRefs = append(expectedRefs, toRefs(scheme, newObjs...)...)
+				})
+
+				It("should list each of the component references in alphabetical order by namespace and name", func() {
+					Eventually(func() ([]operatorsv1.RichReference, error) {
+						err := k8sClient.Get(ctx, name, operator)
+						return operator.Status.Components.Refs, err
+					}, timeout, interval).Should(ConsistOf(expectedRefs))
```
timflannagan, line 230: Should we split this out into it's own individual spec?

pkg/controller/operators/operator_controller_test.go
```diff
@@ -204,6 +204,44 @@ var _ = Describe("Operator Controller", func() {
...
+					for _, ref := range operator.Status.Components.Refs {
+						if ref.Kind != rbacv1.ServiceAccountKind {
+							continue
+						}
+
+						Expect(expectedServiceAccountOrder).Should(Not(BeEmpty()))
+						Expect(ref.Name).Should(Equal(expectedServiceAccountOrder[0]))
+						expectedServiceAccountOrder = expectedServiceAccountOrder[1:]
+					}
+					Expect(expectedServiceAccountOrder).Should(BeEmpty())
```
timflannagan, line 241: Do we need an auxiliary type here? We have access to the loop iteration value, so we could capture that vs. using a blank identifier for the index value.

```go
for i, ref := range operator.Status.Components.Refs {
    if ref.Kind != rbacv1.ServiceAccountKind { continue }
    Expect(ref.Name).Should(Equal(fmt.Sprintf("sa-%d", i))
```

Thoughts?

pkg/controller/operators/operator_controller.go
dtfranz: In my limited experience calls to the api like this could result in ephemeral errors that may clear themselves on the next attempt. I guess it really depends on what kind of error you're getting back though.

timflannagan: I think my original comment wasn't super clear: I was basically asking whether we should be specifying `ctrl.Result{Requeue: true}, nil` here vs. `ctrl.Result{}, err` given we already know we're handling a non-nil error in this conditional.

This isn't a blocker though, as both of those implementations are semantically equal given they will result in a requeue being processed. The latter is more prevalent w.r.t the OLM repositories, so that implementation would remain consistent with the rest of the OLM controlled codebases.

timflannagan: And I also think an implementation that logs, and then returns an error is an anti-pattern given you can expect the call site to determine whether the error should be logged.

dtfranz: Oh! Thanks for the explanation, your comment makes more sense now. I had missed the `err` vs `nil` return there on my first look through.

awgreene: That's a good point @timflannagan , I didn't change the update call from the original code but I agree that we ought to just return the error. I can make a quick update.

pkg/controller/operators/operator_controller_test.go
awgreene: Sorry, it's own individual spec?

pkg/controller/operators/operator_controller_test.go
awgreene: That seems reasonable, I still want to make sure that all the serviceAccounts are found, so I might make a slight modification:

```
serviceAccountCount := 0
					for i, ref := range operator.Status.Components.Refs {
						if ref.Kind != rbacv1.ServiceAccountKind {
							continue
						}

						Expect(ref.Name).Should(Equal(fmt.Sprintf("sa-%d", i)))
						serviceAccountCount++
					}
					Expect(serviceAccountCount).Should(Equal(9))
				})
```

pkg/controller/operators/operator_controller.go
awgreene: I started updating this, but this pattern happens all over the operator controller, I'm gonna do this in a separate PR.

pkg/controller/operators/operator_controller_test.go
awgreene: This doesn't work if there is any other object associated with the operator, had to update the code slightly:
```
serviceAccountCount := 0
for _, ref := range operator.Status.Components.Refs {
	if ref.Kind != rbacv1.ServiceAccountKind {
		continue
	}

	Expect(ref.Name).Should(Equal(fmt.Sprintf("sa-%d", serviceAccountCount)))
	serviceAccountCount++
}
Expect(serviceAccountCount).Should(Equal(9))
```

timflannagan: Yep, that works for me.

pkg/controller/operators/operator_controller.go
```diff
@@ -134,6 +135,8 @@ func (r *OperatorReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 		}
 	}
 
+	log.V(1).Info("operator", "comp", in.Status)
```
timflannagan, line 138: Is this a debug artifact?

pkg/controller/operators/operator_controller.go
timflannagan: Yeah if this is a pattern throughout the operator controller, then this implementation is fine with me for now.

timflannagan:
LGTM once https://github.com/operator-framework/operator-lifecycle-manager/pull/2880#discussion_r1009891967 is resolved and tests are green. Just want to make sure we're not introducing a debug artifact into the reconciliation logic, and I think we could get the current status view from inspecting the actual object itself.

pkg/controller/operators/operator_controller_test.go
timflannagan: Err, s/"individual spec"/an "It" container if that makes more sense?

Basically, move this check into an "It" container:

```go
It("should eventually contain a status.Components.Refs that match the expected ref count", func() {
	Eventually(func() ([]operatorsv1.RichReference, error) {
		err := k8sClient.Get(ctx, name, operator)
		return operator.Status.Components.Refs, err
	}, timeout, interval).Should(ConsistOf(expectedRefs))
})
...
```

But I don't think this is super important given the nature of this bug fix.

pkg/controller/operators/operator_controller.go
awgreene: Yep, my bad

awgreene:
Thanks for the review @timflannagan, we should be good to go!

PR #2893, opened 2022-11-11
Add channel entries to the packagemanifests API

joelanford:
Add list of channel entries to the package manifest API, where each entry includes the CSV name, version, and upgrade edges (replaces, skips, and skipRange) 

Surface list of available bundles and their upgrade edges, which can help users understand what can be installed (e.g. answer to queston: "what can i set startingCSV to?").

See also: 
- https://github.com/operator-framework/kubectl-operator/issues/1
- https://github.com/operator-framework/kubectl-operator/issues/2
- https://issues.redhat.com/browse/OLM-1594

**Architectural changes:**
None. Only notable change is that the package server calls `ListBundles` to get this information when syncing a catalog source.

**Testing remarks:**

**Reviewer Checklist**


pkg/package-server/provider/registry_test.go
```diff
@@ -512,6 +512,22 @@ func TestRegistryProviderGet(t *testing.T) {
 								require.NoError(t, json.Unmarshal([]byte(etcdCSVJSON), &csv))
 								return operators.CreateCSVDescription(&csv, etcdCSVJSON)
 							}(),
+							Entries: []operators.ChannelEntry{
```
grokspawn, line 515: Given that the entry list for test prometheus and etcd operators is reused throughout, does it make sense to declare them as objects which we can reuse?

joelanford: Done!

pkg/package-server/apis/operators/packagemanifest_types.go
```diff
@@ -81,6 +80,17 @@ type PackageChannel struct {
 
 	// CurrentCSVSpec holds the spec of the current CSV
 	CurrentCSVDesc CSVDescription
+
+	// Entries lists all CSVs in the channel, with their upgrade edges.
+	Entries []ChannelEntry
```
perdasilva, line 85: nit: would refactoring to `Entries` to `ChannelEntries` improve readability?

joelanford: I could go either way. We're already in the `PackageChannel` struct, and none of the other fields have a `Channel` prefix, so my instinct was just `Entries`.

I _did_ originally have the struct type `ChannelEntry` called `Entry`. I changed from `Entry` to `ChannelEntry` in that case because the type is defined at the package level, and it made sense to me for usage of the struct from outside the package to look like `operators.ChannelEntry` rather than `operators.Entry`

PR #2900, opened 2022-11-24
Dump cache to logs on resolution failed

perdasilva:

Dump cache contents to log in case of resolution failure

From time to time we're running into strange resolution errors that we suspect might be related to cache issues. This adds extra logging to give support extra data points when debugging

**Architectural changes:**

**Testing remarks:**
Tested manually locally

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -997,6 +997,7 @@ func (o *Operator) syncResolvingNamespace(obj interface{}) error {
 		// not-satisfiable error
 		if _, ok := err.(solver.NotSatisfiable); ok {
 			logger.WithError(err).Debug("resolution failed")
+			o.resolver.DumpCache(namespace)
 			_, updateErr := o.updateSubscriptionStatuses(
 				o.setSubsCond(subs, v1alpha1.SubscriptionCondition{
 					Type:    v1alpha1.SubscriptionResolutionFailed,
--- a/pkg/controller/registry/resolver/resolver.go
+++ b/pkg/controller/registry/resolver/resolver.go
@@ -55,6 +55,46 @@ func (w *debugWriter) Write(b []byte) (int, error) {
 	return n, nil
 }
 
+// DumpCache writes out the contents of the cache to logs
+// This is useful for debugging resolution errors
+func (r *Resolver) DumpCache(namespaces []string) {
+	if r.log == nil {
+		return
+	}
+
+	if r.cache == nil {
+		r.log.Warningln("resolver cache is nil")
+		return
+	}
+
+	cacheEntries := r.cache.Namespaced(namespaces...).Find(cache.True())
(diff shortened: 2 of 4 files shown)


pkg/controller/registry/resolver/resolver.go
```diff
@@ -55,6 +55,46 @@ func (w *debugWriter) Write(b []byte) (int, error) {
...
+	if r.log == nil {
+		return
+	}
+
+	if r.cache == nil {
+		r.log.Warningln("resolver cache is nil")
+		return
+	}
+
+	cacheEntries := r.cache.Namespaced(namespaces...).Find(cache.True())
```
benluddy, line 70: Would it be more useful to print the same cache view that was used when resolution failed? I am concerned about wild goose chases stemming from logs containing the failure followed immediately by a cache dump that may not represent the inputs to the failure.

pkg/controller/operators/catalog/operator.go
```diff
@@ -997,6 +997,7 @@ func (o *Operator) syncResolvingNamespace(obj interface{}) error {
 		// not-satisfiable error
 		if _, ok := err.(solver.NotSatisfiable); ok {
 			logger.WithError(err).Debug("resolution failed")
+			o.resolver.DumpCache(namespace)
```
dinhxuanvu, line 1000: While I can guess the reason behind exposing more information regarding what operators are available in the namespace for diagnostic reason, the fact there are lot of operators in default catalogsources which are available for all namespaces, this print-out can potentially overwhelm the logs for catalog-operator which is already crowded. You may want to consider if this is really a `Infof` or `Debugf` depending what the targeted audience this info is for. Admittedly, if this is debug log, it will not be present in log unless it is enabled manually which can be a pain to ask from the customer.

Secondly, this particular condition check will return a nil AKA this is not a terminal condition. So this will lead to retry on resolver. Continuously printing this info over and over again on the log is certainly not a good idea.

Lastly, I don't see how this dump will help with identifying with where this issue comes from in term of "is this catalog's fault AKA missing satisfiable candidates or is this existing operator's fault AKA conflicting stuff". It is important for logging to be helpful and not just for sake of getting the haystack first and hopefully find the needle later because it is a difficult task either way.

dinhxuanvu (commented):
Just a quick thought. ;)

pkg/controller/registry/resolver/resolver.go
```diff
@@ -55,6 +55,46 @@ func (w *debugWriter) Write(b []byte) (int, error) {
 	return n, nil
 }
 
+// DumpCache writes out the contents of the cache to logs
+// This is useful for debugging resolution errors
+func (r *Resolver) DumpCache(namespaces []string) {
```
awgreene, line 60: In addition to seeing what is in the cache, I would like to see what entries are returned by the lister, as it could be feeding stale entries into the cache.

PR #2905, opened 2022-12-06
packagemanifests: do not list deprecated channel entries

joelanford:
Ignore bundles with property `olm.deprecated` when listing channel entries in the packagemanifests API.

Follow up to #2893, which added the channel entry listing to the packagemanifests API. This is necessary to avoid misleading users about the set of bundles that available to install.

**Architectural changes:**

**Testing remarks:**

**Reviewer Checklist**
- [x] Tests are comprehensible, e.g. Ginkgo DSL is being used appropriately
- [x] Code is properly formatted

### DIFF
--- a/pkg/package-server/provider/registry.go
+++ b/pkg/package-server/provider/registry.go
@@ -13,6 +13,7 @@ import (
 	"github.com/blang/semver/v4"
 	operatorsv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 	"github.com/operator-framework/operator-registry/pkg/api"
+	orregistry "github.com/operator-framework/operator-registry/pkg/registry"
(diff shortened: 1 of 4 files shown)


joelanford:
@grokspawn This bit of added code ensures the 0.6.0 bundle is deprecated: https://github.com/joelanford/operator-lifecycle-manager/blob/d69c934f4e47429ee25b837edc2541a96a428e3e/pkg/package-server/provider/registry_test.go#L68-L72

exdx (approved):

grokspawn:
> 



> Would this be considered a change of behavior and subsequently require a version bump?

I don't think so.  In this same version timeframe we exposed the channel edges in this API and then restricted the edges we would expose.

PR #2906, opened 2022-12-13
Default to legacy PSA settings

awgreene:
Problem: OLM recently introduced a few changes to default to running its workloads in a restricted mode. As a part of these changes, catalogSources built with earlier versions of OPM will not run as expected unless the catalogSource yaml is configured to run in a legacy version. Unfortunately, these legacy catalogs cannot be ran in restricted namespaces, which includes the `olm` namespace which is used to define global catalogSources.

Solution: Provide users ample time to convert to the new restricted fromat by defaulting to legacy restrictions and reclassify the `olm` namespace as a baseline privilege namespace.


deploy/chart/templates/0000_50_olm_00-namespace.yaml
```diff
@@ -6,6 +6,8 @@ metadata:
     {{- if .Values.namespace_psa }}
     pod-security.kubernetes.io/enforce: {{ .Values.namespace_psa.enforceLevel }}
     pod-security.kubernetes.io/enforce-version: {{ .Values.namespace_psa.enforceVersion }}
+    pod-security.kubernetes.io/audit: restricted
+    pod-security.kubernetes.io/warn: restricted
```
awgreene, line 10: @joelanford I _think_ we want these here as well, thoughts?

deploy/chart/templates/0000_50_olm_00-namespace.yaml
```diff
@@ -17,4 +19,6 @@ metadata:
     {{- if .Values.operator_namespace_psa }}
     pod-security.kubernetes.io/enforce: {{ .Values.operator_namespace_psa.enforceLevel }}
     pod-security.kubernetes.io/enforce-version: {{ .Values.operator_namespace_psa.enforceVersion }}
+    pod-security.kubernetes.io/audit: restricted
+    pod-security.kubernetes.io/warn: restricted
```
awgreene, line 23: @joelanford I _think_ we want these here as well, thoughts? CC @anik120

deploy/chart/templates/0000_50_olm_00-namespace.yaml
joelanford: +1

deploy/chart/templates/0000_50_olm_00-namespace.yaml
joelanford: +1

test/e2e/catalog_e2e_test.go
```diff
@@ -1451,6 +1451,7 @@ var _ = Describe("Starting CatalogSource e2e tests", func() {
 				}).Should(Succeed())
 			})
 			It("The registry pod fails to become come up because of lack of permission", func() {
+				Skip("This test will not work until catalogSources run in restricted mode by default")
```
joelanford, line 1454: Non-blocking for follow-up. It seems like we could still exercise this case by explicitly setting the securityContextConfig on the catalog source rather than depending on defaulting?

awgreene: That makes sense to me, the test will now focus on ensuring that restricted catalogSources can't run with non-root images.

deploy/chart/templates/0000_50_olm_00-namespace.yaml
anik120: +1 we do want these here, but a nit is to have these also be populated from the chart value like the enforce values, to make things easier during downstream. eg if we want different values in downstream, we'll simply change the chart values.

It's a hygiene nit since we can always change the hardcoded value too, but that seems like confusing paper trail potentially.

deploy/chart/templates/0000_50_olm_00-namespace.yaml
anik120: Same comment as above.

deploy/chart/values.yaml
```diff
@@ -2,7 +2,7 @@ rbacApiVersion: rbac.authorization.k8s.io
 namespace: operator-lifecycle-manager
 # see https://kubernetes.io/docs/concepts/security/pod-security-admission/ for more details
 namespace_psa: 
-  enforceLevel: restricted
+  enforceLevel: baseline
```
anik120, line 5: eg we'll just include a new fields after this 

```
- auditLevel: restricted
```

pkg/controller/registry/reconciler/reconciler.go
```diff
@@ -195,11 +195,7 @@ func Pod(source *operatorsv1alpha1.CatalogSource, name string, image string, saN
 		},
 	}
 
-	if source.Spec.GrpcPodConfig != nil {
-		if source.Spec.GrpcPodConfig.SecurityContextConfig == operatorsv1alpha1.Restricted {
-			addSecurityContext(pod, runAsUser)
-		}
-	} else {
+	if source.Spec.GrpcPodConfig != nil && source.Spec.GrpcPodConfig.SecurityContextConfig == operatorsv1alpha1.Restricted {
```
anik120, line 198: 👍🏽

anik120 (commented):
Just a couple of nits otherwise lgtm

PR #2908, opened 2022-12-15
OCPBUGS-4955: Set ImagePullPolicy of bundle unpacker to "IfNotPresent" for image digests

dtfranz:

Captures the `imagePullPolicy` selection logic into a func, then utilizes it both in its original location as well as in the bundle unpacker. This allows users to set `imagePullPolicy` to `IfNotPresent` by supplying an image digest rather than a tag.

OCPBUGS-4955

### DIFF
--- a/pkg/controller/bundle/bundle_unpacker_test.go
+++ b/pkg/controller/bundle/bundle_unpacker_test.go
@@ -33,7 +33,7 @@ const (
 	opmImage    = "opm-image"
 	utilImage   = "util-image"
 	bundlePath  = "bundle-path"
-	digestImage = "bundle@sha256:54d626e08c1c802b305dad30b7e54a82f102390cc92c7d4db112048935236e9c"
+	digestPath  = "bundle-path@sha256:54d626e08c1c802b305dad30b7e54a82f102390cc92c7d4db112048935236e9c"
 	runAsUser   = 1001
 )
 
@@ -50,8 +50,8 @@ func TestConfigMapUnpacker(t *testing.T) {
 	defaultUnpackTimeoutSeconds := int64(defaultUnpackDuration.Seconds())
 
 	// Custom timeout to override the default cmdline flag ActiveDeadlineSeconds value
-	customAnnotationDuration := 2 * time.Minute
-	customAnnotationTimeoutSeconds := int64(customAnnotationDuration.Seconds())
+	//customAnnotationDuration := 2 * time.Minute
+	//customAnnotationTimeoutSeconds := int64(customAnnotationDuration.Seconds())
 
 	type fields struct {
 		objs []runtime.Object
@@ -77,7 +77,7 @@ func TestConfigMapUnpacker(t *testing.T) {
 		args        args
 		expected    expected
 	}{
-		{
+		/*{
 			description: "NoCatalogSource/NoConfigMap/NoJob/NotCreated/Pending",
 			fields:      fields{},
 			args: args{
@@ -397,20 +397,20 @@ func TestConfigMapUnpacker(t *testing.T) {
 					},
 				},
 			},
-		},
+		},*/
 		{
 			description: "CatalogSourcePresent/ConfigMapPresent/JobPresent/DigestImage/Unpacked",
 			fields: fields{
 				objs: []runtime.Object{
 					&batchv1.Job{
 						ObjectMeta: metav1.ObjectMeta{
-							Name:      pathHash,
+							Name:      hash(digestPath),
 							Namespace: "ns-a",
 							OwnerReferences: []metav1.OwnerReference{
 								{
 									APIVersion:         "v1",
 									Kind:               "ConfigMap",
-									Name:               pathHash,
+									Name:               hash(digestPath),
 									Controller:         &blockOwnerDeletion,
 									BlockOwnerDeletion: &blockOwnerDeletion,
 								},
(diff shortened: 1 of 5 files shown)


jianzhangbjz:
Hi @dtfranz , I cannot test this PR since it's for upstream, as follows:
image

I will test it once it  was submitted to downstream: https://github.com/openshift/operator-framework-olm

jianzhangbjz:
The CI test fail at:
```go
?   	github.com/operator-framework/operator-lifecycle-manager/pkg/api/wrappers/wrappersfakes	[no test files]
--- FAIL: TestConfigMapUnpacker (0.61s)
    --- FAIL: TestConfigMapUnpacker/CatalogSourcePresent/ConfigMapPresent/JobPresent/DigestImage/Unpacked (0.20s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x28ddf46]

goroutine 115 [running]:
testing.tRunner.func1.2({0x2b025e0, 0x4[32](https://github.com/operator-framework/operator-lifecycle-manager/actions/runs/3708381393/jobs/6285883596#step:5:33)1ba0})
	/opt/hostedtoolcache/go/1.19.4/x64/src/testing/testing.go:1396 +0x[37](https://github.com/operator-framework/operator-lifecycle-manager/actions/runs/3708381393/jobs/6285883596#step:5:38)2
testing.tRunner.func1()
```

dtfranz:
> The CI test fail at:
> 
> ```go
> ?   	github.com/operator-framework/operator-lifecycle-manager/pkg/api/wrappers/wrappersfakes	[no test files]
> --- FAIL: TestConfigMapUnpacker (0.61s)
>     --- FAIL: TestConfigMapUnpacker/CatalogSourcePresent/ConfigMapPresent/JobPresent/DigestImage/Unpacked (0.20s)
> panic: runtime error: invalid memory address or nil pointer dereference [recovered]
> 	panic: runtime error: invalid memory address or nil pointer dereference
> [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x28ddf46]
> 
> goroutine 115 [running]:
> testing.tRunner.func1.2({0x2b025e0, 0x4[32](https://github.com/operator-framework/operator-lifecycle-manager/actions/runs/3708381393/jobs/6285883596#step:5:33)1ba0})
> 	/opt/hostedtoolcache/go/1.19.4/x64/src/testing/testing.go:1396 +0x[37](https://github.com/operator-framework/operator-lifecycle-manager/actions/runs/3708381393/jobs/6285883596#step:5:38)2
> testing.tRunner.func1()
> ```

Thanks Jian, yes, this test passed before I pushed but that was a fluke. I've updated the test so it passes now.

pkg/controller/bundle/bundle_unpacker_test.go
```diff
@@ -398,18 +399,18 @@ func TestConfigMapUnpacker(t *testing.T) {
...
 		},
 		{
-			description: "CatalogSourcePresent/ConfigMapPresent/JobPresent/Unpacked",
+			description: "CatalogSourcePresent/ConfigMapPresent/JobPresent/DigestImage/Unpacked",
 			fields: fields{
 				objs: []runtime.Object{
 					&batchv1.Job{
 						ObjectMeta: metav1.ObjectMeta{
-							Name:      pathHash,
+							Name:      hash(digestPath),
```
awgreene, line 407: Rather than calling `hash(digestPath)` multiple times, could we define `digestHash` once upfront like we did for `pathHast`

pkg/controller/bundle/bundle_unpacker_test.go
```diff
@@ -488,8 +489,8 @@ func TestConfigMapUnpacker(t *testing.T) {
 										},
 										{
 											Name:            "pull",
-											Image:           bundlePath,
-											ImagePullPolicy: "Always",
+											Image:           digestPath,
+											ImagePullPolicy: "IfNotPresent",
```
awgreene, line 493: I think this fails to test that we have the pull policy set to `always` if a bundlePath is provided, is this captured elsewhere?

awgreene (commented):
Nice work on this @dtfranz, had one question and one suggestion, but this looks good otherwise.

pkg/controller/bundle/bundle_unpacker_test.go
dtfranz: All of the other test cases in this file use a 'normal' (non-digest) bundlePath and expect the resulting pull policy to be `Always`. Because of that I figured we'd only need a single case in here that covers the `IfNotPresent` scenario.

pkg/controller/bundle/bundle_unpacker_test.go
dtfranz: Yes good call out, updated

pkg/controller/bundle/bundle_unpacker_test.go
awgreene: That's probably sufficient, in more complex workflows we'd want to ensure that each "type" of image is tested in all places, but I think this one is pretty safe.

pkg/controller/bundle/bundle_unpacker_test.go
```diff
@@ -77,7 +77,7 @@ func TestConfigMapUnpacker(t *testing.T) {
 		args        args
 		expected    expected
 	}{
-		{
+		/*{
```
perdasilva: is this test no longer needed? maybe we should either remove, or if there's something we need to track, create an issue and tag it here?

pkg/controller/bundle/bundle_unpacker_test.go
```diff
@@ -77,7 +77,7 @@ func TestConfigMapUnpacker(t *testing.T) {
 		args        args
 		expected    expected
 	}{
-		/*{
+		{
```
perdasilva, line 81: nevermind XD

PR #2913, opened 2023-01-09
OCPBUGS-5523: Catalog, fatal error: concurrent map read and map write

dtfranz:
Protected subscriptionSyncCounters access to prevent concurrent map writes

The metrics pkg uses a shared map var which is not protected from concurrent read/write access. This PR adds a mutex to lock and unlock before and after each map access instance. 

OCPBUGS-5523

### DIFF
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -1,6 +1,7 @@
 package metrics
 
 import (
+	"sync"
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
@@ -203,6 +204,8 @@ var (
 	// Subscription objects. The key of a record is the Subscription name, while the value
 	//  is struct containing label values used in the counter
 	subscriptionSyncCounters = make(map[string]subscriptionSyncLabelValues)
+
+	subscriptionSyncCountersLock sync.Mutex
 )
 
 type subscriptionSyncLabelValues struct {
@@ -280,6 +283,8 @@ func EmitSubMetric(sub *operatorsv1alpha1.Subscription) {
 	if sub.Spec == nil {
 		return
 	}
+	subscriptionSyncCountersLock.Lock()
+	defer subscriptionSyncCountersLock.Unlock()
 	SubscriptionSyncCount.WithLabelValues(sub.GetName(), sub.Status.InstalledCSV, sub.Spec.Channel, sub.Spec.Package, string(sub.Spec.InstallPlanApproval)).Inc()
 	if _, present := subscriptionSyncCounters[sub.GetName()]; !present {
 		subscriptionSyncCounters[sub.GetName()] = subscriptionSyncLabelValues{
@@ -302,6 +307,8 @@ func UpdateSubsSyncCounterStorage(sub *operatorsv1alpha1.Subscription) {
 	if sub.Spec == nil {
 		return
 	}
+	subscriptionSyncCountersLock.Lock()
+	defer subscriptionSyncCountersLock.Unlock()
 	counterValues := subscriptionSyncCounters[sub.GetName()]
 	approvalStrategy := string(sub.Spec.InstallPlanApproval)
 
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -209,7 +209,7 @@ var (
 // protected by mutex.
 type subscriptionSyncCounter struct {
 	counters     map[string]subscriptionSyncLabelValues
-	countersLock sync.Mutex
+	countersLock sync.RWMutex
 }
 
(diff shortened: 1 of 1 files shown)


pkg/metrics/metrics.go
```diff
@@ -199,12 +200,37 @@ var (
...
+
+func (s *subscriptionSyncCounter) setValues(key string, val subscriptionSyncLabelValues) {
+	s.countersLock.Lock()
+	defer s.countersLock.Unlock()
+	s.counters[key] = val
+}
+
+func (s *subscriptionSyncCounter) readValues(key string) (subscriptionSyncLabelValues, bool) {
+	s.countersLock.Lock()
+	defer s.countersLock.Unlock()
```
joelanford, line 229:
```suggestion
	s.countersLock.RLock()
	defer s.countersLock.RUnlock()
```

pkg/metrics/metrics.go
```diff
@@ -199,12 +200,37 @@ var (
...
+	subscriptionSyncCounters = newSubscriptionSyncCounter()
 )
 
+// subscriptionSyncCounter keeps a record of the Prometheus counters emitted by
+// Subscription objects. The key of a record is the Subscription name, while the value
+// is struct containing label values used in the counter. Read and Write access are
+// protected by mutex.
+type subscriptionSyncCounter struct {
+	counters     map[string]subscriptionSyncLabelValues
+	countersLock sync.Mutex
```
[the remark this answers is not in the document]
dtfranz: Updated.

pkg/metrics/metrics.go
dtfranz: Updated.

pkg/metrics/metrics.go
```diff
@@ -199,12 +200,37 @@ var (
...
+	s.countersLock.Lock()
+	defer s.countersLock.Unlock()
+	s.counters[key] = val
+}
+
+func (s *subscriptionSyncCounter) readValues(key string) (subscriptionSyncLabelValues, bool) {
+	s.countersLock.RLock()
+	defer s.countersLock.RUnlock()
+	val, ok := s.counters[key]
+	return val, ok
```
awgreene, line 231:
```suggestion
	return s.counters[key]
```

awgreene (approved):
Seems reasonable to me, could we squash the commits? It'll make it easier for downstreaming.

pkg/metrics/metrics.go
```diff
@@ -199,12 +200,37 @@ var (
...
+	subscriptionSyncCounters = newSubscriptionSyncCounter()
 )
 
+// subscriptionSyncCounter keeps a record of the Prometheus counters emitted by
+// Subscription objects. The key of a record is the Subscription name, while the value
+// is struct containing label values used in the counter. Read and Write access are
+// protected by mutex.
+type subscriptionSyncCounter struct {
+	counters     map[string]subscriptionSyncLabelValues
+	countersLock sync.RWMutex
```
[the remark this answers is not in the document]
awgreene: /hold
Should this be a sync.Mutex? I believe that hte sync.RWMutex requires that only a single routine is classified as the writer, and I believe that controllers can be configured to run multiple routines in parallel. I can't recall if we configured OLM to reconcile subscriptions in parallel offhand, but I'll check tomorrow.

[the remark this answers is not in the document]
awgreene: This is incorrect, only a single writer can HOLD the lock at a given time.
/unhold

pkg/metrics/metrics.go
awgreene: Wow, I would have thought that this behaved the same way as it does for [functions](https://go.dev/play/p/J6HLJ0ZZQH_6), but for some reason [the bool isn't returned](https://go.dev/play/p/JSvaVQCXiKX) doesn't work when trying to use the result of a map lookup to satisfy

awgreene: I suspect this is because you don't have to receive the value as shown [here](https://go.dev/play/p/gXjkM1_zVPs).

awgreene: Courtesy of @benluddy:

[From the Go Docs](https://go.dev/ref/spec#Index_expressions):
> An index expression on a map a of type map[K]V used in an [assignment statement](https://go.dev/ref/spec#Assignment_statements) or initialization of the special form
> v, ok = a[x]
> v, ok := a[x]
> var v, ok = a[x]
> yields an additional untyped boolean value.

tmshort (approved):
/lgtm

awgreene:
@dtfranz this needs a rebase.

tmshort (approved):
/lgtm

PR #2918, opened 2023-02-01
Introduce Thread Safety test for UpdateSubsSyncCounterStorage

awgreene:
Introduces a test that ensure that the UpdateSubsSyncCounterStorage function is thread safe and avoids concurrent map writes.


dtfranz:
Test passes for me with the thread safety changes in, and fails without them
/lgtm

PR #2925, opened 2023-02-22
Sort channels in lexicographical order in packagemanifest

anik120:
See https://github.com/operator-framework/operator-registry/issues/1069 for more details 

**Architectural changes:**
None

**Reviewer Checklist**


joelanford:
Nice! Should we add a test for this to avoid regressions?

pkg/package-server/provider/registry_test.go
```diff
@@ -476,6 +477,99 @@ func TestToPackageManifest(t *testing.T) {
...
+		Name: "etcd",
+		Channels: []*api.Channel{
+			{
+				Name:    "alpha",
+				CsvName: "etcdoperator.v0.9.2",
+			},
+			{
+				Name:    "beta",
+				CsvName: "etcdoperator.v0.10.1",
+			},
```
joelanford, line 491: Does this test pass without the sort change? Seems like we may want to reverse the order of these here and check that they always end up as [alpha, beta].

joelanford: Reason being: pkg.GetChannels() returns in whatever order we define here I think:

https://github.com/operator-framework/operator-registry/blob/0a7ff74f9fc6c92be12a48a0f8e67ecfb6a8e60e/pkg/api/registry.pb.go#L175-L180

anik120: That's fair. I've reversed the order, the input is now [beta,alpha] with the expected output as [alpha, beta]. And with this setup, when the sort change is reversed, this test fails.

joelanford (approved):
/lgtm

PR #2928, opened 2023-02-23
Registry Pod Controller Flag

dtfranz:
Sets the controller flag on the registry pod owner references to true in an attempt to remove the need for forceful shutdown when draining nodes.


ankitathomas (approved):
/lgtm

PR #2942, opened 2023-03-22
Prevent OLM from creating `InstallPlan`s when bundle unpack fails

m1kola:
Prerequisites:
* https://github.com/operator-framework/api/pull/282
* https://github.com/operator-framework/operator-lifecycle-manager/pull/2954
* https://github.com/operator-framework/operator-lifecycle-manager/pull/2952

---

This PR changes how `InstallPlan`s are being created.

Today OLM creates InstallPlans even if bundle images are not available for some reason. For example, catalog source contains references to an unreachable registry.

Consider this scenario:
* User installs an operator and enables auto upgrades
* Catalog source receives an update which references unreachable bundle images (e.g. pre-release from a private registry)
* OLM creates an `InstallPlan` for an upgrade which features unreachable bundle images
* At this point upgrade is stuck since the cluster can not pull bundle images
* Catalog source receives a corrective update which contains only reachable bundle images

Workaround: to resolve this issue users have to manually remove a failed `InstallPlan`. After that OLM is expected to create a new `InstallPlan` based on the latest state of catalog sources.

The goal: We would like to find a solution which prevents `InstallPlan` from being created with unreachable images in the first place. This eliminates the need to manually delete `InstallPlan` in scenarios like the one described above.

**Architectural changes:**

Before the change:

* On a `Namespace` reconciliation (`syncResolvingNamespace`) OLM creates an `InstallPlan` if required (e.g. new sub is created or a new update in a catalog is available for an existing sub)
* On `InstallPlan` reconciliation (`syncInstallPlans`) OLM creates a `Job` to unpack a relevant bundle.
  * On unsuccessful job unpacking OLM reports failures as `InstallPlan` conditions

After this change:

* On a `Namespace` reconciliation (`syncResolvingNamespace`):
  * OLM first creates a `Job` to unpack a relevant bundle and waits for the job to sucessfully complete
    * On unsuccessful job unpacking OLM reports failures as `Subscription` conditions (set on all subs in the namespace)
  * Only after successfull completion OLM creates an `InstallPlan`

**Testing remarks:**

To reproduce:

* Create a test cluster and install OLM (`make run-local` from this repo can be helpful)
* Create a catalog source:
    ```yaml
    kind: CatalogSource
    metadata:
      name: alex-operators
      namespace: olm
    spec:
      image: quay.io/agreene/index:unreachable-bundle-upgrade
      displayName: Alex Operators
      priority: -100
      publisher: Alex
      sourceType: grpc
      updateStrategy:
        registryPoll:
          interval: 10m0s
    ```
* And create a subscription:
    ```yaml
    kind: Subscription
    metadata:
      name: quay
      namespace: operators
    spec:
      channel: unavailable-image-upgrade
      name: quay
      source: alex-operators
      sourceNamespace: olm
      startingCSV: quay-operator.v3.8.3
    ```

In master - you will see two `InstallPlan`s: one for `quay-operator.v3.8.3` (successfull install) and one for `quay-operator.v3.8.4` which will eventually (10min by default) report failure to unpack in bundle lookup conditions.

In this branch you will see only one `InstallPlan` for `quay-operator.v3.8.3` (successfull install). Eventually `Subscription` will get a condition reporting failure to unpack (`SubscriptionBundleUnpackFailed`). During unpacking it should report `UnpackingInProgress`.

**Reviewer Checklist**


pkg/controller/operators/catalog/operator.go
```diff
@@ -1028,6 +1064,55 @@ func (o *Operator) syncResolvingNamespace(obj interface{}) error {
...
+		// which means bundle lookup has failed and subscriptions need to be updated
+		// with a condition indicating the failure.
+		isFailed, cond := hasBundleLookupFailureCondition(bundleLookups)
+		if isFailed {
+			err := fmt.Errorf("bundle unpacking failed. Reason: %v, and Message: %v", cond.Reason, cond.Message)
+			logger.Infof("%v", err)
+
+			_, updateErr := o.updateSubscriptionStatuses(
+				o.setSubsCond(subs, v1alpha1.SubscriptionCondition{
+					// TODO: Decide if we need a separate type and/or reason for unpack failures
```
m1kola, line 1089: Most likely we will be exposing two conditions on subs. Something like: `SubscriptionBundleUnpacking` and `SubscriptionBundleUnpackFailed`.

pkg/controller/bundle/bundle_unpacker.go
```diff
@@ -39,7 +39,7 @@ const (
 	BundleLookupFailed operatorsv1alpha1.BundleLookupConditionType = "BundleLookupFailed"
```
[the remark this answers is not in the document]
tmshort: The actual code is not that large; it's the tests...

pkg/controller/operators/catalog/operator.go
```diff
@@ -1028,6 +1034,80 @@ func (o *Operator) syncResolvingNamespace(obj interface{}) error {
...
+						Status:  corev1.ConditionTrue,
+					}))
+				if updateErr != nil {
+					logger.WithError(updateErr).Debug("failed to update subs conditions")
+					return updateErr
+				}
+				return nil
+			}
+			// Retry sync if non-fatal error
+			return fmt.Errorf("bundle unpacking failed with an error: %v", err)
```
tmshort, line 1061: Should probably use `%w` rather than `%v`

tmshort: (In general with respect to printing errors)

test/e2e/catalog_e2e_test.go
```diff
@@ -1028,30 +1028,29 @@ var _ = Describe("Starting CatalogSource e2e tests", func() {
...
-
 		source, err := crc.OperatorsV1alpha1().CatalogSources(source.GetNamespace()).Create(context.Background(), source, metav1.CreateOptions{})
 		Expect(err).ShouldNot(HaveOccurred())
 		defer func() {
 			err := crc.OperatorsV1alpha1().CatalogSources(source.GetNamespace()).Delete(context.Background(), source.GetName(), metav1.DeleteOptions{})
 			Expect(err).ShouldNot(HaveOccurred())
 		}()
 
-		// Wait for the CatalogSource to be ready
+		By("waiting for the CatalogSource to be ready")
```
tmshort, line 1038: Nit: Changing these comments to `By()` calls is not strictly necessary. Nice, though.

m1kola: I had some issues (timeouts, IIRC) when running this test locally. `By()` prints steps and it helps to narrow down the piece of the test which had these issues. They act like docs and logs at the same time.

tmshort: Oh, I know. Just saying that these changes aren't strictly related to the primary code change. That being said, no need to change it.

pkg/controller/operators/catalog/operator.go
tmshort: Agreed.

tmshort (commented):
A lot of the testing changes were just tests being moved around to account for the different behavior. I'm still looking , but it looks reasonable.

pkg/controller/operators/catalog/operator.go
m1kola: Had to update Had to update again: I messed up expected error in the test. Only updated [this line](https://github.com/operator-framework/operator-lifecycle-manager/pull/2942/files#diff-ae673094a18efa85f1d053631f5dfe6770527c3c4ac4f7cef194a44f75c511daR639).

pkg/controller/operators/catalog/operator.go
```diff
@@ -1062,6 +1142,15 @@ func (o *Operator) syncResolvingNamespace(obj interface{}) error {
 		logger.Debugf("no subscriptions were updated")
 	}
 
+	// Make sure that we no longer indicate unpacking progress
```
dtfranz, line 1145: Apologies if I'm not understanding the order of operations correctly here, but does this condition not become true _before_ calling `ensureInstallPlan`? Or in other words shouldn't these status updates come on line 1110?  

Or else, are we saying that a bundle unpack isn't successful until an `InstallPlan` is successfully created with the unpacked bundles?

dtfranz:
Looks good! Had a couple questions. 

You mentioned above:
> Only after successfull completion OLM creates an InstallPlan

In the example scenario then, where the CatalogSource is updated again with correct images, will OLM self-heal?

In the case where the bundle unpack is successful and then the following `InstallPlan` fails for an unrelated reason, what happens to the bundle? Can it be re-used on subsequent `InstallPlan` attempts, or will it need to be unpacked again on each retry?

pkg/controller/operators/catalog/operator.go
m1kola: > Apologies if I'm not understanding the order of operations correctly here, but does this condition not become true _before_ calling `ensureInstallPlan`? Or in other words shouldn't these status updates come on line 1110?

Your understanding is correct. Order is as follows:
* We first create an unpack job within `o.unpackBundles(...)`
* Based on the result from the above we either set `SubscriptionBundleUnpackFailed` (in case of errors) or `SubscriptionBundleUnpacking` (in case when we still waiting on the unpack job) to `True`. I believe `SubscriptionBundleUnpacking` will be the most common on the first run when syncing with a new bundle.
* Once unpacking is done - we create an `InstallPlan` in `o.ensureInstallPlan(...)`
  * We just create an `InstallPlan` here, but syncing of the `InstallPlan` is a separate process - (see `o.syncInstallPlans`)
* All being well - we make sure to remove `SubscriptionBundleUnpackFailed` on subs (if any exist) and we set `SubscriptionBundleUnpacking` to `False`.
* Update subs


> Or else, are we saying that a bundle unpack isn't successful until an `InstallPlan` is successfully created with the unpacked bundles?

Technically unpack can be successfull without successfull `InstallPlan` creation. But in terms of current implementation in this PR - you are right.

I considered removing `SubscriptionBundleUnpackFailed` and setting `SubscriptionBundleUnpacking` to `False` before `o.ensureInstallPlan(...)`, but it makes syncing a bit more complex becaues we would have to do two updates to the subs. We would have to implement something like this:

* Unpacking (first two steps from the list above)
* Removing `SubscriptionBundleUnpackFailed` and setting `SubscriptionBundleUnpacking` to `False`
* `o.updateSubscriptionStatuses(subs)` <--- First update
* `o.ensureInstallPlan(...)` and `o.setIPReference(...)`
* `o.updateSubscriptionStatuses(subs)` <--- Second update

Instead we just do:

* Unpacking
* `o.ensureInstallPlan(...)` and `o.setIPReference(...)`
* Removing `SubscriptionBundleUnpackFailed` and setting `SubscriptionBundleUnpacking` to `False`
* `o.updateSubscriptionStatuses(subs)` <--- The only update to subs at the end

So it is a bit of optimisation, but I'm happy to reconsider. Let me know if you see an issue with this.

m1kola:
> In the example scenario then, where the CatalogSource is updated again with correct images, will OLM self-heal?

@dtfranz the answer is yes, but... I was going to point you to a e2e test demonstrating this and it looks like it is missing. I was sure that I covered it, but it looks like I didn't. I'll look into adding an e2e to demonstrate this. Good catch! 👍

I'll _probably_ submit it as a follow up PR. Given this PR already has decent coverage and is already too big it might be good idea to do that separately. In any way - I'll let you know once I add extra coverage here or elsewhere.

> In the case where the bundle unpack is successful and then the following `InstallPlan` fails for an unrelated reason, what happens to the bundle? Can it be re-used on subsequent `InstallPlan` attempts, or will it need to be unpacked again on each retry?

I did not touch the bit concerning unpack jobs (not directly at least), but I believe you are correct: unpack jobs are being reused. See in `pkg/controller/bundle`:
* Before creating a new job we first check whether job already exists by name [here](https://github.com/operator-framework/operator-lifecycle-manager/blob/cf13b7731a2a272f2eb38a5bef3b02c5fdca50cf/pkg/controller/bundle/bundle_unpacker.go#L628-L635)
* Name of the job comes from the name of config map via [this](https://github.com/operator-framework/operator-lifecycle-manager/blob/cf13b7731a2a272f2eb38a5bef3b02c5fdca50cf/pkg/controller/bundle/bundle_unpacker.go#L467) ref
* And the config name gets its name from [this](https://github.com/operator-framework/operator-lifecycle-manager/blob/cf13b7731a2a272f2eb38a5bef3b02c5fdca50cf/pkg/controller/bundle/bundle_unpacker.go#L87) hash which is based on bundle lookup path
* So as long as bundle lookup path (image reference) stays the same - jobs should be reusable

test/e2e/fail_forward_e2e_test.go
```diff
@@ -102,18 +100,11 @@ var _ = Describe("Fail Forward Upgrades", func() {
...
-			By("verifying the subscription is referencing a new installplan")
-			subscription, err = fetchSubscription(crclient, subscription.GetNamespace(), subscription.GetName(), subscriptionHasInstallPlanDifferentChecker(originalInstallPlanRef.Name))
-			Expect(err).Should(BeNil())
-
-			By("waiting for the bad InstallPlan to report a failed installation state")
-			ref := subscription.Status.InstallPlanRef
-			_, err = fetchInstallPlan(GinkgoT(), crclient, ref.Name, ref.Namespace, buildInstallPlanPhaseCheckFunc(operatorsv1alpha1.InstallPlanPhaseFailed))
-			Expect(err).To(BeNil())
+			By("verifying that the subscription is still on v0.1.0")
+			Consistently(subscriptionCurrentCSVGetter(crclient, subscription.GetNamespace(), subscription.GetName())).Should(Equal("example-operator.v0.1.0"))
```
m1kola, line 107: I think this is probably not the right check for this test (and some other tests) as `subscriptionCurrentCSVGetter` checks `CurrentCSV`.

pkg/controller/operators/catalog/operator.go
dtfranz: Thanks for answering, I appreciate the level of detail!  
I'm fine with what you have here to keep API calls to a minimum. We _probably_ don't need such a fine-grained level of tracing for the process anyway - I mostly wanted to make sure there was intention behind it. We can keep it the way you have it :)

dtfranz:
> I'll _probably_ submit it as a follow up PR. Given this PR already has decent coverage and is already too big it might be good idea to do that separately. In any way - I'll let you know once I add extra coverage here or elsewhere.

I agree with you there, I'm fine with this being submitted in a follow-up. Thanks for answering those questions as well!
/lgtm

m1kola:
Putting this on hold as I had several unexpected `ConstraintsNotSatisfiable` while looking into additional E2E coverage.

/hold

pkg/controller/operators/catalog/operator.go
```diff
@@ -1062,6 +1142,15 @@ func (o *Operator) syncResolvingNamespace(obj interface{}) error {
...
+	subs = o.setSubsCond(subs, v1alpha1.SubscriptionCondition{
+		Type:   v1alpha1.SubscriptionBundleUnpacking,
+		Status: corev1.ConditionFalse,
+	})
+
+	// Remove BundleUnpackFailed condition from subscriptions
+	subs = o.removeSubsCond(subs, v1alpha1.SubscriptionBundleUnpackFailed)
```
m1kola, line 1152: Ok, I found a bug with this line. `o.removeSubsCond` only returns a list of subs which were updated.

After hitting this line in successfull installation from the first time - we will have `nil` in `subs` aftere this line. And it might result in `ConstraintsNotSatisfiable` or other conditions to not be updated.

m1kola (commented):
Will move the PR into a draft state as I found a bug.

pkg/controller/operators/catalog/operator.go
m1kola: We now do `o.removeSubsCond(subs, v1alpha1.SubscriptionBundleUnpackFailed)` and `o.removeSubsCond(subs, v1alpha1.SubscriptionResolutionFailed)` (not overriding `subs` slice).

I believe `v1alpha1.SubscriptionResolutionFailed` had the same issue.

m1kola:
Fixed the issue mentioned above (details are [here](https://github.com/operator-framework/operator-lifecycle-manager/pull/2942#discussion_r1186045060)). Extra E2E coverage added in a [follow up PR](https://github.com/operator-framework/operator-lifecycle-manager/pull/2961) just to not make things even more complicated in this one.

/hold cancel

pkg/controller/operators/catalog/operator.go
```diff
@@ -1028,6 +1034,80 @@ func (o *Operator) syncResolvingNamespace(obj interface{}) error {
...
+	// Attempt to unpack bundles before installing
+	// Note: This should probably use the attenuated client to prevent users from resolving resources they otherwise don't have access to.
+	if len(bundleLookups) > 0 {
+		logger.Debug("unpacking bundles")
+
+		var unpacked bool
+		unpacked, steps, bundleLookups, err = o.unpackBundles(namespace, steps, bundleLookups, unpackTimeout)
+		if err != nil {
+			// If the error was fatal capture and fail
+			if olmerrors.IsFatal(err) {
```
pgodowski, line 1046: Sorry for stupid question, but is this type of failure the 'fatal' one?

```yaml
bundle unpacking failed. Reason: DeadlineExceeded, and Message: Job was
        active longer than specified deadline
```

pkg/controller/operators/catalog/operator.go
```diff
@@ -1028,6 +1034,80 @@ func (o *Operator) syncResolvingNamespace(obj interface{}) error {
...
+					Type:    v1alpha1.SubscriptionBundleUnpackFailed,
+					Reason:  "BundleUnpackFailed",
+					Message: err.Error(),
+					Status:  corev1.ConditionTrue,
+				}))
+			if updateErr != nil {
+				logger.WithError(updateErr).Debug("failed to update subs conditions")
+				return updateErr
+			}
+			// Since this is likely requires intervention we do not want to
```
pgodowski, line 1083: We hit the error
`bundle unpacking failed. Reason: DeadlineExceeded, and Message: Job was
        active longer than specified deadline`


frequently after `CatalogSource` update, but for sure the bundle images are correct. Could we handle this and requeue rather than looking for the manual intervention?

pgodowski (commented):
Asking questions about DeadlineExceeded unpack Job retries

pkg/controller/operators/catalog/operator.go
m1kola: No, failure from your example will fall under `if isFailed` branch.

Here is an example of `FatalError`s:

https://github.com/operator-framework/operator-lifecycle-manager/blob/858c7cdb2f4cfe3d351225972aede54e3b114ce4/pkg/controller/registry/resolver/steps.go#L122-L129

m1kola:
Had to rebase since something else got merged into master

---

@pgodowski thank you for the example. Very helpful. I think I mislead you on the community call yesterday because I misunderstood what you were asking about. I will try to explain the behaviour for this case better now.

> catalog source was updated with the GA content which surely contains valid operator bundles and then for some reason, the unpack Job fails.
> Today when unpack Job fails, we have to manually remove it.

Nothing changes in this scenario apart from the fact that there will be no `InstallPlan` created now. Failure will be reported on a subscription via the `BundleUnpackFailed` condition. You will still have to take a manual action in cases like this as `DeadlineExceeded` might indicate a legit issue (for example, cluster gets 403 when pulling a bundle image).

Your options are:

* Increase unpack timeout using `--bundle-unpack-timeout` (10m by default). This might be helpful if you are aware of a reason why unpacking takes longer than default timeout on your clusters. Might happen due to a variety of reasons such as insufficient resources for unpacking Pod or slow image pulling due to unreliable networking, etc. Increased timeout will allow Job to stay alive for longer period of time. This timeout translates into [Job's `.spec.activeDeadlineSeconds`][activeDeadlineSeconds]. Keep in mind that OLM sets `.spec.backoffLimit` to 3 still. 
* Keep removing the Job manually. OLM will recreated the Job for the bundle on the next reconciliation.

However once this PR merged it will be easier to introduce logic where we can instruct OLM to recreated unpack failed jobs automatically. E.g. "recreated failed unpack jobs that are older than 1 day". It will be easier because we won't have to worry about transitioning `InstallPlan` from the failed terminal state into non-terminal state since after this PR we won't be creating `InstallPlans` at all. I think it will come down to modifying [`ensureJob` here][ensureJob] + adding a param. But I personally not a big fan of this idea as it sounds like hiding root cause behind the retries.

It might be a good idea to create an issue for this and include details of root cause(s) for unpack jobs failures in your clusters.

To summarise - We are not trying to improve this case here. This PR aims at solving an issue when catalog source gets updated with a corrected references to bundle images.



[activeDeadlineSeconds]: https://kubernetes.io/docs/concepts/workloads/controllers/job/#job-termination-and-cleanup
[ensureJob]: https://github.com/operator-framework/operator-lifecycle-manager/blob/858c7cdb2f4cfe3d351225972aede54e3b114ce4/pkg/controller/bundle/bundle_unpacker.go#L626-L645

pkg/controller/operators/catalog/operator.go
m1kola: Replied [here](https://github.com/operator-framework/operator-lifecycle-manager/pull/2942#issuecomment-1551184652).

pgodowski:
> To summarise - We are not trying to improve this case here. This PR aims at solving an issue when catalog source gets updated with a corrected references to bundle images.

Understood - thanks.

test/e2e/fail_forward_e2e_test.go
```diff
@@ -189,14 +207,13 @@ var _ = Describe("Fail Forward Upgrades", func() {
...
 
-			By("waiting for the subscription to maintain the example-operator.v0.2.0 status.updatedCSV")
-			Consistently(func() string {
-				subscription, err := crclient.OperatorsV1alpha1().Subscriptions(subscription.GetNamespace()).Get(context.Background(), subscription.GetName(), metav1.GetOptions{})
-				if err != nil || subscription == nil {
-					return ""
-				}
-				return subscription.Status.CurrentCSV
-			}).Should(Equal("example-operator.v0.2.0"))
+			By("waiting for the subscription to maintain the example-operator.v0.2.0 status.currentCSV")
```
ankitathomas, line 210: nit: By references 0.2.0 instead of 0.1.0

perdasilva:
@m1kola I've rebased from this side

tmshort (approved):

test/e2e/fail_forward_e2e_test.go
```diff
@@ -102,40 +103,30 @@ var _ = Describe("Fail Forward Upgrades", func() {
...
 			Expect(err).To(BeNil())
 
-			By("waiting for the subscription to have the example-operator.v0.2.1 status.updatedCSV")
-			subscription, err = fetchSubscription(crclient, subscription.GetNamespace(), subscription.GetName(), subscriptionHasCurrentCSV("example-operator.v0.2.1"))
-			Expect(err).Should(BeNil())
-
-			By("waiting for the bad v0.2.1 InstallPlan to report a failed installation state")
-			ref := subscription.Status.InstallPlanRef
-			_, err = fetchInstallPlan(GinkgoT(), crclient, ref.Name, ref.Namespace, buildInstallPlanPhaseCheckFunc(operatorsv1alpha1.InstallPlanPhaseFailed))
-			Expect(err).To(BeNil())
```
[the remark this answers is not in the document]
m1kola: Actually, I'm having second thoughts on this. How I rearanged tests to be "when InstallPlan is not being created due to failed upgrade" just provides extra coverage for what we cover elsewhere (in `test/e2e/subscription_e2e_test.go` and and extra tests defined in #2961).

> This should test our ability to recover from a failed installPlan, so it may make sense to remove fail-forward tests around missing bundle images.

I think this is the right approach.

> shouldn't we verify the failed installPlan state as a prerequisite for testing fail-forward?

I'll explore getting back to "an InstallPlan is reporting a failed state" and emulating failed `InstallPlan` execution somewhere in `BeforeEach`.

test/e2e/fail_forward_e2e_test.go
```diff
@@ -102,40 +103,30 @@ var _ = Describe("Fail Forward Upgrades", func() {
...
+			By("waiting for the subscription to maintain the example-operator.v0.1.0 status.currentCSV")
+			Consistently(subscriptionCurrentCSVGetter(crclient, subscription.GetNamespace(), subscription.GetName())).Should(Equal("example-operator.v0.1.0"))
 
+			By("verifying the subscription is referencing the same InstallPlan")
+			subscription, err = fetchSubscription(crclient, subscription.GetNamespace(), subscription.GetName(), subscriptionHasInstallPlanChecker)
+			Expect(err).Should(BeNil())
+			Expect(subscription.Status.InstallPlanRef.Name).To(Equal(originalInstallPlanRef.Name))
```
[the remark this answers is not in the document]
m1kola: Resolving this thread. Let's continue in https://github.com/operator-framework/operator-lifecycle-manager/pull/2942/files#r1196646809

m1kola:
I updated to address feedback.

Here is the short summary of the diff since the last push which ([diff can be found here](https://github.com/operator-framework/operator-lifecycle-manager/compare/67b8fc69b1d1945a4243f4d6110ddfcbee5b6e62..9ed3d8363609854f272963e4781e67548bcaf8f4)). It is arguably easier to comparae modified files to the master still.

* Updated `test/e2e/fail_forward_e2e_test.go` as per descusion [here](https://github.com/operator-framework/operator-lifecycle-manager/pull/2942#discussion_r1196646809): reverted (with necessary changes) to having failed `InstallPlan`.
* Updated `test/e2e/subscription_e2e_test.go` to add extra coverage to show how OLM auto-recovers as per [this](https://github.com/operator-framework/operator-lifecycle-manager/pull/2942#issuecomment-1527315191). Initially I wanted it to be a separate PR (#2961) but to address fail forward I needed extra test data file which I already had in that separate PR. So I squashed them together here.
* Rearanged test data in `test/e2e/testdata/fail-forward/` for above two to work.

No changes to the controller itself. Only tests.

I hate large PRs like this, but changes changes in `test/e2e/fail_forward_e2e_test.go` were necessary to ensure that we do not merge something which fails to catch a regression in fail forward feature. And it dragged other things in.

@tmshort @dtfranz @pgodowski @ankitathomas @perdasilva please take another look.

perdasilva (approved):
lgtm - excellent work!! users will be pleased ^^

PR #2949, opened 2023-04-06
Option to skip the local role lift for the global operator group

alebedev87:
**Note**
It's just a draft which can be a starting point for the further discussion.

Ability to skip the lifting of the local roles to the cluster level for the global operator group if a special annotation is set on CSV.

The operators are forced to migrate to `AllNamespaces` installmode when they start to use the conversion webhook. This may be undesirable for some of the operators, especially for those which were designed to not be the cluster level operators. The allnamespaces installmode has a very inconvenient side effect of lifting the local permissions specified in CSV to the cluster level. This may give excessive permissions to the operator.

**Architectural changes:**
The default behavior is the same as before, the new annotation can be applied optionally to those operators which wish to opt out of the local roles lifting.

**Testing remarks:**

**Reviewer Checklist**


perdasilva:
closing PR as stale. Please re-open if it's still important.

PR #2952, opened 2023-04-13
Moves bundle unpack timeout into OperatorGroup

m1kola:

Moves `operatorframework.io/bundle-unpack-timeout` annotation from `InstallPlan` to `OperatorGroup`

`operatorframework.io/bundle-unpack-timeout` is an internal annotation used for E2E testing.

We need to move this out of `InstallPlan` in preparation to changes in the unpacking process (see #2942): OLM will soon be creating unpack jobs before creating `InstallPlan` so we need to find a new place where we can set this annotation.

**Testing remarks:**

Updates existing E2E tests

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/bundle/bundle_unpacker.go
+++ b/pkg/controller/bundle/bundle_unpacker.go
@@ -26,6 +26,7 @@ import (
 
 	"github.com/operator-framework/api/pkg/operators/reference"
 	operatorsv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
+	v1listers "github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/listers/operators/v1"
 	listersoperatorsv1alpha1 "github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/listers/operators/v1alpha1"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/install"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/registry/resolver/projection"
@@ -770,4 +771,31 @@ func getCondition(job *batchv1.Job, conditionType batchv1.JobConditionType) (con
 		}
 	}
 	return
+}
+
+// OperatorGroupBundleUnpackTimeout returns bundle timeout from annotation if specified.
+// If the timeout annotation is not set, return timeout < 0 which is subsequently ignored.
+// This is to overrides the --bundle-unpack-timeout flag value on per-OperatorGroup basis.
+func OperatorGroupBundleUnpackTimeout(ogLister v1listers.OperatorGroupNamespaceLister) (time.Duration, error) {
+	ignoreTimeout := -1 * time.Minute
+
+	ogs, err := ogLister.List(k8slabels.Everything())
+	if err != nil {
+		return ignoreTimeout, err
+	}
+	if len(ogs) != 1 {
+		return ignoreTimeout, fmt.Errorf("found %d operatorGroups, expected 1", len(ogs))
+	}
+
+	timeoutStr, ok := ogs[0].GetAnnotations()[BundleUnpackTimeoutAnnotationKey]
+	if !ok {
+		return ignoreTimeout, nil
+	}
+
+	d, err := time.ParseDuration(timeoutStr)
+	if err != nil {
+		return ignoreTimeout, fmt.Errorf("failed to parse unpack timeout annotation(%s: %s): %w", BundleUnpackTimeoutAnnotationKey, timeoutStr, err)
+	}
+
+	return d, nil
 }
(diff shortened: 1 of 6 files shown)


perdasilva (approved):
lgtm

ankitathomas (approved):
/lgtm

pkg/controller/bundle/bundle_unpacker.go
```diff
@@ -768,3 +769,31 @@ func getCondition(job *batchv1.Job, conditionType batchv1.JobConditionType) (con
...
+
+// OperatorGroupBundleUnpackTimeout returns bundle timeout from annotation if specified.
+// This is to overrides the --bundle-unpack-timeout flag value on per-OperatorGroup basis.
+// If the timeout cannot be parsed it's set to < 0 and subsequently ignored.
+func OperatorGroupBundleUnpackTimeout(logger *logrus.Logger, ogLister v1listers.OperatorGroupNamespaceLister) (time.Duration, error) {
+	ignoreTimeout := -1 * time.Minute
+
+	ogs, err := ogLister.List(k8slabels.Everything())
+	if err != nil || len(ogs) == 0 {
+		return ignoreTimeout, nil
```
tmshort, line 781: Shouldn't this return an `err`?

tmshort: Also, the check for `len(ogs) == 0` matches the check below. Are we treating 0 different than not 1?

pkg/controller/bundle/bundle_unpacker.go
```diff
@@ -768,3 +769,31 @@ func getCondition(job *batchv1.Job, conditionType batchv1.JobConditionType) (con
...
+
+	timeoutStr, ok := ogs[0].GetAnnotations()[BundleUnpackTimeoutAnnotationKey]
+	if !ok {
+		return ignoreTimeout, nil
+	}
+
+	d, err := time.ParseDuration(timeoutStr)
+	if err != nil {
+		logger.Errorf("failed to parse unpack timeout annotation(%s: %s): %v", BundleUnpackTimeoutAnnotationKey, timeoutStr, err)
+		return ignoreTimeout, nil
```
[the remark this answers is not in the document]
m1kola: This works in the same way it was before the refactoring [here](https://github.com/operator-framework/operator-lifecycle-manager/pull/2952/files#diff-a1760d9b7ac1e93734eea675d8d8938c96a50e995434b163c6f77c91bace9990L1422): we log an error and ignore the value.

The only case when we want to return an error in this func - is where there are two `OperatorGroup`s in which case we can not proceed.

perdasilva:
Rebased from this side

pkg/controller/bundle/bundle_unpacker.go
perdasilva: I wonder if we ought to return the error if the `List` call fails and remove the `|| len(ogs) == 0`. This can let the caller decide whether they want to retry, or are happy with using the default value. Maybe we could also add a ticket for the old code so we get this rechecked. I'm worried silent errors around this code could lead to hard to determine bugs. wdyt?

m1kola: I made this function to return an error from `List` and removed `|| len(ogs) == 0`. From what I know about OLM so far - `OperatorGroup` is required and without it resolution will fail eventually. In case of these errors - sync will be requeued (no extra code on caller required).

I'll try to the same for old code in `IsFailForwardEnabled`, but in a separate PR. Will see how it goes.

m1kola: Created https://github.com/operator-framework/operator-lifecycle-manager/pull/2957 for `IsFailForwardEnabled`.

test/e2e/installplan_e2e_test.go
```diff
@@ -3485,11 +3487,9 @@ var _ = Describe("Install Plan", func() {
...
 		It("should show an error on the bundlelookup condition for a non-existent bundle image", func() {
 			// We wait for some time over the bundle unpack timeout (i.e ActiveDeadlineSeconds) so that the Job can eventually fail
-			// Since the default --bundle-unpack-timeout=10m, we override with a shorter timeout via the
-			// unpack timeout annotation on the InstallPlan
-			annotations := make(map[string]string)
-			annotations[bundle.BundleUnpackTimeoutAnnotationKey] = "1m"
-			ip.SetAnnotations(annotations)
+			// Since the default --bundle-unpack-timeout=10m, we override with a shorter timeout
+			By("patching the OperatorGroup to reduce the bundle unpacking timeout")
+			addBundleUnpackTimeoutOGAnnotation(context.Background(), ctx.Ctx().Client(), types.NamespacedName{Name: ogName, Namespace: ns.GetName()}, "1m")
```
tmshort, line 3492: What happens when `ogName` is `nil`?

tmshort: Nevermind; initialized before each test.

PR #2970, opened 2023-05-26
Fixes a typo

m1kola:
Just a typo (or a grammar mistake 🤷‍♂️) in a comment.


stevekuznetsov (commented):
If plural, subscriptions *need* to be updated. Your change would be appropriate if it were a singular subscription that needed to be updated.

PR #2979, opened 2023-06-14
Retry initialization error conditions

tmshort:

When the api server is flakey (e.g. during a cluster install), it is possible for some of the OLM initialization to fail. When this happens, OLM gets into a bad state (e.g. a monitoring go routine terminates) and can't recover without a restart.

There were at least two places I found where a retry mechanism is needed to handle intialization errors. This was as far as I peeled the onion. It's not an exponential backoff retry, but a 1 minute retry interval should be sufficient (no other backoffs are exponential).

Downstream bug report.

**Architectural changes:**

None.

**Testing remarks:**

**Reviewer Checklist**

### DIFF
--- a/pkg/lib/operatorstatus/monitor.go
+++ b/pkg/lib/operatorstatus/monitor.go
@@ -129,9 +129,18 @@ func (m *monitor) Run(stopCh <-chan struct{}) {
 	m.logger.Infof("initializing clusteroperator resource(s) for %s", m.names)
 
 	for _, name := range m.names {
-		if err := m.init(name); err != nil {
-			m.logger.Errorf("initialization error - %v", err)
-			break
+		for {
+			if err := m.init(name); err != nil {
+				m.logger.Errorf("initialization error - %v", err)
+			} else {
+				m.logger.Infof("initialized cluster resource - %s", name)
+				break
+			}
+			select {
+			case <-time.After(defaultProbeInterval):
+			case <-stopCh:
+				return
+			}
 		}
 	}
 
--- a/pkg/lib/queueinformer/queueinformer_operator.go
+++ b/pkg/lib/queueinformer/queueinformer_operator.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"sync"
+	"time"
 
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/kubestate"
 	"github.com/pkg/errors"
@@ -11,6 +12,10 @@ import (
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/tools/cache"
+)
+
+const (
+	defaultServerVersionInterval = 1 * time.Minute
 )
 
 // ExtensibleOperator describes a Reconciler that can be extended with additional informers and queue informers
@@ -194,6 +199,16 @@ func (o *operator) start(ctx context.Context) error {
 	go func() {
 		defer close(errs)
 		v, err := o.serverVersion.ServerVersion()
+		if err == nil {
+			o.logger.Infof("connection established. cluster-version: %v", v)
+			return
+		}
+		select {
+		case <-time.After(defaultServerVersionInterval):
+		case <-ctx.Done():
+			return
+		}
+		v, err = o.serverVersion.ServerVersion()
 		if err != nil {
 			select {
 			case errs <- errors.Wrap(err, "communicating with server failed"):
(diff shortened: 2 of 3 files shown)


pkg/lib/queueinformer/queueinformer_operator_test.go
```diff
@@ -78,7 +78,7 @@ func TestOperatorRunChannelClosure(t *testing.T) {
 
 			o.Run(ctx)
 
-			timeout := time.After(time.Second)
+			timeout := time.After(2 * time.Minute)
```
ankitathomas, line 81: is this meant to represent `2*defaultServerVersionInterval` ?

pkg/lib/operatorstatus/monitor.go
```diff
@@ -129,9 +129,18 @@ func (m *monitor) Run(stopCh <-chan struct{}) {
...
+				m.logger.Errorf("initialization error - %v", err)
+			} else {
+				m.logger.Infof("initialized cluster resource - %s", name)
+				break
+			}
+			select {
+			case <-time.After(defaultProbeInterval):
+			case <-stopCh:
+				return
+			}
```
ankitathomas, line 143: Do we want to try initialization indefinitely?

pkg/lib/queueinformer/queueinformer_operator_test.go
tmshort: Yeah, it probably should be.

pkg/lib/operatorstatus/monitor.go
tmshort: If this fails, then OLM is unable to monitor and update the `packageserver` cluster operator. After chatting with some people, it was decided to keep on retrying, rather than terminate the OLM process (which would restart the pod), which could look worse in metrics.

The other retry was added, but is restricted because it impacts unit tests.

pkg/lib/queueinformer/queueinformer_operator_test.go
tmshort: Fixed

pkg/lib/queueinformer/queueinformer_operator.go
```diff
@@ -194,6 +199,16 @@ func (o *operator) start(ctx context.Context) error {
...
+		if err == nil {
+			o.logger.Infof("connection established. cluster-version: %v", v)
+			return
+		}
+		select {
+		case <-time.After(defaultServerVersionInterval):
+		case <-ctx.Done():
+			return
+		}
+		v, err = o.serverVersion.ServerVersion()
```
awgreene, line 211: So if I'm reading this correctly, we attempt to get the server version twice. If the first attempt fails, we retry after the `defaultServerVersionInterval`. If we ever succeed we return the serverVersion. Could you share your reasoning why we make two attempts here but infinite attempts above?

pkg/lib/operatorstatus/monitor.go
```diff
@@ -129,9 +129,18 @@ func (m *monitor) Run(stopCh <-chan struct{}) {
 	m.logger.Infof("initializing clusteroperator resource(s) for %s", m.names)
 
 	for _, name := range m.names {
-		if err := m.init(name); err != nil {
-			m.logger.Errorf("initialization error - %v", err)
-			break
+		for {
```
awgreene, line 132: As I read the code, I thought we might want to consider running this code in parallel for each "name", but I suspect that there isn't much of a performance gain given that:

1. I think the length of `m.names` is 3 or so entries.
2. Even if we ran this in parallel, I suspect that all three would fail if the controller was unable to initialize, so we'd just have 3 concurrent waits.

pkg/lib/queueinformer/queueinformer_operator.go
tmshort: Unit tests... the unit tests actually wait for this initialization to complete (success or failure), which is guaranteed to fail.

pkg/lib/operatorstatus/monitor.go
tmshort: I'm not sure it's worth it.

pkg/lib/queueinformer/queueinformer_operator.go
tmshort: (I would've liked to have had an infinite wait, but then the unit-tests timeout)

pkg/lib/operatorstatus/monitor.go
tmshort: Presumably, if the first one fails, then it makes sense to push off trying subsequent names until the first one succeeds. This puts less pressure on the API server.

ankitathomas (approved):
/lgtm

ankitathomas (approved):
/lgtm

PR #3003, opened 2023-08-02
OCPBUGS-17157: improve watching semantics around CSVs

stevekuznetsov:
operators/olm: use a partial object metadata watch for copied CSVs

All we ever ned to know about copied CSVs is their metadata. No need to
prune objects in memory, it's better to never allocate the memory to
deserilize them in the first place.

---

operators/catalog: don't watch copied CSVs

As far as I can tell, we never do anything with them.

---

### DIFF
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -204,7 +204,9 @@ func NewOperator(ctx context.Context, kubeconfigPath string, clock utilclock.Clo
 	// Fields are pruned from local copies of the objects managed
 	// by this informer in order to reduce cached size.
 	prunedCSVInformer := cache.NewSharedIndexInformer(
-		pruning.NewListerWatcher(op.client, metav1.NamespaceAll, func(*metav1.ListOptions) {}, pruning.PrunerFunc(func(csv *v1alpha1.ClusterServiceVersion) {
+		pruning.NewListerWatcher(op.client, metav1.NamespaceAll, func(options *metav1.ListOptions) {
+			options.LabelSelector = fmt.Sprintf("!%s", v1alpha1.CopiedLabelKey)
+		}, pruning.PrunerFunc(func(csv *v1alpha1.ClusterServiceVersion) {
 			*csv = v1alpha1.ClusterServiceVersion{
 				TypeMeta: csv.TypeMeta,
 				ObjectMeta: metav1.ObjectMeta{
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -204,29 +204,31 @@ func NewOperator(ctx context.Context, kubeconfigPath string, clock utilclock.Clo
 	// Fields are pruned from local copies of the objects managed
 	// by this informer in order to reduce cached size.
 	prunedCSVInformer := cache.NewSharedIndexInformer(
-		pruning.NewListerWatcher(op.client, metav1.NamespaceAll, func(options *metav1.ListOptions) {
-			options.LabelSelector = fmt.Sprintf("!%s", v1alpha1.CopiedLabelKey)
-		}, pruning.PrunerFunc(func(csv *v1alpha1.ClusterServiceVersion) {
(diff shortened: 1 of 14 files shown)


go.mod
```diff
@@ -40,11 +40,11 @@ require (
 	google.golang.org/grpc v1.53.0
 	gopkg.in/yaml.v2 v2.4.0
 	helm.sh/helm/v3 v3.12.2
-	k8s.io/api v0.27.2
+	k8s.io/api v0.27.4
```
stevekuznetsov, line 43: This is not technically necessary, can revert if required.

pkg/controller/operators/olm/operator.go
```diff
@@ -1195,17 +1172,16 @@ func (a *Operator) handleClusterServiceVersionDeletion(obj interface{}) {
...
 		"id":          queueinformer.NewLoopID(),
 		"csv":         csv.GetName(),
 		"namespace":   csv.GetNamespace(),
-		"phase":       csv.Status.Phase,
 		"labels":      csv.GetLabels(),
 		"annotations": csv.GetAnnotations(),
 	})
 
-	if !csv.IsCopied() {
+	if !IsCopied(csv) {
```
stevekuznetsov, line 1184: TODO: import from `api` repo once the dependent PR merges

go.mod
tmshort: Since everything else is at 0.27.2, I'd suggest reverting.

tmshort: If possible, I'd suggest reverting everything back to 0.27.2, but I realize that may not always be possible.
Having different versions of the k8s modules across the api/operator-registry/operator-lifecycle-manager repos will cause us headaches when downstreaming.

pkg/controller/operators/catalog/operator.go
```diff
@@ -204,7 +204,9 @@ func NewOperator(ctx context.Context, kubeconfigPath string, clock utilclock.Clo
 	// Fields are pruned from local copies of the objects managed
 	// by this informer in order to reduce cached size.
 	prunedCSVInformer := cache.NewSharedIndexInformer(
-		pruning.NewListerWatcher(op.client, metav1.NamespaceAll, func(*metav1.ListOptions) {}, pruning.PrunerFunc(func(csv *v1alpha1.ClusterServiceVersion) {
+		pruning.NewListerWatcher(op.client, metav1.NamespaceAll, func(options *metav1.ListOptions) {
+			options.LabelSelector = fmt.Sprintf("!%s", v1alpha1.CopiedLabelKey)
+		}, pruning.PrunerFunc(func(csv *v1alpha1.ClusterServiceVersion) {
```
tmshort, line 209: I'd ask for a newline after the `},` to make it clear that these are separate arguments to the `NewSharedIndexInformer()` function, rather than a continuation of the `func` definition of the first argument.
(Unless `go fmt` insists it be this way...)

pkg/controller/operators/olm/operator.go
```diff
@@ -1683,18 +1661,23 @@ func (a *Operator) createCSVCopyingDisabledEvent(csv *v1alpha1.ClusterServiceVer
...
 		return
 	}
 	return
 }
 
+func IsCopied(o metav1.Object) bool {
+	_, ok := o.GetLabels()[v1alpha1.CopiedLabelKey]
+	return ok
+}
+
```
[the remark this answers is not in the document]
stevekuznetsov: Did this to make the diff easier - importing it from `operatorsv1alpha1` now that the dependent PR is in

perdasilva (approved):
7f5

Nice one! Thank you ^^
We have a downstream only plug-in that creates a [copiedcsv informer](https://github.com/openshift/operator-framework-olm/blob/master/staging/operator-lifecycle-manager/pkg/controller/operators/olm/plugins/downstream_csv_namespace_labeler_plugin.go#L283) - do you think we'd benefit with this change there as well?

stevekuznetsov:
@perdasilva interesting! Yeah, I'll get that when I downstream this, thanks for the call-out.

PR #3005, opened 2023-08-04
OCPBUGS-17157:  operators/olm: record and expose informers

stevekuznetsov:
Plugins should re-use informers that the host started, since there's no need to double the number of watches and caches we're doing to reconcile the same set of objects in one process.


awgreene:
Thanks for performance improvement @stevekuznetsov.

PR #3006, opened 2023-08-04
reconcile clusteroperator status more frequently

sanchezl:
Needed for openshift/cluster-kube-apiserver-operator#1423.

If status is modified, the operator must set status to proper values. This ensures that accidents cannot permanently reset status and gives a clear indication that the operator is "live". This came up as important when operators were NOT live, during cert rotations and we had no indication of problems.

Ideally this should happen on every sync (e.g. move this logic to its own crontroller loop, as it was done for the `operator-lifecycle-manager` ) but setting the backoff interval on sucess small enough to avoid triggering the future stale condition controller from flagging these conditions as stale / `Unknown` should be sufficient as a quick fix.


stevekuznetsov:
@joelanford can we get a review on this one?

perdasilva (approved):
lgtm

PR #3007, opened 2023-08-08
OCPBUGS-17157: queueinformer: don't double register informers

stevekuznetsov:
The abstraction here doesn't let you re-use an informer without also causing it to be re-registered and re-started, which is invalid.


awgreene (approved):
/approve

dinhxuanvu (commented):
/lgtm

PR #3014, opened 2023-08-17
OCPBUGS-17157: pkg/controller: use a metadata watch for CRDs

stevekuznetsov:
Using a full LIST+WATCH is an optimization, with trade-offs. Holding the state of the world for CRDs in memory when we rarely, if ever, actually need to access them is a bad use of that trade-off, especially when the sum total size of CRDs on even the most basic cluster is O(20MiB).


Makefile
```diff
@@ -64,7 +64,7 @@ ifeq (, $(wildcard $(KUBEBUILDER_ASSETS)/kube-apiserver))
 endif
 
 cover.out:
-	go test $(MOD_FLAGS) -tags "json1" -v -race -coverprofile=cover.out -covermode=atomic \
```
stevekuznetsov, line 67: Not necessary but the DX of watching 10k lines of output scroll by on success makes it impossible to find test failures and should be an anti-pattern.

pkg/controller/operators/catalog/operator.go
```diff
@@ -443,13 +450,23 @@ func NewOperator(ctx context.Context, kubeconfigPath string, clock utilclock.Clo
 		return nil, err
 	}
 
-	// Register CustomResourceDefinition QueueInformer
-	crdInformer := extinf.NewSharedInformerFactory(op.opClient.ApiextensionsInterface(), resyncPeriod()).Apiextensions().V1().CustomResourceDefinitions()
-	op.lister.APIExtensionsV1().RegisterCustomResourceDefinitionLister(crdInformer.Lister())
+	// Register CustomResourceDefinition QueueInformer. Object metadata requests are used
```
stevekuznetsov, line 453: I can't find any users of the CRD lister/informer in the catalog operator, but this code can sometimes be :spaghetti: so I'm not confident enough to just delete it.

pkg/controller/operators/olm/operator.go
```diff
@@ -1183,7 +1193,7 @@ func (a *Operator) handleClusterServiceVersionDeletion(obj interface{}) {
 		}
 
 		for i, crdName := range desc.ConversionCRDs {
-			crd, err := a.lister.APIExtensionsV1().CustomResourceDefinitionLister().Get(crdName)
+			crd, err := a.opClient.ApiextensionsInterface().ApiextensionsV1().CustomResourceDefinitions().Get(context.TODO(), crdName, metav1.GetOptions{})
```
stevekuznetsov, line 1196: This is the crux of the trade-off. In the rare case that a CSV is deleted, we incur server cost and latency to do a live GET for the data we need. Better than holding onto 20MiB of every CRD ever when we would only ever be concerned with a tiny subset, and with the spec of that tiny subset in an even smaller set of cases.

Makefile
joelanford: +1000

pkg/controller/operators/olm/operator.go
tmshort: But it's an improvement...

tmshort (approved):
/lgtm
(Assuming CI passes)

stevekuznetsov:
Let me dig into e2e, this passes locally for me ... :|

stevekuznetsov:
@tmshort should be good to go now

PR #3015, opened 2023-08-18
OCPBUGS-17791: controller/registry: don't set memory limits for target

stevekuznetsov:
It's not allowed for core components to have memory limits, to ensure that that source of crash-looping pods does not occur to the core payload.


tmshort:
Need to update the bug to include `4.14.0` to fix the jira label.

stevekuznetsov:
Had it on the wrong bug, oops

PR #3016, opened 2023-08-22
retry unpacking jobs on failure

ankitathomas:
Recreate failed bundle unpack jobs to allow for automatic retries on unpacking failure.

Bundle unpack jobs may fail due to network or configuration issues in the cluster that may be transient or resolved with user intervention. Since unpack jobs have deterministic names referencing the bundle they correspond to, recovery from unpack failures requires manual intervention for deleting the associated unpack jobs.

This PR automates recreation of failed unpack jobs indefinitely with a minimum guaranteed interval between jobs if specified by a new operatorGroup annotation.

### DIFF
--- a/pkg/controller/bundle/bundle_unpacker.go
+++ b/pkg/controller/bundle/bundle_unpacker.go
@@ -659,6 +659,14 @@ func (c *ConfigMapUnpacker) ensureJob(cmRef *corev1.ObjectReference, bundlePath 
 
 		return
 	}
+	// Cleanup old unpacking job and retry
+	if _, isFailed := getCondition(job, batchv1.JobFailed); isFailed {
+		err = c.client.BatchV1().Jobs(job.GetNamespace()).Delete(context.TODO(), job.GetName(), metav1.DeleteOptions{})
+		if err == nil {
+			job, err = c.client.BatchV1().Jobs(fresh.GetNamespace()).Create(context.TODO(), fresh, metav1.CreateOptions{})
+		}
+		return
+	}
 
 	if equality.Semantic.DeepDerivative(fresh.GetOwnerReferences(), job.GetOwnerReferences()) && equality.Semantic.DeepDerivative(fresh.Spec, job.Spec) {
 		return
--- a/pkg/controller/bundle/bundle_unpacker.go
+++ b/pkg/controller/bundle/bundle_unpacker.go
@@ -41,6 +41,13 @@ const (
 	// e.g 1m30s
 	BundleUnpackTimeoutAnnotationKey = "operatorframework.io/bundle-unpack-timeout"
 	BundleUnpackPodLabel             = "job-name"
+
+	// BundleUnpackRetryMinimumIntervalAnnotationKey sets a minimum interval to wait before
+	// attempting to recreate a failed unpack job for a bundle.
+	BundleUnpackRetryMinimumIntervalAnnotationKey = "operatorframework.io/bundle-unpack-min-retry-interval"
+
+	// bundleUnpackRefLabel is used to filter for all unpack jobs for a specific bundle.
+	bundleUnpackRefLabel = "operatorframework.io/bundle-unpack-ref"
 )
 
 type BundleUnpackResult struct {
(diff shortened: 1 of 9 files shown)


pkg/controller/bundle/bundle_unpacker.go
```diff
@@ -651,6 +651,14 @@ func (c *ConfigMapUnpacker) ensureJob(cmRef *corev1.ObjectReference, bundlePath
...
 		return
 	}
+	// Cleanup old unpacking job and retry
+	if _, isFailed := getCondition(job, batchv1.JobFailed); isFailed {
+		err = c.client.BatchV1().Jobs(job.GetNamespace()).Delete(context.TODO(), job.GetName(), metav1.DeleteOptions{})
+		if err == nil {
+			job, err = c.client.BatchV1().Jobs(fresh.GetNamespace()).Create(context.TODO(), fresh, metav1.CreateOptions{})
+		}
+		return
+	}
```
tmshort, line 661: This appears to retry without limit?

pkg/controller/bundle/bundle_unpacker.go
```diff
@@ -651,6 +651,14 @@ func (c *ConfigMapUnpacker) ensureJob(cmRef *corev1.ObjectReference, bundlePath
 
 		return
 	}
+	// Cleanup old unpacking job and retry
```
ankitathomas, line 654: If we don't care about persisting the failed job at all, this can be simplified to deleting the job immediately after failure and waiting for the next resolver run.

perdasilva: I think we should persist the failed job - we need a debug trail of some sort

pkg/controller/bundle/bundle_unpacker.go
perdasilva: What's the retry cadence? is it exp backoff?
Maybe it's ok to retry forever as long as we're not hammering the apiserver

ankitathomas: This runs whenever olm syncs resolves a namespace - we use the [default client-go workqueue](https://github.com/kubernetes/client-go/blob/v0.28.1/util/workqueue/default_rate_limiters.go#L39) so we have exp backoff up to ~15 min.

pkg/controller/bundle/bundle_unpacker.go
```diff
@@ -651,6 +651,14 @@ func (c *ConfigMapUnpacker) ensureJob(cmRef *corev1.ObjectReference, bundlePath
...
+	if _, isFailed := getCondition(job, batchv1.JobFailed); isFailed {
+		err = c.client.BatchV1().Jobs(job.GetNamespace()).Delete(context.TODO(), job.GetName(), metav1.DeleteOptions{})
```
varshaprasad96, line 656: why delete it manually and not set TTL? (https://kubernetes.io/docs/concepts/workloads/controllers/job/#ttl-mechanism-for-finished-jobs). 

I haven't looked into the entire code, but I'm assuming the controller should re-create a new Job incase it has not been unpacked.

ankitathomas: iirc, the current implementation [requires completed jobs to persist](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/controller/bundle/bundle_unpacker.go#L533) to indicate an unpacked bundle.

pkg/controller/bundle/bundle_unpacker.go
```diff
@@ -41,6 +41,13 @@ const (
 	// e.g 1m30s
 	BundleUnpackTimeoutAnnotationKey = "operatorframework.io/bundle-unpack-timeout"
 	BundleUnpackPodLabel             = "job-name"
+
+	// BundleUnpackRetryMinimumIntervalAnnotationKey sets a minimum interval to wait before
+	// attempting to recreate a failed unpack job for a bundle.
+	BundleUnpackRetryMinimumIntervalAnnotationKey = "operatorframework.io/bundle-unpack-min-retry-interval"
```
dtfranz, line 47: Will you have a follow-up PR to document how to use this field?

ankitathomas: I'll follow up with https://github.com/operator-framework/olm-docs/issues/313 once the PR is merged

tmshort (approved):
/lgtm

PR #3017, opened 2023-08-22
 OCPBUGS-17157: label non-OLM resources

stevekuznetsov:
HACK: add a round-tripper to ensure we label non-OLM resources

We can either flag-gate this to dev only, or delete it before merge.

---

*: label non-OLM resources

Today, our controllers use un-filtered LIST+WATCH calls to monitor the
state of the cluster. For OLM-specific resource types, that's fine,
since we need to know (for instance) about every CSV. For non-OLM
resource groups, though, that is needlessly wasteful in memory
consumption and makes our controller's footprint scale with the size of
the cluster itself, irrespective of the usage of OLM. Adding a label to
every resource we create is the first step in being able to filter down
all of those requests to only those objects with our label.

---

### DIFF
--- a/cmd/olm/main.go
+++ b/cmd/olm/main.go
@@ -11,6 +11,7 @@ import (
 
 	configclientset "github.com/openshift/client-go/config/clientset/versioned"
 	configv1client "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/operators/validatingroundtripper"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/pflag"
 	corev1 "k8s.io/api/core/v1"
@@ -139,6 +140,9 @@ func main() {
 	}
 	config := mgr.GetConfig()
 
+	// create a config that validates we're creating objects with labels
+	validatingConfig := validatingroundtripper.Wrap(config)
+
 	versionedConfigClient, err := configclientset.NewForConfig(config)
 	if err != nil {
 		logger.WithError(err).Fatal("error configuring openshift proxy client")
@@ -147,7 +151,7 @@ func main() {
 	if err != nil {
 		logger.WithError(err).Fatal("error configuring config client")
 	}
-	opClient, err := operatorclient.NewClientFromRestConfig(config)
+	opClient, err := operatorclient.NewClientFromRestConfig(validatingConfig)
 	if err != nil {
 		logger.WithError(err).Fatal("error configuring operator client")
 	}
--- a/pkg/controller/bundle/bundle_unpacker.go
+++ b/pkg/controller/bundle/bundle_unpacker.go
@@ -86,11 +86,19 @@ func newBundleUnpackResult(lookup *operatorsv1alpha1.BundleLookup) *BundleUnpack
 
 func (c *ConfigMapUnpacker) job(cmRef *corev1.ObjectReference, bundlePath string, secrets []corev1.LocalObjectReference, annotationUnpackTimeout time.Duration) *batchv1.Job {
 	job := &batchv1.Job{
+		ObjectMeta: metav1.ObjectMeta{
+			Labels: map[string]string{
+				install.OLMManagedLabelKey: install.OLMManagedLabelValue,
+			},
+		},
 		Spec: batchv1.JobSpec{
 			//ttlSecondsAfterFinished: 0 // can use in the future to not have to clean up job
 			Template: corev1.PodTemplateSpec{
 				ObjectMeta: metav1.ObjectMeta{
 					Name: cmRef.Name,
+					Labels: map[string]string{
+						install.OLMManagedLabelKey: install.OLMManagedLabelValue,
+					},
 				},
 				Spec: corev1.PodSpec{
 					// With restartPolicy = "OnFailure" when the spec.backoffLimit is reached, the job controller will delete all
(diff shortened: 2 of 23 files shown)


ncdc:
Was the hack commit for your local testing, to confirm things were doing what you expected, and that you didn't miss adding labels anywhere in the code base?

cmd/olm/main.go
```diff
@@ -82,6 +85,33 @@ func init() {
...
 
+type validatingRoundTripper struct {
+	delegate http.RoundTripper
+}
+
+func (rt *validatingRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	if req.Method == "POST" {
+		b, err := req.GetBody()
+		if err != nil {
+			panic(err)
```
perdasilva, line 96: Why `panic` vs `return nil, err`? will this kill the process?

pkg/controller/bundle/bundle_unpacker.go
```diff
@@ -86,11 +86,19 @@ func newBundleUnpackResult(lookup *operatorsv1alpha1.BundleLookup) *BundleUnpack
...
+			Labels: map[string]string{
+				install.OLMManagedLabelKey: install.OLMManagedLabelValue,
+			},
+		},
 		Spec: batchv1.JobSpec{
 			//ttlSecondsAfterFinished: 0 // can use in the future to not have to clean up job
 			Template: corev1.PodTemplateSpec{
 				ObjectMeta: metav1.ObjectMeta{
 					Name: cmRef.Name,
+					Labels: map[string]string{
```
perdasilva, line 99: non-blocking - just expressing a thought out loud: is it worth moving the addition of this label to a function? I doubt that the assignment of the label value will evolve much, if at all...

perdasilva (approved):
lgtm - just a couple of non-blocking questions. If, in your judgement, they are not worth addressing, I'm happy to merge

stevekuznetsov:
@ncdc yes, but I could see it being generally useful moving forward to not mistakenly re-introduce the bug. Once the informers are filtered, you'd end up knowing anyway since your controller would not do the right thing but the round-tripper failure would be much more direct.

cmd/olm/main.go
stevekuznetsov: I wanted the stacktrace to clearly show where the offending code was. Yes, this does cause process exit. If we want to keep this around, I'd make it gated on some dev-only flag (`$CI=true`).

pkg/controller/bundle/bundle_unpacker.go
stevekuznetsov: I think it might be more invasive and perhaps less readable to have to pipe every possible occurrence of the label into a mutation function - so I just used literals instead.

cmd/olm/main.go
ncdc: I'd say yes, keep it around, and gate on CI

stevekuznetsov: Done

stevekuznetsov:
@ncdc @perdasilva this should be ready for review

test/e2e/operator_groups_e2e_test.go
```diff
@@ -442,34 +442,38 @@ var _ = Describe("Operator Group", func() {
 			}(ns)
 		}
 
-		// Generate operatorGroupA - OwnNamespace
```
stevekuznetsov, line 445: If you're going to have concepts in the code like "operatorGroupA" but the value in the cluster will be auto-generated, it's imperative to print that relationship for folks that debug your test.

test/e2e/operator_groups_e2e_test.go
```diff
@@ -426,8 +426,8 @@ var _ = Describe("Operator Group", func() {
 		// Create crd so csv succeeds
 		// Ensure clusterroles created and aggregated for access provided APIs
 
-		// Generate namespaceA
```
stevekuznetsov, line 429: If you're going to have phases of a test, use a logger to make them appear in the output to make debugging failures easier.

test/e2e/util.go
```diff
@@ -346,7 +346,7 @@ func catalogSourceRegistryPodSynced(catalog *operatorsv1alpha1.CatalogSource) bo
 	if connState != nil {
 		state = connState.LastObservedState
 	}
-	fmt.Printf("waiting for catalog pod %v to be available (for sync) - %s\n", catalog.GetName(), state)
+	fmt.Printf("waiting for catalog pod %s/%s to be available (for sync) - %s\n", catalog.GetNamespace(), catalog.GetName(), state)
```
stevekuznetsov, line 349: Always fully qualify the object you're waiting on so that folks debugging a failing test know where to look.

test/e2e/csv_e2e_test.go
```diff
@@ -4432,17 +4432,18 @@ func fetchCSV(c versioned.Interface, name, namespace string, checker csvConditio
 	var fetchedCSV *operatorsv1alpha1.ClusterServiceVersion
 	var err error
 
+	ctx.Ctx().Logf("waiting for CSV %s/%s to reach condition", namespace, name)
```
stevekuznetsov, line 4435: If a function like `fetchCSV` is going to validate conditions, at least log something to make it less surprising.

perdasilva (approved):
lgtm

ncdc (approved):

awgreene (approved):
This looks good to me and I wish that we had done this earlier. @perdasilva and I were recently discussing how OLM "adopts/overwrites" existing k8s resources if there is a name collision. Had we considered using labels in the past we could have written OLM such that it wouldn't delete/update resources missing these labels. Now that OLM is out in the wild there isn't a good migration path.

pkg/controller/operators/catalog/operator.go
```diff
@@ -2252,6 +2281,10 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
 					}
 					rb.SetOwnerReferences(updated)
 					rb.SetNamespace(namespace)
+					if rb.Labels == nil {
+						rb.Labels = map[string]string{}
+					}
+					rb.Labels[install.OLMManagedLabelKey] = install.OLMManagedLabelValue
```
tmshort, line 2287: I am seeing this idiom so much, it almost makes sense to create a utility API to initialize the labels and then add the label.

e.g. 
```
addlabel(&rbLabels, install.OLMManagedLabelKey, install.OLMManagedLabelValue)

tmshort (approved):
/lgtm
Some minor comments that don't really need changes.

PR #3020, opened 2023-08-25
OCPBUGS-17157: *: label k8s objects we own

stevekuznetsov:


stevekuznetsov:
I think I will test this once I add logic to startup that ensures we reach a steady-state where we can filter down the WATCH streams. Will be best to test on upgrade.

pkg/controller/operators/catalog/operator.go
```diff
@@ -363,21 +371,85 @@ func NewOperator(ctx context.Context, kubeconfigPath string, clock utilclock.Clo
...
+	)); err != nil {
+		return nil, err
+	}
+
 	// Wire Services
 	serviceInformer := k8sInformerFactory.Core().V1().Services()
 	op.lister.CoreV1().RegisterServiceLister(metav1.NamespaceAll, serviceInformer.Lister())
 	sharedIndexInformers = append(sharedIndexInformers, serviceInformer.Informer())
 
+	// TODO(skuznets): some services don't seem to have any marker to key off of, but they match the operator name
```
ncdc, line 442: Since I haven't been involved deeply in this - could you please explain this comment?

pkg/controller/operators/catalog/operator.go
```diff
@@ -392,6 +464,23 @@ func NewOperator(ctx context.Context, kubeconfigPath string, clock utilclock.Clo
 	op.lister.CoreV1().RegisterPodLister(metav1.NamespaceAll, csPodInformer.Lister())
 	sharedIndexInformers = append(sharedIndexInformers, csPodInformer.Informer())
 
+	if err := labelObjects(corev1.SchemeGroupVersion.WithResource("pods"), csPodInformer.Informer(), labeller.ObjectLabeler[*corev1.Pod, *corev1applyconfigurations.PodApplyConfiguration](
+		ctx, op.logger, func(object metav1.Object) bool {
+			if labels := object.GetLabels(); labels != nil {
+				if _, ok := labels[reconciler.CatalogSourceLabelKey]; ok {
```
ncdc, line 470: I take it the only pods we care about are those labeled as a catalog source?

pkg/controller/operators/catalog/operator.go
```diff
@@ -416,6 +505,27 @@ func NewOperator(ctx context.Context, kubeconfigPath string, clock utilclock.Clo
 	jobInformer := k8sInformerFactory.Batch().V1().Jobs()
 	sharedIndexInformers = append(sharedIndexInformers, jobInformer.Informer())
 
+	if err := labelObjects(batchv1.SchemeGroupVersion.WithResource("jobs"), jobInformer.Informer(), labeller.ObjectLabeler[*batchv1.Job, *batchv1applyconfigurations.JobApplyConfiguration](
+		ctx, op.logger, func(object metav1.Object) bool {
+			job, ok := object.(*batchv1.Job)
+			if !ok {
+				return false
```
ncdc, line 512: Would we ever get the wrong type here?

pkg/controller/operators/labeller/labels.go
```diff
@@ -0,0 +1,153 @@
...
+type ApplyConfig[T any] interface {
+	WithLabels(map[string]string) T
+}
+
+type Client[A ApplyConfig[A], T metav1.Object] interface {
+	Apply(ctx context.Context, cfg ApplyConfig[A], opts metav1.ApplyOptions) (result T, err error)
+}
+
+func hasLabel(obj metav1.Object) bool {
+	if labels := obj.GetLabels(); labels != nil {
```
ncdc, line 33: Can omit the nil check

pkg/controller/operators/labeller/labels.go
```diff
@@ -0,0 +1,153 @@
...
+
+func ObjectLabeler[T metav1.Object, A ApplyConfig[A]](
+	ctx context.Context,
+	logger *logrus.Logger,
+	check func(metav1.Object) bool,
+	applyConfigFor func(name, namespace string) A,
+	apply func(namespace string, ctx context.Context, cfg A, opts metav1.ApplyOptions) (T, error),
+) func(
+	obj interface{},
+) error {
```
ncdc, line 49: Maybe, for readability, give this a custom type?

pkg/controller/operators/catalog/operator.go
stevekuznetsov: I don't know, and I would love for someone with more knowledge to help here. It looks like there's a set of service objects here that cannot be identified as having been created by OLM with any definitive data, rather it may be possible to infer it by their name, though.

pkg/controller/operators/labeller/labels.go
```diff
@@ -0,0 +1,153 @@
...
+	return func(obj interface{}) error {
+		cast, ok := obj.(T)
+		if !ok {
```
ncdc, line 52: Would it be correct to assume that if we get a cast failure, it's entirely programmer error? And panicking would be better?

pkg/controller/operators/catalog/operator.go
stevekuznetsov: There's four sets of Pods we care about, but the rest all have dedicated informers already and are labeled. This particular one is just for catalog sources, yes.

pkg/controller/operators/catalog/operator.go
stevekuznetsov: No, but I'd rather write code that did not panic and crash-loop just in case. What would you suggest?

pkg/controller/operators/labeller/labels.go
```diff
@@ -0,0 +1,153 @@
...
+		}
+	}
+	return false
+}
+
+func ObjectLabeler[T metav1.Object, A ApplyConfig[A]](
+	ctx context.Context,
+	logger *logrus.Logger,
+	check func(metav1.Object) bool,
+	applyConfigFor func(name, namespace string) A,
```
ncdc, line 45: I usually put ns first (scope from broadest to narrowest)

pkg/controller/operators/labeller/labels.go
```diff
@@ -0,0 +1,153 @@
...
+			logger.WithError(err).Error("casting failed")
+			return fmt.Errorf("casting failed: %w", err)
+		}
+
+		if !check(cast) || hasLabel(cast) {
+			return nil
+		}
+
+		cfg := applyConfigFor(cast.GetName(), cast.GetNamespace())
+		cfg.WithLabels(map[string]string{
```
ncdc, line 63: I'm guessing in the implementations, this modifies the receiver?

pkg/controller/operators/labeller/labels.go
stevekuznetsov: Sorry, a bit hard to trace what you're commenting on here - the `func(obj interface{}) error` ?

ncdc: Yes, the return

pkg/controller/operators/labeller/labels.go
ncdc: I think I'd prefer a panic such that we fail as quickly as possible in a noticeable manner, vs. hoping someone notices a log message. WDYT?

stevekuznetsov: I don't have a strong opinion, just modeled this after what controllers normally do when e.g. casting the queue key to a string before parsing it as `namespace/name` - if any of that fails, it's definitely programmer error, but none of it panics

ncdc: Ok, let's not introduce panics to just a portion. Fine to leave as you have it. Thanks.

pkg/controller/operators/catalog/operator.go
perdasilva: @kevinrizza any pointers?

pkg/controller/operators/catalog/operator.go
```diff
@@ -416,6 +501,27 @@ func NewOperator(ctx context.Context, kubeconfigPath string, clock utilclock.Clo
...
+			}
+			for _, container := range job.Spec.Template.Spec.Containers {
+				if strings.Join(container.Command[0:3], " ") == "opm alpha bundle extract" {
```
joelanford, line 511: > container.Command[0:3]

This will panic if the container command does not have at least 4 items, right?

pkg/controller/operators/catalog/operator.go
```diff
@@ -416,6 +501,27 @@ func NewOperator(ctx context.Context, kubeconfigPath string, clock utilclock.Clo
...
+					return true
```
joelanford, line 512: If something/someone else runs a job that looks like this, what are the implications? We'll apply a label to it and then start getting events/copies in our cache?

Does (or, in the future, could) that label imply OLM ownership and therefore other side-effects that shouldn't happen with what is intended to be a non-OLM-owned job?

stevekuznetsov:
> And this is where I'm confused: Why do we need informers and syncers for this? Shouldn't the thing(s) that create/sync these objects in the first place be where the labels are applied?

We need both - #3017 went in to label new resources we create, but we need this PR to label resources created by older versions of OLM that upgrade to this one.

pkg/controller/operators/catalog/operator.go
kevinrizza: I'm honestly not sure without more info. Are there any examples of the services you're talking about? What operators are you installing to check? Certainly it's possible that an operator happens to create this themselves in some cases

pkg/controller/operators/olm/operator.go
```diff
@@ -481,6 +519,18 @@ func newOperatorWithConfig(ctx context.Context, config *operatorConfig) (*Operat
 		return nil, err
 	}
 
+	if err := labelObjects(rbacv1.SchemeGroupVersion.WithResource("clusterroles"), clusterRoleInformer.Informer(), labeller.ObjectLabeler[*rbacv1.ClusterRole, *rbacv1applyconfigurations.ClusterRoleApplyConfiguration](
+		ctx, op.logger, labeller.HasOLMOwnerRef,
+		func(name, _ string) *rbacv1applyconfigurations.ClusterRoleApplyConfiguration {
```
awgreene, line 524: Cool, I haven't seen `_ string` used before.

awgreene (approved):
/approve

awgreene:
Nice work @stevekuznetsov, thanks for the PR.

pkg/controller/operators/labeller/labels.go
```diff
@@ -0,0 +1,160 @@
...
+				return true
+			}
+		}
+	}
+	hasOLMOwnerLabels := true
+	for _, label := range []string{ownerutil.OwnerKey, ownerutil.OwnerNamespaceKey, ownerutil.OwnerKind} {
+		_, exists := object.GetLabels()[label]
+		hasOLMOwnerLabels = hasOLMOwnerLabels && exists
+	}
+	return hasOLMOwnerLabels
```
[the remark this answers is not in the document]
stevekuznetsov: Since N=3 I don't know that it makes any meaningful difference

PR #3021, opened 2023-08-28
OCPBUGS-17157: *: filter informers when preconditions are met

stevekuznetsov:
*: filter informers when preconditions are met

When we can detect at startup time that all of the objects we're about
to look at have the labels we're expecting, we can filter our informer
factories upfront.

---

depends on #3020

### DIFF
--- a/pkg/controller/operators/labeller/filters.go
+++ b/pkg/controller/operators/labeller/filters.go
@@ -1,0 +1,112 @@
+package labeller
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"sync"
+
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/registry/reconciler"
+	"github.com/sirupsen/logrus"
+	"golang.org/x/sync/errgroup"
+	appsv1 "k8s.io/api/apps/v1"
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/metadata"
+
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/operators/internal/alongside"
+)
+
+func Filter(gvr schema.GroupVersionResource) func(metav1.Object) bool {
+	if f, ok := filters[gvr]; ok {
+		return f
+	}
+	return func(object metav1.Object) bool {
+		return false
+	}
+}
+
+func JobFilter(getConfigMap func(namespace, name string) (metav1.Object, error)) func(object metav1.Object) bool {
+	return func(object metav1.Object) bool {
+		for _, ownerRef := range object.GetOwnerReferences() {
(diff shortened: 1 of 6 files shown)


pkg/controller/operators/labeller/filters.go
```diff
@@ -0,0 +1,106 @@
+package labeller
```
perdasilva, line 1: should we add some unit tests around this package?

pkg/controller/operators/labeller/filters.go
```diff
@@ -0,0 +1,106 @@
...
+				return true
+			}
+		}
+		return false
+	},
+	appsv1.SchemeGroupVersion.WithResource("deployments"):          HasOLMOwnerRef,
+	rbacv1.SchemeGroupVersion.WithResource("roles"):                HasOLMOwnerRef,
+	rbacv1.SchemeGroupVersion.WithResource("rolebindings"):         HasOLMOwnerRef,
+	rbacv1.SchemeGroupVersion.WithResource("clusterroles"):         HasOLMOwnerRef,
+	rbacv1.SchemeGroupVersion.WithResource("clusterrolebindingss"): HasOLMOwnerRef,
```
perdasilva, line 62: If I'm getting this right, we're applying filters to identify the resources OLM manages. Will all the current installations' managed resources have the necessary labels/annotations? I guess the right question is, what's the risk of this upgrading and losing track of previously managed resources? My thought it that is should low, but I don't know enough about the guts to have a high level of certainty there isn't something I'm missing =S

stevekuznetsov: It's a good question. I hope someone with more OLMv0 knowledge can chime in here - I've been doing ad-hoc testing on existing clusters with v0 and in my tests these filters are sufficient to find all existing data OLM created in the past, but I don't know if OLMv0 had every code path exercised there.

stevekuznetsov: @awgreene @joelanford @kevinrizza PTAL

pkg/controller/operators/labeller/filters.go
```diff
@@ -0,0 +1,106 @@
...
+		for key := range object.GetAnnotations() {
+			if strings.HasPrefix(key, alongside.AnnotationPrefix) {
+				return true
+			}
+		}
+		return false
+	},
+}
+
+func Validate(ctx context.Context, logger *logrus.Logger, metadataClient metadata.Interface) (bool, error) {
```
ncdc, line 73: Will it be safe to remove this check at some point, e.g. after we know everything has been labeled?

stevekuznetsov: Yes, whenever we are at a version where all upgrade paths have gone through some state that has the labeling controller in, we can remove this precondition, the label controllers, etc, and just filter by default.

ncdc: Would it be useful to place a sentinel ConfigMap somewhere that indicates 100% labeling, and check for that, so we don't have to list all the things each time we startup?

stevekuznetsov: There's no capital C consistent way to ensure that is always correct, since storage can be sharded by resource. We're only doing partial object metadata lists here, so I *think* it should be ok?

stevekuznetsov: Actually ... for Jobs we need the spec :facep

stevekuznetsov:
I can't get any of the fail-forward e2e tests to fail locally :|

tmshort (approved):
/lgtm
Seems reasonable.

PR #3024, opened 2023-08-31
Valid service from catsrc with periods in name

awgreene:
Problem: Creating a catalogSource with a period in the name is valid as the Kuberenetes API catalogSources to adhere to the following naing conventions:
- Start and end with a lower case alphanumeric character
- Consist only of alphanumeric characters, `.`, or `-`

Unfortunately, the service created by OLM for the catalogSource must adhere to the following naming conventions:
- Start and end with a lower case alphanumeric character
- Consist only of alphanumeric characters or `-`

This causes OLM to constantly recreate the catalogSource as the service it attempts to create is rejected from the Kubernetes API service due to the `.` in the name.

Solution: When naming the service, replace all instances of `.` in the name with `-`.

### DIFF
--- a/pkg/controller/registry/reconciler/grpc.go
+++ b/pkg/controller/registry/reconciler/grpc.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"hash/fnv"
+	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


oceanc80:
nit: We may want to add/update docs to make users aware of this so that they aren't confused when they go to look for their service. 
/lgtm

PR #3028, opened 2023-09-07
OCPBUGS-17157: *: detect when all objects are labelled, restart

stevekuznetsov:
When all of the k8s objects that need labels have them, we are good to exit the process. The next Pod that start up will detect that all labels are present and be able to filter informers going forward.

### DIFF
--- a/cmd/olm/manager.go
+++ b/cmd/olm/manager.go
@@ -3,10 +3,15 @@ package main
 import (
 	"context"
 
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/selection"
+	apiregistrationv1 "k8s.io/kube-aggregator/pkg/apis/apiregistration/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -51,8 +56,37 @@ func Manager(ctx context.Context, debug bool) (ctrl.Manager, error) {
 		Scheme:             scheme,
 		MetricsBindAddress: "0", // TODO(njhale): Enable metrics on non-conflicting port (not 8080)
 		Cache: cache.Options{
-			DefaultLabelSelector: labels.SelectorFromValidatedSet(map[string]string{install.OLMManagedLabelKey: install.OLMManagedLabelValue}),
 			ByObject: map[client.Object]cache.ByObject{
+				&appsv1.Deployment{}: {
+					Label: labels.SelectorFromValidatedSet(map[string]string{install.OLMManagedLabelKey: install.OLMManagedLabelValue}),
+				},
+				&corev1.Service{}: {
+					Label: labels.SelectorFromValidatedSet(map[string]string{install.OLMManagedLabelKey: install.OLMManagedLabelValue}),
+				},
+				&apiextensionsv1.CustomResourceDefinition{}: {
+					Label: labels.SelectorFromValidatedSet(map[string]string{install.OLMManagedLabelKey: install.OLMManagedLabelValue}),
+				},
+				&apiregistrationv1.APIService{}: {
+					Label: labels.SelectorFromValidatedSet(map[string]string{install.OLMManagedLabelKey: install.OLMManagedLabelValue}),
+				},
+				&corev1.ConfigMap{}: {
+					Label: labels.SelectorFromValidatedSet(map[string]string{install.OLMManagedLabelKey: install.OLMManagedLabelValue}),
+				},
+				&corev1.ServiceAccount{}: {
+					Label: labels.SelectorFromValidatedSet(map[string]string{install.OLMManagedLabelKey: install.OLMManagedLabelValue}),
+				},
+				&rbacv1.Role{}: {
+					Label: labels.SelectorFromValidatedSet(map[string]string{install.OLMManagedLabelKey: install.OLMManagedLabelValue}),
+				},
+				&rbacv1.RoleBinding{}: {
+					Label: labels.SelectorFromValidatedSet(map[string]string{install.OLMManagedLabelKey: install.OLMManagedLabelValue}),
+				},
+				&rbacv1.ClusterRole{}: {
+					Label: labels.SelectorFromValidatedSet(map[string]string{install.OLMManagedLabelKey: install.OLMManagedLabelValue}),
+				},
+				&rbacv1.ClusterRoleBinding{}: {
+					Label: labels.SelectorFromValidatedSet(map[string]string{install.OLMManagedLabelKey: install.OLMManagedLabelValue}),
+				},
 				&operatorsv1alpha1.ClusterServiceVersion{}: {
 					Label: copiedLabelDoesNotExist,
 				},
--- a/cmd/olm/manager.go
+++ b/cmd/olm/manager.go
@@ -3,11 +3,15 @@ package main
 import (
 	"context"
 
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/selection"
+	apiregistrationv1 "k8s.io/kube-aggregator/pkg/apis/apiregistration/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -53,6 +57,36 @@ func Manager(ctx context.Context, debug bool) (ctrl.Manager, error) {
 		MetricsBindAddress: "0", // TODO(njhale): Enable metrics on non-conflicting port (not 8080)
 		Cache: cache.Options{
 			ByObject: map[client.Object]cache.ByObject{
+				&appsv1.Deployment{}: {
+					Label: labels.SelectorFromValidatedSet(map[string]string{install.OLMManagedLabelKey: install.OLMManagedLabelValue}),
+				},
+				&corev1.Service{}: {
+					Label: labels.SelectorFromValidatedSet(map[string]string{install.OLMManagedLabelKey: install.OLMManagedLabelValue}),
+				},
+				&apiextensionsv1.CustomResourceDefinition{}: {
+					Label: labels.SelectorFromValidatedSet(map[string]string{install.OLMManagedLabelKey: install.OLMManagedLabelValue}),
+				},
+				&apiregistrationv1.APIService{}: {
+					Label: labels.SelectorFromValidatedSet(map[string]string{install.OLMManagedLabelKey: install.OLMManagedLabelValue}),
+				},
+				&corev1.ConfigMap{}: {
+					Label: labels.SelectorFromValidatedSet(map[string]string{install.OLMManagedLabelKey: install.OLMManagedLabelValue}),
+				},
+				&corev1.ServiceAccount{}: {
+					Label: labels.SelectorFromValidatedSet(map[string]string{install.OLMManagedLabelKey: install.OLMManagedLabelValue}),
+				},
+				&rbacv1.Role{}: {
+					Label: labels.SelectorFromValidatedSet(map[string]string{install.OLMManagedLabelKey: install.OLMManagedLabelValue}),
+				},
+				&rbacv1.RoleBinding{}: {
+					Label: labels.SelectorFromValidatedSet(map[string]string{install.OLMManagedLabelKey: install.OLMManagedLabelValue}),
+				},
+				&rbacv1.ClusterRole{}: {
+					Label: labels.SelectorFromValidatedSet(map[string]string{install.OLMManagedLabelKey: install.OLMManagedLabelValue}),
+				},
+				&rbacv1.ClusterRoleBinding{}: {
+					Label: labels.SelectorFromValidatedSet(map[string]string{install.OLMManagedLabelKey: install.OLMManagedLabelValue}),
+				},
 				&corev1.Secret{}: {
 					Label: labels.SelectorFromValidatedSet(map[string]string{install.OLMManagedLabelKey: install.OLMManagedLabelValue}),
 				},
--- a/cmd/olm/manager.go
+++ b/cmd/olm/manager.go
@@ -3,15 +3,10 @@ package main
 import (
 	"context"
 
-	appsv1 "k8s.io/api/apps/v1"
-	corev1 "k8s.io/api/core/v1"
-	rbacv1 "k8s.io/api/rbac/v1"
-	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/selection"
-	apiregistrationv1 "k8s.io/kube-aggregator/pkg/apis/apiregistration/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -56,40 +51,8 @@ func Manager(ctx context.Context, debug bool) (ctrl.Manager, error) {
 		Scheme:             scheme,
 		MetricsBindAddress: "0", // TODO(njhale): Enable metrics on non-conflicting port (not 8080)
 		Cache: cache.Options{
+			DefaultLabelSelector: labels.SelectorFromValidatedSet(map[string]string{install.OLMManagedLabelKey: install.OLMManagedLabelValue}),
 			ByObject: map[client.Object]cache.ByObject{
-				&appsv1.Deployment{}: {
-					Label: labels.SelectorFromValidatedSet(map[string]string{install.OLMManagedLabelKey: install.OLMManagedLabelValue}),
-				},
-				&corev1.Service{}: {
-					Label: labels.SelectorFromValidatedSet(map[string]string{install.OLMManagedLabelKey: install.OLMManagedLabelValue}),
-				},
-				&apiextensionsv1.CustomResourceDefinition{}: {
-					Label: labels.SelectorFromValidatedSet(map[string]string{install.OLMManagedLabelKey: install.OLMManagedLabelValue}),
-				},
-				&apiregistrationv1.APIService{}: {
-					Label: labels.SelectorFromValidatedSet(map[string]string{install.OLMManagedLabelKey: install.OLMManagedLabelValue}),
-				},
-				&corev1.ConfigMap{}: {
-					Label: labels.SelectorFromValidatedSet(map[string]string{install.OLMManagedLabelKey: install.OLMManagedLabelValue}),
-				},
-				&corev1.ServiceAccount{}: {
-					Label: labels.SelectorFromValidatedSet(map[string]string{install.OLMManagedLabelKey: install.OLMManagedLabelValue}),
-				},
-				&rbacv1.Role{}: {
-					Label: labels.SelectorFromValidatedSet(map[string]string{install.OLMManagedLabelKey: install.OLMManagedLabelValue}),
-				},
-				&rbacv1.RoleBinding{}: {
-					Label: labels.SelectorFromValidatedSet(map[string]string{install.OLMManagedLabelKey: install.OLMManagedLabelValue}),
-				},
-				&rbacv1.ClusterRole{}: {
-					Label: labels.SelectorFromValidatedSet(map[string]string{install.OLMManagedLabelKey: install.OLMManagedLabelValue}),
-				},
-				&rbacv1.ClusterRoleBinding{}: {
-					Label: labels.SelectorFromValidatedSet(map[string]string{install.OLMManagedLabelKey: install.OLMManagedLabelValue}),
-				},
-				&corev1.Secret{}: {
-					Label: labels.SelectorFromValidatedSet(map[string]string{install.OLMManagedLabelKey: install.OLMManagedLabelValue}),
-				},
 				&operatorsv1alpha1.ClusterServiceVersion{}: {
 					Label: copiedLabelDoesNotExist,
 				},
--- a/pkg/controller/install/certresources.go
+++ b/pkg/controller/install/certresources.go
@@ -456,7 +456,7 @@ func (i *StrategyDeploymentInstaller) installCertRequirementsForDeployment(deplo
 			Name:     "system:auth-delegator",
 		},
 	}
-	authDelegatorClusterRoleBinding.SetName(service.GetName() + "-system:auth-delegator")
+	authDelegatorClusterRoleBinding.SetName(AuthDelegatorClusterRoleBindingName(service.GetName()))
 	authDelegatorClusterRoleBinding.SetLabels(map[string]string{OLMManagedLabelKey: OLMManagedLabelValue})
 
 	existingAuthDelegatorClusterRoleBinding, err := i.strategyClient.GetOpLister().RbacV1().ClusterRoleBindingLister().Get(authDelegatorClusterRoleBinding.GetName())
@@ -504,7 +504,7 @@ func (i *StrategyDeploymentInstaller) installCertRequirementsForDeployment(deplo
 			Name:     "extension-apiserver-authentication-reader",
 		},
 	}
-	authReaderRoleBinding.SetName(service.GetName() + "-auth-reader")
+	authReaderRoleBinding.SetName(AuthReaderRolebindingName(service.GetName()))
 	authReaderRoleBinding.SetNamespace(KubeSystem)
 	authReaderRoleBinding.SetLabels(map[string]string{OLMManagedLabelKey: OLMManagedLabelValue})
 
@@ -541,6 +541,14 @@ func (i *StrategyDeploymentInstaller) installCertRequirementsForDeployment(deplo
 	// is used by the apiserver if not hot reloading.
 	SetCAAnnotation(&depSpec, caHash)
 	return &depSpec, caPEM, nil
+}
+
+func AuthDelegatorClusterRoleBindingName(serviceName string) string {
+	return serviceName + "-system:auth-delegator"
+}
+
+func AuthReaderRolebindingName(serviceName string) string {
+	return serviceName + "-auth-reader"
 }
 
 func SetCAAnnotation(depSpec *appsv1.DeploymentSpec, caHash string) {
--- a/pkg/controller/install/certresources.go
+++ b/pkg/controller/install/certresources.go
@@ -504,7 +504,7 @@ func (i *StrategyDeploymentInstaller) installCertRequirementsForDeployment(deplo
 			Name:     "extension-apiserver-authentication-reader",
 		},
 	}
-	authReaderRoleBinding.SetName(AuthReaderRolebindingName(service.GetName()))
+	authReaderRoleBinding.SetName(AuthReaderRoleBindingName(service.GetName()))
 	authReaderRoleBinding.SetNamespace(KubeSystem)
 	authReaderRoleBinding.SetLabels(map[string]string{OLMManagedLabelKey: OLMManagedLabelValue})
 
@@ -547,7 +547,7 @@ func AuthDelegatorClusterRoleBindingName(serviceName string) string {
 	return serviceName + "-system:auth-delegator"
 }
 
-func AuthReaderRolebindingName(serviceName string) string {
+func AuthReaderRoleBindingName(serviceName string) string {
 	return serviceName + "-auth-reader"
 }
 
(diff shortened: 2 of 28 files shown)


pkg/controller/operators/labeller/labels.go
```diff
@@ -40,28 +41,49 @@ func ObjectLabeler[T metav1.Object, A ApplyConfig[A]](
...
+					return nil
+				}
+				gvrFullyLabelled := true
+				for _, item := range items {
+					gvrFullyLabelled = gvrFullyLabelled && (!check(item) || hasLabel(item))
+				}
+				if gvrFullyLabelled {
+					allObjectsLabelled := done()
+					if allObjectsLabelled {
+						logrus.Fatal("detected that every object is labelled, exiting...")
```
perdasilva, line 73: nit: should we use Info or Debug? Fatal could be misinterpreted here as somethings gone wrong. (maybe add exiting _gracefully_ - to further indicate all is well?)

pkg/controller/operators/labeller/labels.go
```diff
@@ -71,58 +93,77 @@ func ObjectPatchLabeler(
...
+					}
+				}
+				return nil
+			}
 
-		uid := cast.GetUID()
-		rv := cast.GetResourceVersion()
+			uid := cast.GetUID()
```
perdasilva, line 130: is it worth moving this step into its own function just to reduce the footprint of this function a bit?

perdasilva: this = 130 - 165

pkg/controller/operators/olm/operator.go
```diff
@@ -449,18 +450,31 @@ func newOperatorWithConfig(ctx context.Context, config *operatorConfig) (*Operat
 		}
 	}
 
-	labelObjects := func(gvr schema.GroupVersionResource, informer cache.SharedIndexInformer, sync queueinformer.LegacySyncHandler) error {
+	complete := map[schema.GroupVersionResource]bool{}
+	completeLock := &sync.RWMutex{}
```
perdasilva, line 454: nit: RW lock that doesn't seem to use RLock(). Could probably just use regular Mutex.

pkg/controller/operators/olm/operator.go
```diff
@@ -449,18 +450,31 @@ func newOperatorWithConfig(ctx context.Context, config *operatorConfig) (*Operat
...
 		op.k8sLabelQueueSets[gvr] = workqueue.NewRateLimitingQueueWithConfig(workqueue.DefaultControllerRateLimiter(), workqueue.RateLimitingQueueConfig{
 			Name: gvr.String(),
 		})
 		queueInformer, err := queueinformer.NewQueueInformer(
 			ctx,
 			queueinformer.WithLogger(op.logger),
 			queueinformer.WithInformer(informer),
-			queueinformer.WithSyncer(sync.ToSyncer()),
+			queueinformer.WithSyncer(sync(func() bool {
+				completeLock.Lock()
```
perdasilva, line 469: suggestion: I like the lock() defer unlock() pattern because it's robust against changes b/w the lock() and unlock() calls.

pkg/controller/operators/labeller/labels.go
stevekuznetsov: I'm using `Fatal` to exit the process, not sure Debug + `os.Exit()` would be better or worse TBH

pkg/controller/operators/labeller/labels.go
stevekuznetsov: If possible, we prefer not to touch this code - the PR I made upstream to get applyconfigurations created for CRDs is in, and backported, so the next time we bump k8s deps we can delete this function

pkg/controller/operators/olm/operator.go
stevekuznetsov: I've seen both as a matter of opinion. I prefer the non-defer case in small critical sections with trivial control flow - not sure if true still but also in the past the defer approach was a couple orders of magnitude slower, which is not great when placed into a high-churn callsite.

pkg/controller/operators/labeller/labels.go
perdasilva: Conceptually, I'm worried about the exit code. If you `os.Exit(0)` the replica set should still bring the pod back, right? Alternatively, just s/exiting/restarting OLM to filter informers (or some such) - When something is interrupted, I think there's value in knowing (ambiguously) that it's doing what it's meant to do (keeping in mind that not all log readers are maintain the code).

On a different point, I wanted to sanity check that there isn't a race condition L61-L70, could we just think that every object was detected?

stevekuznetsov: I see. I'll update the log + exit code to be more explicit, but FWIW it is not relevant to bringing back a new Pod.

> On a different point, I wanted to sanity check that there isn't a race condition L61-L70, could we just think that every object was detected?

What race do you think we have? We lock around the `done()` call. In any case, when this exits and a new process re-starts, it will re-evaluate the cluster state as opposed to trusting the last process.

perdasilva: It's just a sanity check, and my thought process might be grounded on bad assumptions (maybe we cleared this up with the previous discussion on how the queues are drained). But, just to lay out my reasoning, I just read:
1. query
2. process result
3. bail the process if 2. checks out

I just didn't know if the query results can change while you're doing 2 and 3. From the comment (L58), I gathered that the operator determines it's in a particular state (i.e. all resource labeled), and restarts the pod to manage the queues differently (filtering informers) - so, all good ^^

pkg/controller/operators/olm/operator.go
perdasilva: Great call out - I didn't know that. Thank you ^^

pkg/controller/operators/labeller/labels.go
stevekuznetsov: It's a good question - the fact that we re-check on startup is key

perdasilva (approved):
lgtm

stevekuznetsov:
This should now only fail one e2e - `User defined service account [It] with retry`. That test does the following:

 - create an `OperatorGroup` with a user-provided service account
 - omit any permissions for that service account
 - create a `Subscription` and wait for the `InstallPlan` to enter a terminal failure state
 - create permissions
 - expect the `InstallPlan` to exit the terminal state

We either need to not wait for the terminal state or allow permissions errors to be re-tried forever. @joelanford please advise.

pkg/controller/operators/olm/apiservices.go
```diff
@@ -164,60 +162,27 @@ func (a *Operator) checkAPIServiceResources(csv *v1alpha1.ClusterServiceVersion,
...
-		serviceAccount, err := a.lister.CoreV1().ServiceAccountLister().ServiceAccounts(deployment.GetNamespace()).Get(serviceAccountName)
+		_, err = a.opClient.KubernetesInterface().CoreV1().ServiceAccounts(deployment.GetNamespace()).Get(context.TODO(), serviceAccountName, metav1.GetOptions{})
 		if err != nil {
-			logger.WithField("serviceaccount", serviceAccountName).Warnf("could not retrieve ServiceAccount")
+			logger.WithError(err).WithField("serviceaccount", serviceAccountName).Warnf("could not retrieve ServiceAccount")
 			errs = append(errs, err)
-			continue
 		}
 
-		// Ensure RBAC permissions for the APIService are correct
```
stevekuznetsov, line 174: Note ... the changes here have us no longer checking for the (Cluster)Role in `kube-system` namespace. I don't see what the point of that logic is - there's no remediation if it's missing, and any conformant Kube installation will have them. If we _really_ think we need to check for those, we can use a live client.

pkg/controller/operators/catalog/operator.go
```diff
@@ -380,10 +380,22 @@ func NewOperator(ctx context.Context, kubeconfigPath string, clock utilclock.Clo
...
 
-	labelObjects := func(gvr schema.GroupVersionResource, informer cache.SharedIndexInformer, sync queueinformer.LegacySyncHandler) error {
+	complete := map[schema.GroupVersionResource][]bool{}
+	completeLock := &sync.Mutex{}
+
+	labelObjects := func(gvr schema.GroupVersionResource, informer cache.SharedIndexInformer, sync func(done func() bool) queueinformer.LegacySyncHandler) error {
 		if canFilter {
 			return nil
 		}
+		var idx int
```
ncdc, line 390: This code hurts my brain. Help me understand it? 😄

pkg/controller/operators/labeller/labels.go
```diff
@@ -40,28 +42,50 @@ func ObjectLabeler[T metav1.Object, A ApplyConfig[A]](
...
+				// the label already has it; in which case we should exit the process, so the Pod that succeeds us can filter
+				// the informers used to drive the controller and stop having to track extraneous objects
+				items, err := list(labels.Everything())
+				if err != nil {
+					logger.WithError(err).Warn("failed to list all objects to check for labelling completion")
+					return nil
+				}
+				gvrFullyLabelled := true
+				for _, item := range items {
+					gvrFullyLabelled = gvrFullyLabelled && (!check(item) || hasLabel(item))
```
ncdc, line 69: can break early if !gvrFullyLabelled?

pkg/controller/operators/catalog/operator.go
```diff
@@ -392,7 +404,18 @@ func NewOperator(ctx context.Context, kubeconfigPath string, clock utilclock.Clo
...
 			queueinformer.WithLogger(op.logger),
 			queueinformer.WithInformer(informer),
-			queueinformer.WithSyncer(sync.ToSyncer()),
+			queueinformer.WithSyncer(sync(func() bool {
+				completeLock.Lock()
+				complete[gvr][idx] = true
+				allDone := true
+				for _, items := range complete {
+					for _, done := range items {
+						allDone = allDone && done
```
ncdc, line 413: break early if false?

pkg/controller/operators/catalog/operator.go
ncdc: 1. We likely call labelObjects multiple times for the same GVR
2. Keep track of syncs for all GVRs for all invocations
3. For each invocation, when the informer for the GVR has synced, set its spot in `complete` to true

HasSynced func:
1. Grab lock
2. Set the complete spot to true
3. Loop through all tracked syncs by GVR
    1. Loop through all invocations for each GVR
4. If all invocations across all GVRs have synced and marked completed, all GVRs we're labeling **in this file** have been synced

?

ncdc: Oh but it's more subtle than that, because `queueinformer.WithSyncer(sync(func() bool {` is actually calling `sync` and the inline function here is the definition for `done`

ncdc: I see what you're doing now, and why. I get it, but it's hard to grok.

pkg/controller/operators/olm/operator.go
```diff
@@ -102,12 +103,57 @@ type Operator struct {
 	clientFactory                clients.Factory
 	plugins                      []plugins.OperatorPlugin
 	informersByNamespace         map[string]*plugins.Informers
+	informersFiltered            bool
+
+	ruleChecker     func(*v1alpha1.ClusterServiceVersion) *install.CSVRuleChecker
+	ruleCheckerLock *sync.RWMutex
```
ncdc, line 109: I think this can be a non-pointer unless you're passing it into other places or you have non-pointer receivers in here?

pkg/controller/install/certresources.go
```diff
@@ -543,6 +543,14 @@ func (i *StrategyDeploymentInstaller) installCertRequirementsForDeployment(deplo
 	return &depSpec, caPEM, nil
 }
 
+func AuthDelegatorClusterRoleBindingName(serviceName string) string {
+	return serviceName + "-system:auth-delegator"
+}
+
+func AuthReaderRolebindingName(serviceName string) string {
```
ncdc, line 550:
```suggestion
func AuthReaderRoleBindingName(serviceName string) string {
```

pkg/controller/operators/olm/requirements.go
```diff
@@ -358,6 +411,59 @@ func (a *Operator) permissionStatus(strategyDetailsDeployment *v1alpha1.Strategy
...
+	if err != nil {
+		return false, fmt.Errorf("failed to hash permission rules: %w", err)
+	}
+	roleSelectorMap := ownerutil.OwnerLabel(csv, v1alpha1.ClusterServiceVersionKind)
+	roleSelectorMap[resolver.ContentHashLabelKey] = ruleHash
+	roleSelectorSet := labels.Set{}
+	for key, value := range roleSelectorMap {
+		roleSelectorSet[key] = value
+	}
+	roleSelector := labels.SelectorFromSet(roleSelectorSet)
```
ncdc, line 431:
```suggestion
	roleSelector := labels.SelectorFromSet(labels.Set(roleSelectorMap))
```

should work?

pkg/controller/operators/olm/requirements.go
```diff
@@ -358,6 +411,59 @@ func (a *Operator) permissionStatus(strategyDetailsDeployment *v1alpha1.Strategy
...
+	}
+
+	if len(roles) == 0 {
+		return false, nil
+	}
+
+	// then, find the (cluster)rolebinding, if we found the role
+	bindingHash, err := resolver.RoleReferenceAndSubjectHashLabelValue(rbacv1.RoleRef{
+		Kind:     "Role",
+		Name:     roles[0].GetName(),
```
ncdc, line 444: We only ever expect 1, or ?

test/e2e/subscription_e2e_test.go
```diff
@@ -2889,9 +2890,9 @@ func fetchSubscription(crc versioned.Interface, namespace, name string, checker
 			return false, err
 		}
 		thisState, thisCSV, thisInstallPlanRef := fetchedSubscription.Status.State, fetchedSubscription.Status.CurrentCSV, fetchedSubscription.Status.InstallPlanRef
-		if thisState != lastState || thisCSV != lastCSV || equality.Semantic.DeepEqual(thisInstallPlanRef, lastInstallPlanRef) {
+		if thisState != lastState || thisCSV != lastCSV || !equality.Semantic.DeepEqual(thisInstallPlanRef, lastInstallPlanRef) {
```
ncdc, line 2893: Seems fairly important 😉

pkg/controller/operators/labeller/labels.go
stevekuznetsov: Yes, but the computation done here is trivial enough that breaking early won't make any meaningful difference in performance and will increase the code length/control flow.

pkg/controller/operators/catalog/operator.go
stevekuznetsov: Yes, but the computation done here is trivial enough that breaking early won't make any meaningful difference in performance and will increase the code length/control flow.

pkg/controller/operators/olm/operator.go
stevekuznetsov: Good point - I think I was passing it in a previous impl.

test/e2e/subscription_e2e_test.go
stevekuznetsov: I wrote this bad code last week so thankfully a small blast radius :)

pkg/controller/operators/olm/requirements.go
stevekuznetsov: Given that one of our labels in the selector is the hash of the rules, we do just expect the one. The workqueue will dedupe keys so I'd expect one worker in the controller to be creating RBAC, not many - so I don't know how we'd race with ourselves to create more than one instance of the role?

pkg/controller/operators/catalog/operator.go
joelanford: Also still trying to grok. I'd suggest adding comments that explain what's going on. This is definitely something I would be re-grok-ing if I have to look at it again later.

pkg/controller/operators/labeller/filters.go
```diff
@@ -49,15 +52,18 @@ func JobFilter(getConfigMap func(namespace, name string) (metav1.Object, error))
 	}
 }
 
+func ServiceAccountFilter(isServiceAccountReferenced func(namespace, name string) bool) func(object metav1.Object) bool {
+	return func(object metav1.Object) bool {
+		return HasOLMOwnerRef(object) || HasOLMLabel(object) || isServiceAccountReferenced(object.GetNamespace(), object.GetName())
+	}
+}
+
```
joelanford, line 60: Am I correct in assuming that these filters that are split out are used in special cases where we don't want (or can't) use the `filters` map?

If so, maybe call out the different use cases somehow? I'm a little concerned that a future commit might add something to the `filters` map that ends up being overwritten by some other code that blindly changes the map key to point to another function.

pkg/controller/operators/catalog/operator.go
stevekuznetsov: Added comments

pkg/controller/operators/labeller/filters.go
stevekuznetsov: Yeah, these are special filters that can't be expressed in a simple way since they require other state inputs. When we're running these filters in the labelling controller, we can answer their other state needs with informers. When we're doing the at-startup check, though, we give them live client calls. I agree that it's easy to break this if you're not careful but I would suggest that we guard against that with careful review to this code in the future rather than some sort of super-clever coding paradigm. I don't know that this needs to get any more complex than it already is.

cmd/olm/manager.go
```diff
@@ -53,6 +57,36 @@ func Manager(ctx context.Context, debug bool) (ctrl.Manager, error) {
 		MetricsBindAddress: "0", // TODO(njhale): Enable metrics on non-conflicting port (not 8080)
 		Cache: cache.Options{
 			ByObject: map[client.Object]cache.ByObject{
+				&appsv1.Deployment{}: {
+					Label: labels.SelectorFromValidatedSet(map[string]string{install.OLMManagedLabelKey: install.OLMManagedLabelValue}),
+				},
```
joelanford, line 62: Nit/question: at this point, is it possible/does it make sense to make this the default selector for everything we cache? And then one-off selectors for objects that don't have this label?

cmd/olm/manager.go
```diff
@@ -53,6 +57,36 @@ func Manager(ctx context.Context, debug bool) (ctrl.Manager, error) {
...
+				&rbacv1.RoleBinding{}: {
+					Label: labels.SelectorFromValidatedSet(map[string]string{install.OLMManagedLabelKey: install.OLMManagedLabelValue}),
+				},
+				&rbacv1.ClusterRole{}: {
+					Label: labels.SelectorFromValidatedSet(map[string]string{install.OLMManagedLabelKey: install.OLMManagedLabelValue}),
+				},
+				&rbacv1.ClusterRoleBinding{}: {
+					Label: labels.SelectorFromValidatedSet(map[string]string{install.OLMManagedLabelKey: install.OLMManagedLabelValue}),
+				},
```
joelanford, line 89: Do we need to include selectors for everything in this list as well?

https://github.com/operator-framework/operator-lifecycle-manager/blob/49533b50e9d2ccac34e38f278cfebb262a9cb276/pkg/controller/operators/catalog/supportedresources.go#L16-L24

These are kinds of objects that can be present in bundles that OLM will install as part of an install plan.

pkg/controller/install/deployment.go
```diff
@@ -152,11 +152,11 @@ func (i *StrategyDeploymentInstaller) deploymentForSpec(name string, spec appsv1
 	dep.Spec.Template.SetAnnotations(annotations)
 
 	// Set custom labels before CSV owner labels
```
joelanford, line 154: Nit: this comment seems backwards now?

pkg/controller/operators/labeller/filters.go
joelanford: > can't be expressed in a simple way since they require other state inputs.

Ah I see that now.

> I would suggest that we guard against that with careful review to this code in the future rather than some sort of super-clever coding paradigm. I don't know that this needs to get any more complex than it already is.

Totally agree! But I don't think a helper function would add too much complexity:
```go
type filtersMap map[schema.GroupVersionResource]func(metav1.Object) bool

func mustAddFilter(m filtersMap, gvr schema.GroupVersionResource, filter func(metav1.Object) bool) {
	if _, ok := m[gvr]; ok {
 		panic("programmer error: cannot add filter to map: already contains key " + gvr.String())
 	}
 	m[gvr] = filter
}
```

pkg/controller/operators/labeller/labels.go
```diff
@@ -40,28 +42,50 @@ func ObjectLabeler[T metav1.Object, A ApplyConfig[A]](
...
+				return fmt.Errorf("casting failed: %w", err)
+			}
 
-		if !check(cast) || hasLabel(cast) {
-			return nil
-		}
+			if !check(cast) || hasLabel(cast) {
+				// if the object we're processing does not need us to label it, it's possible that every object that requires
+				// the label already has it; in which case we should exit the process, so the Pod that succeeds us can filter
+				// the informers used to drive the controller and stop having to track extraneous objects
```
joelanford, line 61: After we exit, how are we sure that no new objects snuck in while we were restarting? Is the assumption that new items will only be added by OLM itself? And since we're running _this_ version of OLM, we know that all new items will be labelled correctly?

pkg/controller/operators/labeller/labels.go
```diff
@@ -40,28 +42,50 @@ func ObjectLabeler[T metav1.Object, A ApplyConfig[A]](
...
+				}
+				gvrFullyLabelled := true
+				for _, item := range items {
+					gvrFullyLabelled = gvrFullyLabelled && (!check(item) || hasLabel(item))
+				}
+				if gvrFullyLabelled {
+					allObjectsLabelled := done()
+					if allObjectsLabelled {
+						logrus.Info("detected that every object is labelled, exiting to re-start the process...")
+						os.Exit(0)
```
joelanford, line 75: Other similar functions are using `logrus.Fatal` which will yield a different exit code. Be consistent?

pkg/controller/operators/olm/apiservices.go
```diff
@@ -164,60 +162,27 @@ func (a *Operator) checkAPIServiceResources(csv *v1alpha1.ClusterServiceVersion,
 		if serviceAccountName == "" {
 			serviceAccountName = "default"
 		}
-		serviceAccount, err := a.lister.CoreV1().ServiceAccountLister().ServiceAccounts(deployment.GetNamespace()).Get(serviceAccountName)
+		_, err = a.opClient.KubernetesInterface().CoreV1().ServiceAccounts(deployment.GetNamespace()).Get(context.TODO(), serviceAccountName, metav1.GetOptions{})
```
joelanford, line 165: How horrible is it to pass a context down?

stevekuznetsov: Horrible, Todd and I looked at this last week and it would be threaded up to the top, missing literally everywhere in this codebase. Happy to tackle the whole cahuna in a follow-up.

pkg/controller/operators/labeller/labels.go
stevekuznetsov: On startup, we call `labeller.Validate()`, so it's not an assumption on that front. We do expect to be labelling only objects we ourselves create, so yes after the call is made to filter informers it's not expected to have someone else making up new objects we need to see but can't.

cmd/olm/manager.go
stevekuznetsov: I don't see anywhere that OLM lists & watches those - do you?

cmd/olm/manager.go
stevekuznetsov: Is that a suggestion for a change? As of now at least to me it seems like the code in the PR does exactly that.

pkg/controller/operators/labeller/filters.go
stevekuznetsov: That function covers the one case of error, but the other and more insidious one is that you need to start & run every labeller you check for in `Validate()`

pkg/controller/operators/olm/apiservices.go
joelanford: Looking at this code, it seems like the idea was that:
- k8s comes with some predefined roles that are generally applicable to apiservices?
- instead of hardcoding the rules from those roles that we expect our serviceaccount to have, we can lookup those rules (in case they change from one k8s version to the next)
- then once we've compiled all the rules we expect our service account to have, run it through the rule checker, which doesn't actually care which roles/bindings made it happen.

And the new code is: "assume RBAC (Cluster)Roles contain the right rules (either OLM created it, or it is present in a conformant cluster); assume bindings bind the service account to the right (Cluster)Roles (OLM created it); now just look to see if the RBAC objects exist"

Am I getting that right? If so, I agree with you that the new code seems sufficient.

pkg/controller/operators/olm/operator.go
```diff
@@ -102,12 +103,57 @@ type Operator struct {
...
+	clusterRolesLister := sif.Rbac().V1().ClusterRoles().Lister()
+	clusterRoleBindingsLister := sif.Rbac().V1().ClusterRoleBindings().Lister()
+
+	done := make(chan struct{})
+	go func() {
+		<-a.ctx.Done()
+		done <- struct{}{}
+	}()
+	sif.Start(done)
+	sif.WaitForCacheSync(done)
```
joelanford, line 145: Nit: We can't just pass `a.ctx.Done()` directly into `Start` and `WaitForCacheSync`?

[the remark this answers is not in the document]
ncdc: It returns a map of sync status, but this function blocks until all caches are synced, or the stop channel has been closed.

pkg/controller/operators/olm/operator.go
joelanford: > not sure if true still but also in the past the defer approach was a couple orders of magnitude slower,

FWIW, this is no longer true. 

From Go 1.14 release notes:
> This release improves the performance of most uses of defer to incur almost zero overhead compared to calling the deferred function directly. As a result, defer can now be used in performance-critical code without overhead concerns.

pkg/controller/operators/olm/operator.go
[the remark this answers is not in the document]
joelanford: Right, so there's no difference in what we do between the two possible outcomes?
- caches synced successfully
- the stop channel was closed

cmd/olm/manager.go
joelanford: In controller-runtime, you can say "use this selector for ALL objects if there is not an object-specific selector". So I'm wondering if we can do that instead of explicitly listing the same selector for each kind.

cmd/olm/manager.go
joelanford: Good point. I don't think it does since InstallPlan is generally a one-shot sort of thing.

pkg/controller/operators/olm/operator.go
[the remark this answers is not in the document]
ncdc: Typically you're waiting on the factory to sync before parking your main goroutine via `<-ctx.Done()`. But because here we aren't in the main goroutine and this function is returning a value, I agree we should act differently if the stop channel is closed. What does that look like, though - return nil and handle the nil upstream, panic, ...?

test/e2e/operator_test.go
```diff
@@ -228,7 +246,7 @@ var _ = Describe("Operator API", func() {
 			return err
 		}).Should(Succeed())
 
-		// Ensure that o is consistently not found
+		By("Ensure that o is consistently not found")
```
joelanford, line 249: This one already has a `By`. Reduce to 1?

pkg/controller/operators/openshift/suite_test.go
```diff
@@ -108,5 +108,5 @@ var _ = BeforeSuite(func() {
 var _ = AfterSuite(func() {
 	By("tearing down the test environment")
 	close(syncCh)
-	Expect(testEnv.Stop()).To(Succeed())
```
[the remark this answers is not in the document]
stevekuznetsov: I added a commit message for it - there's a years-long thread where you (lol) have trouble reproducing it. The process exit kills the test env anyway so I honestly just wanted to get this out of the way and run tests locally.

test/e2e/operator_test.go
joelanford: Looks like a few others as well.

stevekuznetsov: If it's worth calling out in a comment, it's worth calling out in a `By()`.

pkg/controller/operators/olm/operator.go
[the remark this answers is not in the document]
stevekuznetsov: Just crash the process & restart? Done context is cancelled when we get SIGINT or similar

test/e2e/operator_test.go
joelanford: Doesn't ginkgo group code from one `By` to the next in some way? I think with two consecutive `By` lines, you're ending up with empty groupings? Not positive on this since I've never seen two consecutive `By`s though.

pkg/controller/operators/openshift/suite_test.go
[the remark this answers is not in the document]
joelanford: Lol. I have literally no recollection of that. But sgtm.

pkg/controller/operators/labeller/filters.go
joelanford: No super strong feelings here as long as we're confident we can page this knowledge in fairly easily in the future. Maybe just more well-placed comments is enough.

pkg/controller/operators/olm/operator.go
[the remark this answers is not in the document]
joelanford: I probably wouldn't `os.Exit` here in case there's some cleanup code (now or in the future) that thinks it's going to get a chance to run.

pkg/controller/operators/catalog/operator.go
joelanford: Comments really helped! Thanks.

test/e2e/operator_test.go
stevekuznetsov: Yes, but only if you append a closure to the items in `By()`. This is a no-op informative log.

cmd/olm/manager.go
stevekuznetsov: Ah - sorry. I think they may have gotten rid of that? I can't remember. Let me look.

pkg/controller/operators/olm/operator.go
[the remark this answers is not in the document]
stevekuznetsov: What should we do other than return from the function?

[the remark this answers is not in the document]
ncdc: This is only used by `func (a *Operator) permissionStatus(strategyDetailsDeployment *v1alpha1.StrategyDetailsDeployment, targetNamespace string, csv *v1alpha1.ClusterServiceVersion) (bool, []v1alpha1.RequirementStatus, error) {`, I think, so we could have that function check the return value, and if it's nil, return an error. Or this function could also return an error.

Issue #3019, opened 2023-08-25
Stop Relying on `opm` Embedded in Catalog Index Images


stevekuznetsov:
Today, when a `CatalogSource` is created, the `Pod` that serves catalog content uses `opm` to do so; the version of the server that's used is whichever version happens to be bundled into the index image. This design has a couple of draw-backs:

 - it's very difficult to reason about which versions of `opm` are used
 - updating `opm` requires rebuilding all catalogs
 - testing changes to `opm` requires the same
 - downstream consumers of the catalog cannot make many assumptions around what version of `opm` they might encounter in a catalog and what they can assume of it

When the data model was tightly coupled to the binary reading it (at the time from SQLite), bundling both the content and the server made a lot of sense. Today, though, it's possible we're not in that state any longer, as the move to FBC has given us a) a stable format and b) a non-obfuscated storage mechanism.

---

The `catalog-operator` already has an `--opmImage` flag to consume a well-known image containing `opm` to use, so implementing this feature will only require us to add a new field to the `CatalogSource` to opt catalogs into this:

```yaml
apiVersion: operators.coreos.com/v1alpha1
kind: CatalogSource
metadata:
  name: redhat-operators
  namespace: openshift-marketplace
spec:
  sourceType: grpc
  grpcPodConfig:
    extractContent:
      cacheDir: /var/cache/
      configDir: /var/config/
```

Then, the `catalog-operator` can create a `Pod` that:

 - runs an init-container with the index image, which copies the cache and config dirs into a shared EmptyDir
 - run the well-known `--opmImage` in a container, passing the two directories to it.

---

Open Questions:

- today, `opm` is deserializing the content of the catalog on disk to validate the cache; if new fields are added to FBC, an older `opm` will not be able to understand those fields and will fail to reproduce the hash - what should we do?
  - we could add an init container with the provided index image to validate the hash, and forgo the check in the main server as long as `opm` faithfully serves cache content without pruning unknown fields
  - we could not make additions to the FBC schema (are we generic enough today that new schema won't occur?)
- in general, what other concerns are there for version skew? what's our support statement for catalogs on older clusters? can we guarantee that older catalogs are built with older `opm` versions?

kevinrizza:
Five hundred feet, I think this is a reasonable change. It's more or less what we want to do anyway now that FBC is a well known format. So a +1 to the concept.

[...]

PR #3029, opened 2023-09-07
controller/registry: implement content extraction for catalog sources

stevekuznetsov:
go.mod: update the api dependency

---

controller/registry: implement content extraction for catalog sources

---

Implements https://github.com/operator-framework/operator-lifecycle-manager/issues/3019


pkg/controller/registry/reconciler/reconciler.go
```diff
@@ -236,6 +240,40 @@ func Pod(source *operatorsv1alpha1.CatalogSource, name string, img string, saNam
...
+			const catalogPath = "/catalog"
+			volumeMount := corev1.VolumeMount{
+				Name:      "catalog-content",
+				MountPath: catalogPath,
+			}
+			pod.Spec.InitContainers = append(pod.Spec.InitContainers, corev1.Container{
+				Name:    "extract-content",
+				Image:   img,
+				Command: []string{"/bin/sh", "-c"},
+				Args: []string{fmt.Sprintf("cp %s %s/catalog && cp %s %s/cache",
```
grokspawn, line 261: `cp -r`, maybe even `cp -pr`?

pkg/controller/registry/reconciler/reconciler.go
```diff
@@ -97,18 +100,19 @@ func (r *registryReconcilerFactory) ReconcilerForSource(source *operatorsv1alpha
...
 		OpClient:             opClient,
 		ConfigMapServerImage: configMapServerImage,
 		SSAClient:            ssaClient,
 		createPodAsUser:      createPodAsUser,
+		opmImage:             opmImage,
 	}
 }
 
-func Pod(source *operatorsv1alpha1.CatalogSource, name string, img string, saName string, labels map[string]string, annotations map[string]string, readinessDelay int32, livenessDelay int32, runAsUser int64) *corev1.Pod {
+func Pod(source *operatorsv1alpha1.CatalogSource, name string, opmImg, img string, saName string, labels map[string]string, annotations map[string]string, readinessDelay int32, livenessDelay int32, runAsUser int64) *corev1.Pod {
```
grokspawn, line 115: nit:
```suggestion
func Pod(source *operatorsv1alpha1.CatalogSource, name string, opmImg string, img string, saName string, labels map[string]string, annotations map[string]string, readinessDelay int32, livenessDelay int32, runAsUser int64) *corev1.Pod {
```

pkg/controller/registry/reconciler/reconciler_test.go
```diff
@@ -167,7 +167,183 @@ func TestPodMemoryTarget(t *testing.T) {
...
+							Name:         "catalog-content",
+							VolumeSource: corev1.VolumeSource{EmptyDir: &corev1.EmptyDirVolumeSource{}},
+						},
+					},
+					InitContainers: []corev1.Container{
+						{
+							Name:         "extract-content",
+							Image:        "image",
+							Command:      []string{"/bin/sh", "-c"},
+							Args:         []string{"cp /catalog /catalog/catalog && cp /tmp/cache /catalog/cache"},
```
grokspawn, line 285: How do we handle this with custom catalogs?  For e.g. RH catalogs all expose the FBC catalog in `/configs`, looking like
```tree
configs
├── abinitio-runtime-operator
│   └── catalog.json
├── acc-operator
│   └── catalog.json
...
```

which is expressed in the entrypoint+cmd of the image, e.g. 
```bash
            ...
            "Cmd": [
                "serve",
                "/configs",
                "--cache-dir=/tmp/cache"
            ],
            ...
            "Entrypoint": [
                "/bin/opm"
            ],
            ...
```

but it is very much not a requirement to use these specific paths.  If someone did something like 
`opm serve /var/spool/cats --cache-dir=/var/cache/catalog`
would it work?

pkg/controller/registry/reconciler/reconciler.go
stevekuznetsov: Why? Go specifically gives you this behavior to not repeat yourself.

pkg/controller/registry/reconciler/reconciler_test.go
stevekuznetsov: The directories are part of the config in the CatalogSource, so the user can pass in whatever they want

pkg/controller/registry/reconciler/reconciler.go
grokspawn: Hey, if you want to make the effort to stack all the other params in the signature, that's OK too.  
But leaving it un-stacked everywhere except the new param hurts readability.

pkg/controller/operators/catalog/operator_test.go
```diff
@@ -1929,7 +1929,7 @@ func toManifest(t *testing.T, obj runtime.Object) string {
 }
 
 func pod(s v1alpha1.CatalogSource) *corev1.Pod {
-	pod := reconciler.Pod(&s, "registry-server", s.Spec.Image, s.GetName(), s.GetLabels(), s.GetAnnotations(), 5, 10, 1001)
+	pod := reconciler.Pod(&s, "registry-server", "ocp-payload-opm", s.Spec.Image, s.GetName(), s.GetLabels(), s.GetAnnotations(), 5, 10, 1001)
```
grokspawn, line 1932: I thought I left this comment, but now I can't find it ... can we upstream-ify this reference so that we're not introducing downstream-ish elements into upstream?

pkg/controller/registry/reconciler/reconciler_test.go
grokspawn: Duh!  I thought this was a filesystem ref, and missed where the init container would set this as a destination for the copy (with source from catsrc config).  
So if you're just consistently `cp -r[p]` everwhere, then you're good regardless.

pkg/controller/registry/reconciler/reconciler.go
```diff
@@ -236,6 +240,40 @@ func Pod(source *operatorsv1alpha1.CatalogSource, name string, img string, saNam
...
+				MountPath: catalogPath,
+			}
+			pod.Spec.InitContainers = append(pod.Spec.InitContainers, corev1.Container{
+				Name:    "extract-content",
+				Image:   img,
+				Command: []string{"/bin/sh", "-c"},
+				Args: []string{fmt.Sprintf("cp -r %s %s/catalog && cp -r %s %s/cache",
+					grpcPodConfig.ExtractContent.CatalogDir, catalogPath,
+					grpcPodConfig.ExtractContent.CacheDir, catalogPath,
+				)},
```
joelanford, line 264: This requires the FBC catalog image to have `/bin/sh`, right?

Is there different way to do this, where we:
- initContainer: copy a `cp` binary or equivalent into an emptyDir called `copy-binary`
- initContainer(catalog image): mount `copy-binary`, run `cp` to copy catalog and cache into another emptydir called `catalog-content`
- container(opm): mount `catalog-content`, run `opm serve` pointing to mounted content dirs.

That way, we make no assumptions about the contents of the catalog image, other than the catalog and cache dirs.

joelanford: > That way, we make no assumptions about the contents of the catalog image, other than the catalog and cache dirs.

That leads me to more questions:
- Do we need to go back to the api and make `catalogDir` required?
- If `cacheDir` is unset, (FBC images don't have to have caches), do we handle that appropriately here?

stevekuznetsov: Why do you think it's more onerous to expect `sh` and `cp` than it is to expect FBC? We certainly could use some other image (which ... ?) to inject them, but it's not clear to me what niche that targets.

stevekuznetsov: > https://github.com/operator-framework/operator-lifecycle-manager/pull/3029#discussion_r1320139215

Both are required (optional is opt-in, right?)

Why should we bother to handle the case where the cache is missing? I'd rather enforce that everyone has one if they want to use this feature.

stevekuznetsov: Yikes, didn't realize my `omitempty` autocomplete had made them implciitly optional. Marking as required: https://github.com/operator-framework/api/pull/295

joelanford: > Why do you think it's more onerous to expect sh and cp

Those boil down to architecture-specific binaries/libraries being in the image, right? Which means we would have to keep building multi-arch catalog images. It would be _much_ nicer, IMO if we could have noarch `scratch`-based images that are literally just the catalogs and cache.

> Why should we bother to handle the case where the cache is missing? I'd rather enforce that everyone has one if they want to use this feature.

I'm not that fussed about it. If it starts off required, we can always come back later and make it optional if we need to.

joelanford: > We certainly could use some other image (which ... ?) 

Maybe the `opm` image or the OLM image? But that's a good question. For bundle unpacking, OLM has its own `cpb` binary that it uses. Not sure how easy it would be to copy over the standard `cp` executable and its dependencies.

stevekuznetsov: Let's start with the `opm` image and add additional configuration to the catalog operator as we see fit.

stevekuznetsov:
@joelanford made the changes you suggested. I've used this approach often in the past and found it useful, but only with statically-compiled binaries under my own control. If anything in `sh` or `cp` is dynamically linked, this won't work and we will be needing to write a Go program. Not hard, but just something to keep in mind.

test/e2e/e2e_test.go
```diff
@@ -38,7 +38,7 @@ var (
 
 	communityOperators = flag.String(
 		"communityOperators",
-		"quay.io/operator-framework/upstream-community-operators@sha256:098457dc5e0b6ca9599bd0e7a67809f8eca397907ca4d93597380511db478fec",
+		"quay.io/operatorhubio/catalog:latest",
```
stevekuznetsov, line 41: @joelanford unfortunately the specific SHAs for this image are not kept around for very long, as there are no long-lived tags. We will have to suffice with `:latest`, which might break tests if they are asserting on specifics.

joelanford: Seems fine for now. It'd be nice if we did the thing we are now doing in [catalogd](https://github.com/operator-framework/catalogd/blob/a1663ec2b64f560330ce57f78a8018357d400656/Makefile#L150-L152) and [operator-controller](https://github.com/operator-framework/operator-controller/blob/5cbc9a0d5e3bbeab3ce200a9fd1da66453f0247b/Makefile#L154-L165) where we have all of our test artifacts actually come out of the repo.

Definitely not a blocker for this though, IMO.

pkg/controller/registry/reconciler/grpc.go
```diff
@@ -196,16 +198,25 @@ func (c *GrpcRegistryReconciler) currentPodsWithCorrectImageAndSpec(source grpcC
 	found := []*corev1.Pod{}
 	newPod := source.Pod(saName)
 	for _, p := range pods {
-		if p.Spec.Containers[0].Image == source.Spec.Image && podHashMatch(p, newPod) {
+		if correctImages(source, p) && podHashMatch(p, newPod) {
```
joelanford, line 201: I know you're just updating based on what's already here, but I do wonder if `podHashMatch` alone would be sufficient here. Wouldn't that cover the image references in the pod spec?

stevekuznetsov: I don't have enough knowledge of the system to understand why that was not the case before, so I did not want to change it. Perhaps a follow-up? Knowing this code-base it will break some obscure edge cases and be hard to track down :)

PR #3033, opened 2023-09-14
remove dangling unused constant

perdasilva:
Found this dangling unused constant while working on a different PR. This one just removes it.


tmshort (approved):
/lgtm

PR #3035, opened 2023-09-18
Rename ClusterRoles created by OperatorGroups

tmshort:
When an OperatorGroup creates a ClusterRole, it's based directly on the OG name with a suffix, this causes two issues:
1. same-named OGs in different namespaces overwrite each others CRs
2. there are some very important CRs that could be overwritten by OG

Tests added.

The ClusterRoles created by an OperatorGroup can conflict with existing CRs and with CRs created by OGs.

**Architectural changes:**

Naming of these OG CRs change, but since they're aggregated, admins shouldn't really be using them, so the new names shouldn't matter.

There will be a need to clean up the old roles, but that could potentially be dangerous.

**Testing remarks:**

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/olm/operator_test.go
+++ b/pkg/controller/operators/olm/operator_test.go
@@ -4560,8 +4560,7 @@ func TestSyncOperatorGroups(t *testing.T) {
 				"": {
 					&rbacv1.ClusterRole{
 						ObjectMeta: metav1.ObjectMeta{
-							ResourceVersion: "",
-							Name:            "olm.og.operator-group-1.admin-8rdAjL0E35JMMAkOqYmoorzjpIIihfnj3DcgDU",
+							Name: "olm.og.operator-group-1.admin-8rdAjL0E35JMMAkOqYmoorzjpIIihfnj3DcgDU",
 							Labels: map[string]string{
 								"olm.owner":           "operator-group-1",
 								"olm.owner.namespace": "operator-ns",
@@ -4571,8 +4570,7 @@ func TestSyncOperatorGroups(t *testing.T) {
 					},
 					&rbacv1.ClusterRole{
 						ObjectMeta: metav1.ObjectMeta{
-							ResourceVersion: "",
-							Name:            "olm.og.operator-group-1.edit-9lBEUxqAYE7CX7wZfFEPYutTfQTo43WarB08od",
+							Name: "olm.og.operator-group-1.edit-9lBEUxqAYE7CX7wZfFEPYutTfQTo43WarB08od",
 							Labels: map[string]string{
 								"olm.owner":           "operator-group-1",
 								"olm.owner.namespace": "operator-ns",
@@ -4582,8 +4580,7 @@ func TestSyncOperatorGroups(t *testing.T) {
 					},
 					&rbacv1.ClusterRole{
 						ObjectMeta: metav1.ObjectMeta{
-							ResourceVersion: "",
-							Name:            "olm.og.operator-group-1.view-1l6ymczPK5SceF4d0DCtAnWZuvmKn6s8oBUxHr",
+							Name: "olm.og.operator-group-1.view-1l6ymczPK5SceF4d0DCtAnWZuvmKn6s8oBUxHr",
 							Labels: map[string]string{
 								"olm.owner":           "operator-group-1",
 								"olm.owner.namespace": "operator-ns",
@@ -4624,8 +4621,7 @@ func TestSyncOperatorGroups(t *testing.T) {
 					},
 					&rbacv1.ClusterRole{
 						ObjectMeta: metav1.ObjectMeta{
-							ResourceVersion: "",
-							Name:            "operator-group-1-admin",
+							Name: "operator-group-1-admin",
 							Labels: map[string]string{
 								"olm.owner":           "operator-group-1",
 								"olm.owner.namespace": "operator-ns",
@@ -4635,8 +4631,7 @@ func TestSyncOperatorGroups(t *testing.T) {
 					},
 					&rbacv1.ClusterRole{
 						ObjectMeta: metav1.ObjectMeta{
-							ResourceVersion: "",
-							Name:            "operator-group-1-view",
+							Name: "operator-group-1-view",
 							Labels: map[string]string{
 								"olm.owner":           "operator-group-1",
 								"olm.owner.namespace": "operator-ns",
@@ -4646,8 +4641,7 @@ func TestSyncOperatorGroups(t *testing.T) {
 					},
 					&rbacv1.ClusterRole{
 						ObjectMeta: metav1.ObjectMeta{
-							ResourceVersion: "",
-							Name:            "operator-group-1-edit",
+							Name: "operator-group-1-edit",
 							Labels: map[string]string{
 								"olm.owner":           "operator-group-1",
 								"olm.owner.namespace": "operator-ns",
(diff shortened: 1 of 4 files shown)


pkg/controller/operators/olm/operator_test.go
```diff
@@ -4521,6 +4523,431 @@ func TestSyncOperatorGroups(t *testing.T) {
...
+			},
+			expectedStatus: operatorsv1.OperatorGroupStatus{
+				Namespaces:  []string{targetNamespace},
+				LastUpdated: &now,
+			},
+			final: final{objects: map[string][]runtime.Object{
+				"": {
+					&rbacv1.ClusterRole{
+						ObjectMeta: metav1.ObjectMeta{
+							ResourceVersion: "",
```
stevekuznetsov, line 4563: nit: can remove these empty fields

pkg/controller/operators/olm/operator_test.go
```diff
@@ -5366,7 +5793,10 @@ func RequireObjectsInCache(t *testing.T, lister operatorlister.OperatorLister, n
 			require.Failf(t, "couldn't find expected object", "%#v", object)
 		}
 		if err != nil {
-			return fmt.Errorf("namespace: %v, error: %v", namespace, err)
+			if apierrors.IsNotFound(err) {
```
stevekuznetsov, line 5796: What's the meaning of this change?

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -981,59 +996,88 @@ func (a *Operator) updateNamespaceList(op *operatorsv1.OperatorGroup) ([]string,
 	return namespaceList, nil
 }
 
-func (a *Operator) ensureOpGroupClusterRole(op *operatorsv1.OperatorGroup, suffix string, apis cache.APISet) error {
-	clusterRole := &rbacv1.ClusterRole{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:   strings.Join([]string{op.GetName(), suffix}, "-"),
-			Labels: map[string]string{install.OLMManagedLabelKey: install.OLMManagedLabelValue},
-		},
+func (a *Operator) getClusterRoleName(op *operatorsv1.OperatorGroup, roleType string) (string, error) {
```
stevekuznetsov, line 999: note: in the degenerate case where there's 0 characters available for the `op.GetName()` the role name does not collide with others since roleType is in the hash

pkg/controller/operators/olm/operator_test.go
tmshort: It was in @perdasilva's change that I pulled in directly. Line numbers are probably a bit different due to reverting format-only changes.
Looks as though it's trying to avoid double-wrapping the error.

pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -981,59 +996,88 @@ func (a *Operator) updateNamespaceList(op *operatorsv1.OperatorGroup) ([]string,
...
 
-	if _, err := a.opClient.UpdateClusterRole(clusterRole); err != nil {
-		a.logger.WithError(err).Errorf("Update existing cluster role failed: %v", clusterRole)
+	if err := ownerutil.AddOwnerLabels(clusterRole, op); err != nil {
 		return err
 	}
-	return nil
+
+	a.logger.Infof("creating cluster role: %s owned by operator group: %s/%s", clusterRole.GetName(), op.GetNamespace(), op.GetName())
+	_, err = a.opClient.KubernetesInterface().RbacV1().ClusterRoles().Create(context.TODO(), clusterRole, metav1.CreateOptions{})
```
stevekuznetsov, line 1058: Is it hard to plumb a context in here for the call? If so, ignore comment.

pkg/lib/ownerutil/util.go
```diff
@@ -271,6 +271,11 @@ func CSVOwnerSelector(owner *operatorsv1alpha1.ClusterServiceVersion) labels.Sel
 	return labels.SelectorFromSet(OwnerLabel(owner, operatorsv1alpha1.ClusterServiceVersionKind))
 }
 
+// OperatorGroupOwnerSelector returns a label selector to find generated objects owned by owner
+func OperatorGroupOwnerSelector(owner *operatorsv1.OperatorGroup) labels.Selector {
+	return labels.SelectorFromSet(OwnerLabel(owner, "OperatorGroup"))
```
stevekuznetsov, line 276: nit: there is an `operatorsv1.OperatorGroupKind` constant

pkg/controller/operators/olm/operatorgroup.go
tmshort: I can certainly look.

test/e2e/operator_groups_e2e_test.go
```diff
@@ -340,27 +340,31 @@ var _ = Describe("Operator Group", func() {
...
-		editRole, err := c.KubernetesInterface().RbacV1().ClusterRoles().Get(context.TODO(), operatorGroup.Name+"-edit", metav1.GetOptions{})
+		existingClusterRoleList, err := c.KubernetesInterface().RbacV1().ClusterRoles().List(context.TODO(), metav1.ListOptions{
+			LabelSelector: labels.SelectorFromSet(ownerutil.OwnerLabel(&operatorGroup, "OperatorGroup")).String(),
+		})
 		require.NoError(GinkgoT(), err)
-		editPolicyRules := []rbacv1.PolicyRule{
-			{Verbs: []string{"create", "update", "patch", "delete"}, APIGroups: []string{mainCRD.Spec.Group}, Resources: []string{mainCRDPlural}},
-		}
-		require.Equal(GinkgoT(), editPolicyRules, editRole.Rules)
+		require.Len(GinkgoT(), existingClusterRoleList.Items, 3)
```
stevekuznetsov, line 347: FWIW try making this fail, the error output is pretty bad. If this is not 3, what went wrong? Can we encode that in the failure message for the dev that breaks this next? What do they need to go fix? It's not clear what is creating the objects - is it test code?

test/e2e/operator_groups_e2e_test.go
```diff
@@ -340,27 +340,31 @@ var _ = Describe("Operator Group", func() {
...
-		require.NoError(GinkgoT(), err)
-		viewPolicyRules := []rbacv1.PolicyRule{
-			{Verbs: []string{"get"}, APIGroups: []string{"apiextensions.k8s.io"}, Resources: []string{"customresourcedefinitions"}, ResourceNames: []string{mainCRD.Name}},
-			{Verbs: []string{"get", "list", "watch"}, APIGroups: []string{mainCRD.Spec.Group}, Resources: []string{mainCRDPlural}},
+		for _, role := range existingClusterRoleList.Items {
+			if strings.HasSuffix(role.Name, "admin") {
+				adminPolicyRules := []rbacv1.PolicyRule{
+					{Verbs: []string{"*"}, APIGroups: []string{mainCRD.Spec.Group}, Resources: []string{mainCRDPlural}},
+				}
+				require.Equal(GinkgoT(), adminPolicyRules, role.Rules)
```
stevekuznetsov, line 354: if possible, do something like

```go
if diff := cmp.Diff(adminPolicyRules, role.Rules); diff != "" {
    fmt.Println(diff) // Ginkgo mangles this output horribly ...
    GinkgoT().Fail("invalid rules on admin role")
}
```

The next person to break this will thank you

pkg/controller/operators/olm/operatorgroup.go
stevekuznetsov: Unit tests would help!

stevekuznetsov:
Content looks good, just nits.

/lgtm

pkg/controller/operators/olm/operatorgroup.go
tmshort: But... mathematically impossible for `nameLimit` to be `<0`. The only thing not a constant value is the roleType.

pkg/controller/operators/olm/operatorgroup.go
tmshort: The whole `olm` package uses `context.TODO()` everywhere (with a rare `context.Background()`), the test files do use `WithCancel(context.TODO())` in some spots, but I'm not seeing it in non-test code. Plumbling in a proper context, either from the user of this package, or inside the "root" of this package is definitely out-of-scope for this.

test/e2e/operator_groups_e2e_test.go
tmshort: if `assert` were imported, this would be a smidgen easier...

PR #3042, opened 2023-09-21
Use generated namespaces in e2e tests

awgreene:
This commit updates the OLM e2e test framework to create the test namespace if it does not already exist.

### DIFF
--- a/test/e2e/csv_e2e_test.go
+++ b/test/e2e/csv_e2e_test.go
@@ -38,9 +38,9 @@ import (
 
 var _ = Describe("ClusterServiceVersion", func() {
 	var (
-		ns  corev1.Namespace
-		c   operatorclient.ClientInterface
-		crc versioned.Interface
+		generatedNamespace corev1.Namespace
+		c                  operatorclient.ClientInterface
+		crc                versioned.Interface
 	)
 
 	BeforeEach(func() {
@@ -49,14 +49,14 @@ var _ = Describe("ClusterServiceVersion", func() {
 	})
 
 	AfterEach(func() {
-		TeardownNamespace(ns.GetName())
+		TeardownNamespace(generatedNamespace.GetName())
 	})
 
 	Context("OwnNamespace OperatorGroup", func() {
 
 		BeforeEach(func() {
 			nsName := genName("csv-e2e-")
-			ns = SetupGeneratedTestNamespace(nsName, nsName)
+			generatedNamespace = SetupGeneratedTestNamespace(nsName, nsName)
 		})
 
 		When("a CustomResourceDefinition was installed alongside a ClusterServiceVersion", func() {
@@ -73,7 +73,7 @@ var _ = Describe("ClusterServiceVersion", func() {
 					ObjectMeta: metav1.ObjectMeta{
 						Name: apifullname,
 						Annotations: map[string]string{
-							"operatorframework.io/installed-alongside-0": fmt.Sprintf("%s/associated-csv", ns.GetName()),
+							"operatorframework.io/installed-alongside-0": fmt.Sprintf("%s/associated-csv", generatedNamespace.GetName()),
 						},
 					},
 					Spec: apiextensionsv1.CustomResourceDefinitionSpec{
@@ -112,7 +112,7 @@ var _ = Describe("ClusterServiceVersion", func() {
 				associated := operatorsv1alpha1.ClusterServiceVersion{
 					ObjectMeta: metav1.ObjectMeta{
 						Name:      "associated-csv",
-						Namespace: ns.GetName(),
+						Namespace: generatedNamespace.GetName(),
 					},
 					Spec: operatorsv1alpha1.ClusterServiceVersionSpec{
 						CustomResourceDefinitions: operatorsv1alpha1.CustomResourceDefinitions{
@@ -170,7 +170,7 @@ var _ = Describe("ClusterServiceVersion", func() {
 				unassociated := operatorsv1alpha1.ClusterServiceVersion{
 					ObjectMeta: metav1.ObjectMeta{
 						Name:      "unassociated-csv",
-						Namespace: ns.GetName(),
+						Namespace: generatedNamespace.GetName(),
 					},
 					Spec: operatorsv1alpha1.ClusterServiceVersionSpec{
 						CustomResourceDefinitions: operatorsv1alpha1.CustomResourceDefinitions{
@@ -194,7 +194,7 @@ var _ = Describe("ClusterServiceVersion", func() {
 				associated := operatorsv1alpha1.ClusterServiceVersion{
 					ObjectMeta: metav1.ObjectMeta{
 						Name:      "associated-csv",
-						Namespace: ns.GetName(),
+						Namespace: generatedNamespace.GetName(),
 					},
 					Spec: operatorsv1alpha1.ClusterServiceVersionSpec{
 						CustomResourceDefinitions: operatorsv1alpha1.CustomResourceDefinitions{
@@ -229,7 +229,7 @@ var _ = Describe("ClusterServiceVersion", func() {
 				unassociated := operatorsv1alpha1.ClusterServiceVersion{
 					ObjectMeta: metav1.ObjectMeta{
 						Name:      "unassociated-csv",
-						Namespace: ns.GetName(),
+						Namespace: generatedNamespace.GetName(),
 					},
 					Spec: operatorsv1alpha1.ClusterServiceVersionSpec{
 						CustomResourceDefinitions: operatorsv1alpha1.CustomResourceDefinitions{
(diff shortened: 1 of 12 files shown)


Makefile
```diff
@@ -139,7 +139,7 @@ endif
 E2E_OPTS ?= $(if $(E2E_SEED),-seed '$(E2E_SEED)') $(if $(SKIP), -skip '$(SKIP)') $(if $(TEST),-focus '$(TEST)') $(if $(ARTIFACT_DIR), -output-dir $(ARTIFACT_DIR) -junit-report junit_e2e.xml) -flake-attempts $(E2E_FLAKE_ATTEMPTS) -nodes $(E2E_NODES) -timeout $(E2E_TIMEOUT) -v -randomize-suites -race -trace -progress
 E2E_INSTALL_NS ?= operator-lifecycle-manager
 E2E_CATALOG_NS ?= $(E2E_INSTALL_NS)
-E2E_TEST_NS ?= operators
+E2E_TEST_NS ?= nonexistent-namespace
```
awgreene, line 142: This isn't technically necessary, but it allows us to ensure that the e2e tests work when a namespace is specified that doesn't exist.

test/e2e/e2e_test.go
```diff
@@ -91,13 +93,32 @@ var _ = BeforeSuite(func() {
...
+		return nil
+	}).Should(Succeed())
+
 	var groups operatorsv1.OperatorGroupList
 	Expect(ctx.Ctx().Client().List(context.Background(), &groups, client.InNamespace(testNamespace))).To(Succeed())
 	if len(groups.Items) == 0 {
 		og := operatorsv1.OperatorGroup{
 			ObjectMeta: metav1.ObjectMeta{
-				Name:      "opgroup",
+				Name:      "global-operators",
```
awgreene, line 117: Some of the tests expect the default `operatorGroup` in the `testNamespace`, so I've changed the name here. Alternatively, I could write the tests to use any `operatorGroup` in the namespace, but there are some assumptions about the default scope of that `operatorGroup`.

Makefile
ncdc: In kcp, we generated unique names for our outermost container in all tests (which was a kcp workspace, but the equivalent here would be a namespace). How would you feel about every standalone test use its own dedicated unique namespace name?

test/e2e/util/e2e_client.go
```diff
@@ -59,6 +59,11 @@ func (m *E2EKubeClient) Reset() error {
 			break
 		}
 
+		// Allow resources to opt-out of cleanup at the end of the test.
```
awgreene, line 62:
```suggestion
		// Allow resources to opt-out of being deleted by the e2e client.
```

Makefile
awgreene: I'm not opposed to that option, would we want to deprecate the ability to specify a testNamespace or update that feature to override those generated namespaces? Would we rather do this now or in a followup?

ncdc: Followup is fine if you don't want to do it now, but I don't see any utility in supporting specifying a test namespace name.

stevekuznetsov: The only reason it would be nice to allow specifying it is if you think that someone who is working on the test could hot-loop the test process and bypass whatever setup is needed by providing an existing namespace.

test/e2e/util/e2e_client.go
stevekuznetsov: In my recent work on e2es I've added `$SKIP_CLEANUP` checks to stop cleaning up - would that approach work here?

test/e2e/csv_e2e_test.go
```diff
@@ -401,6 +401,10 @@ var _ = Describe("ClusterServiceVersion", func() {
 	})
 
 	Context("AllNamespaces OperatorGroup", func() {
+		BeforeEach(func() {
+			nsName := genName("csv-e2e-")
+			generatedNamespace = SetupGeneratedTestNamespace(nsName)
```
stevekuznetsov, line 406: Why is this valid? Were the tests reusing one namespace in the past?

test/e2e/installplan_e2e_test.go
```diff
@@ -3320,15 +3320,6 @@ var _ = Describe("Install Plan", func() {
 	It("compresses installplan step resource manifests to configmap references", func() {
 		// Test ensures that all steps for index-based catalogs are references to configmaps. This avoids the problem
 		// of installplans growing beyond the etcd size limit when manifests are written to the ip status.
-
-		og := &operatorsv1.OperatorGroup{}
```
stevekuznetsov, line 3331: It's not clear why this block was removed?

test/e2e/installplan_e2e_test.go
```diff
@@ -3716,24 +3691,22 @@ var _ = Describe("Install Plan", func() {
...
-				Name:      genName("og-"),
-				Namespace: generatedNamespace.GetName(),
-			},
-			Spec: operatorsv1.OperatorGroupSpec{
-				ServiceAccountName: sa.GetName(),
-			},
-		}
-		_, err = crc.OperatorsV1().OperatorGroups(generatedNamespace.GetName()).Create(context.Background(), og, metav1.CreateOptions{})
-		Expect(err).ToNot(HaveOccurred())
+		// Update the existing OG to use the ServiceAccount
```
stevekuznetsov, line 3694: (just reviewing the last commit) really not clear where the existing OG comes from - why is it shared? Can it be test-local?

test/e2e/packagemanifest_e2e_test.go
```diff
@@ -36,9 +36,7 @@ var _ = Describe("Package Manifest API lists available Operators from Catalog So
 		pmc = newPMClient()
 		c = ctx.Ctx().KubeClient()
 
-		BeforeEach(func() {
-			generatedNamespace = SetupGeneratedTestNamespace(genName("package-manifest-e2e"))
-		})
+		generatedNamespace = SetupGeneratedTestNamespace(genName("package-manifest-e2e"))
```
stevekuznetsov, line 39: Why do we want one ns for all of them vs one per test?

awgreene: Hmm I might be missing what you're trying to point out, could you describe what you mean by "all of them" vs "one per test"? Running the tests in this describe block shows that multiple namespaces are generated, specifically one for each test.

```
$ k get ns -w 
...
...
package-manifest-e2ect2wt   Active   0s
package-manifest-e2ehsr64   Active   0s
package-manifest-e2erptpp   Active   0s
package-manifest-e2edgq5m   Active   0s
```

Makefile
awgreene: I like Steve's proposal, but I'm concerned about scope creep for the PR. How would we feel about focusing this PR on using generated namespaces for e2e tests. To do so we would:
1. Drop the first commit (which includes this change) from the PR
2. Create an issue to update how e2e tests consume the testNamespace parameter.

test/e2e/util/e2e_client.go
awgreene: Good idea! If we decide to split this out as proposed [here](https://github.com/operator-framework/operator-lifecycle-manager/pull/3042#discussion_r1337494471) I'll share this idea in the ticket.

test/e2e/csv_e2e_test.go
awgreene: This is actually a no-op change, so I'm going to revert it with the exception of changing  `ns =` to `generatedNamespace =`

test/e2e/installplan_e2e_test.go
awgreene: I removed this as [this code](https://github.com/operator-framework/operator-lifecycle-manager/pull/3042/files#diff-eddfb8217b77c532b776defca62eb7d6e4710bb15ceb8bf8a4d8de408df3a6a4R72) will create a fresh namespace with an operator group in it.

test/e2e/installplan_e2e_test.go
awgreene: [This line](https://github.com/operator-framework/operator-lifecycle-manager/pull/3042/files#diff-eddfb8217b77c532b776defca62eb7d6e4710bb15ceb8bf8a4d8de408df3a6a4R72) is ran before each it, which creates the namespace and operatorGroup, so we don't need to create new resources here.

test/e2e/packagemanifest_e2e_test.go
awgreene: Silly Pseudo Code used to explain what is happening:
```
Describe("e2e suite")...{
   BeforeEach{
      fmt.Printf("Before\n)
   }
   AfterEach{
      fmt.Printf("After\n)
   }
   It(1)...{
      fmt.Printf("Test 1\n)
   }
   It(2)...{
      fmt.Printf("Test 2\n)
   }
}
```

If you run the tests above, you'd see:
1. Before
2. Test1
3. After
4. Before
5. Test2
6. After

Makefile
awgreene: Removed the first commit.

awgreene:
>[FAIL] Fail Forward Upgrades when a CSV resource is in a failed state [It] eventually reports a successful state when using replaces

Worked locally, retesting

awgreene:
>[FAIL] Fail Forward Upgrades when a CSV resource is in a failed state [It] eventually reports a successful state when using skip ranges

New failure that didn't happen locally, retesting.

stevekuznetsov:
@awgreene both of those are the known flakes that we tracked down, can sync on Slack for them but just re-test your heart away for now

PR #3043, opened 2023-09-28
CRD Compatibility Check Fix

### LINKED ISSUE #3032: CRD compatibility check logic is incorrect
The logic to validate CRD compatibility is incorrect. It looks like it used to be ok with v1beta1 CRDs, but with the move to v1, it doesn't handle multi-version schemas correctly.

This code does the following, attempting to short-circuit and return early to avoid listing all CRs if possible:

1. Get the last version in the updated CRD. There is no guarantee this is the "latest" or "newest" version, however. I don't believe Kubernetes sorts this.
2. If and only if all of the existing CRD's version schemas are identical to the last version in the updated CRD, return early (this is unlikely to ever be true)

https://github.com/operator-framework/operator-lifecycle-manager/blob/5c9ae5a284da563051d4df977190f9a8b3e257f9/pkg/controller/operators/catalog/operator.go#L1993-L2004

This code loops through all the existing CRD's versions, and for each one that is served, tries to validate all the existing CRs. But that logic

https://github.com/operator-framework/operator-lifecycle-manager/blob/5c9ae5a284da563051d4df977190f9a8b3e257f9/pkg/controller/operators/catalog/operator.go#L2054-L2071

is incorrect. Specifically, this line

https://github.com/operator-framework/operator-lifecycle-manager/blob/5c9ae5a284da563051d4df977190f9a8b3e257f9/pkg/controller/operators/catalog/operator.go#L2061

Most likely always yields a validator that always returns success / no error, because in most if not all v1 CRDs, `spec.validation` is nil. And `validation.NewSchemaValidator` returns a no-op validator in that case.

This is what I think the correct behavior should be:

1. Loop through each of the **updated** CRD's versions. For each that is served:
2. Construct a validator for that specific version's schema
3. List all the CRs at that version
4. Attempt to validate each one

dtfranz:

Updates the validation code for CRD updates to validate all CRs present on the cluster against the schema of **each** served version in the new CRD. 

Previously this code checked **only** the `newCRD.Spec.Validation` field for new schema, which when using the `apiextensions.k8s.io/v1` API should not be used.

**Testing remarks:**

We did not have any tests using `apiextensions.k8s.io/v1` CRDs previously, which is likely how this became an oversight; I've added a small suite of tests to check functionality using the up-to-date CRD API.

Closes #3032

### DIFF
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -32,6 +32,7 @@ import (
 	"k8s.io/apimachinery/pkg/selection"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	"k8s.io/apimachinery/pkg/util/yaml"
 	batchv1applyconfigurations "k8s.io/client-go/applyconfigurations/batch/v1"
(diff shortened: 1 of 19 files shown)


pkg/controller/operators/catalog/operator.go
```diff
@@ -2013,59 +2013,77 @@ func transitionInstallPlanState(log logrus.FieldLogger, transitioner installPlan
...
+			versionMap[version.Name] = struct{}{}
 		}
 	}
 
-	logrus.Debugf("Successfully validated CRD %s\n", newCRD.Name)
+	for _, schemaVersion := range convertedCRD.Spec.Versions {
+		if !schemaVersion.Served {
+			// skip unserved versions
+			continue
+		} else if _, ok := versionMap[schemaVersion.Name]; !ok {
```
ncdc, line 2033: nit: don't need `else`

pkg/controller/operators/catalog/operator.go
```diff
@@ -2013,59 +2013,77 @@ func transitionInstallPlanState(log logrus.FieldLogger, transitioner installPlan
...
 	}
 
-	if oldCRD.Spec.Version != "" {
-		gvr := schema.GroupVersionResource{Group: oldCRD.Spec.Group, Version: oldCRD.Spec.Version, Resource: oldCRD.Spec.Names.Plural}
-		err := validateExistingCRs(dynamicClient, gvr, convertedCRD)
+	for _, version := range convertedCRD.Spec.Versions {
+		if !version.Served {
+			// skip unserved versions
+			continue
+		} else if _, ok := versionMap[version.Name]; !ok {
```
ncdc, line 2073: nit: don't need else

pkg/controller/operators/catalog/operator.go
```diff
@@ -2013,59 +2013,77 @@ func transitionInstallPlanState(log logrus.FieldLogger, transitioner installPlan
...
+			// cannot validate versions which do not already exist in the cluster
+			continue
+		}
+		gvr := schema.GroupVersionResource{Group: convertedCRD.Spec.Group, Version: schemaVersion.Name, Resource: convertedCRD.Spec.Names.Plural}
+		schema := schemaVersion.Schema
+		if schema == nil {
+			// If the per-version schema is nil, use the global schema
+			schema = convertedCRD.Spec.Validation
+		}
+		err := validateExistingCRs(dynamicClient, gvr, schema)
```
ncdc, line 2043: if err := ...; err != nil

pkg/controller/operators/catalog/operator.go
```diff
@@ -2013,59 +2013,77 @@ func transitionInstallPlanState(log logrus.FieldLogger, transitioner installPlan
...
+	for _, version := range convertedCRD.Spec.Versions {
+		if !version.Served {
+			// skip unserved versions
+			continue
+		} else if _, ok := versionMap[version.Name]; !ok {
+			// cannot validate versions which do not already exist in the cluster
+			continue
+		}
+		gvr := schema.GroupVersionResource{Group: convertedCRD.Spec.Group, Version: version.Name, Resource: convertedCRD.Spec.Names.Plural}
+		err := validateExistingCRs(dynamicClient, gvr, convertedCRD.Spec.Validation)
```
ncdc, line 2078: if err := ...; err != nil

pkg/controller/operators/catalog/operator.go
```diff
@@ -2013,59 +2013,77 @@ func transitionInstallPlanState(log logrus.FieldLogger, transitioner installPlan
...
+		if err != nil {
+			return err
+		}
+	}
+
+	if convertedCRD.Spec.Version != "" {
+		gvr := schema.GroupVersionResource{Group: convertedCRD.Spec.Group, Version: convertedCRD.Spec.Version, Resource: convertedCRD.Spec.Names.Plural}
+		err := validateExistingCRs(dynamicClient, gvr, convertedCRD.Spec.Validation)
```
ncdc, line 2086: if err := ...; err != nil

pkg/controller/operators/catalog/operator.go
```diff
@@ -2074,20 +2092,24 @@ func validateV1Beta1CRDCompatibility(dynamicClient dynamic.Interface, oldCRD *ap
...
+func validateExistingCRs(dynamicClient dynamic.Interface, gvr schema.GroupVersionResource, crdValidation *apiextensions.CustomResourceValidation) error {
 	crList, err := dynamicClient.Resource(gvr).List(context.TODO(), metav1.ListOptions{})
 	if err != nil {
 		return fmt.Errorf("error listing resources in GroupVersionResource %#v: %s", gvr, err)
 	}
+
+	// create validator from given crdValidation
+	validator, _, err := validation.NewSchemaValidator(crdValidation)
+	if err != nil {
+		return fmt.Errorf("error creating validator for schema %#v: %s", crdValidation, err)
```
ncdc, line 2105: Not sure we want to dump the full schema in the error message

pkg/controller/operators/catalog/operator.go
```diff
@@ -2074,20 +2092,24 @@ func validateV1Beta1CRDCompatibility(dynamicClient dynamic.Interface, oldCRD *ap
...
+	// validate each CR against this version schema
 	for _, cr := range crList.Items {
-		validator, _, err := validation.NewSchemaValidator(newCRD.Spec.Validation)
-		if err != nil {
-			return fmt.Errorf("error creating validator for schema %#v: %s", newCRD.Spec.Validation, err)
-		}
 		err = validation.ValidateCustomResource(field.NewPath(""), cr.UnstructuredContent(), validator).ToAggregate()
 		if err != nil {
-			return fmt.Errorf("error validating custom resource against new schema for %s %s/%s: %v", newCRD.Spec.Names.Kind, cr.GetNamespace(), cr.GetName(), err)
+			return fmt.Errorf("error validating custom resource against new schema for version %s of %s %s/%s: %v", cr.GetAPIVersion(), cr.GroupVersionKind().Kind, cr.GetNamespace(), cr.GetName(), err)
```
ncdc, line 2112: If it's cluster-scoped, namespace will be empty. Let's format the fully qualified name either as %s or %s/%s based on cluster vs. namespace scoped.

dtfranz:
@ncdc Updated per your comments, thanks for reviewing! Let me know if you see anything else.

pkg/controller/operators/catalog/operator.go
```diff
@@ -2013,81 +2013,108 @@ func transitionInstallPlanState(log logrus.FieldLogger, transitioner installPlan
...
 		err = validation.ValidateCustomResource(field.NewPath(""), cr.UnstructuredContent(), validator).ToAggregate()
 		if err != nil {
-			return fmt.Errorf("error validating custom resource against new schema for %s %s/%s: %v", newCRD.Spec.Names.Kind, cr.GetNamespace(), cr.GetName(), err)
+			var namespacedName string
+			if cr.GetNamespace() == "" {
+				namespacedName = cr.GetName()
+			} else {
+				namespacedName = fmt.Sprintf("%s/%s", cr.GetNamespace(), cr.GetName())
+			}
+			return fmt.Errorf("error validating custom resource against new schema for version %s of %s %s: %v", cr.GetAPIVersion(), cr.GroupVersionKind().Kind, namespacedName, err)
```
ncdc, line 2117:
```suggestion
			return fmt.Errorf("error validating %s %q: updated validation is too restrictive: %v ", cr.GroupVersionKind(), namespacedName, err)
```

Looking for something like this:
```
error validating $group/$version, Kind=$kind "$ns/$name": updated validation is too restrictive: $err
```

pkg/controller/operators/catalog/operator.go
```diff
@@ -2013,81 +2013,108 @@ func transitionInstallPlanState(log logrus.FieldLogger, transitioner installPlan
...
+		if !version.Served {
+			// skip unserved versions
+			continue
+		}
+		if _, ok := versionMap[version.Name]; !ok {
+			// cannot validate versions which do not already exist in the cluster
+			continue
+		}
+		gvr := schema.GroupVersionResource{Group: convertedCRD.Spec.Group, Version: version.Name, Resource: convertedCRD.Spec.Names.Plural}
+		if err := validateExistingCRs(dynamicClient, gvr, convertedCRD.Spec.Validation); err != nil {
```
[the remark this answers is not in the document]
joelanford: It is possible to extract the common logic into a helper function that is used by both the `v1` and` `v1beta1` variants?

ncdc:
I don't think there is anything we can really do. This is "best effort".

pkg/controller/operators/catalog/operator.go
```diff
@@ -2013,81 +2013,108 @@ func transitionInstallPlanState(log logrus.FieldLogger, transitioner installPlan
...
+			return err
+		}
+	}
+
+	if convertedCRD.Spec.Version != "" {
+		gvr := schema.GroupVersionResource{Group: convertedCRD.Spec.Group, Version: convertedCRD.Spec.Version, Resource: convertedCRD.Spec.Names.Plural}
+		if err := validateExistingCRs(dynamicClient, gvr, convertedCRD.Spec.Validation); err != nil {
 			return err
 		}
 	}
```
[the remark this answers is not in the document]
dtfranz: Ah, understood. For some reason I was under the impression that `spec.version` and `spec.versions` were mutually exclusive; I'll correct this.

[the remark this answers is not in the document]
dtfranz: For `apiextensionsv1` CRDs, the `spec.version` field does not exist and we can ignore this special case. For `apiextensionsv1beta1`, either or both can be used, but the schema within `spec.versions` or `spec.validation` are mutually exclusive. I've updated the conversion function to handle these cases and added an additional unit test for the case where `spec.versions` is unused.

pkg/controller/operators/catalog/operator.go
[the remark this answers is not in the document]
dtfranz: See my comment [here](https://github.com/operator-framework/operator-lifecycle-manager/pull/3043#discussion_r1346891592); the logic is different between the two versions, so we can simplify things when dealing with `v1`.

dtfranz:
@ncdc @joelanford Thank you both again for looking through this, I think I've managed to take care of all the special v1beta1 crd cases but I would appreciate another look. 
Meanwhile the `flaky-e2e-tests` are failing which I cannot reproduce locally. I'm looking into that but I don't think it's necessarily a cause for concern (same failures as a different PR run [here](https://github.com/operator-framework/operator-lifecycle-manager/actions/runs/6413568062/job/17412684244))

pkg/controller/operators/catalog/operator.go
```diff
@@ -2013,81 +2013,101 @@ func transitionInstallPlanState(log logrus.FieldLogger, transitioner installPlan
...
 
 // Validate all existing served versions against new CRD's validation (if changed)
 func validateV1Beta1CRDCompatibility(dynamicClient dynamic.Interface, oldCRD *apiextensionsv1beta1.CustomResourceDefinition, newCRD *apiextensionsv1beta1.CustomResourceDefinition) error {
 	logrus.Debugf("Comparing %#v to %#v", oldCRD.Spec.Validation, newCRD.Spec.Validation)
 
-	// TODO return early of all versions are equal
-	convertedCRD := &apiextensions.CustomResourceDefinition{}
-	if err := apiextensionsv1beta1.Convert_v1beta1_CustomResourceDefinition_To_apiextensions_CustomResourceDefinition(newCRD, convertedCRD, nil); err != nil {
-		return err
+	oldVersionSet := make(map[string]struct{})
```
ncdc, line 2042: You can use sets.Set[string].

pkg/controller/operators/catalog/operator.go
```diff
@@ -2013,81 +2013,101 @@ func transitionInstallPlanState(log logrus.FieldLogger, transitioner installPlan
...
+			schema := newVersion.Schema
+			if schema == nil {
+				// apiextensionsv1beta1 special case: if per-version schema is absent, use the global validation
+				schema = newCRD.Spec.Validation
+			}
+			convertedValidation := &apiextensions.CustomResourceValidation{}
+			if err := apiextensionsv1beta1.Convert_v1beta1_CustomResourceValidation_To_apiextensions_CustomResourceValidation(schema, convertedValidation, nil); err != nil {
+				return err
+			}
+			validationsMap[newVersion.Name] = convertedValidation
```
ncdc, line 2078: You can either have per-version schemas, or a global schema, but not a mix. It might be worth adjusting this a bit to take that into account. In other words, if you had this:

- oldVersion and newVersion both have:
  - 3 versions, no schema for any of them
  - global schema validation

I think the way it's currently coded, it will use the same global validation for all 3 versions, making 3 list calls, and validating everything 3 times against the same global schema... right?

pkg/controller/operators/catalog/operator.go
dtfranz: Using this now, thanks!

pkg/controller/operators/catalog/operator.go
dtfranz: That's right; good point. I've updated the loop to validate once and return early when the global validation is set. This should prevent unnecessary lists and validations.

joelanford:
Re: race condition of new unvalidated CR sneaking in

> I don't think there is anything we can really do. This is "best effort".

The only thing I could imagine would be for us to run a validating webhook for all CRDs managed by OLM where we temporarily reject CR creation during the critical section.

PR #3044, opened 2023-09-29
controller/registry/resolver: detect incoherent state

stevekuznetsov:


pkg/controller/registry/resolver/source_csvs.go
```diff
@@ -85,6 +91,20 @@ func (s *csvSource) Snapshot(ctx context.Context) (*cache.Snapshot, error) {
 			continue
 		}
 
+		if cachedSubscription, ok := csvSubscriptions[csv]; !ok || cachedSubscription == nil {
```
ncdc, line 94: I think you can just check for existence (`ok`)?

stevekuznetsov: belt + suspenders :D

pkg/controller/registry/resolver/source_csvs.go
```diff
@@ -85,6 +91,20 @@ func (s *csvSource) Snapshot(ctx context.Context) (*cache.Snapshot, error) {
...
+			// we might be in an incoherent state, so let's check with live clients to make sure
+			realSubscriptions, err := s.client.OperatorsV1alpha1().Subscriptions(csv.Namespace).List(ctx, metav1.ListOptions{})
+			if err != nil {
+				return nil, fmt.Errorf("failed to list subscriptions: %w", err)
+			}
+			for _, realSubscription := range realSubscriptions.Items {
+				if realSubscription.Status.InstalledCSV == csv.Name {
+					// oops, live cluster state is coherent
+					return nil, fmt.Errorf("lister caches incoherent for CSV %s/%s - found owning Subscription %s/%s", csv.Namespace, csv.Name, realSubscription.Namespace, realSubscription.Name)
```
ncdc, line 103: nit: inconsistent rather than incoherent? 😄

stevekuznetsov: cache coherency is a term of art?

pkg/controller/registry/resolver/source_csvs.go
```diff
@@ -85,6 +93,20 @@ func (s *csvSource) Snapshot(ctx context.Context) (*cache.Snapshot, error) {
...
 
+		if cachedSubscription, ok := csvSubscriptions[csv]; !ok || cachedSubscription == nil {
+			// we might be in an incoherent state, so let's check with live clients to make sure
+			realSubscriptions, err := s.listSubscriptions(ctx)
+			if err != nil {
+				return nil, fmt.Errorf("failed to list subscriptions: %w", err)
+			}
+			for _, realSubscription := range realSubscriptions.Items {
+				if realSubscription.Status.InstalledCSV == csv.Name {
+					// oops, live cluster state is coherent
```
grokspawn, line 104:
```suggestion
					// oops, live cluster state is incoherent
```

perdasilva:
If I'm getting it right, we're introducing a live list call to check whether the lister cache is stale. I'm not getting why

```
if realSubscription.Status.InstalledCSV == csv.Name
```

gives us that. What's stale in the lister? Has the subscription gone to the new csv but the cached version (of the sub) still points to an old csv?

Is there any concern around the added load on the API server?

pkg/controller/registry/resolver/source_csvs.go
stevekuznetsov: That's not correct - if we're here, the live cluster state is coherent - i.e. we found that a subscription really does exist with the CSV in status.

stevekuznetsov:
@perdasilva the `if` check to enter the block is predicated on that equality _not_ being true, based on the values we got in our listers. If we do a live list and find the equality to be true, we know our listers were incoherent.

pkg/controller/registry/resolver/source_csvs.go
grokspawn: That feels weird that the comment says if we got here the cluster is coherent, but the error message immediately following it says the lister caches are incoherent.  
Isn't the mismatch our basis for determining that the cluster state is incoherent?  (i.e. would the live listers ever be wrong, as suggested by the error below?)

stevekuznetsov: The cluster - and the live `List()` calls - cannot, by definition, be incoherent. Our *caches* are incoherent, which is what the error message says.

PR #3045, opened 2023-09-29
Add support for packageServerSyncInterval

tmshort:
Add support for packageServerSyncInterval

The API was updated to include `packageServerSyncInterval`

In the packageserver, read the OLMConfig at init and set the sourceProvider timeout from that (or default).

Watch the OLMConfig resource and if the sync interval changes, restart the sourceProvider.

Update to the latest api.

**Architectural changes:**

**Testing remarks:**

**Reviewer Checklist**


pkg/package-server/server/server.go
```diff
@@ -239,3 +290,24 @@ func (o *PackageServerOptions) Run(ctx context.Context) error {
...
+func (a *Operator) syncOLMConfig(obj interface{}) error {
+	olmConfig, ok := obj.(*operatorsv1.OLMConfig)
+	if !ok {
+		return fmt.Errorf("casting OLMConfig failed")
+	}
+	// restart the pod on change
+	if olmConfig.Spec.Features == nil || olmConfig.Spec.Features.PackageServerSyncInterval == nil {
+		if a.options.WakeupInterval != DefaultWakeupInterval {
+			log.Warnf("Change to olmConfig: '%v' != default '%v'", a.options.WakeupInterval, DefaultWakeupInterval)
+			os.Exit(0)
```
tmshort, line 303: @stevekuznetsov not sure I like how this causes a restart, but I couldn't figure out a way to reconfigure or restart the `sourceProvider`

stevekuznetsov: Yeah, thankfully this is a very commonplace way to achieve this in k8s so I think you're good

pkg/package-server/server/server.go
```diff
@@ -82,7 +97,7 @@ func NewPackageServerOptions(out, errOut io.Writer) *PackageServerOptions {
 		Authorization:  genericoptions.NewDelegatingAuthorizationOptions(),
 		Features:       genericoptions.NewFeatureOptions(),
 
-		WakeupInterval: 5 * time.Minute,
+		WakeupInterval: DefaultWakeupInterval,
```
joelanford, line 100: In order to avoid an entire restart, could we change the `WakeupInterval` field to a `WakeupIntervalFunc` function. We could call the function every time we need the value, and we could have that function interrogate our OLMConfig cache to get the up-to-date value.

tmshort: The only time the value matters is when we pass it into:
```
sourceProvider, err := provider.NewRegistryProvider(ctx, crClient, queueOperator, o.WakeupInterval, o.GlobalNamespace)
```
Not sure how a function would be of benefit here?

joelanford: Ah, well we'd have to plumb the function through to the place where the wakeup interval is actually used.

tmshort: Yup, it's the plumbing that would need to be introduced, and eventually into k8s code(!). I gave up looking at it when the value reached the k8s code.

tmshort:
A bit of cleanup in the recent update.

stevekuznetsov (approved):

PR #3051, opened 2023-10-04
Fix default sync interval

tmshort:
The sync default sync interval was changed to 12 hours, but there is also a CLI option that also configures the sync interval.

The OLMConfig reconcile code did not properly handle the value coming in via the CLI, and changing the sync interval from the default without an OLMConfig value specified.

The CLI option now changes the default sync interval, when the OLMConfig is not present. The 12h default sync interval is used when neither the CLI nor the OLMConfig is specified. The OLMConfig reconcile code now uses the CLI option as the default.

**Architectural changes:**

**Testing remarks:**

**Reviewer Checklist**

### DIFF
--- a/pkg/package-server/server/server.go
+++ b/pkg/package-server/server/server.go
@@ -52,7 +52,7 @@ func NewCommandStartPackageServer(ctx context.Context, defaults *PackageServerOp
 	}
 
 	flags := cmd.Flags()
-	flags.DurationVar(&defaults.SyncInterval, "interval", defaults.SyncInterval, "interval at which to re-sync CatalogSources")
+	flags.DurationVar(&defaults.DefaultSyncInterval, "interval", defaults.DefaultSyncInterval, "default interval at which to re-sync CatalogSources")
 	flags.StringVar(&defaults.GlobalNamespace, "global-namespace", defaults.GlobalNamespace, "Name of the namespace where the global CatalogSources are located")
 	flags.StringVar(&defaults.Kubeconfig, "kubeconfig", defaults.Kubeconfig, "path to the kubeconfig used to connect to the Kubernetes API server and the Kubelets (defaults to in-cluster config)")
 	flags.BoolVar(&defaults.Debug, "debug", defaults.Debug, "use debug log level")
@@ -71,8 +71,9 @@ type PackageServerOptions struct {
 	Authorization  *genericoptions.DelegatingAuthorizationOptions
 	Features       *genericoptions.FeatureOptions
 
-	GlobalNamespace string
-	SyncInterval    time.Duration
+	GlobalNamespace     string
+	DefaultSyncInterval time.Duration
+	CurrentSyncInterval time.Duration
 
 	Kubeconfig   string
 	RegistryAddr string
@@ -95,7 +96,8 @@ func NewPackageServerOptions(out, errOut io.Writer) *PackageServerOptions {
 		Authorization:  genericoptions.NewDelegatingAuthorizationOptions(),
 		Features:       genericoptions.NewFeatureOptions(),
 
-		SyncInterval: DefaultWakeupInterval,
+		DefaultSyncInterval: DefaultWakeupInterval,
+		CurrentSyncInterval: DefaultWakeupInterval,
 
 		DisableAuthForTesting: false,
 		Debug:                 false,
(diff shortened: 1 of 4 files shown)


pkg/package-server/server/server.go
```diff
@@ -52,7 +52,7 @@ func NewCommandStartPackageServer(ctx context.Context, defaults *PackageServerOp
 	}
 
 	flags := cmd.Flags()
-	flags.DurationVar(&defaults.SyncInterval, "interval", defaults.SyncInterval, "interval at which to re-sync CatalogSources")
+	flags.DurationVar(&defaults.DefaultSyncInterval, "interval", defaults.DefaultSyncInterval, "default interval at which to re-sync CatalogSources")
```
awgreene, line 55: Argh, defaults.DefaultSyncInterval :P

awgreene (commented):
One non-blocking nit
/approve

pkg/package-server/server/server.go
tmshort: It makes more sense when the structure is not referenced as `defaults`.

tmshort:
Waiting until some additional testing is performed.

dinhxuanvu:
Hold for reviewing
/hold

pkg/package-server/server/server.go
```diff
@@ -95,7 +96,8 @@ func NewPackageServerOptions(out, errOut io.Writer) *PackageServerOptions {
 		Authorization:  genericoptions.NewDelegatingAuthorizationOptions(),
 		Features:       genericoptions.NewFeatureOptions(),
 
-		SyncInterval: DefaultWakeupInterval,
+		DefaultSyncInterval: DefaultWakeupInterval,
+		CurrentSyncInterval: DefaultWakeupInterval,
```
dinhxuanvu, line 100: I'm not seeing the value of splitting one `SyncInterval` into 2 variables. On line 255, you can simply compare `o.SyncInterval` to `DefaultWakeupInterval` to detect if SyncInterval is using CLI variable and log the change accordingly. You don't even need to do this `o.CurrentSyncInterval = o.DefaultSyncInterval` as it is already assigned when `NewPackageServerOptions` is initialized.

For OLMConfig part, you simply need to compare this `o.SyncInterval` to OLConfig spec to detect the change and log it. There is no need for a "temporary" field to hold the current interval given the current one should be the one in use regardless it comes from CLI or OLMConfig. The default one is also the const that is specified in the file.

dinhxuanvu (commented):
Looking good but I have some feedbacks on the SyncInternal field part.

pkg/package-server/server/server.go
tmshort: That is what the original code was doing, and it was causing a crash loop.

Originally, during initialization, the `o.SyncInterval` was set to the `DefaultWakeupInterval`. If the `--interval` CLI option was set, it updated `o.SyncInterval`. When reconciliation happens and the OLMConfig is not present or the `packageServerSyncInterval` field is not present, we needed to check against the default. But the `o.SyncInterval` is no longer the default (because it was set via `--interval`); so a comparison to the default (i.e. `DefaultWakeupInterval`) will indicate that a change has occurred (which it hasn't), so a restart needs to happen.

That caused a crash loop.

The `NewPackageServerOptions` function initializes both values to the constant default (i.e. `DefaultWakeupInterval` or `12h`). However, `o.SyncInterval` is changed by the `NewCommandStartPackageServer` function processing of the flags, specifically the `--interval` flag, so it is no longer starts at the default.

The CLI `--interval` parameter effectively changes the default; we can't compare against the constant default anymore. Thus, the need for two values (current and default).

tmshort: TL;DR: we need to recognize `o.SyncInterval` being set by `--interval` when OLMConfig is not present.

dinhxuanvu: I see what you were saying. I start to think this `--interval` setting is a problem that needs to go away. It was helpful when OLMConfig wasn't a thing but now it is, this setting is obsolete. This becomes a 3-way battle for the ultimate control of resync usage. The precedence on this area is not well-defined which leads to this messy conditional loop. I would think one of them should be removed ideally the `--interval`. However, given the packageserver may be gone soon, I think it may not be worth it the effort. I think I'm fine with this change.

dinhxuanvu (approved):
/lgtm
/hold cancel

tmshort:
I still want to wait until other (ahem, downstream) testing has completed. It's looking good so far.

PR #3066, opened 2023-10-16
test/e2e/fail-forward: use internal catalogs, change failure mode

stevekuznetsov:


test/e2e/csv_e2e_test.go
```diff
@@ -4289,39 +4287,12 @@ func buildAPIServiceCleanupFunc(c operatorclient.ClientInterface, apiServiceName
...
-	}
-	_, err := c.ApiextensionsInterface().ApiextensionsV1().CustomResourceDefinitions().Create(context.TODO(), out, metav1.CreateOptions{})
-	if err != nil {
-		return nil, err
-	}
-
-	return buildCRDCleanupFunc(c, crd.GetName()), nil
+func createCRD(c operatorclient.ClientInterface, crd apiextensionsv1.CustomResourceDefinition) (cleanupFunc, error) {
+	return createV1CRD(c, crd)
 }
```
tmshort, line 4292: This could just be removed now and replaced with `createV1CRD()` everywhere it's used.

stevekuznetsov: made that change in #3067

test/e2e/fail_forward_e2e_test.go
```diff
@@ -102,14 +199,39 @@ var _ = Describe("Fail Forward Upgrades", func() {
...
+				}
+				if len(fetched.Items) == 0 {
+					return true, nil
+				}
+				ctx.Ctx().Logf("waiting for the catalog source %s pod to be deleted...", fetched.Items[0].GetName())
+				return false, nil
+			}).Should(BeTrue())
+
+			By("updating the catalog with a v0.2.0 bundle image that has an invalid CRD")
+			pms, crds, csvs, err = loadData("v0.1.0", "v0.2.0-invalid-crd")
```
tmshort, line 224: Behavior change? Removing the prior catalog rather than replacing, and using `v0.2.0-invalid-crd` rather than `v0.2.0`?

stevekuznetsov: > Removing the prior catalog rather than replacing

If you read the implementation of the magic catalog, it's identical: 

https://github.com/operator-framework/operator-lifecycle-manager/blob/1ca6d9e799cd7da2bc8da287b310da76846efeb2/test/e2e/magic_catalog.go#L84-L91

> using v0.2.0-invalid-crd rather than v0.2.0?

That is the point of the PR, yes. We trigger a failing InstallPlan by having invalid data in the bundle as opposed to missing permissions.

scripts/build_local.sh
```diff
@@ -29,4 +29,18 @@ if [ -n "$KIND" ]; then
   kind load docker-image quay.io/operator-framework/olm:local ${KIND_FLAGS}
   kind load docker-image quay.io/operator-framework/olm-e2e:local ${KIND_FLAGS}
   kind load docker-image hang:10 ${KIND_FLAGS}
+  kind load docker-image quay.io/operatorhubio/catalog:latest ${KIND_FLAGS}
```
tmshort, line 32: This image does not exist, i.e. it was not `docker pull`d. Does this need to happen? Or does it need to be created?

stevekuznetsov:
@tmshort will do - was ironing out the last weirdness with Alex, it should be pushed & ready for review this morning.

scripts/build_local.sh
tmshort: It _does_ make sense to either build these locally, or pull them down earlier to avoid quay.io timeouts, but completely fair to focus on the most pressing issues.

test/e2e/fail_forward_e2e_test.go
```diff
@@ -119,36 +214,41 @@ var _ = Describe("Fail Forward Upgrades", func() {
...
-			Expect(err).To(BeNil())
+			By("updating the catalog with a v0.2.1 bundle that has an invalid CSV")
+			cleanup, deployError := updateCatalogSource(generatedNamespace.GetName(), catalogSourceName, "v0.1.0", "v0.2.0-invalid", "v0.2.1-invalid")
+			Expect(deployError).To(BeNil())
+			if cleanup != nil {
+				cleanups = append(cleanups, cleanup)
+			}
 
-			By("waiting for the subscription to maintain the example-operator.v0.2.0 status.currentCSV")
-			Consistently(subscriptionCurrentCSVGetter(crclient, subscription.GetNamespace(), subscription.GetName())).Should(Equal("example-operator.v0.2.0"))
```
stevekuznetsov, line 141: This was only valid when the bundle failed to unpack (and, anyway, in the distributed system is not a good way to prove something unless you wait forever :) )

test/e2e/fail_forward_e2e_test.go
```diff
@@ -119,36 +214,41 @@ var _ = Describe("Fail Forward Upgrades", func() {
...
+			By("waiting for the subscription to have the example-operator.v0.2.1&invalid status.currentCSV")
+			_, err := fetchSubscription(crclient, subscription.GetNamespace(), subscription.GetName(), subscriptionHasCurrentCSV("example-operator.v0.2.1&invalid"))
+			Expect(err).Should(BeNil())
 
-			By("patching the OperatorGroup to increase the bundle unpacking timeout")
-			addBundleUnpackTimeoutOGAnnotation(context.Background(), c, types.NamespacedName{Name: ogName, Namespace: generatedNamespace.GetName()}, "5m")
```
stevekuznetsov, line 144: The bundle unpacking logic here was necessary when the failure mode was expecting the non-existent bundle image tag to fail to pull & unpack. Not necessary any longer.

stevekuznetsov:
/retitle @stevekuznetsov
[test/e2e/fail-forward: use internal catalogs, change failure mode](https://github.com/operator-framework/operator-lifecycle-manager/pull/3066/commits/91f6de46a5870f4230f31cb78066db6e80eec005)

stevekuznetsov:
/retitle test/e2e/fail-forward: use internal catalogs, change failure mode

test/e2e/util.go
```diff
@@ -382,12 +382,16 @@ func fetchCatalogSourceOnStatus(crc versioned.Interface, name, namespace string,
...
-			fmt.Printf("failed to fetch catalogSource %s/%s: %v\n", namespace, name, err)
+			err = fmt.Errorf("failed to fetch catalogSource %s/%s: %v\n", namespace, name, err)
+			fmt.Println(err.Error())
 			return false, err
 		}
 		return check(fetched), nil
 	})
 
+	if err != nil {
+		err = fmt.Errorf("failed to wati for catalog source to reach intended state: %w", err)
```
tmshort, line 393:
```suggestion
		err = fmt.Errorf("failed to wait for catalog source to reach intended state: %w", err)
```

tmshort:
/lgtm
Although there is a typo.
We have mostly passage!

tmshort (approved):

test/e2e/util.go
awgreene: Let's fix this in a followup.

awgreene:
Exciting!
/lgtm

PR #3067, opened 2023-10-16
test/e2e: don't use internal versions of api extensions

stevekuznetsov:
The internal versions of the API Extensions client objects are what client calls get converted into for processing within the server. This mechanism allows many different clients to use many different outward-facing versions, while the code within the server only ever runs against one type. There is no utility to these types outside of the server, especially for code that's making client calls. We can simply use the client-facing types and not require coversion anywhere.


test/e2e/csv_e2e_test.go
```diff
@@ -4289,39 +4287,8 @@ func buildAPIServiceCleanupFunc(c operatorclient.ClientInterface, apiServiceName
...
-	scheme := runtime.NewScheme()
-	if err := apiextensions.AddToScheme(scheme); err != nil {
-		return nil, err
-	}
-	if err := apiextensionsv1.AddToScheme(scheme); err != nil {
-		return nil, err
-	}
-	if err := scheme.Convert(&crd, out, nil); err != nil {
-		return nil, err
-	}
```
tmshort, line 4323: So, this setup is unnecessary?

tmshort (approved):
/lgtm
Just a single question.

test/e2e/csv_e2e_test.go
stevekuznetsov: Yes, entirely.

stevekuznetsov: (the commit message explains why)

PR #3069, opened 2023-10-18
Do not derive installplan.spec.clusterServiceNames from bundle IDs

awgreene:
Problem: Historically, when creating an installPlan it's spec.ClusterServiceVersionNames was derived from two sources:
1. The names of CSVs found in "steps" of the installPlan's status.plan
2. The metadata associated with the bundle image

These sources couldn't result in duplicate entries as the unpacking job would finish after the installPlan was created and the steps weren't populated until the unpacking job finished.

OLM was recently updated to complete the unpacking jobs prior to creating the installPlan, which means that the two sources can cause duplicate entries to appear in the array.

Solution: Now that OLM creates the installPlan after successfully unpacking the bundles, we no longer need to derive the names of the CSV from the bundle metadata.


tmshort (approved):
/lgtm

PR #3070, opened 2023-10-19
flaky-e2e-tests fix

dtfranz:
Fixing flake tests so they don't fail 100% of the time.

E2E tests were updated recently to use a generated namespace with an OperatorGroup created from the beginning of each test. However some of the tests were not updated to reflect this and were still creating and deleting resources in an additional namespace, or began to fail because the test environment required that no OperatorGroups exist from the beginning of the test.

This commit addresses two tests that have been failing at a rate of 100% due to the issues stated above.

### DIFF
--- a/test/e2e/installplan_e2e_test.go
+++ b/test/e2e/installplan_e2e_test.go
@@ -3076,22 +3076,6 @@ var _ = Describe("Install Plan", func() {
 			GinkgoT().Logf("%s: %s", time.Now().Format("15:04:05.9999"), s)
 		}
 
-		ns := &corev1.Namespace{}
-		ns.SetName(genName("ns-"))
-
-		// Create a namespace an OperatorGroup
-		ns, err := c.KubernetesInterface().CoreV1().Namespaces().Create(context.Background(), ns, metav1.CreateOptions{})
-		require.NoError(GinkgoT(), err)
-		deleteOpts := &metav1.DeleteOptions{}
-		defer func() {
-			require.NoError(GinkgoT(), c.KubernetesInterface().CoreV1().Namespaces().Delete(context.Background(), generatedNamespace.GetName(), *deleteOpts))
-		}()
-
-		og := &operatorsv1.OperatorGroup{}
-		og.SetName("og")
-		_, err = crc.OperatorsV1().OperatorGroups(generatedNamespace.GetName()).Create(context.Background(), og, metav1.CreateOptions{})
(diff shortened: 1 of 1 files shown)


test/e2e/installplan_e2e_test.go
```diff
@@ -3223,17 +3208,11 @@ var _ = Describe("Install Plan", func() {
...
 
 		BeforeEach(func() {
-			ns = &corev1.Namespace{}
-			ns.SetName(genName("ns-"))
-
-			// Create a namespace
-			Eventually(func() error {
-				return ctx.Ctx().Client().Create(context.Background(), ns)
-			}, timeout, interval).Should(Succeed(), "could not create Namespace")
+			// Make sure there are no OGs in the namespace already
```
dtfranz, line 3214: OGs are created in the `generatedNamespace` before each test in this file but this particular test relies on a blank slate to work consistently. This is a quick-and-dirty way to make sure we get that clean slate.

test/e2e/installplan_e2e_test.go
```diff
@@ -3247,13 +3226,12 @@ var _ = Describe("Install Plan", func() {
 			outIP.Status = ip.Status
 			_, err = crc.OperatorsV1alpha1().InstallPlans(generatedNamespace.GetName()).UpdateStatus(context.Background(), outIP, metav1.UpdateOptions{})
 			Expect(err).NotTo(HaveOccurred())
+
```
dinhxuanvu, line 3229: This is no needed.

test/e2e/installplan_e2e_test.go
dinhxuanvu: I understand the intention is to reuse one namespace all for tests under this umbrella but should you take care of the fact a new InstallPlan still being created for each test in the same namespace?

dinhxuanvu (changes requested):

test/e2e/installplan_e2e_test.go
dinhxuanvu: Ignore this comment, the AfterEach should take care of this. So we are cool here.

test/e2e/installplan_e2e_test.go
dtfranz: Woops, yeah I forgot to push up that change.

dinhxuanvu (approved):
/approve
/lgtm

tmshort (approved):

PR #3075, opened 2023-10-25
Constrain RBACs requested by OLM

awgreene:

PR #3076, opened 2023-10-25
Specify exact verbs that OLM requires

awgreene:
This will specifically allow OLM to avoid the "use" verb for SCCs that it wasn't configured to work with on OpenShift clusters.


awgreene:
/hold
I'm considering this as an option over https://github.com/operator-framework/operator-lifecycle-manager/pull/3075 as it will be less disruptive.

awgreene:
/unhold
Discussed with the team and decided we're gonna go with this change in favor of #3075

ncdc (approved):

Issue #3062, opened 2023-10-11
[FAIL] Install Plan [It] creation with permissions


stevekuznetsov:
```
 • [FAILED] [363.523 seconds]
Install Plan [It] creation with permissions
/home/runner/work/operator-lifecycle-manager/operator-lifecycle-manager/test/e2e/installplan_e2e_test.go:2713

  Captured StdOut/StdErr Output >>
  waiting 103.588823ms for catalog pod install-plan-e2e-dhqvf/nginx-catalogvqw54 to be available (for sync) - NO_CONNECTION
  waiting 701.742488ms for catalog pod install-plan-e2e-dhqvf/nginx-catalogvqw54 to be available (for sync) - IDLE
  waiting 355.207008ms for catalog pod install-plan-e2e-dhqvf/nginx-catalogvqw54 to be available (for sync) - CONNECTING
  waiting 598.533766ms for catalog pod install-plan-e2e-dhqvf/nginx-catalogvqw54 to be available (for sync) - TRANSIENT_FAILURE
  waiting 2.399265723s for catalog pod install-plan-e2e-dhqvf/nginx-catalogvqw54 to be available (for sync) - CONNECTING
  waiting 20.000601951s for catalog pod install-plan-e2e-dhqvf/nginx-catalogvqw54 to be available (for sync) - TRANSIENT_FAILURE
  waiting 2.799121894s for catalog pod install-plan-e2e-dhqvf/nginx-catalogvqw54 to be available (for sync) - CONNECTING
  waiting 20.000701415s for catalog pod install-plan-e2e-dhqvf/nginx-catalogvqw54 to be available (for sync) - TRANSIENT_FAILURE
  waiting 4.600022137s for catalog pod install-plan-e2e-dhqvf/nginx-catalogvqw54 to be available (for sync) - READY
  probing catalog nginx-catalogvqw54 pod with address nginx-catalogvqw54.install-plan-e2e-dhqvf.svc:50051
  skipping health check
  Using the kubectl kubectl binary
  Using the artifacts/install-plan-e2e-dhqvf output directory
  Storing the test artifact output in the artifacts/install-plan-e2e-dhqvf directory
  Collecting get catalogsources -o yaml output...
  Collecting get subscriptions -o yaml output...
  Collecting get operatorgroups -o yaml output...
  Collecting get clusterserviceversions -o yaml output...
  Collecting get installplans -o yaml output...
  Collecting get pods -o wide output...
[...]

PR #3077, opened 2023-10-25
Fix some of the e2e loops

tmshort:
Some of the e2e loops are a bit flakey, make them more robust
* Retry on certain errors
* Use Eventually() consistently (avoid wait.Poll())
* Clarify logging
* Reduce some logging
* Change csvExists() to waitForCsvToDelete(), as that's how it's used
* Change awaitCSV() to fetchCSV()

**Reviewer Checklist**

### DIFF
--- a/test/e2e/catalog_e2e_test.go
+++ b/test/e2e/catalog_e2e_test.go
@@ -214,7 +214,7 @@ var _ = Describe("Starting CatalogSource e2e tests", func() {
 			return err
 		}).Should(Succeed())
 
-		_, err = awaitCSV(crc, generatedNamespace.GetName(), mainCSV.GetName(), csvSucceededChecker)
+		_, err = fetchCSV(crc, generatedNamespace.GetName(), mainCSV.GetName(), csvSucceededChecker)
 		Expect(err).ShouldNot(HaveOccurred())
 
 		// Update manifest
@@ -364,7 +364,7 @@ var _ = Describe("Starting CatalogSource e2e tests", func() {
 		subscription, err := fetchSubscription(crc, generatedNamespace.GetName(), subscriptionName, subscriptionStateAtLatestChecker())
 		Expect(err).ShouldNot(HaveOccurred())
 		Expect(subscription).ShouldNot(BeNil())
-		_, err = fetchCSV(crc, subscription.Status.CurrentCSV, generatedNamespace.GetName(), buildCSVConditionChecker(v1alpha1.CSVPhaseSucceeded))
+		_, err = fetchCSV(crc, generatedNamespace.GetName(), subscription.Status.CurrentCSV, buildCSVConditionChecker(v1alpha1.CSVPhaseSucceeded))
 		Expect(err).ShouldNot(HaveOccurred())
 
 		ipList, err := crc.OperatorsV1alpha1().InstallPlans(generatedNamespace.GetName()).List(context.Background(), metav1.ListOptions{})
@@ -455,7 +455,7 @@ var _ = Describe("Starting CatalogSource e2e tests", func() {
 		subscription, err := fetchSubscription(crc, generatedNamespace.GetName(), subscriptionName, subscriptionStateAtLatestChecker())
 		Expect(err).ShouldNot(HaveOccurred())
 		Expect(subscription).ToNot(BeNil())
-		_, err = fetchCSV(crc, subscription.Status.CurrentCSV, generatedNamespace.GetName(), buildCSVConditionChecker(v1alpha1.CSVPhaseSucceeded))
+		_, err = fetchCSV(crc, generatedNamespace.GetName(), subscription.Status.CurrentCSV, buildCSVConditionChecker(v1alpha1.CSVPhaseSucceeded))
 		Expect(err).ShouldNot(HaveOccurred())
 	})
 
@@ -597,7 +597,7 @@ var _ = Describe("Starting CatalogSource e2e tests", func() {
 		subscription, err := fetchSubscription(crc, generatedNamespace.GetName(), subscriptionName, subscriptionStateAtLatestChecker())
 		Expect(err).ShouldNot(HaveOccurred())
 		Expect(subscription).ShouldNot(BeNil())
-		_, err = fetchCSV(crc, subscription.Status.CurrentCSV, generatedNamespace.GetName(), csvSucceededChecker)
+		_, err = fetchCSV(crc, generatedNamespace.GetName(), subscription.Status.CurrentCSV, csvSucceededChecker)
 		Expect(err).ShouldNot(HaveOccurred())
 
 		// Update the catalog's address to point at the other registry pod's cluster ip
@@ -613,7 +613,7 @@ var _ = Describe("Starting CatalogSource e2e tests", func() {
 		}).Should(Succeed())
 
 		// Wait for the replacement CSV to be installed
-		_, err = awaitCSV(crc, generatedNamespace.GetName(), replacementCSV.GetName(), csvSucceededChecker)
+		_, err = fetchCSV(crc, generatedNamespace.GetName(), replacementCSV.GetName(), csvSucceededChecker)
 		Expect(err).ShouldNot(HaveOccurred())
 	})
 
@@ -928,7 +928,7 @@ var _ = Describe("Starting CatalogSource e2e tests", func() {
 		Expect(subscription).ShouldNot(BeNil())
 
 		// Wait for csv to succeed
-		_, err = fetchCSV(crc, subscription.Status.CurrentCSV, subscription.GetNamespace(), csvSucceededChecker)
+		_, err = fetchCSV(crc, subscription.GetNamespace(), subscription.Status.CurrentCSV, csvSucceededChecker)
 		Expect(err).ShouldNot(HaveOccurred())
 
 		registryCheckFunc := func(podList *corev1.PodList) bool {
@@ -1033,7 +1033,7 @@ var _ = Describe("Starting CatalogSource e2e tests", func() {
 		Expect(subscription).ShouldNot(BeNil())
 
 		// Wait for csv to succeed
-		csv, err := fetchCSV(crc, subscription.Status.CurrentCSV, subscription.GetNamespace(), csvSucceededChecker)
+		csv, err := fetchCSV(crc, subscription.GetNamespace(), subscription.Status.CurrentCSV, csvSucceededChecker)
 		Expect(err).ShouldNot(HaveOccurred())
 
 		// check version of running csv to ensure the latest version (0.9.2) was installed onto the cluster
@@ -1144,7 +1144,7 @@ var _ = Describe("Starting CatalogSource e2e tests", func() {
 		Expect(subscription).ShouldNot(BeNil())
 
 		By("waiting for busybox v2 csv to succeed and check the replaces field")
-		csv, err := fetchCSV(crc, subscription.Status.CurrentCSV, subscription.GetNamespace(), csvSucceededChecker)
+		csv, err := fetchCSV(crc, subscription.GetNamespace(), subscription.Status.CurrentCSV, csvSucceededChecker)
 		Expect(err).ShouldNot(HaveOccurred())
 		Expect(csv.Spec.Replaces).To(Equal("busybox.v1.0.0"))
 
@@ -1157,7 +1157,7 @@ var _ = Describe("Starting CatalogSource e2e tests", func() {
 		Expect(subscription).ShouldNot(BeNil())
 
 		By("waiting for busybox-dependency v2 csv to succeed and check the replaces field")
-		csv, err = fetchCSV(crc, subscription.Status.CurrentCSV, subscription.GetNamespace(), csvSucceededChecker)
+		csv, err = fetchCSV(crc, subscription.GetNamespace(), subscription.Status.CurrentCSV, csvSucceededChecker)
 		Expect(err).ShouldNot(HaveOccurred())
 		Expect(csv.Spec.Replaces).To(Equal("busybox-dependency.v1.0.0"))
 	})
--- a/test/e2e/csv_e2e_test.go
+++ b/test/e2e/csv_e2e_test.go
@@ -10,7 +10,6 @@ import (
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
-	operatorsv1 "github.com/operator-framework/api/pkg/operators/v1"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/clientset/versioned"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/install"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/operatorclient"
@@ -30,6 +29,7 @@ import (
 	apiregistrationv1 "k8s.io/kube-aggregator/pkg/apis/apiregistration/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	operatorsv1 "github.com/operator-framework/api/pkg/operators/v1"
 	operatorsv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 	"github.com/operator-framework/operator-lifecycle-manager/test/e2e/ctx"
 )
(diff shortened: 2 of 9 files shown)


test/e2e/csv_e2e_test.go
```diff
@@ -4474,47 +4465,65 @@ func awaitCSV(c versioned.Interface, namespace, name string, checker csvConditio
...
 
 func waitForDeployment(namespace string, c operatorclient.ClientInterface, name string) error {
-	return wait.Poll(pollInterval, pollDuration, func() (bool, error) {
-		_, err := c.GetDeployment(namespace, name)
+	var err error
+
+	Eventually(func() (bool, error) {
+		_, err = c.GetDeployment(namespace, name)
 		if err != nil {
-			if apierrors.IsNotFound(err) {
```
ncdc, line 4480: I'm not sure it's safe to remove this?

tmshort: The reasoning is that regardless of the error ("not found" or "i/o error"), we have to continue to retry until the deployment is found. Note that this has changed from `wait.Poll()` to `Eventually()` which have ever-so-slightly different semantics.
* `Eventually()` continues until the first return value meets the test (`Should(BeTrue())`) and all subsequent return values are "0".
* `wait.Poll()` continues until the `bool` return value is `true` or the `error` return value is `non-nil`. In this case, the `wait.Poll()` was terminating on any error other than "not found", including "i/o errors", which should be retried.

tmshort: (Also, this is a WIP, so this is not yet final. 😄)

test/e2e/csv_e2e_test.go
```diff
@@ -4433,88 +4427,83 @@ var csvAnyChecker = buildCSVConditionChecker(operatorsv1alpha1.CSVPhasePending,
...
-		fetchedCSV, err = c.OperatorsV1alpha1().ClusterServiceVersions(namespace).Get(context.TODO(), name, metav1.GetOptions{})
+	Eventually(func() bool {
+		fetched, err = c.OperatorsV1alpha1().ClusterServiceVersions(namespace).Get(context.TODO(), name, metav1.GetOptions{})
 		if err != nil {
-			return false, err
+			ctx.Ctx().Logf("error getting csv %s: %v", name, err)
+			return false
 		}
-		return checker(fetchedCSV), nil
+		ctx.Ctx().Logf("%s - %s (%s): %s", name, fetched.Status.Phase, fetched.Status.Reason, fetched.Status.Message)
```
stevekuznetsov, line 4440: This kind of log message will spam on every iteration, please only log on edge transitions.

test/e2e/csv_e2e_test.go
stevekuznetsov: Do we really suffer from failing to reach the API server during tests? Yikes.

test/e2e/csv_e2e_test.go
```diff
@@ -4433,88 +4427,83 @@ var csvAnyChecker = buildCSVConditionChecker(operatorsv1alpha1.CSVPhasePending,
...
-
-	Eventually(func() (bool, error) {
-		ctx.Ctx().Logf("waiting for deployment %s to delete", name)
+	ctx.Ctx().Logf("waiting for deployment %s to delete", name)
+	Eventually(func() bool {
 		_, err := c.GetDeployment(namespace, name)
-		if apierrors.IsNotFound(err) {
-			ctx.Ctx().Logf("deleted %s", name)
-			return true, nil
+		if apierrors.IsNotFound(err) || apierrors.IsGone(err) {
```
stevekuznetsov, line 4472: 410 Gone is not something this call can return and not something you should check for here, that's a different semantic for time-travel queries.

test/e2e/csv_e2e_test.go
```diff
@@ -4433,88 +4427,83 @@ var csvAnyChecker = buildCSVConditionChecker(operatorsv1alpha1.CSVPhasePending,
...
+			ctx.Ctx().Logf("deployment %s deleted", name)
```
stevekuznetsov, line 4473: please ensure that namespaced resources are referred to with namespaced identifiers (namespace/name)

test/e2e/csv_e2e_test.go
```diff
@@ -4433,88 +4427,83 @@ var csvAnyChecker = buildCSVConditionChecker(operatorsv1alpha1.CSVPhasePending,
...
+		csv, err := c.OperatorsV1alpha1().ClusterServiceVersions(namespace).Get(context.TODO(), name, metav1.GetOptions{})
+		if apierrors.IsNotFound(err) || apierrors.IsGone(err) {
+			ctx.Ctx().Logf("csv %s deleted", name)
+			return true
+		}
+		if err != nil {
+			ctx.Ctx().Logf("error getting csv %s: %v", name, err)
+		}
+		if csv != nil {
+			ctx.Ctx().Logf("csv %s status: %s (%s): %s", name, csv.Status.Phase, csv.Status.Reason, csv.Status.Message)
```
stevekuznetsov, line 4500: similarly, this is spammy, please log on edges

test/e2e/csv_e2e_test.go
tmshort: 1) WIP
2) Many of these logs ARE currently on every iteration
3) I had tried on transitions (which I agree is better), and it wasn't working.

I'll clean this up when I'm ready to commit.

test/e2e/csv_e2e_test.go
tmshort: The API documentation is not clear on that, thanks.

test/e2e/csv_e2e_test.go
tmshort: Makes sense; this is a simple rewording of the prior message.

test/e2e/csv_e2e_test.go
```diff
@@ -4433,88 +4427,83 @@ var csvAnyChecker = buildCSVConditionChecker(operatorsv1alpha1.CSVPhasePending,
...
-			return false, err
+			ctx.Ctx().Logf("error getting deployment %s: %v", name, err)
 		}
-		return false, nil
-	}).Should(BeTrue())
-	return err
+		return false
+	}).WithTimeout(pollDuration).WithPolling(pollInterval).Should(BeTrue())
+
+	// Keep signature, Eventually() will error exit on timeout, this only happens on success
```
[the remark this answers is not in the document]
tmshort: For this WIP, I didn't want to make such a dramatic change until I was sure. But yeah, there are other functions that are basically dups of each other, and the signatures have unnecessary returns.

tmshort:
Definitely seeing #3062 a lot, so I marked it a flake.

test/e2e/csv_e2e_test.go
```diff
@@ -4432,89 +4426,86 @@ var csvFailedChecker = buildCSVConditionChecker(operatorsv1alpha1.CSVPhaseFailed
 var csvAnyChecker = buildCSVConditionChecker(operatorsv1alpha1.CSVPhasePending, operatorsv1alpha1.CSVPhaseSucceeded, operatorsv1alpha1.CSVPhaseReplacing, operatorsv1alpha1.CSVPhaseDeleting, operatorsv1alpha1.CSVPhaseFailed)
 var csvCopiedChecker = buildCSVReasonChecker(operatorsv1alpha1.CSVReasonCopied)
 
-func fetchCSV(c versioned.Interface, name, namespace string, checker csvConditionChecker) (*operatorsv1alpha1.ClusterServiceVersion, error) {
-	var fetchedCSV *operatorsv1alpha1.ClusterServiceVersion
-	var err error
+func fetchCSV(c versioned.Interface, name, namespace string, checker csvConditionChecker) (fetched *operatorsv1alpha1.ClusterServiceVersion, err error) {
```
grokspawn, line 4429: > (fetched *operatorsv1alpha1.ClusterServiceVersion, err error)
I don't typically see go return types given instance names.  Is there a case where that is preferred?

tmshort: So, AFAICT (based on testing and documentation) `Eventually()` is weird because it terminates the test function via an assertion, so the code following it is not necessarily invoked (i.e. the `return`). Naming the arguments seems to allow the error to be passed back to the caller of this function to assert on as well.
Without the named arguments, I was seeing panics on the return values when a timeout occured.

tmshort: I suppose one could argue that either:
a. `Eventually()` should be used to call this function rather than having `Eventually()` inside the function (there are dozens if not hundreds of invocations)
b. `wait.Poll()` should be used instead, taking advantage of the (existing) error handling by the caller.

test/e2e/catalog_e2e_test.go
```diff
@@ -214,7 +214,7 @@ var _ = Describe("Starting CatalogSource e2e tests", func() {
 			return err
 		}).Should(Succeed())
 
-		_, err = awaitCSV(crc, generatedNamespace.GetName(), mainCSV.GetName(), csvSucceededChecker)
+		_, err = fetchCSV(crc, mainCSV.GetName(), generatedNamespace.GetName(), csvSucceededChecker)
```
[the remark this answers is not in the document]
stevekuznetsov: FWIW `namespace, name` is the most common & understood order.

test/e2e/csv_e2e_test.go
```diff
@@ -4432,89 +4426,86 @@ var csvFailedChecker = buildCSVConditionChecker(operatorsv1alpha1.CSVPhaseFailed
...
+	return fetched, nil
 }
 
-func awaitCSV(c versioned.Interface, namespace, name string, checker csvConditionChecker) (*operatorsv1alpha1.ClusterServiceVersion, error) {
-	var fetched *operatorsv1alpha1.ClusterServiceVersion
-	var err error
-
-	Eventually(func() (bool, error) {
-		fetched, err = c.OperatorsV1alpha1().ClusterServiceVersions(namespace).Get(context.TODO(), name, metav1.GetOptions{})
+func waitForDeployment(namespace string, c operatorclient.ClientInterface, name string) (err error) {
```
stevekuznetsov, line 4451: what's the utility of this func that does not return the delpoyment or wait for any particular condition?

test/e2e/csv_e2e_test.go
stevekuznetsov: Named returns are one of the most insidious and pernicious features in Go, often seen as an anti-pattern. Perhaps we just use control flow that is grok-able and doesn't require this?

test/e2e/csv_e2e_test.go
```diff
@@ -4432,89 +4426,86 @@ var csvFailedChecker = buildCSVConditionChecker(operatorsv1alpha1.CSVPhaseFailed
...
-func fetchCSV(c versioned.Interface, name, namespace string, checker csvConditionChecker) (*operatorsv1alpha1.ClusterServiceVersion, error) {
-	var fetchedCSV *operatorsv1alpha1.ClusterServiceVersion
-	var err error
+func fetchCSV(c versioned.Interface, name, namespace string, checker csvConditionChecker) (fetched *operatorsv1alpha1.ClusterServiceVersion, err error) {
+	var lastMessage string
 
 	ctx.Ctx().Logf("waiting for CSV %s/%s to reach condition", namespace, name)
-	Eventually(func() (bool, error) {
-		fetchedCSV, err = c.OperatorsV1alpha1().ClusterServiceVersions(namespace).Get(context.TODO(), name, metav1.GetOptions{})
+	Eventually(func() bool {
```
stevekuznetsov, line 4433: Can you please provide the reasoning behind your changes? I've stared at the `awaitCSV` -> `fetchCSV` change for a bit and I'm, not really sure what the difference is.

test/e2e/subscription_e2e_test.go
```diff
@@ -3390,7 +3389,11 @@ func updateInternalCatalog(t GinkgoTInterface, c operatorclient.ClientInterface,
 			fmt.Println("catalog updated")
 			return true
 		}
-		fmt.Printf("waiting for catalog pod %v to be available (after catalog update) - %s\n", catalog.GetName(), catalog.Status.GRPCConnectionState.LastObservedState)
+		msg := fmt.Sprintf("waiting for catalog pod %v to be available (after catalog update) - %s\n", catalog.GetName(), catalog.Status.GRPCConnectionState.LastObservedState)
+		if msg != logMsg {
```
stevekuznetsov, line 3393: global nit: the other funcs expose time stamps to help communicate the passage of time

test/e2e/webhook_e2e_test.go
```diff
@@ -690,8 +689,12 @@ var _ = Describe("CSVs with a Webhook", func() {
 			cleanupSubscription := createSubscriptionForCatalog(crc, source.GetNamespace(), subscriptionName, source.GetName(), packageName, channelName, "", operatorsv1alpha1.ApprovalAutomatic)
 			defer cleanupSubscription()
 
+			_, err = fetchSubscription(crc, source.GetNamespace(), subscriptionName, subscriptionHasInstallPlanChecker())
+			//_, err = fetchSubscription(crc, source.GetNamespace(), subscriptionName, subscriptionHasCurrentCSV(csvName))
```
stevekuznetsov, line 693: leftover comment?

test/e2e/catalog_e2e_test.go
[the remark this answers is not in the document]
tmshort: I agree re: operand order. The problem here is that there are WAY MORE calls to `fetchCSV` than to `awaitCSV` which is why `s/awaitCSV/fetchCSV/g` occurred, rather than the other way around.

test/e2e/csv_e2e_test.go
tmshort: This was a preexisting function; it's just waiting for the deployment to be created.

test/e2e/catalog_e2e_test.go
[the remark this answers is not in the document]
tmshort: Originally:
* awaitCSV: 49
* fetchCSV: 113

test/e2e/csv_e2e_test.go
tmshort: Would you prefer `time.poll` then? `Eventually()` has some odd behaviors, admittedly.

test/e2e/csv_e2e_test.go
stevekuznetsov: Did you remove one of them, or is there a change to the code within?

tmshort: Yes to both. `awaitCSV` is gone. `fetchCSV` is more robust.

test/e2e/catalog_e2e_test.go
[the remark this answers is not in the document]
tmshort: `sed` is my friend.

test/e2e/csv_e2e_test.go
tmshort: Added a return argument, even though it's not used.

test/e2e/csv_e2e_test.go
tmshort: And this was basically a loop around a function that loops.

test/e2e/subscription_e2e_test.go
```diff
@@ -3091,10 +3090,12 @@ func fetchSubscription(crc versioned.Interface, namespace, name string, checker
 	var lastCSV string
 	var lastInstallPlanRef *corev1.ObjectReference
 
-	err = wait.Poll(pollInterval, pollDuration, func() (bool, error) {
+	err := wait.Poll(pollInterval, pollDuration, func() (bool, error) {
+		var err error
 		fetchedSubscription, err = crc.OperatorsV1alpha1().Subscriptions(namespace).Get(context.Background(), name, metav1.GetOptions{})
-		if err != nil || fetchedSubscription == nil {
```
stevekuznetsov, line 3098: removing the check for subscription == nil might mean the following bits can panic as they access innards

tmshort: Possible, but unlikely; if `err == nil` (i.e. success), then `fetchedSubscription` would be set. I can restore the check, but the action might be different.

stevekuznetsov: ack, seems unlikely

stevekuznetsov (approved):

tmshort:
@stevekuznetsov I just updated because I *did* have one place where I checked for both!

tmshort:
@stevekuznetsov, needs LGTM?

stevekuznetsov (approved):

PR #3093, opened 2023-11-06
Fix Image Update test

awgreene:
This change grants the serviceAccount used by the catalogSource permissions to pull from the internal registry.


stevekuznetsov:
/hold cancel

We can sleuth out-of-band by running a build with this commit removed :)

PR #3096, opened 2023-11-07
Modify Image Update Test

awgreene:
This commit reverts a change to the Image Update test and instead updates the test's logic to delete the registry pod if it is unable to pull the image from the internal registry.


awgreene:
/hold
I'd really like to find out the root cause of the failed internal pulls before moving forward on this.

tmshort (commented):
Do any of these returned errors get printed? I think only in the case of the loop terminating, so you don't get to see transient conditions.

PR #3102, opened 2023-11-09
Fix catsrc pod hash logic

awgreene:


pkg/controller/registry/reconciler/grpc.go
```diff
@@ -231,20 +231,27 @@ func (c *GrpcRegistryReconciler) EnsureRegistryServer(catalogSource *v1alpha1.Ca
...
-	// recreate the pod if no existing pod is serving the latest image or correct spec
-	overwritePod := overwrite || len(c.currentPodsWithCorrectImageAndSpec(source, sa.GetName())) == 0
-
 	if err != nil && !apierrors.IsAlreadyExists(err) {
 		return errors.Wrapf(err, "error ensuring service account: %s", source.GetName())
 	}
-	if err := c.ensurePod(source, sa.GetName(), overwritePod); err != nil {
-		return errors.Wrapf(err, "error ensuring pod: %s", source.Pod(sa.Name).GetName())
+
+	// The serviceAccount should have a imagePullSecret before creating the pod
```
stevekuznetsov, line 238: The Pod should have all of the ImagePullSecrets that the ServiceAccount does (which may be none)

pkg/controller/registry/reconciler/reconciler.go
```diff
@@ -192,6 +201,9 @@ func Pod(source *operatorsv1alpha1.CatalogSource, name, opmImg, utilImage, img,
 				"kubernetes.io/os": "linux",
 			},
 			ServiceAccountName: saName,
+			// If this field is not set, there is a chance that the pod is admitted to the cluster by the admission webhook without
+			// the imagePullSecrets defined for it's serviceAccount.
```
stevekuznetsov, line 205: In the race, the service account does not have secrets yet - you are guaranteed to get them if they exist.

pkg/controller/registry/reconciler/reconciler_test.go
```diff
@@ -167,13 +167,22 @@ func TestPodMemoryTarget(t *testing.T) {
...
+func serviceAccount(namespace, name string) *corev1.ServiceAccount {
+	return &corev1.ServiceAccount{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: namespace,
+			Name:      name,
+		},
+	}
+}
+
 func TestPodExtractContent(t *testing.T) {
```
stevekuznetsov, line 186: Can we add a test somewhere in this file for the new behavior of `Pod()`?

pkg/controller/registry/reconciler/grpc.go
```diff
@@ -478,9 +486,23 @@ func (c *GrpcRegistryReconciler) removePods(pods []*corev1.Pod, namespace string
...
+	// returning a IsNotFound error will cause the controller to exit and never create the resources, so we should
+	// only return an error if it is something other than a NotFound error.
+	serviceAccount := source.ServiceAccount()
+	serviceAccount, getError := c.OpClient.GetServiceAccount(serviceAccount.GetNamespace(), serviceAccount.GetName())
+	if getError != nil {
+		healthy = false
+		if !apierrors.IsNotFound(getError) {
+			err = getError
+		}
+		return
```
stevekuznetsov, line 500: Naked named returns are one of Go's silliest features - given that the `healthy` variable only ever exists to return, would be OK in this PR changing this function to do a more normal flow?


```go
if getError != nil {
    if apierrors.IsNotFound(getError) {
        return false, nil
    }
    return false, getError
}

File: pkg/controller/registry/reconciler/reconciler.go (hunk 2 above)
Comment (awgreene) [reply pkg/controller/registry/reconciler/reconciler.go:205]: How about:
```
/// If this field is not set, there that the imagePullSecrets defined in the pod
// and its serviceAccount will be out of sync.
```

File: pkg/controller/registry/reconciler/grpc.go (hunk 1 above)
Comment (awgreene) [reply pkg/controller/registry/reconciler/grpc.go:238]: I think I'm gonna remove this comment, no value add.

File: pkg/controller/registry/reconciler/reconciler.go (hunk 2 above)
Comment (stevekuznetsov) [reply pkg/controller/registry/reconciler/reconciler.go:205]: sure, something like that works

File: pkg/controller/registry/reconciler/reconciler_test.go (hunk 3 above)
Comment (awgreene) [reply pkg/controller/registry/reconciler/reconciler_test.go:187]: Done.

**Comment by stevekuznetsov** [comment]
Er sorry - `TestPodExtractContent` is a specific test of `Pod()` for extracting content. I think you want to add a test case to a general test for `Pod()` or add a new `TestPodServiceAccountImagePullSecrets()`

**Comment by awgreene** [comment]
>Er sorry - TestPodExtractContent is a specific test of Pod() for extracting content. I think you want to add a test case to a general test for Pod() or add a new TestPodServiceAccountImagePullSecrets()

@stevekuznetsov I can move it to its own test suite, but functionally the existing implementation guarantees that the pod is constructed with the proper imagePullSecrets defined by the serviceAccount, is the reason for the change request just to avoid muddying the purpose of a test?

**Comment by stevekuznetsov** [comment]
Yes, that's all.

**Review by stevekuznetsov — DISMISSED** [review]

**Comment by awgreene** [comment]
/unhold

Good to go!

File: pkg/controller/registry/reconciler/reconciler.go [hunk 5]
```diff
@@ -192,6 +201,9 @@ func Pod(source *operatorsv1alpha1.CatalogSource, name, opmImg, utilImage, img,
 				"kubernetes.io/os": "linux",
 			},
 			ServiceAccountName: saName,
+			// If this field is not set, there that the imagePullSecrets defined in the pod
+			// and its serviceAccount will be out of sync.
+			ImagePullSecrets: saImagePullSecrets,
```
awgreene, line 206: This is basically the logic we wanted to introduce to the PR.

PR #3106, opened 2023-11-10
Update Subscription Status with Deprecation Warnings

### LINKED ISSUE #3099: Add 'deprecated' condition to Subscription status
As an OLM user, I would like to know when a Subscription has a deprecated Operator installed.

Acceptance Criteria:

- When the catalog entry used by a Subscription is marked as "deprecated", the "Deprecated" condition will be added to the Subscription status.
- Tests are written

dtfranz:

Adds the `OperatorDeprecated` Condition to the status of Subscriptions when the bundle has been deprecated.

~~Doing this in the step resolver is convenient because:~~
- ~~We have ready access to the cache entries.~~
- ~~Subscription status is already updated in this area.~~
- ~~Deprecated status should be added promptly if a catalog update occurs which adds the property to a bundle.~~

Modifying the subscription status within the resolver was both problematic and risky. We've changed strategies and are now updating status during catalog state update. This carries much lower risk.

**Remaining Tasks:**
- [x] Populate cache entries with deprecation info.
- [x] Add additional deprecation condition types (package, channel, bundle).
- [x] ~~Update logic in catalog operator's `syncResolvingNamespace` to account for subs updated with only `OperatorDeprecated` and do not necessarily require InstallPlan updates in the namespace (currently it assumes all updates are new bundles).~~
- [x] Update test and/or add more tests with additional condition types (as above).
- [x] Update go.mod to use merged/released API updates to `api` and `registry`.
- [x] Squash commits where appropriate.

Closes #3099 
[Brief](https://docs.google.com/document/d/1ipnf5uMiGZh5GXt2ey3oxziFx9lZv0qiFmB8g54EFwQ/edit?usp=sharing)
[RFC](https://docs.google.com/document/d/1SLgERsnNwNN7PuABdzysmxYxzD-hiyh7yuJc1m2Hy98/edit?usp=sharing)

**Architectural changes:**

### DIFF
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -346,6 +346,7 @@ func NewOperator(ctx context.Context, kubeconfigPath string, clock utilclock.Clo
 		subscription.WithAppendedReconcilers(subscription.ReconcilerFromLegacySyncHandler(op.syncSubscriptions, nil)),
 		subscription.WithRegistryReconcilerFactory(op.reconciler),
 		subscription.WithGlobalCatalogNamespace(op.namespace),
+		subscription.WithSourceProvider(resolverSourceProvider),
 	)
 	if err != nil {
 		return nil, err
--- a/pkg/controller/operators/catalog/subscription/config.go
+++ b/pkg/controller/operators/catalog/subscription/config.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/clientset/versioned"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/registry/reconciler"
+	resolverCache "github.com/operator-framework/operator-lifecycle-manager/pkg/controller/registry/resolver/cache"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/kubestate"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/operatorlister"
 )
@@ -26,6 +27,7 @@ type syncerConfig struct {
 	reconcilers               kubestate.ReconcilerChain
 	registryReconcilerFactory reconciler.RegistryReconcilerFactory
 	globalCatalogNamespace    string
+	sourceProvider            resolverCache.SourceProvider
 }
 
 // SyncerOption is a configuration option for a subscription syncer.
@@ -125,6 +127,12 @@ func WithRegistryReconcilerFactory(r reconciler.RegistryReconcilerFactory) Synce
 func WithGlobalCatalogNamespace(namespace string) SyncerOption {
 	return func(config *syncerConfig) {
 		config.globalCatalogNamespace = namespace
+	}
+}
+
+func WithSourceProvider(provider resolverCache.SourceProvider) SyncerOption {
+	return func(config *syncerConfig) {
+		config.sourceProvider = provider
 	}
 }
 
--- a/pkg/controller/operators/catalog/subscription/reconciler.go
+++ b/pkg/controller/operators/catalog/subscription/reconciler.go
@@ -5,8 +5,10 @@ import (
 	"errors"
 	"fmt"
 	"sort"
+	"strings"
 
 	"github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
@@ -18,8 +20,10 @@ import (
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/clientset/versioned"
 	listers "github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/listers/operators/v1alpha1"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/registry/reconciler"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/registry/resolver/cache"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/kubestate"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/queueinformer"
+	"github.com/operator-framework/operator-registry/pkg/api"
 )
 
 // ReconcilerFromLegacySyncHandler returns a reconciler that invokes the given legacy sync handler and on delete funcs.
@@ -57,6 +61,7 @@ type catalogHealthReconciler struct {
 	catalogLister             listers.CatalogSourceLister
 	registryReconcilerFactory reconciler.RegistryReconcilerFactory
 	globalCatalogNamespace    string
+	sourceProvider            cache.SourceProvider
 }
 
 // Reconcile reconciles subscription catalog health conditions.
@@ -84,7 +89,15 @@ func (c *catalogHealthReconciler) Reconcile(ctx context.Context, in kubestate.St
 					break
 				}
 
-				next, err = s.UpdateHealth(c.now(), c.client.OperatorsV1alpha1().Subscriptions(ns), catalogHealth...)
+				var healthUpdated, deprecationUpdated bool
+				next, healthUpdated = s.UpdateHealth(c.now(), catalogHealth...)
+				deprecationUpdated, err = c.updateDeprecatedStatus(ctx, s.Subscription())
+				if err != nil {
+					return next, err
+				}
+				if healthUpdated || deprecationUpdated {
+					_, err = c.client.OperatorsV1alpha1().Subscriptions(ns).UpdateStatus(ctx, s.Subscription(), metav1.UpdateOptions{})
+				}
 			case SubscriptionExistsState:
 				if s == nil {
 					err = errors.New("nil state")
@@ -107,6 +120,119 @@ func (c *catalogHealthReconciler) Reconcile(ctx context.Context, in kubestate.St
 	out = next
 
 	return
+}
+
+// updateDeprecatedStatus adds deprecation status conditions to the subscription when present in the cache entry then
+// returns a bool value of true if any changes to the existing subscription have occurred.
+func (c *catalogHealthReconciler) updateDeprecatedStatus(ctx context.Context, sub *v1alpha1.Subscription) (bool, error) {
+	if c.sourceProvider == nil {
+		return false, nil
+	}
+
+	source, ok := c.sourceProvider.Sources(sub.Namespace)[cache.SourceKey{
+		Name:      sub.Spec.CatalogSource,
+		Namespace: sub.Namespace,
+	}]
+	if !ok {
+		return false, nil
+	}
+	snapshot, err := source.Snapshot(ctx)
+	if err != nil {
+		return false, err
+	}
+	if len(snapshot.Entries) == 0 {
+		return false, nil
+	}
+
+	changed := false
+	rollupMessages := []string{}
+	var deprecations *cache.Deprecations
+
+	found := false
+	for _, entry := range snapshot.Entries {
+		// Find the cache entry that matches this subscription
+		if entry.SourceInfo == nil || entry.Package() != sub.Spec.Package {
+			continue
+		}
+		if sub.Spec.Channel != "" && entry.Channel() != sub.Spec.Channel {
+			continue
+		}
+		if sub.Status.InstalledCSV != entry.Name {
+			continue
+		}
+		deprecations = entry.SourceInfo.Deprecations
+		found = true
+		break
+	}
+	if !found {
+		// No matching entry found
+		return false, nil
+	}
+	conditionTypes := []v1alpha1.SubscriptionConditionType{
+		v1alpha1.SubscriptionPackageDeprecated,
+		v1alpha1.SubscriptionChannelDeprecated,
+		v1alpha1.SubscriptionBundleDeprecated,
+	}
+	for _, conditionType := range conditionTypes {
+		oldCondition := sub.Status.GetCondition(conditionType)
+		var deprecation *api.Deprecation
+		if deprecations != nil {
+			switch conditionType {
+			case v1alpha1.SubscriptionPackageDeprecated:
+				deprecation = deprecations.Package
+			case v1alpha1.SubscriptionChannelDeprecated:
+				deprecation = deprecations.Channel
+			case v1alpha1.SubscriptionBundleDeprecated:
+				deprecation = deprecations.Bundle
+			}
+		}
+		if deprecation != nil {
+			if conditionType == v1alpha1.SubscriptionChannelDeprecated && sub.Spec.Channel == "" {
+				// Special case: If optional field sub.Spec.Channel is unset do not apply a channel
+				// deprecation message and remove them if any exist.
+				sub.Status.RemoveConditions(conditionType)
+				if oldCondition.Status == corev1.ConditionTrue {
+					changed = true
+				}
+				continue
+			}
+			newCondition := v1alpha1.SubscriptionCondition{
+				Type:               conditionType,
+				Message:            deprecation.Message,
+				Status:             corev1.ConditionTrue,
+				LastTransitionTime: c.now(),
+			}
+			rollupMessages = append(rollupMessages, deprecation.Message)
+			if oldCondition.Message != newCondition.Message {
+				// oldCondition's message was empty or has changed; add or update the condition
+				sub.Status.SetCondition(newCondition)
+				changed = true
+			}
+		} else if oldCondition.Status == corev1.ConditionTrue {
+			// No longer deprecated at this level; remove the condition
+			sub.Status.RemoveConditions(conditionType)
+			changed = true
+		}
+	}
+
+	if !changed {
+		// No need to update rollup condition if no other conditions have changed
+		return false, nil
+	}
+	if len(rollupMessages) > 0 {
+		rollupCondition := v1alpha1.SubscriptionCondition{
+			Type:               v1alpha1.SubscriptionDeprecated,
+			Message:            strings.Join(rollupMessages, "; "),
+			Status:             corev1.ConditionTrue,
+			LastTransitionTime: c.now(),
+		}
+		sub.Status.SetCondition(rollupCondition)
+	} else {
+		// No rollup message means no deprecation conditions were set; remove the rollup if it exists
+		sub.Status.RemoveConditions(v1alpha1.SubscriptionDeprecated)
+	}
+
+	return true, nil
 }
 
 // catalogHealth gets the health of catalogs that can affect Susbcriptions in the given namespace.
--- a/pkg/controller/registry/resolver/cache/operators.go
+++ b/pkg/controller/registry/resolver/cache/operators.go
@@ -140,11 +140,18 @@ type OperatorSourceInfo struct {
 	StartingCSV    string
 	Catalog        SourceKey
 	DefaultChannel bool
+	Deprecations   *Deprecations
 	Subscription   *v1alpha1.Subscription
 }
 
 func (i *OperatorSourceInfo) String() string {
 	return fmt.Sprintf("%s/%s in %s/%s", i.Package, i.Channel, i.Catalog.Name, i.Catalog.Namespace)
+}
+
+type Deprecations struct {
+	Package *api.Deprecation
+	Channel *api.Deprecation
+	Bundle  *api.Deprecation
 }
 
 type Entry struct {
--- a/pkg/controller/registry/resolver/source_registry.go
+++ b/pkg/controller/registry/resolver/source_registry.go
@@ -109,7 +109,7 @@ func (s *registrySource) Snapshot(ctx context.Context) (*cache.Snapshot, error) 
 	// Fetching default channels this way makes many round trips
 	// -- may need to either add a new API to fetch all at once,
 	// or embed the information into Bundle.
-	defaultChannels := make(map[string]string)
+	packages := make(map[string]*api.Package)
 
 	it, err := s.client.ListBundles(ctx)
 	if err != nil {
@@ -118,21 +118,39 @@ func (s *registrySource) Snapshot(ctx context.Context) (*cache.Snapshot, error) 
 
 	var operators []*cache.Entry
 	for b := it.Next(); b != nil; b = it.Next() {
-		defaultChannel, ok := defaultChannels[b.PackageName]
+		p, ok := packages[b.PackageName]
 		if !ok {
-			if p, err := s.client.GetPackage(ctx, b.PackageName); err != nil {
+			if p, err = s.client.GetPackage(ctx, b.PackageName); err != nil {
 				s.logger.Printf("failed to retrieve default channel for bundle, continuing: %v", err)
 				continue
 			} else {
-				defaultChannels[b.PackageName] = p.DefaultChannelName
-				defaultChannel = p.DefaultChannelName
+				packages[b.PackageName] = p
 			}
 		}
-		o, err := newOperatorFromBundle(b, "", s.key, defaultChannel)
+		o, err := newOperatorFromBundle(b, "", s.key, p.DefaultChannelName)
 		if err != nil {
 			s.logger.Printf("failed to construct operator from bundle, continuing: %v", err)
 			continue
 		}
+		var deprecations *cache.Deprecations
+		if p.Deprecation != nil {
+			deprecations = &cache.Deprecations{Package: &api.Deprecation{Message: fmt.Sprintf("olm.package/%s: %s", p.Name, p.Deprecation.GetMessage())}}
+		}
+		for _, c := range p.Channels {
+			if c.Name == b.ChannelName && c.Deprecation != nil {
+				if deprecations == nil {
+					deprecations = &cache.Deprecations{}
+				}
+				deprecations.Channel = &api.Deprecation{Message: fmt.Sprintf("olm.channel/%s: %s", c.Name, c.Deprecation.GetMessage())}
+			}
+		}
+		if b.Deprecation != nil {
+			if deprecations == nil {
+				deprecations = &cache.Deprecations{}
+			}
+			deprecations.Bundle = &api.Deprecation{Message: fmt.Sprintf("olm.bundle/%s: %s", b.CsvName, b.Deprecation.GetMessage())}
+		}
+		o.SourceInfo.Deprecations = deprecations
 		o.ProvidedAPIs = o.ProvidedAPIs.StripPlural()
 		o.RequiredAPIs = o.RequiredAPIs.StripPlural()
 		o.Replaces = b.Replaces
--- a/test/e2e/subscription_e2e_test.go
+++ b/test/e2e/subscription_e2e_test.go
@@ -2534,6 +2534,151 @@ var _ = Describe("Subscription", func() {
 				_, err = fetchSubscription(crc, generatedNamespace.GetName(), subName, subscriptionHasCurrentCSV("example-operator.v0.2.0"))
 				Expect(err).Should(BeNil())
 			})
+
+			It("should report deprecation conditions when package, channel, and bundle are referenced in an olm.deprecations object", func() {
+				By("patching the OperatorGroup to reduce the bundle unpacking timeout")
+				ogNN := types.NamespacedName{Name: operatorGroup.GetName(), Namespace: generatedNamespace.GetName()}
+				addBundleUnpackTimeoutOGAnnotation(context.Background(), ctx.Ctx().Client(), ogNN, "5m")
+
+				By("updating the catalog with a fixed v0.2.0 bundle image marked deprecated")
+				provider, err := NewFileBasedFiledBasedCatalogProvider(filepath.Join(testdataDir, subscriptionTestDataBaseDir, "example-operator.v0.2.0-deprecations.yaml"))
+				Expect(err).To(BeNil())
+				err = magicCatalog.UpdateCatalog(context.Background(), provider)
+				Expect(err).To(BeNil())
+
+				By("waiting for the subscription to have v0.2.0 installed with a Bundle Deprecated condition")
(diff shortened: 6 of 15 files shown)


pkg/controller/registry/resolver/step_resolver.go
```diff
@@ -193,14 +193,32 @@ func (r *OperatorStepResolver) ResolveSteps(namespace string) ([]*v1alpha1.Step,
 			})
 		}
 
-		// add steps for subscriptions for bundles that were added through resolution
+		// add steps for subscriptions for bundles that were added through resolution and bundles that have been marked deprecated
 		for sub := range existingSubscriptions {
-			if sub.Status.CurrentCSV == op.Name {
-				continue
+			updated := false
+			if len(op.Deprecations) == 0 && corev1.ConditionStatus(sub.Status.GetCondition(v1alpha1.SubscriptionOperatorDeprecated).Type) != corev1.ConditionUnknown {
```
joelanford, line 199: `sub.Status.GetCondition` can return `nil`, so this will panic when trying to reference `.Type`. But separately, why not remove the condition if it is `Unknown`?

joelanford, line 199: This seems like it deserves a comment since it is non-obvious that `GetCondition` will return a populated condition set to `Unknown` when the type is not found.

However, do we even need the `Unknown` check? Can we just always remove it if there are no deprecations?

pkg/controller/registry/resolver/step_resolver.go
```diff
@@ -193,14 +193,32 @@ func (r *OperatorStepResolver) ResolveSteps(namespace string) ([]*v1alpha1.Step,
...
+				updated = true
+			} else if len(op.Deprecations) > 0 {
+				// Remove old conditions first in case of update
+				sub.Status.RemoveConditions(v1alpha1.SubscriptionOperatorDeprecated)
+				for _, deprecation := range op.Deprecations {
+					sub.Status.SetCondition(v1alpha1.SubscriptionCondition{
+						Type:    v1alpha1.SubscriptionOperatorDeprecated,
+						Message: deprecation.Message,
+					})
+				}
```
joelanford, line 211: It looks like the result of this will be that the resulting subscription conditions will contain just the last deprecation from `op.Deprecations`. If `type` already exists in the conditions slice `SetCondition` overwrites it.

pkg/controller/registry/resolver/step_resolver.go
dtfranz: Yeah this func is a bit odd in how it behaves. We could also use a more generic func for this op but I believe we will have to do conversions to a generic condition type.

The reason I did the check was primarily to avoid doing a no-op removal then adding the subscription to the list of `updatedSubs` when potentially nothing has changed. If we don't care about that though we can just do it regardless.

I'll also note however I did check `Type` here instead of `Status` so I will fix that :+1:

pkg/controller/registry/resolver/step_resolver.go
dtfranz: Yes, this part is in a messy in-between state because it's behaving as though we're going to add each deprecation with a different type (package, channel, bundle) but the separate types aren't in there yet so it just overwrites. I'll go ahead and create another `api` PR for the additional condition types then I can clean this up.

pkg/controller/operators/catalog/operator.go
```diff
@@ -1181,6 +1181,7 @@ func (o *Operator) syncResolvingNamespace(obj interface{}) error {
 	// TODO: parallel
 	maxGeneration := 0
 	subscriptionUpdated := false
+	subsSet := sets.New[string]()
```
joelanford, line 1184: Could we make this a set of a [`types.NamespacedName`](https://pkg.go.dev/k8s.io/apimachinery/pkg/types#NamespacedName) instead?
```suggestion
	subsSet := sets.New[types.NamespacedName]()
```

That way we don't have to worry about string concatenation resulting in inadvertent key reuse. E.g.
- `namepace: foo, name: bar-bazz` => `key: foo-bar-bazz`
- `namespace: foo-bar, name: bazz` => `key: foo-bar-bazz`

pkg/controller/operators/catalog/operator.go
```diff
@@ -1341,8 +1344,31 @@ func (o *Operator) syncResolvingNamespace(obj interface{}) error {
 		}
 	}
 
+	newInstallPlanSubs := map[string]*v1alpha1.Subscription{}
+	statusUpdatedSubs := map[string]*v1alpha1.Subscription{}
```
joelanford, line 1348: Same comment here about using `types.NamespacedName` as the key.

pkg/controller/registry/resolver/cache/operators.go
```diff
@@ -158,6 +164,9 @@ type Entry struct {
 	SourceInfo   *OperatorSourceInfo
 	Properties   []*api.Property
 	BundlePath   string
+	// Since there are separate cache entries for the same bundle in different channels,
+	// This should contain at most one deprecation for package, channel and bundle.
+	Deprecations []Deprecation
```
joelanford, line 169: Two things:
1. An entry is already specific to a package, channel, and bundle version, right? (looking at `SourceInfo`, that seems to be the case). Given that, it seems like we could make `Deprecations` be a single struct that has pointer fields for `Package`, `Channel`, and `Bundle` `api.Deprecation` structs, and then we don't need strings to encode `Type` and `Name`?
2. Would it make sense to put this inside the `OperatorSourceInfo` struct, since that already seems to have information related the package and channel catalog metadata?

pkg/controller/registry/resolver/source_registry.go
```diff
@@ -10,6 +10,7 @@ import (
 	"github.com/blang/semver/v4"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/registry"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/registry/resolver/cache"
+	"github.com/operator-framework/operator-registry/alpha/declcfg"
```
joelanford, line 13: Let's avoid this dependency. OLMv0 is unaware of FBC concepts/schemas, etc.

Like I said in another comment, I think we can just have `Package`, `Channel` and `Bundle`/`CSVName` fields for each deprecation layer.

pkg/controller/registry/resolver/step_resolver.go
```diff
@@ -193,14 +194,56 @@ func (r *OperatorStepResolver) ResolveSteps(namespace string) ([]*v1alpha1.Step,
...
+							Type:    v1alpha1.SubscriptionPackageDeprecated,
+							Message: deprecation.Message,
+							Status:  corev1.ConditionTrue,
+						})
+					case declcfg.SchemaChannel:
+						sub.Status.SetCondition(v1alpha1.SubscriptionCondition{
+							Type:    v1alpha1.SubscriptionChannelDeprecated,
+							Message: deprecation.Message,
+							Status:  corev1.ConditionTrue,
+						})
```
joelanford, line 221: It looks like `subscription.spec.channel` is an optional field. We should only set the channel deprecation condition if `subscription.spec.channel` is set and that channel is deprecated.

pkg/controller/registry/resolver/step_resolver.go
```diff
@@ -193,14 +194,56 @@ func (r *OperatorStepResolver) ResolveSteps(namespace string) ([]*v1alpha1.Step,
...
+					}
+					rollupMessage = fmt.Sprintf("%s; %s", rollupMessage, deprecation.Message)
+				}
+				sub.Status.SetCondition(v1alpha1.SubscriptionCondition{
+					Type:    v1alpha1.SubscriptionDeprecated,
+					Message: rollupMessage,
+					Status:  corev1.ConditionTrue,
+				})
+				updated = true
```
joelanford, line 236: If the sub already has this existing set of deprecations and none of the messages changed, it seems like we should not set `updated = true`?

pkg/controller/registry/resolver/cache/operators.go
dtfranz: Thanks for suggesting this, I think it cleans things up a lot, particularly when constructing the conditions and detecting required changes.

pkg/controller/registry/resolver/step_resolver.go
dtfranz: Yes, I'm updating this to detect if a change occurs and only update when required.

pkg/controller/operators/catalog/subscription/reconciler.go
```diff
@@ -84,7 +89,14 @@ func (c *catalogHealthReconciler) Reconcile(ctx context.Context, in kubestate.St
 					break
 				}
 
+				// TODO: reduce this to one update call
 				next, err = s.UpdateHealth(c.now(), c.client.OperatorsV1alpha1().Subscriptions(ns), catalogHealth...)
+				if err != nil {
+					return out, err
+				}
+				if c.sourceProvider != nil {
+					err = updateDeprecatedStatus(s.Subscription(), c.now(), c.client.OperatorsV1alpha1().Subscriptions(ns), c.sourceProvider.Sources(s.Subscription().GetNamespace()))
```
everettraven, line 98: Reuse the `ns` variable here as well?
```suggestion
					err = updateDeprecatedStatus(s.Subscription(), c.now(), c.client.OperatorsV1alpha1().Subscriptions(ns), c.sourceProvider.Sources(ns))
```

pkg/controller/operators/catalog/subscription/reconciler.go
```diff
@@ -109,6 +121,104 @@ func (c *catalogHealthReconciler) Reconcile(ctx context.Context, in kubestate.St
...
 
+// updateDeprecatedStatus adds deprecation status conditions to the subscription when present in the cache entry
+func updateDeprecatedStatus(sub *v1alpha1.Subscription, time *metav1.Time, client clientv1alpha1.SubscriptionInterface, source map[cache.SourceKey]cache.Source) error {
+	changed := false
+	rollupMessage := ""
+	sourceKey := cache.SourceKey{
+		Name:      sub.Spec.CatalogSource,
+		Namespace: sub.Namespace,
+	}
+	entries, err := source[sourceKey].Snapshot(context.Background())
```
everettraven, line 132: Nit: to avoid the potentially confusing `entries.Entries` usage below and from the `.Snapshot()` function returning `(cache.Snapshot, err)` instead of `([]cache.Entry, err)`
```suggestion
	snapshot, err := source[sourceKey].Snapshot(context.Background())
```

test/e2e/subscription_e2e_test.go
```diff
@@ -2534,6 +2534,147 @@ var _ = Describe("Subscription", func() {
...
+				Expect(err).To(BeNil())
+				err = magicCatalog.UpdateCatalog(context.Background(), provider)
+				Expect(err).To(BeNil())
+
+				By("waiting for the subscription to have v0.3.0 installed")
+				sub, err = fetchSubscription(crc, generatedNamespace.GetName(), subName, subscriptionHasCurrentCSV("example-operator.v0.3.0"))
+				Expect(err).Should(BeNil())
+
+				By("checking for the deprecated conditions")
+				// Operators is deprecated at all three levels in the catalog
```
everettraven, line 2622: This comment doesn't align with the code

test/e2e/subscription_e2e_test.go
```diff
@@ -2534,6 +2534,147 @@ var _ = Describe("Subscription", func() {
...
+				packageCondition := sub.Status.GetCondition(operatorsv1alpha1.SubscriptionPackageDeprecated)
+				Expect(packageCondition.Status).To(Equal(corev1.ConditionTrue))
+				channelCondition := sub.Status.GetCondition(operatorsv1alpha1.SubscriptionChannelDeprecated)
+				Expect(channelCondition.Status).To(Equal(corev1.ConditionTrue))
+				bundleCondition = sub.Status.GetCondition(operatorsv1alpha1.SubscriptionBundleDeprecated)
+				Expect(bundleCondition.Status).To(Equal(corev1.ConditionUnknown))
+
+				By("verifying that a roll-up condition is present not containing bundle deprecation condition")
+				// Roll-up condition should be present and contain deprecation messages from all three levels
```
everettraven, line 2631: Same here

pkg/controller/operators/catalog/subscription/config.go
```diff
@@ -1,17 +1,17 @@
 package subscription
 
 import (
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/clientset/versioned"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/registry/reconciler"
+	resolverCache "github.com/operator-framework/operator-lifecycle-manager/pkg/controller/registry/resolver/cache"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/kubestate"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/operatorlister"
```
awgreene, line 8: Typically we keep imports groups in three sets:
- Go packages
- third party packages
- repo packages

pkg/controller/operators/catalog/subscription/reconciler.go
```diff
@@ -84,7 +89,12 @@ func (c *catalogHealthReconciler) Reconcile(ctx context.Context, in kubestate.St
 					break
 				}
 
+				// TODO: reduce this to one update call
```
awgreene, line 92: +1

pkg/controller/operators/catalog/subscription/reconciler.go
```diff
@@ -84,7 +89,12 @@ func (c *catalogHealthReconciler) Reconcile(ctx context.Context, in kubestate.St
...
 				next, err = s.UpdateHealth(c.now(), c.client.OperatorsV1alpha1().Subscriptions(ns), catalogHealth...)
+				if err != nil {
+					return out, err
```
awgreene, line 95: I think this is a named return value

pkg/controller/operators/catalog/subscription/reconciler.go
```diff
@@ -109,6 +119,114 @@ func (c *catalogHealthReconciler) Reconcile(ctx context.Context, in kubestate.St
...
+func (c *catalogHealthReconciler) updateDeprecatedStatus(ctx context.Context, sub *v1alpha1.Subscription) error {
+	if c.sourceProvider == nil {
+		return nil
+	}
+
+	sourceKey := cache.SourceKey{
+		Name:      sub.Spec.CatalogSource,
+		Namespace: sub.Namespace,
+	}
+	source, ok := c.sourceProvider.Sources(sub.Namespace)[sourceKey]
```
awgreene, line 132:
```suggestion
	source, ok := c.sourceProvider.Sources(sub.Namespace)[cache.SourceKey{
		Name:      sub.Spec.CatalogSource,
		Namespace: sub.Namespace,
	}]
```

pkg/controller/operators/catalog/subscription/reconciler.go
```diff
@@ -109,6 +119,114 @@ func (c *catalogHealthReconciler) Reconcile(ctx context.Context, in kubestate.St
...
+	if !ok {
+		return nil
+	}
+	snapshot, err := source.Snapshot(ctx)
+	if err != nil {
+		return err
+	}
+	if len(snapshot.Entries) == 0 {
+		return nil
+	}
```
awgreene, line 142:
```suggestion
        // return err if it isn't nil or err (which is nil) if the entries is an empty list.
	if err != nill || len(snapshot.Entries) == 0 {
		return err
	}
```

pkg/controller/operators/catalog/subscription/reconciler.go
```diff
@@ -109,6 +119,114 @@ func (c *catalogHealthReconciler) Reconcile(ctx context.Context, in kubestate.St
...
+	var deprecations *cache.Deprecations
+	for _, entry := range snapshot.Entries {
+		// Find the cache entry that matches this subscription
+		if entry.SourceInfo == nil {
+			continue
+		}
+		if entry.Package() == sub.Spec.Package {
+			deprecations = &entry.SourceInfo.Deprecations
+			break
+		}
```
awgreene, line 155:
```suggestion
		if entry.SourceInfo == nil || entry.Package() != sub.Spec.Package {
			continue
		}
		deprecations = &entry.SourceInfo.Deprecations
		break
```

pkg/controller/registry/resolver/source_registry.go
```diff
@@ -109,7 +109,7 @@ func (s *registrySource) Snapshot(ctx context.Context) (*cache.Snapshot, error)
 	// Fetching default channels this way makes many round trips
 	// -- may need to either add a new API to fetch all at once,
 	// or embed the information into Bundle.
-	defaultChannels := make(map[string]string)
+	packages := make(map[string]*api.Package)
```
awgreene, line 113: So these changes to the resolver packages are still required to make the deprecation information available in the entry, right?

test/e2e/subscription_e2e_test.go
```diff
@@ -2534,6 +2534,147 @@ var _ = Describe("Subscription", func() {
 				_, err = fetchSubscription(crc, generatedNamespace.GetName(), subName, subscriptionHasCurrentCSV("example-operator.v0.2.0"))
 				Expect(err).Should(BeNil())
 			})
+
+			It("should report deprecation conditions when package, channel, and bundle are referenced in an olm.deprecations object", func() {
```
awgreene, line 2538: Great test.

awgreene (commented):
Great work @dtfranz!

pkg/controller/operators/catalog/subscription/reconciler.go
dtfranz: Updated!

pkg/controller/registry/resolver/source_registry.go
dtfranz: That's right, these are still needed.

pkg/controller/operators/catalog/subscription/reconciler.go
joelanford: IMO, this suggestion makes the code less readable because readers would need to track the value of `err` (or trust the comment) to understand that `return err` is sometimes actually `return nil`. I'm not sure the reduced LoC makes up for the decrease in readability.

pkg/controller/operators/catalog/subscription/reconciler.go
```diff
@@ -84,7 +89,15 @@ func (c *catalogHealthReconciler) Reconcile(ctx context.Context, in kubestate.St
...
 
-				next, err = s.UpdateHealth(c.now(), c.client.OperatorsV1alpha1().Subscriptions(ns), catalogHealth...)
+				var healthUpdated, deprecationUpdated bool
+				next, healthUpdated = s.UpdateHealth(c.now(), catalogHealth...)
+				deprecationUpdated, err = c.updateDeprecatedStatus(ctx, s.Subscription())
+				if err != nil {
+					return
+				}
+				if healthUpdated || deprecationUpdated {
+					_, err = c.client.OperatorsV1alpha1().Subscriptions(ns).UpdateStatus(context.TODO(), s.Subscription(), metav1.UpdateOptions{})
```
everettraven, line 99: It looks like this function has a context in it's parameters. Would probably be better to use that here instead of `context.TODO()`

awgreene (approved):
Nice work on this @dtfranz

pkg/controller/operators/catalog/subscription/reconciler.go
```diff
@@ -109,6 +122,105 @@ func (c *catalogHealthReconciler) Reconcile(ctx context.Context, in kubestate.St
 	return
 }
 
+// updateDeprecatedStatus adds deprecation status conditions to the subscription when present in the cache entry
+func (c *catalogHealthReconciler) updateDeprecatedStatus(ctx context.Context, sub *v1alpha1.Subscription) (bool, error) {
+	if c.sourceProvider == nil {
+		return false, nil
+	}
```
everettraven, line 129: Should we have some kind of logging or return a non-nil error here? I can see it being useful to look at the logs and get some insights for debugging any issues with deprecation status not being set.

My same question applies to all locations where we return a nil error.

Without more familiarity with the overall flow that OLM takes here I would lean more towards some kind of logging and keeping the nil error returns, but if anyone more familiar with this logic thinks it is better to do non-nil error returns I'm happy with that as well.

joelanford: Do we ever expect `c.sourceProvider` to be `nil`? If not, I'd suggest a `panic` here.

pkg/controller/operators/catalog/subscription/reconciler.go
```diff
@@ -109,6 +122,105 @@ func (c *catalogHealthReconciler) Reconcile(ctx context.Context, in kubestate.St
...
+
+	changed := false
+	rollupMessages := []string{}
+	var deprecations *cache.Deprecations
+	for _, entry := range snapshot.Entries {
+		// Find the cache entry that matches this subscription
+		if entry.SourceInfo == nil || entry.Package() != sub.Spec.Package {
+			continue
+		}
```
joelanford, line 151: If `sub.Spec.Channel` is set, we need to also make sure the `entry` is for that channel.

pkg/controller/registry/resolver/cache/operators.go
```diff
@@ -140,13 +140,20 @@ type OperatorSourceInfo struct {
 	StartingCSV    string
 	Catalog        SourceKey
 	DefaultChannel bool
+	Deprecations   Deprecations
```
joelanford, line 143: Also make this a pointer?
```suggestion
	Deprecations   *Deprecations
```

pkg/controller/operators/catalog/subscription/reconciler.go
```diff
@@ -109,6 +122,105 @@ func (c *catalogHealthReconciler) Reconcile(ctx context.Context, in kubestate.St
...
+		deprecations = &entry.SourceInfo.Deprecations
+		break
+	}
+	if deprecations == nil {
+		// No matching entry found
+		return false, nil
+	}
```
joelanford, line 158: In all of the cases that we `return false, nil` here and up, that means that whatever the subscription currently says is what we continue going with, right?

I wonder if we need to specifically set the subscription deprecation conditions to `Unknown` in some (or all?) of these cases?

test/e2e/subscription_e2e_test.go
```diff
@@ -2534,6 +2534,147 @@ var _ = Describe("Subscription", func() {
...
+				By("verifying that a roll-up condition is present not containing bundle deprecation condition")
+				// Roll-up condition should be present and contain deprecation messages from Package and Channel levels
+				rollUpCondition := sub.Status.GetCondition(operatorsv1alpha1.SubscriptionDeprecated)
+				Expect(rollUpCondition.Status).To(Equal(corev1.ConditionTrue))
+				Expect(rollUpCondition.Message).To(ContainSubstring(packageCondition.Message))
+				Expect(rollUpCondition.Message).To(ContainSubstring(channelCondition.Message))
+				Expect(rollUpCondition.Message).ToNot(ContainSubstring(bundleDeprecatedMessage))
+			})
+
+			It("should report deprecated status when catalog is updated to deprecate an installed bundle", func() {
```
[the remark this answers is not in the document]
dtfranz: This test case differs in that it runs a catalog update which does not result in an operator upgrade. I left it because before we changed our strategy to update status in the Subscription reconciler this test case would fail because the cache entries of installed bundles did not contain deprecation info.

pkg/controller/operators/catalog/subscription/reconciler.go
everettraven: I'm personally fine with the nil check, but I don't like the idea of this "failing" silently. I'd at the least have a log message.

test/e2e/subscription_e2e_test.go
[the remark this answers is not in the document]
everettraven: Ah, gotcha. Thanks for the additional context!

pkg/controller/operators/catalog/subscription/reconciler.go
dtfranz: @everettraven Yes I think you're correct regarding Joe's meaning; however i don't think that setting these conditions to `Unknown` is necessary. It adds additional complexity to this logic which is already getting somewhat complex, and additionally in my mind it's not even necessarily correct.

If an installed bundle is deprecated, then it will be marked as such on the cluster. If that bundle is then removed from the catalog, isn't the bundle still deprecated? The only situation I can think of where this might be useful is a package/channel/bundle is marked deprecated in a catalog, removed from the catalog, but then they decide later that it should not have been deprecated, but do not put it back in the catalog.

It would also require some refactoring because the Subscription Status' `GetCondition` method relies on the `Unknown` Condition to work correctly (although it shouldn't because of this kind of situation).

If I'm misunderstanding here definitely let me know though!

pkg/controller/operators/catalog/subscription/reconciler.go
dtfranz: Since we don't at this moment know what kind of logging impact there would be from logging here, we will take this as a follow-up.

pkg/controller/operators/catalog/subscription/reconciler.go
dtfranz: Updated; I've made sure that the cache entry matches what is installed by the Subscription completely.

pkg/controller/operators/catalog/subscription/reconciler.go
joelanford: Yes, sorry I was suggesting using `status: Unknown` on the existing types, not suggesting a new types.

I'm actually curious what OLMv0 does with subscriptions that have been successfully installed, but then later have those that subscription's package, channel, or bundle removed from the catalog. Do we surface that information somehow? I assume at a minimum we'd somehow say, "couldn't resolve this subscription anymore".

But regardless, I'm also concerned about having "sticky" state that depends on what happened in the past. Consider two users (A and B) and three versions of a catalog:
1. Has bundle foo.v1 with no deprecation
2. Has bundle foo.v1 with a deprecation
3. Removes bundle foo.v1 entirely

If user A sees version 1, 2, and 3 of the catalog, their subscription will be stuck on "foo.v1 is deprecated"
If user B sees only versions 1 and 3 of the catalog, their subscription will be stuck on  "foo.v1 is not deprecated"

That strikes me as (a) undesirable in general and (b) totally possible, especially with disconnected clusters which often pick up catalog changes on some infrequent basis and do heavy catalog pruning during the mirroring step to minimize time/size of the transfer.

everettraven: I agree with @dtfranz here about the unknown status being ripe for confusion and solving the "sticky" state as a follow up.

For me this raises the question of whether or not we have thought of best practices/recommendations for catalog maintainers to follow? It strikes me as odd that we are attempting to solve for the many edge cases around how someone might create and modify their catalog and the versioning (after all it _is_ an OCI image and we don't have any control over how users make changes to it) instead of coming up with a set of guidelines for how OLM interacts with the catalogs and what "expectations" are in place. This way we can note that if a catalog maintainer deviates from the recommendations/best practices things may not function as expected due to the assumptions/expectations that have been made in OLM to provide the best user experience possible. No need to have this conversation in this PR, but something I would definitely be interested in having a continued discussion about as I think it also affects the OLMv1 work.

pkg/controller/operators/catalog/subscription/reconciler.go
awgreene: Could we create an issue to investigate the logging and test changes discussed in this thread?

pkg/controller/operators/catalog/subscription/reconciler.go
joelanford: I'm suppose I'm okay with existing behavior since my scenario is an edge case (though not a total edge case because it _is_ expected for a deprecated bundle to eventually be removed from the catalog)

Can we capture a follow-up issue that captures both of the following?
- what we do generally to inform users that their currently installed stuff is no longer present in the catalog.
- specifically, what the implications are for the new deprecation condition types.

pkg/controller/operators/catalog/subscription/reconciler.go
```diff
@@ -109,6 +122,118 @@ func (c *catalogHealthReconciler) Reconcile(ctx context.Context, in kubestate.St
 	return
 }
 
+// updateDeprecatedStatus adds deprecation status conditions to the subscription when present in the cache entry
```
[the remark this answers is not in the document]
dtfranz: Thanks for catching, updated!

pkg/controller/operators/catalog/subscription/reconciler.go
dtfranz: Thank you @joelanford! I've created issue #3122 to continue this discussion.

awgreene: >I'm actually curious what OLMv0 does with subscriptions that have been successfully installed, but then later have thier package, channel, or bundle removed from the catalog. Do we surface that information somehow? I assume at a minimum we'd somehow say, "couldn't resolve this subscription anymore".

@joelanford I believe that the following is correct, but I'll rely on @kevinrizza to keep me honest:

- in the case of a package or channel removal, the resolution failure is surfaced in the subscription.
- the removed bundle case is more nuanced. Resolution will succeed if the version falls within the skip-range of a separate bundle. The bundle shouldn't have been removed if replaces or skips was used, so it'd be the case of an invalid catalog, the resolution would fail and be surfaced in the subscription.

>But regardless, I'm also concerned about having "sticky" state that depends on what happened in the past. >Consider two users (A and B) and three versions of a catalog:

    > Has bundle foo.v1 with no deprecation
    > Has bundle foo.v1 with a deprecation
    > Removes bundle foo.v1 entirely

>If user A sees version 1, 2, and 3 of the catalog, their subscription will be stuck on "foo.v1 is deprecated"
>If user B sees only versions 1 and 3 of the catalog, their subscription will be stuck on "foo.v1 is not deprecated"

>That strikes me as (a) undesirable in general and (b) totally possible, especially with disconnected clusters which often pick up catalog changes on some infrequent basis and do heavy catalog pruning during the mirroring step to minimize time/size of the transfer.

Agreed, worth tracking in an issue as suggested in this thread.

PR #3115, opened 2023-11-16
*: use a stable hash for client-side short-circuits

stevekuznetsov:
The previous approach here was to use go-spew, which, unfortunately, is not a stable and germane encoding for objects. We are encoding objects that we send to the Kubernetes API server, so features of go-spew like printing the capacity of slices are not germane. This led to hashes of otherwise identical (as far as the server was concerned) objects to be determined as hashing differently.

### DIFF
--- a/pkg/controller/registry/reconciler/grpc.go
+++ b/pkg/controller/registry/reconciler/grpc.go
@@ -3,12 +3,17 @@ package reconciler
 import (
 	"context"
 	"fmt"
-	"hash/fnv"
 	"strings"
 	"time"
 
 	"github.com/google/go-cmp/cmp"
+	"github.com/operator-framework/api/pkg/operators/v1alpha1"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/install"
+	controllerclient "github.com/operator-framework/operator-lifecycle-manager/pkg/lib/controller-runtime/client"
+	hashutil "github.com/operator-framework/operator-lifecycle-manager/pkg/lib/kubernetes/pkg/util/hash"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/operatorclient"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/operatorlister"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/ownerutil"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
@@ -16,14 +21,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/intstr"
-	"k8s.io/apimachinery/pkg/util/rand"
-
-	"github.com/operator-framework/api/pkg/operators/v1alpha1"
-	controllerclient "github.com/operator-framework/operator-lifecycle-manager/pkg/lib/controller-runtime/client"
-	hashutil "github.com/operator-framework/operator-lifecycle-manager/pkg/lib/kubernetes/pkg/util/hash"
-	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/operatorclient"
-	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/operatorlister"
-	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/ownerutil"
 )
 
 const (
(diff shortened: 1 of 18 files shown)


pkg/lib/kubernetes/pkg/util/hash/hash.go
```diff
@@ -17,21 +17,36 @@ limitations under the License.
...
+
+	raw, err := json.Marshal(obj)
+	if err != nil {
+		return "", fmt.Errorf("couldn't encode object: %w", err)
 	}
-	printer.Fprintf(hasher, "%#v", objectToWrite)
+
+	// base62(sha224(bytes)) is a useful hash and encoding for adding the contents of this
+	// to a Kubernetes identifier or other field which has length and character set requirements
+	hash := sha256.Sum224(raw)
```
joelanford, line 47: We could cut out the extra `raw` byte buffer if we use `json.NewEncoder(hash)`, I think?

pkg/controller/registry/reconciler/grpc.go
```diff
@@ -237,9 +247,14 @@ func correctImages(source grpcCatalogSourceDecorator, pod *corev1.Pod) bool {
 // EnsureRegistryServer ensures that all components of registry server are up to date.
 func (c *GrpcRegistryReconciler) EnsureRegistryServer(logger *logrus.Entry, catalogSource *v1alpha1.CatalogSource) error {
 	source := grpcCatalogSourceDecorator{CatalogSource: catalogSource, createPodAsUser: c.createPodAsUser, opmImage: c.opmImage, utilImage: c.utilImage}
+	fmt.Printf("EnsureRegistryServer source.Status.RegistryServiceStatus: %#v\n", source.CatalogSource.Status.RegistryServiceStatus)
```
joelanford, line 250: Remove this line.

pkg/controller/registry/reconciler/grpc_test.go
```diff
@@ -348,6 +348,7 @@ func TestGrpcRegistryReconciler(t *testing.T) {
 			factory, client := fakeReconcilerFactory(t, stopc, withNow(now), withK8sObjs(tt.in.cluster.k8sObjs...), withK8sClientOptions(clientfake.WithNameGeneration(t)))
 			rec := factory.ReconcilerForSource(tt.in.catsrc)
 
+			fmt.Printf("test source.Status.RegistryServiceStatus: %#v\n", tt.in.catsrc.Status.RegistryServiceStatus)
```
joelanford, line 351: Remove this line

pkg/controller/registry/resolver/rbac.go
```diff
@@ -4,31 +4,29 @@ import (
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 const maxNameLength = 63
 
-func generateName(base string, o interface{}) string {
-	hasher := fnv.New32a()
-	hashutil.DeepHashObject(hasher, o)
-	hash := utilrand.SafeEncodeString(fmt.Sprint(hasher.Sum32()))
+func generateName(base string, o interface{}) (string, error) {
```
joelanford, line 20: Oh one other thing comes to mind here. I feel like I've seen users ask us about RBAC naming and whether it is something they can depend on, potentially programmatically.

It seems unreasonable to depend on something like this, but I want to make sure that we haven't made any promises about it.

Maybe @kevinrizza or @awgreene remembers details?

stevekuznetsov: I don't imagine the new approach will be any more or less dependable than the previous one.

pkg/lib/kubernetes/pkg/util/hash/hash.go
```diff
@@ -17,21 +17,39 @@ limitations under the License.
...
+	// we often need to operate on snippets. Checking with the experts and the implementation,
+	// we can see that the serializers are a thin wrapper over json.Marshal for encoding:
+	// https://github.com/kubernetes/kubernetes/blob/8509ab82b96caa2365552efa08c8ba8baf11c5ec/staging/src/k8s.io/apimachinery/pkg/runtime/serializer/json/json.go#L216-L247
+	// Therefore, we can be confident that using json.Marshal() here will:
+	//  1. be stable & idempotent - the library sorts keys, etc.
+	//  2. be germane to our needs - only fields that serialize and are sent to the server
+	//     will be encoded
+
+	hasher := sha256.New224()
 	hasher.Reset()
```
tmshort, line 41: Nit: Not sure a `Reset()` is needed immediately after a `New224()`.

stevekuznetsov: `sha256.Sum224()` does it after allocating a new digest, so I figured I'd be good to follow their lead:

```go
// Sum224 returns the SHA224 checksum of the data.
func Sum224(data []byte) [Size224]byte {
	if boring.Enabled {
		return boring.SHA224(data)
	}
	var d digest
	d.is224 = true
	d.Reset()
	d.Write(data)
	sum := d.checkSum()
	ap := (*[Size224]byte)(sum[:])
	return *ap
}
```

PR #3110, opened 2023-11-14
PackageManifests Surface Deprecations

awgreene:
Demo:
[asciicast](https://asciinema.org/a/xlDhS5xd8ARTqbJ9MVboIPIlK)

CatalogSource.yaml used in the demo:
```
apiVersion: operators.coreos.com/v1alpha1
kind: CatalogSource
metadata:
  name: deprecated-catalog
  namespace: olm
spec:
  displayName: Deprecated Operators
  grpcPodConfig:
    securityContextConfig: restricted
  image: docker.io/grokspawn/deprecation-catalog
  publisher: OperatorHub.io
  sourceType: grpc
  updateStrategy:
    registryPoll:
      interval: 60m
```

PR #3116, opened 2023-11-16
Bump o-f/operator-registry and o-f/api dependencies

awgreene:


awgreene:
Created to check if this broke the unit test suite which is surfacing in https://github.com/operator-framework/operator-lifecycle-manager/pull/3110

PR #3133, opened 2023-12-18
deploy: use Recreate strategy for Deployments

stevekuznetsov:
deploy: use Recreate strategy for Deployments

Our controllers must not share runtime with controllers of previous
versions, as we expect that the set of actors on the cluster to be
coherent with respect to what they're trying to do. RollingUpdate
strategies do not guarantee this and lead to cases where previous
versions of the operators issue spurious mutating calls that the
current operator needs to un-do.

---


ncdc (approved):

everettraven (approved):
/lgtm

ncdc:
Manually merging to get past CI flakes.

PR #3137, opened 2023-12-20
test/e2e: emit helpful comments as logs

stevekuznetsov:


stevekuznetsov:
```
   It looks like you are calling By outside of a running spec.  Make sure you
  call By inside a runnable node such as It or BeforeEach and not inside the
  body of a container such as Describe or Context.
```

stevekuznetsov:
:eyeroll:

ncdc (approved):

PR #3144, opened 2024-01-03
[chore] Update google.golang.org/grpc to v1.59.0

tmshort:
require grpc@v1.59.0 was replaced by v1.40.0, which is now out-of-date.

This will allow additional dependency updates.

**Reviewer Checklist**


go.sum
```diff
@@ -1209,8 +1225,29 @@ google.golang.org/genproto/googleapis/api v0.0.0-20230822172742-b8732ec3820d h1:
...
+google.golang.org/grpc v1.30.0/go.mod h1:N36X2cJ7JwdamYAgDz+s+rVMFjt3numwzf/HckM8pak=
+google.golang.org/grpc v1.31.0/go.mod h1:N36X2cJ7JwdamYAgDz+s+rVMFjt3numwzf/HckM8pak=
+google.golang.org/grpc v1.31.1/go.mod h1:N36X2cJ7JwdamYAgDz+s+rVMFjt3numwzf/HckM8pak=
+google.golang.org/grpc v1.33.1/go.mod h1:fr5YgcSWrqhRRxogOsw7RzIpsmvOZ6IcH4kBYTpR3n0=
+google.golang.org/grpc v1.33.2/go.mod h1:JMHMWHQWaTccqQQlmk3MJZS+GWXOdAesneDmEnv2fbc=
+google.golang.org/grpc v1.34.0/go.mod h1:WotjhfgOW/POjDeRt8vscBtXq+2VjORFy659qA51WJ8=
+google.golang.org/grpc v1.35.0/go.mod h1:qjiiYl8FncCW8feJPdyg3v6XW24KsRHe+dy9BAGRRjU=
+google.golang.org/grpc v1.36.0/go.mod h1:qjiiYl8FncCW8feJPdyg3v6XW24KsRHe+dy9BAGRRjU=
+google.golang.org/grpc v1.36.1/go.mod h1:qjiiYl8FncCW8feJPdyg3v6XW24KsRHe+dy9BAGRRjU=
+google.golang.org/grpc v1.38.0/go.mod h1:NREThFqKR1f3iQ6oBuvc5LadQuXVGo9rkm5ZGrQdJfM=
```
grokspawn, line 1248: Is this going to result in a bunch of dependabot/synk/other vuln scans to throw up because of these earlier versions?  Would we instead need to change the require to v1.59.0 [here](https://github.com/operator-framework/operator-lifecycle-manager/pull/3144/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6L60)?

tmshort: I don't think so, because the "requiresd" version used is v1.59.0, and that's the version within `vendor`. We ought to remove the `replaces` line, and putting in the same version doesn't make much sense.

joelanford:
Oh interesting. When I was investigating #3146, I didn't see the `replace`, so my toy client was using 1.59.0.

It looks like the behavior of the client changes quite a bit from 1.40 to 1.59 in that:
- in 1.40, the status changes from READY to CONNECTING/TRANSIENT_FAILURE when the catalog pod is killed, and eventually goes back to READY when the new catalog pod comes back up.
- in 1.59, the status changes from READY to IDLE when the catalog pod is killed and does not attempt reconnection.


That might explain why we're using a `replace`. Ultimately, I agree we should drop the replace and resolve these underlying issues.

joelanford:
I did a little bit of testing and found that we can get the client to reconnect when we detect the idle state by adding the following code snippet in `pkg/controller/registry/grpc/source.go` in the `watch()` function (around line 231, just before the `notify` channel send) 

```go
// always try to reconnect if the connection goes idle
if newState == connectivity.Idle {
    source.Conn.Connect()
}
```

My guess is that this will resolve the issues with e2es in this PR and it _may_ resolve the issue described in #3146 as well.

joelanford (approved):
Looks like it all passed, and you've also accounted for @stevekuznetsov's comment from my PR.

Shall we go ahead and merge this one? (I mainly put mine up just to get a quicker signal earlier this afternoon)

joelanford:
#3147 just merged, so this one is no longer necessary.

Issue #3143, opened 2024-01-03
Add CSV finalizers to OLM


tmshort:
## Feature Request

**Is your feature request related to a problem? Please describe.**

See #3108 and corresponding discussion:
https://kubernetes.slack.com/archives/C0181L6JYQ2/p1704293361617039

With various performance improvements, the test case in #3108 started to fail more frequently, as the Subscription would re-create the CSV faster, subsequently blocking the CR/CRBs from being deleted, and thus causing the test to fail.

However, finalizers are not used (in general) for cleanup, and should be added.

**Describe the solution you'd like**
Add finalizers for appropriate deletion actions.

***CSV Finalizer***
For instance, the CSV needs a finalizer to cleanup CRs/CRBs.

Example Scenario:

1. do all the normal setup
2. scale the controller down to 0 replicas
3. delete the CSV
4. scale the controller up to 1 replicas

No cleanup will occur.

Other finalizers may also be necessary, but we want to ensure these are resolved.

tmshort:
See: https://github.com/operator-framework/operator-lifecycle-manager/issues/3130#issuecomment-1845529275

PR #3145, opened 2024-01-04
Improve condition check for installplan

dtfranz:
This test is currently flakey because it will occasionally delete the installplan before the subscription has been properly linked up. This results in a state where the subscription has an `InstallPlanRef` but the referenced install plan does not exist. Additionally, a new installplan will not be created automatically when this happens.

Long-term, a finalizer should be added to the installplan to prevent a user from getting their subscription into a limbo state by deleting an installplan rapidly. I will add details to #3143 to capture this.


tmshort (approved):
/lgtm

PR #3148, opened 2024-01-05
CSV Finalizer

### LINKED ISSUE #3143: Add CSV finalizers to OLM

tmshort:
Fix #3143
Add a finalizer for CSVs for cleanup, rather than the existing method.

**Architectural changes:**

**Testing remarks:**

**Reviewer Checklist**

### DIFF
--- a/cmd/olm/manager.go
+++ b/cmd/olm/manager.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 
+	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
@@ -85,6 +86,12 @@ func Manager(ctx context.Context, debug bool) (ctrl.Manager, error) {
 					Label: labels.SelectorFromValidatedSet(map[string]string{install.OLMManagedLabelKey: install.OLMManagedLabelValue}),
 				},
 				&rbacv1.ClusterRoleBinding{}: {
+					Label: labels.SelectorFromValidatedSet(map[string]string{install.OLMManagedLabelKey: install.OLMManagedLabelValue}),
+				},
+				&admissionregistrationv1.MutatingWebhookConfiguration{}: {
+					Label: labels.SelectorFromValidatedSet(map[string]string{install.OLMManagedLabelKey: install.OLMManagedLabelValue}),
+				},
+				&admissionregistrationv1.ValidatingWebhookConfiguration{}: {
 					Label: labels.SelectorFromValidatedSet(map[string]string{install.OLMManagedLabelKey: install.OLMManagedLabelValue}),
 				},
 				&operatorsv1alpha1.ClusterServiceVersion{}: {
--- a/pkg/controller/operators/olm/operator.go
+++ b/pkg/controller/operators/olm/operator.go
@@ -11,7 +11,6 @@ import (
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/operators/labeller"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/operators/olm/plugins"
 	"github.com/sirupsen/logrus"
-	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
@@ -50,7 +49,6 @@ import (
 	csvutility "github.com/operator-framework/operator-lifecycle-manager/pkg/lib/csv"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/event"
 	index "github.com/operator-framework/operator-lifecycle-manager/pkg/lib/index"
-	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/kubestate"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/labeler"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/operatorclient"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/operatorlister"
@@ -86,7 +84,6 @@ type Operator struct {
 	olmConfigQueue               workqueue.RateLimitingInterface
 	csvCopyQueueSet              *queueinformer.ResourceQueueSet
 	copiedCSVGCQueueSet          *queueinformer.ResourceQueueSet
-	objGCQueueSet                *queueinformer.ResourceQueueSet
 	nsQueueSet                   workqueue.RateLimitingInterface
 	apiServiceQueue              workqueue.RateLimitingInterface
 	csvIndexers                  map[string]cache.Indexer
@@ -202,7 +199,6 @@ func newOperatorWithConfig(ctx context.Context, config *operatorConfig) (*Operat
 		olmConfigQueue:               workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "olmConfig"),
 		csvCopyQueueSet:              queueinformer.NewEmptyResourceQueueSet(),
 		copiedCSVGCQueueSet:          queueinformer.NewEmptyResourceQueueSet(),
-		objGCQueueSet:                queueinformer.NewEmptyResourceQueueSet(),
 		apiServiceQueue:              workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "apiservice"),
 		resolver:                     config.strategyResolver,
 		apiReconciler:                config.apiReconciler,
@@ -480,21 +476,6 @@ func newOperatorWithConfig(ctx context.Context, config *operatorConfig) (*Operat
 		if err := op.RegisterQueueInformer(serviceAccountQueueInformer); err != nil {
 			return nil, err
 		}
-
-		objGCQueue := workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), fmt.Sprintf("%s/obj-gc", namespace))
-		op.objGCQueueSet.Set(namespace, objGCQueue)
-		objGCQueueInformer, err := queueinformer.NewQueue(
-			ctx,
-			queueinformer.WithLogger(op.logger),
-			queueinformer.WithQueue(objGCQueue),
-			queueinformer.WithSyncer(queueinformer.LegacySyncHandler(op.syncGCObject).ToSyncer()),
-		)
-		if err != nil {
-			return nil, err
-		}
-		if err := op.RegisterQueueInformer(objGCQueueInformer); err != nil {
-			return nil, err
-		}
 	}
 
 	complete := map[schema.GroupVersionResource][]bool{}
@@ -565,22 +546,6 @@ func newOperatorWithConfig(ctx context.Context, config *operatorConfig) (*Operat
 			return op.opClient.KubernetesInterface().AppsV1().Deployments(namespace).Apply(ctx, cfg, opts)
 		},
 	)); err != nil {
-		return nil, err
-	}
-
-	// add queue for all namespaces as well
-	objGCQueue := workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), fmt.Sprintf("%s/obj-gc", ""))
-	op.objGCQueueSet.Set("", objGCQueue)
-	objGCQueueInformer, err := queueinformer.NewQueue(
-		ctx,
-		queueinformer.WithLogger(op.logger),
-		queueinformer.WithQueue(objGCQueue),
-		queueinformer.WithSyncer(queueinformer.LegacySyncHandler(op.syncGCObject).ToSyncer()),
-	)
-	if err != nil {
-		return nil, err
-	}
-	if err := op.RegisterQueueInformer(objGCQueueInformer); err != nil {
 		return nil, err
 	}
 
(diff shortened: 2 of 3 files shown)


tmshort:
* I have not added a test file for the new reconciler, either.

pkg/controller/operators/cleanupcsv_controller.go
```diff
@@ -0,0 +1,151 @@
...
+			ownerSelector := ownerutil.CSVOwnerSelector(in)
+			deleteOptions := client.DeleteAllOfOptions{
+				ListOptions: client.ListOptions{
+					LabelSelector: ownerSelector,
+				},
+			}
+			// Look for resources owned by this CSV, and delete them.
+			log.WithValues("selector", ownerSelector).Info("Cleaning up resources after CSV deletion")
+
+			err := r.Client.DeleteAllOf(ctx, &rbacv1.ClusterRoleBinding{}, &deleteOptions)
```
stevekuznetsov, line 110: Probably arguments either way, but it may make more sense to attempt to send all of these calls even if one fails, instead of bailing on the first error. WDYT?

pkg/controller/operators/cleanupcsv_controller.go
```diff
@@ -0,0 +1,151 @@
...
+			}
+
+			err = r.Client.DeleteAllOf(ctx, &admissionregistrationv1.ValidatingWebhookConfiguration{}, &deleteOptions)
+			if client.IgnoreNotFound(err) != nil {
+				log.Error(err, "Deleting ValidatingWebhookConfigurations on CSV delete")
+				return reconcile.Result{}, err
+			}
+
+			// remove our finalizer from the csv and update
+			controllerutil.RemoveFinalizer(in, myFinalizerName)
```
stevekuznetsov, line 141: If you don't set a foreground deletion policy, non-nil errors from `r.Client.DeleteAllOf()` don't necessarily mean that the objects are gone. Setting that, though, means that the handler in this controller will hang, potentially denying service if other resources are slow to clean up.

I believe all of the resources we're deleting are resources that we have informers on, so I'd suggest checking for dependent objects in the cache, and only removing the finalizer if none are present.

cmd/olm/manager.go
```diff
@@ -123,6 +123,19 @@ func Manager(ctx context.Context, debug bool) (ctrl.Manager, error) {
 		return nil, err
 	}
 
+	cleanupCsvReconciler, err := operators.NewCleanupCsvReconciler(
```
stevekuznetsov, line 126: I'll defer to @joelanford and @awgreene on this - while I understand why using the `controller-runtime` approach may be appealing here, it might be premature to move. It's a duplicate in-memory cache of all of these objects, which is a crying shame, and to make the deletion logic robust we'd need to add listers for mutating and validating admission webhooks, which moves us in the wrong direction. Then again, getting this right in the byzantine framework from v0 may be more work than it's worth.

pkg/controller/operators/olm/operator.go
```diff
@@ -948,98 +913,6 @@ func (a *Operator) EnsureCSVMetric() error {
 	return nil
 }
 
-func (a *Operator) syncGCObject(obj interface{}) (syncError error) {
```
stevekuznetsov, line 951: note to self: events on these objects by definition cannot be triggers for cleanup, so losing the cross-queueing should in theory not make any difference

pkg/controller/operators/cleanupcsv_controller.go
```diff
@@ -0,0 +1,151 @@
...
+
+		log:     log,
+		factory: factory,
+	}, nil
+}
+
+// ReconcileClusterServiceVersion projects the component labels of a given CSV onto all resources owned by it.
+func (r *CleanupCsvReconciler) ReconcileClusterServiceVersion(ctx context.Context, req ctrl.Request) (reconcile.Result, error) {
+	// Set up a convenient log object so we don't have to type request over and over again
+	log := r.log.WithValues("request", req)
```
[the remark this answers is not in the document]
tmshort: Yeah, this was copied from `adoption_controller.go`

pkg/metrics/experimental_declarations.go
```diff
@@ -19,6 +20,10 @@ func EmitOperatorReconcile(namespace, name string) {
 	emitReconcile(operatorController, namespace, name)
 }
 
+func EmitCleanupCSVReconcile(namespace, name string) {
```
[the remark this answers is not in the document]
tmshort: This is done for all the other controllers, so I was being consistent.

pkg/controller/operators/cleanupcsv_controller.go
```diff
@@ -0,0 +1,151 @@
...
+
+		return reconcile.Result{}, err
+	}
+
+	// Ignore copied csvs
+	if in.IsCopied() {
+		return reconcile.Result{}, nil
+	}
+
+	myFinalizerName := "olm.csv.cleanup.controller/finalizer"
```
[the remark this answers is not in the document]
tmshort: Yeah, this was mostly for testing, I had yet to figure out domain/name to use.

cmd/olm/manager.go
tmshort: This is strongly based on (i.e. copied from) `adoption_controller.go`, which already uses `controller-runtime` and reconciles CSVs already... BUT is only used when a specific feature flag is enabled. So, my original plan to piggy back on an existing CSV reconciler (there are two already!), was abandoned, and I created a new one. The `controller-runtime` seems to be easier to use/manage, especially with finalizers.

pkg/controller/operators/cleanupcsv_controller.go
tmshort: Certainly an option, and I think I may have done that in another iteration.

pkg/controller/operators/cleanupcsv_controller.go
```diff
@@ -0,0 +1,151 @@
...
+				return reconcile.Result{}, err
+			}
+
+			err = r.Client.DeleteAllOf(ctx, &rbacv1.ClusterRole{}, &deleteOptions)
+			if client.IgnoreNotFound(err) != nil {
+				log.Error(err, "Deleting ClusterRoles on CSV delete")
+				return reconcile.Result{}, err
+			}
+
+			deleteOptions = client.DeleteAllOfOptions{
```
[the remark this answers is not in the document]
tmshort: the selector value is actually different, or at least, I thought it was. I should double check.

pkg/controller/operators/cleanupcsv_controller.go
```diff
@@ -0,0 +1,151 @@
...
+	if in.IsCopied() {
+		return reconcile.Result{}, nil
+	}
+
+	myFinalizerName := "olm.csv.cleanup.controller/finalizer"
+	if in.ObjectMeta.DeletionTimestamp.IsZero() {
+		// not being deleted, add finalizer if not present
+		if !controllerutil.ContainsFinalizer(in, myFinalizerName) {
+			controllerutil.AddFinalizer(in, myFinalizerName)
+			if err := r.Update(ctx, in); err != nil {
```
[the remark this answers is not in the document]
tmshort: Makes sense, I was following the various online finalizer code examples with used `Update`. Not sure I can do that when deleting it, though?

pkg/controller/operators/cleanupcsv_controller.go
```diff
@@ -0,0 +1,151 @@
...
+				return reconcile.Result{}, err
+			}
+		}
+	} else {
```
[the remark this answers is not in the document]
tmshort: Same with C, but in this case (which I'd have to look at closer), it may not be a `return` (and doesn't appear to be from the code snippet above) for the whole `if` statement, meaning that the `else` block shouldn't necessarily be executed. (There's really not enough context here without diving back into the code, and it's Sunday, and I'm cleaning up after a snowstorm.)

tmshort:
Thanks @stevekuznetsov for your comments!

pkg/controller/operators/cleanupcsv_controller.go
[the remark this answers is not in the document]
stevekuznetsov: You will be able to use PATCH or SSA for both actions.

cmd/olm/manager.go
stevekuznetsov: Yes, I think the ease of coding is certainly on the side of the `controller-runtime` approach - my concern is with duplicating the memory footprint of those informers.

tmshort: Understood; the reconciller in `operatorconditiongenerator_controller.go` already reconciles the CSV, so the impact should be minimal. And I could probably move the code into there.

pkg/controller/operators/cleanupcsv_controller.go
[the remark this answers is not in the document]
tmshort: I was wrong, they're the same...

pkg/controller/operators/operatorconditiongenerator_controller.go
```diff
@@ -170,3 +179,100 @@ func (r *OperatorConditionGeneratorReconciler) ensureOperatorCondition(operatorC
...
+
+	err = r.Client.DeleteAllOf(ctx, &admissionregistrationv1.ValidatingWebhookConfiguration{}, &deleteOptions)
+	if client.IgnoreNotFound(err) != nil {
+		log.Error(err, "Deleting ValidatingWebhookConfigurations on CSV delete")
+		errs = append(errs, err)
+	}
+
+	// Make sure things are deleted
+	crbList := &rbacv1.ClusterRoleBindingList{}
+	if _ = r.Client.List(ctx, crbList, &listOptions); len(crbList.Items) != 0 {
```
stevekuznetsov, line 244: We need to use the lister for this, not a live client.

pkg/controller/operators/operatorconditiongenerator_controller.go
```diff
@@ -170,3 +179,100 @@ func (r *OperatorConditionGeneratorReconciler) ensureOperatorCondition(operatorC
...
+		errs = append(errs, fmt.Errorf("waiting for ClusterRoleBindings to delete"))
+	}
+
+	crList := &rbacv1.ClusterRoleList{}
+	if _ = r.Client.List(ctx, crList, &listOptions); len(crList.Items) != 0 {
```
stevekuznetsov, line 249: When there's an error, you will have no items, but that is not a success - the code as written would gloss over that.

pkg/controller/operators/operatorconditiongenerator_controller.go
```diff
@@ -170,3 +179,100 @@ func (r *OperatorConditionGeneratorReconciler) ensureOperatorCondition(operatorC
...
+// Return values, err, ok; ok == true: continue Reconcile, ok == false: exit Reconcile
+func (r *OperatorConditionGeneratorReconciler) processFinalizer(ctx context.Context, csv *operatorsv1alpha1.ClusterServiceVersion) (error, bool) {
+	myFinalizerName := "operators.coreos.com/csv-cleanup"
+	log := r.log.WithValues("name", csv.GetName()).WithValues("namespace", csv.GetNamespace())
+
+	if csv.ObjectMeta.DeletionTimestamp.IsZero() {
+		// CSV is not being deleted, add finalizer if not present
+		if !controllerutil.ContainsFinalizer(csv, myFinalizerName) {
+			patch := client.MergeFrom(csv.DeepCopy())
+			controllerutil.AddFinalizer(csv, myFinalizerName)
```
stevekuznetsov, line 192: Mutate your deep copy, not the object you got from the cache

cmd/olm/manager.go
stevekuznetsov: Either one is missing listers for some of these resources and will have the same downside. @joelanford do we care?

pkg/controller/operators/operatorconditiongenerator_controller.go
tmshort: "need" or "highly desirable"? There's no code in place for informers/indexers/listers in this part of the code. I'll have to pull in a lot more infrastructure.

stevekuznetsov: Need, with a capital N. The server load doing live lists is extremely hard. This is why I made this comment d458b66414

tmshort: OK, to be clear, it's not needed for _functionality_ (as it works already), but needed for _performance_. CSV deletion doesn't occur very often, but it could hit the API server fairly hard if deleting these resources takes longer than expected.

stevekuznetsov: Not sure the distinction means much, to be honest. It's needed to meet the quality bar we have for controllers in OCP. Full-cluster resource lists are a good way to DoS the cluster.

tmshort: OK, before I dive too deep into this, let's make sure it's not already using a cache.
The client being used comes from https://github.com/operator-framework/operator-lifecycle-manager/blob/920a92140b3aeb581c48b03635839406ec9624cd/cmd/olm/manager.go#L114

Which  comes from a Controller-Runtime Manager, which is created with a Cache:
https://github.com/operator-framework/operator-lifecycle-manager/blob/920a92140b3aeb581c48b03635839406ec9624cd/cmd/olm/manager.go#L55

AFAICT, the Client from the Manager uses the provided cache. In such a case, I shouldn't have to add separate informers, and should just have to add the appropriate resources to the cache.

https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/manager#Options

tmshort: That's what I thought. Thanks!

stevekuznetsov:
@tmshort can you dupe the issue / create one for the InstallPlan work since this PR will close #3143 but not implement that half?

tmshort:
> @tmshort can you dupe the issue / create one for the InstallPlan work since this PR will close #3143 but not implement that half?

Yes.

PR #3149, opened 2024-01-05
Snapshot Fix for Deprecation Info

dtfranz:
When initially implemented, in order to retrieve deprecation metadata the subscription resolver was modified to add an optional sourceProvider and we incorrectly used the aggregate `OperatorGroupToggleSourceProvider`. This resulted in errors when taking snapshots from namespaces in which no `OperatorGroups` were present. 

To fix, we just use the `RegistrySourceProvider` directly instead of the aggregate which is what we actually needed in the first place. This results in silencing the errors mentioned above and also resolves an issue where the deprecation info will not be added to `Subscription` status when the `CatalogSource`'s namespace does not contain an `OperatorGroup`.

### DIFF
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -346,7 +346,7 @@ func NewOperator(ctx context.Context, kubeconfigPath string, clock utilclock.Clo
 		subscription.WithAppendedReconcilers(subscription.ReconcilerFromLegacySyncHandler(op.syncSubscriptions, nil)),
 		subscription.WithRegistryReconcilerFactory(op.reconciler),
 		subscription.WithGlobalCatalogNamespace(op.namespace),
-		subscription.WithSourceProvider(resolverSourceProvider),
+		subscription.WithSourceProvider(op.sourceInvalidator),
 	)
 	if err != nil {
 		return nil, err


tmshort (approved):
/lgtm

pkg/controller/operators/catalog/operator.go
```diff
@@ -346,7 +346,7 @@ func NewOperator(ctx context.Context, kubeconfigPath string, clock utilclock.Clo
 		subscription.WithAppendedReconcilers(subscription.ReconcilerFromLegacySyncHandler(op.syncSubscriptions, nil)),
 		subscription.WithRegistryReconcilerFactory(op.reconciler),
 		subscription.WithGlobalCatalogNamespace(op.namespace),
-		subscription.WithSourceProvider(resolverSourceProvider),
+		subscription.WithSourceProvider(op.sourceInvalidator),
```
awgreene, line 349: So just to clarify, we needed the list of catalogSources in https://github.com/operator-framework/operator-lifecycle-manager/commit/b9638dcaaa14d4005b12abe63feb2055a49ca793 and we just used the wrong source provider?

dtfranz: Yeah that's right, the `OperatorGroupToggleSourceProvider` (from what I understand) is essentially a filter which allows users to exclude global catalogs from resolution. But I don't believe that this `sourceProvider` was intended to be used when taking cache snapshots of the `CatalogSource` namespace, instead it's used to allow for instance a user to ignore a Catalog update by adding an annotation to the `OperatorGroup` of a separate namespace in which their operator is installed.

In OCP deployments, the `openshift-marketplace` namespace doesn't have an `OperatorGroup`, so if you use this source provider it will error out when taking snapshots and give you nothing back.

dtfranz:
Any idea what's going on with tide? I haven't seen this before:
`tide Pending — Not mergeable. No Tide query for branch master found. `

tmshort:
> Any idea what's going on with tide? I haven't seen this before: `tide Pending — Not mergeable. No Tide query for branch master found. `

It's disabled for the master branch. It's fine to merge.

pkg/controller/operators/catalog/operator.go
awgreene: Ah got it, thanks Daniel!

PR #3159, opened 2024-01-18
ClusterRoleBindings + Services: Optimistic Create

dtfranz:
Modify installation steps for ClusterRoleBindings and Services to attempt creation first then update/replace if existing in order to avoid race conditions.

### DIFF
--- a/pkg/controller/install/certresources.go
+++ b/pkg/controller/install/certresources.go
@@ -253,8 +253,15 @@ func (i *StrategyDeploymentInstaller) installCertRequirementsForDeployment(deplo
 	ownerutil.AddNonBlockingOwner(service, i.owner)
 	service.SetLabels(map[string]string{OLMManagedLabelKey: OLMManagedLabelValue})
 
-	existingService, err := i.strategyClient.GetOpLister().CoreV1().ServiceLister().Services(i.owner.GetNamespace()).Get(service.GetName())
-	if err == nil {
+	// Optimistically create the Service
+	_, err := i.strategyClient.GetOpClient().CreateService(service)
+	if apierrors.IsAlreadyExists(err) {
+		// If the Service already exists, attempt to replace it if safe to do so.
+		existingService, err := i.strategyClient.GetOpLister().CoreV1().ServiceLister().Services(i.owner.GetNamespace()).Get(service.GetName())
+		if err != nil {
+			return nil, nil, fmt.Errorf("could not retrieve existing service %s: %s", service.GetName(), err.Error())
+		}
+
 		if !ownerutil.Adoptable(i.owner, existingService.GetOwnerReferences()) {
 			return nil, nil, fmt.Errorf("service %s not safe to replace: extraneous ownerreferences found", service.GetName())
 		}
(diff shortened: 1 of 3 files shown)


pkg/controller/operators/olm/operatorgroup.go
```diff
@@ -585,7 +585,7 @@ func (a *Operator) ensureSingletonRBAC(operatorNamespace string, csv *v1alpha1.C
 					Name:     r.RoleRef.Name,
 				},
 			}
-			// TODO: this should do something smarter if the cluster role binding already exists
+			// here
```
kevinrizza, line 588: ?

dtfranz: Oops, that's embarrassing

pkg/controller/install/certresources.go
```diff
@@ -459,32 +465,33 @@ func (i *StrategyDeploymentInstaller) installCertRequirementsForDeployment(deplo
...
+	// Optimistically create the ClusterRoleBinding with CSV as the labelled owner
+	if err := ownerutil.AddOwnerLabels(authDelegatorClusterRoleBinding, i.owner); err != nil {
+		return nil, nil, err
+	}
+	_, err = i.strategyClient.GetOpClient().CreateClusterRoleBinding(authDelegatorClusterRoleBinding)
+	if apierrors.IsAlreadyExists(err) {
+		// ClusterRoleBinding already exists; update if safe to do so.
+		existingAuthDelegatorClusterRoleBinding, err := i.strategyClient.GetOpLister().RbacV1().ClusterRoleBindingLister().Get(authDelegatorClusterRoleBinding.GetName())
+		if err != nil {
+			return nil, nil, err
```
everettraven, line 477: I like the more descriptive error given on https://github.com/operator-framework/operator-lifecycle-manager/pull/3159/files#diff-bc7b779225ba70ff0e1f63ff55cb2dd3d71b5ae4f96c158947266da9f8fde60dR262 - could that be done here as well?

pkg/controller/install/certresources.go
```diff
@@ -253,8 +253,15 @@ func (i *StrategyDeploymentInstaller) installCertRequirementsForDeployment(deplo
 	ownerutil.AddNonBlockingOwner(service, i.owner)
 	service.SetLabels(map[string]string{OLMManagedLabelKey: OLMManagedLabelValue})
 
-	existingService, err := i.strategyClient.GetOpLister().CoreV1().ServiceLister().Services(i.owner.GetNamespace()).Get(service.GetName())
-	if err == nil {
+	// Optimistically create the Service
+	_, err := i.strategyClient.GetOpClient().CreateService(service)
+	if apierrors.IsAlreadyExists(err) {
+		// If the Service already exists, attempt to replace it if safe to do so.
```
everettraven, line 259: I know that this is building on top of existing logic to delete and then recreate, but is there a specific reason why it needs to be a delete followed by a create? Could the same goal be accomplished with a patch/update?

dtfranz: IME this is because there are many rules for the service spec around updates and it's very easy to run into errors unless you delete then recreate them. I can't really say with certainty if this was or is a best-practice of any kind, just trying to give a potential reason why this code was set up this way.

pkg/controller/install/certresources.go
dtfranz: Absolutely!

dtfranz: Updated, thanks!

pkg/controller/install/certresources.go
everettraven: Yeah, that makes sense. This should definitely not block this PR since this is _generally_ the same logic. Was curious if there was any historical reasoning for this behavior that you might be aware of since I noticed there is also a `UpdateService()` method on the service client interface when combing through the code

everettraven (commented):
I'm not very familiar with this codebase so I'll leave explicit approval to someone more familiar, but the changes LGTM

m1kola (approved):
I'm not very well versed in this part of OLM, but the changes look reasonable to me.

I left one comment, but it is not blocking.

pkg/controller/install/certresources.go
```diff
@@ -263,15 +270,14 @@ func (i *StrategyDeploymentInstaller) installCertRequirementsForDeployment(deplo
 		// Delete the Service to replace
 		deleteErr := i.strategyClient.GetOpClient().DeleteService(service.GetNamespace(), service.GetName(), &metav1.DeleteOptions{})
 		if deleteErr != nil && !apierrors.IsNotFound(deleteErr) {
-			return nil, nil, fmt.Errorf("could not delete existing service %s", service.GetName())
+			return nil, nil, fmt.Errorf("could not delete existing service %s: %w", service.GetName(), err)
```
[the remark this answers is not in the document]
dtfranz: Fair enough, there's probably not a need to wrap them.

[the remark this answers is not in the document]
dtfranz: Updated!

m1kola (approved):
/lgtm

pkg/controller/install/certresources.go
stevekuznetsov: Server side apply will make sure you only need one call and will do the right thing (e.g. create if not exist) in case you want to consider it.

PR #3166, opened 2024-01-31
Clear (existing) error cond from Subscription, once error resolved

### LINKED ISSUE #3162: Resolution error `constraints not satisfiable: no operator found from catalog` is not cleared after CatalogSource is made available again
## Bug Report

**What did you do?**
1. Run olm in a kind cluster and install an operator: 

```sh 
$ make run-local 

$ kubectl apply -f - <<EOF
kind: OperatorGroup
metadata:
  name: my-group
  namespace: default
EOF
operatorgroup.operators.coreos.com/my-group created

$ kubectl apply -f -<<EOF
kind: Subscription
metadata:
  name: camel-k
  namespace: default
spec:
  channel: stable-v2
  name: camel-k
  source: operatorhubio-catalog
  sourceNamespace: olm
EOF
subscription.operators.coreos.com/camel-k created
```

2. Wait for operator to be installed successfully: 

```sh
$ kubectl get csv 
NAME                      DISPLAY            VERSION   REPLACES                  PHASE
camel-k-operator.v2.2.0   Camel K Operator   2.2.0     camel-k-operator.v2.1.0   Succeeded

$ kubectl get subscription.operators -A -o custom-columns='NAMESPACE:.metadata.namespace,NAME:.metadata.name,ResolutionFailed:.status.conditions[?(@.type=="ResolutionFailed")].status,MSG:.status.conditions[?(@.type=="ResolutionFailed")].message'
NAMESPACE   NAME      ResolutionFailed   MSG
default     camel-k   <none>             <none>
```

3. Force a "CatalogSource not found" resolution error: 

```sh 
$  kubectl delete catalogsource operatorhubio-catalog -n olm 
catalogsource.operators.coreos.com "operatorhubio-catalog" deleted
$ k get pods -n olm 
NAME                                READY   STATUS    RESTARTS   AGE
catalog-operator-86664778f9-pmcxc   1/1     Running   0          7m32s
olm-operator-9d6d59964-gvqt6        1/1     Running   0          7m32s
packageserver-7fdd5ccd9f-rbsrz      1/1     Running   0          7m31s

$ kubectl delete pod catalog-operator-86664778f9-pmcxc 
pod "catalog-operator-86664778f9-pmcxc" deleted

$ kubectl get subscription.operators -A -o custom-columns='NAMESPACE:.metadata.namespace,NAME:.metadata.name,ResolutionFailed:.status.conditions[?(@.type=="ResolutionFailed")].status,MSG:.status.conditions[?(@.type=="ResolutionFailed")].message' 
NAMESPACE   NAME      ResolutionFailed   MSG
default     camel-k   True               constraints not satisfiable: no operators found from catalog operatorhubio-catalog in namespace olm referenced by subscription camel-k, subscription camel-k exists
```

4. Recreate the CatalogSource
```
$ kubectl apply -f deploy/upstream/manifests/latest/0000_50_olm_17-upstream-operators.catalogsource.yaml 
catalogsource.operators.coreos.com/operatorhubio-catalog created
```

After deleting the new catalog-operator pod to force another resolution, the error message on the subscription to go away
```sh
$ kubectl delete pod catalog-operator-86664778f9-mmmqg -n olm 
pod "catalog-operator-86664778f9-mmmqg" deleted
```

Error message still exists
```
$ kubectl get subscription.operators -A -o custom-columns='NAMESPACE:.metadata.namespace,NAME:.metadata.name,ResolutionFailed:.status.conditions[?(@.type=="ResolutionFailed")].status,MSG:.status.conditions[?(@.type=="ResolutionFailed")].message' 
NAMESPACE   NAME      ResolutionFailed   MSG
default     camel-k   True               constraints not satisfiable: no operators found from catalog operatorhubio-catalog in namespace olm referenced by subscription camel-k, subscription camel-k exists
```

**Environment**

master

```
Client Version: version.Info{Major:"1", Minor:"25", GitVersion:"v1.25.0", GitCommit:"a866cbe2e5bbaa01cfd5e969aa3e033f3282a8a2", GitTreeState:"clean", BuildDate:"2022-08-23T17:44:59Z", GoVersion:"go1.19", Compiler:"gc", Platform:"darwin/arm64"}
Kustomize Version: v4.5.7
Server Version: version.Info{Major:"1", Minor:"26", GitVersion:"v1.26.3", GitCommit:"9e644106593f3f4aa98f8a84b23db5fa378900bd", GitTreeState:"clean", BuildDate:"2023-03-30T06:36:32Z", GoVersion:"go1.19.7", Compiler:"gc", Platform:"linux/arm64"}
```
* Kubernetes cluster kind: 

kind v0.18.0 go1.20.2 darwin/arm64

**Possible Solution**

**Additional context**

anik120:

The func `removeSubsCond` takes in a list of pointers to Subscription objects, modifies the
objects that the pointers point to, but return a new list of those pointers. A [PR](https://github.com/operator-framework/operator-lifecycle-manager/pull/2942) included in
the v0.25.0 release [changed the way the output of that function was being used](https://github.com/operator-framework/operator-lifecycle-manager/pull/2942/files#diff-a1760d9b7ac1e93734eea675d8d8938c96a50e995434b163c6f77c91bace9990R1146-R1155) leading to a regression. 
This PR fixes the `removeSubsCond` function, fixing the regression as a result.

Closes #3162

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -1393,16 +1393,14 @@ func (o *Operator) syncResolvingNamespace(obj interface{}) error {
 	}
 
 	// Make sure that we no longer indicate unpacking progress
-	subs = o.setSubsCond(subs, v1alpha1.SubscriptionCondition{
-		Type:   v1alpha1.SubscriptionBundleUnpacking,
-		Status: corev1.ConditionFalse,
-	})
+	o.removeSubsCond(subs, v1alpha1.SubscriptionBundleUnpacking)
 
 	// Remove BundleUnpackFailed condition from subscriptions
 	o.removeSubsCond(subs, v1alpha1.SubscriptionBundleUnpackFailed)
 
 	// Remove resolutionfailed condition from subscriptions
 	o.removeSubsCond(subs, v1alpha1.SubscriptionResolutionFailed)
+
 	newSub := true
 	for _, updatedSub := range updatedSubs {
 		updatedSub.Status.RemoveConditions(v1alpha1.SubscriptionResolutionFailed)
@@ -1679,13 +1677,9 @@ func (o *Operator) setSubsCond(subs []*v1alpha1.Subscription, cond v1alpha1.Subs
 	return subList
 }
 
-// removeSubsCond will remove the condition to the subscription if it exists
-// Only return the list of updated subscriptions
-func (o *Operator) removeSubsCond(subs []*v1alpha1.Subscription, condType v1alpha1.SubscriptionConditionType) []*v1alpha1.Subscription {
-	var (
-		lastUpdated = o.now()
-	)
-	var subList []*v1alpha1.Subscription
+// removeSubsCond removes the given condition from all of the subscriptions in the input
+func (o *Operator) removeSubsCond(subs []*v1alpha1.Subscription, condType v1alpha1.SubscriptionConditionType) {
+	lastUpdated := o.now()
 	for _, sub := range subs {
 		cond := sub.Status.GetCondition(condType)
 		// if status is ConditionUnknown, the condition doesn't exist. Just skip
@@ -1694,9 +1688,7 @@ func (o *Operator) removeSubsCond(subs []*v1alpha1.Subscription, condType v1alph
 		}
 		sub.Status.LastUpdated = lastUpdated
 		sub.Status.RemoveConditions(condType)
-		subList = append(subList, sub)
 	}
-	return subList
 }
 
 func (o *Operator) updateSubscriptionStatuses(subs []*v1alpha1.Subscription) ([]*v1alpha1.Subscription, error) {
--- a/pkg/controller/operators/catalog/operator_test.go
+++ b/pkg/controller/operators/catalog/operator_test.go
@@ -1261,13 +1261,6 @@ func TestSyncResolvingNamespace(t *testing.T) {
 					Status: v1alpha1.SubscriptionStatus{
 						CurrentCSV: "",
 						State:      "",
-						Conditions: []v1alpha1.SubscriptionCondition{
-							{
-								Type:   v1alpha1.SubscriptionBundleUnpacking,
-								Status: corev1.ConditionFalse,
-							},
-						},
-						LastUpdated: now,
 					},
 				},
 			},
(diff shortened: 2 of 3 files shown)


pkg/controller/operators/catalog/operator.go
```diff
@@ -1681,11 +1679,8 @@ func (o *Operator) setSubsCond(subs []*v1alpha1.Subscription, cond v1alpha1.Subs
 
 // removeSubsCond will remove the condition to the subscription if it exists
 // Only return the list of updated subscriptions
```
m1kola, line 1681:
```suggestion
```

pkg/controller/operators/catalog/operator.go
```diff
@@ -1694,9 +1689,7 @@ func (o *Operator) removeSubsCond(subs []*v1alpha1.Subscription, condType v1alph
 		}
 		sub.Status.LastUpdated = lastUpdated
 		sub.Status.RemoveConditions(condType)
-		subList = append(subList, sub)
 	}
-	return subList
```
m1kola, line 1699: Since #2942 we no longer use the return value of `removeSubsCond` - how removing return value helps to fix #3162?

I think I'm missing something. I read the explanation in the PR description, but I still do not understand the connection. Could you please explain it to me in more details?

tmshort:
Generally looks good to me. But @m1kola has some quyestions.

anik120:
> /retest

@tmshort there was a change I needed to do in the e2e test to reflect "BundleUnpacking cond is now removed instead of being set to false" in the e2e test 😅 Made that change, all tests should pass now.

pkg/controller/operators/catalog/operator.go
anik120: @m1kola so it's definitely a little bit confusing to me too. The thing that actually fixed the problem isn't just "removing the return value" though. It's the fact that `setSubsCond` is not being used here and instead has been switched over to `removeSubsCond`. From what I can tell by trying to keep track of everything going on with original pointers being modified and new list being returned, it looked like there were two lists we were dealing with [here](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/controller/operators/catalog/operator.go#L1396-L1405), and one of those lists were actually reaching [`UpdateSubscriptionStatues`](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/controller/operators/catalog/operator.go#L1424), leading to the `removeSubsCond(resolutionError)` to fly under the radar without making any dent. 

So the thing that actually fixed the problem, is getting rid of the confusions by using the list of pointers like it's supposed to be used, which it seems was the intention in #2942 too.

m1kola: Thanks for the explanation, I did not take the fact that `setSubsCond` is being used into account.

Resolving this thread.

pkg/controller/operators/catalog/operator.go
```diff
@@ -1393,16 +1393,14 @@ func (o *Operator) syncResolvingNamespace(obj interface{}) error {
 	}
 
 	// Make sure that we no longer indicate unpacking progress
-	subs = o.setSubsCond(subs, v1alpha1.SubscriptionCondition{
-		Type:   v1alpha1.SubscriptionBundleUnpacking,
-		Status: corev1.ConditionFalse,
-	})
+	o.removeSubsCond(subs, v1alpha1.SubscriptionBundleUnpacking)
```
m1kola, line 1396: If we want to keep the condition, _I think_ alternative fix will be something like this:

```suggestion
	o.setSubsCond(subs, v1alpha1.SubscriptionCondition{
		Type:   v1alpha1.SubscriptionBundleUnpacking,
		Status: corev1.ConditionFalse,
	})
```

Because `setSubsCond` returns a new slice of pointers containing updated subs only (similar to old implementation of `removeSubsCond`).

m1kola: Removing condition can potentially break clients, but given the fact:
1. It is relatively recent change
2. And not consistent with other conditions

I think it might be ok to remove it.

m1kola (approved):
I'm a bit on the fence about removing the condition instead of setting it to False (see comments below), but other than that - looks good.

pkg/controller/operators/catalog/operator.go
anik120: My vote would be to remove it. Reasons I say that are: 
* When I asked around I wasn't able to get a "a user asked for it and hence it was added in there" answer.
* Which feeds into the doubts about the usefulness of clients needing to know if BundleUnpacking is true or false. As opposed to for example, the usefulness of clients needing to know if `ResolutionFailed` is true or false.

m1kola: Ship it! :shipit:

PR #3170, opened 2024-02-05
Reflect BundleUnpacking cond removal in missed e2e test

anik120:
PR #3166 removes the `BundleUnpacking` condition once resolution is successful. PR #3166 [modified an e2e test](https://github.com/operator-framework/operator-lifecycle-manager/commit/54da66a9996632315827ba37e14823de6405b4d9#diff-11f70fc71ac22d725767916f562789de88d06eb9ebe19f337a59fd7035a3ca2dR2448) to reflect that change. However, the test being fixed in this PR is skipped for upstream, and runs only downstream.This PR is a follow up to #3166 to reflect the `BundleUnpacking` condition removal in the remaining test.

**Architectural changes:**

**Testing remarks:**

**Reviewer Checklist**


test/e2e/subscription_e2e_test.go
```diff
@@ -2806,7 +2806,7 @@ properties:
 				if err != nil {
 					return err
 				}
-				if cond := fetched.Status.GetCondition(v1alpha1.SubscriptionBundleUnpacking); cond.Status != corev1.ConditionFalse {
+				if cond := fetched.Status.GetCondition(v1alpha1.SubscriptionBundleUnpacking); cond.Status != corev1.ConditionUnknown {
```
stevekuznetsov, line 2809: sidenote that a condition being present with conditionunknown is not the same thing as a condition being absent, the `GetCondition()` impl is sufficiently surprising ... we might want to clean up that semantic in the future

stevekuznetsov (approved):

test/e2e/subscription_e2e_test.go
joelanford: I suspect the upstream API conventions are the reason for that logic:
> Controllers should apply their conditions to a resource the first time they visit the resource, even if the status is Unknown. This allows other components in the system to know that the condition exists and the controller is making progress on reconciling that resource.
> 
> Not all controllers will observe the previous advice about reporting "Unknown" or "False" values. For known conditions, the absence of a condition status should be interpreted the same as Unknown, and typically indicates that reconciliation has not yet finished (or that the resource state may not yet be observable).

https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties

anik120: Ah okay. So setting the conditions as `False` is in fact the right way to go. Which means, in Ankita's PR, she was doing the right thing by setting the condition to false. However, "the mess" I kept uncovering was because after Ankita did the right thing for one condition, there existed a mixture of how we handle the conditions in the CR, some were being removed, while one was being set to false. And this confusion of some conditions being removed and one condition being set to False led to the original "why isn't the ResolutionFailed condition being reset even after the error was resolved by means of a Healthy catalog?".

Creating an issue to capture the task of "set conditions to false", so that we can follow upstream API conventions for our conditions in future releases, while avoiding confusing bugs.

anik120: https://github.com/operator-framework/operator-lifecycle-manager/issues/3171

stevekuznetsov: I think given the context your change is perfectly acceptable. Orthogonally to what Joe's citing, a method that searches for a condition should just return `nil` or add some boolean to the signature and return `false` when no such condition exists on the object. Returning some default value and imposing some semantic on top is surprising. For instance, if we were to be writing a controller using this `GetCondition()` function and we wanted to know - have we already set the thing to `Unknown` or do we need to add one? Today, not possible with `GetCondition()`.

PR #3182, opened 2024-02-28
ClusterRoleBindings and Services Installation: SSA

dtfranz:
Modify installation steps for ClusterRoleBindings and Services to use SSA in order to avoid race conditions.

~In draft until I fix a failing unit test.~
Fake Client and envtest do not currently support SSA, but creating the objects beforehand functions as a workaround for now. I've left comments in the PR pointing out areas where I've used this.

### DIFF
--- a/pkg/controller/install/certresources.go
+++ b/pkg/controller/install/certresources.go
@@ -12,6 +12,8 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	corev1ac "k8s.io/client-go/applyconfigurations/core/v1"
+	rbacv1ac "k8s.io/client-go/applyconfigurations/rbac/v1"
 
 	"github.com/operator-framework/api/pkg/operators/v1alpha1"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/certs"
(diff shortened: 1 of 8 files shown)


pkg/controller/install/certresources.go
```diff
@@ -241,43 +244,45 @@ func CalculateCertRotatesAt(certExpirationTime time.Time) time.Time {
...
+		portsApplyConfig = append(portsApplyConfig, ac)
+	}
+
+	ownerRef := ownerutil.NonBlockingOwner(i.owner)
+	ownerRefAC := metav1ac.OwnerReference().
+		WithAPIVersion(ownerRef.APIVersion).
+		WithKind(ownerRef.Kind).
+		WithUID(ownerRef.UID).
+		WithName(ownerRef.Name)
+	if ownerRef.BlockOwnerDeletion != nil {
```
stevekuznetsov, line 264: the `NonBlockingOwner()` is explicitly setting these, always - if I were maintaining this codebase I'd perfer to see some function in the `ownerutil` package that exposes an `OwnerReferenceApplyConfiguration` and closes over details like this

pkg/controller/install/certresources_test.go
```diff
@@ -167,7 +169,35 @@ func TestInstallCertRequirementsForDeployment(t *testing.T) {
...
+					ownerRefAC.WithController(*ownerRef.Controller)
+				}
+				svcApplyConfig := corev1ac.Service(service.Name, service.Namespace).
+					WithSpec(corev1ac.ServiceSpec().
+						WithPorts(portsApplyConfig...).
+						WithSelector(selector(t, "test=label").MatchLabels)).
+					WithOwnerReferences(ownerRefAC).
+					WithLabels(map[string]string{OLMManagedLabelKey: OLMManagedLabelValue})
+
+				mockOpClient.EXPECT().ApplyService(svcApplyConfig, metav1.ApplyOptions{Force: true, FieldManager: "olm.install"}).Return(&service, nil)
```
stevekuznetsov, line 200: I don't know if the mock client testing bits support apply :|

dtfranz: Fortunately, these calls do work properly.

PR #3185, opened 2024-03-04
OCPBUGS-29729: Refactor security context configuration in pod reconciler

bentito:
This change updates the logic for setting security contexts within the OLM pod reconciler. Now, it differentiates between 'Restricted' and 'Legacy' security contexts more explicitly. The 'Restricted' security context applies default security settings unless overridden, while the 'Legacy' context clears all security settings. When no security context is configured, it defaults to restricted. Additionally, the related tests have been updated to reflect these changes and ensure correct behavior.

OCPBUGS-29729

**Architectural changes:**
None

**Testing remarks:**

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/registry/reconciler/reconciler.go
+++ b/pkg/controller/registry/reconciler/reconciler.go
@@ -206,8 +206,21 @@ func Pod(source *operatorsv1alpha1.CatalogSource, name, opmImg, utilImage, img s
 		},
 	}
 
-	if source.Spec.GrpcPodConfig != nil && source.Spec.GrpcPodConfig.SecurityContextConfig == operatorsv1alpha1.Restricted {
-		addSecurityContext(pod, runAsUser)
+	// Check if GrpcPodConfig is defined
+	if source.Spec.GrpcPodConfig != nil {
+		// Determine the security context configuration
+		securityContextConfig := source.Spec.GrpcPodConfig.SecurityContextConfig
+
+		// Apply 'restricted' security settings if SecurityContextConfig is 'Restricted' or not set
+		if securityContextConfig == operatorsv1alpha1.Restricted || securityContextConfig == "" {
+			addSecurityContext(pod, runAsUser) // Apply restricted settings
+		} else {
+			// For 'Legacy' or any unspecified value, clear all security contexts
+			clearAllSecurityContexts(pod)
+		}
+	} else {
+		// If GrpcPodConfig is nil, apply 'restricted' security settings by default
+		addSecurityContext(pod, runAsUser) // Default to restricted settings
 	}
 
 	// Override scheduling options if specified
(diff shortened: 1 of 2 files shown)


pkg/controller/registry/reconciler/reconciler.go
```diff
@@ -206,8 +206,21 @@ func Pod(source *operatorsv1alpha1.CatalogSource, name, opmImg, utilImage, img s
...
+	if source.Spec.GrpcPodConfig != nil {
+		// Determine the security context configuration
+		securityContextConfig := source.Spec.GrpcPodConfig.SecurityContextConfig
+
+		// Apply 'restricted' security settings if SecurityContextConfig is 'Restricted' or not set
+		if securityContextConfig == operatorsv1alpha1.Restricted || securityContextConfig == "" {
+			addSecurityContext(pod, runAsUser) // Apply restricted settings
+		} else {
+			// For 'Legacy' or any unspecified value, clear all security contexts
+			clearAllSecurityContexts(pod)
```
joelanford, line 219: I'm curious why we need to clear existing security contexts. Aren't we building the "expected" pod from scratch in this function?

pkg/controller/registry/reconciler/reconciler.go
```diff
@@ -206,8 +206,21 @@ func Pod(source *operatorsv1alpha1.CatalogSource, name, opmImg, utilImage, img s
...
+		}
+	} else {
+		// If GrpcPodConfig is nil, apply 'restricted' security settings by default
+		addSecurityContext(pod, runAsUser) // Default to restricted settings
```
joelanford, line 223: I mentioned this elsewhere, but the implication here is that we will break existing catalog sources that reference catalog images that are not compatible with the securityContext required by the PSA restricted mode.

In general, are these implications well understood and gamed out? Any chance we have a design doc about this change of the default behavior?

bentito: ok, that approach makes sense. I will rework this fix to check the namespace PSA config first.

tmshort: Are we still waiting for an update on this?

joelanford: I would definitely avoid making a client call inside the `Pod` function, especially since we call it so many times.

Steve ran into issues where the resolver would use clients to fetch data from the cache and/or apiserver. The result was really strange race conditions because the cache and apiserver can change during the course of a single resolution.

This sounds like it could run into similar problems. I think we should find a way to call `getNamespacePSAConfig` just once during a reconciliation and then use its result anywhere it is needed.

pkg/controller/registry/reconciler/reconciler.go
bentito: Currently in the construction we set one item, `ReadOnlyRootFilesystem`.

pkg/controller/registry/reconciler/reconciler.go
bentito: The last commit (f19d68a) adds checking of the namespace PSA restrictions. Currently, I am at about 8 files changed with many changes in each file (git diff is 871 lines). I am close, but still have some failing unit tests.

My basic approach was looking at the call graph and trying to minimize changes. Given that, I am trying to change the two wrapper `Pod()` functions in `configmap.go` and `grpc.go`. These locations already have access to the client and seem like a good place to put a fetch of the namespace object. We're checking if it's cached already to meet the only call once per reconcile loop goal.

I guess my question is, does this seem worth continuing to pursue for this bug fix, hammer out the rest of the unit test problems? Or, try some other approach...

PR #3186, opened 2024-03-05
Add slash command access to make unit

bentito:
This should allow PR comments of the form:
`/retest unit` to trigger `make unit`

Prow offers PR comments to trigger fine-grained retest capabilities for tests. This trials adding this capability for the unit test target.

**Architectural changes:**

None

**Testing remarks:**

**Reviewer Checklist**

### DIFF
--- a/.github/workflows/unit.yml
+++ b/.github/workflows/unit.yml
@@ -8,8 +8,12 @@ on:
   pull_request:
   workflow_dispatch:
   merge_group:
+  issue_comment:
+    types: [created]  # Triggers the workflow when a comment is created. see `if` section
 jobs:
   unit:
+    if: >-
+      github.event_name != 'issue_comment' || startsWith(github.event.comment.body, '/retest unit')
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3


.github/workflows/unit.yml
```diff
@@ -8,8 +8,13 @@ on:
   pull_request:
   workflow_dispatch:
   merge_group:
+  issue_comment:
+    types: [created]
 jobs:
   unit:
+    if: >-
+      github.event_name == 'pull_request' || 
+      (github.event_name == 'issue_comment' && startsWith(github.event.comment.body, '/retest unit'))
```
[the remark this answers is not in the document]
bentito: `contains` might trigger on comments unintentionally, startsWith forces to a smaller trigger.

This syntax is meant to be like Prow commands which many users of Prow based workflow are used to.

Was hoping to just start with a simple one here and we can add other commands as time and interest allows.

kevinrizza (dismissed):

.github/workflows/unit.yml
m1kola, line 17: I think this will mean we do not run unit tests on merge queue, etc.

m1kola (changes requested):
I think this change makes merge queue and other events to skip unit tests.

.github/workflows/unit.yml
bentito: in the `on` section we have `on:
on-push-branches:- master`  that should trigger the workflow on commits pushed to master, including merges. But just to be sure, and more clear, we add other events and comments to the `on` section in f6b3a1e4

m1kola: I think `on` section was fine as it was originally. I would suggest reverting back.

My concern is within `if` which makes the `unit` job conditional. As it is right now - I think the workflow will be triggered on multiple events (push to `master`, pull request, merge queue, comment, etc), but the `unit` job will only be run on pull request events and comment events).

Or am I missing something?

bentito: Ah, I think I get the problem you're finding. Yeah I think `on` can get reverted but `if` section needs to get modded to allow in all the other events we want unit to be run:
```yaml
jobs:
  unit:
    if: >-
      github.event_name == 'push' ||
      github.event_name == 'pull_request' ||
      github.event_name == 'workflow_dispatch' ||
      github.event_name == 'merge_group' ||
      (github.event_name == 'issue_comment' && startsWith(github.event.comment.body, '/retest unit'))
```
Seems right?

m1kola: Maybe something like `github.event_name != 'issue_comment' || startsWith(github.event.comment.body, '/retest unit')`?

So we don't have to update in two places if we change something in `on`?

bentito: Yep, that's better! 39d866f cleans up the on section and adds that to if section

m1kola (approved):
Looks good to me. I'm approving, but if you could squash into one commit - that would be great.

bentito:
e2e (3) seems to fail on a "Subscription when unpacking bundle with bundle which OLM will fail to unpack [It] should report deprecated status when catalog is updated to deprecate an installed bundle"?

Seems unrelated to this PR. Merging again as first check on whether this is a flaky e2e test or something that needs fixing to unblock merges.

PR #3201, opened 2024-04-11
OCPBUGS-32183: catalog-operator: delete catalog pods stuck in Terminating state due to unreachable node

joelanford:

**Architectural changes:**

**Testing remarks:**

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/registry/reconciler/grpc.go
+++ b/pkg/controller/registry/reconciler/grpc.go
@@ -2,19 +2,22 @@ package reconciler
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"slices"
 	"strings"
 	"time"
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/operator-framework/api/pkg/operators/v1alpha1"
-	"github.com/pkg/errors"
+	pkgerrors "github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/utils/ptr"
 
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/install"
 	controllerclient "github.com/operator-framework/operator-lifecycle-manager/pkg/lib/controller-runtime/client"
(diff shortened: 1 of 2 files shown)


pkg/controller/registry/reconciler/grpc.go
```diff
@@ -338,13 +339,30 @@ func isRegistryServiceStatusValid(source *grpcCatalogSourceDecorator) (bool, err
 }
 
 func (c *GrpcRegistryReconciler) ensurePod(logger *logrus.Entry, source grpcCatalogSourceDecorator, serviceAccount *corev1.ServiceAccount, overwrite bool) error {
-	// currentLivePods refers to the currently live instances of the catalog source
-	currentLivePods := c.currentPods(logger, source)
-	if len(currentLivePods) > 0 {
+	// currentPods refers to the current pod instances of the catalog source
+	currentPods := c.currentPods(logger, source)
+
+	tmpPods := currentPods[:0]
```
perdasilva, line 345: nifty! didn't know about this one

joelanford: Actually, this reminds me there's a new kid on the block (`slices.DeleteFunc`). I'll switch to that as the intent is more clear.

grokspawn:
would love to better understand the need for the errors --> pkgerrors change, to know if that's a pattern we should look for elsewhere.

joelanford:
`errors` is the name of a very popular package in the Go standard library.

This package was authored prior to standard library Go having `fmt.Errorf("blah blah: %w", err)`. Before Go supported that, `pkg/errors.Wrapf(err, "message")` was a popular way to wrap errors.

I made this change to give the `errors` name to stdlib `errors` over the third-party errors package.

pkg/controller/registry/reconciler/grpc.go
```diff
@@ -523,6 +545,29 @@ func imageChanged(logger *logrus.Entry, updatePod *corev1.Pod, servingPods []*co
 	return false
 }
 
+func isPodDead(pod *corev1.Pod) bool {
+	for _, check := range []func(*corev1.Pod) bool{
```
grokspawn, line 549: This is a clever way to provide an extensible approach to a series of checks... but do we need it?  It seems we would have the same benefits by doing `s/isPodDead/IsPodDeletedByTaintManager/g`

joelanford: I'm somewhat confident we'll find another way for pods to be dead. We've seen similar issues in `operator-lib` leader for life.

So I figured I'd make things super easy for ourselves next time around.

tmshort: Yeah, do we think we'll need to expand this in the future?

grokspawn: Well, I ain't agin' it, but I generally like to supply it when it's needed. 
/lgtm
from me

tmshort: I'm not objecting... just asking.

PR #3202, opened 2024-04-12
Update if AlreadyExists

dtfranz:
Adds code to the client calls for Create to run an Update if the object already exists. This allows us to bypass issues where the object was not found in cache but already exists in the cluster.

### DIFF
--- a/pkg/controller/bundle/bundle_unpacker.go
+++ b/pkg/controller/bundle/bundle_unpacker.go
@@ -734,6 +734,9 @@ func (c *ConfigMapUnpacker) ensureRole(cmRef *corev1.ObjectReference) (role *rba
 	if err != nil {
 		if apierrors.IsNotFound(err) {
 			role, err = c.client.RbacV1().Roles(fresh.GetNamespace()).Create(context.TODO(), fresh, metav1.CreateOptions{})
+			if apierrors.IsAlreadyExists(err) {
+				role, err = c.client.RbacV1().Roles(fresh.GetNamespace()).Update(context.TODO(), fresh, metav1.UpdateOptions{})
+			}
 		}
 
 		return
@@ -778,6 +781,9 @@ func (c *ConfigMapUnpacker) ensureRoleBinding(cmRef *corev1.ObjectReference) (ro
 	if err != nil {
 		if apierrors.IsNotFound(err) {
 			roleBinding, err = c.client.RbacV1().RoleBindings(fresh.GetNamespace()).Create(context.TODO(), fresh, metav1.CreateOptions{})
+			if apierrors.IsAlreadyExists(err) {
+				roleBinding, err = c.client.RbacV1().RoleBindings(fresh.GetNamespace()).Update(context.TODO(), fresh, metav1.UpdateOptions{})
+			}
 		}
 
 		return
--- a/pkg/controller/operators/operatorcondition_controller.go
+++ b/pkg/controller/operators/operatorcondition_controller.go
@@ -150,7 +150,10 @@ func (r *OperatorConditionReconciler) ensureOperatorConditionRole(operatorCondit
 		if !apierrors.IsNotFound(err) {
 			return err
 		}
-		return r.Client.Create(context.TODO(), role)
+		err = r.Client.Create(context.TODO(), role)
+		if apierrors.IsAlreadyExists(err) {
+			return r.Client.Update(context.TODO(), role)
+		}
 	}
 
 	if ownerutil.IsOwnedBy(existingRole, operatorCondition) &&
@@ -199,7 +202,10 @@ func (r *OperatorConditionReconciler) ensureOperatorConditionRoleBinding(operato
 		if !apierrors.IsNotFound(err) {
 			return err
 		}
-		return r.Client.Create(context.TODO(), roleBinding)
+		err = r.Client.Create(context.TODO(), roleBinding)
+		if apierrors.IsAlreadyExists(err) {
+			return r.Client.Update(context.TODO(), roleBinding)
+		}
 	}
 
 	if ownerutil.IsOwnedBy(existingRoleBinding, operatorCondition) &&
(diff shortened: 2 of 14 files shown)


pkg/lib/operatorclient/apiservice.go
```diff
@@ -12,7 +13,11 @@ import (
 
 // CreateAPIService creates the APIService.
 func (c *Client) CreateAPIService(ig *apiregistrationv1.APIService) (*apiregistrationv1.APIService, error) {
-	return c.ApiregistrationV1Interface().ApiregistrationV1().APIServices().Create(context.TODO(), ig, metav1.CreateOptions{})
+	createdAS, err := c.ApiregistrationV1Interface().ApiregistrationV1().APIServices().Create(context.TODO(), ig, metav1.CreateOptions{})
+	if err != nil && apierrors.IsAlreadyExists(err) {
```
stevekuznetsov, line 17: nit: no need to check for `!= nil`, must be true if it's an already exists err

tmshort:
Would it make sense to add the `olm.managed` label to make sure everything is labeled everywhere we create resources?

tmshort:
> @tmshort every create call should already be setting that in the object payload.

"should"

It would mean that we guarantee adding the label, without having to remember everywhere... just a thought. Not necessary for this PR.

stevekuznetsov:
@tmshort we spent quite a lot of time making sure it was present everywhere. If you have any examples of that not being the case, please open a PR. Locality of behavior is a nice property - if you're looking at code that creates an object, you can see what that object looks like without spelunking into hack-ey wrapper bits. Respectfully, I understand where you're coming from but I don't think we have any evidence that we missed any spots where the label should have been present in a create call.

kuiwang02:
@dtfranz do we need to change code of CreateServiceAccount in serviceaccount.go to handle `already exist`? thanks

dtfranz:
@kuiwang02 I did this initially but adding that behavior universally to `CreateServiceAccount` caused an infinite creation loop for `CatalogSource` secrets, SAs, and pods because the registry code could not properly respond to `AlreadyExists` errors. For that reason `ServiceAccounts` and also a few other places in the code have been modified to use simple `Create()` calls because they already handle `AlreadyExists` errors in their own unique ways.

pkg/controller/bundle/bundle_unpacker.go
```diff
@@ -734,6 +734,9 @@ func (c *ConfigMapUnpacker) ensureRole(cmRef *corev1.ObjectReference) (role *rba
 	if err != nil {
 		if apierrors.IsNotFound(err) {
 			role, err = c.client.RbacV1().Roles(fresh.GetNamespace()).Create(context.TODO(), fresh, metav1.CreateOptions{})
+			if apierrors.IsAlreadyExists(err) {
```
dtfranz, line 737: `bundle_unpacker.go` does not have the wrapper client; I made the changes this way instead of spending the time to plumb the client in to keep the PR simpler and spend time on more important changes.

pkg/controller/operators/operatorcondition_controller.go
```diff
@@ -150,7 +150,10 @@ func (r *OperatorConditionReconciler) ensureOperatorConditionRole(operatorCondit
 		if !apierrors.IsNotFound(err) {
 			return err
 		}
-		return r.Client.Create(context.TODO(), role)
+		err = r.Client.Create(context.TODO(), role)
```
dtfranz, line 153: As in `bundler_unpacker.go`, client wrapper is not available here.

pkg/lib/operatorclient/apiservice.go
```diff
@@ -12,7 +13,11 @@ import (
 
 // CreateAPIService creates the APIService.
 func (c *Client) CreateAPIService(ig *apiregistrationv1.APIService) (*apiregistrationv1.APIService, error) {
-	return c.ApiregistrationV1Interface().ApiregistrationV1().APIServices().Create(context.TODO(), ig, metav1.CreateOptions{})
+	createdAS, err := c.ApiregistrationV1Interface().ApiregistrationV1().APIServices().Create(context.TODO(), ig, metav1.CreateOptions{})
```
dtfranz, line 16: Client wrapper also not available here.

pkg/lib/operatorclient/customresources.go
```diff
@@ -104,7 +105,11 @@ func (c *Client) CreateCustomResourceRawIfNotFound(apiGroup, version, namespace,
...
 	err = c.CreateCustomResourceRaw(apiGroup, version, namespace, kind, data)
-	if err != nil {
+	if apierrors.IsAlreadyExists(err) {
+		if err = c.UpdateCustomResourceRaw(apiGroup, version, namespace, kind, name, data); err != nil {
+			return false, err
+		}
+	} else if err != nil {
 		return false, err
 	}
 	return true, nil
```
tmshort, line 115: Nit, This could be:
```
return err == nil, err
```
For most, if not all, of the returns.

pkg/lib/operatorclient/deployment.go
```diff
@@ -25,10 +25,15 @@ func (c *Client) GetDeployment(namespace, name string) (*appsv1.Deployment, erro
...
 }
 
-// CreateDeployment creates the Deployment object.
+// CreateDeployment creates the Deployment object or Updates if it already exists.
 func (c *Client) CreateDeployment(dep *appsv1.Deployment) (*appsv1.Deployment, error) {
 	klog.V(4).Infof("[CREATE Deployment]: %s:%s", dep.Namespace, dep.Name)
-	return c.AppsV1().Deployments(dep.Namespace).Create(context.TODO(), dep, metav1.CreateOptions{})
+	createdDep, err := c.AppsV1().Deployments(dep.Namespace).Create(context.TODO(), dep, metav1.CreateOptions{})
+	if apierrors.IsAlreadyExists(err) {
+		updatedDep, _, err := c.UpdateDeployment(dep)
```
tmshort, line 33: Three returns? Compared to the others that just have two? (Just whining)

dtfranz:
Removed hold comment from above but I guess it needs a whole new comment.
/unhold

PR #3204, opened 2024-04-17
Fix unpack job cache issue

kevinrizza:
In the same vein as https://github.com/operator-framework/operator-lifecycle-manager/pull/3202, use update if the unpack job already exists but isn't cached


grokspawn:
[here](https://github.com/operator-framework/operator-lifecycle-manager/pull/3204/files#diff-83d16ba56fcfb7a91b215cb6fa5dadc24e04d25453bd8dfbe80a81840bdf2596R690) too?

joelanford (approved):

PR #3206, opened 2024-04-19
OCPBUGS-29729: Updates default security context behavior for catalog source pods

perdasilva:
Before: by default CatalogSource would stamp out pods with the "legacy" security profile
Now: by default CatalogSource stamps out "restricted" security profile pods when in appropriately labeled namespaces and "legacy" otherwise

Helps users be secure-by-default on PSA clusters when deploying in  "restricted" namespaces (no change for older clusters)

**Architectural changes:**

**Testing remarks:**

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/registry/reconciler/configmap.go
+++ b/pkg/controller/registry/reconciler/configmap.go
@@ -25,8 +25,7 @@ import (
 // configMapCatalogSourceDecorator wraps CatalogSource to add additional methods
 type configMapCatalogSourceDecorator struct {
 	*v1alpha1.CatalogSource
-	Reconciler *ConfigMapRegistryReconciler
-	runAsUser  int64
+	runAsUser int64
 }
 
 const (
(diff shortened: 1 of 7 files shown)


pkg/controller/registry/reconciler/configmap.go
```diff
@@ -288,6 +288,11 @@ func (c *ConfigMapRegistryReconciler) EnsureRegistryServer(logger *logrus.Entry,
 	overwrite := source.Status.RegistryServiceStatus == nil
 	overwritePod := overwrite
 
+	podSecurityConfig, err := getNamespaceSecurityContextConfig(c.OpClient, catalogSource.GetNamespace())
```
joelanford, line 291: WDYT about renaming this everywhere throughout the stack to `defaultPodSecurityConfig` to make it a bit more clear that this could be overridden if it is explicitly set in the catsrc?

pkg/controller/registry/reconciler/grpc.go
```diff
@@ -273,8 +274,13 @@ func (c *GrpcRegistryReconciler) EnsureRegistryServer(logger *logrus.Entry, cata
 		return err
 	}
 
+	registryPodSecurityConfig, err := getNamespaceSecurityContextConfig(c.OpClient, catalogSource.GetNamespace())
```
joelanford, line 277: Same here (and down the stack) re: `defaultPodSecurityConfig`

m1kola (approved):
This looks reasonable to me, still might be a good idea to get more eyes on it.

pkg/controller/registry/reconciler/grpc.go
```diff
@@ -33,6 +34,18 @@ const (
...
+	SecurityContextConfig v1alpha1.SecurityConfig
+}
+
+type PodOptionFunc func(option *PodOption)
+
+func WithSecurityContextConfig(securityContextConfig v1alpha1.SecurityConfig) PodOptionFunc {
+	return func(option *PodOption) {
+		option.SecurityContextConfig = securityContextConfig
+	}
+}
```
joelanford, line 47: This naming is still something that I think could lead to confusion. This is the default security context config, right?

This name implies that whatever value is here is what will be used. But it isn't until we evaluate the catalog source that we decide what the actual value to use is.

I'm also having trouble understanding why in some cases we use the `options ...PodOptionFunc` parameter and in others we use the `podSecurityConfig v1alpha1.SecurityConfig` parameter.

In this case, my suggestions would be:
- drop the option func (my reasoning is: calling a function that takes this variadic parameter means that passing that parameter is optional, which means we need a default for the default, which just complicates things even more).
- use `defaultPodSecurityConfig` as the name of the variables and parameters everywhere, except where we finally decide the `securityContext` we're going to put into the pod.

bentito:
This PR is a positive step towards better handling of security contexts in the OLM pod reconciler, making it easier to manage and test security configurations comprehensively. LGTM

pkg/controller/registry/reconciler/grpc.go
perdasilva: see if that's better. I just really wanted to make as few changes as possible XD but you make good points.

joelanford (approved):

PR #3209, opened 2024-04-23
OCPBUGS-31522: Warn and allow CRD upgrade if validation fails but new CRD specifies a conversion strategy

everettraven:
- Updates the CRD Upgrade validation logic to warn when the validations for a CRD upgrade fails but the new CRD specifies a conversion strategy

- When the new version of the CRD specifies a conversion strategy it is assumed that this conversion strategy should be used to perform conversion. There is no guarantee existing conversion strategies match this conversion strategy and thus OLM should assume the conversion would result in successful validation and allow the upgrade through with a warning.

**Architectural changes:**

**Testing remarks:**

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -1585,9 +1585,7 @@ func (o *Operator) ensureSubscriptionCSVState(logger *logrus.Entry, sub *v1alpha
 }
 
 func (o *Operator) setIPReference(subs []*v1alpha1.Subscription, gen int, installPlanRef *corev1.ObjectReference) []*v1alpha1.Subscription {
-	var (
-		lastUpdated = o.now()
-	)
+	lastUpdated := o.now()
 	for _, sub := range subs {
 		sub.Status.LastUpdated = lastUpdated
 		if installPlanRef != nil {
@@ -2204,6 +2202,10 @@ func validateV1Beta1CRDCompatibility(dynamicClient dynamic.Interface, oldCRD *ap
 	return validateExistingCRs(dynamicClient, gr, validationsMap)
 }
 
+type ValidationError struct {
+	error
+}
+
 // validateExistingCRs lists all CRs for each version entry in validationsMap, then validates each using the paired validation.
 func validateExistingCRs(dynamicClient dynamic.Interface, gr schema.GroupResource, validationsMap map[string]*apiextensions.CustomResourceValidation) error {
 	for version, schemaValidation := range validationsMap {
@@ -2212,7 +2214,6 @@ func validateExistingCRs(dynamicClient dynamic.Interface, gr schema.GroupResourc
 		if err != nil {
 			return fmt.Errorf("error creating validator for schema version %s: %s", version, err)
 		}
-
 		gvr := schema.GroupVersionResource{Group: gr.Group, Version: version, Resource: gr.Resource}
 		crList, err := dynamicClient.Resource(gvr).List(context.TODO(), metav1.ListOptions{})
 		if err != nil {
@@ -2229,7 +2230,7 @@ func validateExistingCRs(dynamicClient dynamic.Interface, gr schema.GroupResourc
 				} else {
 					namespacedName = fmt.Sprintf("%s/%s", cr.GetNamespace(), cr.GetName())
 				}
-				return fmt.Errorf("error validating %s %q: updated validation is too restrictive: %v", cr.GroupVersionKind(), namespacedName, err)
+				return ValidationError{fmt.Errorf("error validating %s %q: updated validation is too restrictive: %v", cr.GroupVersionKind(), namespacedName, err)}
 			}
 		}
 	}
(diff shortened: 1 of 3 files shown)


pkg/controller/operators/catalog/operator.go
```diff
@@ -2204,6 +2202,10 @@ func validateV1Beta1CRDCompatibility(dynamicClient dynamic.Interface, oldCRD *ap
 	return validateExistingCRs(dynamicClient, gr, validationsMap)
 }
 
+type ValidationError struct {
+	error
+}
```
joelanford, line 2207: Can this be unexported?

everettraven: Sure

everettraven: Done in https://github.com/operator-framework/operator-lifecycle-manager/pull/3209/commits/81d1ed44802c5218563f17a15c7d4665db578f84

pkg/controller/operators/catalog/step.go
```diff
@@ -140,7 +140,21 @@ func (b *builder) NewCRDV1Step(client apiextensionsv1client.ApiextensionsV1Inter
 					currentCRD, _ := client.CustomResourceDefinitions().Get(context.TODO(), crd.GetName(), metav1.GetOptions{})
 					crd.SetResourceVersion(currentCRD.GetResourceVersion())
 					if err = validateV1CRDCompatibility(b.dynamicClient, currentCRD, crd); err != nil {
-						return fmt.Errorf("error validating existing CRs against new CRD's schema for %q: %w", step.Resource.Name, err)
+						vErr := &ValidationError{}
+						// if the conversion strategy in the new CRD is "None" OR the error is not a ValidationError
+						// return an error. This will catch and return any errors that occur unrelated to actual validation.
+						// For example, the API server returning an error when performing a list operation
+						if crd.Spec.Conversion == nil || crd.Spec.Conversion.Strategy == apiextensionsv1.NoneConverter || !errors.As(err, vErr) {
+							return fmt.Errorf("error validating existing CRs against new CRD's schema for %q: %w", step.Resource.Name, err)
```
joelanford, line 148: I think we need to explicitly check for Webhook Converter to switch to the warn logic. If another converter type is added in the future, I think that should be an error, not a warning. If it turns out that new converter type is something we can't evaluate we could opt that into the warning logic at that time.

pkg/controller/operators/catalog/step.go
```diff
@@ -140,7 +140,21 @@ func (b *builder) NewCRDV1Step(client apiextensionsv1client.ApiextensionsV1Inter
...
+						}
+						// If the conversion strategy in the new CRD is not "None" and the error that occurred
+						// is an error related to validation, warn that validation failed but that we are trusting
+						// that the conversion strategy specified by the author will successfully convert to a format
+						// that passes validation and allow the upgrade to continue
+						b.logger.Warnf("trusting conversion strategy detected in new CRD, but failed validation of existing CRs against new CRD's schema for %q: %s",
```
joelanford, line 154: Two things:
1. This is likely something that needs some pretty descriptive messaging. Can you try to explain a bit more about why this isn't being treated as an error.
2. I think we need to use an EventRecorder to emit the warning as a kubernetes event against the InstallPlan object.

joelanford: (But keep this. Logging to the olm-operator log is helpful too)

everettraven: Should be done in https://github.com/operator-framework/operator-lifecycle-manager/pull/3209/commits/209d78371dbc47cf4030ccf5221cf9e8468f362d

pkg/controller/operators/catalog/step.go
```diff
@@ -140,7 +144,27 @@ func (b *builder) NewCRDV1Step(client apiextensionsv1client.ApiextensionsV1Inter
...
+						// is an error related to validation, warn that validation failed but that we are trusting
+						// that the conversion strategy specified by the author will successfully convert to a format
+						// that passes validation and allow the upgrade to continue
+						warnTempl := `validation of existing CRs against new CRD's schema failed, but a webhook conversion strategy was specified.
+allowing the CRD upgrade to occur as we can't run the webhook, but is assumed that it will successfully convert existing CRs
+to a format that would have passed validation.
+
+CRD: %q
+Validation Error: %s
+`
```
joelanford, line 164:
```suggestion
						warnTempl := `Validation of existing CRs against the new CRD's schema failed, but a webhook conversion strategy was specified in the new CRD. The new webhook will only start after the bundle is upgraded, so we must assume that it will successfully convert existing CRs to a format that would have passed validation.

CRD: %q
Validation Error: %s
`
```

everettraven: Updated in https://github.com/operator-framework/operator-lifecycle-manager/pull/3209/commits/91b7f5fe469059b97bd3d0803605b912b96af344

dtfranz (approved):
LGTM!

PR #3213, opened 2024-04-26
Update to go v1.22

perdasilva:
Updates to go version 1.22

**Architectural changes:**

**Testing remarks:**

**Reviewer Checklist**


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/operator-framework/operator-lifecycle-manager
 
-go 1.21
+go 1.22
```
m1kola, line 3: I was expecting to see toolchain directive here too.

perdasilva: great call out! I didn't know about toolchains. Looking at the [documentation](https://go.dev/doc/toolchain#get), I think it says that if you don't specify a toolchain it will take tool chain for that version:

```
If the toolchain line is omitted, the module or workspace is considered to have an implicit toolchain goV line, where V is the Go version from the go line.

For example, a go.mod that says go 1.21.0 with no toolchain line is interpreted as if it had a toolchain go1.21.0 line.
```

Should we start updating the Go z-version and as well, do you think?

m1kola: Good to know about implicit toolchain. I asked because I think `go mod tidy` adds it by default.

I think explicit is better than implicit, but I think we can go ahead with this PR.

perdasilva: I agree. I'll make it explicit.

m1kola (approved):

PR #3217, opened 2024-05-03
[OCPBUGS-25341]: perform operator apiService certificate validity checks directly

ankitathomas:
Cert updates can occasionally fail silently, updating only the timestamps on the CSV without any changes to the underlying cert secret. This PR uses the cert expiry times directly to retry the refresh.

### DIFF
--- a/pkg/controller/install/certresources.go
+++ b/pkg/controller/install/certresources.go
@@ -157,6 +157,13 @@ func ServiceName(deploymentName string) string {
 	return deploymentName + "-service"
 }
 
+func HostnamesForService(serviceName, namespace string) []string {
+	return []string{
+		fmt.Sprintf("%s.%s", serviceName, namespace),
+		fmt.Sprintf("%s.%s.svc", serviceName, namespace),
+	}
+}
+
 func (i *StrategyDeploymentInstaller) getCertResources() []certResource {
 	return append(i.apiServiceDescriptions, i.webhookDescriptions...)
 }
@@ -223,13 +230,72 @@ func (i *StrategyDeploymentInstaller) CertsRotated() bool {
 	return i.certificatesRotated
 }
 
-func ShouldRotateCerts(csv *v1alpha1.ClusterServiceVersion) bool {
+// shouldRotateCerts indicates whether an apiService cert should be rotated due to being
+// malformed, invalid, expired, inactive or within a specific freshness interval (DefaultCertMinFresh) before expiry.
+func shouldRotateCerts(certSecret *corev1.Secret, hosts []string) bool {
 	now := metav1.Now()
-	if !csv.Status.CertsRotateAt.IsZero() && csv.Status.CertsRotateAt.Before(&now) {
+	caPEM, ok := certSecret.Data[OLMCAPEMKey]
+	if !ok {
+		// missing CA cert in secret
+		return true
+	}
+	certPEM, ok := certSecret.Data["tls.crt"]
+	if !ok {
+		// missing cert in secret
 		return true
 	}
 
+	ca, err := certs.PEMToCert(caPEM)
+	if err != nil {
+		// malformed CA cert
+		return true
+	}
+	cert, err := certs.PEMToCert(certPEM)
+	if err != nil {
+		// malformed cert
+		return true
+	}
+
+	// check for cert freshness
+	if !certs.Active(ca) || now.Time.After(CalculateCertRotatesAt(ca.NotAfter)) ||
+		!certs.Active(cert) || now.Time.After(CalculateCertRotatesAt(cert.NotAfter)) {
+		return true
+	}
+
+	// Check validity of serving cert and if serving cert is trusted by the CA
+	for _, host := range hosts {
+		if err := certs.VerifyCert(ca, cert, host); err != nil {
+			return true
+		}
+	}
 	return false
+}
+
+func (i *StrategyDeploymentInstaller) ShouldRotateCerts(s Strategy) (bool, error) {
+	strategy, ok := s.(*v1alpha1.StrategyDetailsDeployment)
+	if !ok {
+		return false, fmt.Errorf("attempted to install %s strategy with deployment installer", strategy.GetStrategyName())
+	}
+
+	hasCerts := map[string]struct{}{}
+	for _, c := range i.getCertResources() {
+		hasCerts[c.getDeploymentName()] = struct{}{}
+	}
+	for _, sddSpec := range strategy.DeploymentSpecs {
+		if _, ok := hasCerts[sddSpec.Name]; ok {
+			certSecret, err := i.strategyClient.GetOpLister().CoreV1().SecretLister().Secrets(i.owner.GetNamespace()).Get(SecretName(ServiceName(sddSpec.Name)))
+			if err == nil {
+				if shouldRotateCerts(certSecret, HostnamesForService(ServiceName(sddSpec.Name), i.owner.GetNamespace())) {
+					return true, nil
+				}
+			} else if apierrors.IsNotFound(err) {
+				return true, nil
+			} else {
+				return false, err
+			}
+		}
+	}
+	return false, nil
 }
 
 func CalculateCertExpiration(startingFrom time.Time) time.Time {
@@ -267,10 +333,7 @@ func (i *StrategyDeploymentInstaller) installCertRequirementsForDeployment(deplo
 	}
 
 	// Create signed serving cert
-	hosts := []string{
-		fmt.Sprintf("%s.%s", serviceName, i.owner.GetNamespace()),
-		fmt.Sprintf("%s.%s.svc", serviceName, i.owner.GetNamespace()),
-	}
+	hosts := HostnamesForService(serviceName, i.owner.GetNamespace())
 	servingPair, err := certGenerator.Generate(expiration, Organization, ca, hosts)
 	if err != nil {
 		logger.Warnf("could not generate signed certs for hosts %v", hosts)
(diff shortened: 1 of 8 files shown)


test/e2e/webhook_e2e_test.go
```diff
@@ -1169,3 +1181,70 @@ func newV1CRD(plural string) apiextensionsv1.CustomResourceDefinition {
...
+	Expect(err).Should(BeNil())
+
+	caDetails := &x509.Certificate{
+		SerialNumber: caSerial,
+		Subject: pkix.Name{
+			CommonName:   fmt.Sprintf("olm-selfsigned-%x", caSerial),
+			Organization: []string{install.Organization},
+		},
+		NotBefore:             time.Now(),
+		NotAfter:              time.Now(),
```
tmshort, line 1196: Wondering if you really want these times in the past...

pkg/controller/operators/olm/apiservices.go
```diff
@@ -93,17 +93,12 @@ func (a *Operator) checkAPIServiceResources(csv *v1alpha1.ClusterServiceVersion,
...
 		if err != nil {
 			logger.Warnf("could not convert APIService CA bundle to x509 cert")
 			errs = append(errs, err)
 			continue
 		}
-		if !certs.Active(ca) {
-			logger.Warnf("CA cert not active")
-			errs = append(errs, fmt.Errorf("found the CA cert is not active"))
-			continue
-		}
```
tmshort, line 106: We're no longer checking the certs (here and below)?

ankitathomas: I've combined the cert expiry checks with [freshness](https://github.com/operator-framework/operator-lifecycle-manager/pull/3217/files#diff-bc7b779225ba70ff0e1f63ff55cb2dd3d71b5ae4f96c158947266da9f8fde60dR274) so we don't have a state needing manual cert refreshes for already expired certs

pkg/controller/install/certresources.go
```diff
@@ -223,15 +230,74 @@ func (i *StrategyDeploymentInstaller) CertsRotated() bool {
...
+			return true
+		}
+	}
 	return false
 }
 
+func (i *StrategyDeploymentInstaller) ShouldRotateCerts(s Strategy) (bool, error) {
+	strategy, ok := s.(*v1alpha1.StrategyDetailsDeployment)
+	if !ok {
+		return false, fmt.Errorf("attempted to install %s strategy with deployment installer", strategy.GetStrategyName())
```
dtfranz, line 277: Could we make this error a little more descriptive? To me it's just saying "we attempted an install" rather than "install failed because: ..." From what I can tell, looks like a failure due to a missing strategy? But I'm not super familiar with this code.

pkg/controller/install/certresources.go
```diff
@@ -223,15 +230,74 @@ func (i *StrategyDeploymentInstaller) CertsRotated() bool {
...
+	}
+
+	hasCerts := map[string]struct{}{}
```
dtfranz, line 280: nitpick and not required: we're using `k8s.io/apimachinery/pkg/util/sets` in v1 and inconsistently in v0. Might make things a little cleaner but it's not a big deal to me.

pkg/controller/install/certresources.go
```diff
@@ -157,6 +158,13 @@ func ServiceName(deploymentName string) string {
 	return deploymentName + "-service"
 }
 
+func HostnamesForService(serviceName, namespace string) []string {
+	return []string{
+		fmt.Sprintf("%s.%s", serviceName, namespace),
+		fmt.Sprintf("%s.%s.svc", serviceName, namespace),
```
tmshort, line 164: not adding `.cluster` nor `.cluster.local`?

pkg/controller/install/certresources.go
```diff
@@ -223,15 +231,74 @@ func (i *StrategyDeploymentInstaller) CertsRotated() bool {
...
+		if hasCerts.Has(sddSpec.Name) {
+			certSecret, err := i.strategyClient.GetOpLister().CoreV1().SecretLister().Secrets(i.owner.GetNamespace()).Get(SecretName(ServiceName(sddSpec.Name)))
+			if err == nil {
+				if shouldRotateCerts(certSecret, HostnamesForService(ServiceName(sddSpec.Name), i.owner.GetNamespace())) {
+					return true, nil
+				}
+			} else if apierrors.IsNotFound(err) {
+				return true, nil
+			} else {
+				return false, err
```
tmshort, line 295: If there's some other error, what happens? This does not rotate the certificate.

pkg/controller/operators/olm/operator.go
```diff
@@ -2207,18 +2207,18 @@ func (a *Operator) transitionCSVState(in v1alpha1.ClusterServiceVersion) (out *v
 			return
 		}
 		if installErr := a.updateInstallStatus(out, installer, strategy, v1alpha1.CSVPhaseInstalling, v1alpha1.CSVReasonWaiting); installErr != nil {
-			// Re-sync if kube-apiserver was unavailable
-			if apierrors.IsServiceUnavailable(installErr) {
-				logger.WithError(installErr).Info("could not update install status")
-				syncError = installErr
-				return
-			}
```
tmshort, line 2215: Why was this moved lower? If there's an error, shouldn't we check it first?

tmshort (commented):
Just a few questions right now.
However, _and I realize this is pre-existing code_, I don't like creating a certificate before checking it. The process should really be:
1. Does the cert exist? No -> create the cert
2. Is the cert need to be rotated? Yes -> rotate/create the cert

Right now, a cert is created _every time_ the rotate check is to be done, and that is a **BIG** waste of CPU.

pkg/controller/operators/olm/operator.go
joelanford: I had a similar thought. Maybe we shouldn't timeout the install plan if the reason for things taking so long is that KAS is unavailable.

pkg/controller/install/certresources.go
```diff
@@ -223,15 +231,74 @@ func (i *StrategyDeploymentInstaller) CertsRotated() bool {
...
+	// Check validity of serving cert and if serving cert is trusted by the CA
+	for _, host := range hosts {
+		if err := certs.VerifyCert(ca, cert, host); err != nil {
+			return true
+		}
+	}
 	return false
 }
 
+func (i *StrategyDeploymentInstaller) ShouldRotateCerts(s Strategy) (bool, error) {
```
joelanford, line 275: (Similar to @tmshort's comment about existing code and reluctance to change it)...

It looks like this function is setup as an all or nothing thing? If _any_ of the certs needs to be rotated, we're going to rotate all of them?

In practice, maybe this doesn't actually make a difference because all of the certs are created initially at the same time with the same expiration. Thus generally all or none will need rotating?

pkg/controller/install/certresources.go
```diff
@@ -223,15 +231,74 @@ func (i *StrategyDeploymentInstaller) CertsRotated() bool {
...
+	strategy, ok := s.(*v1alpha1.StrategyDetailsDeployment)
+	if !ok {
+		return false, fmt.Errorf("failed to install %s strategy with deployment installer: unsupported deployment install strategy", strategy.GetStrategyName())
+	}
+
+	hasCerts := sets.NewString()
```
joelanford, line 281: I believe this is deprecated. The `sets` package now supports/prefers the generic type.
```suggestion
	hasCerts := sets.New[string]()
```

pkg/controller/install/certresources.go
joelanford: I assume it means we enter exponential backoff (at least until we hit the 5m install plan timeout)?

Any other error is likely to be either temporary (in which case it's hopefully resolved next time around), or not (in which case we would likely fail in a similar way trying to rotate the cert).

The only scenario that is coming to mind where `Get` secret fails, but a subsequent `Create` or `Update` would succeed would be if we lack read permission but have write permission.

pkg/controller/install/certresources.go
ankitathomas: That's just the test to see if we need to transition the CSV from a Success/Failed state, there's a separate check before deciding to reuse/refresh certs for each deployment separately [here](https://github.com/operator-framework/operator-lifecycle-manager/pull/3217/files#diff-bc7b779225ba70ff0e1f63ff55cb2dd3d71b5ae4f96c158947266da9f8fde60dR381)

pkg/controller/install/certresources.go
ankitathomas: These should resolve even without the cluster domain specified, though we may need to test for the current cluster domain before adding `.cluster` or `.cluster.local` in case it's different from the default.

ankitathomas:
>     * re-reconcile if the cert secret changes
We already do this for all [olm managed secrets](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/controller/operators/olm/operator.go#L430)

>     * re-reconcile at a time in the future base on the current secrets' expiration time.
With each reconcile, we check all certs for anything that expires in a day or less and rotate all of those, including the ones that are already expired. The problem was that we were checking the cert freshness timestamps on the CSV to make those checks, and those were sometimes being incorrectly updated when the cert rotate hadn't really succeeded.

pkg/controller/install/certresources.go
joelanford: Got it. Thanks!

joelanford:
> With each reconcile, we check all certs for anything that expires in a day or less and rotate all of those, including the ones that are already expired.

But I'm wondering if there's a scenario where:
1. I install an operator that needs a cert
2. OLM creates the secrets, sets expiration to `now + N`
3. Everything reconciles, steady state is achieved, no further changes are made to the CSV.
4. `N` time passes, and the certs expire. Still no changes, so still no cert renewal?

Is there something that forces a re-reconcile inside the time window where:
- The certs are not yet expired, but
- They are close enough to expiration that we'll rotate them.

tmshort (approved):
/lgtm

Although we may want to look at the order of operations when a cert is to be rotated. As, this could lead to CPU spikes.

PR #3222, opened 2024-05-05
[WIP] clean-up: remove e2e environment provisioning

perdasilva:
Originally, the e2e test binary could be compiled with "helm" and "kind" tags enabling it to provision a kind cluster and install OLM before running the test suite. This PR removes this functionality from the e2e test suite and moves the provisioning/setup to the Makefile. Additionally, it adds bingo to help manage the tools the build process needs (kind, helm, setup_envtest, etc.).

In order to make this work, I've had to remove parallelization of the test run (make E2E_NODES=1). The reason is having tests run in parallel seems to cause flakes. It didn't happen before because each test process would provision and install its own cluster (i.e. have its own cluster to run the tests against == no noisy neighbor).

This PR also: updates kind to v0.22.0 and sets up kind with a node image comparable to the version of client-go (i.e. we're no longer just testing against the default node image)

The e2e test suite is not the right place to maintain provisioning and installation code. 

**Architectural changes:**
e2e test binary no longer provisions clusters and installs OLM (Makefile does that now)

**Testing remarks:**

**Reviewer Checklist**


.github/workflows/e2e-tests.yml
```diff
@@ -13,15 +13,15 @@ jobs:
...
-        parallel-id: [0, 1, 2, 3]
+        parallel-id: [0, 1, 2, 3, 4, 5, 6, 7]
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-go@v5
         with:
           go-version-file: "go.mod"
       - run: mkdir -p artifacts
-      - run: make e2e-local E2E_TEST_CHUNK=${{ matrix.parallel-id }} E2E_TEST_NUM_CHUNKS=${{ strategy.job-total }} E2E_NODES=2 ARTIFACT_DIR=./artifacts/ SKIP='\[FLAKE\]'
```
tmshort, line 24: Why is `E2E_NODES` decreased from 2 to 1?

test/e2e/split/main.go
```diff
@@ -145,7 +143,7 @@ func createChunkRegexp(numChunks, printChunk int, specs []string) (string, error
 	// Create chunks of size ceil(number of specs/number of chunks) in alphanumeric order.
 	// This is deterministic on inputs.
 	chunks := make([][]string, numChunks)
-	interval := int(math.Ceil(float64(numSpecs) / float64(numChunks)))
```
tmshort, line 148: Removal of `math.Ceil()`?
May want to update the comment.

tmshort:
While there isn't a lot of data, previously, when there were 4 e2e jobs (0 1 2 3), the average time to complete them was 20 minutes (range was from 18 to 23 minutes).
Looking at the timings here, we have e2es taking a much longer time on average, and a very long pole: [7 39 24 8 44 12 5 7].
If the goal was to reduce the overall time it took to perform the tests, this PR has failed.
Could this be due to `E2E_NODES` being reduced?
Logically, splitting the test into 8 with 1 node each, uses the same resources as 4 splits using 2 nodes. But the timings don't reflect any benefit.

perdasilva:
The goal is to remove cluster provisioning and installation from the e2e suite. Doesn't make sense for it to be doing all of that. The e2e suite doesn't run well in parallel. The reason it worked before was because each test process would provision it's own test cluster. Without that, the specs tread all over each other.

I'm still playing with figuring out how to bring those times down. There's a lot of variability between the specs in terms of runtime. Might just separate out the super long ones and distribute the others.

test/e2e/split/main.go
perdasilva: I'm going to revert this. Was experimental.

.github/workflows/e2e-tests.yml
perdasilva: Having tests run in parallel seems to cause flakes. It didn't happen before because each test process would provision and install its own cluster (i.e. have its own cluster to run the tests against == no noisy neighbor).

perdasilva:
trying to understand which specs have issues running in // and fixing them

PR #3262, opened 2024-05-17
 OCPBUGS-32439: Fix: Ensure jobs without unpack label are considered

varshaprasad96:
If an unpacked job without the relevant unpack label exists, then proceed with the install steps. 

In previous versions of OLM, the unpack label in jobs was not set. Which means - if there is a job existing on cluster which already has unpacked contents, and OLM tries installing the same operator, we error out.

**Architectural changes:**

**Testing remarks:**

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/bundle/bundle_unpacker.go
+++ b/pkg/controller/bundle/bundle_unpacker.go
@@ -669,6 +669,19 @@ func (c *ConfigMapUnpacker) ensureJob(cmRef *corev1.ObjectReference, bundlePath 
 	if err != nil {
 		return
 	}
+
+	// This is to ensure that we account for any existing unpack jobs that may be missing the label
+	jobWithoutLabel, err := c.jobLister.Jobs(fresh.GetNamespace()).Get(cmRef.Name)
+	if err != nil && !apierrors.IsNotFound(err) {
+		return
+	}
+	if jobWithoutLabel != nil {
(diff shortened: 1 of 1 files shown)


varshaprasad96:
cc: @dtfranz could you please take a look?

dtfranz:
I'm assuming the `olmlog.txt` file was accidental? :smile:

pkg/controller/bundle/bundle_unpacker.go
```diff
@@ -669,6 +669,19 @@ func (c *ConfigMapUnpacker) ensureJob(cmRef *corev1.ObjectReference, bundlePath
 	if err != nil {
 		return
 	}
+
+	// This is to ensure that any unpack job that was successful but did not contain the label was taken into account.
```
dtfranz, line 673: nitpick, since existing jobs could also potentially be failed:
```suggestion
	// This is to ensure that we account for any existing unpack jobs that may be missing the label
```

varshaprasad96:
Addressed the changes. The previous e2e failures didn't seem to be related to the changes, hoping that it passes CI.

perdasilva:
Any unit tests we can add to make sure we don't regress?

perdasilva (approved):
spoke to Dev, there will be follow up PRs for test. Need this in master to unblock.

PR #3275, opened 2024-05-23
:seedling: parameterizes the e2e target and pins opm images

perdasilva:
The previous refactored parametrization of the e2e target was too restrictive. This PR re-introduces parameters to configure the e2e test command.

It has also come to our attention that opm:latest images are being used in e2e tests. This PR also pins the opm images to the relevant version of operator-registry

**Architectural changes:**

**Testing remarks:**

**Reviewer Checklist**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -163,7 +163,7 @@ deploy:
 		--set olm.image.pullPolicy=IfNotPresent \
 		--set catalog.image.ref=$(OLM_IMAGE) \
 		--set catalog.image.pullPolicy=IfNotPresent \
-		--set catalog.commandArgs=--configmapServerImage=$(CONFIGMAP_SERVER_IMAGE)\
+		--set catalog.commandArgs=--configmapServerImage=$(CONFIGMAP_SERVER_IMAGE) \
 		--set catalog.opmImageArgs=--opmImage=$(OPERATOR_REGISTRY_IMAGE) \
 		--set package.image.ref=$(OLM_IMAGE) \
 		--set package.image.pullPolicy=IfNotPresent \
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -144,21 +157,22 @@ kind-create: kind-clean
...
+		--set catalog.opmImageArgs=--opmImage=$(OPERATOR_REGISTRY_IMAGE) \
 		--set package.image.ref=$(OLM_IMAGE) \
 		--set package.image.pullPolicy=IfNotPresent \
 		$(HELM_INSTALL_OPTS) \
 		--wait;
 
 .PHONY: e2e-build
 e2e-build: BUILD_TAGS="json1 e2e experimental_metrics"
-e2e-build: export GOOS=linux
-e2e-build: export GOARCH=amd64
```
tmshort, line 161: Are you sure you don't need the `export`? We added that recently because it was causing problems with test builds on macs

Makefile
```diff
@@ -144,21 +157,22 @@ kind-create: kind-clean
 OLM_IMAGE := quay.io/operator-framework/olm:local
 deploy:
 	$(KIND) load docker-image $(OLM_IMAGE) --name $(KIND_CLUSTER_NAME); \
-	$(HELM) install olm deploy/chart \
+	$(HELM) upgrade --install olm deploy/chart \
```
everettraven, line 160: Could you elaborate on why this change was needed?

Makefile
perdasilva: I think we can drop the GOARCH as that could screw up M-class cpu mac people
I'll fix GOOS tho =D

Makefile
perdasilva: it was just convenient during development - if I just wanted to deploy and skip the whole kind create step

deploy/chart/values.yaml
```diff
@@ -40,6 +40,7 @@ catalog:
   setWorkloadUserID: true
   replicaCount: 1
   commandArgs: --configmapServerImage=quay.io/operator-framework/configmap-operator-registry:latest
+  opmImageArgs: --opmImage=quay.io/operator-framework/opm:latest
```
everettraven, line 43: I'm not very familiar with this code base, could you provide any resources about how this is injecting this image as the opm image to use?

perdasilva: It's [this](https://github.com/operator-framework/operator-lifecycle-manager/pull/3275/commits/af279d96d68467c3808fc33e6aaaa4b5b7632ae0#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R257) bit that does the injection. This just sets a reasonable default for the value.

everettraven: Apologies, I think I could have phrased this better - Where is this `--opmImage` flag defined and where is the provided value used within OLM?

By no means should this comment block this PR, but since I'm still relatively new to the codebase the additional context would be helpful for me :)

everettraven (approved):
Everything looks good to me from a code change perspective, but I'm not to familiar with this code base and how effective this change is. If we trust our e2e tests then LGTM :)

tmshort:
unit tests are now failing :(

deploy/chart/values.yaml
perdasilva: All good, man. I'm learning myself 😂

This flag is defined by the catalog operator binary (in the cmd package)

perdasilva:
> unit tests are now failing :( And the only change was the export‽‽

Flake (already investing XD - I'll get this repo stable if it kills me!)

PR #3285, opened 2024-05-27
:seedling: cleanup/tools

perdasilva:
This PR moves tools that are orthogonal to OLM (e.g. kind, helm, golangci-lint, etc.) to .bingo. It only keeps certain necessary tools in `tools.go`

Running into an issue where the helm version is skewing with our containerd version. So, the time has come to try to depollute the vendor directory.

**Architectural changes:**

**Testing remarks:**

**Reviewer Checklist**


.golangci.yaml
```diff
@@ -15,11 +15,6 @@ linters:
   - errcheck
 
 linters-settings:
-  # This stanza is needed to avoid getting the warning
```
perdasilva, line 18: golangci-lint got updated in the meantime and this is no longer an issue =D

Makefile
```diff
@@ -13,6 +13,40 @@ $(warning Undefining GOFLAGS set in CI)
...
+# a) we need some resource (e.g. yaml file)
+# b) need to run a script
+# c) need to track tools closely with OLM
+# the tools belongs in tools.go
+TOOL_EXEC := GO111MODULE=on go run -mod=vendor
+GINKGO := $(TOOL_EXEC) github.com/onsi/ginkgo/v2/ginkgo
+
+.PHONY: g
+g:
+	$(GINKGO)
```
m1kola, line 48: I'm not following - why do we need this one?

Makefile
```diff
@@ -13,6 +13,40 @@ $(warning Undefining GOFLAGS set in CI)
...
+#  b) the tool's version doesn't need to track closely with OLM
+# the tool goes here
+include .bingo/Variables.mk
+
+# tools.go manages type 2 tools. If
+# a) we need some resource (e.g. yaml file)
+# b) need to run a script
+# c) need to track tools closely with OLM
+# the tools belongs in tools.go
+TOOL_EXEC := GO111MODULE=on go run -mod=vendor
```
m1kola, line 43: Is `GO111MODULE=on` still necessary

Makefile
```diff
@@ -13,6 +13,40 @@ $(warning Undefining GOFLAGS set in CI)
...
+#   - golangci-lint
+#   - helm
+#   - kind
+#   - setup-envtest
+#   - yq
+# 2. tools.go: imports modules that need to be vendored and/or have a low tolerance for skew with the main module
+#              or because they are already imported by the main module, but we want to pull down the whole dependency
+#              to run some cmd or script, or get some resource (e.g. CRD yaml files).
+#   - OLM API CRDs
+#   - Test libraries, e.g. ginkgo
```
m1kola, line 29: As far as I see we do not have ginkgo in `tools.go` - only here in `Makefile`.

Any reason why not to have it managed by bingo?

m1kola (commented):
A few comments.

Makefile
perdasilva: oh whoops - that was a test target!!

Makefile
perdasilva: I forgot I removed it to see if it was necessary XD
I was thinking it would be good to execute the tests with the same library version as what is in go.mod. That's why I decided to not bingo it. I'd like to have a few things as possible in bingo because dependabot doesn't track those for updates =P (although I'm sure we could come up with a bot of our own)

m1kola: @perdasilva I see the point about dependabot. But right now we do `go run` in `Makefile` - I think it will have the same effect since on dependabot since ginkgo won't end up in go.mod. Or am I missing something?

perdasilva: ginkgo is in go.mod anyway because we import it in the e2e and for some unit tests - so I think dependabot should pick it up right? So, we could have skew between bingo gingko and olm ginkgo

m1kola (approved):
/lgtm

PR #3297, opened 2024-05-29
Unpack Job Creation Failure Test

dtfranz:
Extends an existing unit test to cover the scenario where an old, failed job missing the label added in later OLM versions is not found with the filtered list call, resulting in an 'AlreadyExists' error when the new job is created.

Addresses this comment:
https://github.com/operator-framework/operator-lifecycle-manager/pull/3262#issuecomment-2121996915

Test **fail** before #3262 merge [here](https://github.com/operator-framework/operator-lifecycle-manager/actions/runs/9290731611/job/25567634441?pr=3297)

Test **pass** after merge [here](https://github.com/operator-framework/operator-lifecycle-manager/actions/runs/9292499460/job/25573528070?pr=3297)

### DIFF
--- a/pkg/controller/bundle/bundle_unpacker_test.go
+++ b/pkg/controller/bundle/bundle_unpacker_test.go
@@ -1218,14 +1218,15 @@ func TestConfigMapUnpacker(t *testing.T) {
 			},
 		},
 		{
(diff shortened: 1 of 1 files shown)


perdasilva (approved):
lgtm ^^ thank you!!

PR #3301, opened 2024-05-31
:seedling: patch installplan e2e

perdasilva:
This particular installplan e2e checks that updating to a package with fewer authorizations leads to an attenuated service account. Previously, the test would also check that the number secrets in the namespace wouldn't change. This PR removes that check because I don't think it is material to the test. What happens to the secrets (which I think are created by ocp?) isn't important. What is is that after the first csv is installed the SA is allowed to do something and after the upgrade it isn't.

Reduce flakes and false negatives

**Architectural changes:**

**Testing remarks:**

**Reviewer Checklist**


tmshort (approved):
Other than the mixing of `Ginkgo` and `require`, which was already in the base file... _sigh..._

/lgtm

PR #3332, opened 2024-06-27
(fix) prevent hash key collision while generating bundle unpacking job name

anik120:


anik120:
/hold 

need more clarification to provide justification for the modifications being done in this PR.

tmshort:
The likelyhood of a cache collision with sha256 is quite low. Looking at the issues in question, this might be a caching problem rather than a collision problem. This will not resolve a caching problem.

PR #3349, opened 2024-07-26
(fix) Resolver: list CatSrc using client, instead of referring to registry-server cache

anik120:
Using "available CatalogSources" information from the registry-client cache was causing cache inconsistency problems.

This has showed up multiple times in production environments over the years, manifesting itself in the form of the all subscriptions in a namespace being transitioned into an error state when a Catalogsource that the cache claims to exist, has actually been deleted from the cluster, but the cache was not updated.

The Subscriptions are transitioned to an error state because of the deleted catalogsource with the follwing error message:

"message": "failed to populate resolver cache from source <deleted-catalogsource>: failed to list
bundles: rpc error: code = Unavailable desc = connection error: desc = \"transport:
Error while dialing dial tcp: lookup <deleted-catalogsource>.<ns>.svc on 172.....: no such host\"",
                "reason": "ErrorPreventedResolution",
                "status": "True",
                "type": "ResolutionFailed"

This PR switches the information lookup from the cache, to using a client to list the CatalogSources present in the cluster.

**Architectural changes:**

**Testing remarks:**

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/catalog/subscription/reconciler.go
+++ b/pkg/controller/operators/catalog/subscription/reconciler.go
@@ -91,11 +91,16 @@ func (c *catalogHealthReconciler) Reconcile(ctx context.Context, in kubestate.St
 
 				var healthUpdated, deprecationUpdated bool
 				next, healthUpdated = s.UpdateHealth(c.now(), catalogHealth...)
+				if healthUpdated {
+					if _, err := c.client.OperatorsV1alpha1().Subscriptions(ns).UpdateStatus(ctx, s.Subscription(), metav1.UpdateOptions{}); err != nil {
+						return nil, err
+					}
+				}
 				deprecationUpdated, err = c.updateDeprecatedStatus(ctx, s.Subscription())
 				if err != nil {
 					return next, err
 				}
-				if healthUpdated || deprecationUpdated {
+				if deprecationUpdated {
 					_, err = c.client.OperatorsV1alpha1().Subscriptions(ns).UpdateStatus(ctx, s.Subscription(), metav1.UpdateOptions{})
 				}
 			case SubscriptionExistsState:
--- a/pkg/controller/registry/resolver/cache/cache.go
+++ b/pkg/controller/registry/resolver/cache/cache.go
@@ -139,7 +139,7 @@ func (c *NamespacedOperatorCache) Error() error {
 		err := snapshot.err
 		snapshot.m.RUnlock()
 		if err != nil {
-			errs = append(errs, fmt.Errorf("failed to populate resolver cache from source %v: %w", key.String(), err))
+			errs = append(errs, fmt.Errorf("error using catalogsource %s/%s: %w", key.Namespace, key.Name, err))
 		}
 	}
 	return errors.NewAggregate(errs)
--- a/pkg/controller/registry/resolver/step_resolver.go
+++ b/pkg/controller/registry/resolver/step_resolver.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/operator-framework/api/pkg/operators/v1alpha1"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/clientset/versioned"
+	v1listers "github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/listers/operators/v1"
 	v1alpha1listers "github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/listers/operators/v1alpha1"
 	controllerbundle "github.com/operator-framework/operator-lifecycle-manager/pkg/controller/bundle"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/registry/resolver/cache"
@@ -16,10 +17,12 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 )
 
 const (
 	BundleLookupConditionPacked v1alpha1.BundleLookupConditionType = "BundleLookupNotPersisted"
+	exclusionAnnotation         string                             = "olm.operatorframework.io/exclude-global-namespace-resolution"
 )
 
 // init hooks provides the downstream a way to modify the upstream behavior
@@ -32,6 +35,7 @@ type StepResolver interface {
 type OperatorStepResolver struct {
 	subLister              v1alpha1listers.SubscriptionLister
 	csvLister              v1alpha1listers.ClusterServiceVersionLister
+	ogLister               v1listers.OperatorGroupLister
 	client                 versioned.Interface
 	globalCatalogNamespace string
 	resolver               *Resolver
@@ -69,6 +73,7 @@ func NewOperatorStepResolver(lister operatorlister.OperatorLister, client versio
 	stepResolver := &OperatorStepResolver{
 		subLister:              lister.OperatorsV1alpha1().SubscriptionLister(),
 		csvLister:              lister.OperatorsV1alpha1().ClusterServiceVersionLister(),
+		ogLister:               lister.OperatorsV1().OperatorGroupLister(),
 		client:                 client,
 		globalCatalogNamespace: globalCatalogNamespace,
 		resolver:               NewDefaultResolver(cacheSourceProvider, catsrcPriorityProvider{lister: lister.OperatorsV1alpha1().CatalogSourceLister()}, log),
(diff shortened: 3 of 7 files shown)


anik120:
Picking up https://github.com/operator-framework/operator-lifecycle-manager/pull/2779

anik120:
/hold 

I want to test this a little more before I open it up for reviews.

benluddy:
> This PR switches the information lookup from the **cache**, to using a client to list the CatalogSources present in the cluster.

Bolded to emphasize that the list of available catalogs is coming from a cache _of registry-server clients_, not a cache of CatalogSources. In other words, it wasn't reading stale cache entries, it was reading the wrong cache. Fixing this to actually be based on CatalogSources is a huge improvement.

pkg/controller/operators/catalog/subscription/reconciler.go
```diff
@@ -91,11 +91,13 @@ func (c *catalogHealthReconciler) Reconcile(ctx context.Context, in kubestate.St
 
 				var healthUpdated, deprecationUpdated bool
 				next, healthUpdated = s.UpdateHealth(c.now(), catalogHealth...)
-				deprecationUpdated, err = c.updateDeprecatedStatus(ctx, s.Subscription())
-				if err != nil {
-					return next, err
+				if healthUpdated {
+					if _, err := c.client.OperatorsV1alpha1().Subscriptions(ns).UpdateStatus(ctx, s.Subscription(), metav1.UpdateOptions{}); err != nil {
+						return nil, err
+					}
```
dtfranz, line 97: I was curious about the reasoning behind separating the `UpdateStatus` calls, then realized that making sure the health status update happens error-free before running the deprecation check+update is more important than potentially saving us an extra `UpdateStatus` call. Is that right?

anik120: @dtfranz so that is why I made this change while I was reading this code, but I've removed the change from this PR now. It go in by mistake 😅; it's an irrelevant change that should be a PR of it's own. Which will most likely happen soon since there's a bug that's been reported in the `update deprecated status` area, so I can make this change when I write up the PR for that bug.

dtfranz: Thanks for explaining! That makes sense, since this change is pretty tangential to this PR's objectives.

anik120:
@benluddy thanks for chiming in! It's been a fun read for a few days, but I think I have a clearer understanding of what's going on now. Many different caches at play here. I'm thinking it's probably a good idea to revisit https://github.com/operator-framework/operator-lifecycle-manager/pull/2337 and the efforts around it (probably https://github.com/operator-framework/operator-lifecycle-manager/pull/2632 too ). Will behoove us to devote some effort in reeling down the code base, for ease of maintenance going forward. 

A very simple: `1. list all the resources needed from the cluster, 2. convert them to the desired formats (Installables, or whatever the latest term is), 3. "Solve"`, without too many caches at play sounds much more maintainable. 

That's not for this PR though 😬 
Opening up the PR for reviews. 

/hold cancel

pkg/controller/operators/catalog/subscription/reconciler.go
anik120: @dtfranz actually, read through it again, and looks like it didn't get in by mistake 😅. The `Deprecated status` area runs into a "could not find registry server for CatalogSource problem"  in [these tests](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/test/e2e/subscription_e2e_test.go#L730-L1029) with the introduction of the changes in this PR. Possibly because of the use of the cache.SourceProvider in its implementation. 

I remember adding this change because it's the path of least resistance: we get the error surfaced to the Catalog status, as attested by the tests "Surface catalog health when there is a problem", and 

`
making sure the health status update happens error-free before running the deprecation check+update is more important than potentially saving us an extra UpdateStatus call
`

anik120: I added it back in PTAL

pkg/controller/registry/resolver/cache/cache.go
```diff
@@ -139,7 +139,7 @@ func (c *NamespacedOperatorCache) Error() error {
 		err := snapshot.err
 		snapshot.m.RUnlock()
 		if err != nil {
-			errs = append(errs, fmt.Errorf("failed to populate resolver cache from source %v: %w", key.String(), err))
+			errs = append(errs, fmt.Errorf("error using catalog %s (in namespace %s): %w", key.Name, key.Namespace, err))
```
bentito, line 142: This is for handling a deleted catalog, right? Presumably the deletion happened intentionally. At this point in the code we simply can't find it. Would it be better to Warn instead?

benluddy: No, if a transient network failure prevents the catalog operator from reaching a particular registry server, ignoring the error means also ignoring that we weren't able to fetch all of a catalog's content. That could result in different operators being installed.

bentito: wasn't talking about a logic change. just the log level, Error -> Warn. It's something we're going to retry or that possibly a catalog was intentionally removed. Was just thinking Warn might be more appropriate for the situation.

anik120: >This is for handling a deleted catalog, right?

@bentito not really. This is the resolver saying "I cannot use this catalog, for some reason or the other". [This test case](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/controller/registry/resolver/step_resolver_test.go#L1203-L1225) demonstrates that well. The catalog cannot be used in that case because "csv" (tbh on the first read through I don't exactly get what that means 😄, probably a needs to be worked on a bit to make it more clear, but the point is because of "3EntryReplacementChain" and "ReplacementChainBroken" and something something, the catalog cannot be used")

bentito: sounds like really should be reported as an error then.

pkg/controller/operators/catalog/subscription/reconciler.go
```diff
@@ -91,11 +91,16 @@ func (c *catalogHealthReconciler) Reconcile(ctx context.Context, in kubestate.St
...
+				}
 				deprecationUpdated, err = c.updateDeprecatedStatus(ctx, s.Subscription())
 				if err != nil {
 					return next, err
 				}
-				if healthUpdated || deprecationUpdated {
+				if deprecationUpdated {
```
tmshort, line 103: So, this case we don't check for error... but the health case we do now?

pkg/controller/registry/resolver/cache/cache.go
tmshort, line 142: This error is kinda word, would it make more sense to be the following?
```suggestion
			errs = append(errs, fmt.Errorf("error using catalogsource %s/%s: %w", key.Namespace, key.Name, err))
```
i.e. refer to the actual resource type (catalogsource) and also use namespace/name format

pkg/controller/operators/catalog/subscription/reconciler.go
anik120: Daniel and I had a conversation about this here https://github.com/operator-framework/operator-lifecycle-manager/pull/3349#discussion_r1705814505

pkg/controller/registry/resolver/cache/cache.go
tmshort: Noting that my suggestion was used below for another error you added.

pkg/controller/registry/resolver/step_resolver.go
```diff
@@ -16,10 +17,12 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 )
 
 const (
 	BundleLookupConditionPacked v1alpha1.BundleLookupConditionType = "BundleLookupNotPersisted"
+	exclusionAnnotation         string                             = "olm.operatorframework.io/exclude-global-namespace-resolution"
```
tmshort, line 25: So, this is now part of the "API"? Is this documented?

pkg/controller/registry/resolver/step_resolver_test.go
```diff
@@ -1218,7 +1218,7 @@ func TestResolver(t *testing.T) {
 				steps: [][]*v1alpha1.Step{},
 				subs:  []*v1alpha1.Subscription{},
 				errAssert: func(t *testing.T, err error) {
-					assert.Contains(t, err.Error(), "failed to populate resolver cache from source @existing/catsrc-namespace: csv")
+					assert.Contains(t, err.Error(), "error using catalog @existing (in namespace catsrc-namespace): csv")
```
tmshort, line 1221: If you use my suggestion above, this error message check will need to change as well.

pkg/controller/operators/catalog/subscription/reconciler.go
anik120: If you mean to say that this [error](https://github.com/operator-framework/operator-lifecycle-manager/pull/3349/files#diff-1452f55a2f3e0dd96fcb5d63a2a0de2244790bcd067225f4989044330006803aL99) is unchecked, I noticed it too, but left it alone since it isn't part of the PR. But since I'm in this area I could totally just add that in if that's what you were asking for.

pkg/controller/registry/resolver/cache/cache.go
anik120: That's fair, updated the PR with your suggestion.

pkg/controller/registry/resolver/step_resolver.go
anik120: It's coming from [this PR that added the feature](https://github.com/operator-framework/operator-lifecycle-manager/pull/2788)  (see [ref](https://github.com/operator-framework/operator-lifecycle-manager/pull/2788/files#diff-11677c62612a788aa62feb63e14540b299fbc9dcd8b35ca7e3576b5fd6058e82R29)). 

So it's an existing feature, I'm just moving the usage to a different area in this PR. However, now that you bring it up, fyi I also have an experiment going on [here](https://github.com/operator-framework/operator-lifecycle-manager/pull/3360) to essentially redo the PR that added this feature, using only the change I made in this PR.

anik120: Avoided removing the `og-source-provider` in this PR because it needs to be back ported and wanted to keep the scope small. But I think that makes a lot of sense as a follow up immediately.

pkg/controller/operators/catalog/subscription/reconciler.go
dtfranz: It's probably not clear because the diff cuts off the func definition, but the error is unchecked because this func uses named returns, and this is the last potential error-causing statement, so that error will be returned if there is one. I'm not defending it, I really don't like named returns, just explaining why it looks like that.

anik120: Ah no that's sounds fine to me. It's one of those "there's 2 ways of doing this and both are right", but it's not related to this PR.

dtfranz (approved):
/lgtm

PR #3366, opened 2024-08-14
(fix) registry pods do not come up again after node failure

anik120:

[PR 3201](https://github.com/operator-framework/operator-lifecycle-manager/pull/3201) attempted to solve for the issue by deleting the pods stuck in
`Terminating` due to unreachable node. However, the logic to do that was
included in `EnsureRegistryServer`, which only gets executed if polling in
requested by the user.

This PR moves the logic of checking for dead pods out of `EnsureRegistryServer`,
and puts it in `CheckRegistryServer` instead. This way, if there are any dead pods
detected during `CheckRegistryServer`, the value of `healthy` is returned `false`,
which inturn triggers `EnsureRegistryServer`.

**Architectural changes:**

**Testing remarks:**

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/registry/reconciler/configmap.go
+++ b/pkg/controller/registry/reconciler/configmap.go
@@ -3,11 +3,12 @@ package reconciler
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/install"
 	hashutil "github.com/operator-framework/operator-lifecycle-manager/pkg/lib/kubernetes/pkg/util/hash"
-	"github.com/pkg/errors"
+	pkgerrors "github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
@@ -15,6 +16,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/utils/ptr"
 
 	"github.com/operator-framework/api/pkg/operators/v1alpha1"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/operatorclient"
@@ -327,27 +329,27 @@ func (c *ConfigMapRegistryReconciler) EnsureRegistryServer(logger *logrus.Entry,
 
 	//TODO: if any of these error out, we should write a status back (possibly set RegistryServiceStatus to nil so they get recreated)
 	if err := c.ensureServiceAccount(source, overwrite); err != nil {
-		return errors.Wrapf(err, "error ensuring service account: %s", source.serviceAccountName())
+		return pkgerrors.Wrapf(err, "error ensuring service account: %s", source.serviceAccountName())
 	}
 	if err := c.ensureRole(source, overwrite); err != nil {
-		return errors.Wrapf(err, "error ensuring role: %s", source.roleName())
+		return pkgerrors.Wrapf(err, "error ensuring role: %s", source.roleName())
 	}
 	if err := c.ensureRoleBinding(source, overwrite); err != nil {
-		return errors.Wrapf(err, "error ensuring rolebinding: %s", source.RoleBinding().GetName())
+		return pkgerrors.Wrapf(err, "error ensuring rolebinding: %s", source.RoleBinding().GetName())
 	}
 	pod, err := source.Pod(image, defaultPodSecurityConfig)
 	if err != nil {
 		return err
 	}
 	if err := c.ensurePod(source, defaultPodSecurityConfig, overwritePod); err != nil {
-		return errors.Wrapf(err, "error ensuring pod: %s", pod.GetName())
+		return pkgerrors.Wrapf(err, "error ensuring pod: %s", pod.GetName())
 	}
 	service, err := source.Service()
 	if err != nil {
 		return err
 	}
 	if err := c.ensureService(source, overwrite); err != nil {
-		return errors.Wrapf(err, "error ensuring service: %s", service.GetName())
+		return pkgerrors.Wrapf(err, "error ensuring service: %s", service.GetName())
 	}
 
 	if overwritePod {
@@ -420,7 +422,7 @@ func (c *ConfigMapRegistryReconciler) ensurePod(source configMapCatalogSourceDec
 		}
 		for _, p := range currentPods {
 			if err := c.OpClient.KubernetesInterface().CoreV1().Pods(pod.GetNamespace()).Delete(context.TODO(), p.GetName(), *metav1.NewDeleteOptions(1)); err != nil && !apierrors.IsNotFound(err) {
-				return errors.Wrapf(err, "error deleting old pod: %s", p.GetName())
+				return pkgerrors.Wrapf(err, "error deleting old pod: %s", p.GetName())
 			}
 		}
 	}
@@ -428,7 +430,7 @@ func (c *ConfigMapRegistryReconciler) ensurePod(source configMapCatalogSourceDec
 	if err == nil {
 		return nil
 	}
-	return errors.Wrapf(err, "error creating new pod: %s", pod.GetGenerateName())
+	return pkgerrors.Wrapf(err, "error creating new pod: %s", pod.GetGenerateName())
 }
 
 func (c *ConfigMapRegistryReconciler) ensureService(source configMapCatalogSourceDecorator, overwrite bool) error {
@@ -512,6 +514,34 @@ func (c *ConfigMapRegistryReconciler) CheckRegistryServer(logger *logrus.Entry, 
 		return
 	}
 
-	healthy = true
-	return
+	podsAreLive, e := detectAndDeleteDeadPods(logger, c.OpClient, pods, source.GetNamespace())
+	if e != nil {
+		return false, fmt.Errorf("error deleting dead pods: %v", e)
+	}
+	return podsAreLive, nil
+}
+
+// detectAndDeleteDeadPods determines if there are registry client pods that are in the deleted state
+// but have not been removed by GC (eg the node goes down before GC can remove them), and attempts to
+// force delete the pods. If there are live registry pods remaining, it returns true, otherwise returns false.
+func detectAndDeleteDeadPods(logger *logrus.Entry, client operatorclient.ClientInterface, pods []*corev1.Pod, sourceNamespace string) (bool, error) {
+	var forceDeletionErrs []error
+	livePodFound := false
+	for _, pod := range pods {
+		if !isPodDead(pod) {
+			livePodFound = true
+			logger.WithFields(logrus.Fields{"pod.namespace": sourceNamespace, "pod.name": pod.GetName()}).Debug("pod is alive")
+			continue
+		}
+		logger.WithFields(logrus.Fields{"pod.namespace": sourceNamespace, "pod.name": pod.GetName()}).Info("force deleting dead pod")
+		if err := client.KubernetesInterface().CoreV1().Pods(sourceNamespace).Delete(context.TODO(), pod.GetName(), metav1.DeleteOptions{
+			GracePeriodSeconds: ptr.To[int64](0),
+		}); err != nil && !apierrors.IsNotFound(err) {
+			forceDeletionErrs = append(forceDeletionErrs, err)
+		}
+	}
+	if len(forceDeletionErrs) > 0 {
+		return false, errors.Join(forceDeletionErrs...)
+	}
+	return livePodFound, nil
 }
--- a/pkg/controller/registry/reconciler/grpc.go
+++ b/pkg/controller/registry/reconciler/grpc.go
@@ -2,9 +2,7 @@ package reconciler
 
 import (
 	"context"
-	"errors"
 	"fmt"
-	"slices"
 	"strings"
 	"time"
 
@@ -24,7 +22,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/intstr"
-	"k8s.io/utils/ptr"
 )
 
 const (
@@ -348,25 +345,6 @@ func isRegistryServiceStatusValid(source *grpcCatalogSourceDecorator) (bool, err
 func (c *GrpcRegistryReconciler) ensurePod(logger *logrus.Entry, source grpcCatalogSourceDecorator, serviceAccount *corev1.ServiceAccount, defaultPodSecurityConfig v1alpha1.SecurityConfig, overwrite bool) error {
 	// currentPods refers to the current pod instances of the catalog source
 	currentPods := c.currentPods(logger, source)
-
-	var forceDeleteErrs []error
-	currentPods = slices.DeleteFunc(currentPods, func(pod *corev1.Pod) bool {
-		if !isPodDead(pod) {
-			logger.WithFields(logrus.Fields{"pod.namespace": source.GetNamespace(), "pod.name": pod.GetName()}).Debug("pod is alive")
-			return false
-		}
-		logger.WithFields(logrus.Fields{"pod.namespace": source.GetNamespace(), "pod.name": pod.GetName()}).Info("force deleting dead pod")
-		if err := c.OpClient.KubernetesInterface().CoreV1().Pods(source.GetNamespace()).Delete(context.TODO(), pod.GetName(), metav1.DeleteOptions{
-			GracePeriodSeconds: ptr.To[int64](0),
-		}); err != nil && !apierrors.IsNotFound(err) {
-			forceDeleteErrs = append(forceDeleteErrs, pkgerrors.Wrapf(err, "error deleting old pod: %s", pod.GetName()))
-		}
-		return true
-	})
-	if len(forceDeleteErrs) > 0 {
-		return errors.Join(forceDeleteErrs...)
-	}
-
 	if len(currentPods) > 0 {
 		if !overwrite {
 			return nil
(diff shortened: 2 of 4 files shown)


perdasilva:
So far it looks pretty good =D should we add any unit tests for the new behaviour? and maybe the test we missed in the original bug fix?

pkg/controller/operators/catalog/operator.go
```diff
@@ -992,6 +992,12 @@ func (o *Operator) syncRegistryServer(logger *logrus.Entry, in *v1alpha1.Catalog
 		out.SetError(v1alpha1.CatalogSourceRegistryServerError, syncError)
 		return
 	}
+	err := srcReconciler.CleanRegistryServer(logger, in)
+	if err != nil {
+		syncError = fmt.Errorf("could not clean dead resources: %s", err)
```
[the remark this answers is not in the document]
anik120: That's fair, changed it to `could not clean stale resources`

perdasilva:
What's the signal

> > So far it looks pretty good =D should we add any unit tests for the new behaviour? and maybe the test we missed in the original bug fix?
> 
> I was scared you'll ask this. :p
> 
> I did think about it, looked into it, and it looks like it's going to be a little bit of effort to write a test for this. Mainly because it's not really unit testable so an e2e test is the viable option. However, I'm not sure we can mimic a node going down from within our tests.
> 
> I just figured the squeeze might not be worth the juice, however if someone has a better idea (or really insist we include a test for this even if it'll require some effort), I'm happy to include one in this PR.

What's the signal we get in the code that a node has gone down? Are some pods unreachable?

anik120:
I may have found a way to e2e test this......working on it now....

perdasilva:
> t discover pod/s that have been `Deleted`

Would it be possible to mock those client responses? e.g. list says they are there, get says they aren't?

anik120:
Okay I have some tests now. 

> I may have found a way to e2e test this......working on it now....

This did not work out. I was thinking about using `gracefulDeletionPeriod` as a hack, but that's not going to be a very reliable test (since `gracefulDeletionPeriod` doesn't necessarily force the pod to hang around for the period specified after it's been deleted, like I was hoping it would. Instead, it looks like as soon as the container in the pod stops, the GC comes and deletes it, which in our case is very quick. ie the registry server is shut down immediately so the pod will be cleaned up immediately too.) 


> Would it be possible to mock those client responses? e.g. list says they are there, get says they aren't?

I went with the unit tests route, that mocks these interactions. So the entire feature is tested by 3 components of tests: 

1. The `CleanRegistryServer` for both `Grpc` and `ConfigMap` is tested individually in the `reconciler` package 
note: it's more clear that the `CleanRegistryServer` is getting called because of the presence of a "deleted" pod when the test is run in verbose mode: 
```
go test ./pkg/controller/registry/reconciler -run TestGrpcRegistryCleaner -v
=== RUN   TestGrpcRegistryCleaner
=== RUN   TestGrpcRegistryCleaner/Grpc/ExistingRegistry/DeletedPod
time="2024-08-19T16:42:20-04:00" level=info msg="force deleting dead pod" pod.name= pod.namespace=testns
--- PASS: TestGrpcRegistryCleaner (0.10s)
    --- PASS: TestGrpcRegistryCleaner/Grpc/ExistingRegistry/DeletedPod (0.10s)
PASS
ok  	github.com/operator-framework/operator-lifecycle-manager/pkg/controller/registry/reconciler	0.939s
```

`force deleting dead pod` is logged. 

2. A level above, in the operator package (that uses the reconciler implementations), a unit test is put in place to check that `reconciler.CleanRegistryServer` is called.

This is the absolute best we can do. e2e isn't possible without much more digging in, and frankly at this point not necessary at all.

anik120:
Pivoted the PR to include the logic inside `CheckRegistryServer`, instead of adding a `CleanRegistryServer` component to the interface.

pkg/controller/registry/reconciler/configmap.go
```diff
@@ -513,5 +515,23 @@ func (c *ConfigMapRegistryReconciler) CheckRegistryServer(logger *logrus.Entry,
...
 
 	healthy = true
+	var forceDeletionErrs []error
+	for _, pod := range pods {
+		if !isPodDead(pod) {
+			logger.WithFields(logrus.Fields{"pod.namespace": source.GetNamespace(), "pod.name": pod.GetName()}).Debug("pod is alive")
+			continue
+		}
+		healthy = false
+		forceDeletionErrs = append(forceDeletionErrs, pkgerrors.Errorf("found %s in a deleted but not removed state", pod.Name))
```
joelanford, line 525: Do we need to return an error from this function when we find a wedged pod and successfully delete it? My initial thought is that we would only error from this function if:
- we failed to determine healthy
- we failed to delete wedged pods.

pkg/controller/registry/reconciler/grpc.go
```diff
@@ -628,15 +608,35 @@ func (c *GrpcRegistryReconciler) CheckRegistryServer(logger *logrus.Entry, catal
...
+		}); err != nil && !apierrors.IsNotFound(err) {
+			forceDeletionErrs = append(forceDeletionErrs, pkgerrors.Wrapf(err, "error deleting old pod: %s", pod.GetName()))
+		}
+	}
+	if len(forceDeletionErrs) > 0 {
+		return false, errors.Join(forceDeletionErrs...)
+	}
+	if foundDeadPod {
+		return false, nil
+	}
```
joelanford, line 639: How much of this can be extracted into a helper function used by both the configmap and grpc implementations?

pkg/controller/registry/reconciler/configmap.go
anik120: I started out thinking the same, but then realized that that'd be an "artificial" error. If we include both of the scenarios you mentioned in the definition of "error", we're sort of sullying the definition which is "something went wrong". In the case of "we failed to determine healthy", we are already sending that signal through the boolean variable. So decided to not include both, and only include the second scenario of "we failed to delete the wedged pods".

pkg/controller/registry/reconciler/grpc.go
anik120: Fair point, let me give it a shot...

anik120: Okay I was able to refractor it out, PTAL!

pkg/controller/registry/reconciler/configmap.go
```diff
@@ -513,5 +515,36 @@ func (c *ConfigMapRegistryReconciler) CheckRegistryServer(logger *logrus.Entry,
...
+		healthy = false
+		err = e
+	}
+	return
+}
+
+// detectAndDeleteDeadPods determines if there are registry client pods that are in the deleted state
+// but have not been removed by GC (eg the node goes down before GC can remove them), and attempts to
+// force delete the pods
+func detectAndDeleteDeadPods(logger *logrus.Entry, client operatorclient.ClientInterface, pods []*corev1.Pod, sourceNamespace string) (foundDeadPod bool, err error) {
```
joelanford, line 528: Nit (but sorta major nit): I know there are other places in the old OLM code that use named/naked returns, but these almost always hurt code readability. Not suggesting we go and fix all of those, but I think it would be good to avoid adding more.

WDYT about using `(bool, error)` in the signature and have explicit returns?

pkg/controller/registry/reconciler/configmap.go
```diff
@@ -513,5 +515,36 @@ func (c *ConfigMapRegistryReconciler) CheckRegistryServer(logger *logrus.Entry,
...
+	var forceDeletionErrs []error
+	foundDeadPod = false
+	for _, pod := range pods {
+		if !isPodDead(pod) {
+			logger.WithFields(logrus.Fields{"pod.namespace": sourceNamespace, "pod.name": pod.GetName()}).Debug("pod is alive")
+			continue
+		}
+		foundDeadPod = true
```
joelanford, line 536: This is trickier than I originally thought. If we have:
1. at least one alive pod
2. the alive pods are otherwise deemed healthy
3. we successfully delete the dead pods

What should we return from `CheckRegistryServer`? Seems like we could say healthy in that case?

pkg/controller/registry/reconciler/grpc.go
```diff
@@ -347,31 +344,12 @@ func isRegistryServiceStatusValid(source *grpcCatalogSourceDecorator) (bool, err
...
-			forceDeleteErrs = append(forceDeleteErrs, pkgerrors.Wrapf(err, "error deleting old pod: %s", pod.GetName()))
-		}
-		return true
-	})
-	if len(forceDeleteErrs) > 0 {
-		return errors.Join(forceDeleteErrs...)
-	}
-
-	if len(currentPods) > 0 {
+	currentLivePods := c.currentPods(logger, source)
```
joelanford, line 347: Nit: Should this still be called simply `currentPods`? Including "live" in the name makes it seem like we've checked for dead-ness and excluded them in the logic of `c.currentPods()`, but that's not what's happening.

I realize it is likely they are all alive at this point because we should have already deleted the dead ones, but that happens in a separate code path and might not always be the case.

pkg/controller/registry/reconciler/grpc_test.go
```diff
@@ -70,6 +71,25 @@ func grpcCatalogSourceWithName(name string) *v1alpha1.CatalogSource {
 	return catsrc
 }
 
+func withPodDeletedButNotRemoved(objs []runtime.Object) []runtime.Object {
+	var out []runtime.Object
+	t := reflect.TypeOf(&corev1.Pod{})
+	for _, obj := range objs {
+		o := obj.DeepCopyObject()
+		if reflect.TypeOf(o) == t {
```
joelanford, line 79: Can the reflection library be replaced by a type assertion (which you already have), but that checks the type assertions bool return value.
```go
for _, obj := range objs {
    if pod, ok := obj.(*corev1.Pod); ok {
        // do stuff with pod
    }
}
```

pkg/controller/registry/reconciler/configmap.go
anik120: Yea this is a bit confusing, I'm thinking if we detect even one dead pod, we just force delete all the pods, and let `EnsureRegistryServer` recreate everything. Otherwise we're in a very non-deterministic state...

pkg/controller/registry/reconciler/configmap.go
anik120: Sure I can do that, I don't like named returns either, just went with it since `CheckRegistryServer` implementation for ConfigMap already has that. But I'll update the PR with just `(bool, error)` in my next update...

pkg/controller/registry/reconciler/grpc.go
anik120: Sure, I was just signaling that this is being reverted back to what it was originally, but I'm happy to keep it as `currentPods`.

pkg/controller/registry/reconciler/grpc_test.go
anik120: Done.

pkg/controller/registry/reconciler/grpc.go
anik120: Done.

pkg/controller/registry/reconciler/configmap.go
anik120: Done.

pkg/controller/registry/reconciler/configmap.go
joelanford: That seems like that could have significant performance implications. There's a lot of disk/cpu/memory costs that are paid when catalog pods are started, so we should minimized that as much as possible.

I think it is okay to force delete the dead pods, and if there are any alive pods, we just move forward as if those were the only pods that were there to begin with.

anik120: That's fair too. I was a bit confused about there being multiple pods in the first place, since my impression is that we only create one registry pod per catalog, but turns out we have `pods` mainly because we use the lister to list via label selector. In any case this should return one pod, unless I'm unaware of some feature that allows us to spin up multiple registry pods per catalog. 

Changed it back to just deleting dead pods.

joelanford: There are definitely multiple pods when we're polling. In that case, we spin up a new pod to see what the digest is, and we compare the digests to see if we need to use the new pod (if it has a new digest) or keep using the old pod (when the digests match)

There may also be multiple pods when the pod spec changes (e.g. when the catalog source image is changed).

anik120: Right, both are transitory states though, and we're in that state because `EnsureRegistryServer` has already been called. But yes I think this is in a good state now.

tmshort:
@joelanford are you good with this?

pkg/controller/registry/reconciler/grpc.go
```diff
@@ -628,15 +606,17 @@ func (c *GrpcRegistryReconciler) CheckRegistryServer(logger *logrus.Entry, catal
...
 	if err != nil {
 		return false, err
 	}
-	if len(current) < 1 ||
+	if len(currentPods) < 1 ||
 		service == nil || c.currentServiceAccount(source) == nil {
 		return false, nil
 	}
-
+	if deadPodsDetected, e := detectAndDeleteDeadPods(logger, c.OpClient, currentPods, source.GetNamespace()); deadPodsDetected {
```
joelanford, line 617: I still think we should return `true, nil` from this function in the case where:
- There is at least one healthy pod
- There are multiple "dead" pods
- We deleted all of the "dead" pods successfully.

That sort of behavior would mean that we would short circuit and avoid calling `EnsureRegistryServer`, which would:
  - end up making some no-op calls to the apiserver:
  	- e.g. https://github.com/anik120/operator-lifecycle-manager/blob/20a171690cf96fcea177e5fa90426e7c234a14d3/pkg/controller/registry/reconciler/grpc.go#L459
  - (maybe, not sure) have some strange issues related to caching, where subsequent calls to `currentPods` would still return the dead pods because the deletion has no propagated back to our informer cache yet (I'm assuming `c.Lister` is backed by a cache. If it is not, then we'd be making more apiserver calls unnecessarily)

joelanford: Essentially, I think both of the following should result in identical behavior:
1. There are 1 or more alive pods, and no dead pods
2. There are 1 ore more alive pods, and we deleted a bunch of dead pods (hence: there are no dead pods, so this is actually a variant of (1))

anik120: @joelanford thanks for the discussion yesterday. Thought about it a little more and we can do this. I've had to add an additional change though, highlighting it in the next comment...

pkg/controller/operators/catalog/operator.go
```diff
@@ -994,7 +994,7 @@ func (o *Operator) syncRegistryServer(logger *logrus.Entry, in *v1alpha1.Catalog
 	}
 
 	healthy, err := srcReconciler.CheckRegistryServer(logger, in)
-	if err != nil {
+	if err != nil && !healthy {
```
anik120, line 997: @joelanford had add in this change to make sure we're sending the reconciler back if there were errors force deleting the dead pods, AND `CheckRegistryServer` returns `healthy=false`. In case we detect dead pods but run into errors deleting them, but we find healthy pod, we are going forward with the current loop.

pkg/controller/registry/reconciler/configmap.go
```diff
@@ -513,5 +515,39 @@ func (c *ConfigMapRegistryReconciler) CheckRegistryServer(logger *logrus.Entry,
 	}
 
 	healthy = true
+	if deadPodsDetected, livePodsCount, e := detectAndDeleteDeadPods(logger, c.OpClient, pods, source.GetNamespace()); deadPodsDetected {
+		if livePodsCount == 0 {
+			healthy = false
+		}
+		err = e
+	}
```
joelanford, line 523: Nit: do we really care if dead pods are detected? Seems like this could be simplified to:
```go
livePodsCount, deleteDeadPodsErr := detectAndDeleteDeadPods(logger, c.OpClient, pods, source.GetNamespace())
if deleteDeadPodsErr != nil {
	return false, deleteDeadPodsErr
}
return livePodsCount > 0, nil
```

joelanford: In fact, we don't even care how many live pods there are. We could have `detectAndDeleteDeadPods` return `(bool, error)`, where the boolean value is `haveAnyAlivePods`.

pkg/controller/operators/catalog/operator.go
anik120: If we return `healthy=true` from `CheckRegistryServer`, but there was a problem deleting some dead pod, isn't the whole point of returning `healthy=true` moot?

pkg/controller/registry/reconciler/configmap.go
anik120: Fair.

pkg/controller/registry/reconciler/configmap.go
```diff
@@ -513,5 +515,36 @@ func (c *ConfigMapRegistryReconciler) CheckRegistryServer(logger *logrus.Entry,
 	}
 
 	healthy = true
+	podsAreLive, e := detectAndDeleteDeadPods(logger, c.OpClient, pods, source.GetNamespace())
+	if !podsAreLive {
+		healthy = false
+		err = e
+	}
```
anik120, line 522: Okay this seems like it's the best we can do. `detectAndDeleteDeadPods` attempts to delete and dead pods, and returns the errors if there were any, but returns true if it finds a healthy pod. 

In `CheckRegistryServer`, if we get `false` from `detectAndDeleteDeadPods`, then we set `healthy` to `false` and return `error` if there were any. 

If there were error deleting dead pods, but we found a healthy pod, we still return `healthy=true`. The next time the loop is run we'll attempt to delete the dead pods again.

pkg/controller/operators/catalog/operator.go
joelanford: If we encounter an error doing stuff, we should bubble that error up regardless. My original point is that we should be able to return `healthy=true, err=nil` when there is a live pod and we successfully delete the dead pods.

Pretend like we don't own the implementation of `CheckRegistryServer`. If we add `!healthy` to this check, then we'll completely ignore errors that occur during `CheckRegistryServer` if it returns `healthy=true`.

If any error is returned we _should_ propagate that error even if there are other healthy pods, because those errors might be indicative of a separate problem that needs a user to intervene.

pkg/controller/registry/reconciler/configmap.go
joelanford: I think we need to propagate errors all the way up the stack regardless of the boolean value returned with it. If any error occurs, we are automatically considering things unhealthy.

Structure all of these like this:
```
boolVal, err := doThing()
if err != nil {
    return false, fmt.Errorf("error doing thing: %v", err)
}
return boolVal, nil
```

PR #3377, opened 2024-08-27
fix: call TokenRequest API when service account token secret is missing

### LINKED ISSUE #3376: Handle missing Service Account Token Secret
## Bug Report

**What did you do?**

An attempt was made to install an operator.

The operator would install successfully.

Installation of the operator failed and an error was returned 

```Bash
time="2024-07-22T06:03:01Z" level=error msg="failed to get a client for operator deployment - the service account does not have any API secret sa=npv-dev-34/some-ansible-operator"
```

**Environment**

```Bash
$ oc get csv packageserver -n openshift-operator-lifecycle-manager
NAME            DISPLAY          VERSION          REPLACES   PHASE
packageserver   Package Server   0.0.1-snapshot              Succeeded
```

OpenShift v4.16

```Bash
$ oc version
Client Version: 4.16.0-202406052127.p0.ga245041.assembly.stream-a245041
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: 4.16.0
Kubernetes Version: v1.29.5+29c95f3
$
```

* Kubernetes cluster kind: 

**Possible Solution**

When a user attempts to install an operator, the Operator Lifecycle Manager should be able to create a service account token by making a request against the Token Request API if it can not find the service account token secret associated with the target service account.

**Additional context**
Versions of Kubernetes prior to v1.22 automatically created credentials for accessing the Kubernetes API which would be stored in a secret with a similar name as the service account. The secrets would be mounted onto pods and used to access the Kubernetes API.

However, since Kubernetes v1.24, credentials are acquired directly from the Token Request API and mounted onto pods via a projected volume.

OchiengEd:

This pull requests introduces code that will attempt to request a service account token via the TokenRequest API whenever an error is return regarding a missing service account token secret.

Beyond Kubernetes 1.22, the [service account token secret](https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/#manual-secret-management-for-serviceaccounts) is not automatically created. Therefore, unless manually created, the service account token secret is expect to be missing.

As a result, it is necessary to update the Operator Lifecycle manager(OLM) code to account for the above change in the Kubernetes behavior.

**Architectural changes:**

**Testing remarks:**

**Reviewer Checklist**

Closes #3376

### DIFF
--- a/pkg/lib/scoped/token_retriever.go
+++ b/pkg/lib/scoped/token_retriever.go
@@ -37,12 +37,15 @@ func (r *BearerTokenRetriever) Retrieve(reference *corev1.ObjectReference) (toke
 	}
 
 	if secret == nil {
-		err = fmt.Errorf("the service account does not have any API secret sa=%s/%s", sa.GetNamespace(), sa.GetName())
-		token, _ = requestSAToken(r.kubeclient, sa)
-		if token != "" {
-			err = nil
+		token, err = requestSAToken(r.kubeclient, sa)
+		if err != nil {
+			err = fmt.Errorf("creating service account token from TokenRequest API for sa=%s/%s; %v",
+				sa.GetNamespace(),
+				sa.GetName(),
+				err,
+			)
+			return
 		}
-		return
 	}
 
 	token = string(secret.Data[corev1.ServiceAccountTokenKey])
--- a/pkg/lib/scoped/token_retriever.go
+++ b/pkg/lib/scoped/token_retriever.go
@@ -44,8 +44,8 @@ func (r *BearerTokenRetriever) Retrieve(reference *corev1.ObjectReference) (toke
 				sa.GetName(),
 				err,
 			)
-			return
 		}
+		return
 	}
 
 	token = string(secret.Data[corev1.ServiceAccountTokenKey])
--- a/pkg/lib/scoped/token_retriever.go
+++ b/pkg/lib/scoped/token_retriever.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/operatorclient"
 	"github.com/sirupsen/logrus"
+	authv1 "k8s.io/api/authentication/v1"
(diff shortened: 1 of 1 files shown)


pkg/lib/scoped/token_retriever.go
```diff
@@ -37,6 +38,10 @@ func (r *BearerTokenRetriever) Retrieve(reference *corev1.ObjectReference) (toke
 
 	if secret == nil {
 		err = fmt.Errorf("the service account does not have any API secret sa=%s/%s", sa.GetNamespace(), sa.GetName())
+		token, _ = requestSAToken(r.kubeclient, sa)
+		if token != "" {
+			err = nil
```
perdasilva, line 43: could it be worthwhile to add a log line with the underlying error here? I'm thinking it might be good to know from a debugging perspective to understand that there was an attempt to acquire the token but it failed because of ...

perdasilva:
P.S. if you rebase, the flake will go away =D

OchiengEd:
@perdasilva All done. You should be good to review

pkg/lib/scoped/token_retriever.go
```diff
@@ -36,8 +37,15 @@ func (r *BearerTokenRetriever) Retrieve(reference *corev1.ObjectReference) (toke
...
-		return
+		token, err = requestSAToken(r.kubeclient, sa)
+		if err != nil {
+			err = fmt.Errorf("creating service account token from TokenRequest API for sa=%s/%s; %v",
+				sa.GetNamespace(),
+				sa.GetName(),
+				err,
+			)
+			return
+		}
```
tmshort, line 48: Doesn't the token get immediately overridden by the line below? Should the `return` be on line 48?

OchiengEd: Good catch. I moved the `return` down a line which will ensure the token and error are returned before evaluating subsequent lines.

perdasilva (approved):
/lgtm

PR #3414, opened 2024-10-18
Add more logging around catalog source sync

perdasilva:

A recent live debugging session revealed that the catalog source reconciliation loop is under logged. This PR adds more logging. Generally, I've tried to:
 - log start and end of method calls
 - log errors as close as possible to source
 - call out steps in the methods

**Architectural changes:**

**Testing remarks:**

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -912,6 +912,7 @@ func validateSourceType(logger *logrus.Entry, in *v1alpha1.CatalogSource) (out *
 		err = fmt.Errorf("unknown sourcetype: %s", sourceType)
 	}
 	if err != nil {
+		logger.WithError(err).Error("error validating catalog source type")
 		out.SetError(v1alpha1.CatalogSourceSpecInvalidError, err)
 		return
 	}
@@ -923,7 +924,6 @@ func validateSourceType(logger *logrus.Entry, in *v1alpha1.CatalogSource) (out *
 		}
 	}
 	continueSync = true
-
 	return
 }
 
@@ -936,27 +936,22 @@ func (o *Operator) syncConfigMap(logger *logrus.Entry, in *v1alpha1.CatalogSourc
 
 	out = in.DeepCopy()
 
-	logger = logger.WithFields(logrus.Fields{
-		"configmap.namespace": in.Namespace,
-		"configmap.name":      in.Spec.ConfigMap,
-	})
-	logger.Info("checking catsrc configmap state")
-
 	var updateLabel bool
 	// Get the catalog source's config map
 	configMap, err := o.lister.CoreV1().ConfigMapLister().ConfigMaps(in.GetNamespace()).Get(in.Spec.ConfigMap)
 	// Attempt to look up the CM via api call if there is a cache miss
 	if apierrors.IsNotFound(err) {
+		// TODO: Don't reach out via live client if its not found in the cache (https://github.com/operator-framework/operator-lifecycle-manager/issues/3415)
 		configMap, err = o.opClient.KubernetesInterface().CoreV1().ConfigMaps(in.GetNamespace()).Get(context.TODO(), in.Spec.ConfigMap, metav1.GetOptions{})
 		// Found cm in the cluster, add managed label to configmap
 		if err == nil {
-			labels := configMap.GetLabels()
-			if labels == nil {
-				labels = make(map[string]string)
+			cmLabels := configMap.GetLabels()
+			if cmLabels == nil {
+				cmLabels = make(map[string]string)
 			}
 
-			labels[install.OLMManagedLabelKey] = "false"
-			configMap.SetLabels(labels)
+			cmLabels[install.OLMManagedLabelKey] = "false"
+			configMap.SetLabels(cmLabels)
 			updateLabel = true
 		}
 	}
@@ -973,12 +968,9 @@ func (o *Operator) syncConfigMap(logger *logrus.Entry, in *v1alpha1.CatalogSourc
 			out.SetError(v1alpha1.CatalogSourceConfigMapError, syncError)
 			return
 		}
-
-		logger.Info("adopted configmap")
 	}
 
 	if in.Status.ConfigMapResource == nil || !in.Status.ConfigMapResource.IsAMatch(&configMap.ObjectMeta) {
-		logger.Info("updating catsrc configmap state")
 		// configmap ref nonexistent or updated, write out the new configmap ref to status and exit
 		out.Status.ConfigMapResource = &v1alpha1.ConfigMapResourceReference{
 			Name:            configMap.GetName(),
@@ -998,7 +990,6 @@ func (o *Operator) syncConfigMap(logger *logrus.Entry, in *v1alpha1.CatalogSourc
 func (o *Operator) syncRegistryServer(logger *logrus.Entry, in *v1alpha1.CatalogSource) (out *v1alpha1.CatalogSource, continueSync bool, syncError error) {
 	out = in.DeepCopy()
 
-	logger.Info("synchronizing registry server")
 	sourceKey := registry.CatalogKey{Name: in.GetName(), Namespace: in.GetNamespace()}
 	srcReconciler := o.reconciler.ReconcilerForSource(in)
 	if srcReconciler == nil {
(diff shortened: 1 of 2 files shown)


pkg/controller/operators/catalog/operator.go
```diff
@@ -912,41 +917,44 @@ func validateSourceType(logger *logrus.Entry, in *v1alpha1.CatalogSource) (out *
...
+		"configmap.name":      in.Spec.ConfigMap,
+	})
+	logger.Info("checking catsrc configmap state")
+
 	out = in
 	if !(in.Spec.SourceType == v1alpha1.SourceTypeInternal || in.Spec.SourceType == v1alpha1.SourceTypeConfigmap) {
 		continueSync = true
+		logger.Info("catalog type is neither internal not configmap - bailing without error")
 		return
 	}
```
joelanford, line 948: I think I'd do the following. Otherwise, it sounds like we'll:
1. Get "checking catsrc configmap state" on every CatalogSource, even if it isn't a configmap type
2. A "catalog type is neither internal..." on all GRPC-based CatalogSources (which are basically all of them)

```suggestion

	out = in
	if !(in.Spec.SourceType == v1alpha1.SourceTypeInternal || in.Spec.SourceType == v1alpha1.SourceTypeConfigmap) {
		continueSync = true
		return
	}
	logger.Info("checking catsrc configmap state")
```

pkg/controller/operators/catalog/operator.go
```diff
@@ -992,30 +1001,39 @@ func (o *Operator) syncConfigMap(logger *logrus.Entry, in *v1alpha1.CatalogSourc
...
 }
 
 func (o *Operator) syncRegistryServer(logger *logrus.Entry, in *v1alpha1.CatalogSource) (out *v1alpha1.CatalogSource, continueSync bool, syncError error) {
 	out = in.DeepCopy()
 
-	logger.Info("synchronizing registry server")
+	logger = logger.WithField("step", "syncRegistryServer")
+	logger.Info("registry server sync start")
+	defer func() {
+		logger.WithError(syncError).WithField("continueSync", continueSync).Info("registry server sync end")
```
joelanford, line 1014: What happens with `WithError(syncError)` if `syncError` is nil?

pkg/controller/operators/catalog/operator.go
```diff
@@ -897,6 +897,11 @@ func (o *Operator) handleCatSrcDeletion(obj interface{}) {
 }
 
 func validateSourceType(logger *logrus.Entry, in *v1alpha1.CatalogSource) (out *v1alpha1.CatalogSource, continueSync bool, _ error) {
+	logger = logger.WithField("step", "validateSourceType")
+	logger.Info("validate source type start")
```
stevekuznetsov, line 901: I think in practice you'll find method book-ending is extremely chatty - on the order of MiB/s logging rate during high churn, especially when the system is just verifying correctness of current state and not changing. Not sure if you really require it. The logs for error cases and outcomes are very useful and will not emit if "nothing" is going on.

pkg/controller/operators/catalog/operator.go
stevekuznetsov: Your suggestion is identical to the previous state of the code, by the looks of it.

pkg/controller/operators/catalog/operator.go
```diff
@@ -912,41 +917,44 @@ func validateSourceType(logger *logrus.Entry, in *v1alpha1.CatalogSource) (out *
...
-		"configmap.name":      in.Spec.ConfigMap,
-	})
-	logger.Info("checking catsrc configmap state")
-
 	var updateLabel bool
 	// Get the catalog source's config map
 	configMap, err := o.lister.CoreV1().ConfigMapLister().ConfigMaps(in.GetNamespace()).Get(in.Spec.ConfigMap)
 	// Attempt to look up the CM via api call if there is a cache miss
 	if apierrors.IsNotFound(err) {
+		logger.Info("configmap not found in cache - attempting with live client")
```
stevekuznetsov, line 957: n.b. this is not a safe thing to do, as you no longer operate against a consistent view of cluster state that existed in one point in time (your lister is behind - but you got an event for this configmap? how?)

pkg/controller/operators/catalog/operator.go
perdasilva: come out as <nil>

pkg/controller/operators/catalog/operator.go
perdasilva: yeah - I tend to agree. The verbosity is high for fairly low utility. I'll cull most of the non-error logs. At least in these paths. Wondering if just leaving pre-return log would still be a good idea to get signal that it went into the method - even if only for the methods higher up the stack....?

pkg/controller/operators/catalog/operator.go
perdasilva: I see what you mean. Thanks for calling that out. I'll poke at it to see if we can just re-enqueue and let the next iteration pick it up from the cache, rather than try to jump ahead of it. Though the code here can get so convoluted its hard to know what taking this jenga piece out will do to the tower XD

pkg/controller/operators/catalog/operator.go
perdasilva: Looking at it now, I'm wondering if we should use WithField("syncError") just to have more fidelity with the code and helping the next person that has to do a live manual debug XD

pkg/controller/operators/catalog/operator.go
perdasilva: btw, how's Bill treating you?

pkg/controller/operators/catalog/operator.go
perdasilva: okie - see if that sparks joy, or if you reckon it's still too verbose

stevekuznetsov: Try it out and see! It might still be too chatty to be useful, but you can't know without running an end-to-end test.

pkg/controller/operators/catalog/operator.go
stevekuznetsov: It's a wild and whacky time! Best of luck with this Jenga :)

pkg/controller/registry/reconciler/grpc.go
```diff
@@ -213,20 +211,10 @@ func (c *GrpcRegistryReconciler) currentPodsWithCorrectImageAndSpec(logger *logr
...
 	}
 	for _, p := range pods {
-		images, hash := correctImages(source, p), podHashMatch(p, newPod)
-		logger = logger.WithFields(logrus.Fields{
-			"current-pod.namespace": p.Namespace, "current-pod.name": p.Name,
-			"correctImages": images, "correctHash": hash,
-		})
-		logger.Info("evaluating current pod")
-		if !hash {
-			logger.Infof("pod spec diff: %s", cmp.Diff(p.Spec, newPod.Spec))
```
joelanford, line 223: This log seems like it could be useful in understanding why catalog pods are being recycled unexpectedly.

pkg/controller/registry/reconciler/grpc.go
```diff
@@ -213,20 +211,10 @@ func (c *GrpcRegistryReconciler) currentPodsWithCorrectImageAndSpec(logger *logr
...
-		}
 		if correctImages(source, p) && podHashMatch(p, newPod) {
 			found = append(found, p)
 		}
 	}
-	logger.Infof("of %d pods matching label selector, %d have the correct images and matching hash", len(pods), len(found))
```
joelanford, line 229: This also seems like it could still be useful?

pkg/controller/registry/reconciler/grpc.go
```diff
@@ -252,64 +240,71 @@ func (c *GrpcRegistryReconciler) EnsureRegistryServer(logger *logrus.Entry, cata
...
 	// recreate the pod if no existing pod is serving the latest image or correct spec
 	current, err := c.currentPodsWithCorrectImageAndSpec(logger, source, sa, defaultPodSecurityConfig)
 	if err != nil {
+		logger.WithError(err).Error("error ensuring registry server: could not get current pods with correct image and spec")
 		return err
 	}
 	overwritePod := overwrite || len(current) == 0
-	if overwritePod {
-		logger.Info("registry pods invalid, need to overwrite")
-	}
```
joelanford, line 286: Is there some other logging of the fact that we are going to override the pod (or at least some logging that only happens if we overwrite the pod), even in the happy path case?

joelanford (approved):

PR #3416, opened 2024-10-18
 fix: buggy `nextY` logic for maxOCPVersion

anik120:
The handling logic of versions that are pre-releases by the
`nextY()` func (that determines the next Y release) was erroneous.
    
Eg: `nextY("4.16.0")` returns "4.17" correctly, but `nextY("4.16.0-rc1")`
returns "4.16" (the correct value is still "4.17").
    
This PR fixes the `nextY` function.
    
Also has improvement for the "not-upgradeable to next OCP" version message.

**Architectural changes:**

**Testing remarks:**

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/openshift/helpers.go
+++ b/pkg/controller/operators/openshift/helpers.go
@@ -91,7 +91,12 @@ func (s skews) String() string {
 		j--
 	}
 
-	return "ClusterServiceVersions blocking cluster upgrade: " + strings.Join(msg, ",")
+	// it is safe to ignore the error here, with the assumption
+	// that we build each skew object only after verifying that the
+	// version string is parseable safely.
+	maxOCPVersion, _ := semver.ParseTolerant(s[0].maxOpenShiftVersion)
+	nextY := nextY(maxOCPVersion).String()
+	return fmt.Sprintf("ClusterServiceVersions blocking minor version upgrades to %s or higher:\n%s", nextY, strings.Join(msg, "\n"))
 }
 
 type skew struct {
@@ -103,10 +108,9 @@ type skew struct {
 
 func (s skew) String() string {
 	if s.err != nil {
-		return fmt.Sprintf("%s/%s has invalid %s properties: %s", s.namespace, s.name, MaxOpenShiftVersionProperty, s.err)
+		return fmt.Sprintf("- %s/%s has invalid %s properties: %s", s.namespace, s.name, MaxOpenShiftVersionProperty, s.err)
 	}
-
-	return fmt.Sprintf("%s/%s is incompatible with OpenShift minor versions greater than %s", s.namespace, s.name, s.maxOpenShiftVersion)
+	return fmt.Sprintf("- maximum supported OCP version for %s/%s is %s", s.namespace, s.name, s.maxOpenShiftVersion)
 }
 
 type transientError struct {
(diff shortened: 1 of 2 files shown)


pkg/controller/operators/openshift/helpers.go
```diff
@@ -105,8 +105,12 @@ func (s skew) String() string {
...
 		return fmt.Sprintf("%s/%s has invalid %s properties: %s", s.namespace, s.name, MaxOpenShiftVersionProperty, s.err)
 	}
-
-	return fmt.Sprintf("%s/%s is incompatible with OpenShift minor versions greater than %s", s.namespace, s.name, s.maxOpenShiftVersion)
+	// is it safe to ignore the error here, with the assumption
+	// that we build a skew object only after verifying that the
+	// version string is parseable safely.
+	maxOCPVersion, _ := semver.ParseTolerant(s.maxOpenShiftVersion)
+	nextY := nextY(maxOCPVersion).String()
+	return fmt.Sprintf("ClusterServiceVersions blocking upgrade to %s or higher. The maximum supported OCP version for %s/%s is %s", nextY, s.namespace, s.name, s.maxOpenShiftVersion)
```
anik120, line 113: For ref, [this](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/controller/operators/openshift/helpers.go#L144-L168) is where the skew objects are created, where [s.maxOpenshiftVersion](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/controller/operators/openshift/helpers.go#L165) is set AFTER the value of the field from the CSV is parsed [here](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/controller/operators/openshift/helpers.go#L154), and `maxOpenshiftVersion()` [checks that the value is parseable and returns the error if it's not](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/controller/operators/openshift/helpers.go#L281-L284)

grokspawn (approved):

pkg/controller/operators/openshift/helpers.go
joelanford, line 113: Am I reading things correctly that this message is printed for each CSV that is blocking the upgrade. If so, it would be duplicative to repeatedly state in our Upgradeable=False message that "ClusterServiceVersions are blocking minor version upgrades to X.Y+1 or higher"

WDYT about the final message looking something like this?
```yaml
conditions:
- type: Upgradeable
  status: False
  reason: IncompatibleOperators
  message: |
    ClusterServiceVersions are blocking minor version upgrades to X.Y+1 or higher.
      - openshift-operators/foo is supported through X.Y
      - openshift-operators/bar is supported through X.Y
      - openshift-operators/baz is supported through X.Y
```

anik120: Ah good catch. Updated the PR: got the exact format you mentioned, [see this](https://go.dev/play/p/QZvCpbQzlbM)

pkg/controller/operators/openshift/helpers.go
```diff
@@ -105,8 +110,7 @@ func (s skew) String() string {
 	if s.err != nil {
 		return fmt.Sprintf("%s/%s has invalid %s properties: %s", s.namespace, s.name, MaxOpenShiftVersionProperty, s.err)
```
joelanford, line 111:
```suggestion
		return fmt.Sprintf("- %s/%s has invalid %s properties: %s", s.namespace, s.name, MaxOpenShiftVersionProperty, s.err)
```

joelanford (approved):

Issue #3419, opened 2024-10-25
0.29.0 never becomes ready


dgn:
## Bug Report

**What did you do?**
We regularly install the olm-operator on KinD using `operator-sdk olm install`. This has started failing since 0.29.0 has come out. 0.28.0 installs just fine. We're using the latest version of `operator-sdk` (v1.37.0)

Here's a log:
```
INFO[0010] Creating OLM resources...                    
INFO[0010]   Creating Namespace "olm"                   
INFO[0012]   Namespace "olm" created                    
INFO[0012]   Creating Namespace "operators"             
INFO[0013]   Namespace "operators" created              
[...snip...]
INFO[0024]   Creating CatalogSource "olm/operatorhubio-catalog" 
INFO[0025]   CatalogSource "olm/operatorhubio-catalog" created 
INFO[0025] Waiting for deployment/olm-operator rollout to complete 
INFO[0026]   Waiting for Deployment "olm/olm-operator" to rollout: 0 of 1 updated replicas are available 
FATA[0120] Failed to install OLM version "latest": deployment/olm-operator failed to rollout: context deadline exceeded 
```

The containers fail to start with `CreateContainerConfigError`. Here's the relevant event log:

```
  Type     Reason     Age                 From               Message
  ----     ------     ----                ----               -------
  Warning  Failed     5s (x11 over 2m1s)  kubelet            Error: container has runAsNonRoot and image will run as root (pod: "olm-operator-7b6ffc567-nnlm7_olm(5b55cb5e-8d14-4b71-abd7-8fd8644c28be)", container: olm-operator)
  Normal   Pulled     5s (x10 over 2m)    kubelet            Container image "quay.io/operator-framework/olm@sha256:cfdda06700fae2962ac89e677c1968517f8c5fae838ca31774b909f545d7be37" already present on machine
```

**What did you expect to see?**
A working installation of olm-operator using `operator-sdk olm install`

**What did you see instead? Under which circumstances?**
Deployment never becomes ready.

**Environment**
* operator-lifecycle-manager version:

0.29.0

* Kubernetes version information:

v1.29.2

* Kubernetes cluster kind: 

KinD cluster

**Possible Solution**
Looks like `container has runAsNonRoot and image will run as root` from the event log is the issue, but there might be more.

**Additional context**

jmazzitelli:
FWIW I see the same problem on minikube:

```
$ minikube version
minikube version: v1.34.0
[...]

PR #3434, opened 2024-11-05
Fix canonical image ref resolution

m1kola:

Previously in release scripts we were using `docker inspect` which seems to sort .RepoDigests and it makes selection of the correct digest difficult because platform-specific digest can appear at a different index depending on the digests in the list.

`olm.yaml` is currently being generated with platform-specific image ref. This PR fixes resolution of canonical image refs.

Related to #3419

**Architectural changes:**

N/A

**Testing remarks:**

Create a new tag to trigger release automation.

**Reviewer Checklist**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -306,14 +306,8 @@ verify: vendor verify-codegen verify-mockgen verify-manifests #HELP Run all veri
 
 #SECTION Release
 
-.PHONY: pull-opm
-pull-opm:
-	docker pull $(OPERATOR_REGISTRY_IMAGE)
-
 .PHONY: package
 package: $(YQ) $(HELM) #HELP Package OLM for release
-package: OLM_RELEASE_IMG_REF=$(shell docker inspect --format='{{index .RepoDigests 0}}' $(IMAGE_REPO):$(RELEASE_VERSION))
-package: OPM_IMAGE_REF=$(shell docker inspect --format='{{index .RepoDigests 0}}' $(OPERATOR_REGISTRY_IMAGE))
 package:
 ifndef TARGET
 	$(error TARGET is undefined)
@@ -321,12 +315,6 @@ endif
 ifndef RELEASE_VERSION
 	$(error RELEASE_VERSION is undefined)
 endif
-	@echo "Getting operator registry image"
-	docker pull $(OPERATOR_REGISTRY_IMAGE)
-	$(YQ) w -i deploy/$(TARGET)/values.yaml olm.image.ref $(OLM_RELEASE_IMG_REF)
-	$(YQ) w -i deploy/$(TARGET)/values.yaml catalog.image.ref $(OLM_RELEASE_IMG_REF)
-	$(YQ) w -i deploy/$(TARGET)/values.yaml package.image.ref $(OLM_RELEASE_IMG_REF)
-	$(YQ) w -i deploy/$(TARGET)/values.yaml -- catalog.opmImageArgs "--opmImage=$(OPM_IMAGE_REF)"
 	./scripts/package_release.sh $(RELEASE_VERSION) deploy/$(TARGET)/manifests/$(RELEASE_VERSION) deploy/$(TARGET)/values.yaml
 	ln -sfFn ./$(RELEASE_VERSION) deploy/$(TARGET)/manifests/latest
 ifeq ($(PACKAGE_QUICKSTART), true)
@@ -334,9 +322,8 @@ ifeq ($(PACKAGE_QUICKSTART), true)
 endif
 
 .PHONY: release
-release: pull-opm manifests # pull the opm image to get the digest
+release: manifests
 	@echo "Generating the $(RELEASE_VERSION) release"
-	docker pull $(IMAGE_REPO):$(RELEASE_VERSION)
 	$(MAKE) TARGET=upstream RELEASE_VERSION=$(RELEASE_VERSION) PACKAGE_QUICKSTART=true package
 
 .PHONY: FORCE
--- a/util/image-canonical-ref/main.go
+++ b/util/image-canonical-ref/main.go
@@ -1,0 +1,78 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	"github.com/containers/image/v5/docker"
+	"github.com/containers/image/v5/docker/reference"
+	"github.com/containers/image/v5/manifest"
+	"github.com/containers/image/v5/types"
+)
+
+// This is a simple tool to resolve canonical reference of the image.
+// E.g. this resolves quay.io/operator-framework/olm:v0.28.0 to
+// quay.io/operator-framework/olm@sha256:40d0363f4aa684319cd721c2fcf3321785380fdc74de8ef821317cd25a10782a
+func main() {
+	ctx := context.Background()
+
+	if len(os.Args) != 2 {
+		fmt.Fprintf(os.Stderr, "usage: %s <image reference>\n", os.Args[0])
+		os.Exit(1)
+	}
+
+	ref := os.Args[1]
+
+	if err := run(ctx, ref); err != nil {
+		fmt.Fprintf(os.Stderr, "error running the tool: %s\n", err)
+		os.Exit(1)
+	}
+}
+
+func run(ctx context.Context, ref string) error {
+	imgRef, err := reference.ParseNamed(ref)
+	if err != nil {
+		return fmt.Errorf("error parsing image reference %q: %w", ref, err)
+	}
+
+	sysCtx := &types.SystemContext{}
+	canonicalRef, err := resolveCanonicalRef(ctx, imgRef, sysCtx)
+	if err != nil {
+		return fmt.Errorf("error resolving canonical reference: %w", err)
+	}
+
+	fmt.Println(canonicalRef.String())
+	return nil
+}
+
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -306,23 +306,19 @@ verify: vendor verify-codegen verify-mockgen verify-manifests #HELP Run all veri
...
+package: OPM_IMAGE_REF=$(shell go run util/image-canonical-ref/main.go $(OPERATOR_REGISTRY_IMAGE))
 package:
 ifndef TARGET
 	$(error TARGET is undefined)
 endif
 ifndef RELEASE_VERSION
 	$(error RELEASE_VERSION is undefined)
 endif
-	@echo "Getting operator registry image"
-	docker pull $(OPERATOR_REGISTRY_IMAGE)
```
m1kola, line 325: We no longer need to run `docker pull` in order to resolve the canonical digest so I removed this `docker pull`, one below and `pull-opm` target (no used anywhere in the scripts and I doubt that it is being used by humans).

acornett21:
I'm not sure this PR is needed, also I don't really like the idea of maintaining custom code for this. I think the issue lies with using `podman` locally and `docker` in the CI. So the results you get locally do not match the CI, since this tools hand back different information for the query in the CI. ie `podman` hands back 2 digests (the system arc digest and manifest-list digest), where `docker` only hands back the manifest-list.

acornett21:
I verified that this PR isn't needed, and that `podman` and `docker` function differently, so what developers see locally (when using podman) will be different then what they see in `docker` (locally or in the pipeline)

sample gha
```
name: docker-sha-test

on:
  push:
    branches:
      - main

jobs:
  test-docker:
    name: test docker sha
    runs-on: ubuntu-latest
    steps:
      - name: Checkout
        uses: actions/checkout@v4
      
      - name: Set up QEMU
        uses: docker/setup-qemu-action@v3

      - name: Set up Docker Buildx
        uses: docker/setup-buildx-action@v3

      - name: Inspect Sha
        run: |
          docker pull quay.io/operator-framework/olm:v0.30.0
          docker inspect --format='{{index .RepoDigests 0}}' quay.io/operator-framework/olm:v0.30.0
```

the output
```
08cf91027712: Pull complete
Digest: sha256:79b31e9a42f5c1a4f4eb85521aa4675fb82869ce8009a4476f4f2c5b7ae945a8
Status: Downloaded newer image for quay.io/operator-framework/olm:v0.30.0
quay.io/operator-framework/olm:v0.30.0
quay.io/operator-framework/olm@sha256:79b31e9a42f5c1a4f4eb85521aa4675fb82869ce8009a4476f4f2c5b7ae945a8
```

m1kola:
@acornett21 I don't think it is podman vs docker. You can see the sha for in release job for `v0.30.0` tag [here](https://github.com/operator-framework/operator-lifecycle-manager/actions/runs/11663682945/job/32472621010#step:8:217): it prints `36b43038867f685a32ef862e619da04e27b6b0dfa21e3201655d17aa45b26db5` which is `v0.30.0-amd64`. And we use docker in GH action.

In the logs:

```
docker pull quay.io/operator-framework/olm:v0.30.0
v0.30.0: Pulling from operator-framework/olm
Digest: sha256:79b31e9a42f5c1a4f4eb85521aa4675fb82869ce8009a4476f4f2c5b7ae945a8
```

And then it goes:
```
/home/runner/go/bin/yq-v3.0.0-20201202084205-8846255d1c37 w -i deploy/upstream/values.yaml olm.image.ref quay.io/operator-framework/olm@sha256:36b43038867f685a32ef862e619da04e27b6b0dfa21e3201655d17aa45b26db5
/home/runner/go/bin/yq-v3.0.0-20201202084205-8846255d1c37 w -i deploy/upstream/values.yaml catalog.image.ref quay.io/operator-framework/olm@sha256:36b43038867f685a32ef862e619da04e27b6b0dfa21e3201655d17aa45b26db5
/home/runner/go/bin/yq-v3.0.0-20201202084205-8846255d1c37 w -i deploy/upstream/values.yaml package.image.ref quay.io/operator-framework/olm@sha256:36b43038867f685a32ef862e619da04e27b6b0dfa21e3201655d17aa45b26db5
/home/runner/go/bin/yq-v3.0.0-20201202084205-8846255d1c37 w -i deploy/upstream/values.yaml -- catalog.opmImageArgs "--opmImage=quay.io/operator-framework/opm@sha256:077e654da39d055ea56edc91faa2c36d5e8f4dec58db2017fed24a80246b6ca7
```

The last bit is where we replace image ref to a canonical one with yq. Resulting `olm.yaml` had this sha.

It seems like `docker inspect` results are non-deterministic.

I also tried other tools: I tried `skapeo` : it requires platform override on mac. There is also docker `buildx imagetools` which seems to be doing something similar, but it is docker-specific (e.g. not compatible with podman).

I think this is a small and simple enough to not cause a lot of maintenance burden. And as we learnt with docker inspect - using pre-existing tools doesn't save from head ache :slightly_smiling_face:

Makefile
```diff
@@ -334,9 +330,8 @@ ifeq ($(PACKAGE_QUICKSTART), true)
 endif
 
 .PHONY: release
-release: pull-opm manifests # pull the opm image to get the digest
+release: manifests # pull the opm image to get the digest
```
perdasilva, line 333:
```suggestion
release: manifests

File: util/image-canonical-ref/main.go [hunk 3]
```diff
@@ -0,0 +1,78 @@
...
+)
+
+// This is a simple tool to resolve canonical reference of the image.
+// E.g. this resolves quay.io/operator-framework/olm:v0.28.0 to
+// quay.io/operator-framework/olm@sha256:40d0363f4aa684319cd721c2fcf3321785380fdc74de8ef821317cd25a10782a
+func main() {
+	ctx := context.Background()
+
+	if len(os.Args) != 2 {
+		fmt.Fprint(os.Stderr, "missing argument: provide image reference\n")
```
perdasilva, line 21: nit: if we have > 2 arguments, the message could be misleading. Maybe something like:
```suggestion
		fmt.Fprint(os.Stderr, fmt.Sprintf("usage: %s <image reference>\n", args[0])")
```
?

Makefile
```diff
@@ -306,23 +306,19 @@ verify: vendor verify-codegen verify-mockgen verify-manifests #HELP Run all veri
...
-.PHONY: pull-opm
-pull-opm:
-	docker pull $(OPERATOR_REGISTRY_IMAGE)
-
 .PHONY: package
 package: $(YQ) $(HELM) #HELP Package OLM for release
-package: OLM_RELEASE_IMG_REF=$(shell docker inspect --format='{{index .RepoDigests 0}}' $(IMAGE_REPO):$(RELEASE_VERSION))
-package: OPM_IMAGE_REF=$(shell docker inspect --format='{{index .RepoDigests 0}}' $(OPERATOR_REGISTRY_IMAGE))
+package: OLM_RELEASE_IMG_REF=$(shell go run util/image-canonical-ref/main.go $(IMAGE_REPO):$(RELEASE_VERSION))
+package: OPM_IMAGE_REF=$(shell go run util/image-canonical-ref/main.go $(OPERATOR_REGISTRY_IMAGE))
```
perdasilva, line 312: dumb question: would we bail on failure here?

m1kola: Good question! It used to fail on `docker pull`, but now that it is gone - we should probably add extra checks. Or return `docker pull`s. I'l look into it.

m1kola: I reverted most of the changes and switched to `docker images` instead of `docker inspect`. So this is no longer a concern: we will fail on `docker pull` should something o wrong.

Makefile
```diff
@@ -312,8 +312,8 @@ pull-opm:
 
 .PHONY: package
 package: $(YQ) $(HELM) #HELP Package OLM for release
-package: OLM_RELEASE_IMG_REF=$(shell docker inspect --format='{{index .RepoDigests 0}}' $(IMAGE_REPO):$(RELEASE_VERSION))
-package: OPM_IMAGE_REF=$(shell docker inspect --format='{{index .RepoDigests 0}}' $(OPERATOR_REGISTRY_IMAGE))
+package: OLM_RELEASE_IMG_REF=$(shell docker images --format='{{.Digest}}' $(IMAGE_REPO):$(RELEASE_VERSION))
+package: OPM_IMAGE_REF=$(shell docker images --format='{{.Digest}}' $(OPERATOR_REGISTRY_IMAGE))
```
m1kola, line 316: Argh! Needs more formatting

/hold

m1kola:
I switched from `docker inspect` to getting the digest `docker images` which should hopefully give consistent results between `podman` and `docker`.

@acornett21 @perdasilva please take another look.

m1kola:
We found out that there is difference between `docker` and `podman` in `images` result so it is not compatible.

/hold

m1kola:
Moving back to the original PR with a custom go utility to resolve a canonical reference. This should be portable and we are re-using an existing dependency.

I however moved utility and `YQ` calls into `scripts/package_release.sh` for a few reasons:
1. So we exit on an error without a lot of complicated checks in `Makefile`
2. So we do not run resolution multiple times (in `master` we run `docker inspect` multiple times for the same values)

perdasilva (approved):
thank you!!! I know it's been a journey XD

Issue #3010, opened 2023-08-14
Problems arising from deleting and rapidly creating a subscription


dtfranz:
### Description
I've noticed that occasionally, after a subscription is updated, then deleted and immediately recreated, the newly created subscription will be updated with the status of the old, deleted subscription. This will halt installation within the namespace, as the status will link the new subscription to the installPlan which was garbage-collected as a result of deleting the original subscription.

### Workaround
The issue can be resolved by simply deleting the subscription, then re-creating it after giving the controllers enough time to register the deletion event. Creating it with a different name should also ensure that the issue doesn't happen at all.

### Possible Cause
I believe this occurs because items in the cache are keyed by `namespace/name`, and it may therefore be possible for a controller to update the new subscription with an old status using a stale entry from the cache.

### Example
Following is an example of a subscription in this state. Note that no CSVs or InstallPlans were present in the namespace at the time. This was reproduced in an OpenShift 4.13 cluster with a catalog-operator image built from this repo as of commit hash `2be5e58`:

```yaml
apiVersion: operators.coreos.com/v1alpha1
kind: Subscription
metadata:
  creationTimestamp: "2023-08-10T18:23:19Z"
  generation: 1
  labels:
    operators.coreos.com/project-quay.openshift-operators: ""
  name: project-quay
  namespace: openshift-operators
  resourceVersion: "334099"
  uid: cab0d6e8-a551-4e39-ad2e-2f3c0a2caf27
spec:
  channel: stable-3.6
  installPlanApproval: Automatic
  name: project-quay
  source: community-operators
  sourceNamespace: openshift-marketplace
status:
  catalogHealth:
  - catalogSourceRef:
      apiVersion: operators.coreos.com/v1alpha1
      kind: CatalogSource
      name: certified-operators
      namespace: openshift-marketplace
      resourceVersion: "307403"
      uid: b1f09bd9-df57-4b1a-8520-a70f2038886b
    healthy: true
    lastUpdated: "2023-08-10T18:22:53Z"
  - catalogSourceRef:
      apiVersion: operators.coreos.com/v1alpha1
      kind: CatalogSource
      name: community-operators
      namespace: openshift-marketplace
      resourceVersion: "319530"
      uid: b3de47e6-43e6-4436-af68-ebaed5a8a7cd
    healthy: true
    lastUpdated: "2023-08-10T18:22:53Z"
  - catalogSourceRef:
      apiVersion: operators.coreos.com/v1alpha1
[...]

PR #3464, opened 2024-12-12
[WIP] cache-fixes

perdasilva:

**Architectural changes:**

**Testing remarks:**

**Reviewer Checklist**

### DIFF
--- a/pkg/lib/queueinformer/queueinformer_operator.go
+++ b/pkg/lib/queueinformer/queueinformer_operator.go
@@ -54,6 +54,8 @@ type Operator interface {
 	// RunInformers starts the Operator's underlying Informers.
 	RunInformers(ctx context.Context)
 
+	ResyncInformers() error
+
 	// Run starts the Operator and its underlying Informers.
 	Run(ctx context.Context)
 }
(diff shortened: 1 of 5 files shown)


pkg/lib/queueinformer/queueinformer_operator.go
```diff
@@ -197,6 +199,17 @@ func (o *operator) Run(ctx context.Context) {
 	})
 }
 
+func (o *operator) ResyncInformers() error {
```
stevekuznetsov, line 202: No client using LIST + WATCH should ever have to do this, FWIW.

stevekuznetsov:
Using a `Deployment` or anything other than a `Pod` is likely the better approach that won't cause grief in the future.

pkg/lib/queueinformer/queueinformer_operator.go
perdasilva: In case you're curious: the issue is we have a situation in which the namespaces gets nuked and immediately recreated with the subscription, and we're ending up with unsat...

perdasilva:
> Using a `Deployment` or anything other than a `Pod` is likely the better approach that won't cause grief in the future.

Sorry - the PR description was misleading because I fumbled the commit (--amend). What I'm trying to understand is this: https://github.com/operator-framework/operator-lifecycle-manager/issues/3010

pkg/lib/queueinformer/queueinformer_operator.go
stevekuznetsov: Are you using UID checks to detect deletion + recreation? Do you care that it deleted & recreated, or no?

perdasilva: in the end the RC was: the typed queues were of _any_ type and we were enqueuing two different types of objects (string and ResourceEvent). Two semantically equal events come in and get sync'ed concurrently because we break dedupe =P

PR #3486, opened 2025-01-10
bump k8s libs to v0.32.0

perdasilva:

Closes #3472 
Closes #3470 
Closes #3469 
Closes #3468 

Bumps k8s libs to v0.32.0. In this version, the apiserver speaks protobuf. So, the validatingroundtripper implementation had to be revisited to support it.

Generated code was regenerated for the new libraries with the `make gen-all` target.

**Architectural changes:**

**Testing remarks:**

**Reviewer Checklist**


pkg/controller/operators/validatingroundtripper/validating_round_tripper.go
```diff
@@ -13,23 +18,74 @@ import (
...
+	codecs   serializer.CodecFactory
+}
+
+func (rt *validatingRoundTripper) decodeYAMLOrJSON(data []byte) (*unstructured.Unstructured, error) {
+	bodyBuffer := bytes.NewBuffer(data)
+	dec := yaml.NewYAMLOrJSONDecoder(bodyBuffer, 10)
+	unstructuredObject := &unstructured.Unstructured{}
+	if err := dec.Decode(unstructuredObject); err != nil {
+		return nil, fmt.Errorf("error decoding yaml/json object to an unstructured object: %w: %+v", err, string(data))
+	}
```
joelanford, line 30: Same question for YAML/JSON/Protobuf: is it always a guarantee that the request body contains a single object? Or do we need to handle multiple objects and continuously call `Decode` until we see `io.EOF`?

pkg/controller/operators/validatingroundtripper/validating_round_tripper.go
```diff
@@ -13,23 +18,74 @@ import (
...
+	b, err := req.GetBody()
+	if err != nil {
+		panic(fmt.Errorf("failed to get request body: %w", err))
+	}
+	defer b.Close()
+
+	data, err := io.ReadAll(b)
+	if err != nil {
+		panic(fmt.Errorf("failed to read request body: %w", err))
+	}
```
joelanford, line 61: Is it possible to pass the decoders an `io.Reader` rather than reading everything into memory at once?

pkg/controller/operators/validatingroundtripper/validating_round_tripper.go
perdasilva: The old code seems to expect a single object. I guess because it's a POST? does that assumption stand?

pkg/controller/operators/validatingroundtripper/validating_round_tripper.go
joelanford: Looks like the UniversalDeserializer wants a `[]byte`, but the YAMLOrJSON decoder accepts an `io.Reader`. So maybe pass readers to both, but then have `decodeProtobuf` be the only thing that does `io.ReadAll` (and YAMLOrJSON can directly read from the request body)

pkg/controller/operators/validatingroundtripper/validating_round_tripper.go
joelanford: If old code expects a single object, we're probably safe to continue with that assumption.

pkg/controller/operators/validatingroundtripper/validating_round_tripper.go
perdasilva: I'll pick this up on Monday. I've converted it to a draft for now. There's something else bothering me as well: https://github.com/operator-framework/operator-lifecycle-manager/pull/3486/commits/2cc85d4184f17b921bbb3ed1b8a64621fe714f40

need to find a nicer way to do this =S the problem is here:

```
opClient, err := operatorclient.NewClientFromRestConfig(validatingConfig)
```

```
func NewClientFromRestConfig(config *rest.Config) (client ClientInterface, err error) {
	kubernetes, err := kubernetes.NewForConfig(config)
	if err != nil {
		return
	}

	apiextensions, err := apiextensions.NewForConfig(config)
	if err != nil {
		return
	}

	apiregistration, err := apiregistration.NewForConfig(config)
	if err != nil {
		return
	}

	client = &Client{
		kubernetes,
		apiextensions,
		apiregistration,
	}

	return
}
```

perdasilva: In the end, it seems easiest and best to put it in the catalog operator constructor. Not the prettiest, but at least it bunches it up in the same place making it easy to spot.

camilamacedo86 (dismissed):
/lgtm

pkg/controller/operators/catalog/operator.go
```diff
@@ -149,7 +151,9 @@ func NewOperator(ctx context.Context, kubeconfigPath string, clock utilclock.Clo
 	}
 
 	// create a config that validates we're creating objects with labels
-	validatingConfig := validatingroundtripper.Wrap(config)
+	_ = apiextensionsv1.AddToScheme(scheme)   // required by opClient
+	_ = apiregistrationv1.AddToScheme(scheme) // required by opClient
```
joelanford, line 155: Does this mutate the manager's scheme, or is this mutating a copy for use only by `NewOperator`? If the former (as I suspect), I think I'd put this in `main` because it would apply to anything else that uses the manager's scheme.

pkg/controller/operators/validatingroundtripper/validating_round_tripper.go
```diff
@@ -13,23 +17,71 @@ import (
...
+	var unstructuredObject *unstructured.Unstructured
+	switch req.Header.Get("Content-Type") {
+	case "application/vnd.kubernetes.protobuf":
+		unstructuredObject, err = rt.decodeProtobuf(b)
+	default:
+		unstructuredObject, err = rt.decodeYAMLOrJSON(b)
+	}
+
+	if err != nil {
+		panic(fmt.Errorf("failed to decode request body: %w", err))
```
joelanford, line 68: What was the determining factor between `panic` vs returning an error? Since we're dealing with network calls, it feels incorrect to panic.

joelanford: I guess the theory is that because this is a `RoundTripper` that operates on an `http.Request` that was built in memory by the same process, then there is always the expectation that we have a valid body.

However, a `RoundTripper` _could_ be used in a proxy scenario where the request is actually coming in from the network.

LalatenduMohanty:
https://github.com/operator-framework/operator-lifecycle-manager/pull/3486/commits/74a356dfb1e7be5b46fba7e136f3ae0a5abbe606 is for `update generated code`. Can you please add more information in the commit about how you updated the generated code. It will be helpful for future when someone trying to figure-out how it was done in the past.

pkg/controller/operators/validatingroundtripper/validating_round_tripper.go
perdasilva: I think this was originally done by SteveK with the memory optimization and resource labeling calls. I think this is specifically done on the outgoing side when creating resources. I think the idea was to ensure that all resources being created carried the labels (or die trying). I think, in the original intent, this is before the n/w call is made.

pkg/controller/operators/catalog/operator.go
perdasilva: Main just passes the standard scheme from client-go. Though, I think your comment still applies.

camilamacedo86 (approved):
I think the points raised are addressed now 
It seems great for me

/lgtm

camilamacedo86 (approved):
/lgtm

PR #3519, opened 2025-02-19
add missing pod disruption reasons to isPodDead

anik120:
isPodDead had checks for TerminationByKubelet and DeletionByTaintManager. It was missing checks for

PreemptionByScheduler
EvictionByEvictionAPI
DeletionByPodGC

ref: https://kubernetes.io/docs/concepts/workloads/pods/disruptions/#pod-disruption-conditions

This PR adds those reasons.

**Architectural changes:**

**Testing remarks:**

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/registry/reconciler/grpc.go
+++ b/pkg/controller/registry/reconciler/grpc.go
@@ -541,37 +541,12 @@ func imageChanged(logger *logrus.Entry, updatePod *corev1.Pod, servingPods []*co
 	return false
 }
 
+// isPodDead checks if the pod has the DisruptionTarget condition set to true,
+// which indicates that the Pod is about to be deleted due to a disruption.
+// ref: https://kubernetes.io/docs/concepts/workloads/pods/disruptions/#pod-disruption-conditions
 func isPodDead(pod *corev1.Pod) bool {
-	for _, check := range []func(*corev1.Pod) bool{
-		isPodDeletedByTaintManager,
-		isPodTerminatedByKubelet,
-	} {
-		if check(pod) {
-			return true
-		}
-	}
-	return false
-}
-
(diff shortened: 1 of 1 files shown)


perdasilva (approved):

pkg/controller/registry/reconciler/grpc.go
```diff
@@ -541,37 +541,12 @@ func imageChanged(logger *logrus.Entry, updatePod *corev1.Pod, servingPods []*co
...
-}
-
-// This reason is set when the Pod was evicted due to resource pressure on the Node
-func isPodTerminatedByKubelet(pod *corev1.Pod) bool {
-	if pod.DeletionTimestamp == nil {
-		return false
-	}
 	for _, condition := range pod.Status.Conditions {
-		if condition.Type == corev1.DisruptionTarget && condition.Reason == "TerminationByKubelet" && condition.Status == corev1.ConditionTrue {
+		if condition.Type == corev1.DisruptionTarget && condition.Status == corev1.ConditionTrue {
```
perdasilva, line 549: should we add some unit tests to make sure we don't regress?

anik120: We already have a test for this [here](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/controller/registry/reconciler/configmap_test.go#L551-L561) that tests this with the `DeleteionByTaintManager` reason [here](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/controller/registry/reconciler/grpc_test.go#L81).

ie that test is in fact making sure that the changes in this PR is not causing a regression :) 


We could technically expand the tests to perform the same test once for each reason, but I think that'll be overkill.

PR #3524, opened 2025-02-25
🐛 (fix): 'pod-security.kubernetes.io/*-version' should use supported k8s version instead of latest

camilamacedo86:
In this commit, a new makefile target was created to allow us to  automatically update the k8s PSA version values based on the K8s API version defined in the go.mod


kevinrizza:
@camilamacedo86 I think this change looks okay, but if we're going to touch these anyway, does it make sense for us to revisit https://github.com/operator-framework/operator-lifecycle-manager/pull/2906 at this point? We've default to baseline enforcement for the last ~2 years. At this point I expect that everyone should be using catalog binaries that can handle restricted enforcement

Maybe not part of this pr, but should we create an issue?

perdasilva (approved):
Thank you ^^

anik120 (approved):
/lgtm 

> At this point I expect that everyone should be using catalog binaries that can handle restricted enforcement

Note that we'll have to dig up why exactly the catalog binaries weren't able to handle the restricted environment, to confirm to ourselves that our customers are off of that issue (I can't remember off the top of my head what exactly the issue was)

jianzhangbjz:
Based on https://redhat-internal.slack.com/archives/C06KP34REFJ/p1741224828143489?thread_ts=1739880491.760029&cid=C06KP34REFJ, we are fine to leave our namespace manifest with latest since it doesn't appear hypershift is using it, "latest" means the-version-of-the-control-plane-that-is-running-psa-logic-for-this-namespace. So these pinning PRs appear not needed. So, we should revert this PR.

we should revert this change.

PR #3526, opened 2025-02-28
⚠️ Enforce PSA Restricted Level for CatalogSources (Drop Support for Catalogs built with opm < v1.21.0)

camilamacedo86:
#### What this PR does

This PR updates the OLM behavior to **enforce the Pod Security Admission (PSA) level `restricted`** instead of `baseline`. As a result, **OLM will no longer support CatalogSources built with `opm` versions older than `v1.21.0`**.

#### Why this matters

- **Security**: Enforcing `restricted` PSA level aligns with modern security best practices, significantly reducing pod permissions and enhancing container isolation.
- **Compatibility**: Older `opm` versions do not generate catalogs compatible with the `restricted` profile.
- **Modernization**: The `restricted` profile is already the default in OpenShift 4.11+ (which aligns with our vendor versions), and support for older setups is increasingly unnecessary.

#### Impact

- This change **breaks compatibility** with CatalogSources built using `opm` versions `< v1.21.0`.
- One existing test checks backward compatibility with these versions. This PR **removes that test** as the use case is deprecated.

#### Motivation

For the last two years, we've defaulted to baseline enforcement. At this point, I expect everyone to use catalog binaries that can handle restricted enforcement.

Motivated by: https://github.com/operator-framework/operator-lifecycle-manager/pull/3524#issuecomment-2683283636


camilamacedo86:
Hi @ @kevinrizza @perdasilva 

The follow up requested :-)

anik120 (changes requested):
Can we confirm that we're okay with moving to `restricted`? 

I'm concerned we'll take these changes downstream, and then hit the same issue we were facing 2 years ago and be forced to revert (and potentially back port)

anik120:
Looks like we might be facing an issue upstream too: 

```
Summarizing 1 Failure:
  [FAIL] Starting CatalogSource e2e tests when The namespace is labled as Pod Security Admission policy enforce:restricted when A CatalogSource built with opm v1.21.0 (<v1.23.2)is created with spec.GrpcPodConfig.SecurityContextConfig set to legacy [It] The registry pod comes up successfully [CatalogSource]
```

Could be a flake too. 
Again, without remembering what exactly the issue was 2 years ago it's hard to make an educated guess.

camilamacedo86:
@anik120 

I updated this test: https://github.com/operator-framework/operator-lifecycle-manager/pull/3526#issuecomment-2691298676 probably we need to remove or I did something wrong. But I agree we need to check this one properly.. we cannot only update. Thank you for call it out :-)

anik120:
> @anik120
> 
> I updated this test: [#3526 (comment)](https://github.com/operator-framework/operator-lifecycle-manager/pull/3526#issuecomment-2691298676) probably we need to remove or I did something wrong. But I agree we need to check this one properly.. we cannot only update. Thank you for call it out :-)

Right, it looks like the test is possibly actually telling us we're not ready to move to `restricted` yet.

camilamacedo86:
Hi @anik120 

See the error:

>  [FAIL] Starting CatalogSource e2e tests when The namespace is labled as Pod Security Admission policy enforce:restricted when A CatalogSource built with opm v1.21.0 (<v1.23.2)is created with spec.GrpcPodConfig.SecurityContextConfig set to legacy [It] The registry pod comes up successfully 

https://github.com/operator-framework/operator-lifecycle-manager/actions/runs/15064702811/job/42346975535?pr=3526

This failure is only in a test that checks if OLM works with CatalogSources built using older opm versions.

If we decide to move forward with this change motivated by the comment  [this comment](https://github.com/operator-framework/operator-lifecycle-manager/pull/3524#issuecomment-2683283636) , **we should remove that test.**

The key question is:
**Are we okay dropping support in OLM latest for CatalogSources built with opm versions older than v1.21.0?**

As mentioned in [this comment](https://github.com/operator-framework/operator-lifecycle-manager/pull/3524#issuecomment-2683283636) by @kevinrizza, we may now be ready to do that.

✅ Yes — remove the test and move on.
❌ No — close the PR for now and revisit later.

**Why this matters:**

- Older opm versions don’t work with enforceLevel: restricted,
- We want to follow better security practices,
- This would reduce permissions and make pods more secure.

Downside:

- OLM will no longer support CatalogSources built with opm < v1.21 which requires baseline. 

Would love to hear from folks who can help make the call.
@joelanford @tmshort @perdasilva WDYT?

perdasilva:
> Hi @anik120
> 
> See the error:
> 
> > [FAIL] Starting CatalogSource e2e tests when The namespace is labled as Pod Security Admission policy enforce:restricted when A CatalogSource built with opm v1.21.0 (<v1.23.2)is created with spec.GrpcPodConfig.SecurityContextConfig set to legacy [It] The registry pod comes up successfully
> 
> https://github.com/operator-framework/operator-lifecycle-manager/actions/runs/15064702811/job/42346975535?pr=3526
> 
> This failure is only in a test that checks if OLM works with CatalogSources built using older opm versions.
> 
> If we decide to move forward with this change motivated by the comment [this comment](https://github.com/operator-framework/operator-lifecycle-manager/pull/3524#issuecomment-2683283636) , **we should remove that test.**
> 
> The key question is: **Are we okay dropping support in OLM latest for CatalogSources built with opm versions older than v1.21.0?** (which seems to me fine 👍 - It is a very old version, and the changes to enforce PodSecurityContext were introduced in the OCP vendor 4.11 and 4.12)
> 
> As mentioned in [this comment](https://github.com/operator-framework/operator-lifecycle-manager/pull/3524#issuecomment-2683283636) by @kevinrizza, we may now be ready to do that.
> 
> ✅ Yes — remove the test and move on. ❌ No — close the PR for now and revisit later.
> 
> **Why this matters:**
> 
>     * Older opm versions don’t work with enforceLevel: restricted,
> 
>     * We want to follow better security practices,
> 
>     * This would reduce permissions and make pods more secure.
> 
> 
> Would love to hear from folks who can help make the call. @joelanford @tmshort @perdasilva WDYT?

I think that's fine with me. It makes sense to move (and to only support) the more secure opm. It shouldn't be easily enough to move catalogs to a newer version of opm if this breaks some old installations somewhere, and to the benefit of the user... imo, I'm good with this.

anik120 (dismissed):
Yup, ack. Thanks for the context @camilamacedo86 

/lgtm

camilamacedo86:
@perdasilva the test is failing in the test removed already, see; https://github.com/operator-framework/operator-lifecycle-manager/blob/d7aaeb11ab80f862fc9c70dfb7865158bb67e28c/test/e2e/catalog_e2e_test.go#L1629-L1674

```
Summarizing 1 Failure:
  [FAIL] Starting CatalogSource e2e tests when The namespace is labled as Pod Security Admission policy enforce:restricted when A CatalogSource built with opm v1.21.0 (<v1.23.2)is created with spec.GrpcPodConfig.SecurityContextConfig set to legacy [It] The registry pod comes up successfully [CatalogSource]
  /home/runner/work/operator-lifecycle-manager/operator-lifecycle-manager/test/e2e/catalog_e2e_test.go:1699

Ran 59 of 207 Specs in 594.500 seconds
FAIL! -- 58 Passed | 1 Failed | 5 Pending | 143 Skipped
```

Shows that the test has not been executed with the PR code :-( I did not get it. Help 🤚

camilamacedo86:
I closed and re-opened to trigger all tests again, lets see if now the action uses the source code 👍

joelanford:
This definitely needs an RFC since it is a breaking change. We need to carefully think through all of the implications. And we would ideally make a data driven decision about this. For example, Red Hat OCP collects telemetry. Can we tell how many CatalogSources are in baseline namespaces or how many explicitly set `spec.grpcPodConfig.securityContextConfig: legacy`?


Also this PR doesn't seem to actually do anything different with CatalogSources that use spec.grpcPodConfig.securityContextConfig: legacy`, so it doesn't feel like we're actually dropping support for those older catalogs. It's hard to tell what the intention is here based on a comparison of the PR changes and the PR description.

joelanford (changes requested):
See my above comment. I think we need a bit more due diligence and discussion to decide to make a breaking change.

camilamacedo86:
Hi @joelanford 

> Can we tell how many CatalogSources are in baseline namespaces or how many explicitly set spec.grpcPodConfig.securityContextConfig: legacy?

All Catalogues from OCP 4.11/4/12 should no longer use the OPM version < v1.21.0, which will result in it, right? 

> Also this PR doesn't seem to actually do anything different with CatalogSources that use spec.grpcPodConfig.securityContextConfig: legacy

This PR only changes the permissions; instead of the baseline, we use restricted.
We are just removing the tests that ensure exactly it, we are still supporting catalog source built with OPM version < v1.21.0
Motivation: https://github.com/operator-framework/operator-lifecycle-manager/pull/3524#issuecomment-2683283636
PR (context): https://github.com/operator-framework/operator-lifecycle-manager/pull/2906

After the above summary, do we still need to check if any CatalogSource distributed from 4.19 was built with OPM < 1.21, which requires baseline SCC and cannot run with restricted permissions? Would an RFC still be required in this case?

camilamacedo86:
Due to OLMv0 being in maintenance mode, we decided not to introduce this breaking change. 
Closing it as deferred. ([more](https://redhat-internal.slack.com/archives/C06KP34REFJ/p1747838662808939))

Issue #3284, opened 2024-05-27
Unpin google.golang.org/grpc from go.mod


perdasilva:
### Description

Version 1.64.0 of google.golang.org/grpc changes the way the CatalogSource connection state behaves. 
previously it would automatically reach READY - but with the new library, it seems to stay on IDLE unless there is data going down the pipe this is breaking many CatalogSource related e2e-tests and also affecting the way CatalogSource metrics work (since it introduces a more frequent transitions between IDLE and READY the READY/NOTREADY gauges go a bit loopy)

### Possible Fixes

Either: 
 - A CatalogSourceState of "READY" would now mean gRPC connection state in {READY, IDLE} (unclear whether this makes sense)
 - Or, introduce a periodic health-check (but, Pods already have health-checks!)
 - Or, switch to looking at the pod for CatalogSourceState (i.e. READY == Pod is READY, TRANSIENT_ERROR == Pod not ready) (But, what about the address only CatalogSources...?)

camilamacedo86:
Closing this one since we remove it already: https://github.com/camilamacedo86/operator-lifecycle-manager/commit/87be2e2fd9f315dc4eee31443b4ba19959088e9d

PR #3529, opened 2025-03-05
🌱 remove no longer required grpc replace (issue: #3284)

camilamacedo86:
Closes: #3284


stevekuznetsov:
I would appreciate it if the robots could be taught not to request my review these days as I no longer work on this project.

pkg/controller/registry/grpc/source.go
```diff
@@ -163,7 +163,7 @@ func grpcConnection(address string) (*grpc.ClientConn, error) {
 		}))
 	}
 
-	return grpc.Dial(address, dialOptions...)
+	return grpc.NewClient(address, dialOptions...)
```
joelanford, line 166: I'm not sure we should make this change. I see that `grpc.Dial` is deprecated, but reading the GoDoc of `grpc.DialContext` (which is called by `grpc.Dial`), it sounds like there are subtle differences that may matter to us:

https://github.com/grpc/grpc-go/blob/c52d02553f2649c28d0279bca17909a3a08de022/clientconn.go#L228-L236

> // One subtle difference between NewClient and Dial and DialContext is that the
> // former uses "dns" as the default name resolver, while the latter use
> // "passthrough" for backward compatibility.  **This distinction should not matter
> // to most users, but could matter to legacy users that specify a custom dialer
> // and expect it to receive the target string directly.**

We _do_ use a custom dialer if there is a proxyURL.

PR #3556, opened 2025-04-10
✨ handle optionality of spec.grpcPodConfig.extractContent.cacheDir

grokspawn:
Allows grpcPodConfig.extractContent.cacheDir to be optional, in case folks are 
1. using custom catalog images w/o an existing `opm serve` cache; or
2. wish to force re-generation of catalog cache based on copied catalog content

To reduce the size of catalogsource images, we'd like to be able to provide OCI images which contain neither binary nor prebuilt cache.  With improvements in cache-build pod startup, we should be able to operate without a pre-existing cache. 

**Architectural changes:**

**Testing remarks:**

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/registry/reconciler/reconciler.go
+++ b/pkg/controller/registry/reconciler/reconciler.go
@@ -275,6 +275,17 @@ func Pod(source *operatorsv1alpha1.CatalogSource, name, opmImg, utilImage, img s
 				Name:      "catalog-content",
 				MountPath: catalogPath,
 			}
+			// init container to copy catalog info.
+			// ExtractContent.CatalogDir is mandatory when ExtractContent is provided
+			// ExtractContent.CacheDir is optional, so we only add it if it is set
+			var extractArgs = []string{
+				"--catalog.from=" + grpcPodConfig.ExtractContent.CatalogDir,
+				"--catalog.to=" + fmt.Sprintf("%s/catalog", catalogPath),
+			}
+			if grpcPodConfig.ExtractContent.CacheDir != "" {
+				extractArgs = append(extractArgs, "--cache.from="+grpcPodConfig.ExtractContent.CacheDir)
+				extractArgs = append(extractArgs, "--cache.to="+fmt.Sprintf("%s/cache", catalogPath))
+			}
 			pod.Spec.InitContainers = append(pod.Spec.InitContainers, corev1.Container{
 				Name:                     "extract-utilities",
 				Image:                    utilImage,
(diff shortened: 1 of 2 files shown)


grokspawn:
Needs a PR in API to make this field optional before this will be functional, so hold this until that has merged.

pkg/controller/registry/reconciler/reconciler.go
```diff
@@ -275,6 +275,17 @@ func Pod(source *operatorsv1alpha1.CatalogSource, name, opmImg, utilImage, img s
 				Name:      "catalog-content",
 				MountPath: catalogPath,
 			}
+			// init container to copy catalog info.
+			// ExtractContent.CatalogDir is mandatory when ExtractContent is provided
+			// ExtractContent.CacheDir is optional, so we only add it if it is set
+			var extractArgs = []string{
+				"--catalog.from=" + grpcPodConfig.ExtractContent.CatalogDir,
+				"--catalog.to=" + fmt.Sprintf("%s/catalog", catalogPath),
```
camilamacedo86, line 283: Do we already have a check for this?

If ExtractContent is not empty and ExtractContent.CatalogDir is empty, we should raise an error with the following message:

"CatalogDir must be set when ExtractContent is provided."

grokspawn: We do have that because CatalogDir is [listed as required](https://github.com/operator-framework/operator-lifecycle-manager/pull/3556/files#diff-cfdc2cf5582e0f68ab2efd954f992741d2d3053a5358c5a106ec64a3ab5d3826R978)

pkg/controller/registry/reconciler/reconciler.go
```diff
@@ -283,22 +294,18 @@ func Pod(source *operatorsv1alpha1.CatalogSource, name, opmImg, utilImage, img s
 				VolumeMounts:             []corev1.VolumeMount{utilitiesVolumeMount},
 				TerminationMessagePolicy: corev1.TerminationMessageFallbackToLogsOnError,
 			}, corev1.Container{
-				Name:            "extract-content",
-				Image:           img,
-				ImagePullPolicy: image.InferImagePullPolicy(img),
-				Command:         []string{utilitiesPath + "/copy-content"},
-				Args: []string{
-					"--catalog.from=" + grpcPodConfig.ExtractContent.CatalogDir,
-					"--catalog.to=" + fmt.Sprintf("%s/catalog", catalogPath),
```
camilamacedo86, line 292: This both we still have 👍

pkg/controller/registry/reconciler/reconciler.go
```diff
@@ -283,22 +294,18 @@ func Pod(source *operatorsv1alpha1.CatalogSource, name, opmImg, utilImage, img s
...
-					"--cache.from=" + grpcPodConfig.ExtractContent.CacheDir,
-					"--cache.to=" + fmt.Sprintf("%s/cache", catalogPath),
```
camilamacedo86, line 294: Thsi both are under `if grpcPodConfig.ExtractContent.CacheDir != "" {` 👍

camilamacedo86 (commented):
Hi @grokspawn 

All shows fine for me here 👍 
We just need rebase it with master
Then, all OK

Just a NIT: The emoji here I think should be ✨ OR 🐛 since 🌱  would be for case scenarios where has 0 impact in end users like ci fixes for example or bumps.

camilamacedo86 (approved):
/lgtm

PR #3558, opened 2025-04-24
🐛 fix(olm): improve error logging for missing olm.managed label

camilamacedo86:
Previously, nonconforming CRDs (missing the  label) were logged as INFO, but this caused the  to enter a CrashLoopBackOff state due to unhandled inconsistencies.

This commit raises the log level to  and enhances the message with actionable advice, clarifying that users should either delete CRDs for uninstalled solutions or label them appropriately.

The root cause of the scenario is: From an old release (ocp 4.15) the component managed by OLM should have the label olm.managed: true added to comply with the new checks.

### DIFF
--- a/pkg/controller/operators/labeller/filters.go
+++ b/pkg/controller/operators/labeller/filters.go
@@ -7,6 +7,7 @@ import (
 	"sync"
 
 	operators "github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/clientset/versioned"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/install"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/registry/reconciler"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/sync/errgroup"
@@ -177,7 +178,11 @@ func Validate(ctx context.Context, logger *logrus.Logger, metadataClient metadat
 				logger.WithFields(logrus.Fields{
 					"gvr":           gvr.String(),
 					"nonconforming": count,
-				}).Info("found nonconforming items")
+				}).Errorf(
+					"found nonconforming items: missing the the required label %q (metadata.labels[\"%s\"]=\"true\"). ",
+					install.OLMManagedLabelKey,
+					install.OLMManagedLabelKey,
+				)
 			}
 			okLock.Lock()
 			ok = ok && count == 0


pkg/controller/operators/labeller/filters.go
```diff
@@ -177,7 +178,12 @@ func Validate(ctx context.Context, logger *logrus.Logger, metadataClient metadat
 				logger.WithFields(logrus.Fields{
 					"gvr":           gvr.String(),
 					"nonconforming": count,
-				}).Info("found nonconforming items")
+				}).Errorf(
+					"found nonconforming items: resource(s) are missing the %q label (metadata.labels[\"%s\"]=\"true\"). "+
+						"You may want to consider deleting it (if this CRD is for an uninstalled solution) or labeling it appropriately.",
+					install.OLMManagedLabelKey,
+					install.OLMManagedLabelKey,
+				)
```
bentito, line 186: Hmm, the phrasing seems awkward to me, how about?:
```go
}).Errorf(
	"Found resources missing the required label %q (`metadata.labels[%s] = \"true\"`). Please label them appropriately, or consider deleting them if they relate to a CRD from an uninstalled solution.",
	install.OLMManagedLabelKey,
	install.OLMManagedLabelKey,
)
```

tmshort: I agree, and I don't think we need the remediation in the log itself.

bentito: @tmshort so just?:
```go
}).Errorf(
	"Found resources missing the required label %q (`metadata.labels[%s] = \"true\"`),
	install.OLMManagedLabelKey,
	install.OLMManagedLabelKey,
)
```

pkg/controller/operators/labeller/filters.go
```diff
@@ -177,7 +178,12 @@ func Validate(ctx context.Context, logger *logrus.Logger, metadataClient metadat
 				logger.WithFields(logrus.Fields{
 					"gvr":           gvr.String(),
 					"nonconforming": count,
-				}).Info("found nonconforming items")
+				}).Errorf(
+					"found nonconforming items: missing the the required label %q (metadata.labels[\"%s\"]=\"true\"). "+
+						"Please label them appropriately, or consider deleting them if they are relate to a CRD from an uninstalled solution.",
```
bentito, line 183:
```suggestion
						"Please label them appropriately, or consider deleting them if they are related to a CRD from an uninstalled solution.",
```

bentito: But I think @tmshort is saying, maybe, we don't want the whole corrective action part, the "Please..." part to show up in the logs.

pkg/controller/operators/labeller/filters.go
tmshort: @bentito, I think your version is better (but still has a type), but it's probably not needed?
Is there a document somewhere with this information?
If this occurs, I think it's safer not to offer remediation, because the advice might change over time (whereas the code won't).

tmshort: If we are going to include a suggestion, propose it as an action rather than familiarly.
```
"Resources (%s) missing required label %q (`metadata.labels[%s] = \"true\"`). Add the label or delete unused resources.",
        <list of missing labels>
	install.OLMManagedLabelKey,
	install.OLMManagedLabelKey,
```
Or something similar. It would be good to list the resources.

camilamacedo86: It is a log so it already has the info about the resource see:
```
logger.WithFields(logrus.Fields{
					"gvr":           gvr.String(), <- HERE IS THE INFO
					"nonconforming": count,
				}).Info("found nonconforming items")
```
Therefore, the suggestion above does not seems fit
By last we should NOT remove the text `"found nonconforming items` that is how people find the KCS.

camilamacedo86 (commented):
Hi @tmshort and @bentito,

Since you preferred not to include the suggestion on how to fix the scenario, that part has been removed. However, just a few points to highlight:

- The information about the GVR is already present
- We **should not remove** the phrase **"found nonconforming items"**, as that is essential for users to find the related KCS article and resolve the issue: [KCS 7112019](https://access.redhat.com/solutions/7112019).

With these adjustments, I hope we can move forward with this small fix.  
Could you please let me know if it has your **LGTM/approval**?

Thank you very much for your help and collaboration!

perdasilva (approved):

PR #3559, opened 2025-04-24
Revert "CVE-2025-27144: pin go-jose/v4@v4.0.5"

anik120:
Reverts operator-framework/operator-lifecycle-manager#3550

go-jose/v4@v4.0.5 is an indirect-indirect dependency, and does not need to be
`replace`-ed since it's absent to begin with.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -190,7 +190,6 @@ require (
 replace google.golang.org/grpc => google.golang.org/grpc v1.63.2
 
 replace (
-	github.com/go-jose/go-jose/v4 => github.com/go-jose/go-jose/v4 v4.0.5 // CVE-2025-27144
 	// controller runtime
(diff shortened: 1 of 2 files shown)


tmshort (approved):
/lgtm

camilamacedo86 (approved):

go.mod
```diff
@@ -190,7 +190,6 @@ require (
 replace google.golang.org/grpc => google.golang.org/grpc v1.63.2
 
 replace (
-	github.com/go-jose/go-jose/v4 => github.com/go-jose/go-jose/v4 v4.0.5 // CVE-2025-27144
 	// controller runtime
```
camilamacedo86, line 193: We can just bump in the indirect call. 
It has no need for replaces

tmshort: That was the point... but in this case there is no reference to it anywhere, so we are removing it.

PR #3568, opened 2025-04-30
✨ Include network policy for all configmap and grpc catalogsources

joelanford:

This change updates OLMv0 to generate and reconcile two `NetworkPolicy` objects for each `CatalogSource` where OLMv0 also manages a catalog pod (i.e. configmap and grpc-based catalog sources).
1. GRPC server pods need to listen and receive requests on port 50051
2. Bundle unpack pods need to make connections to the kubernetes API server.

Related implementation details:
- Informer events from network policies are handled by the catalog operator, so any deletions or changes are reverted.
- Each NetworkPolicy gets an owner reference so that it is automatically cleaned up when the catalog source is deleted.
- Each NetworkPolicy uses a pod selector that specifically targets the appropriate pods, so there should be no accidental blocking of traffic of unrelated pods.

Unit tests are updated to ensure that NetworkPolicy objects are handled correctly.

By adding NetworkPolicy, we can provide more security to mitigate vulnerabilities and avoid accidental data leaks.

**Architectural changes:**

None (unless you count managing NetworkPolicy for CatalogSources as an architectural change)

**Testing remarks:**

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/registry/reconciler/configmap_test.go
+++ b/pkg/controller/registry/reconciler/configmap_test.go
@@ -199,7 +199,8 @@ func objectsForCatalogSource(t *testing.T, catsrc *v1alpha1.CatalogSource) []run
 	switch catsrc.Spec.SourceType {
 	case v1alpha1.SourceTypeInternal, v1alpha1.SourceTypeConfigmap:
 		decorated := configMapCatalogSourceDecorator{catsrc, runAsUser}
-		np := decorated.NetworkPolicy()
+		grpcServerNetworkPolicy := decorated.GRPCServerNetworkPolicy()
+		unpackBundlesNetworkPolicy := decorated.UnpackBundlesNetworkPolicy()
 		service, err := decorated.Service()
 		if err != nil {
 			t.Fatal(err)
@@ -210,7 +211,8 @@ func objectsForCatalogSource(t *testing.T, catsrc *v1alpha1.CatalogSource) []run
 			t.Fatal(err)
 		}
 		objs = append(objs,
-			np,
+			grpcServerNetworkPolicy,
+			unpackBundlesNetworkPolicy,
 			pod,
 			service,
 			serviceAccount,
@@ -218,7 +220,8 @@ func objectsForCatalogSource(t *testing.T, catsrc *v1alpha1.CatalogSource) []run
 	case v1alpha1.SourceTypeGrpc:
 		if catsrc.Spec.Image != "" {
 			decorated := grpcCatalogSourceDecorator{CatalogSource: catsrc, createPodAsUser: runAsUser, opmImage: ""}
-			np := decorated.NetworkPolicy()
+			grpcServerNetworkPolicy := decorated.GRPCServerNetworkPolicy()
+			unpackBundlesNetworkPolicy := decorated.UnpackBundlesNetworkPolicy()
 			serviceAccount := decorated.ServiceAccount()
 			service, err := decorated.Service()
 			if err != nil {
@@ -229,7 +232,8 @@ func objectsForCatalogSource(t *testing.T, catsrc *v1alpha1.CatalogSource) []run
 				t.Fatal(err)
 			}
 			objs = append(objs,
-				np,
+				grpcServerNetworkPolicy,
+				unpackBundlesNetworkPolicy,
 				pod,
 				service,
 				serviceAccount,
@@ -351,7 +355,7 @@ func TestConfigMapRegistryReconciler(t *testing.T) {
 			},
 		},
 		{
-			testName: "ExistingRegistry/BadNetworkPolicy",
+			testName: "ExistingRegistry/BadNetworkPolicies",
 			in: in{
 				cluster: cluster{
 					k8sObjs: append(setLabel(objectsForCatalogSource(t, validCatalogSource), &networkingv1.NetworkPolicy{}, CatalogSourceLabelKey, "wrongValue"), validConfigMap),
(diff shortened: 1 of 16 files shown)


pkg/controller/registry/reconciler/helpers.go
```diff
@@ -0,0 +1,65 @@
...
+					},
+				},
+			},
+		},
+	}
+	ownerutil.AddOwner(np, catalogSource, false, false)
+	return np
+}
+
+func sanitizedDeepEqual(a client.Object, b client.Object) bool {
```
dinhxuanvu, line 46: Given MatchLabels matters here and OLM has tendency to add new fields to labels from time to time, it may cause issue. I recommend adding a check for MatchLabels to ensure OLM will update NetworkPolicy if MatchLabels is updated.

dinhxuanvu (commented):
Looking good. A few comments.

pkg/controller/registry/reconciler/configmap_test.go
```diff
@@ -342,6 +350,24 @@ func TestConfigMapRegistryReconciler(t *testing.T) {
...
+				},
+				catsrc: validCatalogSource,
+			},
+			out: out{
+				status: &v1alpha1.RegistryServiceStatus{
+					CreatedAt:        now(),
+					Protocol:         "grpc",
+					ServiceName:      "cool-catalog",
+					ServiceNamespace: testNamespace,
+					Port:             "50051",
```
dinhxuanvu, line 367: Nit: We use this port number across multiple places for Catalog-related services and such. Would love to have a const for this to synchronize them if we ever decide to change it. However, it is not a big deal. Just a note here.

pkg/controller/registry/reconciler/helpers.go
dinhxuanvu: Now think of it, Semantic.DeepEqual check is probably enough here.

joelanford: Thanks for the review @dinhxuanvu!

If I'm understanding your comment correctly, I think this function would handle the case you are talking about.

What is essentially being ignored here are all the fields that we expect to be defaulted or managed by KAS. Every other diff (like us changing a label) will be different between `a` and `b` and thus cause this function to return `false` (for "not equal", and thus "need to update")

pkg/controller/registry/reconciler/configmap_test.go
joelanford: Noted, and agreed. That change might be better left as its own PR though.

dinhxuanvu (dismissed):
/approve

test/e2e/catalog_e2e_test.go
```diff
@@ -750,6 +753,119 @@ var _ = Describe("Starting CatalogSource e2e tests", Label("CatalogSource"), fun
 		Expect(registryPods.Items).To(HaveLen(1), "unexpected number of replacement registry pods found")
 	})
 
+	It("delete registry network policy triggers recreation", func() {
```
perdasilva, line 756: do we need an e2e? could a unit test cover this case?

perdasilva:
@joelanford is this ready for review? it looks pretty good to me =D

test/e2e/catalog_e2e_test.go
joelanford: Maybe? I didn't look too closely. The main bit of coverage I wanted here was "is there an NP event that the catalog-operator notices and causes there to be a sync of the relevant CatSrc"

It seemed like there were other e2e's for similar catsrc reconcile behaviors, so I followed the pattern.

pkg/controller/registry/reconciler/helpers.go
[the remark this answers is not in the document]
joelanford: Yes, that is possible. I looked at using `equality.Semantic.DeepDerivative()` instead, but I didn't want to allow someone to _add_ an extra network policy rule or change the selector in a way that would say "all good" from a deep derivative perspective.

We could potentially write a diff detector in a different way that would be less susceptible to future API changes, but also still secure in terms of reverting undesirable changes.

joelanford:
> is this ready for review? it looks pretty good to me =D

Yes-ish. I think the catalog pod network policy is reviewable. But I realized I also need to add dynamic network policy handling for bundle unpack pods.

I could do that here, or I could do it in a separate PR. I don't really care which.

I also have this listed as a draft because the RFC is still in review: https://docs.google.com/document/d/1TCb2YsSHaXBUnUUFp44Bt2hdsb2jwLnL77imP25QGVs/edit?tab=t.0#heading=h.x3tfh25grvnv

pkg/controller/registry/reconciler/helpers.go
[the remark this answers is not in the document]
joelanford: I changed this to be specific to network policy and exactly what we care about:
1. Our labels are there (but others can be added)
2. The specs are identical

joelanford:
Ready for review again!

RFC is here: https://docs.google.com/document/d/1TCb2YsSHaXBUnUUFp44Bt2hdsb2jwLnL77imP25QGVs/edit?tab=t.0#heading=h.x3tfh25grvnv

test/e2e/catalog_e2e_test.go
perdasilva: I'm just worried about adding additional e2es XD but I'd also say that if the effort to make it a unit test is too big, it's not worth the effort...let's proceed as it is

perdasilva (approved):

PR #3597, opened 2025-06-10
WIP: Update of PR 3497

tmshort:
Alternative to #3497

**Architectural changes:**

**Testing remarks:**

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -230,38 +230,53 @@ func NewOperator(ctx context.Context, kubeconfigPath string, clock utilclock.Clo
 
 	// Fields are pruned from local copies of the objects managed
 	// by this informer in order to reduce cached size.
-	prunedCSVInformer := cache.NewSharedIndexInformer(
-		pruning.NewListerWatcher(op.client, metav1.NamespaceAll,
+	prunedCSVInformer := cache.NewSharedIndexInformerWithOptions(
+		pruning.NewListerWatcher(
+			op.client,
+			metav1.NamespaceAll,
 			func(options *metav1.ListOptions) {
 				options.LabelSelector = fmt.Sprintf("!%s", v1alpha1.CopiedLabelKey)
 			},
-			pruning.PrunerFunc(func(csv *v1alpha1.ClusterServiceVersion) {
-				*csv = v1alpha1.ClusterServiceVersion{
-					TypeMeta: csv.TypeMeta,
-					ObjectMeta: metav1.ObjectMeta{
-						Name:        csv.Name,
-						Namespace:   csv.Namespace,
-						Labels:      csv.Labels,
-						Annotations: csv.Annotations,
-					},
-					Spec: v1alpha1.ClusterServiceVersionSpec{
-						CustomResourceDefinitions: csv.Spec.CustomResourceDefinitions,
-						APIServiceDefinitions:     csv.Spec.APIServiceDefinitions,
-						Replaces:                  csv.Spec.Replaces,
-						Version:                   csv.Spec.Version,
-					},
-					Status: v1alpha1.ClusterServiceVersionStatus{
-						Phase:  csv.Status.Phase,
-						Reason: csv.Status.Reason,
-					},
-				}
-			})),
+		),
 		&v1alpha1.ClusterServiceVersion{},
-		resyncPeriod(),
-		cache.Indexers{
-			cache.NamespaceIndex: cache.MetaNamespaceIndexFunc,
+		cache.SharedIndexInformerOptions{
+			ResyncPeriod: resyncPeriod(),
+			Indexers: cache.Indexers{
+				cache.NamespaceIndex: cache.MetaNamespaceIndexFunc,
+			},
 		},
 	)
+
+	// Transformed the CSV to be just the necessary data
+	prunedCSVTransformFunc := func(i interface{}) (interface{}, error) {
+		if csv, ok := i.(*v1alpha1.ClusterServiceVersion); ok {
+			*csv = v1alpha1.ClusterServiceVersion{
+				TypeMeta: csv.TypeMeta,
+				ObjectMeta: metav1.ObjectMeta{
+					Name:        csv.Name,
+					Namespace:   csv.Namespace,
+					Labels:      csv.Labels,
+					Annotations: csv.Annotations,
+				},
+				Spec: v1alpha1.ClusterServiceVersionSpec{
+					CustomResourceDefinitions: csv.Spec.CustomResourceDefinitions,
+					APIServiceDefinitions:     csv.Spec.APIServiceDefinitions,
+					Replaces:                  csv.Spec.Replaces,
+					Version:                   csv.Spec.Version,
+				},
+				Status: v1alpha1.ClusterServiceVersionStatus{
+					Phase:  csv.Status.Phase,
+					Reason: csv.Status.Reason,
+				},
+			}
+			return csv, nil
+		}
+		return nil, fmt.Errorf("Unable to convert input to CSV")
+	}
+
+	if err := prunedCSVInformer.SetTransform(prunedCSVTransformFunc); err != nil {
+		return nil, err
+	}
 	csvLister := operatorsv1alpha1listers.NewClusterServiceVersionLister(prunedCSVInformer.GetIndexer())
 	op.lister.OperatorsV1alpha1().RegisterClusterServiceVersionLister(metav1.NamespaceAll, csvLister)
 	if err := op.RegisterInformer(prunedCSVInformer); err != nil {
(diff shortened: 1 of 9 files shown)


grokspawn:
/lgtm
/hold

holding this long enough for some others to take a peek (including @camilamacedo86 who had comments on its predecessor).

tmshort:
I might still need to make a few minor tweaks... specifically with getting the hash from the primary CSV... looking at the code, something doesn't seem right...
EDIT: I could be conflating the catalog and olm operator code (both of which were modified but only the olm code uses the hashes)

pkg/controller/operators/internal/pruning/listerwatcher.go
joelanford, line 1: This looks like a fairly standard listerwatcher at this point right? I wonder if there's something in client-go, apimachinery, or controller-runtime that can replace this entire package now?

tmshort: I was thinking the same, but didn't spend time looking for it. The only thing is the `override()` of the options, but I think that might be pretty standard.

joelanford: I don't see anything. What you have there looks pretty idiomatic to me based on what I see elsewhere. I think my only suggestion then would be to rename the package because it is no longer handling any pruning concepts directly.

joelanford:
I'm testing locally, and I think I'm seeing something unexpected?

I installed OLM from the PR, and then installed the tektoncd-operator in the `operators` namespace. That causes CSV copies to end up in all the other namespaces.

From there I edit the copied CSV in the `default` namespace:
1. I change an annotation value, OLM does not change it back
2. I change a spec value, OLM does not change it back

OLM is supposed to see these changes and revert them, right?

joelanford:
Looking at it more, it looks like the change _eventually_ gets reverted, but it looks like something reconciles the main CSV ([looks like 30s](https://github.com/operator-framework/operator-lifecycle-manager/blob/3775a4d31f6625cce96c7f3e80c96e74038c4a6e/pkg/controller/operators/olm/config.go#L104)) and _that_ sync ends up detecting/reverting the change.

pkg/controller/operators/olm/operator.go
```diff
@@ -293,20 +304,54 @@ func newOperatorWithConfig(ctx context.Context, config *operatorConfig) (*Operat
...
+			pruning.NewListerWatcher(
+				op.client,
+				namespace,
+				func(opts *metav1.ListOptions) {
+					opts.LabelSelector = v1alpha1.CopiedLabelKey
+				},
+			),
+			&v1alpha1.ClusterServiceVersion{},
+			cache.SharedIndexInformerOptions{
+				ResyncPeriod: config.resyncPeriod(),
```
joelanford, line 320: I'm a bit worried about this resync period, if it is 30s as I suspect. That means we'll be re-listing all copied CSVs from the apiserver every 30s, which seems like it would be _much_ worse that re-listing all PartialObjectMetadata for copied CSVs every 30s. Neither is great though.

Do we actually need to resync that often? My feeling is that the resync interval should be something like 10h, but that we should also have a syncer tied to the workqueue such that we react to changes immediately.

pkg/controller/operators/catalog/operator.go
```diff
@@ -230,38 +230,53 @@ func NewOperator(ctx context.Context, kubeconfigPath string, clock utilclock.Clo
...
-					},
-				}
-			})),
+		),
 		&v1alpha1.ClusterServiceVersion{},
-		resyncPeriod(),
-		cache.Indexers{
-			cache.NamespaceIndex: cache.MetaNamespaceIndexFunc,
+		cache.SharedIndexInformerOptions{
+			ResyncPeriod: resyncPeriod(),
```
joelanford, line 243: Same concern here as I mentioned in the OLM operator. I fear that asking for a relist of the full copied CSVs from the apiserver every 30s is a pretty significant performance concern.

tmshort: AFAICT, 15m is the default value in the catalog operator, and was just copied from the original code. I can certainly change it to 6h or 10h or something similar. But you have concerns about the "reverts" not occurring in a reasonable time?

tmshort:
> Looking at it more, it looks like the change _eventually_ gets reverted, but it looks like something reconciles the main CSV ([looks like 30s](https://github.com/operator-framework/operator-lifecycle-manager/blob/3775a4d31f6625cce96c7f3e80c96e74038c4a6e/pkg/controller/operators/olm/config.go#L104)) and _that_ sync ends up detecting/reverting the change.

The `csvCopyQueue` has `syncCopyCSV()` as a syncer, which should trigger the syncing to the main CSV.

pkg/controller/operators/olm/operator.go
tmshort: There seems to be two defaults for the OLM one. The default, default, is `30s`:
https://github.com/operator-framework/operator-lifecycle-manager/blob/3775a4d31f6625cce96c7f3e80c96e74038c4a6e/pkg/controller/operators/olm/config.go#L104

When the command line arguments are taken into account, the default is `5m`:
https://github.com/operator-framework/operator-lifecycle-manager/blob/3775a4d31f6625cce96c7f3e80c96e74038c4a6e/cmd/olm/main.go#L36

pkg/controller/operators/catalog/operator.go
tmshort: https://github.com/operator-framework/operator-lifecycle-manager/blob/3775a4d31f6625cce96c7f3e80c96e74038c4a6e/cmd/catalog/main.go#L30

pkg/controller/operators/internal/pruning/listerwatcher.go
tmshort: There's `cache.NewFilteredListWatchFromClient()` which is _close_, but requires a more generic client.

pkg/controller/operators/catalog/operator.go
joelanford: I mentioned somewhere else, but I looked into the resync period and confirmed that resync is a client-only operation. It causes the informer to add all objects in its cache back into the workqueue. It does _not_ cause a re-list from the API server.

pkg/controller/operators/olm/operator.go
joelanford: I mentioned somewhere else, but I looked into the resync period and confirmed that resync is a client-only operation. It causes the informer to add all objects in its cache back into the workqueue. It does not cause a re-list from the API server.

pkg/controller/operators/olm/operator.go
```diff
@@ -1864,17 +1930,52 @@ func (a *Operator) createCSVCopyingDisabledEvent(csv *v1alpha1.ClusterServiceVer
...
 	}
-	return
+
+	// check for any garbage collection
+	err := a.removeDanglingChildCSVs(csv)
+	if err != nil {
+		return err
+	}
+
+	// Trigger partent reconciliation
```
joelanford:
```suggestion
	// Trigger parent reconciliation
```

pkg/controller/operators/olm/operator.go
```diff
@@ -1864,17 +1930,52 @@ func (a *Operator) createCSVCopyingDisabledEvent(csv *v1alpha1.ClusterServiceVer
...
+	csv, ok := obj.(*v1alpha1.ClusterServiceVersion)
+	if !ok {
+		a.logger.Debugf("casting ClusterServiceVersion failed: wrong type: %#v", obj)
+		return
+	}
+	if !v1alpha1.IsCopied(csv) {
+		return
+	}
+
+	// Trigger partent reconciliation
```
joelanford:
```suggestion
	// Trigger parent reconciliation
```

pkg/controller/operators/olm/operator.go
```diff
@@ -1864,17 +1930,52 @@ func (a *Operator) createCSVCopyingDisabledEvent(csv *v1alpha1.ClusterServiceVer
...
+	// Requeue parent CSV to deal with any changes to the copied CSV
+	copiedFromNamespace, ok := csv.GetLabels()[v1alpha1.CopiedLabelKey]
+	if !ok {
+		a.logger.Infof("no %q label found in CSV, skipping requeue", v1alpha1.CopiedLabelKey)
+		return nil
+	}
+	return a.csvCopyQueueSet.Requeue(copiedFromNamespace, name)
+}
+
+func (a *Operator) deleteCopiedCsv(obj interface{}) {
```
joelanford: This name is misleading. This function doesn't delete the copied CSV, it handles the deletion event when something _else_ deletes the copied CSV. I'd suggest the following:
```suggestion
func (a *Operator) handleCopiedCSVDeletion(obj interface{}) {
```

PR #3599, opened 2025-06-12
Set ImagePullPolicy ifNotPresent for server container

kevinrizza:
When the change to use the copy content sidecar container was added for the extract content option it properly set the ifNotPresent pull policy for the bundle image, but the actual server container is getting the wrong value and is not based on the opm image passed to the catalog operator. This commit updates the main registry server container to variably set the container to pull only if not present.


pkg/controller/registry/reconciler/reconciler_test.go
```diff
@@ -951,7 +977,16 @@ func TestPullPolicy(t *testing.T) {
 	}
 
 	for _, tt := range table {
-		p, err := Pod(source, "catalog", "opmImage", "utilImage", tt.image, serviceAccount("", "service-account"), nil, nil, int32(0), int32(0), int64(workloadUserID), v1alpha1.Legacy)
+		if tt.extractContent {
+			grpcPodConfig := &v1alpha1.GrpcPodConfig{
+				ExtractContent: &v1alpha1.ExtractContentConfig{
+					CacheDir:   "/tmp/cache",
+					CatalogDir: "/catalog",
+				},
```
camilamacedo86, line 985: Good catcher 👍 
I think we should have that now.

camilamacedo86 (approved):

PR #3614, opened 2025-07-24
NO-ISSUE: Enabled readonlyRootFilesystem by default

dusk125:
Description of the change:

Enforce readOnlyRootFilesystem: true for enhanced security (and provide brief justification for false exceptions)
Motivation for the change:

Reviewer Checklist

- [ ] Docs updated or added to /docs

### DIFF
--- a/pkg/controller/registry/reconciler/reconciler.go
+++ b/pkg/controller/registry/reconciler/reconciler.go
@@ -371,7 +371,7 @@ func addSecurityContext(pod *corev1.Pod, runAsUser int64) {
 			pod.Spec.Containers[i].SecurityContext = &corev1.SecurityContext{}
 		}
 		pod.Spec.Containers[i].SecurityContext.AllowPrivilegeEscalation = ptr.To(false)
-		pod.Spec.InitContainers[i].SecurityContext.ReadOnlyRootFilesystem = ptr.To(true)
+		pod.Spec.Containers[i].SecurityContext.ReadOnlyRootFilesystem = ptr.To(true)
 		pod.Spec.Containers[i].SecurityContext.Capabilities = &corev1.Capabilities{
 			Drop: []corev1.Capability{"ALL"},
 		}
--- a/pkg/controller/registry/reconciler/reconciler.go
+++ b/pkg/controller/registry/reconciler/reconciler.go
@@ -191,7 +191,7 @@ func Pod(source *operatorsv1alpha1.CatalogSource, name, opmImg, utilImage, img s
 						},
 					},
 					SecurityContext: &corev1.SecurityContext{
-						ReadOnlyRootFilesystem: ptr.To(false),
+						ReadOnlyRootFilesystem: ptr.To(true),
 					},
 					ImagePullPolicy:          image.InferImagePullPolicy(img),
 					TerminationMessagePolicy: corev1.TerminationMessageFallbackToLogsOnError,
@@ -361,6 +361,7 @@ func addSecurityContext(pod *corev1.Pod, runAsUser int64) {
 			pod.Spec.InitContainers[i].SecurityContext = &corev1.SecurityContext{}
 		}
 		pod.Spec.InitContainers[i].SecurityContext.AllowPrivilegeEscalation = ptr.To(false)
+		pod.Spec.InitContainers[i].SecurityContext.ReadOnlyRootFilesystem = ptr.To(true)
 		pod.Spec.InitContainers[i].SecurityContext.Capabilities = &corev1.Capabilities{
 			Drop: []corev1.Capability{"ALL"},
 		}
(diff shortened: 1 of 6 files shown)


tmshort:
@dusk125 - this seems to be failing all our tests...

pkg/controller/registry/reconciler/reconciler.go
```diff
@@ -191,7 +191,7 @@ func Pod(source *operatorsv1alpha1.CatalogSource, name, opmImg, utilImage, img s
 						},
 					},
 					SecurityContext: &corev1.SecurityContext{
-						ReadOnlyRootFilesystem: ptr.To(false),
+						ReadOnlyRootFilesystem: ptr.To(true),
```
grokspawn: As long as we support legacy access modes (and AFAIK we continue to do so), this needs to continue to be
```suggestion
						ReadOnlyRootFilesystem: ptr.To(false),
```

pkg/controller/registry/reconciler/reconciler.go
```diff
@@ -361,6 +361,7 @@ func addSecurityContext(pod *corev1.Pod, runAsUser int64) {
 			pod.Spec.InitContainers[i].SecurityContext = &corev1.SecurityContext{}
 		}
 		pod.Spec.InitContainers[i].SecurityContext.AllowPrivilegeEscalation = ptr.To(false)
+		pod.Spec.InitContainers[i].SecurityContext.ReadOnlyRootFilesystem = ptr.To(true)
```
grokspawn: This and 374 should be setting this to the same as `pod.Spec.Containers[0].SecurityContext.ReadOnlyRootFilesystem` because we support both legacy and restricted modes here.

dusk125: Thanks for the review, can you elaborate a little more please? It looks like the `addSecurityContext` function is only called when in restricted mode (lines 333-336); wouldn't it make sense to enable readOnly in restricted mode? Or does restricted imply something else?

grokspawn: We've historically coordinated RunAsNonRoot and RunAsUser (circa line 383) together.  Even for restricted mode, we defaulted to RunAsNonRoot (false) if we didn't include RunAsUser (!= 0). 
We would want to do the same for the init containers (but I was looking initially at an older commit, oops!) so here we'd need to either duplicate the logic
```suggestion
               if runAsUser > 0 {
  		  pod.Spec.InitContainers[i].SecurityContext.ReadOnlyRootFilesystem = ptr.To(true)
  	       }
```
or perform [these lines](https://github.com/operator-framework/operator-lifecycle-manager/pull/3614/files#diff-fffdeef1fc140a5dc5dc92dda323f567a6e46fc2ecbb0b91ba907acd02bde50dR380-R389) before the loop to set the init containers' attributes, and then do 
```suggestion
		pod.Spec.InitContainers[i].SecurityContext.ReadOnlyRootFilesystem = pod.Spec.SecurityContext.RunAsNonRoot
```

pkg/controller/registry/reconciler/reconciler.go
```diff
@@ -301,8 +304,12 @@ func Pod(source *operatorsv1alpha1.CatalogSource, name, opmImg, utilImage, img s
 				Args:                     extractArgs,
 				VolumeMounts:             []corev1.VolumeMount{utilitiesVolumeMount, contentVolumeMount},
 				TerminationMessagePolicy: corev1.TerminationMessageFallbackToLogsOnError,
+				SecurityContext: &corev1.SecurityContext{
+					ReadOnlyRootFilesystem: ptr.To(true),
+				},
 			})
 
+			pod.Spec.Containers[0].SecurityContext.ReadOnlyRootFilesystem = ptr.To(true)
```
jianzhangbjz, line 312: Just to confirm, if `grpcPodConfig.ExtractContent == nil`, then `pod.Spec.Containers[0].SecurityContext.ReadOnlyRootFilesystem` will be set to `false`. Is that the expected behavior? Thanks!

grokspawn: Correct.  If we're using extractContent, then we are pulling out the catalog+cache and running it in a catalogsource using an in-cluster opm version.  
If we're not using extractContent, then we're "just running an arbitrary pod", which is expected to speak GRPC on port 50051.

grokspawn:
go version has not changed as a result of this PR, but the go-verdiff assesses any delta between this PR and its destination branch (master).  

I'll override. 

/lgtm

pkg/controller/registry/reconciler/reconciler.go
jianzhangbjz: Thanks for the confirmation! I have one more question—if we don’t use extractContent, the command running in the Pod is `/bin/opm serve /configs --cache-dir=/tmp/cache`, which requires `readOnlyRootFilesystem: false`. However, I believe mounting `/tmp` as a volume should also work if we set `readOnlyRootFilesystem: true`. So, my question is: why not set `readOnlyRootFilesystem: true` when extractContent is not used? Or am I missing something? Thanks!

PR #3630, opened 2025-08-26
Test NetworkPolicy support in e2es

oceanc80:
Now that we've added support for NetworkPolicies, ensure we test that they can be installed correctly.

**Architectural changes:**

**Testing remarks:**

**Reviewer Checklist**


trgeiger:
Latest revision of the PR adds build files for creating the test bundle image and the test catalog/index image. These are pushed to new tags to keep the old ones if ever needed. I also added a script for rebuilding those images and an addition to the documentation for using the script.

DEVELOPMENT.md
```diff
@@ -46,6 +46,19 @@ To run a specific e2e test locally:
 $ make e2e-local TEST=TestCreateInstallPlanManualApproval
 ```
 
+##### Updating test images
```
anik120, line 49: nit: should we also give the disclaimer that you need to be logged into quay from an account that has write access to the olmtest namespace?

anik120 (approved):
/lgtm

scripts/build_e2e_test_images.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
```
anik120, line 1: Actually let's talk about this for a bit. Do we need a new script, or can we add this cleanly to the existing build_test_images.sh script?

anik120: Or maybe what's actually required is proper naming of the two scripts. Right now it's confusing to me what does what if I just read the names. I have to read through the scripts to understand the difference (unnecessary cognitive overload in the workflow). 

Maybe 

`build-update-versioned-index.sh`

and 

`create-single-bundle-index.sh`


?

DEVELOPMENT.md
```diff
@@ -46,6 +46,19 @@ To run a specific e2e test locally:
...
+
+Sometimes you will need to update the index or bundle images used in the unit or e2e tests. To update those images, you can use the following 2 scripts in the ./scripts/ directory:
+
+```sh
+./scripts/build_test_images.sh
+./scripts/build_e2e_test_images.sh
+```
```
anik120: And if we are documenting these scripts, might as well write a short summary of what each script does.

scripts/build_e2e_test_images.sh
trgeiger: I think you're right about combining them, actually. All the images are for different end-to-end tests, so it makes sense to combine them and maybe rename the single script to the name I gave the new one, build_e2e_test_images.sh, to be more descriptive.

DEVELOPMENT.md
trgeiger: Done

scripts/build_e2e_test_images.sh
trgeiger: I combined them. Let me know if you think that works, or if you think it would be better to separate the 2 index image builds.

DEVELOPMENT.md
trgeiger: Done

test/images/single-bundle-image/manifests/networkpolicy.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: busybox-networkpolicy
+spec:
+  podSelector: {}
+  policyTypes:
+    - Ingress
```
perdasilva: nit: can we please add newlines at the end of the files?

trgeiger: done

trgeiger: and finally changed my editor's settings to always add one on save ;)

anik120 (approved):
/lgtm

PR #3640, opened 2025-09-02
rorfs for cacheless lacks write access to /tmp/ to generate cache

grokspawn:
when we are using `extractContent`, but w/o `extractContent.CacheDir`, ensure that we have write permissions to `/tmp/` by creating an emptydir volume there.

When there is no provided `CacheDir`, our recent changes to set `SecurityContext{ReadOnlyFilesystem: ptr.To(true)}` will not allow writing to `/tmp/`.  However, `opm` will need to generate a cache and cannot be told another directory to use.

**Architectural changes:**

**Testing remarks:**
#3614 passed CI with no errors, both [unit](https://github.com/operator-framework/operator-lifecycle-manager/actions/runs/16911030770/job/47929994554) which covered ExtractContent functionality including w/o CacheDir. 

This is because while the unit tests validated the shape of the resulting pods but not the pods' viability in a cluster (i.e. could the pods walk the initcontainer chain successfully and bring the GRPC service up). 

ginkgo test "gRPC address catalog source" _does_ evaluate catsrc pod viability using extractContent, but never as cacheless.

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/registry/reconciler/reconciler.go
+++ b/pkg/controller/registry/reconciler/reconciler.go
@@ -322,13 +322,13 @@ func Pod(source *operatorsv1alpha1.CatalogSource, name, opmImg, utilImage, img s
 			} else {
 				// opm serve does not allow us to specify an empty cache directory, which means that it will
 				// only create new caches in /tmp/, so we need to provide adequate write access there
-				const tmpdirPath = "/tmp/"
+				const tmpdirName = "tmpdir"
 				tmpdirVolumeMount := corev1.VolumeMount{
-					Name:      "tmpdir",
-					MountPath: tmpdirPath,
+					Name:      tmpdirName,
+					MountPath: "/tmp/",
 				}
 				pod.Spec.Volumes = append(pod.Spec.Volumes, corev1.Volume{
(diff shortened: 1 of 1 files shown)


pkg/controller/registry/reconciler/reconciler.go
```diff
@@ -319,6 +319,22 @@ func Pod(source *operatorsv1alpha1.CatalogSource, name, opmImg, utilImage, img s
 			}
 			if grpcPodConfig.ExtractContent.CacheDir != "" {
 				containerArgs = append(containerArgs, "--cache-dir="+filepath.Join(catalogPath, "cache"))
+			} else {
+				// opm serve does not allow us to specify an empty cache directory, which means that it will
+				// only create new caches in /tmp/, so we need to provide adequate write access there
+				const tmpdirPath = "/tmp/"
```
joelanford: Nit: const for `tmpdirPath` seems unnecessary since we don't use it more than once. Although a const for the volume mount name (`"tmpdir"`) would make sense since we use it in both the volume and volumeMount.

perdasilva:
lol whoops

PR #3682, opened 2025-10-23
Fix TOCTOU race condition in ensureInstallPlan

jianzhangbjz:
**Description of the change:**
Fixes a race condition that causes OLM to create duplicate InstallPlans for the same subscription when multiple worker threads reconcile a namespace concurrently.

  #### Problem
  When two worker threads process the namespace queue simultaneously:
  1. Worker 1 calls `listInstallPlans()` and finds no existing InstallPlan
  2. Worker 2 calls `listInstallPlans()` and finds no existing InstallPlan
  3. Worker 1 acquires the mutex lock and creates an InstallPlan
  4. Worker 1 releases the lock
  5. Worker 2 acquires the mutex lock, checks its stale list (from step 2), and creates a duplicate InstallPlan

  This is a classic Time-of-Check-to-Time-of-Use (TOCTOU) vulnerability.

**Motivation for the change:**

**Architectural changes:**
Move the `listInstallPlans()` call **inside** the mutex-protected critical section to ensure atomic check-and-create behavior.

**Testing remarks:**
  Added `TestEnsureInstallPlanConcurrency` which:
  - Launches 10 concurrent goroutines calling `ensureInstallPlan()`
  - Verifies only one InstallPlan is created
  - Verifies all goroutines receive the same InstallPlan reference

**Reviewer Checklist**
- [ ] Implementation matches the proposed design, or proposal is updated to match implementation
- [x] Sufficient unit test coverage
- [ ] Sufficient end-to-end test coverage
- [ ] Bug fixes are accompanied by regression test(s)
- [ ] e2e tests and flake fixes are accompanied evidence of flake testing, e.g. executing the test 100(0) times
- [ ] tech debt/todo is accompanied by issue link(s) in comments in the surrounding code
- [ ] Tests are comprehensible, e.g. Ginkgo DSL is being used appropriately
- [ ] Docs updated or added to `/doc`
- [ ] Commit messages sensible and descriptive
- [ ] Tests marked as `[FLAKE]` are truly flaky and have an issue
- [ ] Code is properly formatted

Assisted-by: Claude Code

PR #3684, opened 2025-10-23
Fix race condition in duplicate InstallPlan prevention

joelanford:

Move the listInstallPlans() call to after the muInstallPlan lock is acquired in ensureInstallPlan(). Previously, the check for existing install plans happened before acquiring the lock, which meant that worker 2 could check for install plans before worker 1 had a chance to create one, even with the lock in place.

This fixes the race condition by ensuring that the sequence is:
1. Worker 1 acquires lock
2. Worker 1 checks for existing install plans
3. Worker 1 creates new install plan (if needed)
4. Worker 1 releases lock
5. Worker 2 acquires lock
6. Worker 2 checks for existing install plans (now sees worker 1's)
7. Worker 2 reuses existing install plan
8. Worker 2 releases lock

Without this fix, both workers could see no existing install plans before either acquired the lock, leading to duplicate install plans being created for the same subscription.

**Architectural changes:**

None

**Testing remarks:**

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -1654,12 +1654,6 @@ func (o *Operator) ensureInstallPlan(logger *logrus.Entry, namespace string, gen
 		return nil, nil
 	}
 
-	// Check if any existing installplans are creating the same resources
-	installPlans, err := o.listInstallPlans(namespace)
-	if err != nil {
-		return nil, err
-	}
-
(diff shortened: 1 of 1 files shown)


perdasilva:
Just saw #3682 solves this in the same way (and adds a unit test). Let's get that one down instead?

joelanford:
Duplicate of #3682 , but that one has a really nice unit test! Closing this one.

PR #3691, opened 2025-10-29
gitignore temp files in .claude folder

anik120:

Ignore temp files created in the `.claude/` folder

**Architectural changes:**

**Testing remarks:**

**Reviewer Checklist**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -455,3 +455,11 @@ apiserver.key
 !vendor/**
 test/e2e/.kube
 dist/
+
+# AI temp/local files
+.claude/settings.local.json
+.claude/history/
+.claude/cache/
+.claude/logs/
+.claude/.session*
+.claude/*.log


.gitignore
```diff
@@ -455,3 +455,7 @@ apiserver.key
 !vendor/**
 test/e2e/.kube
 dist/
+
+# AI Tools
+
+.claude/
```
tmshort: Actually, you should ask Claude... ;)

Take a look at:
https://github.com/operator-framework/operator-controller/pull/2290/files#diff-bc37d034bad564583790a46f19d807abfe519c5671395fd494d8cce506c42947

anik120: Okay, NICE!! 

So then this is an interesting conversation now. The reason I wanted to introduce this gitignore is because I want to create sub-agents at the project level for myself. 

When you create a sub-agent at the project level, it's created at `.claude/agents/*.md`. 

However, there's a use case similar to your command sharing use case, where sub-agents can be shared around (so included in git distribution). 


I guess the best approach here then is to mimic your changes (ie ignore temporary files), and then explore creating agents we don't want to share at the user level (gets created in the `~/.claude/agents/` folder, ie in the root level).

PR #3692, opened 2025-10-30
Fix: OLM should not report Progressing=True during pod disruption from cluster upgrades

jianzhangbjz:

### 1. Pod Disruption Detection

  Added `isAPIServiceBackendDisrupted()` function that checks if APIService unavailability is due to expected pod disruption:

  **Disruption Signals Detected:**
  - Pod has `DeletionTimestamp` set (terminating during node drain)
  - Pod in `Pending` phase (being scheduled/created after eviction)
  - Container in `ContainerCreating` or `PodInitializing` state
  - Deployment has unavailable replicas with pods restarting

  ### 2. RetryableError Type

  Introduced `RetryableError` to distinguish:
  - **RetryableError**: Expected transient failures (pod disruption) → Don't change CSV phase
  - **Normal Error**: Real failures → Mark as Failed as before

  ### 3. Updated Logic

  **`areAPIServicesAvailable()`:**
  - When APIService unavailable, check for pod disruption
  - If disrupted: Return `RetryableError`
  - If not disrupted: Return normal error (existing behavior)

  **`updateInstallStatus()`:**
  - Check if `apiServiceErr` is retryable
  - If retryable: Requeue without changing CSV phase (NO Progressing=True)
  - If not retryable: Mark as Failed (existing behavior)

  ## Impact

  ### Behavior Change

  | Scenario | Before | After |
  |----------|--------|-------|
  | APIService unavailable, pod disrupted during upgrade | CSV→Failed, Progressing=True ❌ | CSV stays current phase, Progressing=False ✅ |
  | APIService unavailable, real failure | CSV→Failed, Progressing=True | CSV→Failed, Progressing=True (unchanged) |
  | APIService available | CSV→Succeeded | CSV→Succeeded (unchanged) |

# Fix OLM Progressing Condition Contract Violation During Cluster Upgrades

  ## Problem

  The OLM packageserver ClusterOperator was violating the documented Progressing condition contract during cluster upgrades.

  ### Test Failure
```console
  [Monitor:legacy-cvo-invariants][bz-OLM] clusteroperator/operator-lifecycle-manager-packageserver
  should stay Progressing=False while MCO is Progressing=True

  Oct 29 06:32:38.018 W clusteroperator/operator-lifecycle-manager-packageserver
    condition/Progressing status/True Working toward 0.0.1-snapshot
```

  ### Root Cause

  During MCO-driven cluster upgrades:
  1. Node reboots (planned maintenance)
  2. PackageServer pod gets evicted and restarts (expected Kubernetes behavior)
  3. APIService temporarily unavailable during pod startup (~14 seconds)
  4. OLM detects unavailability → marks CSV as `Failed` → ClusterOperator reports `Progressing=True`
  5. System self-heals after pod restarts

  **This violates the contract** because:
  - No new code being rolled out (same version)
  - No config changes being propagated
  - Just reconciling to previously known state after expected disruption

**Architectural changes:**

**Testing remarks:**

**Reviewer Checklist**
- [x] Bug fixes are accompanied by regression test(s)

Assisted-by: Claude Code

### DIFF
--- a/pkg/controller/operators/olm/apiservices.go
+++ b/pkg/controller/operators/olm/apiservices.go
@@ -3,10 +3,12 @@ package olm
 import (
 	"context"
 	"fmt"
+	"time"
 
 	hashutil "github.com/operator-framework/operator-lifecycle-manager/pkg/lib/kubernetes/pkg/util/hash"
 	log "github.com/sirupsen/logrus"
 	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
@@ -22,6 +24,11 @@ import (
 const (
 	// Name of packageserver API service
 	PackageserverName = "v1.packages.operators.coreos.com"
+
+	// maxDisruptionDuration is the maximum duration we consider pod disruption as "expected"
+	// (e.g., during cluster upgrade). Beyond this time, we treat the unavailability as a real failure.
+	// This prevents indefinitely waiting for pods that will never recover.
+	maxDisruptionDuration = 5 * time.Minute
 )
 
 // apiServiceResourceErrorActionable returns true if OLM can do something about any one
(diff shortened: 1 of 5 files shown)


camilamacedo86 (changes requested):
Hey @jianzhangbjz — thanks a lot for this PR! 🎉  
I agree with the direction: OLM shouldn’t flip `Progressing=True` just because nodes are draining/rebooting or the API backend has a short wobble. I just added a comment about how identify the scenario see; https://github.com/operator-framework/operator-lifecycle-manager/pull/3692/files#r2476546082

jianzhangbjz:
Please read https://github.com/openshift/api/blob/master/config/v1/types_cluster_operator.go#L159-L164

pkg/controller/operators/olm/apiservices.go
```diff
@@ -168,6 +169,87 @@ func (a *Operator) checkAPIServiceResources(csv *v1alpha1.ClusterServiceVersion,
...
+							return true
+						}
+					}
+				}
+			}
+		}
+	}
+
+	return false
+}
```
camilamacedo86: **OCP rule ( Please, let me know if I misunderstood it )** 
- `Progressing=True` is for when the operator is actually rolling out new code or propagating config (moving between steady states). 
- Reconciling back to a previously known state after node disruption is **not** “progress.” In those moments, reflect the dip with `Available`, not `Progressing`.

#### So, I think the current logic can bring **false positives** in the following scenarios:

- Pods are **Pending** due to **quota / PDB / limits** (not a drain).
- Pods show real failures like **`ImagePullBackOff` / `CrashLoopBackOff`**.
- The **Deployment is rolling out a new template** (that *is* real progress).
- A single pod is **Pending for a long time** while nodes are healthy.
- **`ContainerCreating`** lasts too long (no time cap), masking a real issue.

**Therefore, we might be able to improve the logic for something like** 

It is **transient node disruption** **only if _all_** of the following are true:

- All unready pods are unready **only for drain-like reasons**:
  - `Terminating`, `ContainerCreating`, `PodInitializing`, or
  - `Unschedulable` with drain-ish messages like “unschedulable”, “node(s) had taint … NoSchedule”, “no nodes available”.
- **And** no pod shows hard failures such as `ImagePullBackOff`, `ErrImagePull`, `CrashLoopBackOff`, `RunContainerError`, or repeated non-zero exits.
- **And** the Deployment is **not** rolling out a new template  
  (`status.observedGeneration == metadata.generation`).
- **And** the nodes hosting those pods are **NotReady** (recent transition) or **Unschedulable** (cordoned), or a meaningful share of nodes are affected.
- **And** the condition has lasted **< 30 minutes** (grace window).

**Then** → requeue with a `RetryableError` (with backoff).  
**Else** → follow the normal failure path.

Then: the code could be something like: **Example**

```go
func isExpectedNodeDisruption(ctx context.Context, c client.Client, dep appsv1.Deployment) (bool, string, error) {
    // 1) Not rolling out a new template
    if dep.Generation != dep.Status.ObservedGeneration {
        return false, "deployment rollout in progress", nil
    }

    // 2) Pods for the deployment
    selector, err := metav1.LabelSelectorAsSelector(dep.Spec.Selector)
    if err != nil {
        return false, "", err
    }
    var pods corev1.PodList
    if err := c.List(ctx, &pods, client.InNamespace(dep.Namespace),
        client.MatchingLabelsSelector{Selector: selector}); err != nil {
        return false, "", err
    }

    // 3) Hard failures → not expected disruption
    for _, p := range pods.Items {
        for _, cs := range append(p.Status.InitContainerStatuses, p.Status.ContainerStatuses...) {
            if cs.State.Waiting != nil {
                switch cs.State.Waiting.Reason {
                case "ImagePullBackOff", "ErrImagePull", "CrashLoopBackOff", "RunContainerError":
                    return false, "hard failure: " + cs.State.Waiting.Reason, nil
                }
            }
            if cs.RestartCount >= 5 {
                if t := cs.LastTerminationState.Terminated; t != nil && t.ExitCode != 0 {
                    return false, "repeated non-zero exits", nil
                }
            }
        }
    }

    // 4) Pods unready only for drain-like reasons
    drainLike := false
    for _, p := range pods.Items {
        if isPodReady(p) { continue }
        drainLike = true

        if p.DeletionTimestamp != nil { continue } // terminating (common in drains)
        if hasWaiting(&p, "ContainerCreating") || hasWaiting(&p, "PodInitializing") { continue }

        if cond := podCond(p, corev1.PodScheduled); cond != nil &&
           cond.Status == corev1.ConditionFalse &&
           cond.Reason == corev1.PodReasonUnschedulable &&
           (strings.Contains(cond.Message, "unschedulable") ||
            strings.Contains(cond.Message, "taint") ||
            strings.Contains(cond.Message, "no nodes available")) {
            continue
        }
        return false, "pods unready for non drain-like reasons", nil
    }
    if !drainLike {
        return false, "no drain-like unavailability found", nil
    }

    // 5) Node disruption signal
    affected, why, err := nodeDisruptionSignal(ctx, c, pods.Items)
    if err != nil { return false, "", err }
    if !affected { return false, "no node disruption signal: " + why, nil }

    // 6) Time-based grace window
    first := earliestUnavailability(pods.Items)
    if first.IsZero() { first = metav1.NewTime(time.Now().Add(-5 * time.Minute)) }
    if time.Since(first.Time) > 30*time.Minute {
        return false, "exceeded 30min grace window", nil
    }

    return true, "node disruption (cordon/NotReady) detected", nil
}

Comment (camilamacedo86) [inline pkg/controller/operators/olm/apiservices.go:338]: So, I think the current logic can bring **false positives** in the following scenarios:

- Pods are **Pending** due to **quota / PDB / limits** (not a drain).
- Pods show real failures like **`ImagePullBackOff` / `CrashLoopBackOff`**.
- The **Deployment is rolling out a new template** (that *is* real progress).
- A single pod is **Pending for a long time** while nodes are healthy.
- **`ContainerCreating`** lasts too long (no time cap), masking a real issue.

Can you please check?
Or I am wrong here?

**Review by camilamacedo86 — CHANGES_REQUESTED** [review]

File: pkg/controller/operators/olm/apiservices.go (hunk 1 above)
Comment (camilamacedo86) [reply pkg/controller/operators/olm/apiservices.go:338]: @jianzhangbjz 

Thank you for sharing: https://github.com/openshift/api/blob/master/config/v1/types_cluster_operator.go#L159-L164

So,  after taking another look, I noticed a few pitfalls in the original proposal:  
- `isAPIServiceBackendDisrupted` treats every `PodPending` as “expected,” so real outages (like image pull errors or unschedulable pods) never flip the CSV to `Failed`.  
- The requeue path never updates the CSV status, which can leave the operator looking healthy forever if pods never recover.  
- The disruption check walks every deployment in the CSV, so an unrelated rollout could mask a broken APIService.  
- The new test file mostly mirrors production logic, meaning regressions could still pass undetected.  

Assisted by AI to help refine coverage and clarity, I’ve pushed a small follow-up commit.  
The goal is to make the review smoother and the logic easier to reason about.  

**This update:**  
- Narrows disruption handling to the APIService’s own deployment, trusting only drain-like pod states and enforcing a clear time limit.  
- Adds a short `RequeueAfter` backoff to avoid hot-looping while pods recover.  
- Improves the test coverage with realistic transient and failure scenarios.  

See: https://github.com/operator-framework/operator-lifecycle-manager/pull/3693/commits/af924b798edc2f4de52962dcea3e6792d794e0b4  - from https://github.com/operator-framework/operator-lifecycle-manager/pull/3693 to help out clarify the suggestions and ensure that all passed the tests after those changes.
 
I really think we need a better coverage here. 
And improve the above points. 

Let me know if you’d like me to adjust or open a separate PR comment!

File: pkg/controller/operators/olm/operator.go [hunk 2]
```diff
@@ -2633,7 +2634,21 @@ func (a *Operator) updateInstallStatus(csv *v1alpha1.ClusterServiceVersion, inst
...
 	if apiServiceErr != nil {
+		if olmerrors.IsRetryable(apiServiceErr) {
+			// This is an expected transient failure (e.g., pod disruption during upgrade)
+			// Don't change the CSV phase to Failed or trigger Progressing=True
+			// Just requeue and let it retry
+			a.logger.Infof("APIService temporarily unavailable due to pod disruption, requeueing without changing phase: %v", apiServiceErr)
+			if err := a.csvQueueSet.Requeue(csv.GetNamespace(), csv.GetName()); err != nil {
+				a.logger.Warn(err.Error())
+			}
+			return apiServiceErr
```
camilamacedo86, line 2654:
```suggestion
		if olmerrors.IsRetryable(apiServiceErr) {
			// The backing pods are in the middle of a drain/restart. Requeue after a short delay
			// so we keep checking without hammering the workqueue.
			a.logger.Infof("APIService temporarily unavailable due to pod disruption, requeueing without changing phase: %v", apiServiceErr)
			if err := a.csvQueueSet.RequeueAfter(csv.GetNamespace(), csv.GetName(), retryableAPIServiceRequeueDelay); err != nil {
				a.logger.Warn(err.Error())
			}
			return apiServiceErr
```

camilamacedo86: I think retry after will be better in this scenario
I mean, ensure that we have a small delay on it to give time

pkg/controller/operators/olm/apiservices.go
```diff
@@ -168,6 +169,87 @@ func (a *Operator) checkAPIServiceResources(csv *v1alpha1.ClusterServiceVersion,
 	return utilerrors.NewAggregate(errs)
 }
 
+// isAPIServiceBackendDisrupted checks if the APIService is unavailable due to expected pod disruption
+// (e.g., during node reboot or cluster upgrade) rather than an actual failure.
+// According to the Progressing condition contract, operators should not report Progressing=True
+// only because pods are adjusting to new nodes or rebooting during cluster upgrade.
+func (a *Operator) isAPIServiceBackendDisrupted(csv *v1alpha1.ClusterServiceVersion, apiServiceName string) bool {
```
camilamacedo86, line 182: @jianzhangbjz Here’s what I’d tweak next:

- **Add a short grace window and a retry delay.** Let’s treat the “pods are repainting” window as expected for only a few minutes, then requeue with a brief pause instead of looping nonstop. That keeps us in line with the Progressing contract/requirement without hiding real outages forever.

- **Swap the broad PodPending check for a targeted set of states** (ContainerCreating, PodInitializing, NodeShutdown, etc.). That way we still handle planned reboots **but we stop calling genuine failures “expected.”** ( solve concerns about false negatives, see that we are checking out those scenarios in the proposed changes for the tests as well )

- Watch only the deployment that actually backs the APIService. That keeps other rollouts (metrics, webhook) from masking a broken APIService.

- **Split the helper into small functions**—one to grab pods, one to judge their state, one to look at recent rollout progress. The logic stays the same but the code gets a lot easier to follow, maintain and etc. Low complexity

Ref -> see the whole suggestions proposed here => : https://github.com/operator-framework/operator-lifecycle-manager/commit/af924b798edc2f4de52962dcea3e6792d794e0b4#diff-1c45679686ec7da9a4e3c02d009d84f465130fec2278b9512d27d53e23224f3a

jianzhangbjz:
Thanks! Updated them.

camilamacedo86 (approved):
@jianzhangbjz I think you addressed all concerns.
Thank you for looking on that it is LGTM for me

It would be nice either get a second reviewer from @perdasilva or @tmshort here before we move forward.

Thank you a lot for the nice work 🎉

pkg/controller/operators/olm/apiservices.go
camilamacedo86: Solved

pkg/controller/operators/olm/apiservices.go
camilamacedo86: It seems solved 
Thank you

Issue #3698, opened 2025-11-05
Unable to resolve addresses for GRPC by proxy resolver


mgencur:
## Bug Report

The recent [bump of dependencies](https://github.com/operator-framework/operator-lifecycle-manager/commit/99f5f8bb9fed7b817c2fda66ff2a3b7023851817#diff-27f86220808fdbef9aa6b6da0f975e70a2662f478a12d1bbfd713b63f5a15c38R166) changed the way a GRPC connection is created from `grpc.Dial(address, dialOptions...)` to  `grpc.NewClient(address, dialOptions...)`. 
This caused a problem in scenarios where the target address of Catalog Source should be resolved by a custom proxy resolver like in the case of [Hypershift](https://github.com/openshift/hypershift). The catalog-operator runs on the management cluster and individual catalog sources on a guest cluster. The new behavior tries to resolve the address immediately (resolver scheme is "dns" instead of "passthrough") instead of delegating this to the proxy resolver. See [docs](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/vendor/google.golang.org/grpc/clientconn.go#L232-L238)

**What did you do?**

* Start a hosted cluster using: hypershift create cluster aws ... --olm-catalog-placement=Guest

**What did you expect to see?**
Catalog Sources available in the guest cluster and package manifests available as well.

**What did you see instead? Under which circumstances?**

Catalog Source in guest cluster:
```
- apiVersion: operators.coreos.com/v1alpha1
  kind: CatalogSource
  metadata:
    annotations:
      target.workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
    creationTimestamp: "2025-11-04T08:13:07Z"
    generation: 1
    labels:
      hypershift.openshift.io/managed: "true"
    name: redhat-operators
    namespace: openshift-marketplace
    resourceVersion: "35089"
    uid: 9795a184-7a8e-4198-bb6a-057038116b7b
  spec:
    displayName: Red Hat Operators
    grpcPodConfig:
      securityContextConfig: restricted
    icon:
      base64data: ""
      mediatype: ""
    image: registry.redhat.io/redhat/redhat-operator-index:v4.20
    priority: -100
    publisher: Red Hat
    sourceType: grpc
    updateStrategy:
      registryPoll:
        interval: 10m
  status:
    connectionState:
      address: redhat-operators.openshift-marketplace.svc:50051
      lastConnect: "2025-11-04T09:34:33Z"
[...]

PR #3699, opened 2025-11-05
Allow overriding resolver scheme

mgencur:
Fixes https://github.com/operator-framework/operator-lifecycle-manager/issues/3698

Introduce RESOLVER_SCHEME env variable which can be used to set the resolver scheme explicitly.

The default behaviour of grpc.NewClient does not work in all cases for OLM. The [GRPC fixes](https://github.com/grpc/grpc-go/pull/7881) in this area introduced a new resolver that handles proxy configuration and delegates to child resolvers for target and proxy address resolution. But it is only used when there's no custom dialer. [OLM calls](https://github.com/operator-framework/operator-lifecycle-manager/blob/615d70d67bd537273e4f561ce1ca5c7f633d13fd/pkg/controller/registry/grpc/source.go#L157) WithContextDialer when proxy is used. In that case, GRPC uses the default resolver which uses the default resolver scheme ("dns"), which resolves the address on client side.

There needs to be a way to specify the original "passthrough" resolver scheme which will delegate resolving to the proxy. When the dialer is used it is recommended to use the passthrough resolver to change the resolver behaviour:

> 
> // Note that gRPC by default performs name resolution on the target passed to
> // NewClient. To bypass name resolution and cause the target string to be
> // passed directly to the dialer here instead, use the "passthrough" resolver
> // by specifying it in the target string, e.g. "passthrough:target".

From [GRPC dialoptions docs](https://github.com/grpc/grpc-go/blob/7472d578b15f718cbe8ca0f5f5a3713093c47b03/dialoptions.go#L469).

**Testing remarks:**

Tested the solution with change in Hypershift: [link to commit](https://github.com/mgencur/hypershift/commit/4c72e577a0399ec926548ff80855026f5ff79ec7)
Details: TBD

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/registry/grpc/source.go
+++ b/pkg/controller/registry/grpc/source.go
@@ -163,7 +163,12 @@ func grpcConnection(address string) (*grpc.ClientConn, error) {
 		}))
 	}
 
-	return grpc.NewClient(address, dialOptions...)
+	target := address
+	if scheme := os.Getenv("RESOLVER_SCHEME"); scheme != "" {
+		target = scheme + ":///" + address
+	}
+
+	return grpc.NewClient(target, dialOptions...)
 }
 
 func (s *SourceStore) Add(key registry.CatalogKey, address string) (*SourceConn, error) {


tmshort:
See #3700

mgencur:
Closing in favor of https://github.com/operator-framework/operator-lifecycle-manager/pull/3700

PR #3700, opened 2025-11-05
OCPBUGS-64574: fix(grpc): Use passthrough resolver when proxy is detected

tmshort:
Fix CatalogSource reporting TRANSIENT_FAILURE in Hypershift guest clusters by automatically using the "passthrough" resolver scheme when a proxy is detected.

Root Cause:
The migration from grpc.Dial() to grpc.NewClient() introduced a resolver scheme issue. When grpc.NewClient() is used with WithContextDialer (for proxy support), gRPC defaults to the "dns" resolver which tries to resolve addresses client-side. In Hypershift, the catalog operator runs in the management cluster and connects via SOCKS5 proxy to catalog pods in the guest cluster. Service addresses like "service.namespace.svc:50051" only exist in the guest cluster's DNS and cannot be resolved from the management cluster, causing connections to fail with TRANSIENT_FAILURE.

Solution:
Automatically detect when a proxy is being used (proxyURL != nil) and prepend "passthrough:///" to the target address. The passthrough resolver bypasses client-side DNS resolution and delegates it to the custom dialer (proxy), which resolves addresses in the guest cluster where they exist.

This solution:
- Requires no environment variables or configuration
- Automatically activates only when proxy is used
- Follows gRPC best practices per documentation
- Simpler than alternative env var approaches (e.g., PR #3699)

Fixes: OCPBUGS-64574
Related: OCPBUGS-64631, #3698, #3699

🤖 Generated with [Claude Code](https://claude.com/claude-code) via /jira:solve OCPBUGS-64574

**Description of the change:**

**Motivation for the change:**

**Architectural changes:**

**Testing remarks:**

**Reviewer Checklist**
- [ ] Implementation matches the proposed design, or proposal is updated to match implementation
- [ ] Sufficient unit test coverage
- [ ] Sufficient end-to-end test coverage
- [ ] Bug fixes are accompanied by regression test(s)
- [ ] e2e tests and flake fixes are accompanied evidence of flake testing, e.g. executing the test 100(0) times
- [ ] tech debt/todo is accompanied by issue link(s) in comments in the surrounding code
- [ ] Tests are comprehensible, e.g. Ginkgo DSL is being used appropriately
- [ ] Docs updated or added to `/doc`
- [ ] Commit messages sensible and descriptive
- [ ] Tests marked as `[FLAKE]` are truly flaky and have an issue
- [ ] Code is properly formatted

PR #3702, opened 2025-11-06
🌱 Increase test coverage to cover proxy-backed grpc catalogs

camilamacedo86:
Increase the test coverage. 
Motivation here is the error faced and fixed in https://github.com/operator-framework/operator-lifecycle-manager/pull/3700


tmshort:
might need to rebase?

PR #3707, opened 2025-11-11
Upgrade dependencies; ocp/api*, kube-openapi and indirects

camilamacedo86:
### Dependency Updates

| Package | From | To |
|---------|------|-----|
| `github.com/openshift/api` | `v0.0.0-20221021112143` | `v0.0.0-20251111193948` |
| `github.com/openshift/client-go` | `v0.0.0-20220525160904` | `v0.0.0-20251015124057` |
| `k8s.io/kube-openapi` | (various) | (updated) |
| `golang.org/x/net` | (indirect) | (updated) |
| `golang.org/x/term` | (indirect) | (updated) |
| `github.com/google/pprof` | (indirect) | (updated) |

### Breaking Change

The upgraded `github.com/openshift/api` removed individual CRD YAML files from vendor. 
- `scripts/generate_openshift_crds.sh` creates static YAML files
- Tests load them from `testdata/crds/` (no runtime generation)
- Integrated into `make gen-all`

### DIFF
--- a/pkg/controller/operators/openshift/suite_test.go
+++ b/pkg/controller/operators/openshift/suite_test.go
@@ -46,15 +46,17 @@ const (
 var _ = BeforeSuite(func() {
 	logf.SetLogger(zap.New(zap.WriteTo(GinkgoWriter), zap.UseDevMode(true)))
 
-	base := filepath.Join("..", "..", "..", "..", "vendor", "github.com", "openshift", "api", "config", "v1")
+	// Note: OpenShift CRDs are loaded from testdata instead of vendor.
+	// Breaking change: The github.com/openshift/api package (v0.0.0-20251111193948+)
+	// no longer ships individual CRD YAML files in vendor (they were removed in favor
+	// of a consolidated manifest). We generate minimal CRDs via scripts/generate_openshift_crds.sh
+	// and load them from testdata to keep tests self-contained and work in envtest/kind environments.
 	testEnv = &envtest.Environment{
-		ErrorIfCRDPathMissing: true,
 		CRDs: []*apiextensionsv1.CustomResourceDefinition{
 			crds.ClusterServiceVersion(),
 		},
 		CRDDirectoryPaths: []string{
-			filepath.Join(base, "0000_00_cluster-version-operator_01_clusteroperator.crd.yaml"),
-			filepath.Join(base, "0000_00_cluster-version-operator_01_clusterversion.crd.yaml"),
+			filepath.Join("testdata", "crds"),
 		},
 	}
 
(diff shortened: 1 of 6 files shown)


pkg/controller/operators/openshift/suite_test.go
```diff
@@ -111,3 +116,149 @@ var _ = AfterSuite(func() {
...
+						},
+					},
+					Subresources: &apiextensionsv1.CustomResourceSubresources{
+						Status: &apiextensionsv1.CustomResourceSubresourceStatus{},
+					},
+				},
+			},
+		},
+	}
+}
```
camilamacedo86: @tmshort I need to address a breaking change in the recent ocp/api dependency upgrade.

**Previous Behavior:**
Our tests were loading ClusterOperator and ClusterVersion CRDs from YAML files in the vendor directory (`github.com/openshift/api/config/v1/0000_00_cluster-version-operator_01_*.crd.yaml`) and applying them directly to the envtest environment.

**Breaking Change:**
The upgraded `github.com/openshift/api` package (v0.0.0-20251111193948+) no longer provides individual CRD YAML files - they were removed in favor of a consolidated manifest.

**Solution:**
We're now generating the ClusterOperator and ClusterVersion CRDs programmatically in the test suite. This approach allows us still doing the test for OCP but still running it with kind/envtest. 

**Note:** The CRD strict schema code in `newClusterOperatorCRD()` and `newClusterVersionCRD()` was generated by AI to match the required OpenShift API structure.

oceanc80: Since I'd imagine the schemas a pretty static, does it really make sense to generate them at runtime? It feels like this could increase the time it takes for our tests to run. Could we generate them once instead of every time the tests run? We probably also need to modify the `manifests` [make target](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/Makefile#L324-L326) and its [associated script](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/scripts/copy_crds.sh) anyways if the `github.com/openshift/api` package no longer provides the CRDs. So we could replace `copy_crds.sh` with this CRD generation and not impact testing times.

camilamacedo86: @oceanc80 Great catch, thank you!

You're absolutely right. I've updated it to use your approach - a shell script that generates the CRDs once:

- `scripts/generate_openshift_crds.sh` creates static YAML files
- Tests load them from `testdata/crds/` (no runtime generation)
- Integrated into `make gen-all`

Zero test overhead and much clearer. Thanks for the better solution!

Closing this one as done ! Resolved

PR #3737, opened 2026-01-13
🌱 Add e2e test for operator workload persistence after catalog deletion

camilamacedo86:
This PR adds a comprehensive e2e test to verify that installed operators continue functioning normally after their source catalog is deleted. This ensures catalogs are only needed for installation and upgrades, not for runtime operation.

**Complete Test Flow**

  **Phase 1: Setup & Installation (WITH Catalog)**

  1. Create CRD and CSV with permissions (ServiceAccount, Role, RoleBinding)
  2. Create catalog source
  3. Create subscription to install operator from catalog
  4. Verify CSV reaches Succeeded state
  5. Verify all OLM-managed resources exist:
    - Deployment (with correct replicas)
    - ServiceAccount
    - Role
    - RoleBinding
  6. Record UIDs of all resources

  **Phase 2: Catalog Deletion**

  7. Delete catalog source
  8. Verify catalog source and its pod are completely removed

  **Phase 3: Passive Health Checks (WITHOUT Catalog)**

  9. Consistently verify for 3 minutes:
    - CSV remains in Succeeded phase
    - Deployment stays healthy with same UID (not recreated)
    - Replica counts remain correct

  **Phase 4: Active OLM Management Tests (WITHOUT Catalog)**

  10. Config Management Test:
    - Update subscription to add environment variable
    - Verify OLM propagates it to deployment
    - Proves: OLM actively manages deployment based on subscription config
  11. Reconciliation Test:
    - Delete all OLM-managed resources:
        - Deployment
    - Wait for deletion confirmation
    - Verify OLM recreates ALL of them with new UIDs
    - Verify recreated deployment still has subscription config (env var)
    - Proves: OLM actively reconciles CSV state and recreates the deployment only

  **Phase 5: Final State Validation**

  12. Verify subscription still tracks the installed CSV
  13. Verify CRD still exists and is functional

  **What This Proves**

  The test validates that after catalog deletion, OLM still:
  - ✅ Keeps CSV in healthy state
  - ✅ Manages deployment configuration via subscription
  - ✅ Reconciles and recreates deleted OLM-managed resources (Deployment ONLY)
  - ✅ Maintains subscription tracking
  - ✅ Preserves all installed CRDs

PS.:  ServiceAccount, Role, RoleBinding has not been re-created

### DIFF
--- a/test/e2e/catalog_e2e_test.go
+++ b/test/e2e/catalog_e2e_test.go
@@ -25,6 +25,7 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	networkingv1 "k8s.io/api/networking/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
@@ -98,8 +99,8 @@ var _ = Describe("Starting CatalogSource e2e tests", Label("CatalogSource"), fun
 
 		defer func() {
 			Eventually(func() error {
-				return ctx.Ctx().KubeClient().ApiextensionsInterface().ApiextensionsV1().CustomResourceDefinitions().Delete(context.Background(), crd.GetName(), metav1.DeleteOptions{})
-			}).Should(Or(Succeed(), WithTransform(k8serror.IsNotFound, BeTrue())))
+				return client.IgnoreNotFound(ctx.Ctx().Client().Delete(context.Background(), &crd))
+			}).Should(Succeed())
 			Eventually(func() error {
 				return client.IgnoreNotFound(ctx.Ctx().Client().Delete(context.Background(), &csv))
 			}).Should(Succeed())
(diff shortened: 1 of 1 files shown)


camilamacedo86:
> Curious to know, what motivated this test?

We will remove the marketplace index.
The motivation is to ensure that if any catalog is removed, everything continues to work correctly (except for upgrades that explicitly require that catalog).

Additionally, we should ensure that the installed solution works even if we temporarily cannot reach the catalog—for example, if the registry is unavailable for any reason.

test/e2e/catalog_e2e_test.go
```diff
@@ -1783,6 +1783,282 @@ var _ = Describe("Starting CatalogSource e2e tests", Label("CatalogSource"), fun
...
+			container := deployment.Spec.Template.Spec.Containers[0]
+			for _, env := range container.Env {
+				if env.Name == "TEST_ENV_VAR" && env.Value == "test-value" {
+					return nil
+				}
+			}
+			return fmt.Errorf("TEST_ENV_VAR not found in deployment")
+		}, pollDuration, pollInterval).Should(Succeed())
+
+		By("Delete an operator pod to test auto-healing")
```
perdasilva: Given that this is governed by the Deployment controller, do we need to test it here?

camilamacedo86: We need ensure that OLM will recreate all that is managed by it.
That was the idea. But I think you are right we might need a better one to delete here

perdasilva: But the Pods are managed by the Deployment. Maybe a better test would be if we delete the Deployment, it is recreated?

camilamacedo86: Yes, we need to delete what is managed by OLM to ensure that all still working in this case I think we have:

  1. The Deployment itself - OLM creates and owns this based on the CSV's InstallStrategy
  2. ServiceAccount - Created by OLM for the operator
  3. Roles/RoleBindings - Managed by OLM from the CSV spec
  
  So good catcher. I am changing that.

camilamacedo86: Updated here and the description so marking it as sorted out.

test/e2e/catalog_e2e_test.go
```diff
@@ -1783,6 +1784,469 @@ var _ = Describe("Starting CatalogSource e2e tests", Label("CatalogSource"), fun
...
+			if err != nil {
+				return err
+			}
+			if len(pods.Items) > 0 {
+				return fmt.Errorf("catalog source pod still exists: %d pods found", len(pods.Items))
+			}
+			return nil
+		}, pollDuration, pollInterval).Should(Succeed())
+
+		By("Verify CSV remains in succeeded state after catalog deletion")
```
ankitathomas, line 2005: Should we also check Subscription behavior here for missing catalogsource?

camilamacedo86: That is a great idea 🎉 
Adding that thank you

test/e2e/catalog_e2e_test.go
```diff
@@ -1783,6 +1784,404 @@ var _ = Describe("Starting CatalogSource e2e tests", Label("CatalogSource"), fun
...
+					},
+				},
+			},
+		}
+		strategy := newNginxInstallStrategy(genName("dep-"), permissions, nil)
+		csv := newCSV(packageStable, generatedNamespace.GetName(), "", semver.MustParse("0.1.0"), []apiextensionsv1.CustomResourceDefinition{crd}, nil, &strategy)
+
+		defer func() {
+			Eventually(func() error {
+				return ctx.Ctx().KubeClient().ApiextensionsInterface().ApiextensionsV1().CustomResourceDefinitions().Delete(context.Background(), crd.GetName(), metav1.DeleteOptions{})
```
perdasilva, line 1815: Why not use client.IgnoreNotFound here?

tmshort:
Looks good to me, but might want to address Per's comment.

test/e2e/catalog_e2e_test.go
camilamacedo86: Done thx

PR #3738, opened 2026-01-19
WIP / TEST Fix install plan delete catalog

camilamacedo86:


camilamacedo86:
It is not a bug.
  This is NOT what OLM does today and violates the principle: "We should ONLY recreate what OLM creates" - because OLM (via InstallPlan) doesn't recreate deleted RBAC.

  The test expectations should align with master behavior: RBAC deletion should cause CSV to go Pending, not trigger recreation.

Closing my test

PR #3746, opened 2026-01-22
Mark 'operator workload continues running after catalog source is deleted' test as flaky

tmshort:
The test has been failing pretty consistently since its introduced in #3737

### DIFF
--- a/test/e2e/catalog_e2e_test.go
+++ b/test/e2e/catalog_e2e_test.go
@@ -1785,7 +1785,7 @@ var _ = Describe("Starting CatalogSource e2e tests", Label("CatalogSource"), fun
 		})
 	})
 
-	It("operator workload continues running after catalog source is deleted", func() {
+	It("[FLAKE] operator workload continues running after catalog source is deleted", func() {
 		By("Create CRD and CSV for operator")
 		packageName := genName("nginx-")
 		stableChannel := "stable"


anik120:
There was a huge effort to mark existing tests as "flake" a while back, to assuage the immensely painful (from dev's POV, `blocking` from the business's POV) experience in trying to get anything in. 


I don't think we should be adding on to that anymore. I'm looking at #3737, and I don't understand why that test HAD to be added.   

What do we loose if we revert that PR for now since this test is flaking?

tmshort:
> There was a huge effort to mark existing tests as "flake" a while back, to assuage the immensely painful (from dev's POV, `blocking` from the business's POV) experience in trying to get anything in.
> 
> I don't think we should be adding on to that anymore. I'm looking at #3737, and I don't understand why that test HAD to be added.
> 
> What do we loose if we revert that PR for now since this test is flaking?

@camilamacedo86 added the test recently because we're removing a catalog, and we needed a test for it. She will be looking at it soon. In the meantime, short of reverting #3737, this is the best solution.

anik120:
Oh looks like I had the same question on that PR too 

https://github.com/operator-framework/operator-lifecycle-manager/pull/3737#pullrequestreview-3670994189

and there was an answer 

https://github.com/operator-framework/operator-lifecycle-manager/pull/3737#issuecomment-3760394366

but I didn't get a chance to get around to the discussion. 

The "workload" you get when an operator is installed is handled by a different controller responsible for reconciling the ClusterServiceVersion CRs (olm-operator), than the one responsible for reconciling the catalogs (catalog-operator). 

This doc outlines that info https://github.com/operator-framework/operator-lifecycle-manager/blob/master/doc/design/architecture.md#architecture

ie I'm trying to make the case that "when a catalog is removed, the CSVs are untouched" is an axiom

tmshort:
Possibly defer to #3747

PR #3747, opened 2026-01-22
🌱 (e2e): Enhance e2e test to check workiload resilience when catalogs goes away

camilamacedo86:
Enhance the code for better readability and avoid flakes e.g. https://github.com/operator-framework/operator-lifecycle-manager/actions/runs/21225377668/job/61071090835

Follow up: https://github.com/operator-framework/operator-lifecycle-manager/pull/3737

### DIFF
--- a/test/e2e/catalog_e2e_test.go
+++ b/test/e2e/catalog_e2e_test.go
@@ -1864,28 +1864,23 @@ var _ = Describe("Starting CatalogSource e2e tests", Label("CatalogSource"), fun
 
 		By("Wait for operator deployment to be ready")
 		var operatorDeployment *appsv1.Deployment
-		Eventually(func() error {
+		Eventually(func(g Gomega) {
+			var err error
 			operatorDeployment, err = c.GetDeployment(generatedNamespace.GetName(), deploymentName)
-			if err != nil {
-				return err
-			}
-			if operatorDeployment.Spec.Replicas == nil || *operatorDeployment.Spec.Replicas == 0 {
-				return fmt.Errorf("deployment replicas is not set")
-			}
-			if operatorDeployment.Status.AvailableReplicas != *operatorDeployment.Spec.Replicas {
-				return fmt.Errorf("deployment %s not ready: %d/%d replicas available",
-					deploymentName,
-					operatorDeployment.Status.AvailableReplicas,
-					*operatorDeployment.Spec.Replicas)
-			}
-			if operatorDeployment.Status.ReadyReplicas != *operatorDeployment.Spec.Replicas {
-				return fmt.Errorf("deployment %s not ready: %d/%d replicas ready",
-					deploymentName,
-					operatorDeployment.Status.ReadyReplicas,
-					*operatorDeployment.Spec.Replicas)
-			}
-			return nil
-		}, pollDuration, pollInterval).Should(Succeed())
+			g.Expect(err).ShouldNot(HaveOccurred())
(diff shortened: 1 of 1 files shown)


anik120:
Copying my comment from https://github.com/operator-framework/operator-lifecycle-manager/pull/3746#issuecomment-3786226834:

Oh looks like I had the same question on that PR too https://github.com/operator-framework/operator-lifecycle-manager/pull/3737#pullrequestreview-3670994189 and there was an answer https://github.com/operator-framework/operator-lifecycle-manager/pull/3737#issuecomment-3760394366

>Curious to know, what motivated this test?

>We will remove the marketplace index.
The motivation is to ensure that if any catalog is removed, everything continues to work correctly (except for upgrades that explicitly require that catalog).
>Additionally, we should ensure that the installed solution works even if we temporarily cannot reach the catalog—for example, if the registry is unavailable for any reason.


but I didn't get a chance to get around to the discussion.

The "workload" you get when an operator is installed is handled by a different controller responsible for reconciling the ClusterServiceVersion CRs (olm-operator), than the one responsible for reconciling the catalogs (catalog-operator).

This doc outlines that info https://github.com/operator-framework/operator-lifecycle-manager/blob/master/doc/design/architecture.md#architecture

ie I'm trying to make the case that "when a catalog is removed, the CSVs are untouched" is an axiom

camilamacedo86:
Hi @anik120 

When a catalog is removed — or when we can’t retrieve catalog data — we should still ensure OLM continues to properly manage and reconcile all resources it already manages. From the end-user perspective, the only impact should be that upgrades can’t proceed (because there’s no catalog data). Anything previously installed should continue running and being reconciled normally.

That’s exactly what we’re validating with these tests.

**Example**: As a user, I want to update the Solution configuration and have the changes applied, even if the referenced Catalog no longer exists, so that I can continue managing the installation and OLM can keep reconciling it.

(see the description of the PR: https://github.com/operator-framework/operator-lifecycle-manager/pull/3737 )
Delete the Catalog 
....
Config Management Test:
- Update subscription to add environment variable
- Verify OLM propagates it to deployment
- Proves: OLM actively manages deployment based on subscription config

Same added for OLMv1. See: https://github.com/operator-framework/operator-controller/pull/2439 ( with bug fixes since there we have an issue. We were not reconciling the content prior that PR). I hope that clarifies. 

I hope that clarifies your questions of https://github.com/operator-framework/operator-lifecycle-manager/pull/3747#issuecomment-3786257913.

tmshort:
/lgtm
/approve

Because this is a flake, I'm re-running `e2e / e2e (0)` to be sure.

anik120:
Yea I looked at the test, and I'm still of the opinion that we should discuss this before adding a test. 

I am coming from the perspective of us having decided we won't be adding anything to v0 unless absolutely necessary and as I said, with this whole premise being an axiom, this test is not necessary  


Having said that, it's an e2e test, and won't affect customers. It will only affect developers so I'll leave it up to you guys to figure it out.

camilamacedo86:
Hi @anik120 

This isn’t a feature. It’s an end-to-end test to verify the existing behavior and ensure we don’t break the expected behavior going forward. It does not affect developers it just prevent we broke the expected behaviour and help us to ensure that it works as intent to work. PS. It was discussed and followed the process.

tmshort:
Ugh... this merged before `e2e (0)` finished... waiting for results.

tmshort:
And it passed...

PR #3766, opened 2026-02-11
abandon evaluation of any new catalogsource image which is pathological

grokspawn:

This PR adds a pathological status check for when container status indicates crashloopbackoff. 
If that status matches, OLM will abandon catalog evaluation for that image, dispose of the pod, and pull a new image when the `pollInterval` comes up again.

This will now cause the catalog operator to emit a message of the form 

```sh
time="2026-02-11T22:30:17Z" level=info msg="catalog polling result: update pod poison-pill-9qvmp failed to start" UpdatePod=poison-pill-9qvmp
```

and the offending pod will be deleted.

In the case that a catalogsource defines `.spec.grpcPodConfig.extractContent` it is possible for OLMv0 to get trapped in an evaluation loop if the catalogsource is not compatible with the on-cluster catalogsource service.

This is because the on-cluster catalog services which use `extractContent` define two initContainers and a service container.  When those initContainers succeed, the pod status progresses to RUNNING regardless of the success/failure of the service container.  

If the service container fails, it will halt, and the pod will start being rebooted by kube when it fails readiness/liveness probes.  It will remain in RUNNING status, so OLM will requeue its evaluation without end. 

**Architectural changes:**

**Testing remarks:**

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/registry/reconciler/grpc.go
+++ b/pkg/controller/registry/reconciler/grpc.go
@@ -30,8 +30,8 @@ const (
 	CatalogSourceUpdateKey      = "catalogsource.operators.coreos.com/update"
 	ServiceHashLabelKey         = "olm.service-spec-hash"
 	CatalogPollingRequeuePeriod = 30 * time.Second
-	// pathologicalRestartCountThreshold is the number of container restarts above which a pod is considered failed for catalog polling.
-	pathologicalRestartCountThreshold = 10
+	// containerReasonCrashLoopBackOff is the kubelet Waiting reason when a container is backing off after repeated crashes.
+	containerReasonCrashLoopBackOff = "CrashLoopBackOff"
 )
 
 // grpcCatalogSourceDecorator wraps CatalogSource to add additional methods
@@ -756,15 +756,15 @@ func swapLabels(pod *corev1.Pod, labelKey, updateKey string) *corev1.Pod {
 	return pod
 }
 
-// podPathologicallyRestarting returns true if any container in the pod has restarts exceeding pathologicalRestartCountThreshold.
-func podPathologicallyRestarting(pod *corev1.Pod) bool {
+// podContainersArePathological returns true if any container in the pod has restarts exceeding pathologicalRestartCountThreshold or is in CrashLoopBackOff.
+func podContainersArePathological(pod *corev1.Pod) bool {
 	for _, s := range pod.Status.ContainerStatuses {
-		if s.RestartCount > pathologicalRestartCountThreshold {
+		if s.State.Waiting != nil && s.State.Waiting.Reason == containerReasonCrashLoopBackOff {
 			return true
 		}
 	}
 	for _, s := range pod.Status.InitContainerStatuses {
-		if s.RestartCount > pathologicalRestartCountThreshold {
+		if s.State.Waiting != nil && s.State.Waiting.Reason == containerReasonCrashLoopBackOff {
 			return true
 		}
 	}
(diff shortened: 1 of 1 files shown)


pkg/controller/registry/reconciler/grpc.go
```diff
@@ -754,9 +756,25 @@ func swapLabels(pod *corev1.Pod, labelKey, updateKey string) *corev1.Pod {
 	return pod
 }
 
+// podPathologicallyRestarting returns true if any container in the pod has restarts exceeding pathologicalRestartCountThreshold.
+func podPathologicallyRestarting(pod *corev1.Pod) bool {
```
joelanford: Should this also (or instead) check for `CrashLoopBackoff`?

There are cases like node reboots that cause the container restart counter to increment because something killed the containers, not be cause the containers were crashing.

grokspawn: My concern with doing that is that the threshold would be immediate.  If the pod went CLBO we would immediately destroy the pod and pull a new image at the next interval.  In the past, we have been very careful about creating a state where we would need to pull the image again, but if we're OK with that I can.

grokspawn: At any rate, I added that consideration in a new commit so we could talk about it.  PTAL.

joelanford: > pull a new image at the next interval

At the next interval, there are two possibilities:
1. It's the same digest. If the pod is scheduled to a new node, then we'd re-pull that image to that new node. If the pod is scheduled on a node that had pulled that digest before, no pull necessary. Eventually, all the nodes could end up pulling the bad digest, but each node would only pull the bad digest once.
2. It's a different digest. We _want_ to pull this to get past the digest with the bad data.

pkg/controller/registry/reconciler/grpc.go
```diff
@@ -756,15 +758,20 @@ func swapLabels(pod *corev1.Pod, labelKey, updateKey string) *corev1.Pod {
...
+// containerInCrashLoopBackOff returns true if the container status indicates CrashLoopBackOff.
+func containerInCrashLoopBackOff(s corev1.ContainerStatus) bool {
+	return s.State.Waiting != nil && s.State.Waiting.Reason == containerReasonCrashLoopBackOff
+}
+
+// podContainersArePathological returns true if any container in the pod has restarts exceeding pathologicalRestartCountThreshold or is in CrashLoopBackOff.
+func podContainersArePathological(pod *corev1.Pod) bool {
 	for _, s := range pod.Status.ContainerStatuses {
-		if s.RestartCount > pathologicalRestartCountThreshold {
+		if s.RestartCount > pathologicalRestartCountThreshold || containerInCrashLoopBackOff(s) {
```
joelanford: You could maybe make this `&&` instead of `||` so that both conditions have to be true in order to kill it.

Assuming no restarts due to other reasons, how long does it take a new pod to get to 10 restarts when it is immediately crashing? The 'Backoff' part of `CrashLookBackoff` means that we wait longer and longer between restarts.

grokspawn: Based on your review, I ended up dropping the restart-count-based approach entirely.  The timing previously would be ready-probe+liveness-probe * threshold.  
For my test scenario with a 5m pollInterval yesterday, we hit the threshold on my local machine at about 20 minutes.

pkg/controller/registry/reconciler/grpc.go
```diff
@@ -30,6 +30,8 @@ const (
 	CatalogSourceUpdateKey      = "catalogsource.operators.coreos.com/update"
 	ServiceHashLabelKey         = "olm.service-spec-hash"
 	CatalogPollingRequeuePeriod = 30 * time.Second
+	// containerReasonCrashLoopBackOff is the kubelet Waiting reason when a container is backing off after repeated crashes.
+	containerReasonCrashLoopBackOff = "CrashLoopBackOff"
```
joelanford, line 34: Just checking there isn't already a constant defined for this in corev1 of k8s.io/api?

pkg/controller/registry/reconciler/grpc.go
```diff
@@ -754,9 +756,25 @@ func swapLabels(pod *corev1.Pod, labelKey, updateKey string) *corev1.Pod {
...
+		if s.State.Waiting != nil && s.State.Waiting.Reason == containerReasonCrashLoopBackOff {
+			return true
+		}
+	}
+	for _, s := range pod.Status.InitContainerStatuses {
+		if s.State.Waiting != nil && s.State.Waiting.Reason == containerReasonCrashLoopBackOff {
+			return true
+		}
+	}
+	// TODO: currently no ephemeral containers in a catalogsource, should we add checks anyway?
```
joelanford, line 771: Ephemeral containers should never be part of the equation I don't think.

See: https://kubernetes.io/docs/concepts/workloads/pods/ephemeral-containers/

PR #3770, opened 2026-02-16
OCPBUGS-76971: Update NetworkPolicy egress to support HyperShift custom API ports

rashmigottipati:
**Description of the change:**

This PR updates NetworkPolicy egress rules in OLM v0 to support HyperShift deployments that use custom Kubernetes API server ports.

**Motivation for the change:**

 The current NetworkPolicy rules hardcode port 6443 for Kubernetes API server access, which breaks HyperShift deployments that customize the API server port. Based on the latest guidance across OCP, we should:
  1. Use wildcard egress `{}` for kube-apiserver to allow all IPs/ports
  2. Use explicit DNS egress rules with ports 53 and 5353 (TCP/UDP) for DNS resolution

This approach aligns with related projects (catalogd, operator-controller) and the companion fix in operator-marketplace [(OCPBUGS-66980).](https://issues.redhat.com/browse/OCPBUGS-66980)

**Architectural changes:**

**Testing remarks:**

**Reviewer Checklist**
- [ ] Implementation matches the proposed design, or proposal is updated to match implementation
- [ ] Sufficient unit test coverage
- [ ] Sufficient end-to-end test coverage
- [ ] Bug fixes are accompanied by regression test(s)
- [ ] e2e tests and flake fixes are accompanied evidence of flake testing, e.g. executing the test 100(0) times
- [ ] tech debt/todo is accompanied by issue link(s) in comments in the surrounding code
- [ ] Tests are comprehensible, e.g. Ginkgo DSL is being used appropriately
- [ ] Docs updated or added to `/doc`
- [ ] Commit messages sensible and descriptive
- [ ] Tests marked as `[FLAKE]` are truly flaky and have an issue
- [ ] Code is properly formatted

Closes https://issues.redhat.com/browse/OCPBUGS-76971

PR #3778, opened 2026-02-28
feat(helm): add feature flag to make network policies optional

maxsxu:

- Added a new  feature flag `networkPolicy.enabled` to control network policies. Default to true

It's useful we can have the ability to choose enable or disable NetworkPolicies

**Architectural changes:**

N/A

**Testing remarks:**

**Reviewer Checklist**

### DIFF
--- a/deploy/chart/templates/0000_50_olm_01-networkpolicies.yaml
+++ b/deploy/chart/templates/0000_50_olm_01-networkpolicies.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.networkPolicy.enabled }}
 apiVersion: networking.k8s.io/v1
 kind: NetworkPolicy
 metadata:
@@ -85,3 +86,4 @@ spec:
     - { }
   egress:
     - { }
+{{- end }}
--- a/deploy/chart/values.yaml
+++ b/deploy/chart/values.yaml
@@ -104,6 +104,7 @@ serviceCa:
     secretName: ""
 
 networkPolicy:
+  enabled: true
   dns:
     ports:
       - protocol: TCP


tmshort:
Hi @maxsxu, thanks for the contribution.

Could you provide a use-case for disabling the NetworkPolicies? We're concerned that doing so would negatively impact functionality (due to other NPs in the cluster) and reduce the level of security in the cluster.

maxsxu:
Hi @tmshort the main use cause is for dev purpose in test environment, it's useful to observe the default behavior once we can deploy chart w/o NPs since we found the latest olm will generate NPs from CatalogSource CR.

And the default value is true of this new flag, so won't break existing installed chart.

grokspawn:
Considering that this PR doesn't change the [isSupported](https://github.com/operator-framework/operator-lifecycle-manager/blob/f6925fde7820b275e766af66363b5dab4fb9bff5/pkg/controller/operators/catalog/supportedresources.go#L32) function to prevent operators from installing NPs, I think we cannot reasonably assert OLM's network needs w/o a NP (so that they cannot be broken by an operator install into an OLM namespace which contains a NP in its manifests). 

I think we wouldn't want to provide such a flag -- even with a NOP default case -- without a comprehensive NP killswitch. 

In your practice @maxsxu , are you installing operators to a namespace containing OLM core components or catalogsources?

joelanford:
Also, FWIW, a recent change went in (https://github.com/operator-framework/operator-lifecycle-manager/pull/3770) that relaxes the egress rules for apiserver access because we found that it is not possible to use a more specific rule given the wide variety of network configurations of kube-apiserver in real environments.

In question I haven't seen asked or answered yet: 

> It's useful we can have the ability to choose enable or disable NetworkPolicies

Why? Are you running into an issue where the default NetworkPolicy setup is not working?

PR #3784, opened 2026-03-04
align probe exec timeouts with k8s probe timeouts

grokspawn:

**Architectural changes:**

**Testing remarks:**

**Reviewer Checklist**

### DIFF
--- a/pkg/controller/registry/reconciler/reconciler_test.go
+++ b/pkg/controller/registry/reconciler/reconciler_test.go
@@ -38,7 +38,7 @@ func TestPodMemoryTarget(t *testing.T) {
 				ObjectMeta: metav1.ObjectMeta{
 					GenerateName: "test-",
 					Namespace:    "testns",
-					Labels:       map[string]string{"olm.pod-spec-hash": "8SbHWyYfjbRT8lLcfdZ5ofXNdC1GE6ayztILTF", "olm.managed": "true"},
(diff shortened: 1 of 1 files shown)


grokspawn:
> Have you noticed any issues with probe timeouts that motivated adding `-rpc-timeout=5s`? Or was this mainly to align with Kubernetes defaults?

Yeah, there are a lot of times in the logs/events where the catsrc pods would claim a startupProbe failure because the catalog size prevented its GRPC availability w/in 1s.  The retries always recovered it, but the logs always include a statement about the 1s failure.  

> omc get events -n openshift-marketplace | grep "redhat-operators" | grep -i "probe\|unhealthy" | head -20)
  ⎿  3h35m       Warning   Unhealthy           pod/redhat-operators-2kjzx
      Startup probe failed: timeout: failed to connect service ":50051" within 1s
     3h25m       Warning   Unhealthy           pod/redhat-operators-2kjzx

This is weird, because we don't configure any 1s thresholds, but this is because it's the default for grpc_health_probe. 

Functionally this change has no real effect.  It will just result in avoiding the initial failure log/event for catalogs of a certain threshold size or higher, and the confusing message.

grokspawn:
Thanks @rashmigottipati .  Could I bother you for another lgtm?  I had to fix the whitespace-only formatting on the test file.  (Maybe wait until we see the sanity CI pass so you don't throw away the lgtm because for some reason I can't run `make verify` on my local.)

rashmigottipati (approved):
/lgtm

PR #3798, opened 2026-03-18
:bug: Fix CRD validation to only validate CRs against storage version schema

jianzhangbjz:

  Before:
  ```go
  // Validate against ALL served versions
  for _, newVersion := range newCRD.Spec.Versions {
      if oldVersionSet.Has(newVersion.Name) && newVersion.Served {
          validationsMap[newVersion.Name] = newVersion.Schema
      }
  }
```
  After:
```go
  // Find the storage version
  var oldStorageVersion string
  for _, v := range oldCRD.Spec.Versions {
      if v.Storage {
          oldStorageVersion = v.Name
          break
      }
  }

  // Only validate against storage version's schema
  for _, newVersion := range newCRD.Spec.Versions {
      if newVersion.Name == oldStorageVersion && newVersion.Served {
          validationsMap[newVersion.Name] = newVersion.Schema
          break
      }
  }
```
Why This Is Safe
  1. Storage version semantics: CRs are always stored in the storage version format, so validating against the storage version's schema is the correct behavior.
  2. Existing safety checks: OLM already has `SafeStorageVersionUpgrade` check that prevents unsafe storage version changes.
  3. Webhook conversion handling: The existing code already handles `conversion.strategy: Webhook` by allowing upgrades to proceed with a warning, trusting the webhook to convert data correctly.
  4. User-created CRs with non-storage versions: When users create CRs using a non-storage version (e.g., v1), Kubernetes validates them against that version's schema at creation time. The data will include all required fields, so validation against the storage version's schema (which has fewer required fields) will pass.

When upgrading a CRD that has multiple versions with `conversion.strategy: None`, OLM incorrectly validates existing CRs against **all served versions' schemas**. This causes upgrade failures when different versions have different required fields.
More: https://redhat.atlassian.net/browse/OCPBUGS-78710

**Architectural changes:**

**Testing remarks:**
  - Updated existing unit tests to use storage version CRs, which accurately reflects how Kubernetes stores CRs
  - All TestValidateV1CRDCompatibility tests pass
  - All TestValidateV1Beta1CRDCompatibility tests pass

**Reviewer Checklist**

Assisted-By: Claude-Code

### DIFF
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -2177,80 +2177,119 @@ func transitionInstallPlanState(log logrus.FieldLogger, transitioner installPlan
 	}
 }
 
-// Validate all existing served versions against new CRD's validation (if changed)
+// Validate existing CRs against the new CRD's validation schema for the storage version.
+// When a CRD has multiple versions with conversion.strategy=None, CRs are stored in the
+// storage version format. We should only validate CRs against the storage version's schema,
+// not against all served versions, because without a conversion webhook the CR data won't
+// be converted to match other versions' schemas.
 func validateV1CRDCompatibility(dynamicClient dynamic.Interface, oldCRD *apiextensionsv1.CustomResourceDefinition, newCRD *apiextensionsv1.CustomResourceDefinition) error {
 	logrus.Debugf("Comparing %#v to %#v", oldCRD.Spec.Versions, newCRD.Spec.Versions)
 
-	oldVersionSet := sets.New[string]()
-	for _, oldVersion := range oldCRD.Spec.Versions {
-		if !oldVersionSet.Has(oldVersion.Name) && oldVersion.Served {
-			oldVersionSet.Insert(oldVersion.Name)
+	// Find the storage version from the old CRD - this is the version CRs are actually stored in
+	var oldStorageVersion string
+	for _, v := range oldCRD.Spec.Versions {
+		if v.Storage {
+			oldStorageVersion = v.Name
+			break
 		}
 	}
 
+	if oldStorageVersion == "" {
+		return fmt.Errorf("no storage version found in old CRD %s", oldCRD.Name)
+	}
+
+	// Only validate existing CRs against the new CRD's schema for the old storage version.
+	// This is the correct behavior because:
+	// 1. All existing CRs are stored in the old storage version format
+	// 2. When conversion.strategy is "None", Kubernetes doesn't convert CR data between versions
+	// 3. Validating CRs against other versions' schemas would incorrectly fail when those
+	//    versions have different required fields
 	validationsMap := make(map[string]*apiextensions.CustomResourceValidation, 0)
 	for _, newVersion := range newCRD.Spec.Versions {
-		if oldVersionSet.Has(newVersion.Name) && newVersion.Served {
-			// If the new CRD's version is present in the cluster and still
-			// served then fill the map entry with the new validation
+		if newVersion.Name == oldStorageVersion && newVersion.Served {
 			convertedValidation := &apiextensions.CustomResourceValidation{}
 			if err := apiextensionsv1.Convert_v1_CustomResourceValidation_To_apiextensions_CustomResourceValidation(newVersion.Schema, convertedValidation, nil); err != nil {
 				return err
 			}
 			validationsMap[newVersion.Name] = convertedValidation
+			break
 		}
 	}
+
+	// If the old storage version is no longer served in the new CRD, skip validation.
+	// This scenario should be caught by SafeStorageVersionUpgrade check.
+	if len(validationsMap) == 0 {
+		logrus.Debugf("Old storage version %s not found or not served in new CRD %s, skipping CR validation", oldStorageVersion, newCRD.Name)
+		return nil
+	}
+
 	return validateExistingCRs(dynamicClient, schema.GroupResource{Group: newCRD.Spec.Group, Resource: newCRD.Spec.Names.Plural}, validationsMap)
 }
 
-// Validate all existing served versions against new CRD's validation (if changed)
+// Validate existing CRs against the new CRD's validation schema for the storage version.
+// See validateV1CRDCompatibility for detailed explanation.
 func validateV1Beta1CRDCompatibility(dynamicClient dynamic.Interface, oldCRD *apiextensionsv1beta1.CustomResourceDefinition, newCRD *apiextensionsv1beta1.CustomResourceDefinition) error {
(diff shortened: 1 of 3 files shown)


pkg/controller/operators/catalog/operator.go
```diff
@@ -2177,80 +2177,119 @@ func transitionInstallPlanState(log logrus.FieldLogger, transitioner installPlan
 	}
 }
 
-// Validate all existing served versions against new CRD's validation (if changed)
+// Validate existing CRs against the new CRD's validation schema for the storage version.
+// When a CRD has multiple versions with conversion.strategy=None, CRs are stored in the
+// storage version format. We should only validate CRs against the storage version's schema,
+// not against all served versions, because without a conversion webhook the CR data won't
+// be converted to match other versions' schemas.
```
kevinrizza, line 2184: This comment is oddly phrased. A conversion webhook existing or not doesn't change whether CRs should be validated. Yes, a conversion webhook can allow a client to translate a resource from one version to another, but the lack of a conversion webhook doesn't mean that a client that is trying to access a resource at one version doesn't have to worry about version mismatch if a webhook doesn't exist.

pkg/controller/operators/catalog/operator.go
```diff
@@ -2177,80 +2177,119 @@ func transitionInstallPlanState(log logrus.FieldLogger, transitioner installPlan
...
+	var oldStorageVersion string
+	for _, v := range oldCRD.Spec.Versions {
+		if v.Storage {
+			oldStorageVersion = v.Name
+			break
 		}
 	}
 
+	if oldStorageVersion == "" {
+		return fmt.Errorf("no storage version found in old CRD %s", oldCRD.Name)
```
kevinrizza, line 2198: This isn't a possible case. There must always be exactly one version in the CRD spec where storage is true. If not, this CSV should fail admission.

pkg/controller/operators/catalog/operator.go
```diff
@@ -2177,80 +2177,119 @@ func transitionInstallPlanState(log logrus.FieldLogger, transitioner installPlan
...
+	}
+
+	// Only validate existing CRs against the new CRD's schema for the old storage version.
+	// This is the correct behavior because:
+	// 1. All existing CRs are stored in the old storage version format
+	// 2. When conversion.strategy is "None", Kubernetes doesn't convert CR data between versions
+	// 3. Validating CRs against other versions' schemas would incorrectly fail when those
+	//    versions have different required fields
```
kevinrizza, line 2206: I'm going to push back on this premise a little bit. The whole purpose of this schema validation is to ensure that new versions being introduced do not ratchet validation against existing data on the cluster. But why would a user ever want to introduce a new served version if it cannot actually be served? It is true that this will technically not break anything, but the whole point of this conversion validation is to make sure that data is being migrated to a new valid format in the previous version before introducing the new version.

If we are going to change this behavior, I think we need to write down exactly what test cases we are expecting this to help with in a way that is verifiable and include them as e2e tests that prove that we aren't accidentally regressing in other areas where this validation is still what we want to do.

I would also argue that what probably should happen here is that this should be an opt in for operator developers. The whole point of OLM is to restrict what upgrades we do and don't support to ensure that users that are not experts in CRD lifecycle don't accidentally shoot themselves in the foot. If a user wants to intentionally introduce a CRD version that cannot actually be served because there is no way to convert the data automatically, then they are breaking clients and they should know that they need to tell their users as much.

kevinrizza (changes requested):
I think I need to see an enhancement or design document for how we want to introduce this change as well as a comprehensive test plan in order to approve this pr.

grokspawn:
The goal of the stored+serving validation is that we avoid conditions which can make the resulting resources unmanageable.  A listed resource via the newer CRD API version could not be applied against the newer API, breaking any new client. 
Normally, our expectation would be that there is at least one release between offering the optional fields and requiring them, and during that breaking change timeline the operator team would have guidance to admins that they _must_ ensure that they provide the optional fields before upgrading to the next version.  
And people who don't ... well, they get an error when they attempt it, telling them that they need to $doThing.  

I do agree with Kevin that any attempt to change the current scheme needs to be researched, supported, and evangelized broadly.

jianzhangbjz:
Hi @kevinrizza @grokspawn , many thanks for your comments! I appreciate the design rationale behind OLM's current validation behavior. The intent is to prevent users from ending up in an unmanageable state where `Custom Resources` can be retrieved via a newer API version but cannot be applied back due to schema validation failures. This is a reasonable safeguard.                                                                                                                                                                                              
#### The expected workflow of the current OLM design, as I understand it, is:                                                                                                                          
  1. Operator developers provide comprehensive migration documentation                                                                                                    
  2. Users review the documentation and identify which fields require updates                                                                                             
  3. Users manually update all existing Custom Resources prior to initiating the upgrade                                                                                  
  4. Users proceed with the operator upgrade  
  
####  However, in practice, this workflow faces several challenges:                                                                                                                                                                                                                                                                                 
  - Documentation gaps: Not all Operator developers provide adequate migration guidance                                                                                   
  - Automatic upgrades: When `installPlanApproval` is set to `Automatic`, users have no opportunity to update CRs before OLM performs validation                              
  - Information asymmetry: Users cannot determine which fields the new CRD version requires until the upgrade is attempted, yet the upgrade is blocked — creating a circular dependency                                                                                                                                                     
  - Insufficient transition periods: Some Operators introduce required fields in new API versions **without first offering a transitional release where those fields are optional** 

#### Proposed middle-ground solutions:                                                                                                                                                                                           
  1. Warning instead of blocking: Log a warning about potentially incompatible CRs but allow the upgrade to proceed, giving users visibility without breaking the upgrade path                                                      
  2. Validate only storage version: Since all CRs are stored in the storage version format and `conversion.strategy: None` **does not transform data**, validating against non-storage versions may not reflect actual compatibility                                                                                                               
  3. Pre-upgrade validation hook: Provide a mechanism for users to run validation checks before initiating an upgrade, giving them time to remediate issues
  4. Enhanced error messaging: Include actionable guidance in error messages, such as which specific CRs need updates and what values are expected
                                                                                                                                                
I believe a balanced approach would preserve the safety guarantees while improving the user experience in real-world upgrade scenarios.

grokspawn:
I don't see an argument above which counters the unmanageable state problem.  I think we need a brief on the proposal so we can consider the details.

jianzhangbjz:
Yeah, agree. But given that OLMv0 is in maintenance mode, I'm not sure if it's worth investing in a formal design proposal for this change. The workaround is sufficient for the user's immediate needs.                                                                                  
If the community believes this is worth addressing, perhaps it could be considered as part of OLMv1's design.

PR #3863, opened 2026-07-09
deploy/chart: add static NetworkPolicies for CatalogSource gRPC ingress and bundle unpack egress

### LINKED ISSUE #3676: NetworkPolicy blocks OLM catalog connectivity and bundle unpacking jobs
## Environment
- **OLM Version**: v0.36.0 (using `quay.io/operator-framework/olm:master`) and master branch
- **Kubernetes**: Rancher Desktop (K3s)
- **Catalog Source**: `quay.io/operatorhubio/catalog:latest`
- **Installation method**: Helm chart (`./deploy/chart`)

## Issue Summary

I encountered this issue on both **v0.36.0** and the **master** branch. After troubleshooting with Claude Code, we identified two NetworkPolicy misconfigurations that prevent OLM from functioning properly. Below are the findings and the workarounds that successfully resolved the issues.

## Problem Description

When deploying OLM with NetworkPolicies enabled (specifically a `default-deny-all-traffic` policy), two critical connectivity issues prevent OLM from functioning properly:

### Issue 1: Catalog gRPC connectivity blocked

The `operatorhubio-catalog` pod cannot receive incoming connections on port 50051, causing persistent connection failures:

```
failed to populate resolver cache from source operatorhubio-catalog/operator-lifecycle-manager: 
failed to list bundles: rpc error: code = Unavailable desc = connection error: 
desc = "transport: Error while dialing dial tcp 10.43.119.62:50051: connect: connection refused"
```

**Status observed:**
```
status:
  connectionState:
    lastObservedState: TRANSIENT_FAILURE
```

The Helm chart creates a `default-deny-all-traffic` NetworkPolicy that blocks all ingress by default, but doesn't include a corresponding NetworkPolicy to allow ingress to the catalog pod.

### Issue 2: Bundle unpacking jobs cannot access Kubernetes API

When a subscription attempts to install an operator, the bundle unpacking jobs fail to access the Kubernetes API server:

```
Error: error loading manifests from directory: 
Get "https://10.43.0.1:443/api/v1/namespaces/operator-lifecycle-manager/configmaps/...": 
dial tcp 10.43.0.1:443: connect: connection refused
```

This causes InstallPlans to fail with:
```
status:
  phase: Failed
  conditions:
  - type: BundleLookupFailed
    reason: BackoffLimitExceeded
    message: Job has reached the specified backoff limit
```

The `default-deny-all-traffic` NetworkPolicy blocks egress traffic, preventing jobs from reaching the Kubernetes API server.

## Expected Behavior

When NetworkPolicies are enabled, OLM should include appropriate NetworkPolicies to allow:
1. Ingress to catalog pods on port 50051 for gRPC communication
2. Egress from all pods to access the Kubernetes API server and external registries

## Reproduction Steps

1. Deploy OLM using the Helm chart with default NetworkPolicies
2. Create a CatalogSource pointing to `quay.io/operatorhubio/catalog:latest`
3. Create a Subscription for any operator (e.g., `cloudnative-pg`)
4. Observe catalog connection failures and InstallPlan failures

## Workaround Applied

Two additional NetworkPolicies were required to fix the issues:

### 1. Allow ingress to catalog pods

```
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  name: operatorhubio-catalog
  namespace: operator-lifecycle-manager
spec:
  podSelector:
    matchLabels:
      olm.catalogSource: operatorhubio-catalog
  policyTypes:
  - Ingress
  - Egress
  ingress:
  - from:
    - namespaceSelector: {}
    ports:
    - protocol: TCP
      port: 50051
  egress:
  - {}
```

### 2. Modify default-deny-all-traffic to allow egress

```
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  name: default-deny-all-traffic
  namespace: operator-lifecycle-manager
spec:
  podSelector: {}
  policyTypes:
  - Ingress
  - Egress
  egress:
  - {}  # Allow all egress traffic
```

**Result**: After applying these NetworkPolicies, OLM became fully operational with catalog connectivity working (`lastObservedState: READY`) and operators installing successfully.

## Proposed Solution

The OLM Helm chart should include:

1. **A catalog NetworkPolicy template** that automatically creates ingress rules for any CatalogSource pods, using label selectors like `olm.catalogSource`

2. **Modified default NetworkPolicy** that includes egress rules for:
   - Kubernetes API access (port 443)
   - DNS resolution (port 53 TCP/UDP)
   - Container registry access (port 443 for HTTPS registries)
   - gRPC catalog communication (port 50051)

This would allow OLM to work out-of-the-box in environments with strict NetworkPolicy enforcement, which is a common security requirement in production Kubernetes clusters.

## Additional Context

The existing Helm chart includes NetworkPolicies for `catalog-operator`, `olm-operator`, and `packageserver`, which do include appropriate egress rules. However:
- There's no NetworkPolicy for catalog pods themselves
- The `default-deny-all-traffic` policy is too restrictive for OLM's operational requirements

---

**Files to check:**
- `deploy/chart/templates/networkpolicy.yaml`
- Catalog operator NetworkPolicy definitions

grokspawn:
## Summary

Fixes #3676.

The Helm chart's `default-deny-all-traffic` NetworkPolicy blocked two traffic paths not covered by the existing static policies:

- **CatalogSource gRPC ingress (port 50051):** Registry pods spawned per CatalogSource could not receive inbound gRPC connections. The catalog-operator reconciler does create per-CatalogSource NetworkPolicies for this dynamically, but there is a bootstrapping gap — `default-deny-all-traffic` takes effect immediately on install/upgrade, while the per-CatalogSource policies only exist after the controller reconciles each object.
- **Bundle-unpack job egress:** Job pods created for bundle unpacking could not reach the Kubernetes API server or container registries. The API server port cannot be statically specified (it is implementation-defined), so a wildcard egress rule is used instead.

Two new NetworkPolicies are added to the chart:

| Policy | Selector | Effect |
|---|---|---|
| `olm-catalog-grpc-ingress` | pods with `olm.catalogSource` label (any value) | allow ingress on `catalogGrpcPodPort` (50051) |
| `bundle-unpack-egress` | pods with `olm.managed=true` + `operatorframework.io/bundle-unpack-ref` label | allow all egress |

The `olm-catalog-grpc-ingress` policy only lists `Ingress` in `policyTypes`, so it does not restrict egress of catalog pods (that remains under the per-CatalogSource policies from the controller). The `bundle-unpack-egress` policy only lists `Egress`, so it does not affect ingress on those pods.

## Test plan

- [ ] Deploy OLM via the Helm chart into a cluster with a default-deny NetworkPolicy enforcement (e.g., K3s / Rancher Desktop)
- [ ] Create a CatalogSource pointing to `quay.io/operatorhubio/catalog:latest`; verify `lastObservedState` reaches `READY`
- [ ] Create a Subscription; verify the InstallPlan completes without `BundleLookupFailed`
- [ ] Verify `olm-catalog-grpc-ingress` and `bundle-unpack-egress` NetworkPolicies are present in the OLM namespace after `helm install`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/deploy/chart/templates/0000_50_olm_01-networkpolicies.yaml
+++ b/deploy/chart/templates/0000_50_olm_01-networkpolicies.yaml
@@ -87,11 +87,17 @@ spec:
     - { }
 ---
 # Complements per-CatalogSource NPs from the controller; covers the bootstrapping window before reconciliation.
+# Effective only when catalog_namespace == namespace (the default). When they differ, no OLM-managed
+# deny-all exists in catalog_namespace, so this rule has no effect; adding one there is unsafe since
+# OLM does not exclusively own that namespace.
+# No 'from:' restriction is intentional, matching current dynamic NP behavior. If the controller ever
+# restricts ingress sources, this Helm-owned NP (which the controller cannot delete) will silently
+# override that — treat as a permanent design constraint.
 apiVersion: networking.k8s.io/v1
 kind: NetworkPolicy
 metadata:
-  name: catalog-source-grpc-server
-  namespace: {{ .Values.namespace }}
+  name: olm-catalog-grpc-ingress
+  namespace: {{ .Values.catalog_namespace }}
 spec:
   podSelector:
     matchExpressions:
@@ -105,11 +111,12 @@ spec:
         port: {{ .Values.catalogGrpcPodPort }}
 ---
 # Wildcard egress; API server port omitted intentionally — it is implementation-defined and not statically knowable.
+# Carries the same split-namespace limitation as olm-catalog-grpc-ingress above.
 apiVersion: networking.k8s.io/v1
 kind: NetworkPolicy
 metadata:
   name: bundle-unpack-egress
-  namespace: {{ .Values.namespace }}
+  namespace: {{ .Values.catalog_namespace }}
 spec:
   podSelector:
     matchExpressions:
--- a/deploy/chart/templates/0000_50_olm_01-networkpolicies.yaml
+++ b/deploy/chart/templates/0000_50_olm_01-networkpolicies.yaml
@@ -85,3 +85,41 @@ spec:
     - { }
   egress:
     - { }
+---
+# Complements per-CatalogSource NPs from the controller; covers the bootstrapping window before reconciliation.
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: catalog-source-grpc-server
+  namespace: {{ .Values.namespace }}
+spec:
+  podSelector:
+    matchExpressions:
+      - key: olm.catalogSource
+        operator: Exists
+  policyTypes:
+    - Ingress
+  ingress:
+    - ports:
+      - protocol: TCP
+        port: {{ .Values.catalogGrpcPodPort }}
+---
+# Wildcard egress; API server port omitted intentionally — it is implementation-defined and not statically knowable.
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: bundle-unpack-egress
+  namespace: {{ .Values.namespace }}
+spec:
+  podSelector:
+    matchExpressions:
+      - key: operatorframework.io/bundle-unpack-ref
+        operator: Exists
+      - key: olm.managed
+        operator: In
+        values:
+          - "true"
+  policyTypes:
+    - Egress
+  egress:
+    - { }


deploy/chart/templates/0000_50_olm_01-networkpolicies.yaml
```diff
@@ -85,3 +85,41 @@ spec:
     - { }
   egress:
     - { }
+---
+# Complements per-CatalogSource NPs from the controller; covers the bootstrapping window before reconciliation.
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: catalog-source-grpc-server
```
perdasilva: **NP name collision risk**

The dynamic NP uses `fmt.Sprintf("%s-grpc-server", catalogSource.GetName())` (see `pkg/controller/registry/reconciler/helpers.go:24`). A CatalogSource named `catalog-source` in `catalog_namespace` would produce a dynamic NP named `catalog-source-grpc-server` — identical to this static Helm NP.

Since they have different specs (`matchExpressions` vs `matchLabels`, different labels/ownerReferences), the controller will overwrite the static NP on reconciliation and Helm will fight back on upgrade, causing reconciliation churn.

Consider prefixing the static NP name, e.g. `olm-static-catalog-source-grpc-server` or similar, to avoid any possible collision with the `<name>-grpc-server` naming pattern.

deploy/chart/templates/0000_50_olm_01-networkpolicies.yaml
```diff
@@ -85,3 +85,41 @@ spec:
...
+  namespace: {{ .Values.catalog_namespace }}
```
perdasilva, line 100: **NPs become no-ops when `catalog_namespace != namespace`**

The `default-deny-all-traffic` NP (line 5) is created only in `{{ .Values.namespace }}`. These new NPs target `{{ .Values.catalog_namespace }}`. When these values differ, Kubernetes allows all traffic by default in `catalog_namespace` — the allow-style NPs grant nothing because there is no deny-all baseline to punch through.

The bootstrap gap these NPs are meant to close remains open in split-namespace configurations.

Options:
- Add a `default-deny-all-traffic` NP in `catalog_namespace` when it differs from `namespace`
- Or document this as a known limitation

deploy/chart/templates/0000_50_olm_01-networkpolicies.yaml
```diff
@@ -85,3 +85,41 @@ spec:
     - { }
   egress:
     - { }
+---
+# Complements per-CatalogSource NPs from the controller; covers the bootstrapping window before reconciliation.
```
perdasilva, line 89: **Static broad-selector NPs persist permanently and cannot be narrowed by the controller**

This NP allows TCP ingress on `catalogGrpcPodPort` from **any** source (no `from` restriction) for **all** pods with the `olm.catalogSource` label (`Exists` match). Since Kubernetes NetworkPolicies are additive (union semantics), if the controller later tightens its per-CatalogSource NPs to restrict ingress to specific source pods (e.g., only from `catalog-operator`), this static NP silently defeats that restriction — and the controller has no code path to delete or update Helm-managed resources.

This is acceptable if the ingress source is intentionally unrestricted long-term. Worth noting as a design constraint.

perdasilva (commented):
Code review — 3 findings on the new NetworkPolicy templates.

deploy/chart/templates/0000_50_olm_01-networkpolicies.yaml
grokspawn: The only way to truly de-fang this challenge is to use a unique construction, which would make an accidental overlap less likely but do nothing to prevent a bad actor from seeing the public construction details and manufacturing a conflict.  

So how likely do we feel this would be?  Do we think that we could dump the resources on the cluster and see that there's a catalogsource called 'catalog-source' and determine the network policies are impacting?  

I don't want to underplay this, but I'd like to understand the goal here.  Just using a more complex or static name does not avoid the issue.

deploy/chart/templates/0000_50_olm_01-networkpolicies.yaml
grokspawn: Documented.  I'm not sure we can safely splat a default-deny-all-traffic NP in the other case, since we can't assume we own it.

deploy/chart/templates/0000_50_olm_01-networkpolicies.yaml
grokspawn: Switched the name to see if we can merge this, but it feels like a potential whack-a-mole game.

rashmigottipati (approved):
/approve
/lgtm

PR #3867, opened 2026-07-13
OCPBUGS-98575: watch APIServer TLS profile and restart on change

tmshort:
When neither --tls-min-version nor --tls-cipher-suites is supplied via flags, packageserver now sets up a watch on APIServer.config.openshift.io/cluster after applying the initial TLS profile at startup. If the profile changes, the process exits so it restarts with the updated settings — the same restart pattern used for OLMConfig interval changes.

The watch is intentionally skipped when either TLS flag is present (i.e. flags are injected externally, e.g. by PSM). In that case the external manager is responsible for triggering restarts on profile changes, avoiding unnecessary restarts. Partial-flag configurations (e.g. --tls-cipher-suites only) still get the cluster default applied for the unset field, but the watch is also skipped to avoid an inconsistent comparison baseline.

### DIFF
--- a/pkg/package-server/server/server.go
+++ b/pkg/package-server/server/server.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"net"
 	"os"
+	"slices"
 	"time"
 
 	log "github.com/sirupsen/logrus"
@@ -23,7 +24,9 @@ import (
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/client-go/util/workqueue"
 
+	apiconfigv1 "github.com/openshift/api/config/v1"
 	configv1client "github.com/openshift/client-go/config/clientset/versioned"
+	configinformers "github.com/openshift/client-go/config/informers/externalversions"
 	libcrypto "github.com/openshift/library-go/pkg/crypto"
 	operatorsv1 "github.com/operator-framework/api/pkg/operators/v1"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/client"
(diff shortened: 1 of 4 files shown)


pkg/package-server/server/server.go
```diff
@@ -347,6 +391,42 @@ func (op *Operator) syncOLMConfig(obj interface{}) error {
...
+// handler is never registered.
+func (op *Operator) syncAPIServerTLSProfile(obj interface{}) error {
+	apiServer, ok := obj.(*apiconfigv1.APIServer)
+	if !ok {
+		return fmt.Errorf("casting APIServer failed")
+	}
+
+	newMinVersion, newCipherSuites := olmapiserver.GetSecurityProfileConfig(apiServer.Spec.TLSSecurityProfile)
+	newMinVersionStr := libcrypto.TLSVersionToNameOrDie(newMinVersion)
+	newCipherNames := libcrypto.CipherSuitesToNamesOrDie(newCipherSuites)
```
perdasilva: **Panic risk in watch handler:** `TLSVersionToNameOrDie` and `CipherSuitesToNamesOrDie` both call `panic()` when the input value is not in their known maps (verified in `vendor/github.com/openshift/library-go/pkg/crypto/crypto.go`). The same functions are used at startup in `applyClusterTLSProfileWithClients`, but a panic there fails the process once and is caught by the pod restart policy. Here, this handler fires on **every APIServer CR update**, so if the cluster admin changes to a `Custom` profile that contains a cipher suite or TLS version not yet in library-go's map, the handler will panic on each reconcile, causing a crash loop rather than a clean error.

Consider wrapping with a `recover`, or switching to returning-an-error variants (e.g. `TLSVersionToName`/`CipherSuiteToName`) so the handler can return an error that the queue will log and retry safely:

```go
newMinVersionStr, err := libcrypto.TLSVersionToName(newMinVersion)
if err != nil {
    return fmt.Errorf("unknown TLS version %d in APIServer profile: %w", newMinVersion, err)
}
```

pkg/package-server/server/server.go
```diff
@@ -347,6 +391,42 @@ func (op *Operator) syncOLMConfig(obj interface{}) error {
...
+	slices.Sort(newSorted)
+	currentSorted := slices.Clone(currentCipherNames)
+	slices.Sort(currentSorted)
+
+	if newMinVersionStr == currentMinVersion && slices.Equal(newSorted, currentSorted) {
+		return nil
+	}
+
+	log.Warnf("APIServer TLS profile changed (minVersion: %s → %s, ciphers: %v → %v), restarting to apply new profile",
+		currentMinVersion, newMinVersionStr, currentCipherNames, newCipherNames)
```
perdasilva: **Log shows unsorted ciphers, comparison uses sorted:** The restart decision is correctly made on sorted copies (`newSorted`/`currentSorted`), but the `Warnf` on the next line logs the original unsorted slices (`currentCipherNames`, `newCipherNames`). When a restart fires, an operator reading the log sees the raw cipher order, which may differ from how the sorted comparison actually saw them. Logging the sorted versions makes the log easier to reconcile against the comparison:

```suggestion
	log.Warnf("APIServer TLS profile changed (minVersion: %s → %s, ciphers: %v → %v), restarting to apply new profile",
		currentMinVersion, newMinVersionStr, currentSorted, newSorted)
```

tmshort:
May not be needed.

tmshort:
Solved with an alternative approach.

PR #2919, opened 2023-02-06
OCPBUGS-672: Catalog Pod Startup Probe Timeout

dtfranz:
**Description of the change:**

Updates the Startup Probe timeoutSeconds for catalog pods to be consistent with the timeouts in the readiness and liveness probes as the default of 1s can be too low in some cases. Also adjusted the failure threshold in order to maintain the ~150s startup time failure as originally designed (timeout of 5s + period of 10s = 15s per attempt * 10 failures = 150s).

**Motivation for the change:**

OCPBUGS-672

PR #3876, opened 2026-07-22
OCPBUGS-78095: Restore registry-server startup probe FailureThreshold to 15

emmahone:

Restores \`StartupProbe.FailureThreshold\` for the registry-server container from \`10\` back to \`15\`, returning the startup budget from 100s to 150s. This reverts the threshold portion of 401bfff4df20c97bef08db2dc9fc75bc1879be7a, which changed FailureThreshold from 15 to 10 while adding TimeoutSeconds.

Test fixtures updated to reflect the recomputed \`olm.pod-spec-hash\` values.

The 100s budget (\`FailureThreshold: 10 × PeriodSeconds: 10\`) can be too tight when there are performance constraints or large custom CatalogSources (OCPBUGS-78095). The commit message in 401bfff states the intent was to "maintain the ~150s startup time failure as originally designed," suggesting the reduction from 15 to 10 was unintentional. This change restores \`FailureThreshold\` to \`15\` (150s budget), which provides sufficient margin for these scenarios without being excessive.

**Architectural changes:**

None.

## Summary by CodeRabbit

* **Bug Fixes**
  * Increased the startup probe tolerance for registry pods, helping prevent premature health-check failures during startup.

### DIFF
--- a/pkg/controller/registry/reconciler/reconciler.go
+++ b/pkg/controller/registry/reconciler/reconciler.go
@@ -180,7 +180,7 @@ func Pod(source *operatorsv1alpha1.CatalogSource, name, opmImg, utilImage, img s
 								Command: []string{"grpc_health_probe", "-addr=:50051", "-rpc-timeout=5s"},
 							},
 						},
-						FailureThreshold: 10,
+						FailureThreshold: 15,
 						PeriodSeconds:    10,
 						TimeoutSeconds:   5,
 					},
--- a/pkg/controller/registry/reconciler/reconciler.go
+++ b/pkg/controller/registry/reconciler/reconciler.go
@@ -180,7 +180,7 @@ func Pod(source *operatorsv1alpha1.CatalogSource, name, opmImg, utilImage, img s
 								Port: 50051,
 							},
 						},
-						FailureThreshold: 10,
+						FailureThreshold: 15,
 						PeriodSeconds:    10,
 						TimeoutSeconds:   5,
 					},
--- a/pkg/controller/registry/reconciler/reconciler_test.go
+++ b/pkg/controller/registry/reconciler/reconciler_test.go
@@ -38,7 +38,7 @@ func TestPodMemoryTarget(t *testing.T) {
 				ObjectMeta: metav1.ObjectMeta{
 					GenerateName: "test-",
 					Namespace:    "testns",
-					Labels:       map[string]string{"olm.pod-spec-hash": "KpDeU9UesY5GlqG7ZUg9deo8Jp7mEGdpBhGJn", "olm.managed": "true"},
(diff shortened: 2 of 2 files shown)


pkg/controller/registry/reconciler/reconciler.go
```diff
@@ -180,7 +180,7 @@ func Pod(source *operatorsv1alpha1.CatalogSource, name, opmImg, utilImage, img s
 								Command: []string{"grpc_health_probe", "-addr=:50051", "-rpc-timeout=5s"},
 							},
 						},
-						FailureThreshold: 10,
+						FailureThreshold: 15,
```
wking, line 183: > Restores `StartupProbe.FailureThreshold` for the registry-server container from `10` back to `15`, returning the startup budget from 100s to 150s...

But [in][1] #2919:

> Also adjusted the failure threshold in order to maintain the ~150s startup time failure as originally designed (timeout of 5s + period of 10s = 15s per attempt * 10 failures = 150s).

So maybe you're going up to timeout of 5s + period of 10s = 15s per attempt * 15 failures = 225s)?  I'm not clear on what "as originally designed" is talking about though, so not clear what the downsides would be of going too high here.

[1]: https://github.com/operator-framework/operator-lifecycle-manager/pull/2919#issue-1573313710

emmahone: Today with the probe interval of 10s and failure threshold of 10, my understanding is the 10th probe is sent out at the 90s mark and times out at 5s. So technically the time to failure is 95s. After this change the time to failure would be 145s. 

Versus what occurred prior to 401bfff where the last probe was sent out at 140s and times out based on the default timeout.

EDIT: Just to clarify, 'as originally designed' is just quoted from the commit message in 401bfff. The tuning of these probe interval and failure threshold is fairly arbitrary which is why I am pushing for increasing them slightly rather than simply telling customers to fix the underlying issues causing these issues and ignoring these values. This gives us some resiliency with a small cost in time to failure. 

-mahoney

tmshort:
@emmahone is there a Red Hat Jira ticket associated with this change?

emmahone:
@tmshort This PR is associate with https://redhat.atlassian.net/browse/OCPBUGS-78095. Its already added in jira as a a web link. Do I need to do anything differently?

tmshort:
@emmahone I put it as part of the title.

tmshort:
@CodeRabbit review

tmshort:
And... now this needs a rebase @emmahone

grokspawn:
To be clear, this PR doesn't resolve the bug, but it does correct what looks like a mistake on the math of the original intent.  In the cases where the probe failures are pathological, the probe duration is immaterial.

perdasilva:
> To be clear, this PR doesn't resolve the bug, but it does correct what looks like a mistake on the math of the original intent. In the cases where the probe failures are pathological, the probe duration is immaterial.

@emmahone I've pushed this through but please keep in mind the comment above

PR #3572, opened 2025-05-07
:seedling: Disable FailForward bad resource e2e

perdasilva:
**Description of the change:**

While investigating the FailForward e2e test flakyness, I've discovered we have a hot loop in the CSV state graph where non-terminal errors seem to be retried forever. For the FF e2e, there's a bundle with a bad deployment name. This error was not being marked as terminal. While the csv hot looped: Pending -> InstallReady -> Failed -> Pending -> ..., the test detected the failed state and created the catalog with the fixed bundle, which created the flake as the resolved would fail resolution due to non-referenced csvs =S

Bug has been filed here: #3573

**Motivation for the change:**

**Architectural changes:**

**Testing remarks:**

**Reviewer Checklist**
- [ ] Implementation matches the proposed design, or proposal is updated to match implementation
- [ ] Sufficient unit test coverage
- [ ] Sufficient end-to-end test coverage
- [ ] Bug fixes are accompanied by regression test(s)
- [ ] e2e tests and flake fixes are accompanied evidence of flake testing, e.g. executing the test 100(0) times
- [ ] tech debt/todo is accompanied by issue link(s) in comments in the surrounding code
- [ ] Tests are comprehensible, e.g. Ginkgo DSL is being used appropriately
- [ ] Docs updated or added to `/doc`
- [ ] Commit messages sensible and descriptive
- [ ] Tests marked as `[FLAKE]` are truly flaky and have an issue
- [ ] Code is properly formatted

PR #3878, opened 2026-07-27
fix: classify invalid strategy component errors as unrecoverable

### LINKED ISSUE #3573: CSV Install/Upgrade hot loop on bad bundle resource
## Bug Report

There is a hot loop in the CSV state machine. If the bundle contains a malformed/invalid resource, e.g. if the deployment name is invalid, the CSV will go from Pending -> InstallReady -> Failed -> Pending ...

This was the cause of the flakyness seem in [failforward e2e](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/test/e2e/fail_forward_e2e_test.go#L324).

PR #3572 disables this test until we can fix this.

pujitha24:
Motivation:
When a CSV's install strategy contains a permanently malformed
component (e.g. a Deployment with an invalid name), the Kubernetes API
server rejects the create/update with an "Invalid" (422) error. The
deployment installer returned this error unclassified, so
install.IsErrorUnrecoverable() treated it as retryable. The CSV state
machine in operator.go then cycled InstallReady -> Failed
(CSVReasonComponentFailed, retryable) -> Pending
(CSVReasonNeedsReinstall) -> InstallReady forever, hot-looping instead
of settling in Failed. This was the root cause of the flakiness in the
FailForward e2e test disabled by PR #3572. The effect is wasted
reconciles, CSV phase oscillation, and event/log spam -- the CSV
controller itself does not crash or otherwise affect other CSVs.

Approach:
- Add a classifyInstallError helper in
  pkg/controller/install/deployment.go that maps apierrors.IsInvalid
  (in addition to the existing apierrors.IsForbidden) to a
  StrategyError, and apply it to both the installCertRequirements and
  installDeployments error paths in Install().
- Add StrategyErrReasonComponentInvalid in
  pkg/controller/install/errors.go and register it in
  unrecoverableErrors, so IsErrorUnrecoverable() returns true and the
  CSV is correctly marked CSVReasonComponentFailedNoRetry, which is
  terminal (the existing top-of-transitionCSVState guard in operator.go
  keeps it in Failed rather than reinstalling).

Validation:
go build ./...
go test ./pkg/controller/install/... ./pkg/controller/operators/olm/...
Both pass, including a new regression test,
TestInstallStrategyDeploymentInstallInvalidDeployment, which asserts
that an apierrors.NewInvalid error from the deployment client causes
Install() to return StrategyErrReasonComponentInvalid and that
IsErrorUnrecoverable() reports it as unrecoverable.

Note: the disabled e2e regression test in
test/e2e/fail_forward_e2e_test.go (XWhen at line ~324, referencing this
issue) requires a live cluster and was intentionally left disabled, as
it cannot be validated in this environment.

Fixes #3573

Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>

### DIFF
--- a/pkg/controller/install/errors.go
+++ b/pkg/controller/install/errors.go
@@ -9,7 +9,7 @@ const (
 	StrategyErrReasonUnknown            = "Unknown"
 	StrategyErrBadPatch                 = "PatchUnsuccessful"
 	StrategyErrDeploymentUpdated        = "DeploymentUpdated"
-	StrategyErrInsufficientPermissions  = "InsufficentPermissions"
+	StrategyErrInsufficientPermissions  = "InsufficientPermissions"
 	StrategyErrReasonComponentInvalid   = "ComponentInvalid"
 )
 
--- a/pkg/controller/install/errors.go
+++ b/pkg/controller/install/errors.go
@@ -10,6 +10,7 @@ const (
 	StrategyErrBadPatch                 = "PatchUnsuccessful"
 	StrategyErrDeploymentUpdated        = "DeploymentUpdated"
 	StrategyErrInsufficientPermissions  = "InsufficentPermissions"
+	StrategyErrReasonComponentInvalid   = "ComponentInvalid"
 )
 
 // unrecoverableErrors are the set of errors that mean we can't recover an install strategy
@@ -18,6 +19,7 @@ var unrecoverableErrors = map[string]struct{}{
 	StrategyErrReasonTimeout:           {},
 	StrategyErrBadPatch:                {},
 	StrategyErrInsufficientPermissions: {},
+	StrategyErrReasonComponentInvalid:  {},
 }
 
 // StrategyError is used to represent error types for install strategies
--- a/pkg/controller/install/deployment.go
+++ b/pkg/controller/install/deployment.go
@@ -199,18 +199,28 @@ func (i *StrategyDeploymentInstaller) Install(s Strategy) error {
 	// Install owned APIServices and update strategy with serving cert data
 	updatedStrategy, err := i.installCertRequirements(strategy)
 	if err != nil {
-		return err
+		return classifyInstallError(err)
 	}
 
 	if err := i.installDeployments(updatedStrategy.DeploymentSpecs); err != nil {
-		if apierrors.IsForbidden(err) {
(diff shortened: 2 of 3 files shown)


tmshort:
Note that that `classifyInstallError` now also covers the `installCertRequirements` path; this is a relatively benign behavior change.

pkg/controller/install/errors.go
```diff
@@ -10,6 +10,7 @@ const (
 	StrategyErrBadPatch                 = "PatchUnsuccessful"
 	StrategyErrDeploymentUpdated        = "DeploymentUpdated"
 	StrategyErrInsufficientPermissions  = "InsufficentPermissions"
```
tmshort: Nit: Not part of this PR, but this is misspelled.
```suggestion
	StrategyErrInsufficientPermissions  = "InsufficientPermissions"
```

joelanford:
`IsInvalid` could be returned due to a ValidatingAdmissionWebhook that waits for some precondition that OLM is unaware of (and therefore won't reconcile again when the precondition changes).

OLMv0 notoriously requires the "OLM Dance" (i.e. uninstall/reinstall) when an installation fails.

IMO, it is better to have a retry loop on `IsInvalid` than a permanent failure and the inevitable OLM dance, unless we are absolutely certain that the request was invalid because of something that will never change.

Sounds like the core problem is a hot loop though. Is there a way to fix that without affecting also forcing the CSV into the `Failed` state?

pujitha24:
Thanks for the catch, @joelanford — that's a fair point I hadn't fully weighed. Treating all `apierrors.IsInvalid` responses as unrecoverable does mean we'd also permanently fail an install that would have succeeded once a ValidatingAdmissionWebhook's precondition (unknown to OLM) is satisfied, which is worse than the hot loop in that case. I don't have a reliable way to distinguish "permanently invalid" from "webhook is waiting on something" purely from the error, so I'd rather not guess at that distinction myself.

Would you prefer I (a) narrow this change to just break the specific hot loop from #3573 while leaving `IsInvalid` retryable otherwise, or (b) look at fixing the hot loop itself (e.g. backoff/dedup in the CSV reconcile path) instead of changing error classification at all? Happy to go either direction once you have a preference.

tmshort:
@pujitha24 is there a Red Har Jira ticket associated with this change?

grokspawn:
closing this since OLMv0 is feature-complete, there are known workarounds, and this is a high potential here for side-effects.

PR #3883, opened 2026-07-29
OCPBUGS-82532: Skip adoption of missing or terminal generated-by installplans

haklein:
`ensureSubscriptionInstallPlanState` adopts the installplan named by the `olm.generated-by` annotation whenever a subscription has no installplan reference. That is only valid while the generating plan is still in progress. Once the plan has been GC'd or has reached a terminal phase, adoption wedges the namespace:

- plan missing: the `NotFound` error aborts `syncResolvingNamespace` before resolution, so no installplan is ever created again for any subscription in the namespace.
- plan complete: the adoption resets the subscription to `UpgradePending` with `currentCSV=startingCSV`, which no longer exists after upgrades. The change is never persisted and the sync bails out before resolution on every iteration.

Both are hit by deleting an unapproved installplan (a documented way to get a plan for a newer version when one is pending), on any namespace with dependency-generated subscriptions, e.g. ODF.

Treat a missing plan as "nothing to adopt" and skip adoption of plans in a terminal phase so resolution can create a fresh installplan. Failed plans are still adopted when fail-forward is enabled, which depends on the subscription referencing them.

Fixes OCPBUGS-82532

**Architectural changes:**

n/a

**Testing remarks:**

Unit tests cover the full decision table of the adoption path:

  | generated-by plan | fail-forward | expected |
  |---|---|---|
  | missing | - | no error, no adoption, resolution proceeds |
  | exists, in progress | - | adopted (existing behavior) |
  | lookup fails (non-404) | - | sync error propagated (existing behavior) |
  | exists, Complete | - | not adopted, resolution proceeds |
  | exists, Failed | off | not adopted, resolution proceeds |
  | exists, Failed | on | adopted, subscription marked Failed (existing behavior) |

The two "not adopted"/"no error" rows are regression tests and fail against the previous code.

**Reviewer Checklist**

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved InstallPlan reconciliation when annotated plans are missing, preventing unnecessary namespace resolution failures.
  * Safely handles completed and failed plans by verifying the subscription’s starting version before adoption.
  * Continues reporting unexpected API errors while ignoring expected missing-resource responses.

* **Tests**
  * Added coverage for InstallPlan adoption, fail-forward behavior, no-op scenarios, and transient API or version lookup errors.

### DIFF
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -1668,10 +1668,40 @@ func (o *Operator) ensureSubscriptionInstallPlanState(logger *logrus.Entry, sub 
 
 	ip, err := o.client.OperatorsV1alpha1().InstallPlans(sub.GetNamespace()).Get(context.TODO(), ipName, metav1.GetOptions{})
 	if err != nil {
+		if apierrors.IsNotFound(err) {
+			// The annotated installplan may have been GC'd. Not an error:
+			// there is nothing to adopt, and returning an error here blocks
+			// resolution for the whole namespace.
+			logger.WithField("installplan", ipName).Debug("generating installplan no longer exists, skipping adoption")
+			return sub, false, nil
+		}
 		logger.WithField("installplan", ipName).Warn("unable to get installplan from cache")
 		return nil, false, err
 	}
 	logger.WithField("installplan", ipName).Debug("found installplan that generated subscription")
+
+	// A plan in a terminal phase is only worth adopting right after it
+	// created this subscription, while spec.startingCSV still names an
+	// existing CSV: adoption then links the subscription to the CSV the
+	// plan installed. Once that CSV is gone the plan is a historical
+	// record, and adopting it would reset the subscription to
+	// UpgradePending with a currentCSV that no longer exists, blocking
+	// resolution. Failed plans are still adopted when fail-forward is
+	// enabled, which depends on the subscription referencing them.
+	if ip.Status.Phase == v1alpha1.InstallPlanPhaseComplete || (ip.Status.Phase == v1alpha1.InstallPlanPhaseFailed && !failForwardEnabled) {
+		startingCSVExists := false
+		if sub.Spec.StartingCSV != "" {
+			_, err := o.client.OperatorsV1alpha1().ClusterServiceVersions(sub.GetNamespace()).Get(context.TODO(), sub.Spec.StartingCSV, metav1.GetOptions{})
+			if err != nil && !apierrors.IsNotFound(err) {
+				return nil, false, err
+			}
+			startingCSVExists = err == nil
+		}
+		if !startingCSVExists {
+			logger.WithField("installplan", ipName).Debug("generating installplan in terminal phase and startingCSV not present, skipping adoption")
+			return sub, false, nil
+		}
+	}
 
 	out := sub.DeepCopy()
 	ref, err := reference.GetReference(ip)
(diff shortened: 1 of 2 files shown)


tmshort:
/approve
(Assuming CI passes)

joelanford:
@haklein, the OCPBUGS issue mentions ODF. However ODF does some funky stuff where it creates more Subscriptions in `Manual` approval mode and then has its own logic for automatically approving those manual subscriptions. I don't know if that functionality of ODF is at play with this bug, but this seems like a rather nuanced change in an area of the codebase that even the current SMEs lack deep understanding of the blast radius.

Before we move forward with this, I would ask that we try to reproduce with:
- a "simple" operator that doesn't have dependencies
- another operator that _does_ have dependencies, but not named ODF :)

joelanford:
/hold

For the above request.

tmshort:
/approve cancel
The e2e tests have been failing, so revoking approval.

haklein:
@joelanford thanks a lot for your feedback. Please bear with the slow reply, I was on PTO last week.

> the OCPBUGS issue mentions ODF. However ODF does some funky stuff where it creates more Subscriptions in `Manual` approval mode and then has its own logic for automatically approving those manual subscriptions. 

I've deployed ODF 4.20 in a lab, and so far the approval mode was automatic out of the box. We do have an ODF doc section "Changing the update approval strategy" which shows how to change to manual. So that's likely done by customers on numerous ODF installations

> I don't know if that functionality of ODF is at play with this bug, but this seems like a rather nuanced change in an area of the codebase that even the current SMEs lack deep understanding of the blast radius.

This does reproduce outside of ODF, see below

> 
> Before we move forward with this, I would ask that we try to reproduce with:
> 
>     * a "simple" operator that doesn't have dependencies

This doesn't shop up with a simple operator, only if one depends on another and when OLM itself creates that subscription. Tested with a dummy operator, and replacing the versions v0.1.0 to v0.4.0 (via replaces chain). That didn’t reproduce, as it was missing the `olm.generated-by` annotation. This annotation only gets added by dependency created subscriptions. 

> 
>     * another operator that _does_ have dependencies, but not named ODF :)

Results from a test with two operators, one depending on the other. testop v0.5.0 declares an olm.package dependency on testdep v0.1.0. A second catalog tag adds testop v0.6.0 and testdep v0.2.0, so that one pending InstallPlan references both subscriptions.

1. Create a subscription for testop v0.5.0 with `installPlanApproval: Manual`
2. Approve the InstallPlan. OLM installs both operators. OLM creates the testdep subscription and writes `olm.generated-by` on it
3. Change the catalog to the tag with v0.6.0. OLM creates a pending InstallPlan for both packages
4. Delete the InstallPlan that the annotation references. On longer running clusters, GC would have done this step (when an unapproved plan was lingering around while multiple z releases were released)
5. Delete the pending InstallPlan

Result: OLM does not create a new InstallPlan again. The subscriptions stay without a plan reference. The catalog-operator log shows this loop:
```
level=warning msg="unable to get installplan from cache" installplan=install-mtjsd sub=testdep-stable-testop-catalog-repro-dep
level=info msg="error ensuring installplan state: installplans.operators.coreos.com \"install-mtjsd\" not found"
```
As soon as the `olm.generated-by` is removed, a new plan is created and executed.

haklein:
A new dependency subscription syncs after its generating installplan completes, and it needs adoption of that installplan to link to its installed CSV. The first version skipped all terminal InstallPlans, so these subscriptions failed with `ConstraintsNotSatisfiable`. The guard now skips a terminal InstallPlan only when `spec.startingCSV` no longer names an existing CSV (which is a failure condition of OCPBUGS-82532)
 I added a unit test for the adoption case and rebased the branch onto master.

PR #3884, opened 2026-08-03
reconciler: fix inverted log message and level for UpdateNotReadyErr

pdudley:

When `ensureUpdatePod` returns `UpdateNotReadyErr` — the expected, benign signal that a new update pod hasn't yet reported ready — `EnsureRegistryServer` logged it at `level=error` with the message `"ensure update pod error is not of type UpdateNotReadyErr"`. The `ok` branch fires precisely when the error *is* `UpdateNotReadyErr`, so the message was the exact opposite of what happened.

Also addresses review feedback from #3871:
- Use `pod.GetGenerateName()` in the ensure-pod wrap (desired pods set `GenerateName`, so `GetName()` is empty).
- Reference the catalog source name when wrapping update-pod ensure failures.
- Keep the Debug message free of the word "error" (`registry server update pod not yet ready`).

This mislabeled error was observed flooding logs in production environments where catalog registry pods have long startup times. Every reconcile tick during pod startup hit this path, producing a steady stream of misleading `level=error` messages that appeared to indicate a failed type-assertion, when the actual state was a completely normal pod-not-ready wait. This made triage and support significantly harder.

**Architectural changes:**

None. Control flow is unchanged — `UpdateNotReadyErr` is still returned as-is; only the log message/severity and error-wrap identifiers are corrected.

**Testing remarks:**

Added a regression unit test `Grpc/PollingEnabled/UpdatePodNotReady/ReturnsUpdateNotReadyErr` in `grpc_test.go` that verifies `EnsureRegistryServer` returns `UpdateNotReadyErr` (unmodified, not wrapped) when polling is enabled and an update pod exists but has not yet reported ready. All existing tests continue to pass.

---

Supersedes #3871 (original author unavailable; includes outstanding review feedback).

Credit: @emmahone for the original fix and test; @tmshort for the Debug message wording.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error reporting when registry pods cannot be created or updated.
  * Reconciliation now preserves clear readiness errors for running but not-ready update pods.
  * Error messages include relevant catalog and pod details to make troubleshooting easier.

* **Tests**
  * Added coverage for registry polling scenarios where an update pod is running but not ready.

### DIFF
--- a/pkg/controller/registry/reconciler/grpc.go
+++ b/pkg/controller/registry/reconciler/grpc.go
@@ -340,11 +340,11 @@ func (c *GrpcRegistryReconciler) EnsureRegistryServer(logger *logrus.Entry, cata
 		return pkgerrors.Wrapf(err, "error ensuring pod: %s", pod.GetName())
 	}
 	if err := c.ensureUpdatePod(logger, sa, defaultPodSecurityConfig, source); err != nil {
-		logger.WithError(err).Error("error ensuring registry server: could not ensure update pod")
 		if _, ok := err.(UpdateNotReadyErr); ok {
-			logger.WithError(err).Error("error ensuring registry server: ensure update pod error is not of type UpdateNotReadyErr")
+			logger.WithError(err).Debug("error ensuring registry server: update pod not yet ready")
 			return err
 		}
+		logger.WithError(err).Error("error ensuring registry server: could not ensure update pod")
 		return pkgerrors.Wrapf(err, "error ensuring updated catalog source pod: %s", pod.GetName())
 	}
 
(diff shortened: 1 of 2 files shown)


rashmigottipati (commented):
/ok-to-test

tmshort:
@pdudley @emmahone is there a Red Hat Jira ticket associated with this fix?

emmahone:
No jira associated. This bug was found while investigating a separate issue. If we need to make an OCPBUGS for this let us know and we can create that.

tmshort:
@CodeRabbit review
Was hoping that a close/reopen would trigger CodeRabbit

rashmigottipati (approved):

PR #3891, opened 2026-08-11
Fix infinite loop on cyclic CSV replacement chains

haklein:

Guard the CSV replace finder (`pkg/lib/csv/replace_finder.go`) against self-references and cycles:

- `IsBeingReplaced` skips the input CSV.
- `IsReplacing` returns nil for a self-referencing `spec.replaces` and logs a warning.
- `GetFinalCSVInReplacing` tracks visited CSVs and stops on the first repeat.

A CSV with `spec.replaces` equal to its own name sends `GetFinalCSVInReplacing` into an infinite loop. The CSV watch notification goroutine spins at 100% CPU and writes one log line per iteration. Deletion of the CSV does not stop the loop, only a pod restart does. The same walk loops on any cycle across two or more CSVs. A self-replacing CSV also deadlocks in `Pending`, because the install gate waits for the previous CSV (itself) to reach `Replacing`. After this change, a self-replacing CSV installs normally.

Fixes OCPBUGS-23954.

**Architectural changes:**

None.

**Testing remarks:**

- Regression tests in `pkg/lib/csv/replace_finder_test.go` cover the self-loop, a two-CSV cycle, and a linear chain. The looping calls hang without the fix.
- Reproduced the loop on kind with `--writePackageServerStatusName` set and a labeled self-replacing CSV: CPU 137–147%. With the fix: 0.1%, and the CSV reaches `Succeeded`. Details in OCPBUGS-23954.
- Full local runs of unit, sanity, e2e (176 passed, 0 failed) and `[FLAKE]` suite (14 passed, 0 failed).

**Reviewer Checklist**

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of CSV replacement relationships by ignoring self-replacements.
  - Replacement chains now stop safely when circular references are detected, preventing endless traversal.
  - Added warnings when replacement cycles are encountered.

- **Tests**
  - Added coverage for self-replacements, circular replacement chains, and normal multi-step replacement paths.

### DIFF
--- a/pkg/lib/csv/replace_finder.go
+++ b/pkg/lib/csv/replace_finder.go
@@ -41,6 +41,12 @@ func (r *replace) IsBeingReplaced(in *v1alpha1.ClusterServiceVersion, csvsInName
 			continue
 		}
 
+		// a CSV cannot replace itself
+		if csv.GetName() == in.GetName() {
+			r.logger.WithField("csv", in.GetName()).Debug("ignoring self-referencing spec.replaces")
+			continue
+		}
+
 		r.logger.Debugf("checking %s", csv.GetName())
 
 		if csv.Spec.Replaces == in.GetName() {
@@ -63,6 +69,12 @@ func (r *replace) IsReplacing(in *v1alpha1.ClusterServiceVersion) *v1alpha1.Clus
 		return nil
 	}
 
+	// a CSV cannot replace itself
+	if in.Spec.Replaces == in.GetName() {
+		r.logger.WithField("csv", in.GetName()).Debug("ignoring self-referencing spec.replaces")
+		return nil
+	}
+
 	// using the client instead of a lister; missing an object because of a cache sync can cause upgrades to fail
 	previous, err := r.client.OperatorsV1alpha1().ClusterServiceVersions(in.GetNamespace()).Get(context.TODO(), in.Spec.Replaces, metav1.GetOptions{})
 	if err != nil {
@@ -79,11 +91,19 @@ func (r *replace) IsReplacing(in *v1alpha1.ClusterServiceVersion) *v1alpha1.Clus
 // If the corresponding ClusterServiceVersion is not found nil is returned.
 func (r *replace) GetFinalCSVInReplacing(in *v1alpha1.ClusterServiceVersion, csvsInNamespace map[string]*v1alpha1.ClusterServiceVersion) (replacedBy *v1alpha1.ClusterServiceVersion) {
 	current := in
+	visited := map[string]struct{}{in.GetName(): {}}
 	for {
 		next := r.IsBeingReplaced(current, csvsInNamespace)
 		if next == nil {
 			break
 		}
+
+		// a cycle in the replacement chain would loop forever
+		if _, ok := visited[next.GetName()]; ok {
+			r.logger.WithField("csv", next.GetName()).Debug("cycle detected in replacement chain")
+			break
+		}
+		visited[next.GetName()] = struct{}{}
 
 		replacedBy = next
 		current = next
--- a/pkg/controller/operators/olm/operator.go
+++ b/pkg/controller/operators/olm/operator.go
@@ -2539,7 +2539,10 @@ func (a *Operator) transitionCSVState(in v1alpha1.ClusterServiceVersion) (out *v
 		// If there is a succeeded replacement, mark this for deletion
 		next := a.isBeingReplaced(out, a.csvSet(out.GetNamespace(), v1alpha1.CSVPhaseAny))
 		// Get the newest CSV in the replacement chain if fail forward upgrades are enabled.
-		if operatorGroup.UpgradeStrategy() == operatorsv1.UpgradeStrategyUnsafeFailForward {
+		// next can be nil (e.g. the replacement was deleted, or a self-referencing
+		// spec.replaces is ignored by the finder); fall through to the
+		// "no replacement CSV found" error below instead of dereferencing it.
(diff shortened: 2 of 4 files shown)


pkg/lib/csv/replace_finder.go
```diff
@@ -63,6 +68,12 @@ func (r *replace) IsReplacing(in *v1alpha1.ClusterServiceVersion) *v1alpha1.Clus
 		return nil
 	}
 
+	// a CSV cannot replace itself
+	if in.Spec.Replaces == in.GetName() {
+		r.logger.WithField("csv", in.GetName()).Warn("ignoring self-referencing spec.replaces")
```
tmshort: The new `Warn` fires 2–3 times per reconcile for any self-referencing CSV. `isReplacing` is called at three separate points within a single `CSVPhasePending` pass (operator.go lines 2233, 2251, 2302), so a permanently misconfigured CSV generates hundreds of identical Warn lines per day. Since the controller handles this case silently and no operator action is needed, consider `Debug` instead — or emit a Kubernetes Event (which the API server deduplicates) if you want operator-visible alerting.

pkg/lib/csv/replace_finder.go
```diff
@@ -79,12 +90,20 @@ func (r *replace) IsReplacing(in *v1alpha1.ClusterServiceVersion) *v1alpha1.Clus
...
+	visited := map[string]struct{}{in.GetName(): {}}
 	for {
 		next := r.IsBeingReplaced(current, csvsInNamespace)
 		if next == nil {
 			break
 		}
 
+		// a cycle in the replacement chain would loop forever
+		if _, ok := visited[next.GetName()]; ok {
+			r.logger.WithField("csv", next.GetName()).Warn("cycle detected in replacement chain")
```
tmshort: Same concern: `GetFinalCSVInReplacing` is called on every CSV reconcile event via `csv_handler.go:141`. For a mutual-cycle pair carrying the `olm.clusteroperator.name` label, a status-update write generates another informer event → another reconcile → another `Warn`, with no rate limiting or deduplication anywhere in the path. `Debug` is more appropriate here since OLM breaks the cycle gracefully.

pkg/lib/csv/replace_finder.go
```diff
@@ -41,6 +41,11 @@ func (r *replace) IsBeingReplaced(in *v1alpha1.ClusterServiceVersion, csvsInName
 			continue
 		}
 
+		// a CSV cannot replace itself
+		if csv.GetName() == in.GetName() {
```
tmshort, line 45: `IsReplacing` logs a `Warn` when it detects a self-reference, but the symmetric guard added here in `IsBeingReplaced` is silent. When diagnosing why a self-referencing CSV never transitions to `CSVPhaseReplacing`, only the `IsReplacing` Warn fires — there is no evidence this path ran. Adding even a `Debug` log here would make the two guards consistent and easier to trace.

tmshort:
Worth noting a migration concern: before this change, `IsBeingReplaced` matched a self-referencing CSV against itself and returned it (non-nil), causing `checkReplacementsAndUpdateStatus` to move it to `CSVPhaseReplacing`. Any cluster already running with such a CSV in `CSVPhaseReplacing` will see different behavior after upgrading OLM:

- `isReplacing` now returns nil (new self-ref guard), so the leaf-check at operator.go:2534 no longer fires.
- `isBeingReplaced` also returns nil, so `next` is nil.
- Without `UnsafeFailForward`: the CSV enters a perpetual `syncError = "marked as replacement, but no replacement CSV found"` loop with no recovery path.
- With `UnsafeFailForward`: operator.go:2543 calls `next.GetNamespace()` on nil → panic.

The nil-before-FailForward dereference at line 2543 is a pre-existing latent bug, but this change makes it reachable in practice. A nil guard there (and/or a note in the commit message about the migration impact) would harden the fix.

pkg/lib/csv/replace_finder.go
haklein: I will change that to debug

haklein:
> Worth noting a migration concern: before this change, `IsBeingReplaced` matched a self-referencing CSV against itself and returned it (non-nil), causing `checkReplacementsAndUpdateStatus` to move it to `CSVPhaseReplacing`. Any cluster already running with such a CSV in `CSVPhaseReplacing` will see different behavior after upgrading OLM

Thanks a lot for catching this. This reproduced exactly as you've described:
1. Deploy stock OLM v0.26.0 (pre-fix)
2. Install a CSV and wait for Succeeded
3. Patch spec.replaces to the CSV's own name. OLM moves the CSV to Replacing
4. Upgrade the olm-operator image to the nil guarded one. With UnsafeFailForward on the OperatorGroup, the operator panics at operator.go:2543 on the first sync and enters CrashLoopBackOff. The rollout never completes, so the OLM upgrade hangs.

Per your recommendation I would add:
- a nil guard at the fail-forward branch: if next != nil && ... A nil next falls through to "no replacement CSV found" error, as with the default strategy.
- Two regression cases in TestTransitionCSVFailForward: a self-referencing CSV in Replacing, with fail-forward on and off. The fail-forward case panics without the guard.

with nil guard the upgraded operator stays healthy on the test setup. CSV stays in Replacing with the "no replacement CSV found" sync error, which can be resolved by deleting the CSV. I will add a  note of this migration behavior in the commit message.

@tmshort Shall I squash before merging the new commits?

tmshort:
@haklein 
> @tmshort Shall I squash before merging the new commits?

This is small enough it doesn't matter; AFAICT github will squash upon merge 

EDIT: Although I'd be OK with a squash

haklein:
@tmshort force pushed the squashed commit with the changes as requested, br Hari

PR #3892, opened 2026-08-13
fix(olm): fix ConversionWebhook spec.conversion lifecycle during CSV upgrades

ugiordan:
### 1. spec.conversion written before pods are ready (upgrade race)

During an OLM-managed upgrade, `spec.conversion` is written to CRDs inside `installDeployments()` before any pod from the new deployment is scheduled or ready.

Once `spec.conversion` is set, the apiserver routes conversion calls to the new webhook service endpoint. Since no new pod is serving `/convert` yet, those calls return HTTP 404. This breaks CRD version conversion mid-upgrade.

### 2. spec.conversion not cleared when replacement CSV drops the ConversionWebhook

`handleClusterServiceVersionDeletion` returned unconditionally when any replacement CSV was found, assuming the replacement would manage `spec.conversion` going forward. If the replacement dropped the ConversionWebhook entirely, `spec.conversion` was left pointing at the now-deleted service. All CR conversion requests then fail with connection refused.

### Fix 1: defer spec.conversion write until deployment is ready

Skip `ConversionWebhook` descriptors in `createOrUpdateCertResourcesForDeployment()` so that `spec.conversion` is never written during `Install()`.

Add `EnsureConversionWebhooks()` on `*StrategyDeploymentInstaller` and call it from `areWebhooksAvailable()`, which is only invoked from `updateInstallStatus()` after `CheckInstalled()` confirms the deployment's pods are ready. Gate `areWebhooksAvailable()` behind `strategyInstalled && strategyErr == nil` so `EnsureConversionWebhooks()` is never called before readiness is confirmed.

#### Why not extend the StrategyInstaller interface?

`EnsureConversionWebhooks()` is intentionally a concrete method on `*StrategyDeploymentInstaller` rather than an interface method. Adding it to `StrategyInstaller` would require updating `NullStrategyInstaller` and all generated counterfeiter fakes. The `areWebhooksAvailable()` call site type-asserts to `*StrategyDeploymentInstaller` before calling the method — a safe assert since `NullStrategyInstaller` never has `ConversionWebhook` entries.

### Fix 2: clear spec.conversion for CRDs dropped by the replacement CSV

Instead of returning unconditionally when a replacement CSV is found, build the set of CRDs still covered by a ConversionWebhook in the replacement. Only reset `spec.conversion` to `NoneConverter` for CRDs the new CSV dropped. CRDs the replacement still covers are left intact so in-flight conversion calls keep working during a normal upgrade.

Also adds a nil guard on `crd.Spec.Conversion` before writing to it, fixing a latent panic in the no-replacement path.

## Files changed

- `pkg/controller/install/deployment.go`: skip ConversionWebhook in `createOrUpdateCertResourcesForDeployment()`, add `EnsureConversionWebhooks()`
- `pkg/controller/operators/olm/apiservices.go`: add `installer` param to `areWebhooksAvailable()`, call `EnsureConversionWebhooks()` before checking CRD state
- `pkg/controller/operators/olm/operator.go`: gate `areWebhooksAvailable()` behind `strategyInstalled && strategyErr == nil`; fix `handleClusterServiceVersionDeletion` to only clear `spec.conversion` for CRDs the replacement CSV dropped

## Related

Downstream: https://github.com/openshift/operator-framework-olm/pull/1348

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved installation reliability by applying conversion webhook settings after deployment readiness is confirmed.
  - Installation status checks now validate webhook availability at the appropriate stage and report failures more clearly.
  - Cleanup preserves conversion webhooks still required by replacement components and removes only obsolete configurations.
  - Cleanup now safely handles missing conversion settings and stops when required resource information cannot be retrieved.


tmshort (commented):
Adding inline review comments...

tmshort (commented):
There's a lack of updates to `*_test.go` files, meaning there aren't any tests to validate this change directly. The comments below come from  claude review.

pkg/controller/operators/olm/apiservices.go
```diff
@@ -593,6 +597,14 @@ func (a *Operator) areWebhooksAvailable(csv *v1alpha1.ClusterServiceVersion) (bo
 			}
 			webhookCount = len(webhookList.Items)
 		case v1alpha1.ConversionWebhook:
+			// Write spec.conversion on each target CRD now that the deployment is confirmed
+			// ready. This is deferred from Install() to prevent routing conversion calls to
+			// pods that are not yet serving /convert.
+			if sdi, ok := installer.(*install.StrategyDeploymentInstaller); ok {
+				if err := sdi.EnsureConversionWebhooks(); err != nil {
```
tmshort: The type assertion to `StrategyDeploymentInstaller` silently fails if installer is a different `StrategyInstaller` implementation type. When `ok=false`, `EnsureConversionWebhooks()` is skipped without error indication. The subsequent check at line 614 verifies `crd.Spec.Conversion` (which was never written) and fails with generic "conversionWebhook not ready" error. The CSV hangs indefinitely with no indication of the type mismatch.

**Recommendation**: Add explicit error handling or document that only StrategyDeploymentInstaller types are supported.

pkg/controller/operators/olm/operator.go
```diff
@@ -2614,7 +2638,14 @@ func (a *Operator) updateInstallStatus(csv *v1alpha1.ClusterServiceVersion, inst
 	}
 
 	apiServicesInstalled, apiServiceErr := a.areAPIServicesAvailable(csv)
-	webhooksInstalled, webhookErr := a.areWebhooksAvailable(csv)
+	// Only attempt to write spec.conversion once the deployment is confirmed ready.
+	// areWebhooksAvailable calls EnsureConversionWebhooks, so calling it when
+	// strategyInstalled is false would recreate the upgrade race we are fixing.
+	webhooksInstalled := false
+	var webhookErr error
+	if strategyInstalled && strategyErr == nil {
```
tmshort: The condition `strategyErr == nil` may be too restrictive. If deployment is successfully installed (`strategyInstalled=true`) but `CheckInstalled()` returns a transient error (`strategyErr!=nil`), then `areWebhooksAvailable()` is skipped entirely, preventing `EnsureConversionWebhooks()` from executing. CRD conversion specs remain unwritten, and the CSV reaches Succeeded phase without functional conversion webhooks.

**Recommendation**: Review whether `strategyErr` should be checked here, or if condition should only verify `strategyInstalled`.

pkg/controller/operators/olm/apiservices.go
ugiordan: Fixed. The type assertion now returns an explicit error instead of silently skipping `EnsureConversionWebhooks()`

pkg/controller/operators/olm/operator.go
ugiordan: You're right to flag it. I've dropped the `strategyErr == nil` guard. `CheckInstalled` never returns (true, non-nil error), so it was redundant. I've added a comment explaining that invariant so it's clear why the condition is strategyInstalled alone.

pkg/controller/operators/olm/operator.go
```diff
@@ -2617,7 +2641,16 @@ func (a *Operator) updateInstallStatus(csv *v1alpha1.ClusterServiceVersion, inst
...
-	webhooksInstalled, webhookErr := a.areWebhooksAvailable(csv)
+	// Only attempt to write spec.conversion once the deployment is confirmed ready.
+	// areWebhooksAvailable calls EnsureConversionWebhooks, so calling it when
+	// strategyInstalled is false would recreate the upgrade race we are fixing.
+	// Note: CheckInstalled never returns (true, non-nil error), so strategyInstalled
+	// is sufficient — no need to also gate on strategyErr.
+	webhooksInstalled := false
+	var webhookErr error
+	if strategyInstalled {
+		webhooksInstalled, webhookErr = a.areWebhooksAvailable(csv, installer)
```
perdasilva, line 2652: `areWebhooksAvailable` also calls `cleanUpRemovedWebhooks` - are we also waiting to cleanup removed webhooks until the deployment is up? could this introduce a period of failures due to unbacked validating or mutating webhooks? should we still somehow clean up removed webhooks whether the deployment is up or not?

perdasilva (commented):
I left a small comment. I'd also maybe ask for an e2e to exercise this fix. Wdyt?

pkg/controller/operators/olm/operator.go
ugiordan: You're right, the readiness gate was also delaying the cleanup of removed validating and mutating webhook configurations.  I split cleanup from availability checks. `cleanUpRemovedWebhooks` now runs from `updateInstallStatus` even while the deployment is coming up, while conversion webhook activation remains gated until the deployment is ready. I added a regression test for this case, and the OLM controller and install tests pass.

ugiordan:
> I left a small comment. I'd also maybe ask for an e2e to exercise this fix. Wdyt?

@perdasilva, totally agree; I forgot about it. Added an E2E case covering an unready conversion-webhook replacement.

perdasilva:
@ugiordan I'm missing the E2E in the PR...? Also there seems to be a failure in one of the webhook related e2es =/

ugiordan:
> @ugiordan I'm missing the E2E in the PR...? Also there seems to be a failure in one of the webhook related e2es =/

@perdasilva I missed the e2e. Added them now. 
I also fixed the webhook E2E failure that left the replacement CSV in NeedsReinstall. Focused tests pass, and CI is running again.

PR #3893, opened 2026-08-13
WIP: OCPBUGS-35210: prevent KCM from deleting SA-token secrets created before their SA

rsacherer:
SA-token Secrets included in an operator bundle are placed earlier in the InstallPlan step list than the synthesized ServiceAccount step. Side effects can install an operator with proper secrets configured, or it can install the operator with a missing secret. In my tests (see below document) it's around 40/60 of hit and miss. Therefore the issue is not always reproduceable (see test document for more details). In a nut shell, a rescheduled loop run is able to get the stale RV, which still does not see the Secret as created, but the prior loop created the SA so this time the Secret will not be deleted.

Fix: add a StepperFunc for BundleSecretKind (mirroring the existing CRD StepperFunc pattern). The new NewBundleSecretStep checks whether the SA referenced by the secret already exists before attempting creation:

- SA absent  → return WaitingForAPI; NeedsRequeue() returns true, keeping phase=Installing and triggering a 5-second requeue.
- SA present → create the secret with correct owner refs (live API UID lookup, matching getUpdatedOwnerReferences behaviour) and return Created/Present.

This (still DRAFT) PR also adds structured debug logging to syncInstallPlans (plan resourceVersion, per-step BS/SA status at reconcile start, UpdateStatus call/result) to make the race observable in OLM pod logs during investigation.

Most likely the (or some) additional logging will be removed by further force pushed commits.

Tested: 30-iteration statistical reproducer against OCP 4.17 — 0/30 bug fires with the fix (without the fix we are looking at roughly 60% failure/40% Success).

This PR is still a draft and work in progress.

OLM creates the Secret before the SA exists; the Kubernetes token controller (KCM) immediately deletes any token secret whose referenced ServiceAccount is absent, and OLM then marks the step as Created permanently — preventing any future retry and leaving the operator without its token secret.

**Architectural changes:**
Add a new NewBundleSecretStep StepperFunc and do not create a BundleSecret before it's SA has been created.

Fix rationale and Test documentation:
https://docs.google.com/document/d/1DPNBepg1_tIfvh1uhILMIs5cWt2kjSVtdpvQ6zchoE0/edit?tab=t.mm6y4gofm4

**Testing remarks:**

Still missing, but on my TODO list:
- regression tests
- 

**Reviewer Checklist**
- [X] Implementation matches the proposed design, or proposal is updated to match implementation
- [X] Sufficient end-to-end test coverage
- [x] Bug fixes are accompanied by regression test(s)
- [X] Commit messages sensible and descriptive
- [X] Code is properly formatted

## Summary by CodeRabbit

* **New Features**
  * Added support for creating and updating bundle secrets during catalog installation.
  * Bundle secrets now receive appropriate labels and ownership metadata automatically.
  * Installation waits for required service accounts before processing dependent resources.
  * Required service-account and access-control resources are prepared before dependent catalog resources.

* **Bug Fixes**
  * Improved installation status tracking and handling of failures, conflicts, skipped steps, and terminal states.
  * Added more precise timestamps and enhanced diagnostics for catalog operations.

### DIFF
--- a/cmd/catalog/start.go
+++ b/cmd/catalog/start.go
@@ -49,6 +49,16 @@ func newRootCmd() *cobra.Command {
 			if o.debug {
 				logger.SetLevel(logrus.DebugLevel)
 			}
+			// OCPBUGS-35210: use millisecond timestamps so OLM and audit log
+			// entries can be correlated at sub-second precision.
+			// Set on both the local logger AND the global package logger so that
+			// code using logrus.WithFields() directly also emits milliseconds.
+			msFormatter := &logrus.TextFormatter{
+				TimestampFormat: "2006-01-02T15:04:05.000Z07:00",
+				FullTimestamp:   true,
+			}
+			logger.SetFormatter(msFormatter)
+			logrus.SetFormatter(msFormatter)
 			logger.Infof("log level %s", logger.Level)
 
 			ctx, cancel := context.WithCancel(signals.Context())
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -2100,6 +2100,23 @@ func (o *Operator) syncInstallPlans(obj interface{}) (syncError error) {
 
 	logger.Info("syncing")
 
+	// OCPBUGS-35210: log the step statuses this reconcile sees at start.
+	// Proves whether this loop received a stale cached plan (NotPresent) or
+	// the post-UpdateStatus version (Created) for the BundleSecret step.
+	if len(plan.Status.Plan) > 0 {
+		for i, step := range plan.Status.Plan {
+			if step.Resource.Kind == "BundleSecret" || step.Resource.Kind == "ServiceAccount" {
+				logger.WithFields(logrus.Fields{
+					"resourceVersion": plan.ResourceVersion,
+					"stepIndex":       i,
+					"kind":            step.Resource.Kind,
+					"name":            step.Resource.Name,
+					"status":          step.Status,
+				}).Debug("installplan step status at reconcile start")
+			}
+		}
+	}
+
 	if len(plan.Status.Plan) == 0 && len(plan.Status.BundleLookups) == 0 {
 		logger.Info("skip processing installplan without status - subscription sync responsible for initial status")
 		return
@@ -2107,6 +2124,9 @@ func (o *Operator) syncInstallPlans(obj interface{}) (syncError error) {
 
 	// Complete and Failed are terminal phases
 	if plan.Status.Phase == v1alpha1.InstallPlanPhaseFailed || plan.Status.Phase == v1alpha1.InstallPlanPhaseComplete {
+		// OCPBUGS-35210: log so we can confirm terminal-phase early exit in the timeline.
+		// Loops that see phase=Complete exit here without executing any steps.
+		logger.WithField("phase", plan.Status.Phase).Debug("phase is terminal, skipping execution")
 		return
 	}
 
@@ -2169,8 +2189,26 @@ func (o *Operator) syncInstallPlans(obj interface{}) (syncError error) {
 
 	defer o.requeueSubscriptionForInstallPlan(plan, logger)
 
+	// OCPBUGS-35210: log what step statuses are being persisted and the
+	// resourceVersion. A concurrent reconcile that reads before this write
+	// will have an older resourceVersion and see different step statuses.
+	{
+		fields := logrus.Fields{
+			"resourceVersion": outInstallPlan.ResourceVersion,
+			"phase":           outInstallPlan.Status.Phase,
+		}
+		for i, step := range outInstallPlan.Status.Plan {
+			if step.Resource.Kind == "BundleSecret" || step.Resource.Kind == "ServiceAccount" {
+				fields[fmt.Sprintf("step[%d].%s", i, step.Resource.Kind)] = string(step.Status)
+			}
+		}
+		logger.WithFields(fields).Debug("calling UpdateStatus")
+	}
+
 	// Update InstallPlan with status of transition. Log errors if we can't write them to the status.
-	if _, err := o.client.OperatorsV1alpha1().InstallPlans(plan.GetNamespace()).UpdateStatus(context.TODO(), outInstallPlan, metav1.UpdateOptions{}); err != nil {
+	if updatedPlan, err := o.client.OperatorsV1alpha1().InstallPlans(plan.GetNamespace()).UpdateStatus(context.TODO(), outInstallPlan, metav1.UpdateOptions{}); err != nil {
+		// OCPBUGS-35210: a 409 here means step statuses were NOT persisted.
+		// A concurrent reconcile that already read NotPresent will re-execute the BundleSecret step.
 		logger = logger.WithField("updateError", err.Error())
 		updateErr := errors.New("error updating InstallPlan status: " + err.Error())
 		if syncError == nil {
@@ -2179,6 +2217,13 @@ func (o *Operator) syncInstallPlans(obj interface{}) (syncError error) {
 		}
 		logger.Info("error transitioning InstallPlan")
 		syncError = fmt.Errorf("error transitioning InstallPlan: %s and error updating InstallPlan status: %s", syncError, updateErr)
+	} else {
+		// OCPBUGS-35210: log the new resourceVersion after a successful write.
+		// Any reconcile loop that read a lower resourceVersion saw stale data.
+		logger.WithFields(logrus.Fields{
+			"newResourceVersion": updatedPlan.ResourceVersion,
+			"phase":              updatedPlan.Status.Phase,
+		}).Debug("UpdateStatus succeeded")
 	}
 
 	return
@@ -2474,9 +2519,10 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
 		o.logger.Errorf("failed to get a client for plan execution- %v", err)
 		return err
 	}
-	b := newBuilder(plan, o.lister.OperatorsV1alpha1().ClusterServiceVersionLister(), builderKubeClient, builderDynamicClient, r, o.logger, o.recorder)
+	b := newBuilder(plan, o.lister.OperatorsV1alpha1().ClusterServiceVersionLister(), builderKubeClient, kubeclient, o.client, builderDynamicClient, r, o.logger, o.recorder)
 
 	for i, step := range plan.Status.Plan {
+		beforeStatus := plan.Status.Plan[i].Status
 		if err := func(i int, step *v1alpha1.Step) error {
 			wr.PopWarnings()
 			defer func() {
@@ -2521,14 +2567,25 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
 			}
 
 			switch step.Status {
-			case v1alpha1.StepStatusPresent, v1alpha1.StepStatusCreated, v1alpha1.StepStatusWaitingForAPI:
+			case v1alpha1.StepStatusPresent, v1alpha1.StepStatusCreated:
+				// OCPBUGS-35210: log skipped steps so we can confirm which reconcile
+				// loop sees Created (and skips) vs NotPresent (and re-executes).
+				if step.Resource.Kind == "BundleSecret" || step.Resource.Kind == "ServiceAccount" {
+					o.logger.WithFields(logrus.Fields{
+						"kind":   step.Resource.Kind,
+						"name":   step.Resource.Name,
+						"status": step.Status,
+					}).Debug("skipping step — already Created/Present")
+				}
+				return nil
+			case v1alpha1.StepStatusWaitingForAPI:
 				return nil
 			case v1alpha1.StepStatusUnknown, v1alpha1.StepStatusNotPresent:
 				manifest, err := r.ManifestForStep(step)
 				if err != nil {
 					return err
 				}
-				o.logger.WithFields(logrus.Fields{"kind": step.Resource.Kind, "name": step.Resource.Name}).Debug("execute resource")
+				o.logger.WithFields(logrus.Fields{"kind": step.Resource.Kind, "name": step.Resource.Name, "stepIndex": i}).Debug("execute resource")
 				switch step.Resource.Kind {
 				case v1alpha1.ClusterServiceVersionKind:
 					// Marshal the manifest into a CSV instance.
@@ -2590,40 +2647,6 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
 					sub.Labels[install.OLMManagedLabelKey] = install.OLMManagedLabelValue
 
 					status, err := ensurer.EnsureSubscription(&sub)
-					if err != nil {
-						return err
-					}
-
-					plan.Status.Plan[i].Status = status
-
-				case resolver.BundleSecretKind:
-					var s corev1.Secret
-					err := json.Unmarshal([]byte(manifest), &s)
-					if err != nil {
-						return errorwrap.Wrapf(err, "error parsing step manifest: %s", step.Resource.Name)
-					}
-
-					// add ownerrefs on the secret that point to the CSV in the bundle
-					if step.Resolving != "" {
-						owner := &v1alpha1.ClusterServiceVersion{}
-						owner.SetNamespace(plan.GetNamespace())
-						owner.SetName(step.Resolving)
-						ownerutil.AddNonBlockingOwner(&s, owner)
-					}
-
-					// Update UIDs on all CSV OwnerReferences
-					updated, err := o.getUpdatedOwnerReferences(s.OwnerReferences, plan.Namespace)
-					if err != nil {
-						return errorwrap.Wrapf(err, "error generating ownerrefs for secret %s", s.GetName())
-					}
-					s.SetOwnerReferences(updated)
-					s.SetNamespace(namespace)
-					if s.Labels == nil {
-						s.Labels = map[string]string{}
-					}
-					s.Labels[install.OLMManagedLabelKey] = install.OLMManagedLabelValue
-
-					status, err := ensurer.EnsureBundleSecret(plan.Namespace, &s)
 					if err != nil {
 						return err
 					}
@@ -2915,7 +2938,39 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
 					return notFoundErr
 				}
 			}
+			// OCPBUGS-35210: log the step that caused ExecutePlan to fail.
+			// This error becomes syncError in syncInstallPlans and appears in the
+			// UpdateStatus WRITE log — it is NOT a UpdateStatus error itself.
+			o.logger.WithFields(logrus.Fields{
+				"kind":      step.Resource.Kind,
+				"name":      step.Resource.Name,
+				"stepIndex": i,
+				"error":     err.Error(),
+			}).Debug("step execution failed — ExecutePlan returning error")
 			return err
+		}
+		// OCPBUGS-35210: log each step's outcome. Distinguish between steps that
+		// were actually executed (status changed) and steps that were skipped
+		// because they were already in a terminal state (status unchanged).
+		afterStatus := plan.Status.Plan[i].Status
+		if afterStatus == beforeStatus {
+			msg := "step execution made no progress"
+			if afterStatus == v1alpha1.StepStatusCreated || afterStatus == v1alpha1.StepStatusPresent {
+				msg = "step skipped — already in terminal state"
+			}
+			o.logger.WithFields(logrus.Fields{
+				"kind":      step.Resource.Kind,
+				"name":      step.Resource.Name,
+				"stepIndex": i,
+				"status":    afterStatus,
+			}).Debug(msg)
+		} else {
+			o.logger.WithFields(logrus.Fields{
+				"kind":      step.Resource.Kind,
+				"name":      step.Resource.Name,
+				"stepIndex": i,
+				"result":    afterStatus,
+			}).Debug("step execution result")
 		}
 	}
 
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -2197,6 +2197,9 @@ func (o *Operator) syncInstallPlans(obj interface{}) (syncError error) {
 			"resourceVersion": outInstallPlan.ResourceVersion,
 			"phase":           outInstallPlan.Status.Phase,
 		}
+		if syncError != nil {
+			fields["syncError"] = syncError.Error()
+		}
 		for i, step := range outInstallPlan.Status.Plan {
 			if step.Resource.Kind == "BundleSecret" || step.Resource.Kind == "ServiceAccount" {
 				fields[fmt.Sprintf("step[%d].%s", i, step.Resource.Kind)] = string(step.Status)
@@ -2519,7 +2522,7 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
 		o.logger.Errorf("failed to get a client for plan execution- %v", err)
 		return err
 	}
-	b := newBuilder(plan, o.lister.OperatorsV1alpha1().ClusterServiceVersionLister(), builderKubeClient, kubeclient, o.client, builderDynamicClient, r, o.logger, o.recorder)
+	b := newBuilder(plan, o.lister.OperatorsV1alpha1().ClusterServiceVersionLister(), builderKubeClient, kubeclient, crclient, builderDynamicClient, r, o.logger, o.recorder)
 
 	for i, step := range plan.Status.Plan {
 		beforeStatus := plan.Status.Plan[i].Status
--- a/pkg/controller/operators/catalog/operator.go
+++ b/pkg/controller/operators/catalog/operator.go
@@ -2969,6 +2969,7 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
 			}).Debug(msg)
 		} else {
 			o.logger.WithFields(logrus.Fields{
+				"ip":        plan.GetName(),
 				"kind":      step.Resource.Kind,
 				"name":      step.Resource.Name,
 				"stepIndex": i,
--- a/pkg/controller/operators/catalog/step.go
+++ b/pkg/controller/operators/catalog/step.go
@@ -379,10 +379,12 @@ func (b *builder) NewBundleSecretStep(step *v1alpha1.Step, manifest string) Step
 		s.Labels[install.OLMManagedLabelKey] = install.OLMManagedLabelValue
 
 		// Add the resolving CSV as a non-blocking owner so the secret is GC'd on
-		// uninstall. Use the lister (catalog-operator credentials, avoids extra API
-		// call) — UID is stable once set so a briefly-stale lister entry is safe.
+		// uninstall. Use a live API call — the CSV may have been created in this
+		// same ExecutePlan invocation and will not yet be in the lister cache.
+		// A lister NotFound here would bubble through apierrors.IsNotFound and
+		// incorrectly trigger the discovery-querier path in ExecutePlan.
 		if step.Resolving != "" {
-			csv, err := b.csvLister.ClusterServiceVersions(namespace).Get(step.Resolving)
+			csv, err := b.olmClient.OperatorsV1alpha1().ClusterServiceVersions(namespace).Get(context.TODO(), step.Resolving, metav1.GetOptions{})
 			if err != nil {
 				return v1alpha1.StepStatusUnknown, fmt.Errorf("error getting csv %s for secret owner ref: %w", step.Resolving, err)
 			}
--- a/pkg/controller/operators/catalog/step.go
+++ b/pkg/controller/operators/catalog/step.go
@@ -2,9 +2,11 @@ package catalog
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/install"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/registry/resolver"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
@@ -19,10 +21,12 @@ import (
 	"k8s.io/client-go/util/retry"
 
 	"github.com/operator-framework/api/pkg/operators/v1alpha1"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/clientset/versioned"
 	listersv1alpha1 "github.com/operator-framework/operator-lifecycle-manager/pkg/api/client/listers/operators/v1alpha1"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/controller/operators/internal/alongside"
 	crdlib "github.com/operator-framework/operator-lifecycle-manager/pkg/lib/crd"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/operatorclient"
+	"github.com/operator-framework/operator-lifecycle-manager/pkg/lib/ownerutil"
 )
 
 // Stepper manages cluster interactions based on the step.
@@ -43,6 +47,8 @@ type builder struct {
 	plan             *v1alpha1.InstallPlan
 	csvLister        listersv1alpha1.ClusterServiceVersionLister
 	opclient         operatorclient.ClientInterface
+	attenuatedClient operatorclient.ClientInterface
+	olmClient        versioned.Interface
 	dynamicClient    dynamic.Interface
 	manifestResolver ManifestResolver
 	logger           logrus.FieldLogger
@@ -51,11 +57,13 @@ type builder struct {
 	annotator alongside.Annotator
 }
 
-func newBuilder(plan *v1alpha1.InstallPlan, csvLister listersv1alpha1.ClusterServiceVersionLister, opclient operatorclient.ClientInterface, dynamicClient dynamic.Interface, manifestResolver ManifestResolver, logger logrus.FieldLogger, er record.EventRecorder) *builder {
+func newBuilder(plan *v1alpha1.InstallPlan, csvLister listersv1alpha1.ClusterServiceVersionLister, opclient operatorclient.ClientInterface, attenuatedClient operatorclient.ClientInterface, olmClient versioned.Interface, dynamicClient dynamic.Interface, manifestResolver ManifestResolver, logger logrus.FieldLogger, er record.EventRecorder) *builder {
 	return &builder{
 		plan:             plan,
 		csvLister:        csvLister,
 		opclient:         opclient,
+		attenuatedClient: attenuatedClient,
+		olmClient:        olmClient,
 		dynamicClient:    dynamicClient,
 		manifestResolver: manifestResolver,
 		logger:           logger,
@@ -91,6 +99,8 @@ func (b *builder) create(step v1alpha1.Step) (Stepper, error) {
 		case crdlib.V1Beta1Version:
 			return b.NewCRDV1Beta1Step(b.opclient.ApiextensionsInterface().ApiextensionsV1beta1(), &step, manifest), nil
 		}
+	case resolver.BundleSecretKind:
+		return b.NewBundleSecretStep(&step, manifest), nil
 	}
 	return nil, notSupportedStepperErr{fmt.Sprintf("stepper interface does not support %s", step.Resource.Kind)}
 }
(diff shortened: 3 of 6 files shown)


pkg/controller/operators/catalog/operator.go
rsacherer, line 2514: The fix here does Implement this TODO as well with the new Stepper function

pkg/controller/operators/catalog/operator.go
```diff
@@ -2169,8 +2189,26 @@ func (o *Operator) syncInstallPlans(obj interface{}) (syncError error) {
...
+			}
+		}
+		logger.WithFields(fields).Debug("calling UpdateStatus")
+	}
+
 	// Update InstallPlan with status of transition. Log errors if we can't write them to the status.
-	if _, err := o.client.OperatorsV1alpha1().InstallPlans(plan.GetNamespace()).UpdateStatus(context.TODO(), outInstallPlan, metav1.UpdateOptions{}); err != nil {
+	if updatedPlan, err := o.client.OperatorsV1alpha1().InstallPlans(plan.GetNamespace()).UpdateStatus(context.TODO(), outInstallPlan, metav1.UpdateOptions{}); err != nil {
+		// OCPBUGS-35210: a 409 here means step statuses were NOT persisted.
+		// A concurrent reconcile that already read NotPresent will re-execute the BundleSecret step.
```
joelanford, line 2214: This sounds very similar to a related bug: https://redhat.atlassian.net/browse/OCPBUGS-106160

Something I mentioned in a [Slack conversation](https://redhat-internal.slack.com/archives/C3VS0LV41/p1786561506544429?thread_ts=1786550385.532839&cid=C3VS0LV41) about _that_ bug was:
> There may be some race conditions that cause the CR validation logic to trigger multiple times.
>
> For example, if a conflict occurs on the InstallPlan that causes the CRD step not to be updated to Present (or whatever the enum is for "I successfully applied the CRD"), then the next reconcile of the InstallPlan will see "oh, I need to apply the CRD from this step" and then do the CR validation as a preflight again.

joelanford:
At first glance, this seems like a reasonable patch for this issue.

One thing I'm not clear on: is this bug caused by concurrent reconciliation of the same `InstallPlan`? Or at least quick succession reconciles where the second reconcile runs fast enough that it does not have the result of the first in the cache yet?

I'm curious about is whether there is a deeper issue that we could solve that eliminates multiple concurrent reconciles of the same InstallPlan, and if that would fix this issue at a lower level (and likely address other latent bugs along with it).

rsacherer:
> At first glance, this seems like a reasonable patch for this issue.
> 
> One thing I'm not clear on: is this bug caused by concurrent reconciliation of the same `InstallPlan`? Or at least quick succession reconciles where the second reconcile runs fast enough that it does not have the result of the first in the cache yet?
> 
> I'm curious about is whether there is a deeper issue that we could solve that eliminates multiple concurrent reconciles of the same InstallPlan, and if that would fix this issue at a lower level (and likely address other latent bugs along with it).

Hi Joe, 

the concurrent reconciliation with stale cache entries are actually what helps in the success case:

- Loop A creates Secret and set the step to Created
- milliseconds later KCM deletes the secret because SA is missing
- Loop A, in later steps create the SA
- Loop B starts with stale cache copy (not the RV written by Loop A)
- The secret step still is seen as unknown, secret gets created
- Because Loop A did eventually create the SA the KCM is not deleting the secret anymore


The above explanation leaves out other issues like errors during Loop A before SA creation step etc.

Example (more details are seen in the linked document):
```
══════════════════════════════════════════════════════
Run 2/30  [13:00:55]
══════════════════════════════════════════════════════
[...]
  Step order  : SA=step[23]  BundleSecret=step[10]  (total=28  gap=13 steps)
  Approved at : 2026-08-11T13:01:13.449
[...]  LOOP  0, Approval rv=2396467
    13:01:14.293635  OLM:    [ r6PNH]  LOOP START  phase=RequiresApproval
    13:01:14.293672  OLM:    [ r6PNH]  READ  rv=2396467  step[10].BundleSecret=Unknown  ← BUG STEP
    13:01:14.293692  OLM:    [ r6PNH]  READ  rv=2396467  step[23].ServiceAccount=Unknown
    13:01:14.301565  OLM:    [ r6PNH]  UpdateStatus WRITE  rv=2396467  BS=Unknown  phase=Installing
    13:01:14.303136  AUDIT:  [------]  UpdateStatus [200]  OK -- step statuses written  InstallPlan/install-tphd6
    13:01:14.389304  OLM:    [ r6PNH]  UpdateStatus OK  newRV=2396468  phase=Installing
    13:01:14.389348  OLM:    [ r6PNH]  LOOP END/REQUEUE  phase=RequiresApproval

[...] LOOP U6bRi newRV=2396468

    13:01:14.395715  OLM:    [ U6bRi]  LOOP START  phase=Installing
    13:01:14.395751  OLM:    [ U6bRi]  READ  rv=2396468  step[10].BundleSecret=Unknown  ← BUG STEP
    13:01:14.395776  OLM:    [ U6bRi]  READ  rv=2396468  step[23].ServiceAccount=Unknown

[...] Bundle Secret Step

    13:01:19.735782  OLM:    [    ----]  CREATE BundleSecret  planRV=2396468  step[10]=Unknown  sa=openshift-gitops-operator-controller-manager  ← race starts here
    13:01:19.735842  OLM:    [    ----]  EnsureBundleSecret: creating token secret (OCPBUGS-35210: SA may not exist yet)  sa=openshift-gitops-operator-controller-manager

[...] AUDIT: OLM creates the token

    13:01:19.737278  AUDIT:  [------]  create [201]  Secret/openshift-gitops-operator-metrics-monitor-bearer-token  by OLM
    13:01:19.770741  OLM:    [    ----]  EnsureBundleSecret: Create succeeded -- KCM will delete if SA absent, step marked Created regardless  sa=openshift-gitops-operator-controller-manager
    13:01:19.770790  OLM:    [    ----]  step[10]  BundleSecret                 openshift-gitops-operator-metrics-monitor-bearer-token  → Created  ← TOKEN SECRET

[...] AUDIT: KCM deletes the token, while in memory Secret step is set to created.

    13:01:19.974159  AUDIT:  [------]  delete [200]  Secret/openshift-gitops-operator-metrics-monitor-bearer-token  by KCM/TokensController

[...] Step 23, SA is eventually created in LOOP U6bRi

    13:01:26.834238  OLM:    [    ----]  step[23]  ServiceAccount               openshift-gitops-operator-controller-manager  → Created  ← SA

[...] Loop U6bRi finishes, in this RV, the BundleSecret step is now set to Created

    13:01:27.375351  OLM:    [ U6bRi]  UpdateStatus WRITE  rv=2396468  BS=Created  phase=Complete
    13:01:27.379456  AUDIT:  [------]  UpdateStatus [200]  OK -- step statuses written  InstallPlan/install-tphd6

[...] We write status, newRV is now 2396717

    13:01:27.438555  OLM:    [ U6bRi]  UpdateStatus OK  newRV=2396717  phase=Complete
    13:01:27.438606  OLM:    [ U6bRi]  LOOP END/REQUEUE  phase=Installing

[...] LOOP EVAJt starts, but get's "stale" RV (2396468, therefore BundleSecret is still seen as Unknown

    13:01:27.438700  OLM:    [ EVAJt]  LOOP START  phase=Installing
    13:01:27.438923  OLM:    [ EVAJt]  READ  rv=2396468  step[10].BundleSecret=Unknown  ← BUG STEP

[...] We create the bundle again, as it still shows Unknown, most other steps are seen as `present` already

    13:01:34.783884  OLM:    [    ----]  CREATE BundleSecret  planRV=2396468  step[10]=Unknown  sa=openshift-gitops-operator-controller-manager  ← race starts here
    13:01:34.783944  OLM:    [    ----]  EnsureBundleSecret: creating token secret (OCPBUGS-35210: SA may not exist yet)  sa=openshift-gitops-operator-controller-manager

[...] AUDIT shows OLM creating the secret, this time, no immediate delete because the SA got created in plan U6bRi

    13:01:34.786347  AUDIT:  [------]  create [201]  Secret/openshift-gitops-operator-metrics-monitor-bearer-token  by OLM


Final result:
  step_status : Created
  secret now  : present
  VERDICT     : ok
```

rsacherer:
I think it might be worthwile to fix the stale RV issue, it would prevent useless re-reruns of steps. However, as I understood OLM, it should be idempotent, so there should not be an issue with that particular point. It is the deletion of BundleSecrets with no associated SA that breaks that idempotency and adds the depencendy from secret to SA, which this proposed fix resolves.

I'll try to get another fix in to see if we can wait for another RV version if we know a previous loop has RV number 10 and has written it into RV number 12, then we should not use RV number 10, backoff some X number of ms (100?) and try again until we get 12 || 12+x as RV number. However, need to check if that is actually possible and if we have all that information at hand at that point.

On the other hand, besides running through steps again and seeing they are already done (present) there seems to be no further harm done, as long as all steps are truly idempotent.

pkg/controller/operators/catalog/step.go
```diff
@@ -91,6 +99,8 @@ func (b *builder) create(step v1alpha1.Step) (Stepper, error) {
 		case crdlib.V1Beta1Version:
 			return b.NewCRDV1Beta1Step(b.opclient.ApiextensionsInterface().ApiextensionsV1beta1(), &step, manifest), nil
 		}
+	case resolver.BundleSecretKind:
+		return b.NewBundleSecretStep(&step, manifest), nil
```
tmshort, line 103: **Dead duplicate handler.** Now that `create()` returns a `StepperFunc` for `resolver.BundleSecretKind`, `doStep` is true and `s.Status()` runs — so the existing `case resolver.BundleSecretKind` in the `ExecutePlan` switch in `operator.go` (~2656–2688) is unreachable. Two divergent implementations now coexist: the old block uses `getUpdatedOwnerReferences`, this new one doesn't. Recommend deleting the old switch case so a future maintainer editing owner-ref logic there isn't editing dead code.

pkg/controller/operators/catalog/step.go
```diff
@@ -318,3 +328,80 @@ func setInstalledAlongsideAnnotation(a alongside.Annotator, dst metav1.Object, n
...
+		saName := s.Annotations[corev1.ServiceAccountNameKey]
+		if s.Type == corev1.SecretTypeServiceAccountToken && saName != "" {
+			_, saErr := b.attenuatedClient.KubernetesInterface().CoreV1().
+				ServiceAccounts(namespace).Get(context.TODO(), saName, metav1.GetOptions{})
+			if apierrors.IsNotFound(saErr) {
+				logrus.WithFields(logrus.Fields{
+					"secret": s.Name,
+					"sa":     saName,
+				}).Info("BundleSecretStep: SA not yet created — returning WaitingForAPI (OCPBUGS-35210)")
+				return v1alpha1.StepStatusWaitingForAPI, nil
```
tmshort, line 366: **Forever-`WaitingForAPI` edge case.** If an SA-token secret references an SA that is *not* part of this plan (e.g. an externally-managed SA that never gets created on-cluster), this returns `WaitingForAPI` on every reconcile → `NeedsRequeue()` keeps `phase=Installing` → the install eventually times out to `Failed`. The old path created the secret and completed. Consider bounding the wait (e.g. give up / proceed after N retries, or only gate on SAs that appear later in this plan) so a missing-external-SA case doesn't hang the install indefinitely.

pkg/controller/operators/catalog/step.go
```diff
@@ -318,3 +328,80 @@ func setInstalledAlongsideAnnotation(a alongside.Annotator, dst metav1.Object, n
...
+			s.Labels = map[string]string{}
+		}
+		s.Labels[install.OLMManagedLabelKey] = install.OLMManagedLabelValue
+
+		// Add the resolving CSV as a non-blocking owner so the secret is GC'd when
+		// the operator is uninstalled. Use a live API call (matching the existing
+		// getUpdatedOwnerReferences pattern) so the UID is always current rather
+		// than relying on the informer cache, which may not yet reflect newly-created
+		// objects. Return an error on failure to trigger a retry.
+		if step.Resolving != "" {
```
tmshort, line 386: **Owner-ref UID refresh dropped vs. the old path.** This only adds the resolving CSV as owner, and only when `step.Resolving != ""`. The old `BundleSecret` handler ran `getUpdatedOwnerReferences`, which refreshed/populated the UID of *any* CSV ownerReference already present in the manifest via the live client. If a bundle secret ships its own CSV ownerReference (or `step.Resolving` is empty), those refs keep empty/stale UIDs here and GC-on-uninstall may not work. Worth preserving the `getUpdatedOwnerReferences` pass for pre-existing owner refs.

pkg/controller/registry/resolver/steps.go
```diff
@@ -148,6 +148,8 @@ func NewStepResourceFromBundle(bundle *api.Bundle, namespace, replaces, catalogS
 	}
 	steps := []v1alpha1.StepResource{step}
 
+	// Original ordering: bundle objects first, then synthesized SA/RBAC last.
```
tmshort: **Altitude note (soft — not a blocker).** The comment itself calls this the "UNFIXED ordering." Appending the synthesized SA/RBAC steps *before* the `bundle.Object` loop would fix the deterministic single-pass case at the source, with far less surface than the runtime workaround. That said, a static reorder alone doesn't close the concurrent/stale-cache race (a 409 dropping the SA step's persisted status, or a concurrent reconcile on a cached view, still reintroduces it), so the runtime check has real value. The ideal is likely both: reorder for the common path + keep the existence check for the race. Flagging so the leftover ordering is a conscious decision rather than latent.

pkg/controller/operators/catalog/step.go
```diff
@@ -318,3 +328,80 @@ func setInstalledAlongsideAnnotation(a alongside.Annotator, dst metav1.Object, n
...
+			csv, err := b.olmClient.OperatorsV1alpha1().
+				ClusterServiceVersions(namespace).Get(context.TODO(), step.Resolving, metav1.GetOptions{})
+			if err != nil {
+				return v1alpha1.StepStatusUnknown, fmt.Errorf("error getting csv %s for secret owner ref: %w", step.Resolving, err)
+			}
+			ownerutil.AddNonBlockingOwner(&s, csv)
+		}
+
+		_, createErr := b.attenuatedClient.KubernetesInterface().CoreV1().
```
tmshort: **Duplicated logic.** The create → `IsAlreadyExists` → `SetNamespace` → `UpdateSecret` sequence here duplicates `StepEnsurer.EnsureBundleSecret` (`step_ensurer.go` ~118–139) almost verbatim. Future fixes to secret creation (error wrapping, conflict retry, etc.) would have to be made in two places and can drift. Consider reusing `EnsureBundleSecret` for the create/update half and keeping only the SA-existence gate + owner-ref logic unique to this StepperFunc.

pkg/controller/operators/catalog/step.go
```diff
@@ -318,3 +328,80 @@ func setInstalledAlongsideAnnotation(a alongside.Annotator, dst metav1.Object, n
...
+		namespace := b.plan.GetNamespace()
+
+		var s corev1.Secret
+		if err := json.Unmarshal([]byte(manifest), &s); err != nil {
+			return v1alpha1.StepStatusUnknown, err
+		}
+
+		saName := s.Annotations[corev1.ServiceAccountNameKey]
+		if s.Type == corev1.SecretTypeServiceAccountToken && saName != "" {
+			_, saErr := b.attenuatedClient.KubernetesInterface().CoreV1().
```
tmshort, line 359: **New `get serviceaccounts` requirement on the attenuated client (narrow, but worth a guard).** This unconditional `Get` on the scoped client is a new RBAC surface. The existing SA install path (`EnsureServiceAccount`) only calls `Get` on the `AlreadyExists` branch, so a scoped `AttenuatedServiceAccountRef` Role granting `create` but *not* `get` on serviceaccounts installs fine today and would newly fail here with `Forbidden` → error → retry-to-timeout → `Failed`. It's a narrow triple-conjunction (scoped install + minimal Role + bundle ships an SA-token secret) and non-scoped installs use the cluster-admin client, so low probability — but trivially avoidable: treat a `Forbidden` on this `Get` as "proceed," or fall back to the informer/lister, rather than erroring.

cmd/catalog/start.go
```diff
@@ -49,6 +49,16 @@ func newRootCmd() *cobra.Command {
...
+			// OCPBUGS-35210: use millisecond timestamps so OLM and audit log
+			// entries can be correlated at sub-second precision.
+			// Set on both the local logger AND the global package logger so that
+			// code using logrus.WithFields() directly also emits milliseconds.
+			msFormatter := &logrus.TextFormatter{
+				TimestampFormat: "2006-01-02T15:04:05.000000Z07:00",
+				FullTimestamp:   true,
+			}
+			logger.SetFormatter(msFormatter)
+			logrus.SetFormatter(msFormatter)
```
tmshort, line 61: **Debug/logging scaffolding to remove before merge (tracking all of it here).** This spans several spots — grouping so none get missed:
- **This line** `logrus.SetFormatter(msFormatter)` mutates the **process-global** logrus formatter, overriding whatever any other consumer of the package-global logger (including vendored code) relies on. This is the higher-risk half; line 60's local `logger.SetFormatter` is harmless.
- Scattered temporary `Debug`/`Info` logging across `operator.go` (~2103–2119, ~2189–2214, ~2567–2600, ~2972+) and the `WaitingForAPI` `Info` log at `step.go:362`.
- The debug code compares kinds with hardcoded `"BundleSecret"`/`"ServiceAccount"` string literals (e.g. `operator.go:2108`, 2201, 2573) instead of `resolver.BundleSecretKind` / `serviceAccountKind` — if any of that logging is *kept*, switch to the constants so a rename doesn't silently stop matching.

You already note the logging will be stripped; just flagging the global-formatter side effect specifically since it's easy to overlook.

pkg/controller/operators/catalog/step.go
```diff
@@ -318,3 +328,80 @@ func setInstalledAlongsideAnnotation(a alongside.Annotator, dst metav1.Object, n
...
+			ownerutil.AddNonBlockingOwner(&s, csv)
+		}
+
+		_, createErr := b.attenuatedClient.KubernetesInterface().CoreV1().
+			Secrets(namespace).Create(context.TODO(), &s, metav1.CreateOptions{})
+		if createErr == nil {
+			return v1alpha1.StepStatusCreated, nil
+		}
+		if apierrors.IsAlreadyExists(createErr) {
+			s.SetNamespace(namespace)
```
tmshort: Nit: `s.SetNamespace(namespace)` here is a no-op — the namespace was already set at line 373 (this mirrors a redundant call in `EnsureBundleSecret`). Can be dropped.

tmshort (commented):
Automated code-review pass (draft PR — logging removal and regression tests already on the author's TODO, so those aren't re-raised as blockers).

The overall approach is sound: for the concurrent/stale-cache race this targets, a runtime SA-existence check is a legitimate strategy (arguably more robust than a static step reorder, though ideally you'd do both). The comments below focus on the implementation details. Most actionable before this leaves draft: the dead duplicate `BundleSecret` handler (now-unreachable), the forever-`WaitingForAPI` edge case, and the dropped owner-ref UID refresh.

tmshort:
@coderabbitai review
Because this is draft, CodeRabbit didn't review

pkg/controller/operators/catalog/step.go
rsacherer: Yes, but, the crucial thing here is that the old path *did* create the secret, and with the SA missing the KCM deleted the secret within milliseconds. So we did get the operator installed with the expectation of a created secret, while in reality it is gone.

So to my thinking in this case (e.g. missed creation of an externally managed SA, or a typo in the raw manifest annotation, etc.) it would be better to fail the installation and know it's failed because we where waiting for an SA and give a clue to what is missing.

It also would not `hang` the install indefinitely as th einstall would time out to my knowledge at some point?

pkg/controller/operators/catalog/step.go
rsacherer: I'll be looking into this.

pkg/controller/registry/resolver/steps.go
rsacherer: Yes, that comment is a leftover from first testing, and in my proposal document I mention that we might want to do the re-order anyway. So in aggrement with this comment and I'll remove the comment.

pkg/controller/operators/catalog/step.go
rsacherer: Interesting, I will look into this.

pkg/controller/operators/catalog/step.go
rsacherer: good point

pkg/controller/operators/catalog/step.go
rsacherer: OK

pkg/controller/operators/catalog/step.go
rsacherer: dead code removed, that was a left over from before I created a proper stepper function.

pkg/controller/operators/catalog/step.go
rsacherer: I feel like this should be OK. Because creating a SA-token secret *will* be deleted by KCM no matter what *IF* the SA does not exist, so IMO, it is better to fail the install and get a hint that an SA is missing then to say install is OK, the SA-token secret step is *created*, but it's actually not present.

cmd/catalog/start.go
```diff
@@ -49,6 +49,16 @@ func newRootCmd() *cobra.Command {
 			if o.debug {
 				logger.SetLevel(logrus.DebugLevel)
 			}
+			// OCPBUGS-35210: use millisecond timestamps so OLM and audit log
+			// entries can be correlated at sub-second precision.
+			// Set on both the local logger AND the global package logger so that
+			// code using logrus.WithFields() directly also emits milliseconds.
+			msFormatter := &logrus.TextFormatter{
+				TimestampFormat: "2006-01-02T15:04:05.000000Z07:00",
+				FullTimestamp:   true,
```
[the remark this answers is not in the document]
rsacherer: Changed to milliseconds, although, I'll review this with logging, I think I saw sub millisecond step executions, if that's the case, I'll change the loggingformat to microseconds and take care to update the comment as well.

cmd/catalog/start.go
rsacherer: Yes, at the end of this patch we will have an operator install test utility as well (it handles installations of complete catalogs already in a batched way). So I need to think about which log messages I'd actually would like to keep in DEBUG mode so we can get good output for the test utility.

pkg/controller/operators/catalog/operator.go
```diff
@@ -2474,9 +2519,10 @@ func (o *Operator) ExecutePlan(plan *v1alpha1.InstallPlan) error {
 		o.logger.Errorf("failed to get a client for plan execution- %v", err)
 		return err
 	}
-	b := newBuilder(plan, o.lister.OperatorsV1alpha1().ClusterServiceVersionLister(), builderKubeClient, builderDynamicClient, r, o.logger, o.recorder)
+	b := newBuilder(plan, o.lister.OperatorsV1alpha1().ClusterServiceVersionLister(), builderKubeClient, kubeclient, o.client, builderDynamicClient, r, o.logger, o.recorder)
```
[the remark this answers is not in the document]
rsacherer: After reviewing this, we would need to get the attenuatedClient to pass through here, while actually the same mechanism is already used in getUpdatedOwnerReferences itself, it uses o.client.OperatorsV1alpha1().ClusterServiceVersions(namespace).Get(...) for  the same operation to fetch CSV UIDs for owner references.

Not sure why coderabbit would flag that here and not elsewhere in the code.

pkg/controller/operators/catalog/step.go
rsacherer: We proceed now when we lack permission (Line 370)

pkg/controller/operators/catalog/step.go
rsacherer: Done, we match what getUpdatedOwnerReferences did now.

tmshort:
@coderabbitai review

pkg/controller/operators/catalog/step.go
```diff
@@ -318,3 +328,95 @@ func setInstalledAlongsideAnnotation(a alongside.Annotator, dst metav1.Object, n
...
+					"secret": s.Name,
+					"sa":     saName,
+				}).Info("BundleSecretStep: SA not yet created — returning WaitingForAPI (OCPBUGS-35210)")
+				return v1alpha1.StepStatusWaitingForAPI, nil
+			}
+			// Forbidden means the scoped client lacks get on serviceaccounts; proceed
+			// and attempt secret creation — KCM will gate on SA existence regardless.
+			if saErr != nil && !apierrors.IsForbidden(saErr) {
+				return v1alpha1.StepStatusUnknown, saErr
+			}
```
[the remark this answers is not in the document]
rsacherer: We did discuss that we specifically do *not* want to return an error on forbidden, but just create the secret anyway.

pkg/controller/operators/catalog/step.go
```diff
@@ -318,3 +328,95 @@ func setInstalledAlongsideAnnotation(a alongside.Annotator, dst metav1.Object, n
...
+func refreshCSVOwnerRefUIDs(refs []metav1.OwnerReference, olmClient versioned.Interface, namespace string) ([]metav1.OwnerReference, error) {
+	updated := append([]metav1.OwnerReference(nil), refs...)
+	for i, owner := range refs {
+		if owner.Kind == v1alpha1.ClusterServiceVersionKind {
+			csv, err := olmClient.OperatorsV1alpha1().ClusterServiceVersions(namespace).Get(context.TODO(), owner.Name, metav1.GetOptions{})
+			if err != nil {
+				return nil, err
+			}
+			owner.UID = csv.GetUID()
+			updated[i] = owner
```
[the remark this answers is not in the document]
rsacherer: There is already precedence for read only lookup operations using the olm client, so  unless we definitely want to use the attenuated or scoped client from my point of view using olmClient in this context should be OK.

rsacherer:
FYI, Running my test tool to test installation of one of those operators (between red hat operators and certified operators there seem to be 5 operators which have a token secret in the bundle) we can see clearly the issue with getting stale RVs:

```
  [15:15:32.193523] openshift-gitops-operator           iteration 3/3
  [15:15:32.193548] openshift-gitops-operator           (latest) starting...
  [15:15:37.521168]   ↳ install-fztj2:                 Step order: SA=step[23]  BS=step[15]  total=28  BS before SA by 8 steps ← bug-prone
  [15:15:37.841986]   ↳ install-fztj2:                 loop[+ji5K] → requeue: NeedsRequeue (phase=Installing)
  [15:15:40.132518]   ↳ install-fztj2:                 loop[8WfQE] → requeue: error: error updating CRD \
  [15:15:37.740248] AUDIT  update   [200]  subscriptions/status openshift-gitops-operator                               by OLM
  [15:15:37.773141] AUDIT  update   [200]  subscriptions/status openshift-gitops-operator                               by OLM
  [15:15:35.517]   ↳ install-fztj2:                 loop[je2x8] | RequiresApproval | rv=170049947 | ERR      syncError  Operation cannot be fulfilled on installplans.operators.coreos.com "install-fztj2": the object has been modified; please apply your changes to the latest version and try again
  [15:15:38.789]   ↳ install-fztj2:                 loop[8WfQE] | Installing | rv=170049984 |   0        Created  ClusterServiceVersion/openshift-gitops-operator.v1.21.3
  [15:15:39.275]   ↳ install-fztj2:                 loop[8WfQE] | Installing | rv=170049984 |   1        Present  CustomResourceDefinition/imageupdaters.argocd-image-updater.argoproj.io
  [15:15:39.375]   ↳ install-fztj2:                 loop[8WfQE] | Installing | rv=170049984 |   2  WaitingForApi  CustomResourceDefinition/analysisruns.argoproj.io
  [15:15:39.546]   ↳ install-fztj2:                 loop[8WfQE] | Installing | rv=170049984 |   3  WaitingForApi  CustomResourceDefinition/analysistemplates.argoproj.io
  [15:15:40.054]   ↳ install-fztj2:                 loop[8WfQE] | Installing | rv=170049984 | ERR      syncError  error updating CRD "applications.argoproj.io": customresourcedefinitions.apiextensions.k8s.io "applications.argoproj.io" not found
  [15:15:41.510]   ↳ install-fztj2:                 loop[wD+UA] | Installing | rv=170049984 |   0        Present  ClusterServiceVersion/openshift-gitops-operator.v1.21.3
  [15:15:41.590]   ↳ install-fztj2:                 loop[wD+UA] | Installing | rv=170049984 |   1  WaitingForApi  CustomResourceDefinition/imageupdaters.argocd-image-updater.argoproj.io
  [15:15:42.267]   ↳ install-fztj2:                 loop[wD+UA] | Installing | rv=170049984 |   2        Present  CustomResourceDefinition/analysisruns.argoproj.io
  [15:15:42.992]   ↳ install-fztj2:                 loop[wD+UA] | Installing | rv=170049984 |   3        Present  CustomResourceDefinition/analysistemplates.argoproj.io
  [15:15:43.233]   ↳ install-fztj2:                 loop[wD+UA] | Installing | rv=170049984 |   4  WaitingForApi  CustomResourceDefinition/applications.argoproj.io
  [15:15:43.857]   ↳ install-fztj2:                 loop[wD+UA] | Installing | rv=170049984 |   5  WaitingForApi  CustomResourceDefinition/applicationsets.argoproj.io
  [15:15:43.925]   ↳ install-fztj2:                 loop[wD+UA] | Installing | rv=170049984 |   6  WaitingForApi  CustomResourceDefinition/appprojects.argoproj.io
  [15:15:45.997]   ↳ install-fztj2:                 loop[wD+UA] | Installing | rv=170049984 |   7        Present  CustomResourceDefinition/argocds.argoproj.io
  [15:15:46.115]   ↳ install-fztj2:                 loop[wD+UA] | Installing | rv=170049984 |   8  WaitingForApi  CustomResourceDefinition/clusteranalysistemplates.argoproj.io
  [15:15:46.271]   ↳ install-fztj2:                 loop[wD+UA] | Installing | rv=170049984 |   9  WaitingForApi  CustomResourceDefinition/experiments.argoproj.io
  [15:15:46.353]   ↳ install-fztj2:                 loop[wD+UA] | Installing | rv=170049984 |  10  WaitingForApi  CustomResourceDefinition/namespacemanagements.argoproj.io
  [15:15:46.457]   ↳ install-fztj2:                 loop[wD+UA] | Installing | rv=170049984 |  11  WaitingForApi  CustomResourceDefinition/notificationsconfigurations.argoproj.io
  [15:15:46.543]   ↳ install-fztj2:                 loop[wD+UA] | Installing | rv=170049984 |  12  WaitingForApi  CustomResourceDefinition/rolloutmanagers.argoproj.io
  [15:15:46.672]   ↳ install-fztj2:                 loop[wD+UA] | Installing | rv=170049984 |  13  WaitingForApi  CustomResourceDefinition/rollouts.argoproj.io
  [15:15:46.870]   ↳ install-fztj2:                 loop[wD+UA] | Installing | rv=170049984 |  14        Created  ConfigMap/openshift-gitops-operator-manager-config
  [15:15:46.962]   ↳ install-fztj2:                 loop[wD+UA] | Installing | rv=170049984 |  15  WaitingForApi  BundleSecret/openshift-gitops-operator-metrics-monitor-bearer-token
  [15:15:47.133]   ↳ install-fztj2:                 loop[wD+UA] | Installing | rv=170049984 |  16        Created  ConfigMap/openshift-gitops-operator-metrics-monitor-ca-bundle
  [15:15:47.388]   ↳ install-fztj2:                 loop[wD+UA] | Installing | rv=170049984 |  17        Created  ServiceMonitor/openshift-gitops-operator-metrics-monitor
  [15:15:47.600]   ↳ install-fztj2:                 loop[wD+UA] | Installing | rv=170049984 |  18        Created  Service/openshift-gitops-operator-metrics-service
  [15:15:47.630]   ↳ install-fztj2:                 loop[wD+UA] | Installing | rv=170049984 |  19        Created  Role/openshift-gitops-operator-prometheus
  [15:15:47.693]   ↳ install-fztj2:                 loop[wD+UA] | Installing | rv=170049984 |  20        Created  RoleBinding/openshift-gitops-operator-prometheus
  [15:15:48.022]   ↳ install-fztj2:                 loop[wD+UA] | Installing | rv=170049984 |  21        Created  Service/openshift-gitops-operator-webhook-service
  [15:15:48.069]   ↳ install-fztj2:                 loop[wD+UA] | Installing | rv=170049984 |  22  WaitingForApi  CustomResourceDefinition/gitopsservices.pipelines.openshift.io
  [15:15:48.160]   ↳ install-fztj2:                 loop[wD+UA] | Installing | rv=170049984 |  23        Created  ServiceAccount/openshift-gitops-operator-controller-manager
  [15:15:48.380]   ↳ install-fztj2:                 loop[wD+UA] | Installing | rv=170049984 |  24        Created  Role/openshift-gitops-operato-8Tpsd2i9EEJe5Zg2xxs8hq8ddOIOxZ3b6gHMye
  [15:15:48.485]   ↳ install-fztj2:                 loop[wD+UA] | Installing | rv=170049984 |  25        Created  RoleBinding/openshift-gitops-operato-8Tpsd2i9EEJe5Zg2xxs8hq8ddOIOxZ3b6gHMye
  [15:15:48.580]   ↳ install-fztj2:                 loop[wD+UA] | Installing | rv=170049984 |  26        Created  ClusterRole/openshift-gitops-operato-9UDesDK9ivdPubzvPPAv7oyajP3WcO2A9TTNdV
  [15:15:48.669]   ↳ install-fztj2:                 loop[wD+UA] | Installing | rv=170049984 |  27        Created  ClusterRoleBinding/openshift-gitops-operato-9UDesDK9ivdPubzvPPAv7oyajP3WcO2A9TTNdV
  [15:15:49.633]   ↳ install-fztj2:                 loop[MGLqT] | Installing | rv=170050106 |   2        Created  CustomResourceDefinition/analysisruns.argoproj.io
  [15:15:49.695]   ↳ install-fztj2:                 loop[MGLqT] | Installing | rv=170050106 |   3        Created  CustomResourceDefinition/analysistemplates.argoproj.io
  [15:15:50.253]   ↳ install-fztj2:                 loop[MGLqT] | Installing | rv=170050106 |   4        Present  CustomResourceDefinition/applications.argoproj.io
  [15:15:51.897]   ↳ install-fztj2:                 loop[MGLqT] | Installing | rv=170050106 |   5        Present  CustomResourceDefinition/applicationsets.argoproj.io
  [15:15:52.151]   ↳ install-fztj2:                 loop[MGLqT] | Installing | rv=170050106 |   6        Present  CustomResourceDefinition/appprojects.argoproj.io
  [15:15:52.597]   ↳ install-fztj2:                 loop[MGLqT] | Installing | rv=170050106 |   7  WaitingForApi  CustomResourceDefinition/argocds.argoproj.io
  [15:15:53.308]   ↳ install-fztj2:                 loop[MGLqT] | Installing | rv=170050106 |   8        Present  CustomResourceDefinition/clusteranalysistemplates.argoproj.io
  [15:15:54.040]   ↳ install-fztj2:                 loop[MGLqT] | Installing | rv=170050106 |   9        Present  CustomResourceDefinition/experiments.argoproj.io
  [15:15:54.277]   ↳ install-fztj2:                 loop[MGLqT] | Installing | rv=170050106 |  10        Present  CustomResourceDefinition/namespacemanagements.argoproj.io
  [15:15:54.394]   ↳ install-fztj2:                 loop[MGLqT] | Installing | rv=170050106 |  11        Present  CustomResourceDefinition/notificationsconfigurations.argoproj.io
  [15:15:54.603]   ↳ install-fztj2:                 loop[MGLqT] | Installing | rv=170050106 |  12        Present  CustomResourceDefinition/rolloutmanagers.argoproj.io
  [15:15:55.198]   ↳ install-fztj2:                 loop[MGLqT] | Installing | rv=170050106 |  13        Present  CustomResourceDefinition/rollouts.argoproj.io
  [15:15:55.443]   ↳ install-fztj2:                 loop[MGLqT] | Installing | rv=170050106 |  14        Present  ConfigMap/openshift-gitops-operator-manager-config
  [15:15:55.575]   ↳ install-fztj2:                 loop[MGLqT] | Installing | rv=170050106 |  15        Created  BundleSecret/openshift-gitops-operator-metrics-monitor-bearer-token
  [15:15:55.775]   ↳ install-fztj2:                 loop[MGLqT] | Installing | rv=170050106 |  16        Present  ConfigMap/openshift-gitops-operator-metrics-monitor-ca-bundle
  [15:15:55.963]   ↳ install-fztj2:                 loop[MGLqT] | Installing | rv=170050106 |  17        Present  ServiceMonitor/openshift-gitops-operator-metrics-monitor
  [15:15:56.500]   ↳ install-fztj2:                 loop[MGLqT] | Installing | rv=170050106 |  18        Present  Service/openshift-gitops-operator-metrics-service
  [15:15:56.677]   ↳ install-fztj2:                 loop[MGLqT] | Installing | rv=170050106 |  19        Present  Role/openshift-gitops-operator-prometheus
  [15:15:56.812]   ↳ install-fztj2:                 loop[MGLqT] | Installing | rv=170050106 |  20        Present  RoleBinding/openshift-gitops-operator-prometheus
  [15:15:57.488]   ↳ install-fztj2:                 loop[MGLqT] | Installing | rv=170050106 |  21        Present  Service/openshift-gitops-operator-webhook-service
  [15:15:57.669]   ↳ install-fztj2:                 loop[MGLqT] | Installing | rv=170050106 |  22        Present  CustomResourceDefinition/gitopsservices.pipelines.openshift.io
  [15:15:57.837]   ↳ install-fztj2:                 loop[MGLqT] | Installing | rv=170050106 |  23        Present  ServiceAccount/openshift-gitops-operator-controller-manager
  [15:15:58.124]   ↳ install-fztj2:                 loop[MGLqT] | Installing | rv=170050106 |  24        Present  Role/openshift-gitops-operato-8Tpsd2i9EEJe5Zg2xxs8hq8ddOIOxZ3b6gHMye
  [15:15:58.689]   ↳ install-fztj2:                 loop[MGLqT] | Installing | rv=170050106 |  25        Present  RoleBinding/openshift-gitops-operato-8Tpsd2i9EEJe5Zg2xxs8hq8ddOIOxZ3b6gHMye
  [15:15:58.868]   ↳ install-fztj2:                 loop[MGLqT] | Installing | rv=170050106 |  26        Present  ClusterRole/openshift-gitops-operato-9UDesDK9ivdPubzvPPAv7oyajP3WcO2A9TTNdV
  [15:15:59.003]   ↳ install-fztj2:                 loop[MGLqT] | Installing | rv=170050106 |  27        Present  ClusterRoleBinding/openshift-gitops-operato-9UDesDK9ivdPubzvPPAv7oyajP3WcO2A9TTNdV
  [15:16:00.116]   ↳ install-fztj2:                 loop[+5JUf] | Installing | rv=170050106 |   2        Created  CustomResourceDefinition/analysisruns.argoproj.io
  [15:16:00.171]   ↳ install-fztj2:                 loop[+5JUf] | Installing | rv=170050106 |   3        Created  CustomResourceDefinition/analysistemplates.argoproj.io
  [15:16:00.850]   ↳ install-fztj2:                 loop[+5JUf] | Installing | rv=170050106 |   4        Present  CustomResourceDefinition/applications.argoproj.io
  [15:16:02.443]   ↳ install-fztj2:                 loop[+5JUf] | Installing | rv=170050106 |   5        Present  CustomResourceDefinition/applicationsets.argoproj.io
  [15:16:03.116]   ↳ install-fztj2:                 loop[+5JUf] | Installing | rv=170050106 |   6        Present  CustomResourceDefinition/appprojects.argoproj.io
  [15:16:04.941]   ↳ install-fztj2:                 loop[+5JUf] | Installing | rv=170050106 |   7        Present  CustomResourceDefinition/argocds.argoproj.io
  [15:16:05.807]   ↳ install-fztj2:                 loop[+5JUf] | Installing | rv=170050106 |   8        Present  CustomResourceDefinition/clusteranalysistemplates.argoproj.io
  [15:16:06.562]   ↳ install-fztj2:                 loop[+5JUf] | Installing | rv=170050106 |   9        Present  CustomResourceDefinition/experiments.argoproj.io
  [15:16:06.842]   ↳ install-fztj2:                 loop[+5JUf] | Installing | rv=170050106 |  10        Present  CustomResourceDefinition/namespacemanagements.argoproj.io
  [15:16:06.979]   ↳ install-fztj2:                 loop[+5JUf] | Installing | rv=170050106 |  11        Present  CustomResourceDefinition/notificationsconfigurations.argoproj.io
  [15:16:07.309]   ↳ install-fztj2:                 loop[+5JUf] | Installing | rv=170050106 |  12        Present  CustomResourceDefinition/rolloutmanagers.argoproj.io
  [15:16:07.753]   ↳ install-fztj2:                 loop[+5JUf] | Installing | rv=170050106 |  13        Present  CustomResourceDefinition/rollouts.argoproj.io
  [15:16:07.962]   ↳ install-fztj2:                 loop[+5JUf] | Installing | rv=170050106 |  14        Present  ConfigMap/openshift-gitops-operator-manager-config
  [15:16:08.185]   ↳ install-fztj2:                 loop[+5JUf] | Installing | rv=170050106 |  15        Present  BundleSecret/openshift-gitops-operator-metrics-monitor-bearer-token
  [15:16:08.591]   ↳ install-fztj2:                 loop[+5JUf] | Installing | rv=170050106 |  16        Present  ConfigMap/openshift-gitops-operator-metrics-monitor-ca-bundle
  [15:16:08.813]   ↳ install-fztj2:                 loop[+5JUf] | Installing | rv=170050106 |  17        Present  ServiceMonitor/openshift-gitops-operator-metrics-monitor
  [15:16:09.415]   ↳ install-fztj2:                 loop[+5JUf] | Installing | rv=170050106 |  18        Present  Service/openshift-gitops-operator-metrics-service
  [15:16:09.526]   ↳ install-fztj2:                 loop[+5JUf] | Installing | rv=170050106 |  19        Present  Role/openshift-gitops-operator-prometheus
  [15:16:09.614]   ↳ install-fztj2:                 loop[+5JUf] | Installing | rv=170050106 |  20        Present  RoleBinding/openshift-gitops-operator-prometheus
  [15:16:10.104]   ↳ install-fztj2:                 loop[+5JUf] | Installing | rv=170050106 |  21        Present  Service/openshift-gitops-operator-webhook-service
  [15:16:10.386]   ↳ install-fztj2:                 loop[+5JUf] | Installing | rv=170050106 |  22        Present  CustomResourceDefinition/gitopsservices.pipelines.openshift.io
  [15:16:10.593]   ↳ install-fztj2:                 loop[+5JUf] | Installing | rv=170050106 |  23        Present  ServiceAccount/openshift-gitops-operator-controller-manager
  [15:16:10.726]   ↳ install-fztj2:                 loop[+5JUf] | Installing | rv=170050106 |  24        Present  Role/openshift-gitops-operato-8Tpsd2i9EEJe5Zg2xxs8hq8ddOIOxZ3b6gHMye
  [15:16:10.949]   ↳ install-fztj2:                 loop[+5JUf] | Installing | rv=170050106 |  25        Present  RoleBinding/openshift-gitops-operato-8Tpsd2i9EEJe5Zg2xxs8hq8ddOIOxZ3b6gHMye
  [15:16:11.157]   ↳ install-fztj2:                 loop[+5JUf] | Installing | rv=170050106 |  26        Present  ClusterRole/openshift-gitops-operato-9UDesDK9ivdPubzvPPAv7oyajP3WcO2A9TTNdV
  [15:16:11.450]   ↳ install-fztj2:                 loop[+5JUf] | Installing | rv=170050106 |  27        Present  ClusterRoleBinding/openshift-gitops-operato-9UDesDK9ivdPubzvPPAv7oyajP3WcO2A9TTNdV
  [15:16:12.728]   ↳ install-fztj2:                 loop[g2kby] | Installing | rv=170050992 |   7        Created  CustomResourceDefinition/argocds.argoproj.io
  [15:16:13.993]   ↳ install-fztj2:                 loop[1yVQR] | Installing | rv=170050992 |   7        Created  CustomResourceDefinition/argocds.argoproj.io
  [15:16:30.056783] openshift-gitops-operator           install=Complete(37s) upgrade=Skipped
  Cleaning up 1 operators in parallel...
  [15:16:30.232733] AUDIT  delete   [200]  subscriptions        openshift-gitops-operator                               by catalog-health-check
  Cleanup done.
```

In the above example reconcile loop 8WfQE and wD+UA got the same ResourceVersion, MGLqT and +5JUf as well (as are the final two loops g2kby and  1yVQR). This might not necessarily be a problem, as we are supposed to be idempotent, but we already saw that it had a side effect (e.g. only a stale RV in a reconcile loop allowed the token secret to actually be re-created and be present after the install. 

Apart from that side effect, we are waisting a lot of API calls and resources on them. I am thinking about storing the last RV number, and if we receive the same, it's stale and ask the RV via a life call instead, should cut down on those unnecessary loops as well. 

And that's another point I want to touch on, a couple of those outputs above with my script I only can get by introducing DEBUG level Logs. I want to keep those in. They should not waste resources or pollute the logs, because by default we do not run with --DEBUG, but for testing operator installs or complete operator catalogs these log lines are invaluable to actually see the step executions and reconcile loops, therefore I would like to keep them in. 

BR
Raimund

tmshort:
@rsacherer I'm not seeing any responses/comments to my comments from 8/14?

pkg/controller/operators/catalog/step.go
tmshort: Has anything come of this?

rsacherer:
@coderabbit review

pkg/controller/operators/catalog/step.go
```diff
@@ -318,3 +328,97 @@ func setInstalledAlongsideAnnotation(a alongside.Annotator, dst metav1.Object, n
...
+		}
+		s.Labels[install.OLMManagedLabelKey] = install.OLMManagedLabelValue
+
+		// Add the resolving CSV as a non-blocking owner so the secret is GC'd on
+		// uninstall. Use a live API call — the CSV may have been created in this
+		// same ExecutePlan invocation and will not yet be in the lister cache.
+		// A lister NotFound here would bubble through apierrors.IsNotFound and
+		// incorrectly trigger the discovery-querier path in ExecutePlan.
+		if step.Resolving != "" {
+			csv, err := b.olmClient.OperatorsV1alpha1().ClusterServiceVersions(namespace).Get(context.TODO(), step.Resolving, metav1.GetOptions{})
```
perdasilva, line 401: I think b.olmClient is an attentuated client (I think it comes from crcClient). We should make sure we're using the same client that getUpdatedOwnerReferences uses. The operator client is basically cluster-admin, and the scoped client uses the service account in the OperatorGroup. If we don't use the same client, we could be introducing a regression here.

pkg/controller/operators/catalog/step.go
```diff
@@ -318,3 +328,97 @@ func setInstalledAlongsideAnnotation(a alongside.Annotator, dst metav1.Object, n
...
+			if err != nil {
+				return v1alpha1.StepStatusUnknown, fmt.Errorf("error getting csv %s for secret owner ref: %w", step.Resolving, err)
+			}
+			ownerutil.AddNonBlockingOwner(&s, csv)
+		}
+
+		// Refresh UIDs on any pre-existing CSV owner refs shipped in the bundle
+		// manifest so Kubernetes GC can match them on uninstall.
+		updated, err := refreshCSVOwnerRefUIDs(s.OwnerReferences, b.olmClient, namespace)
```
perdasilva: does `refreshCSVOwnerRefUIDs` somehow duplicate the live CSV get call on line 387 + 391?

PR #3904, opened 2026-08-27
reconciler: fix correctImages panic when pod has no containers

sebrandon1:
## Summary

- `correctImages` accessed `pod.Spec.Containers[0]` without a length guard in the non-`ExtractContent` path, causing a panic for pods with an empty containers slice (e.g. evicted or malformed pods)
- The `ExtractContent` path already had an equivalent `len(pod.Spec.Containers) != 1` guard; this brings the fallback path in line
- Adds `TestCorrectImages` covering both paths, including the empty-containers regression case and an `ExtractContent` serving-image mismatch case

## Test plan

- [ ] `go test ./pkg/controller/registry/reconciler/... -run TestCorrectImages` — all 6 cases pass

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented reconciliation errors when evaluating workloads with no regular containers.
  - Improved image validation across standard and content-extraction configurations.

- **Tests**
  - Added coverage for empty-container scenarios, matching and mismatched images, container layouts, and invalid configurations.

### DIFF
--- a/pkg/controller/registry/reconciler/grpc.go
+++ b/pkg/controller/registry/reconciler/grpc.go
@@ -273,6 +273,9 @@ func correctImages(source grpcCatalogSourceDecorator, pod *corev1.Pod) bool {
 			pod.Spec.InitContainers[1].Image == source.CatalogSource.Spec.Image &&
 			pod.Spec.Containers[0].Image == source.opmImage
 	}
+	if len(pod.Spec.Containers) == 0 {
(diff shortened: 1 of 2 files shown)


grokspawn (approved):

rashmigottipati (commented):
happened to review post merge, lgtm

sebrandon1:
Thanks!!!

PR #3910, opened 2026-09-01
decorators: replace init() panics with lazy error propagation

sebrandon1:
## Summary
- Replace `init()` panics with `sync.Once` lazy initialization for jq query compilation in `decorators/operator.go`
- Failures now surface as reconciliation errors in `Component.Reference()` instead of crashing the operator process at startup
- Add direct test for `jqQueries()` verifying successful compilation and `sync.Once` idempotency

## Test Plan
- [x] `go test ./pkg/controller/operators/decorators/...` — 9/9 pass
- [x] `go vet` clean
- [ ] CI passes

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when initializing component and CSV query processing.
  * Query parsing errors are now reported gracefully instead of causing startup failures.
  * Component condition extraction now properly handles initialization errors.

* **Tests**
  * Added coverage to verify successful query initialization and consistent reuse of initialized queries.

### DIFF
--- a/pkg/controller/operators/decorators/operator.go
+++ b/pkg/controller/operators/decorators/operator.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"sort"
 	"strings"
+	"sync"
 
 	"github.com/go-viper/mapstructure/v2"
 	"github.com/itchyny/gojq"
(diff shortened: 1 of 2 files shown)


rashmigottipati (commented):
/ok-to-test

tmshort:
I'm concerned that the failure of compiling jq queries is really a coding error, and not a run-time error. There is nothing we can do on a failure to recover. Failing in `init()` makes sense as it tells the developer immediately that there's a coding error. Continuing despite the error gives a false sense of security to the developer.

In other words, there's no way to recover, so why not fail immediately?

sebrandon1:
Good point, generally I try to not use `Init()` as much as I can just because I don't like logic firing without being explicitly called but I can see what you're getting at.  Also it let me add unit tests for that JQ parsing.

I'm good either way if you prefer the original Init() I can just close it.  Thanks @tmshort!

pkg/controller/operators/decorators/operator.go
```diff
@@ -32,19 +33,26 @@ const (
...
 
-func init() {
-	var err error
-	if componentConditionsJQ, err = gojq.Parse(".status.conditions"); err != nil {
-		panic(fmt.Errorf("failed to parse component conditions jq: %s", err))
-	}
-	if csvConditionsJQ, err = gojq.Parse(".status | [{\"type\": .phase, \"status\": \"True\", \"reason\": .reason, \"message\": .message, \"lastUpdateTime\": .lastUpdateTime,\"lastTransitionTime\": .lastTransitionTime}]"); err != nil {
-		panic(fmt.Errorf("failed to parse csv conditions jq: %s", err))
-	}
+func jqQueries() (componentQ, csvQ *gojq.Query, err error) {
```
perdasilva, line 44: would it be possible to avoid named returned parameters?

sebrandon1: Happy to change it if you'd prefer, but named returns show up a fair bit in this package already — including on `Reference()`, which calls `jqQueries()`:

- [`OperatorNames`](https://github.com/operator-framework/operator-lifecycle-manager/blob/3f56d52b5cbe4f8f468864f0734d505db526a13e/pkg/controller/operators/decorators/operator.go#L51)
- [`AdoptComponent`](https://github.com/operator-framework/operator-lifecycle-manager/blob/3f56d52b5cbe4f8f468864f0734d505db526a13e/pkg/controller/operators/decorators/operator.go#L236)
- [`DisownComponent`](https://github.com/operator-framework/operator-lifecycle-manager/blob/3f56d52b5cbe4f8f468864f0734d505db526a13e/pkg/controller/operators/decorators/operator.go#L263)
- [`Reference`](https://github.com/operator-framework/operator-lifecycle-manager/blob/3f56d52b5cbe4f8f468864f0734d505db526a13e/pkg/controller/operators/decorators/operator.go#L407)

Same pattern elsewhere too, e.g. [`NewSyncer`](https://github.com/operator-framework/operator-lifecycle-manager/blob/3f56d52b5cbe4f8f468864f0734d505db526a13e/pkg/lib/proxy/syncer.go#L26) and [`SyncOperatorGroup`](https://github.com/operator-framework/operator-lifecycle-manager/blob/3f56d52b5cbe4f8f468864f0734d505db526a13e/pkg/lib/scoped/syncer.go#L53).

If you'd still rather keep this one unnamed, easy enough to drop them.

perdasilva: yeah - its all pretty old code from before people realized  they could introduce some hard to find bugs XD I've had to chase one of those once in this code-base myself. If its all the same, let's drop the named return parameters. Thanks heaps =D

sebrandon1: Done!  ✅

PR #3915, opened 2026-09-13
OCPBUGS-78355: add cpu limits to catalogsource pods

ankitathomas:
On startup, catalogsource pods cause CPU spikes which can degrade node health on clusters. This PR adds resource limits to CPU cores for catalogsource pods to avoid performance degradation. CPU limit set to 2 cores since openshift nodes typically need 4 cores for multi node clusters and 8 cores for single node openshift.

## Summary by CodeRabbit

- **Bug Fixes**
  - Catalog-building pods now have a CPU limit of 2 cores.
  - This helps prevent CPU spikes during startup and reduces the risk of impacting other workloads on the same node.
  - Standard and extracted catalog pod configurations now consistently apply the same resource limit across supported security-context scenarios.


perdasilva:
Hey @ankitathomas could you please fix up the formatting of `pkg/controller/registry/reconciler/reconciler.go`

```
pkg/controller/registry/reconciler/reconciler.go:195:1  gofmt  File is not properly formatted
```

perdasilva:
@ankitathomas I've fixed up the formatting issue - please pull down the latest changes before making any further changes =D (I hope that's ok)

rashmigottipati (approved):
/lgtm

grokspawn:
I'm reverting this, since as a rule we do not limit control plane resources as the limits do not always make sense in different topologies/capacities, and we generally don't make vendor-specific decisions in upstream.

grokspawn:
A better potential solution for the raised issue (package-server cache update caused CPU spike) might be https://github.com/operator-framework/operator-lifecycle-manager/pull/3920

PR #3920, opened 2026-09-16
fix(packageserver): bound and stagger catalog refresh work

stevekuznetsov:
Unbounded package refresh fan-out can overwhelm resource-constrained catalogs. Share a configurable worker limit across refreshes and jitter queued catalog notifications to reduce synchronized request bursts.

Preserve cached packages on interrupted streams, honor cancellation, and report refresh completion. Add concurrency, scheduling, and flag validation tests and document the defaults and unchanged RPC timeout.

## Summary by CodeRabbit

* **New Features**
  * Added configurable package refresh concurrency with `--package-refresh-workers` (default: 4; range: 1–128).
  * Added configurable catalog refresh jitter with `--catalog-refresh-jitter` (default: 0.2; range: 0–1) to stagger periodic and follow-up refreshes.
  * Catalog refreshes now process packages concurrently within the configured limit and respond to cancellation.
  * Invalid refresh settings are rejected before the package server starts.

* **Documentation**
  * Documented refresh worker, jitter, timing, defaults, and command-line configuration in the README.


stevekuznetsov:
E2E failure seems to be some catalog operator flake wrt the ConfigMap, not related to this change. I don't have permission to re-try the job.

pkg/package-server/provider/registry.go
```diff
@@ -278,7 +341,7 @@ func getOperatorDeprecation(in *api.Deprecation) *operators.Deprecation {
 	}
 }
 
-func (p *RegistryProvider) refreshCache(ctx context.Context, client *registryClient) error {
+func (p *RegistryProvider) refreshCache(ctx context.Context, client *registryClient) (result error) {
```
perdasilva, line 344: could we avoid the named return parameter here?

pkg/package-server/provider/registry.go
```diff
@@ -155,11 +189,47 @@ func NewRegistryProvider(ctx context.Context, crClient versioned.Interface, oper
...
+	if err != nil {
+		return nil, err
+	}
 	p.catsrcLister = catsrcInformer.Lister()
+	logrus.WithFields(logrus.Fields{"workers": workers, "resyncInterval": wakeupInterval, "jitter": jitter, "rpcTimeout": cacheTimeout}).Info("configured catalog refresh")
 
 	return p, nil
 }
 
+func (p *RegistryProvider) enqueueRefresh(key registry.CatalogKey, resync bool) {
```
perdasilva, line 215: nit
```suggestion
func (p *RegistryProvider) enqueueRefresh(key registry.CatalogKey, isResync bool) {
```

pkg/package-server/provider/registry.go
```diff
@@ -155,11 +189,47 @@ func NewRegistryProvider(ctx context.Context, crClient versioned.Interface, oper
...
+	_, err = catsrcInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			source := obj.(*operatorsv1alpha1.CatalogSource)
+			p.refreshQueue.Add(types.NamespacedName{Namespace: source.Namespace, Name: source.Name})
+		},
+		UpdateFunc: func(old, current interface{}) {
+			before, source := old.(*operatorsv1alpha1.CatalogSource), current.(*operatorsv1alpha1.CatalogSource)
+			p.enqueueRefresh(registry.CatalogKey{Namespace: source.Namespace, Name: source.Name}, before.ResourceVersion == source.ResourceVersion)
+		},
+		DeleteFunc: p.catalogSourceDeleted,
```
perdasilva, line 204: Does it matter that we're setting `queueinformer.WithDeletionHandler(p.catalogSourceDeleted)` on the queue informer (line 184)? From memory I think this is a no-op since we're no longer setting the informer. But from a maintenance perspective, I'm not sure if it could be a source of confusion or catalogSourceDeleted to be called twice if an informer ever gets set there again...thoughts?

pkg/package-server/provider/registry.go
perdasilva, line 245: should we add a return after the logging the error?

pkg/package-server/provider/registry.go
```diff
@@ -347,45 +419,60 @@ func (p *RegistryProvider) refreshCache(ctx context.Context, client *registryCli
...
 
 			mu.Lock()
 			defer mu.Unlock()
 			added[newPkg.GetName()] = struct{}{}
+			completed.Add(1)
 		}()
 	}
 
 	logger.Debug("caching new packages...")
 	wg.Wait()
```
perdasilva, line 471: the double wg.Wait could be a source of clarity issues. Would it make sense to wrap the for loop in its own function stanza an make the whole thing synchronous returning an error?

